PR #35323, opened 2026-04-15
fix(reindex): eliminate TOCTOU race in ReindexThread BulkProcessor rebuild

### LINKED ISSUE #35313: fix(ReindexThread): TOCTOU race on BulkProcessor rebuild — use AtomicReference + compareAndSet
## Description

`ReindexThread.java` lines 243–248 contains a pre-existing TOCTOU (Time-Of-Check To Time-Of-Use) race condition on the bulk-processor rebuild path.

```java
if (bulkProcessor == null || rebuildBulkIndexer.get()) {   // CHECK
    closeBulkProcessor(bulkProcessor);
    bulkProcessorListener = new BulkProcessorListener();
    bulkProcessor = indexAPI.createBulkProcessor(bulkProcessorListener); // USE
}
bulkProcessorListener.workingRecords.putAll(workingRecords); // line 248
```

Two threads racing through the null-check can both create a new processor. The first processor is discarded, but any in-flight `afterBulk` callbacks still hold a reference to the old `bulkProcessorListener` whose `workingRecords` has already been replaced — potentially losing reindex records silently.

Additionally, `bulkProcessorListener.workingRecords.putAll(workingRecords)` at line 248 runs on the `ReindexThread` main loop while `afterBulk` callbacks can call `workingRecords.get()` on the processor's internal thread simultaneously.

**Identified during review of PR #35289 by @mbiuki.** Left out of scope there because it affects pre-existing code not changed by that PR.

## Acceptance Criteria

- [ ] `bulkProcessor` is held in an `AtomicReference<BulkProcessor>` and replaced via `compareAndSet` to make the check+create atomic
- [ ] `bulkProcessorListener` assignment is coordinated with the processor swap so no callback can observe a mismatched listener
- [ ] Existing reindex integration tests pass (no `ConcurrentModificationException`, no silent record loss)
- [ ] A unit or integration test covers the concurrent-rebuild scenario (or a comment documents why the race is now impossible)

## Additional Context

- Identified in: `dotCMS/src/main/java/com/dotmarketing/common/reindex/ReindexThread.java` lines 243–248
- Related PR: dotCMS/core#35289
- The companion fix (`workingRecords` → `ConcurrentHashMap`, `contentletsIndexed` → `volatile`) was already applied in commit `3fd612924c`

fabrizzio-dotCMS:
## Summary

Fixes the TOCTOU (Time-Of-Check To Time-Of-Use) race condition in `ReindexThread.runReindexLoop()` identified in issue #35313.

### Root cause

Two concurrent hazards on lines 243–248 of `ReindexThread.java`:

**1. Null-check / rebuild not atomic (`rebuildBulkIndexer`)**
```java
if (bulkProcessor == null || rebuildBulkIndexer.get()) {  // CHECK
    closeBulkProcessor(bulkProcessor);
    bulkProcessorListener = new BulkProcessorListener();
    bulkProcessor = indexAPI.createBulkProcessor(bulkProcessorListener); // USE
}
```
An external thread calling `rebuildBulkIndexer()` between the check and the creation can leave callbacks pointing to a stale listener, silently losing reindex records.

**2. `putAll` vs `afterBulk` callback race**
```java
bulkProcessorListener.workingRecords.putAll(workingRecords); // ReindexThread
// meanwhile: afterBulk() reads workingRecords on BulkProcessor callback thread
```
In back-to-back loop iterations, `putAll(batch N+1)` can interleave with `afterBulk(batch N)` reading `workingRecords`, causing `deleteReindexEntry`/`markAsFailed` to resolve entries from the wrong batch.

### Fix

- Create a fresh `BulkProcessorListener` and `IndexBulkProcessor` **per batch** via try-with-resources
- `close()` blocks until `afterBulk` completes before the next batch starts — each batch owns its immutable `workingRecords` snapshot
- `rebuildBulkIndexer` flag, `rebuildBulkIndexer()` method, and `closeBulkProcessor()` helper removed (all made unnecessary by per-batch approach)
- `finalizeReIndex()` simplified: no longer manages a processor parameter
- `BulkProcessorListener.afterBulk`: logs warning on >50% failure rate instead of signalling a rebuild

### Scope

Only 2 files changed (`ReindexThread.java` + `BulkProcessorListener.java`), net **-33 lines**.

## Test plan

- [ ] Existing `ReindexThread` integration tests pass
- [ ] `OSCreateContentIndexIntegrationTest` / `OpenSearchUpgradeSuite` pass
- [ ] Manual concurrent-reindex: trigger `rebuildBulkIndexer()` while reindex is in flight — no `ConcurrentModificationException`, no silent record loss
- [ ] Full reindex switchover still works: queue empties → `finalizeReIndex` → switchover → notification

Closes #35313


dotCMS/src/main/java/com/dotmarketing/common/reindex/ReindexThread.java
```diff
@@ -24,16 +30,12 @@
...
-import com.dotcms.content.index.domain.IndexBulkProcessor;
-
 import java.sql.SQLException;
 import java.time.Duration;
 import java.util.Map;
 import java.util.concurrent.ThreadPoolExecutor;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
+import org.apache.felix.framework.OSGISystem;
```
wezell, line 38: why this import?

dario-daza (approved):

dotCMS/src/main/java/com/dotmarketing/common/reindex/ReindexThread.java
fabrizzio-dotCMS: The import was already there. 
it was pushed down, so it looks like a brand new line was added. Looks like dead code to me

nollymar (approved):

PR #35342, opened 2026-04-15
fix(index): clean up indicies DB rows on all physical delete paths (DO NOT MERGE)

### LINKED ISSUE #35306: [BUG] DELETE /api/v1/esindex/{name} does not clean up indicies table — orphaned DB rows
## Summary

`DELETE /api/v1/esindex/{indexName}` physically removes the index from the search provider (ES or OS) but does **not** delete the corresponding row from the `indicies` database table.

After deletion, the `indicies` table retains stale entries pointing to non-existent indices.

## Observed behavior

1. `DELETE /api/v1/esindex/working_20260414033602` → HTTP 200, index physically deleted from OS
2. `SELECT * FROM indicies;` → row `working_20260414033602` still present
3. On next startup: dotCMS attempts to activate a non-existent index → potential startup failure or silent inconsistency

## Affected providers

Confirmed for both ES and OS providers across all phases.

## Root cause

`ContentletIndexAPIImpl.delete()` calls the physical delete on the search provider but does not follow up with a call to `IndiciesAPI.removeIndiciesByName()` (or equivalent) to remove the DB record.

## Fix

After successful physical index deletion in `ContentletIndexAPIImpl.delete()`, delete the corresponding row from the `indicies` table:

```java
// After successful physical delete:
APILocator.getIndiciesAPI().removeIndiciesByName(indexName);
// or for versioned indices:
versionedIndicesAPI.removeIndicesByName(indexName);
```

Ensure the cleanup handles both `legacyIndiciesAPI` (ES, non-versioned) and `versionedIndicesAPI` (OS, versioned) stores.

## Discovered via

QA test run — TC-06 (Phase 3) and DELETE operations across phases, `dotcms/dotcms:latest` with `single-node-os-migration` stack.

## Related

- Refs #34933
- PR #35289

fabrizzio-dotCMS:
## Summary

Fixes #35306 — `ContentletIndexAPIImpl.delete()` physically removed the index from the search provider but never deleted the corresponding row from the `indicies` table.

After a successful physical delete, a stale DB entry would remain pointing to a non-existent index, which could cause startup failures or silent inconsistencies on the next boot.

## Root cause

The bug only occurs when `delete()` is called on an **active** index without a prior `deactivateIndex()` call (e.g., direct REST call to `DELETE /api/v1/esindex/{name}`). In the normal UI workflow, `deactivateIndex()` clears the DB row via `point()` before `delete()` is called, so no orphan is left.

## Fix

After a successful physical delete, `versionedIndicesAPI.removeByIndexName()` is called to clean up the `indicies` row.

**Phase safety:**
- **Physical delete**: `indexAPI` resolves to `IndexAPIImpl` (the phase-aware router) — fans out to ES, OS, or both depending on the active migration phase.
- **DB cleanup**: `versionedIndicesAPI.removeByIndexName()` issues `DELETE FROM indicies WHERE index_name = ? OR index_name = ?` (bare + `os::` tagged forms), covering both legacy non-versioned rows (ES, phases 0/1/2) and versioned OS rows (phase 3) in a single call. The deprecated `legacyIndiciesAPI` is not involved.

## Test plan

- [ ] `DELETE /api/v1/esindex/{name}` on an active index → row removed from `indicies` table
- [ ] `DELETE /api/v1/esindex/{name}` on an already-deactivated index → no error (row already gone, `removeByIndexName` is a no-op)
- [ ] Verify across all phases (0, 1, 2, 3)

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImpl.java
+++ b/dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImpl.java
@@ -1285,8 +1285,17 @@ public class ContentletIndexAPIImpl implements ContentletIndexAPI {
 
     }
 
-    public boolean delete(String indexName) {
-        return indexAPI.delete(indexName);
+    public boolean delete(final String indexName) {
+        final boolean deleted = indexAPI.delete(indexName);
+        if (deleted) {
+            try {
+                versionedIndicesAPI.removeByIndexName(indexName);
+            } catch (DotDataException e) {
+                Logger.error(this, "Index [" + indexName + "] was physically deleted but the " +
+                        "indices DB row could not be removed: " + e.getMessage(), e);
+            }
+        }
+        return deleted;
     }
 
     public boolean optimize(List<String> indexNames) {
--- a/dotCMS/src/main/java/com/dotcms/content/index/VersionedIndicesAPIImpl.java
+++ b/dotCMS/src/main/java/com/dotcms/content/index/VersionedIndicesAPIImpl.java
@@ -229,6 +229,23 @@ public class VersionedIndicesAPIImpl implements VersionedIndicesAPI {
     /**
      * {@inheritDoc}
      */
+    @WrapInTransaction
+    @Override
+    public void removeByIndexName(final String indexName) throws DotDataException {
+        Logger.debug(this, "Removing indices row for index: " + indexName);
+        indicesFactory.removeByIndexName(indexName);
+        // Flush all index-related caches so no stale names survive the deletion:
+        // 1. VersionedIndicesCache — our own versioned-index cache
+        cache.clearCache();
+        // 2. IndiciesCache — legacy (ES, non-versioned) index cache used by IndiciesFactory
+        CacheLocator.getIndiciesCache().clearCache();
+        // 3. ESQueryCache — cached search queries that may reference the deleted index name
+        CacheLocator.getESQueryCache().clearCache();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
     public void clearCache() {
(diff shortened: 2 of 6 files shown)


dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImpl.java
```diff
@@ -1285,8 +1285,17 @@ private void logSwitchover(final IndiciesInfo oldInfo, final String luckyServer)
 
     }
 
-    public boolean delete(String indexName) {
-        return indexAPI.delete(indexName);
+    public boolean delete(final String indexName) {
+        final boolean deleted = indexAPI.delete(indexName);
```
jsanca, line 1289: I am not sure if this action could be consider thread-safe in a cluster you may consider use the ShedLock, there is an implementation on dotCMS already

dotCMS/src/main/java/com/dotcms/content/index/VersionedIndicesAPIImpl.java
```diff
@@ -226,6 +226,23 @@ public Optional<VersionedIndices> loadDefaultVersionedIndices() throws DotDataEx
...
+     * {@inheritDoc}
+     */
+    @WrapInTransaction
+    @Override
+    public void removeByIndexName(final String indexName) throws DotDataException {
+        Logger.debug(this, "Removing indices row for index: " + indexName);
+        indicesFactory.removeByIndexName(indexName);
+        // Flush all index-related caches so no stale names survive the deletion:
+        // 1. VersionedIndicesCache — our own versioned-index cache
+        cache.clearCache();
```
jsanca, line 239: Usually this things are in a commit listener as far as I recall

PR #35348, opened 2026-04-16
fix(opensearch): ensure OS indices always created with custom analyzers — fixes #35305

### LINKED ISSUE #35305: [BUG] putMapping fails with HTTP 400 on OpenSearch 3.4.0 — ES mapping format incompatibility
## Summary

When dotCMS boots in Phase 3 (OS only), all content-type `putMapping` calls to OpenSearch 3.4.0 fail with **HTTP 400**. As a result, 0 documents are indexed in OS despite the indices being physically created.

## Observed behavior

- OS indices are created successfully at startup
- All `putMapping` calls return HTTP 400 from OS
- Working and live indices remain at 0 documents
- `GET /api/v1/esindex/` shows OS indices with `health: null`, `docs: 0`

## Root cause

The mapping JSON generated by `ESMappingAPIImpl` / `MappingHelper` targets Elasticsearch 7.x format. `IndexMappingRestOperationsOS` (or `MappingOperationsOS.putMapping()`) passes this payload unchanged to OpenSearch 3.4.0, which rejects it.

Known incompatibilities between ES 7.x and OS 3.x mapping formats include:
- `_type` field mappings (removed in OS 3.x)
- `include_type_name` parameter handling
- Some field type aliases

## Fix

Audit the mapping JSON produced by `MappingHelper` / `ESMappingAPIImpl.getMapping()` for OS 3.4.0 compatibility. Add required format transformations in the OS-specific mapping path (`IndexMappingRestOperationsOS`) before submitting to OS.

Start by capturing the exact 400 response body from OS to identify which fields are being rejected.

## Impact

- Phase 3 (OS only) is completely non-functional — no content can be indexed
- Blocks full migration validation

## Discovered via

QA test run — Phase 3 boot against `dotcms/dotcms:latest` with `single-node-os-migration` stack (OS 3.4.0).

## Related

- Refs #34933
- PR #35289

fabrizzio-dotCMS:
## Summary

- **`OSIndexAPIImpl.getDefaultIndexSettings()`**: was loading `opensearch-content-settings.json` (a file that does not exist), silently falling back to minimal hardcoded settings without the `analysis` block. This caused `clearIndex()` to recreate OS indices without `my_analyzer`, making all subsequent `addCustomMapping()` `putMapping` calls fail with HTTP 400 on OS 3.4.0. Changed to load `os-content-settings.json` (the file that actually exists in the classpath).
- **`ContentletIndexOperationsOS.createContentIndex()`**: was swallowing the `IOException` from the settings file load and continuing with `settings = null`, allowing index creation without custom analyzers. The silent `catch` is removed so the failure propagates before any `putMapping` call is made against an under-configured index.

## Root cause

`addCustomMapping` generates explicit field-level `"analyzer":"my_analyzer"` mappings for text fields (`TextField`, `TextAreaField`, `WysiwygField`, etc.). OS 3.4.0 rejects these with HTTP 400 when the index was created without defining `my_analyzer` in its analysis settings:

```
{"error":{"reason":"analyzer [my_analyzer] has not been configured in mappings"},"status":400}
```

## Test plan

- Start the `os-migration` Docker stack (OS 3.4.0 at `localhost:9201`)
- Run the reproduction steps from the [issue comment](https://github.com/dotCMS/core/issues/35305#issuecomment-4261162837) to confirm HTTP 200 on all `putMapping` calls when the index is created with `os-content-settings.json`
- Verify `OSIndexAPIImpl.getDefaultIndexSettings()` now returns settings that include the `analysis` block
- Trigger a `clearIndex()` via admin UI and confirm the recreated index has `my_analyzer` configured (`GET /{index}/_settings`)

This PR fixes: #35305


fabrizzio-dotCMS:
Closing — branch was cut from main instead of master. Recreating correctly.

PR #35355, opened 2026-04-16
refactor(style-editor): restructure style editor schema handling and …

### LINKED ISSUE #35270: Unify Style Editor schema definition: use content type tab as single source of truth for headless and traditional pages
## Description

Currently, the Style Editor supports two **Style Editor Schema** definition paths — **Headless** and **Traditional** — but schema definition works differently for each:

| | Headless | Traditional |
|---|---|---|
| Schema defined in | Code generated schema; uses `registerStyleEditorSchema` in the FE | Content Type tab UI (behind `FEATURE_FLAG_UVE_STYLE_EDITOR`) uses `registerStyleEditorSchema` in the BE |

This dual approach creates inconsistency and maintenance overhead. The goal is to establish a **single source of truth** for Style Editor schema definition: the **content type tab** visual approach.

### How it works today

The schema defined via the Content Type tab is saved in the **Content Type metadata**. Both headless and traditional pages already have access to this metadata. `registerStyleEditorSchema` is the internal mechanism used in both modes to register the schema at runtime.

### What needs to change

The Style Editor schema resolution logic should:

1. **Always prefer** the schema defined in the Content Type tab (stored in Content Type metadata) — for both headless and traditional pages.
2. **Create a new endpoint** to retrieve the Style Editor Schemas defined for Headless approaches, the it requieres an auth token in order to work correctly.
3. **Update the "no schema" documentation/message** to reference the content type tab as the unified approach.

The feature flag `FEATURE_FLAG_UVE_STYLE_EDITOR_FOR_TRADITIONAL_PAGES` has been deleted and now we are using just the `FEATURE_FLAG_UVE_STYLE_EDITOR` to enable/disable the Style Editor for Traditional and Headless.

### Priority behavior

```
Content type tab schema  →  wins (highest priority)
Code-based schema        →  moved to internal usage now the only source of truth are the schemas defined in the Content Types tab
```

## Acceptance Criteria

- [ ] Schema defined via the content type tab is used as the Style Editor schema for both headless and traditional pages that use that content type.
- [ ] Style Editor end-user behavior (rendering, editing experience) is unchanged in both headless and traditional pages after this change.
- [ ] Headless pages whose content type has a schema defined in the content type tab render Style Editor using that schema.
- [ ] The existing "no schema configured" information message/documentation is updated to reference the content type tab as the unified way to define Style Editor schemas.
- [ ] The feature flag `FEATURE_FLAG_UVE_STYLE_EDITOR` is the only one FF that gates the Traditional and Headless Style Editor experience.

## Additional Context

- The new endpoint that shows up the schema definition for Headless needs an Auth token.
- Internal schema registration mechanism (`registerStyleEditorSchema`) is not removed — it is still used internally to register schemas from either source.

KevinDavilaDotCMS:
…remove unused hooks

- Introduced  for normalizing style editor forms.
- Updated  and  to register style editor schemas.
- Removed  hook and associated tests to streamline the codebase.
- Added new types for style editor fields and schemas in .
- Updated various components to utilize the new schema handling, improving maintainability and clarity.

This refactor enhances the style editor's integration and prepares for future feature expansions.

https://github.com/user-attachments/assets/db704465-d8fc-4492-a809-170553c9fdd8

This PR fixes: #35270

This PR fixes: #35270


core-web/libs/sdk/client/src/lib/client/page/page-api.ts
```diff
@@ -183,6 +189,13 @@ export class PageClient extends BaseApiClient {
                 });
             }
 
+            const styleEditorSchemas = await fetchStyleEditorSchemas(
+                pageResponse.page.identifier,
+                this.config,
+                this.requestOptions,
+                this.httpClient
+            );
```
rjvelazco, line 197: Maybe no for this PR but I think we should include this in Graphql

core-web/libs/sdk/uve/src/lib/style-editor/internal.ts
```diff
@@ -224,3 +227,46 @@ export function normalizeForm(form: StyleEditorForm): StyleEditorFormSchema {
...
+        return true;
+    });
+
+    sendMessageToUVE({
+        action: DotCMSUVEAction.REGISTER_STYLE_SCHEMAS,
+        payload: {
+            schemas: validatedSchemas
+        }
+    });
+}
```
rjvelazco, line 272: Is this still needed, @dario-daza ?

core-web/libs/portlets/edit-ema/portlet/src/lib/store/features/flags/withFlags.ts
```diff
@@ -26,7 +26,17 @@ export function withFlags(flags: FeaturedFlags[]) {
...
+                    .pipe(
+                        take(1),
+                        map((rawFlags) =>
+                            Object.fromEntries(
+                                Object.entries(rawFlags).map(([key, value]) => [
+                                    key,
+                                    value === true || value === FEATURE_FLAG_NOT_FOUND
+                                ])
+                            )
+                        )
```
rjvelazco, line 41: Shoud this go here?

rjvelazco (changes requested):

core-web/libs/sdk/client/src/lib/client/page/page-api.ts
dario-daza: Agree, including schemas in the GraphQL response would remove the extra round-trip. It applies also for the PageAPI response, we might not need another endpoint, but I'll review that in another PR.

core-web/libs/sdk/uve/src/lib/style-editor/internal.ts
dario-daza: Yes, it's needed. The schemas are fetched from the server by the SDK, but UVE doesn't have them yet until we call the `registerStyleEditorSchemas`, it is the bridge that delivers them to the UVE editor runtime so the style panel activates.

core-web/libs/portlets/edit-ema/portlet/src/lib/store/features/flags/withFlags.ts
dario-daza: yes, It's the single point where all UVE flags are fetched and stored, so make the validation to tell if a FF is `true` or `NOT_FOUND` means that the FF is enabled, and any other value means that the FF if disabled.

I included a comment clarifying the functionality of FF.

core-web/libs/sdk/uve/src/lib/style-editor/internal.ts
rjvelazco: Weren't we planning only leaving the Style editor for builder as the source of truth? 

cc: @dario-daza @KevinDavilaDotCMS @fmontes

rjvelazco (approved):

core-web/libs/sdk/uve/src/lib/style-editor/internal.ts
fmontes: now.

rjvelazco:
@claude check frontent changes again

core-web/libs/sdk/uve/src/lib/style-editor/internal.ts
dario-daza: @rjvelazco @fmontes @KevinDavilaDotCMS I was a bit confused regarding this so I ask claude to give me an explanation and this is what I understand:

The function is still needed, but for a different reason than before. The source of schemas changed (now they all come from the BE) instead of developer-defined code. But `registerStyleEditorSchemas` isn't about schema definition anymore; it's the postMessage bridge that carries those schemas from the headless app's iframe into the UVE editor's parent window. Without it, `fetchStyleEditorSchemas` fetches the schemas correctly but they stay trapped in the iframe's memory and never reach uveStore.setStyleSchemas(), so the style panel never activates. The communication mechanism is still required even if the source changed.

PR #35363, opened 2026-04-16
feat(graphql): unify Binary, File, and Image field queries via DotBinaryLike #34540

### LINKED ISSUE #34540: GraphQL Support for dotAssets and FileAssets in Image/File Fields
## Description

Currently, when querying content with asset relationships through GraphQL, there are critical issues affecting both dotAssets and FileAssets:

1. **User created fields not accessible** - User created fields on dotAssets and FileAssets cannot be queried through GraphQL, blocking AI tagging workflows and rich metadata access.

2. **Incomplete GraphQL schema** - The DotAsset type definition is missing 20+ properties that are available via `_map`, including: `fileName`, `publishDate`, `mimeType`, `size`, `extension`, `url`, `creationDate`, `name`, `shortyId`, and others. This breaks GraphQL's self-documenting nature and forces developers to rely on the `_map` escape hatch.

- The GraphQL schema for DotAsset exposes all available properties through schema introspection
- Custom fields defined on dotAsset and FileAsset content types are queryable via GraphQL
- Developers can build headless applications without N+1 query patterns or `_map` workarounds

## Target Personas

- [x] Developer teams

## Links

- [Full Context](https://docs.google.com/document/d/1lvxHLTS9xeBrIXWpPPgR8JM7W0ns1g9X_Qq9gPrYtXs/edit?tab=t.f5v2xy55497i)

nollymar:
## Summary

Refs dotCMS/core#34540.

Today `BinaryField` maps to `DotBinary` and `FileField`/`ImageField` map to `DotFileasset` in the GraphQL schema. The two object types expose completely different sub-fields, so a client cannot write a uniform query across the three dotCMS field types:

```graphql
mybinary { versionPath name size mime }   # works
myfile   { versionPath name size mime }   # fails — fields aren't on DotFileasset
myimage  { versionPath name size mime }   # fails — same
```

This PR introduces a shared **`DotBinaryLike`** GraphQL interface implemented by both `DotBinary` and `DotFileasset`. It declares the 12 common binary properties:

`name, title, size, mime, versionPath, idPath, path, sha256, isImage, width, height, modDate`

- **`DotBinary`** — resolves them from the binary map already produced by `BinaryToMapTransformer`. **No behavior change.**
- **`DotFileasset`** — new `FileAssetBinaryPropertyDataFetcher` runs the same transformer against the underlying FileAsset/DotAsset `Contentlet`, so the values returned are identical to those a client gets from the equivalent `mybinary` query.

Type-specific extras stay on the concrete type:
- `DotBinary` keeps `focalPoint`.
- `DotFileasset` keeps `fileName, description, fileAsset, metaData, showOnMenu, sortOrder`.

A `TypeResolver` on the interface dispatches on the source Java type: `Map` → `DotBinary`, `Contentlet` → `DotFileasset`.

### Backward compatibility
- All existing `DotBinary` queries: unchanged.
- All existing `DotFileasset` queries (`myfile { fileAsset { ... } }`, `fileName`, `description`, etc.): unchanged.
- Change is purely additive.

## Test plan
- [x] Unit tests: `CustomFieldTypeBinaryLikeTest` (8 tests, all green) — asserts interface presence on both concrete types, common field set, consistent scalar types, and backward-compat fields.
- [x] Regression: full `com.dotcms.graphql.**` unit-test package passes (24 tests).
- [ ] Manual integration smoke via GraphiQL on a content type with Binary, File, and Image fields — verify all three respond to the same sub-field query.
- [ ] Postman collection for GraphQL binary-field queries (follow-up if missing).

## Commits
1. `test(graphql): failing tests for unified DotBinaryLike interface #34540` — TDD red step.
2. `feat(graphql): unify Binary, File, and Image field queries via DotBinaryLike #34540` — implementation.

This PR fixes: #34540


fabrizzio-dotCMS:
Closing this as superseded by #37594, which takes over the same issue (#34540). Recording the reasoning here rather than just closing, since the work in this PR identified the problem correctly and the decision is about shape, not about it being wrong.

**What changed under it.** #37594 types an asset-pointing field with an *interface* instead of the flat `DotFileasset` object, so a client narrows to the concrete content type the field actually points at. That covers the other limb of #34540 — the customer-defined fields on DOTASSET/FILEASSET-derived types, which block the AI tagging workflow and had no workaround.

**Why this PR's benefit does not survive that.** It exists to spare one level of nesting: `image { size }` instead of `image { fileAsset { size } }`. But once the field returns the asset's real type, that type's binary property is already a fully described `DotBinary` carrying all twelve of those properties plus `focalPoint` — verified by introspection against a running instance. The client still descends exactly one level, just to the correctly named property. And the twelve would have been attached to the flat view, which #37594 removes, so the work would need redoing on the interface either way.

**What #37594 kept from it**, explicitly:

- The problem statement. The flat view really was serving two base types through two fallback ternaries, and this PR named that first.
- The corrected version of `GraphqlAPITest.areFileassetFieldsPresent`. It used `allMatch`, so it asserted the type exposed *no field other than* the six — the opposite of its name, and a lock against ever adding one. That is what this PR's CI has been failing on since 2026-09-10, and it needed fixing regardless of which design landed.

**One thing from this PR that #37594 does *not* do**, worth flagging in case it matters to you: the twelve binary properties are no longer flattened onto the asset-pointing field. A client reads them through the asset's own binary property (`asset` for image-style content, `fileAsset` for file-style). If the flattening is wanted on the new shape, that is a follow-up worth its own issue — it is additive over what #37594 delivers.

@nollymar — #37594 also carries a breaking change and requests an exception to ADR-0022, and asks for your sign-off alongside @fmontes. Details are in the PR description and in the spec's ADR Alignment section.

PR #35380, opened 2026-04-18
fix(ui): round-3 polish — container list, bundle upload, toast severity

hmoreras:
## Summary

Round 3 of UI consistency fixes for the post-Lara-theme portlet polish (part of #35274). Four related areas:

### Container list (`dot-container-list`)
- Right-click anywhere on a row now opens the action context menu — previously `pContextMenuRow` was set on each `<tr>` but no `<p-contextMenu>` was attached, so the directive was a no-op. Fixed by wiring a `<p-contextMenu>` via `viewChild` and a `(contextmenu)` handler that reuses the existing `setContainerActions()` entries, matching the pattern already used by the template list.
- 3-dot action button is now hover-only (Tailwind `group` / `group-hover:opacity-100`), matching other list portlets. `<dot-action-menu-button>` replaced with a `<p-button pi-ellipsis-v>` that opens the same context menu, so right-click and the button share one source of truth.
- Fixed a `@ViewChild('actionsMenu')` regression: once a signal-based `viewChild` was declared for the context menu, the legacy decorator started resolving to the wrong component at runtime (`ContextMenu` instead of `Menu`), breaking `handleActionMenuOpen`. Migrating `actionsMenu` to the signal API restores correct resolution.
- Spec updated: added the `window.matchMedia` jsdom polyfill required by PrimeNG `ContextMenu`, and converted the action-menu DOM assertions to call `setContainerActions()` directly.

### Bundle upload dialog (`view_publish_tool.jsp`)
- Replaced the raw `<input type="file">` with a `dojox.form.Uploader` widget so the dialog matches the dijit styling of the surrounding portlet, and added a localized filename label.
- Worked around two bugs in the custom Dojo build's `Uploader`:
  1. `Uploader._createInput()` does not forward the `accept` attribute to its internal `<input type="file">`, so any file could be picked. `applyBundleAcceptFilter()` walks the widget's `domNode` and sets `accept=".tar.gz,.gz,.tgz"` on the real file input. Re-applied on every dialog open because `uploader.reset()` recreates the input.
  2. The widget only fires an `onChange` event — no visible element is updated — so the label kept reading "No file chosen" even after a pick. `updateBundleFileNameDisplay()` is wired to `onChange` in `dojo.ready()` and reflects the selection.
- Spacing + action-bar margin tweaks so the new controls sit cleanly.

Before 
Screenshot 2026-04-17 at 1 30 54 PM

After 
image

### Toast severity (`dot-message-display`)
- PrimeNG Toast expects severity `'warn'`, but `DotMessageSeverity` emits `'WARNING'`. Map `'warning' → 'warn'` before calling `MessageService.add()` so warning toasts pick up the warn style. SCSS class renamed (`.warning` → `.warn`) to stay in sync.

Before 

Screenshot 2026-04-17 at 1 23 40 PM

After 

Screenshot 2026-04-17 at 1 24 47 PM

### Container service (`dot-containers.service`)
- `copy()` switched from `PUT` to `POST` — the backend `_copy` endpoint is a POST, and the PUT was returning 405.

## Related

Part of #35274

## Test plan

- [ ] `yarn nx test dotcms-ui --testPathPattern=container-list.component.spec` passes
- [ ] Right-click on a container row opens the context menu with the correct actions (Publish/Unpublish/Duplicate for a working container; Unarchive/Delete for an archived one)
- [ ] 3-dot button on a container row is hidden by default and fades in on hover; clicking it opens the same menu
- [ ] Archived/system/file containers do not show the 3-dot or right-click menu (disableInteraction respected)
- [ ] Upload Bundle dialog: the OS file picker only lets you select `.tar.gz`, `.gz`, `.tgz` files
- [ ] After picking a file, the filename replaces "No file chosen" in the dialog
- [ ] Reopening the dialog resets the filename label
- [ ] Copying a container succeeds (previously returned 405)
- [ ] Triggering a warning toast (e.g. `DotMessageDisplayService.push({ severity: DotMessageSeverity.WARNING, ... })`) displays with the warn style rather than the default

This PR fixes: #35274


core-web/apps/dotcms-ui/src/app/portlets/dot-containers/container-list/container-list.component.ts
```diff
@@ -86,8 +87,8 @@ export class ContainerListComponent implements OnDestroy {
     private dialogService = inject(DialogService);
     private siteService = inject(SiteService);
 
-    @ViewChild('actionsMenu')
-    actionsMenu: Menu;
+    actionsMenu = viewChild<Menu>('actionsMenu');
+    rowContextMenu = viewChild<ContextMenu>('rowContextMenu');
```
oidacra, line 83: This PR migrates `actionsMenu` to `viewChild()` and adds `rowContextMenu = viewChild()`, but `@ViewChildren('tableRow') tableRows: QueryList<...>` is left on the legacy decorator API. Since it's the only remaining legacy query (and the only reason `QueryList` + `ViewChildren` are still imported), consider finishing the migration: `tableRows = viewChildren<ElementRef<HTMLTableRowElement>>('tableRow')` and updating `focusFirstRow()` to read from the signal's readonly array.

core-web/apps/dotcms-ui/src/app/portlets/dot-containers/container-list/container-list.component.ts
```diff
@@ -97,6 +98,7 @@ export class ContainerListComponent implements OnDestroy {
     notify$ = this.#store.notify$;
 
     selectedContainers: DotContainer[] = [];
+    contextMenuItems: MenuItem[] = [];
```
oidacra, line 92: `contextMenuItems` is assigned imperatively and is the last non-signal piece of state here after this PR. Consider `contextMenuItems = signal<MenuItem[]>([])` with `[model]="contextMenuItems()"` in the template and `this.contextMenuItems.set(...)` in `setContextMenu()`. The component isn't yet `OnPush`, so this is not a current bug — but it keeps the migration consistent and avoids rework when `OnPush` lands.

core-web/apps/dotcms-ui/src/app/portlets/dot-containers/container-list/container-list.component.html
```diff
@@ -89,15 +89,15 @@
                         <tr
                             (click)="handleRowClick(rowData)"
                             (keyup.enter)="handleRowClick(rowData)"
+                            (contextmenu)="setContextMenu($event, rowData)"
                             [class]="
-                                rowData.disableInteraction
+                                'group transition-colors duration-200 ' +
+                                (rowData.disableInteraction
                                     ? 'bg-gray-200! text-gray-500! pointer-events-none'
```
oidacra, line 96: The `[class]` binding is concatenating unconditional classes (`group transition-colors duration-200`) with a conditional string. Simpler and less error-prone to keep the unconditional classes on a plain `class` attribute and use a conditional `[class]`:

```html
class="group transition-colors duration-200"
[class]="rowData.disableInteraction ? 'bg-gray-200! text-gray-500! pointer-events-none' : ''"
```

or per-class bindings (`[class.bg-gray-200!]="rowData.disableInteraction"` etc.).

core-web/apps/dotcms-ui/src/app/portlets/dot-containers/container-list/container-list.component.spec.ts
```diff
@@ -350,13 +358,13 @@ describe('ContainerListComponent', () => {
                 menuItem: { label: 'Duplicate', command: expect.any(Function) }
             });
 
-            expect(publishContainer.actions).toEqual(actions);
+            expect(comp.setContainerActions(publishedContainer)).toEqual(actions);
```
oidacra, line 361: These tests used to verify that the rendered action-menu received the correct `[actions]` input. Now they call `comp.setContainerActions(container)` directly, which only covers the pure-logic path — the new `setContextMenu → contextMenuItems → <p-contextMenu>` wiring is no longer exercised. Consider keeping the direct-call assertion for the logic and adding one end-to-end test that fires `setContextMenu` via a click or `contextmenu` event and asserts `comp.contextMenuItems.map(i => i.label)` contains the expected entries — so a regression in the template binding still gets caught.

core-web/apps/dotcms-ui/src/app/view/components/dot-message-display/dot-message-display.component.spec.ts
```diff
@@ -110,6 +110,25 @@ describe('DotMessageDisplayComponent', () => {
...
+            message: 'message',
+            portletIdList: [],
+            severity: DotMessageSeverity.WARNING,
+            type: DotMessageType.SIMPLE_MESSAGE
+        });
+
+        expect(messageService.add).toHaveBeenCalledWith({
+            life: 300,
+            detail: 'message',
+            severity: 'warn'
```
oidacra, line 128: This test asserts `messageService.add` receives `severity: 'warn'`, which covers the mapping in the TS. The template also now maps `severity === 'warn'` → icon name `'warning'` — that branch is the actual visual fix but isn't exercised here. Consider adding `fixture.detectChanges()` after the `expect(messageService.add)` call and querying `dot-icon` to assert `name === 'warning'`, so the template mapping is also covered.

core-web/apps/dotcms-ui/src/app/view/components/dot-message-display/dot-message-display.component.scss
```diff
@@ -46,7 +46,7 @@
             color: colors.$error;
         }
 
-        &.warning {
+        &.warn {
```
oidacra, line 49: The `.warning` → `.warn` SCSS rename and the TS `'warning' → 'warn'` mapping are consistent for callers going through `DotMessageDisplayService`. Worth a quick pre-merge grep for any component calling PrimeNG's `MessageService.add({ severity: 'warning' })` directly — if one exists it would now hit a non-existent CSS selector and render without the warn color.

core-web/apps/dotcms-ui/src/app/portlets/dot-containers/container-list/container-list.component.ts
```diff
@@ -187,6 +189,29 @@ export class ContainerListComponent implements OnDestroy {
         return this.#store.getContainerActions(container);
     }
 
+    /**
+     * Open the context menu for a row — used by both right-click and the 3-dot button.
+     *
+     * @param {Event} event
+     * @param {DotContainer} container
+     * @memberof ContainerListComponent
+     */
```
oidacra: Nit: most of this JSDoc restates what the TypeScript signature already conveys (`@param {Event}`, `@param {DotContainer}`, `@memberof`). Only the trailing "used by both right-click and the 3-dot button" carries non-obvious WHY. Could shrink to a single line, e.g. `// Invoked by both right-click and the 3-dot action button.`

core-web/apps/dotcms-ui/src/app/portlets/dot-containers/container-list/container-list.component.ts
hmoreras: Fixed in 463bf539f6 — migrated `tableRows` to `viewChildren<ElementRef<HTMLTableRowElement>>('tableRow')`, removed the legacy `@ViewChildren`/`QueryList` imports, and updated `focusFirstRow()` to call `this.tableRows().find(...)`. Also updated the spec's `comp.tableRows.get(0)` to `comp.tableRows()[0]`.

core-web/apps/dotcms-ui/src/app/portlets/dot-containers/container-list/container-list.component.html
hmoreras: Fixed in 463bf539f6 — split into a static `class="group transition-colors duration-200"` attribute and a separate conditional `[class]` binding.

core-web/apps/dotcms-ui/src/app/view/components/dot-message-display/dot-message-display.component.spec.ts
hmoreras: Fixed in 463bf539f6 — added a new test `'should render warning icon for WARNING severity'` that emits a WARNING message, runs `detectChanges()`, and asserts `dot-icon[name] === 'warning'`, covering the template mapping path.

core-web/apps/dotcms-ui/src/app/portlets/dot-containers/container-list/container-list.component.ts
hmoreras: Fixed in 463bf539f6 — replaced the multi-line JSDoc with a single inline comment keeping only the non-obvious WHY: `// Invoked by both right-click and the 3-dot action button.`

core-web/apps/dotcms-ui/src/app/view/components/dot-message-display/dot-message-display.component.scss
hmoreras: name change double checked no need to cover that scenario.

oidacra (approved):

PR #35407, opened 2026-04-21
Update ESContentletAPIImpl.java

### LINKED ISSUE #35437: Backport #35407 to release-24.12.27_lts: Fix validated on 24.12.27_lts_v19
## Summary

The fix introduced in #35407 (excessive DB queries in `validateRelationships()` when saving contentlets with many MANY_TO_MANY relationships) has been **validated successfully** on the following Docker images:

- `dotcms/dotcms:24.12.27_lts_v19_0c188d2`
- `dotcms/dotcms:24.12.27_lts_v19`

## Action Required

A new patch release must be generated for the **`release-24.12.27_lts`** branch incorporating the fix from #35407.

## Background

When saving a Promotion with many required relationships (e.g. 499 PromotionRegisters), `validateRelationships()` called `getRelatedContent()` for every child contentlet in the `isParent` loop — including MANY_TO_MANY cardinalities where the result was never used. The `limit=1` parameter was silently ignored by the `RelationshipCache` path, causing `filterRelatedContentByLiveAndLanguage()` to fire one DB query per identifier per configured language. With 528 related identifiers and 44 languages this produced ~23K queries per save, accounting for 436 of ~641 seconds observed in Glowroot traces.

## References

- Fix PR: #35407
- Helpdesk ticket: https://helpdesk.dotcms.com/helpdesk/tickets/36004

erickgonzalez:
When saving a Promotion with many required relationships (e.g. 499                                                                                                                                                                                                                                                        
  PromotionRegisters), validateRelationships() called getRelatedContent()
  for every child contentlet in the isParent loop — including MANY_TO_MANY                                                                                                                                                                                                                                                  
  cardinalities where the result was never used. The limit=1 parameter was                                                                                                                                                                                                                                                  
  silently ignored by the RelationshipCache path, causing                                                                                                                                                                                                                                                                   
  filterRelatedContentByLiveAndLanguage() to fire one DB query per                                                                                                                                                                                                                                                          
  identifier per configured language. With 528 related identifiers and 44                                                                                                                                                                                                                                                   
  languages this produced ~23K queries per save, accounting for 436 of                                                                                                                                                                                                                                                      
  ~641 seconds observed in Glowroot traces.                                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                                                                            
  Fix: guard the query with the ONE_TO_MANY cardinality check (skipping it                                                                                                                                                                                                                                                  
  entirely for other cardinalities) and replace getRelatedContent() with                                                                                                                                                                                                                                                    
  FactoryLocator.getRelationshipFactory().dbRelatedContent(..., 1, 0),                                                                                                                                                                                                                                                      
  which queries the tree + contentlet_version_info tables directly with                                                                                                                                                                                                                                                     
  LIMIT 1, bypassing the cache path.                              

https://github.com/user-attachments/assets/1edf2ad7-93db-4af1-b2b1-c4ba13866834

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ESContentletAPIImpl.java
+++ b/dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ESContentletAPIImpl.java
@@ -8261,47 +8261,48 @@ public class ESContentletAPIImpl implements ContentletAPI {
                         cve.addRequiredRelationship(relationship, contentsInRelationship);
                     }
                     for (final Contentlet contentInRelationship : contentsInRelationship) {
-                        try {
-                            // In order to get the related content we should use method getRelatedContent
-                            // that has -boolean pullByParent- as parameter so we can pass -false-
-                            // to get related content where we are parents.
-                            final List<Contentlet> relatedContents = getRelatedContent(
-                                    contentInRelationship, relationship, false,
-                                    APILocator.getUserAPI()
-                                            .getSystemUser(), true, 1, 0, null);
-                            // If there's a 1-N relationship and the parent
-                            // content is relating to a child that already has
-                            // a parent...
-                            if (relationship.getCardinality()
-                                    == RELATIONSHIP_CARDINALITY.ONE_TO_MANY.ordinal()
-                                    && !relatedContents.isEmpty()
-                                    && !relatedContents.get(0).getIdentifier()
-                                    .equals(contentlet.getIdentifier())) {
-                                final StringBuilder error = new StringBuilder();
-                                error.append("ERROR! Parent content [").append(contentletId)
-                                        .append("] cannot be related to child content [")
-                                        .append(contentInRelationship.getIdentifier())
-                                        .append("] because it is already related to parent content [")
-                                        .append(relatedContents.get(0).getIdentifier()).append("]");
-                                Logger.error(this, error.toString());
-                                hasError = true;
-                                cve.addBadCardinalityRelationship(relationship,
-                                        contentsInRelationship);
-                            }
+                       try {                                                                                                                                                                                                                                                                                                 
+                          // For ONE_TO_MANY relationships, check if the child already has a
+                          // different parent. Skip for other cardinalities — the DB query is                                                                                                                                                                                                                               
+                          // unnecessary and causes N×M query explosion via the relationship cache.                                                                                                                                                                                                                         
+                          if (relationship.getCardinality()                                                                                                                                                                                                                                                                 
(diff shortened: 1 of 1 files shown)


fabrizzio-dotCMS:
## Leftover issues not addressed by this fix

### 1. `RelationshipFactoryImpl` — unnecessary Contentlet hydration (`dbRelatedContentByChild` lines 655, 695)

`dbRelatedContentByChild` selects `cvi.working_inode` and then calls `contentletAPI.find(inode, ...)` to hydrate a full `Contentlet` per row. In `validateRelationships`, only `Contentlet.getIdentifier()` is used from the result — which maps directly to `t.parent`, already present in the `tree` table.

A dedicated method returning `List<String>` identifiers instead of `List<Contentlet>` would eliminate the secondary `find()` call entirely:

```java
// proposed signature
List<String> dbRelatedContentIdentifiers(
        Relationship rel, Contentlet child,
        boolean hasParent, int limit, int offset) throws DotDataException;
```

SQL simplification:
```sql
-- current: fetches working_inode, then find(inode) per row
SELECT cvi.working_inode AS inode
FROM tree t, contentlet_version_info cvi
WHERE t.child = ? AND t.relation_type = ?
  AND cvi.working_inode IS NOT NULL
  AND t.parent = cvi.identifier

-- proposed: t.parent IS the identifier, no secondary find() needed
SELECT t.parent
FROM tree t
WHERE t.child = ? AND t.relation_type = ?
```

---

### 2. `ESContentletAPIImpl.getRelatedContent` — root cause of the query explosion is still present (lines 4545–4556)

The fix bypasses the cache for this specific call, but the underlying bug remains in `getRelatedContent`: when the relationship cache is warm the `limit` parameter is silently dropped before entering `getCachedRelatedContentlets`:

```java
if (relatedIds.containsKey(variableName)) {
    // limit is NOT forwarded here  ← root cause
    relatedContentlet = getCachedRelatedContentlets(relatedIds, variableName, language, live);
} else {
    // limit is passed correctly here
    relatedContentlet = getNonCachedRelatedContentlets(..., limit, offset, ...);
}
```

`getCachedRelatedContentlets` then expands every cached identifier across all configured languages with no upper bound:

```java
return relatedIds.get(variableName).stream()
    .flatMap(identifier -> filterRelatedContentByLiveAndLanguage(language, live, identifier))
    .collect(Collectors.toList()); // no .limit(n) anywhere
```

Any other caller of `getRelatedContent()` that passes a `limit > 0` and has a warm cache is still exposed to the same N × languages query explosion. This PR patches the symptom for `validateRelationships` but leaves the root cause intact.

fabrizzio-dotCMS (approved):
I'm approving as it looks sound for an LTS
However, I'm leaving a few notes that should be taken into account when taking this fix to main

PR #35150, opened 2026-03-27
feat(dotAI): Dot AI LangChain4J integration (First provider/OpenAI)

ihoffmann-dot:
## Summary
Replaces the direct OpenAI HTTP client (`OpenAIClient`) with a LangChain4J abstraction layer, enabling multi-provider AI support in dotCMS. This PR covers Phase 1: OpenAI via LangChain4J.

## Changes
- **`LangChain4jAIClient.java`**: New `AIProxiedClient` implementation that delegates chat, embeddings, and image requests to LangChain4J models.
- **`LangChain4jModelFactory.java`**: Factory that builds `ChatModel`, `EmbeddingModel`, and `ImageModel` instances from a `ProviderConfig`. Only place with provider-specific builder logic.
- **`ProviderConfig.java`**: Deserializable POJO for the `providerConfig` JSON secret (per provider section: model, apiKey, endpoint, maxTokens, maxCompletionTokens, etc.).
- **`AppConfig.java`**: Replaced legacy individual-field secrets (apiKey, model, etc.) with a single `providerConfig` JSON string. `isEnabled()` now only checks this field.
- **`AIAppValidator.java`**: Removed the OpenAI `/v1/models` validation call, which is incompatible with the multi-provider architecture.
- **`CompletionsResource.java`**: Updated `/api/v1/ai/completions/config` to derive model names and config values from `AppConfig` getters instead of iterating raw `AppKeys`.
- **`dotAI.yml`**: Removed legacy hidden fields; added `providerConfig` as the single configuration entry point.
- **Tests**: Added unit tests for `ProviderConfig`, `LangChain4jModelFactory`, and `LangChain4jAIClient`; updated `AIProxyClientTest` integration test to use `providerConfig`-based setup.

## Motivation
The previous implementation was tightly coupled to OpenAI's API contract (hardcoded HTTP calls, OpenAI-specific parameters, model validation via `/v1/models`). LangChain4J provides a provider-agnostic model interface, allowing future phases to add Azure OpenAI, AWS Bedrock, and Vertex AI without touching the core request/response flow.

The `providerConfig` JSON secret replaces multiple individual secrets with a single structured configuration, supporting per-section (chat/embeddings/image) provider and model settings.

## Related Issue
This PR fixes [#35183](https://github.com/dotCMS/core/issues/35183)
[EPIC: dotAI Multi-Provider Support #33970](https://github.com/dotCMS/core/issues/33970)

---

> [!NOTE]
> **High Risk**
[...]

PR #35426, opened 2026-04-22
fix(dotAI): Dot AI LangChain4J - ProviderConfig fixes

ihoffmann-dot:
## Summary

Consolidates AI configuration by moving prompts and per-capability settings into the `providerConfig` JSON sections, enabling a single-field setup in the Apps UI. Also adds comma-separated model fallback so any provider can specify multiple models in priority order.

- Make `providerConfig` visible in the Apps UI (`hidden: false`) and remove legacy per-field secrets (`rolePrompt`, `textPrompt`, `imagePrompt`, `imageSize`, `listenerIndexer`) from `dotAI.yml`
- `AppConfig` now reads prompts and settings from inside the `providerConfig` JSON sections instead of from separate app secrets
- `ProviderConfig.allModels()`: parse comma-separated `model` field into a priority list for fallback

## Configuration

```json
{
  "chat": {
    "provider": "openai",
    "apiKey": "sk-...",
    "model": "gpt-4o,gpt-4o-mini",
    "maxTokens": 16384,
    "temperature": 1.0,
    "rolePrompt": "You are dotCMSbot, an AI assistant to help content creators.",
    "textPrompt": "Use Descriptive writing style."
  },
  "embeddings": {
    "provider": "openai",
    "apiKey": "sk-...",
    "model": "text-embedding-ada-002",
    "listenerIndexer": { "default": "blog,news,webPageContent" }
  },
  "image": {
    "provider": "openai",
    "apiKey": "sk-...",
    "model": "dall-e-3",
    "size": "1792x1024",
    "imagePrompt": "Use 16:9 aspect ratio."
  }
}
```

## Notes

- Prompts (`rolePrompt`, `textPrompt`, `imagePrompt`) and embedding settings (`listenerIndexer`) now live inside `providerConfig` sections. Separate App secrets for these fields are no longer supported.
- Multi-model fallback: set `"model": "gpt-4o,gpt-4o-mini"` to try models in order. Sync requests fall back on both initialization and runtime failures; streaming falls back on initialization failures only (cannot retry once bytes are written to the output stream).

## Related Issue

[feat(dotAI): Dot AI LangChain4J integration (First provider/OpenAI)
#35150](https://github.com/dotCMS/core/pull/35150)


dotCMS/src/main/java/com/dotcms/ai/client/langchain4j/LangChain4jAIClient.java
```diff
@@ -153,45 +154,63 @@ public <T extends Serializable> void sendRequest(final AIRequest<T> request, fin
...
 
         final List<ChatMessage> messages = toMessages(payload.optJSONArray(AiKeys.MESSAGES));
         if (messages.isEmpty()) {
             throw new IllegalArgumentException("Chat request must contain at least one message");
         }
 
+        // Fallback is only possible before streaming starts (once bytes are written to output
+        // we cannot retry). Loop through models on initialization failures only.
+        RuntimeException lastException = null;
+        for (final String modelName : models) {
```
dario-daza, line 194: Is this going to throw multiple exceptions per model? or just the last known exception?
isn't better throw an exceptions as soon it occurs?

dotCMS/src/main/java/com/dotcms/ai/client/langchain4j/LangChain4jAIClient.java
```diff
@@ -273,35 +292,57 @@ private void writeToOutput(final String responseJson, final OutputStream output)
...
+            final ProviderConfig baseConfig,
+            final Cache<String, M> modelCache,
+            final Function<ProviderConfig, M> modelBuilder,
+            final Function<M, String> executor) {
+        final List<String> models = baseConfig.allModels();
+        if (models.isEmpty()) {
+            throw new IllegalArgumentException(
+                    "No model configured in providerConfig." + section + " — set 'model'");
+        }
+        RuntimeException lastException = null;
```
dario-daza, line 332: Same here, plus is there some logic that could be reused here?

dotCMS/src/main/java/com/dotcms/ai/client/langchain4j/LangChain4jAIClient.java
```diff
@@ -153,45 +154,63 @@ public <T extends Serializable> void sendRequest(final AIRequest<T> request, fin
...
+                        cacheKeyPrefix + ":chat:streaming:" + modelName,
+                        () -> LangChain4jModelFactory.buildStreamingChatModel(modelConfig));
+            } catch (ExecutionException | UncheckedExecutionException e) {
+                final Throwable cause = e.getCause() != null ? e.getCause() : e;
+                lastException = new IllegalArgumentException(
+                        "Failed to initialize streaming model '" + modelName + "': " + cause.getMessage(), cause);
+                Logger.warn(LangChain4jAIClient.class,
+                        "Streaming model '" + modelName + "' init failed: " + cause.getMessage()
+                        + (models.size() > 1 ? " — trying next model" : ""));
+                continue;
```
jsanca: this continue and return make harder the reading
also you should introduce some observability here, logs are ok but are not telling anything about how long it is take, etc
also you are using system prompt that never change, probably want to introduce a cache entry point to save tokens

dotCMS/src/main/java/com/dotcms/ai/client/langchain4j/LangChain4jAIClient.java
```diff
@@ -273,35 +292,57 @@ private void writeToOutput(final String responseJson, final OutputStream output)
...
+        final List<String> models = baseConfig.allModels();
+        if (models.isEmpty()) {
+            throw new IllegalArgumentException(
+                    "No model configured in providerConfig." + section + " — set 'model'");
+        }
+        RuntimeException lastException = null;
+        for (final String modelName : models) {
+            try {
+                final ProviderConfig modelConfig = ImmutableProviderConfig.copyOf(baseConfig).withModel(modelName);
+                final M model = modelCache.get(
```
jsanca, line 336: for instance this method is easier to read

dotCMS/src/main/java/com/dotcms/ai/client/langchain4j/ProviderConfig.java
```diff
@@ -53,7 +56,31 @@
...
+     */
     @Nullable String model();
+
+    /**
+     * Returns the ordered list of model names parsed from {@link #model()}.
+     * If {@code model} is blank or null, returns an empty list.
+     */
+    default List<String> allModels() {
+        final String m = model();
+        if (m == null || m.isBlank()) {
```
jsanca, line 72: usually if something may be null, Optional is a fantastic choice

dotCMS/src/main/java/com/dotcms/ai/client/langchain4j/LangChain4jAIClient.java
ihoffmann-dot: Good one. It throws the last one, the idea is to have a model fallback as it works currently in production. That's why it waits until the last exception to throw it

ihoffmann-dot: Also, I added a comment in the code explaining that part, and we have a Logger to keep each exception in case we need them in the logs

dotCMS/src/main/java/com/dotcms/ai/client/langchain4j/LangChain4jAIClient.java
ihoffmann-dot: Good catch, I extracted the logic into initStreamingModel() and added timing logs. Regarding the cache, is a good observation but it would need some logic that we could implement in another PR.

dotCMS/src/main/java/com/dotcms/ai/client/langchain4j/LangChain4jAIClient.java
ihoffmann-dot: Sure, executeWithFallback is the reusable helper, all three sync paths (chat, embeddings, image) delegate to it. For the streaming path initStreamingModel() was extracted following the same pattern

dario-daza (approved):

dotCMS/src/test/java/com/dotcms/ai/client/langchain4j/LangChain4jAIClientTest.java
```diff
@@ -147,6 +150,109 @@ public void test_toImageResponseJson_nullUrl_returnsEmptyString() {
...
+                .apiKey("sk-test")
+                .model(models)
+                .build();
+    }
+
+    private static Cache<String, String> freshCache() {
+        return CacheBuilder.newBuilder().build();
+    }
+
+    @Test
```
nollymar, line 169: I suggest adding javadoc to tests. We generally use this format:
```
  /**
     * Method to test: {@link StoryBlockAPI#refreshReferences(Contentlet)}
     * Given Scenario: This will create 2 block contents, adds a rich content to each block content and retrieve the json.
     * ExpectedResult: The new json will contain the rich text data map for each block content.
     */
```

nollymar (approved):

PR #35435, opened 2026-04-23
fix(content-types): reset field editor form state when creating a new field

### LINKED ISSUE #35434: Field editor retains previous field's values when creating a new field
## Problem Statement

In the content type editor (`content-types-angular`), when a user creates a field that has a **Value** textarea (e.g., Radio, Select, Checkbox, Multi Select) and saves it, opening the dialog for a **new** field of the same family immediately shows the previously entered values still populated in the Value textarea.

The new field dialog should always start with an empty Value (and empty Default Value / Hint / Name / flags). Stale state from the previously saved field is leaking into a newly instantiated field form, which can cause:

- Users to accidentally save duplicate / wrong values into the new field.
- Loss of trust in the editor (data from an unrelated field appears out of thin air).
- Real data corruption when the user simply clicks Save without realizing the Value textarea is not empty.

**Environment:** dotCMS UI (Angular), latest `main`. Reproduces in Chrome.

## Steps to Reproduce

1. Edit a content type.
2. Add a new **Radio** field. Fill in the Name (e.g. `radio`) and enter values in the Value textarea:
   ```
   value|1
   value.2|2
   ```
3. Save the Radio field.
4. Add a new field of a related type with the same Value control — e.g. **Select**.
5. Open the new Select field editor.

**Expected:** The Value textarea is empty — this is a brand new field.

**Actual:** The Value textarea is prepopulated with `value|1` and `value.2|2`, i.e. the values from the previously saved Radio field.

Image

https://github.com/user-attachments/assets/4a2fb885-467b-41e1-9185-f6ee01fa54bd

## Acceptance Criteria

- [ ] Creating a new field after saving any field type with a Value textarea opens a dialog with an empty Value textarea.
- [ ] Default Value, Hint, Name, and all boolean flags (Required, User Searchable, System Indexed, Show in List, Unique) are reset for the new field.
- [ ] The bug is verified against Radio → Select, Radio → Radio, Select → Select, Select → Multi Select, and Checkbox → Radio combinations (all field types that expose the Value textarea).
- [ ] Editing an existing field still loads that field's real values (no regression in edit mode).
- [ ] Cancelling a new-field dialog and re-opening it does not resurrect stale values.
- [ ] Unit/Spectator test added that asserts the new-field form state is reset when the dialog opens for a new field.

## dotCMS Version

Latest from `main` branch.

## Severity

## Links

NA

oidacra:
## Summary

Fixes a bug in the content type editor where the field properties dialog leaked state (most visibly the Value textarea) from a previously saved field into a newly opened field dialog. Users would see the prior field's values prepopulated, the textarea would behave as uneditable, and Save would stay disabled until toggling an unrelated checkbox.

Closes #35434

## Root cause

Two independent defects compounded to produce the visible behavior:

1. **`DotTextareaContentComponent.writeValue`** had a truthy guard (`if (value)`) that skipped empty/null/undefined values. When Reactive Forms called `writeValue('')` to reset the control, the Monaco editor's internal model kept the stale content from the previous field.

2. **`DynamicFieldPropertyDirective`** decided whether to recreate the dynamic inner component by comparing `previousFieldId !== currentFieldId`. Since every *new* field has `id === null`, switching between two new fields (e.g. Radio → Select) resulted in `null !== null === false` → the `ValuesPropertyComponent` (and its child Monaco editor) was reused rather than recreated, preserving its stale DOM.

## Changes

- `dot-textarea-content.component.ts` — `writeValue` now writes the value directly (using `?? ''`), so empty/null/undefined correctly reset the control.
- `dynamic-field-property.directive.ts` — compare the previous vs. current field *reference* instead of ids, so the inner component is recreated whenever the field instance changes.
- `dot-textarea-content.component.spec.ts` — new `writeValue` tests asserting the control resets for `''`, `null`, and `undefined`.

## Acceptance criteria

- [x] Creating a new field after saving any field type with a Value textarea opens a dialog with an empty Value textarea.
- [x] Default Value, Hint, Name, and all boolean flags are reset for the new field.
- [x] The Value textarea in the new-field dialog is immediately editable.
- [x] The Save button follows its normal enable/disable rules (no sticky-disabled state from stale form state).
- [x] Editing an existing field still loads that field's stored values (no regression in edit mode).
- [x] Cancelling a new-field dialog and re-opening it does not resurrect stale values.
- [x] Unit test added asserting `writeValue` correctly resets for empty/null/undefined.

## Test plan

- [ ] Edit a content type.
- [ ] Add a Radio field with values like `value|1` / `value.2|2`, save.
- [ ] Add a new Select field — the Value textarea must be empty and editable; Save must enable as soon as required fields are valid.
- [ ] Repeat for Radio → Radio, Select → Select, Select → Multi Select, Checkbox → Radio.
- [ ] Edit an existing saved field and confirm its stored values still load correctly.
- [ ] Open a new field dialog, type some values, Cancel, reopen — the form must be empty.

### DIFF
--- a/core-web/apps/dotcms-ui/src/app/portlets/shared/dot-content-types-edit/components/fields/content-type-fields-properties-form/field-properties/dynamic-field-property-directive/dynamic-field-property.directive.spec.ts
+++ b/core-web/apps/dotcms-ui/src/app/portlets/shared/dot-content-types-edit/components/fields/content-type-fields-properties-form/field-properties/dynamic-field-property-directive/dynamic-field-property.directive.spec.ts
@@ -103,4 +103,50 @@ describe('Directive: DynamicFieldPropertyDirective', () => {
         expect(testComponent.group).toEqual(hostSpectator.hostComponent.group);
         expect(testComponent.helpText).toEqual('helpText');
     });
+
+    describe('shouldRecreate — field reference identity', () => {
+        const buildNewField = (overrides: Partial<DotCMSContentTypeField> = {}) =>
+            ({
+                ...hostSpectator.hostComponent.field,
+                id: null,
+                ...overrides
+            }) as DotCMSContentTypeField;
+
+        beforeEach(() => {
+            fieldPropertyService.getComponent.mockClear();
+        });
+
+        it('recreates the component when switching between two new fields with null id but different references', () => {
+            hostSpectator.hostComponent.field = buildNewField({ variable: 'first' });
+            hostSpectator.detectChanges();
+            const callsAfterFirst = fieldPropertyService.getComponent.mock.calls.length;
+
+            hostSpectator.hostComponent.field = buildNewField({ variable: 'second' });
+            hostSpectator.detectChanges();
+
+            expect(fieldPropertyService.getComponent.mock.calls.length).toBe(callsAfterFirst + 1);
+        });
+
+        it('does not recreate the component when the same field reference is passed again', () => {
+            const sameField = buildNewField({ variable: 'stable' });
+            hostSpectator.hostComponent.field = sameField;
+            hostSpectator.detectChanges();
+            const callsAfterFirst = fieldPropertyService.getComponent.mock.calls.length;
+
+            hostSpectator.hostComponent.field = sameField;
+            hostSpectator.detectChanges();
+
+            expect(fieldPropertyService.getComponent.mock.calls.length).toBe(callsAfterFirst);
+        });
+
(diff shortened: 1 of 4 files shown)


hmoreras (approved):

oidacra:
Added three unit tests in 13adbe9 covering the new reference-based `shouldRecreate` logic in `DynamicFieldPropertyDirective`:

1. Two new fields with `id=null` but different references → recreates the inner component.
2. The same field reference passed again → does not recreate.
3. Switching from a saved field to a new null-id field → recreates.

All 2109 tests pass.

PR #34165, opened 2025-12-24
refactor: replace CoreWebService with HttpClient in various services

nicobytes:
### Summary
This commit refactors multiple services to replace the deprecated CoreWebService with Angular's HttpClient for making HTTP requests. The changes enhance code clarity and maintainability by utilizing Angular's built-in HTTP capabilities.

### Changes Made
- Updated services to use HttpClient for GET, POST, PUT, and DELETE requests.
- Replaced  operator with  for extracting response data.
- Ensured all services maintain their original functionality while improving type safety and reducing boilerplate code.

### Impact
This refactor improves the overall architecture by promoting the use of Angular's HttpClient, aligning with best practices for modern Angular development.

### Testing
All affected services have been tested to ensure that they function correctly with the new implementation.

### Checklist
- [x] Code compiles without errors
- [x] Documentation updated
- [x] Linting & formatting applied
- [x] Ready for review

This PR fixes: #33073

PR #35440, opened 2026-04-23
fix(auth): send JSON object body to /api/v1/changePassword

oidacra:
## Summary

Fixes the "Change Password" button on the Password Recovery page, which silently failed with `HTTP 415 Unsupported Media Type` and a console `TypeError: Cannot read properties of undefined (reading '0')`.

**Root cause:** `LoginService.changePassword` was calling `HttpClient.post(url, JSON.stringify({...}))`. When given a string body, Angular's `HttpClient` auto-sets `Content-Type: text/plain`, but the backend `ResetPasswordResource` declares `@Consumes(MediaType.APPLICATION_JSON)` — so every request was rejected with 415. The error handler at `reset-password.component.ts:125` then crashed reading `errors[0]` on the non-JSON 415 body.

**Fix:** Pass the plain object so Angular serializes it and sets `Content-Type: application/json` automatically. Matches the pattern of every other auth call in `LoginService` (`recoverPassword`, `loginUser`, `loginAs`, etc.).

**Regression origin:** PR #34165 (`df97483c4e`) — the `CoreWebService` → `HttpClient` refactor. `CoreWebService.request()` handled JSON headers internally; the migrated code kept the `JSON.stringify` but lost the headers.

## Verification

Reproduced locally against `http://localhost:8080`:

| Request body | Status | Response |
|---|---|---|
| `JSON.stringify({password, token})` (before) | **415** | `{"message":"HTTP 415 Unsupported Media Type"}` |
| `{password, token}` (after) | **400** | `{"errors":[{"errorCode":"reset-password-token-invalid","message":"Invalid token..."}]}` |

The 400 response shape matches what `reset-password.component.ts` expects — no more TypeError.

Closes #35269

## Acceptance Criteria

- [x] `POST /api/v1/changePassword` is sent with `Content-Type: application/json`
- [x] "Change Password" button triggers the request successfully (no 415)
- [x] Error responses are parsed without a TypeError in the console
- [x] Happy path updates the password and allows login with the new one

## Test plan

- [x] `yarn nx test dotcms-ui` — 2109 passing
- [ ] Manual: trigger password recovery → open email link → submit new password → verify successful change and subsequent login

## Changed files

- `core-web/libs/dotcms-js/src/lib/core/login.service.ts` — remove `JSON.stringify` on the `changePassword` body

This PR fixes: #35269

### DIFF
--- a/core-web/apps/dotcms-ui/src/app/view/components/login/reset-password-component/reset-password.component.spec.ts
+++ b/core-web/apps/dotcms-ui/src/app/view/components/login/reset-password-component/reset-password.component.spec.ts
@@ -1,5 +1,7 @@
 import { throwError } from 'rxjs';
 
+import { provideHttpClient } from '@angular/common/http';
+import { HttpTestingController, provideHttpClientTesting } from '@angular/common/http/testing';
 import { DebugElement } from '@angular/core';
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
@@ -8,8 +10,9 @@ import { ActivatedRoute } from '@angular/router';
 import { RouterTestingModule } from '@angular/router/testing';
 
 import { DotMessageService, DotRouterService } from '@dotcms/data-access';
-import { LoginService } from '@dotcms/dotcms-js';
+import { DotcmsEventsService, LoginService } from '@dotcms/dotcms-js';
 import {
+    DotcmsEventsServiceMock,
     LoginServiceMock,
     MockDotMessageService,
     MockDotRouterService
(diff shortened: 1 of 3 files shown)


oidacra:
Addressed in 59481e6d11:

**1. Silent blank message on non-standard errors** — Done. Added `?? response.error?.message ?? ''` fallback so a 415-style body surfaces the backend message instead of a blank state.

**2. No test for absent-`errors` case** — Done. New test `should not crash when the error body has no "errors" array` in `reset-password.component.spec.ts` flushes a 415 with `{ message: "..." }` and asserts the component survives and shows the fallback message.

**3. No unit test for `changePassword` in the service** — Done (in the component spec, not the service spec). New `describe('ResetPasswordComponent — HTTP contract')` mounts the real `LoginService` with `HttpTestingController` and asserts:
- `POST /api/v1/changePassword`
- `typeof req.request.body === 'object'` (catches any future `JSON.stringify` regression — a string body triggers `text/plain` and the 415)
- Body shape `{ password, token }`

The natural home for this test would be `libs/dotcms-js/login.service.spec.ts`, but that project has `skip:test` and `skip:lint` in `project.json` so karma specs there don't run in CI — which is the structural reason this regression slipped. The component-level HTTP test runs in the existing Jest/CI pipeline and catches the same class of bug.

**4. Deprecated `subscribe(successFn, errorFn)`** — Skipping. Pre-existing, not introduced by this PR; out of scope.

dsolistorres (approved):

PR #35453, opened 2026-04-24
fix(content-type): fix ensure parameter in multi-type search pagination

rjvelazco:
## Summary

Follow-up to #35300. That PR fixed the `ensure` parameter for the **single-type** path in `ContentTypeAPIImpl.search()`. The **multi-type path** (triggered by `/api/v1/contenttype?type=CONTENT&type=WIDGET…`, which routes through `ContentTypeFactoryImpl.searchMultipleTypes()` and a UNION ALL query) still had the same two bugs.

### Reproduction on the demo dataset (before the fix)

```
GET /api/v1/contenttype?orderby=name&direction=ASC&per_page=10
      &ensure=Product,webPageContent,YouTube
      &type=CONTENT&type=WIDGET
```

24 total content types in `CONTENT ∪ WIDGET`:

| Page | Returned | Expected | Issue |
|------|----------|----------|-------|
| 1    | 9 items  | 10       | short by 1 |
| 2    | 8 items  | 10       | `Product`, `webPageContent`, `YouTube` **duplicated from page 1** |
| 3    | 7 items  | 4        | same 3 ensure items duplicated a **third** time |

**6 unique items were never returned at all**: `Banner`, `BlogAuthor`, `BlogComment`, `CategoryFilter`, `forms`, `PhotoGallery`.

### Root cause

`ContentTypeFactoryImpl.searchMultipleTypes()` had two bugs analogous to the single-type path before #35300:

1. The ensure block was adding ensure items to `result` on **every page**, not only page 1, producing duplicates.
2. The UNION query called `dc.setStartRow(offset)` with the raw client offset, without shifting back by the number of ensure items that borrowed slots on page 1 — so items immediately following the displaced slots were permanently skipped.

### Fix

Mirror the single-type fix from #35300 inside `searchMultipleTypes`:

- Always register **all** found ensure IDs in `includedIds` so the UNION merge filter removes them on every page.
- Only add ensure items to the result when `offset == 0`.
- On pages 2+, compute `adjustedOffset = Math.max(0, offset - ensureTypes.size())` and pass it to `dc.setStartRow()`.

### Expected behavior after the fix (same request as above)

| Page | Returned | Contains ensure items? | Duplicates? |
|------|----------|------------------------|-------------|
| 1    | 10       | yes (3) | none |
| 2    | 10       | no      | none |
| 3    | 4        | no      | none |

All 24 unique items returned exactly once, no skipped items.

## Changes

- `dotCMS/src/main/java/com/dotcms/contenttype/business/ContentTypeFactoryImpl.java` — ensure-block rewrite and `adjustedOffset` wiring inside `searchMultipleTypes`.
- `dotcms-integration/src/test/java/com/dotcms/contenttype/test/ContentTypeAPIImplTest.java` — four new integration tests mirroring the single-type ensure tests:
  - `testSearchMultipleTypes_withEnsure_singleItem_appearsOnPage1Only_noSkippedItems`
  - `testSearchMultipleTypes_withEnsure_multipleItems_noDuplicatesNoSkipped`
  - `testSearchMultipleTypes_withoutEnsure_paginatesNormally` (regression)
  - `testSearchMultipleTypes_withEmptyEnsure_paginatesNormally` (regression)

## Known follow-ups (out of scope)

- **Unauthorized ensure items**: if the caller lacks READ permission on an ensure item, the factory still counts it in `adjustedOffset` but `ContentTypeAPIImpl.searchMultipleTypes` filters it out afterwards, leaving page 1 short by one slot and page 2's offset off by one. Same edge case exists in the single-type path; a proper fix requires moving ensure handling out of the factory and into the API layer where permissions are known.
- **`ensure.size() > per_page`**: the extra ensure items are not lost but may appear on a later page at their natural sort position rather than being guaranteed on page 1. Same behavior as the single-type path.

## Test plan

- [ ] `./mvnw verify -pl :dotcms-integration -Dcoreit.test.skip=false -Dit.test="ContentTypeAPIImplTest#testSearchMultipleTypes_withEnsure_singleItem_appearsOnPage1Only_noSkippedItems+testSearchMultipleTypes_withEnsure_multipleItems_noDuplicatesNoSkipped+testSearchMultipleTypes_withoutEnsure_paginatesNormally+testSearchMultipleTypes_withEmptyEnsure_paginatesNormally"` passes
- [ ] Hit the demo endpoint from the Reproduction section above and confirm page 1 = 10, page 2 = 10, page 3 = 4, all 24 items unique
- [ ] Open Content Drive with a URL filter on two base types plus a pre-selected content type; reload and scroll the multiselect — every content type appears exactly once

### Issue Video

https://github.com/user-attachments/assets/ce01c4f5-839c-4f3e-bcb5-71cd1ab5c4aa

This PR fixes: #35213

This PR fixes: #35213


rjvelazco:
Thanks for the thorough review — addressed the primary correctness concern in the latest push (commit e9072ddb96).

**Fix applied (issue #1 + #2):** Added `AND inode.inode NOT IN (?, ?, ...)` inside each SELECT of the UNION ALL query so `setMaxRows` and `setStartRow` operate on the correct (non-ensure) row stream. Pages no longer come up short when an ensure item's natural sort position falls inside the fetch window. The Java STEP 3 filter stays in place as defense in depth (also handles `ct.id()` matches that aren't `inode()` matches).

**Tests added** to close the gap:
- `testSearchMultipleTypes_withEnsure_itemSortsEarly_pageStaysFullSize` — ensure=A (first alphabetically). Would fail without the SQL-level exclusion (page 1 returns 2 items instead of 3).
- `testSearchMultipleTypes_withEnsure_itemInMidFetchWindow_pageStaysFullSize` — 10 items A–J, ensure=F (sorts at position 5, inside page 3's fetch window). Would fail without the fix (page 3 returns 2 items instead of 3).

**Issues kept as follow-ups** (documented in the PR body):
- #3 `adjustedOffset` formula refinement when `ensure.size() > per_page` — same behavior as the single-type path (merged in #35300); worth a single combined fix later.
- #4 `find()` runs every page — performance-only, same pattern as the single-type path.
- #5 Permission-aware offset — requires moving ensure handling into the API layer; mirrors the single-type follow-up.
- #6 Unconditional Java re-sort — pre-existing, out of scope for this bug fix.

E2E failure on the previous run was unrelated — Docker Hub returned 504 Gateway Timeout while pulling `opensearchproject/opensearch:1.3.6`. Will re-run once the new commit's checks complete.

ihoffmann-dot (approved):

dotCMS/src/main/java/com/dotcms/contenttype/business/ContentTypeFactoryImpl.java
```diff
@@ -1439,30 +1439,43 @@ public List<ContentType> searchMultipleTypes(final String search, final Collecti
...
-                    return result;
+                    // Early return if ensure items already fill the page.
+                    if (result.size() >= effectiveLimit) {
+                        return result;
+                    }
+                } else {
+                    // Pages 2+: shift the UNION offset back by the number of ensure items that
+                    // were "borrowed" from pagination on page 1. Without this, items immediately
+                    // following the ensure-displaced slots are permanently skipped.
+                    adjustedOffset = Math.max(0, offset - ensureTypes.size());
```
dario-daza: it is worth to check this claude comment:

```java
// Current (wrong for ensure.size() > effectiveLimit):
adjustedOffset = Math.max(0, offset - ensureTypes.size());

// Correct:
final int ensureSlotsUsedOnPage1 = Math.min(ensureTypes.size(), effectiveLimit);
adjustedOffset = Math.max(0, offset - ensureSlotsUsedOnPage1);
```

plus did you add a test to check what happens when the ensure params exceed the max num of items in the page?

rjvelazco: Good catch — applied in `8d181aeab5`. Used your suggested formula:

```java
final int ensureSlotsUsedOnPage1 = Math.min(ensureTypes.size(), effectiveLimit);
adjustedOffset = Math.max(0, offset - ensureSlotsUsedOnPage1);
```

Without this, when `ensure.size() > perPage`, page 3 over-shifts the offset and overlaps with page 2. Concrete example with `ensure=5 items, perPage=3`:

| Page | offset | Bad formula `offset - 5` | Correct `offset - min(5,3)` |
|------|--------|-------------------------|-----------------------------|
| 2 | 3 | 0 → non-ensure[0..2] | 0 → non-ensure[0..2] |
| 3 | 6 | **1 → non-ensure[1..3]** ← overlap | 3 → non-ensure[3..5] |

Added `testSearchMultipleTypes_withEnsure_moreEnsureItemsThanPageSize_noOverlap` that exercises exactly this case (9 items A–I, ensure=[F,G,H,I], perPage=3) and asserts no overlap between pages 1, 2, and 3. Would have failed with the old formula.

Note: the documented "ensure overflow" follow-up — when `ensure.size() > perPage`, the items beyond the first `perPage` are excluded from `includedIds` and therefore don't appear on any page — is still out of scope. The test handles that by only asserting that non-ensure items A–E are returned somewhere, not that the 4th ensure item I appears.

rjvelazco: Done!

dotCMS/src/main/java/com/dotcms/contenttype/business/ContentTypeFactoryImpl.java
```diff
@@ -1439,30 +1439,50 @@ public List<ContentType> searchMultipleTypes(final String search, final Collecti
...
+                // removes them from the normal paginated stream on every page.
                 for (ContentType ct : ensureTypes) {
-                    if (result.size() < effectiveLimit) {
-                        result.add(ct);
-                        includedIds.add(ct.inode());
-                        includedIds.add(ct.id());
-                    } else {
-                        break;
-                    }
+                    includedIds.add(ct.inode());
```
nollymar: remove `includedIds.add(ct.inode());` as `id=inode` for content types. It is redundant

rjvelazco: Done in `85a6d92ef3` — kept `includedIds.add(ct.id())` and dropped `ct.inode()`. The SQL-level `inode.inode NOT IN (...)` clause and the Java STEP 3 merge filter both still work correctly since `id == inode` for content types.

PR #35456, opened 2026-04-24
fix(dotAI): Dot AI LangChain4J - per-site config support via siteId query param

### LINKED ISSUE #36031: fix(dotAI): Config API always resolved site from HTTP Host header, ignoring target site
## Description

`GET /api/v1/ai/completions/config` and `PUT /api/v1/ai/completions/config` always resolved the target site from the HTTP `Host` header. This made it impossible to read or save configuration for a specific site — including `SYSTEM_HOST` — through the API or from the custom config UI.

Background threads use the `SYSTEM_HOST` configuration for embeddings, so any save made through the UI would silently target the wrong site and leave `SYSTEM_HOST` with stale configuration.

## Changes needed

- Add optional `?siteId=<identifier>` query parameter to both `GET` and `PUT /api/v1/ai/completions/config`
- Accept `SYSTEM_HOST` as a special-case value
- Fall back to HTTP host resolution when `siteId` is not provided (backward-compatible)
- Frontend passes the site identifier from the route param (`dotAI/edit/:id`) on both load and save

## Acceptance criteria

- `GET /api/v1/ai/completions/config?siteId=SYSTEM_HOST` returns the system-level config
- `PUT /api/v1/ai/completions/config?siteId=SYSTEM_HOST` saves to system host
- Omitting `siteId` continues to resolve from HTTP `Host` header (no regression)
- Config UI reads and saves the correct site when navigating from `dotAI/edit/:siteId`

## Related PR

Implemented in #35456

ihoffmann-dot:
## Summary

Config endpoints were always resolving the target host from the HTTP `Host` header, making it impossible to read or save configuration for a specific site (including `SYSTEM_HOST`) through the API.

- `GET /api/v1/ai/completions/config` and `PUT /api/v1/ai/completions/config` now accept an optional `?siteId=<identifier>` query param to scope the operation to a specific site
- `SYSTEM_HOST` is supported as a special-case value
- Falls back to HTTP host resolution when `siteId` is not provided (backward-compatible)
- Frontend passes the site identifier from the route param (`dotAI/edit/:id`) on both load and save

## Notes

- Without this fix, saving config through the custom UI always targeted the site resolved from the HTTP `Host` header (e.g. `demo.dotcms.com`), never `SYSTEM_HOST`. Background threads use `SYSTEM_HOST` config, so embeddings would silently pick up stale configuration.
- `SYSTEM_HOST` can now be explicitly targeted by passing `?siteId=SYSTEM_HOST`.

## Related Issue

This PR fixes https://github.com/dotCMS/core/pull/35150


nollymar (approved):

dotcms-postman/src/main/resources/postman/AI.postman_collection.json
```diff
@@ -3520,6 +3520,289 @@
...
+                "type": "text/javascript"
+              }
+            }
+          ],
+          "request": {
+            "auth": {
+              "type": "basic",
+              "basic": [
+                {
+                  "key": "password",
```
dario-daza, line 3555: Next time you can use an authentication with a Pre-request in the postman collection (like the one in ConfigurationResource.postman_collection.json), this allow you avoid the auth setting in every request and just use the authentication **Inherit auth from parent**

dotCMS/src/main/java/com/dotcms/ai/rest/CompletionsResource.java
```diff
@@ -230,7 +233,12 @@ public Response saveConfig(@Context final HttpServletRequest request,
         }
 
         try {
-            final Host host = WebAPILocator.getHostWebAPI().getCurrentHostNoThrow(request);
+            final Host host = resolveHostStrict(siteId, request, user);
+            if (host == null) {
+                return Response.status(Response.Status.BAD_REQUEST)
```
dario-daza, line 257: Cool

dotCMS/src/main/java/com/dotcms/ai/rest/CompletionsResource.java
```diff
@@ -310,6 +322,58 @@ private static void redactNode(final JsonNode node) {
...
+    }
+
+    /**
+     * Resolves a host from {@code siteId} strictly — no fallback.
+     * Returns {@code null} when siteId is blank (caller uses its own fallback).
+     * Returns {@code null} when the site is not found.
+     * Throws {@link DotSecurityException} when the user lacks permission.
+     * Use for write operations where silently targeting the wrong site is unacceptable.
+     */
+    private static Host resolveHostStrict(final String siteId,
```
dario-daza, line 375: I see that this method is pretty similar to the one above `resolveHost()` the only difference that I can tell is the return null when we go trough the Exception is it any issue if you use just this one?

ihoffmann-dot: Good catch, but actually the two methods have different contracts. 

resolveHost() always returns a valid host, handles all exceptions and falls back to the right host, which is OK for reads. resolveHostStrict is for writes, there is no fallback when siteId is provided, it returns null (400) or throws DotSecurityException (403). 

I think merging them would need a flag or extra logic.

dotcms-postman/src/main/resources/postman/AI.postman_collection.json
ihoffmann-dot: Good one, I take it!

dario-daza (approved):

ihoffmann-dot:
Closes #36031

PR #35465, opened 2026-04-27
feat(content-drive): chip-style filter component for toolbar

### LINKED ISSUE #33524: [TASK] Update the visual style for the filters in content drive
### Description

the filter dropdowns in content drive should be styled like pills or chips. Meli has the newest design but we can wait for the lara theme to be turned on and then we can customize the look and feel of these dropdowns. 

### Acceptance Criteria

the dropdowns should look like chips. 

### Priority

None

### Additional Context

_No response_

zJaaal:
## Summary

- Adds `DotChipFilterComponent` — a pill-shaped, single-line filter trigger that displays a `title` and a derived selection summary (none / one / two-comma-joined / "first and N more")
- Replaces the fixed-column toolbar grid with a wrapping flex row, restoring the PrimeNG multiselect `min-w-56` floor that the new layout would otherwise collapse
- Active state uses the primary palette tints (`primary-100/900`) so multiple active chips read as a group rather than competing fills

Closes #33524

## Test plan

- [ ] `nx test content-drive-ui --testPathPattern=dot-chip-filter`
- [ ] Open content drive, narrow the viewport — filter row wraps cleanly, multiselects keep their 14rem min width
- [ ] Toggle selections via store/state — chip switches between inactive (white + slate border) and active (primary-100 tint) and the icon flips chevron-down ↔ close
- [ ] Click the close icon on an active chip — `removed` fires, `clicked` does NOT (propagation stopped)
- [ ] Inspect labels: empty selections → just title; 1 → `"Title: A"`; 2 → `"Title: A, B"`; 3+ → `"Title: A and N more"` (with i18n)

This PR fixes: #33524


dario-daza (approved):

core-web/libs/portlets/dot-content-drive/ui/src/lib/dot-chip-filter/dot-chip-filter.component.html
```diff
@@ -0,0 +1,11 @@
+<span class="truncate">
+    <span class="font-semibold" data-testid="chip-title">{{ title() }}</span>
+    @if (active()) {
+        <span data-testid="chip-values">: {{ valuesLabel() }}</span>
+    }
+</span>
+@if (active()) {
+    <i class="pi pi-times text-sm leading-none" (click)="onRemove($event)"></i>
```
oidacra: The close icon `<i>` has a `(click)` handler but isn't reachable via keyboard — no `role`, `tabindex`, or keydown handler, and the host click emits `clicked` rather than `removed`. Keyboard-only users can't remove a filter. Consider replacing the `<i>` with a `<button type="button">` so it inherits native keyboard handling and focus, plus an `aria-label` for the action. While here, the chip itself is also a click target with no native semantics — adding `role="button"`, `tabindex="0"`, and `keydown.enter`/`keydown.space` on the host would make the whole component WCAG AA-compliant.

core-web/libs/portlets/dot-content-drive/ui/src/lib/dot-chip-filter/dot-chip-filter.component.ts
```diff
@@ -0,0 +1,53 @@
...
+
+const INACTIVE_CLASSES = 'bg-white text-slate-600 border border-slate-200 hover:border-primary-400';
+
+const ACTIVE_CLASSES = 'bg-primary-100 text-primary-900 hover:bg-primary-200';
+
+@Component({
+    selector: 'dot-chip-filter',
+    standalone: true,
+    imports: [],
+    templateUrl: './dot-chip-filter.component.html',
```
oidacra, line 17: Per `ANGULAR_STANDARDS.md`, don't set `standalone: true` in the decorator — it's implied by default in Angular 19+ and the project convention is to omit it. Just drop this line.

oidacra (approved):

core-web/libs/portlets/dot-content-drive/ui/src/lib/dot-chip-filter/dot-chip-filter.component.html
zJaaal: done

Issue #35403, opened 2026-04-21
Block Editor: Category fields missing from nested contentlet response (hydration gap in StoryBlockAPIImpl)


rjvelazco:
## Problem Statement

When a Block Editor field embeds a contentlet that has a **Category field**, the Category field is completely missing from the API response. Everything else comes back fine (identifier, inode, title, modDate, contentType, Story Block fields, Binary fields, etc.) but Category fields are silently dropped.

This affects headless use cases (GraphQL / REST) where consumers rely on the full field data of embedded contentlets.

**What's happening under the hood:** The Block Editor hydration path in `StoryBlockAPIImpl.refreshContentlet()` rebuilds embedded contentlets by iterating over fields and calling `contentlet.get(field.variable())`, then dropping any `null` values. For `CategoryField`, that call always returns `null` because categories live in the `category`/`tree` tables, not on the contentlet property map. The `CategoryViewStrategy` (which calls `categoryAPI.getParents(...)`) is never called in this path, so categories never make it into the serialized JSON.

This is the same class of bug as #32973 ("Missing properties on Contentlet Block on Block Editor result - Missing Thumbnail"), which fixed missing `hasTitleImage`/`titleImage` properties by adding explicit hydration in `loadCommonContentletProps`. The same fix pattern needs to be applied for Category fields, and likely Tag and Relationship fields too.

**Affected file:** `dotCMS/src/main/java/com/dotcms/contenttype/business/StoryBlockAPIImpl.java`
- `refreshContentlet()` iterates fields and drops nulls, so Category fields are always dropped
- `loadCommonContentletProps()` handles special field types but has no Category handling

**How the fix should look:** `CategoryViewStrategy.java` has the reference implementation using `toolBox.categoryAPI.getParents(contentlet, user, true)` to fetch and map category data by field variable.

## Steps to Reproduce

1. Create a Content Type (e.g. `UsDisclaimer`) with a Category field (variable: `targeting`)
2. Create a second Content Type with a Block Editor field
3. Create a `UsDisclaimer` contentlet with categories assigned to `targeting`
4. Embed that contentlet inside a Block Editor contentlet
5. Fetch the Block Editor contentlet via the headless API (GraphQL or REST)

**Expected:** The nested `UsDisclaimer` JSON includes the `targeting` Category field with its values.

**Actual:** The `targeting` field is absent from the nested JSON. No error is thrown.

## Acceptance Criteria

[...]

PR #35469, opened 2026-04-27
fix(block-editor): guard StoryBlock JSON parsing against non-object scalars

### LINKED ISSUE #35408: Block Editor: Nested contentlet updates don't propagate through relationship chains
## Problem Statement

When a Block Editor field embeds a contentlet that sits behind a relationship chain, updates to the deeply nested content are not reflected in the page JSON response. Everything looks fine after the initial publish, but after editing the nested content the page keeps serving the old (stale) version.

The issue is confirmed to be on the backend — most likely a cache or hydration problem where the re-hydration of relationship fields in Block Editor doesn't propagate changes up the chain.

**Workaround:** Re-publishing the contentlet that holds the relationship field forces the page to reflect the latest content.

A video showing the full reproduction is available in the Slack thread (too large to upload to GitHub): https://dotcms.slack.com/archives/C08F8NZRK9S/p1776800139499559?thread_ts=1776782935.910469&cid=C08F8NZRK9S

## Steps to Reproduce

**Setup — you'll need 3 content types:**
1. **Post** — with a title and a Block Editor field
2. **News** — with a title and a Block Editor field
3. **Similar News** — with a title and a 1:N relationship field pointing to `News`

**Reproduce the issue:**
1. Create a **News** contentlet and add any contentlet to its Block Editor field
2. Create a **Similar News** contentlet and link the News contentlet via the relationship field
3. Create a **Post** contentlet and add the Similar News contentlet to its Block Editor field
4. Add the Post contentlet to a page and check the page JSON response — content looks correct
5. Go back and update the contentlet inside the **News** Block Editor field
6. Check the page JSON response again

**Expected:** The updated content from step 5 is reflected in the page response.

**Actual:** The page still returns the old content. The change only shows up after re-publishing the **Similar News** contentlet (the one holding the relationship field).

## Acceptance Criteria

- [ ] When a contentlet nested inside a Block Editor is updated, the change is reflected in the page JSON response without requiring a manual re-publish of intermediate contentlets
- [ ] The fix covers the full relationship chain: `Page -> Block Editor -> Relationship field -> Block Editor -> nested contentlet`
- [ ] Cache invalidation (if applicable) is triggered correctly when any contentlet in the chain is published
- [ ] An integration test is added covering this multi-level nesting scenario
- [ ] No regression in existing Block Editor or relationship field behavior

## dotCMS Version

Reproducible on current `main` branch.

## Severity

## Links

- Slack thread with reproduction video: https://dotcms.slack.com/archives/C08F8NZRK9S/p1776800139499559?thread_ts=1776782935.910469&cid=C08F8NZRK9S
- Helpdesk tickets: https://helpdesk.dotcms.com/a/tickets/36444 and https://helpdesk.dotcms.com/a/tickets/36152

fmontes:
Fixes regression from #35412: `/api/content/_search` throws `Cannot deserialize value of type LinkedHashMap from Integer` when a Story Block references a contentlet with scalar field values.

## Fix

**1. JSON guard.** `JsonUtil.isValidJSON()` returns true for scalars (`42`, `"foo"`, `[1,2]`). Replaced with `isJsonObject()` (valid JSON + starts with `{`) at the 4 call sites feeding `toMap()`. HTML strings correctly skip refresh (no `{`).

**2. Failure isolation.** One bad contentlet was killing the whole search response. Now:
- `ContentletTransformer.refreshStoryBlockReferences` catches/logs — bad contentlet returns un-refreshed.
- `StoryBlockAPIImpl.isRefreshed` isolates each child block.
- `StoryBlockAPIImpl.refreshContentlet` isolates each field (falls back to raw value).

## QA

Re-run QA from #35408 (PR #35412) and #35403 (PR #35413).

Plus: via Content REST API, create a contentlet with **raw HTML** as the Block Editor field value (e.g. `<p>hello</p>`). Confirm create + update + `_search` + edit-UI all work without `MismatchedInputException` in logs.

## Tests

- `test_refreshStoryBlockValueReferences_with_non_object_json_scalars` — numeric/string/boolean/array/HTML inputs
- `test_refreshStoryBlockValueReferences_isolates_bad_child_block` — inner isolation
- `test_refreshReferences_does_not_throw_on_malformed_nested_block` — API-level resilience

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/contenttype/business/StoryBlockAPIImpl.java
+++ b/dotCMS/src/main/java/com/dotcms/contenttype/business/StoryBlockAPIImpl.java
@@ -227,15 +227,23 @@ public class StoryBlockAPIImpl implements StoryBlockAPI {
         boolean refreshed = false;
         for (final Map<String, Object> contentMap : contentsMap) {
             if (UtilMethods.isSet(contentMap)) {
-                final String type = contentMap.get(TYPE_KEY).toString();
-                if (allowedTypes.contains(type)) { // if somebody adds a story block to itself, we don't want to refresh it
+                // Isolate per-block failures so that one bad nested reference
+                // does not prevent the rest of the Story Block from refreshing.
+                try {
+                    final String type = contentMap.get(TYPE_KEY).toString();
+                    if (allowedTypes.contains(type)) { // if somebody adds a story block to itself, we don't want to refresh it
 
-                    refreshed |= this.refreshStoryBlockMap(contentMap, parentContentletIdentifier);
-                } else {
-                    final Object nestedContent = contentMap.get(CONTENT_KEY);
-                    if (nestedContent instanceof List) {
-                        refreshed |= this.isRefreshed(parentContentletIdentifier, (List<Map<String, Object>>) nestedContent);
+                        refreshed |= this.refreshStoryBlockMap(contentMap, parentContentletIdentifier);
+                    } else {
+                        final Object nestedContent = contentMap.get(CONTENT_KEY);
+                        if (nestedContent instanceof List) {
+                            refreshed |= this.isRefreshed(parentContentletIdentifier, (List<Map<String, Object>>) nestedContent);
+                        }
                     }
+                } catch (final Exception e) {
+                    Logger.warnAndDebug(StoryBlockAPIImpl.class, String.format(
+                            "Skipping Story Block child while refreshing parent '%s': %s",
+                            parentContentletIdentifier, ExceptionUtil.getErrorMessage(e)), e);
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


dotCMS/src/main/java/com/dotcms/contenttype/business/StoryBlockAPIImpl.java
```diff
@@ -570,6 +578,22 @@ private static void addDependencies(final ImmutableList.Builder<StoryBlockDepend
...
+        if (value == null) {
+            return false;
+        }
+        final String trimmed = value.trim();
+        return trimmed.startsWith("{") && JsonUtil.isValidJSON(trimmed);
+    }
+
     @Override
     @SuppressWarnings("unchecked")
     public LinkedHashMap<String, Object> toMap(final Object blockEditorValue) throws JsonProcessingException {
```
wezell, line 599: any way this can return a `LinkedHashMap<String, Serializable>`?

fmontes: Not in this PR — and probably not at all without a separate cleanup. Jackson populates the map via `readValue(json, LinkedHashMap.class)`, so the actual values are whatever the JSON contained (`String`, `Integer`, nested `LinkedHashMap`, `ArrayList`, etc.). Declaring the return as `<String, Serializable>` would be an unchecked cast — the runtime contents wouldn't change, and any caller that started trusting the type could silently break if a non-Serializable value ever leaked in (e.g. a `Contentlet` or a custom POJO without `Serializable`).

It's also a public interface change on `StoryBlockAPI.toMap` that affects every caller and is unrelated to this regression. Happy to open a follow-up ticket if the goal is making cached Story Block payloads cluster-safe — but the right fix there is at the producer side (ensuring inserted values are Serializable), not retyping the API.

wezell: Claude says not without a lot of work.

wezell (approved):
Not going to be worse

PR #35493, opened 2026-04-28
feat(uve): navigate to new edit content editor from Go to Full Editor when the new edit content is active

### LINKED ISSUE #33235: [DEFECT]
### Problem Statement

When then new edit content is enabled for a content type, the expectation is that whenever and wherever that content type is accessed that the new edit content experience will be show as per the user selection. At the moment, when you enable the new edit content for a content type and try to access it via the UVE, the new edit content experience is not launched, instead we show the old edit content. 

### Steps to Reproduce

Enable new edit content for a content type. 
Attempt to open said content type from UVE. 
Notice old edit content is shown as opposed to the new one like the user selected. 

### Acceptance Criteria

When then new edit content is enabled for a content type, the expectation is that whenever and wherever that content type is accessed that the new edit content experience will be show as per the user selection.

### dotCMS Version

evergreen

### Severity

Low - Minor issue or cosmetic

adrianjm-dotCMS:
## Summary

- When the user clicks **Go to Full Editor** in UVE, navigate to `/content/:inode` if the content type has `FEATURE_FLAG_CONTENT_EDITOR2_ENABLED` in its metadata; otherwise fall back to the existing dialog behavior
- Update the UVE breadcrumb to use the full `/dotAdmin/#/edit-page/content?...` URL so users can navigate back to the exact editor page from the breadcrumb trail
- Fix breadcrumb deduplication to search the entire trail (not just the last crumb), enabling correct truncation when navigating back to a previous UVE page
- Fix `contentEdit` regex in breadcrumb utils to prevent `/edit-page/content?...` UVE routes from incorrectly triggering the replace-last-crumb rule
- Add `untracked()` around `breadcrumbs` read inside the navigation effect to prevent infinite reactive loops

https://github.com/user-attachments/assets/c4f0269e-aeda-4535-a765-115896479c69

## Changed Files

- `libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/edit-ema-editor.component.ts` — core "Go to Full Editor" routing logic
- `libs/portlets/edit-ema/portlet/src/lib/dot-ema-shell/dot-ema-shell.component.ts` — UVE breadcrumb now uses full navigable URL
- `libs/global-store/src/lib/features/breadcrumb/breadcrumb.feature.ts` — full-trail deduplication + `untracked()` fix
- `libs/global-store/src/lib/features/breadcrumb/breadcrumb.utils.ts` — regex fix for UVE routes

## Acceptance Criteria

- [x] Clicking "Go to Full Editor" navigates to `/content/:inode` when `FEATURE_FLAG_CONTENT_EDITOR2_ENABLED` is `true` in the content type metadata
- [x] Falls back to the existing dialog when the flag is absent or `false`
- [x] Falls back to dialog when the content type is not in the cache
- [x] Does nothing when the contentlet has no inode
- [x] UVE breadcrumb is clickable and navigates back to the correct editor page
- [x] Navigating back to a previously visited UVE page truncates the breadcrumb trail correctly

## Test Plan

- [x] Unit tests added for all `handleOpenFullEditor` branches (new editor, dialog fallback, cache miss, no inode)
- [x] Breadcrumb tests updated and new truncation test added
- [x] All affected tests pass locally

Closes #33235

### DIFF
--- a/core-web/libs/global-store/src/lib/features/breadcrumb/breadcrumb.utils.spec.ts
+++ b/core-web/libs/global-store/src/lib/features/breadcrumb/breadcrumb.utils.spec.ts
@@ -243,7 +243,7 @@ describe('shouldReplaceLastCrumb', () => {
             ).toBe(true);
         });
 
-        it('should return false when both items match /edit-page/content?{params}', () => {
+        it('should return true when both items match /edit-page/content?{params} (via uvePage rule)', () => {
             expect(
                 shouldReplaceLastCrumb(
                     { label: 'New', url: '/dotAdmin/#/edit-page/content?url=page2' },
--- a/core-web/libs/global-store/src/lib/features/breadcrumb/breadcrumb.utils.spec.ts
+++ b/core-web/libs/global-store/src/lib/features/breadcrumb/breadcrumb.utils.spec.ts
@@ -243,13 +243,13 @@ describe('shouldReplaceLastCrumb', () => {
             ).toBe(true);
         });
 
-        it('should return true when both items match /edit-page/content?{params}', () => {
+        it('should return false when both items match /edit-page/content?{params}', () => {
             expect(
                 shouldReplaceLastCrumb(
                     { label: 'New', url: '/dotAdmin/#/edit-page/content?url=page2' },
                     { label: 'Old', url: '/dotAdmin/#/edit-page/content?url=page1' }
                 )
-            ).toBe(true);
+            ).toBe(false);
         });
 
         it('should return false when only new item matches content-edit', () => {
--- a/core-web/libs/global-store/src/lib/features/breadcrumb/breadcrumb.utils.spec.ts
+++ b/core-web/libs/global-store/src/lib/features/breadcrumb/breadcrumb.utils.spec.ts
@@ -249,7 +249,7 @@ describe('shouldReplaceLastCrumb', () => {
                     { label: 'New', url: '/dotAdmin/#/edit-page/content?url=page2' },
                     { label: 'Old', url: '/dotAdmin/#/edit-page/content?url=page1' }
                 )
-            ).toBe(false);
+            ).toBe(true);
         });
 
         it('should return false when only new item matches content-edit', () => {
--- a/core-web/libs/global-store/src/lib/features/breadcrumb/breadcrumb.utils.ts
+++ b/core-web/libs/global-store/src/lib/features/breadcrumb/breadcrumb.utils.ts
@@ -163,10 +163,19 @@ interface ReplaceLastCrumbRule {
  */
 const REPLACE_LAST_CRUMB_RULES: Record<string, ReplaceLastCrumbRule> = {
     contentEdit: {
+        // Match only root content routes (e.g. /content/{inode}, /content?filter=...),
+        // but not nested routes like /edit-page/content used by UVE breadcrumbs.
         test: (item, last) => {
-            // Match only root content routes (e.g. /content/{inode}, /content?filter=...),
-            // but not nested routes like /edit-page/content used by UVE breadcrumbs.
             const regex = /^\/content(?:[/?].+)?$/;
+            const normalize = (url: string | undefined) => (url ?? '').replace(/^.*#/, '');
+            return regex.test(normalize(item.url)) && regex.test(normalize(last.url));
+        }
+    },
+    uvePage: {
+        // When navigating between different pages in UVE, replace the last crumb
+        // so the trail stays as Pages > [Current Page] instead of accumulating every visited page.
+        test: (item, last) => {
+            const regex = /\/edit-page\/content/;
             const normalize = (url: string | undefined) => (url ?? '').replace(/^.*#/, '');
             return regex.test(normalize(item.url)) && regex.test(normalize(last.url));
         }
--- a/core-web/libs/global-store/src/lib/features/breadcrumb/breadcrumb.utils.ts
+++ b/core-web/libs/global-store/src/lib/features/breadcrumb/breadcrumb.utils.ts
@@ -164,7 +164,9 @@ interface ReplaceLastCrumbRule {
 const REPLACE_LAST_CRUMB_RULES: Record<string, ReplaceLastCrumbRule> = {
     contentEdit: {
         test: (item, last) => {
-            const regex = /\/content[/?].+/;
+            // Match only root content routes (e.g. /content/{inode}, /content?filter=...),
+            // but not nested routes like /edit-page/content used by UVE breadcrumbs.
+            const regex = /^\/content(?:[/?].+)?$/;
             const normalize = (url: string | undefined) => (url ?? '').replace(/^.*#/, '');
             return regex.test(normalize(item.url)) && regex.test(normalize(last.url));
         }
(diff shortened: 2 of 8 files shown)


core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/edit-ema-editor.component.ts
```diff
@@ -1112,7 +1115,17 @@ export class EditEmaEditorComponent implements OnInit, OnDestroy, AfterViewInit
...
+        if (!contentlet?.inode) {
+            return;
+        }
+
+        const contentType = this.uveStore.contentTypeCache()[contentlet.contentType];
+        const isNewEditor =
+            contentType?.metadata?.[FeaturedFlags.FEATURE_FLAG_CONTENT_EDITOR2_ENABLED];
+
+        if (isNewEditor) {
+            this.#router.navigate(['content', contentlet.inode]);
```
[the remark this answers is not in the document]
adrianjm-dotCMS: Fixed — updated to `['/content', contentlet.inode]` with explicit leading slash.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/edit-ema-editor.component.ts
```diff
@@ -1112,7 +1115,17 @@ export class EditEmaEditorComponent implements OnInit, OnDestroy, AfterViewInit
...
         const { contentlet } = this.$contentletEditData();
 
-        if (contentlet?.inode) {
+        if (!contentlet?.inode) {
+            return;
+        }
+
+        const contentType = this.uveStore.contentTypeCache()[contentlet.contentType];
+        const isNewEditor =
+            contentType?.metadata?.[FeaturedFlags.FEATURE_FLAG_CONTENT_EDITOR2_ENABLED];
```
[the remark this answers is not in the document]
adrianjm-dotCMS: Fixed — updated to `=== true` for strict comparison.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/edit-ema-editor.component.spec.ts
```diff
@@ -2449,6 +2449,83 @@ describe('EditEmaEditorComponent', () => {
...
+                            metadata: {
+                                [FeaturedFlags.FEATURE_FLAG_CONTENT_EDITOR2_ENABLED]: true
+                            }
+                        } as never
+                    }
+                });
+
+                spectator.component['handleOpenFullEditor']();
+
+                expect(router.navigate).toHaveBeenCalledWith(['content', 'contentlet-inode-123']);
```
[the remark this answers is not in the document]
adrianjm-dotCMS: Fixed — test expectation updated to `['/content', 'contentlet-inode-123']`.

core-web/libs/global-store/src/lib/features/breadcrumb/breadcrumb.utils.spec.ts
```diff
@@ -243,7 +243,7 @@ describe('shouldReplaceLastCrumb', () => {
             ).toBe(true);
         });
 
-        it('should return true when both items match /edit-page/content?{params}', () => {
+        it('should return false when both items match /edit-page/content?{params}', () => {
```
oidacra: The `it()` title was flipped from 'should return true' to 'should return false', but the assertion body is unchanged and still asserts `.toBe(true)`. The test passes only because the new `uvePage` rule keeps `shouldReplaceLastCrumb` returning `true` for these inputs — but the title now contradicts the assertion. Consider either reverting the title (since the public behavior is unchanged) or splitting into two tests: one verifying that the narrowed `contentEdit` rule no longer matches, and one verifying that the new `uvePage` rule does.

core-web/libs/global-store/src/lib/features/breadcrumb/breadcrumb.utils.ts
```diff
@@ -163,8 +163,19 @@ interface ReplaceLastCrumbRule {
  */
 const REPLACE_LAST_CRUMB_RULES: Record<string, ReplaceLastCrumbRule> = {
     contentEdit: {
+        // Match only root content routes (e.g. /content/{inode}, /content?filter=...),
+        // but not nested routes like /edit-page/content used by UVE breadcrumbs.
```
oidacra, line 167: The regex `^/content(?:[/?].+)?$` still matches `/content?filter=...`, which is owned by the `contentFilter` ROUTE_HANDLER. If `REPLACE_LAST_CRUMB_RULES` runs before the route handler dispatch, navigating from `/content/abc` to `/content?filter=blog` will replace the content-item crumb instead of appending the filter crumb. Tightening to `^/content/[^?#]+` keeps `contentEdit` exclusively for `/content/{inode}` routes and leaves filter URLs to their dedicated handler. A unit test for the `/content/abc → /content?filter=...` transition would lock this in.

core-web/libs/global-store/src/lib/features/breadcrumb/breadcrumb.utils.ts
```diff
@@ -163,8 +163,19 @@ interface ReplaceLastCrumbRule {
  */
 const REPLACE_LAST_CRUMB_RULES: Record<string, ReplaceLastCrumbRule> = {
     contentEdit: {
+        // Match only root content routes (e.g. /content/{inode}, /content?filter=...),
```
oidacra: Minor: this comment lists `/content?filter=...` as an example, but that URL is owned by the `contentFilter` ROUTE_HANDLER, not by this replace-last-crumb rule. Consider rewording to: 'Match content edit routes (e.g. `/content/{inode}`, `/content/{inode}?tab=…`), but not nested routes like `/edit-page/content` used by UVE breadcrumbs.'

core-web/libs/global-store/src/lib/features/breadcrumb/breadcrumb.utils.spec.ts
adrianjm-dotCMS: Fixed — reverted the title to should return true when both items match /edit-page/content?{params} (via uvePage rule). The public behavior is unchanged (still returns true); the uvePage rule now handles UVE routes instead of contentEdit, but the outcome is the same.

core-web/libs/global-store/src/lib/features/breadcrumb/breadcrumb.utils.ts
adrianjm-dotCMS: The scenario described isn't reachable in practice. `shouldReplaceLastCrumb` is only called inside `addNewBreadcrumb`, and none of the `addNewBreadcrumb` call sites ever pass a `/content?filter=...` URL. Filter URLs are added exclusively via the `contentFilter` ROUTE_HANDLER inside `_processUrl` (router navigation events) — a completely separate code path that bypasses `shouldReplaceLastCrumb` entirely. Keeping the regex as-is.

core-web/libs/global-store/src/lib/features/breadcrumb/breadcrumb.utils.ts
adrianjm-dotCMS: Fixed — updated the comment to use `/content/{inode}?tab=...` as the example instead of `/content?filter=...`, which belongs to the `contentFilter` ROUTE_HANDLER.

zJaaal (changes requested):
I get the idea of let the user use the new Edit Content

But before making these changes we need to review the solution with @fmontes there are nuances that we have to consider

Issue #35048, opened 2026-03-20
Make Analytics Infrastructure Production-Ready


freddyDOTCMS:
## Summary

Turn the dotCMS analytics infrastructure POC (#34824) into a production-ready system. The architecture consists of a Browser/JS SDK sending events and queries through dotCMS (single public entrypoint) to an Analytics Microservice backed by ClickHouse for storage and multi-tenant isolation.

## Background

The current system is a proof-of-concept that needs hardening across 8 phases before it can go to production:

- **Phase 1 — Foundation**: Code refactoring, ClickHouse optimizations, test coverage, structured logging, CI/CD
- **Phase 2 — UA Parser**: Replace ClickHouse lookup tables with Java-based UA parsing
- **Phase 3 — Security**: Input validation, automated user provisioning script
- **Phase 4 — Reliability**: Health checks, async timeouts, 503 backpressure, graceful shutdown
- **Phase 5 — Observability**: Connection error monitoring, per-customer rate limiting
- **Phase 6 — Performance**: Externalize configuration, query result caching
- **Phase 7 — Features**: Conversion dashboard endpoints, engagement session summary endpoints
- **Phase 8 — Polish**: Docker image version pinning

## Architecture

- **Browser/JS SDK** sends event registration requests and analytics queries — does not talk to ClickHouse directly
- **dotCMS** acts as single public entrypoint; routes to the microservice and injects customer credentials (`customer_id` + password + environment)
- **Analytics Microservice**: Ingester validates and inserts events; Query API builds SQL programmatically and executes against ClickHouse
- **ClickHouse** stores raw events and aggregated views; enforces multi-tenant isolation via ClickHouse users and Row Policies so every query is automatically filtered by `customer_id`

## Acceptance criteria

- [ ] All 32 child tasks completed and merged
- [ ] System is observable (structured logs, metrics, health checks)
- [ ] Multi-tenancy enforced at ClickHouse layer via Row Policies
- [ ] Full test coverage: unit, integration (Testcontainers), web layer, performance (Gatling)
- [ ] CI/CD pipeline validates every PR
- [ ] Conversion and engagement dashboard endpoints live
- [ ] No hardcoded secrets or configuration values

## References

- POC: #34824

PR #35524, opened 2026-04-30
feat(Content Analytics) dotCMS/core#35066 : Add health checks for Content Analytics across dotCMS and CA Event Manager.

jcastro-dotcms:
## What does this PR do?

Implements the **dotCMS-side health checks** for Content Analytics described in #35066. The system status monitor endpoint (`/api/v1/system/status`) now evaluates Content Analytics readiness through three sequential checks rather than firing a write event to the analytics pipeline.

### New: `Health` enum (`com.dotcms.rest.api.v1.system.monitor`)
Replaces the previous dependency on `ExperimentsAPI.Health` with a dedicated, monitor-scoped enum that makes the distinction between _not set up_ and _misconfigured_ explicit for operators:

| Value | Meaning |
|---|---|
| `OK` | App configured, all env vars present, CA Event Manager reachable |
| `NOT_CONFIGURED` | `dotContentAnalytics-config` app has no secrets for this Site — feature is simply not enabled |
| `CONFIGURATION_ERROR` | App is partially set up but env vars are missing or the `/v1/health` probe failed |

### Updated: `MonitorHelper.isContentAnalytics()`
Replaces the old `EventLogRunnable.sendTestEvent()` approach (which wrote a real event to the analytics pipeline) with three lightweight, read-only checks:

1. **App configuration** — `ContentAnalyticsUtil.getAppSecrets(host)` must return a non-empty secrets map for the current Site. Returns `NOT_CONFIGURED` if absent.
2. **Required environment variables** — `DOT_ANALYTICS_BASE_URL`, `DOT_ANALYTICS_CUSTOMER_ID`, `DOT_ANALYTICS_PASSWORD`, and `DOT_ANALYTICS_ENVIRONMENT` must all be non-empty. Returns `CONFIGURATION_ERROR` with a `WARN`-level log listing every missing property.
3. **Service connectivity** — delegates to `EventAnalyticsProxyHelper.healthCheck()`. Returns `OK` on 2xx + non-empty body, `CONFIGURATION_ERROR` otherwise.

### Updated: `EventAnalyticsProxyHelper`
- **`healthCheck()` (new public method)** — issues an unauthenticated GET to `{DOT_ANALYTICS_BASE_URL}/v1/health`. Skips the HTTP call entirely (logs at DEBUG) when `DOT_ANALYTICS_BASE_URL` is not configured, so sites without Content Analytics don't generate noisy ERROR logs on every health poll.
- **`stripTrailingSlash()` (new private helper)** — extracted from the inline slash-stripping that was copy-pasted between `buildUpstreamUrl` and `healthCheck`.
- **Config constants** promoted from package-private to `public` so `MonitorHelper` can import them via static imports, keeping the string literals in one place.

### Updated: `MonitorStats.getContentAnalytics()` Javadoc
Updated reference from the removed `ExperimentsAPI.Health` to the new `Health` enum.

### New tests in `MonitorHelperTest`
Added coverage for two previously untested, self-contained units:

- **`isStartedUp()`** — verifies `false` when `dotcms.started.up` is absent and `true` when it is present (with proper cleanup via `try/finally`)
- **`FileSystemTest` inner class** — three cases: writable temp directory (happy path), path with a trailing separator (constructor normalization), and a regular file used as the base path (forces `mkdirs()` to fail → `false`)

> **Note:** `isContentAnalytics()` is `private` and relies on three static call sites (`WebAPILocator`, `ContentAnalyticsUtil`, `EventAnalyticsProxyHelper`). Unit tests for that method require a static-mock framework or a live dotCMS context; they are deferred to a follow-up integration test.

## Testing

- [ ] `/api/v1/system/status` returns `"contentAnalytics": "NOT_CONFIGURED"` on a Site with no Content Analytics app secrets
- [ ] Returns `"contentAnalytics": "CONFIGURATION_ERROR"` when secrets exist but one or more of the four env vars is unset
- [ ] Returns `"contentAnalytics": "CONFIGURATION_ERROR"` when env vars are set but the CA Event Manager is unreachable
- [ ] Returns `"contentAnalytics": "OK"` when fully configured and the CA Event Manager `/v1/health` responds 2xx
- [ ] `MonitorHelperTest` passes (5 new tests + 16 existing)

## Related
- Closes #35066
- Part of epic #35048

This PR fixes: #35066


jcastro-dotcms:
> **Claude finished @jcastro-dotcms's task in 1m 31s** —— [View job](https://github.com/dotCMS/core/actions/runs/25182101032)
> 
> ### Code Review
> * [x]  Gather context
> * [x]  Read changed files
> * [x]  Analyze code
> * [x]  Post review findings
> 
> ### Issues Found
> #### 1. Wrong `Health` value when env vars are missing — `MonitorHelper.java:228-233`
> Check 2 returns `NOT_CONFIGURED` when the app secrets are present but env vars are absent. That's semantically wrong. The `Health` enum you added is explicit about the distinction:
> 
> * `NOT_CONFIGURED` — "No action is required unless the feature is expected to be active"
> * `CONFIGURATION_ERROR` — "partially or incorrectly configured"
> 
> If secrets exist, the operator has started configuring the feature. Missing env vars = `CONFIGURATION_ERROR`, not `NOT_CONFIGURED`.
> 
> ```java
> // Current (wrong) — operator sees "not configured" but the app IS partially set up
> if (UtilMethods.isNotSet(baseUrl) || ...) {
>     return Health.NOT_CONFIGURED.name();   // ← should be CONFIGURATION_ERROR
> }
> ```
> 
> [Fix this →](https://claude.ai/code?q=In%20dotCMS%2Fsrc%2Fmain%2Fjava%2Fcom%2Fdotcms%2Frest%2Fapi%2Fv1%2Fsystem%2Fmonitor%2FMonitorHelper.java%20on%20branch%20issue-35066-Add-health-checks-for-Content-Analytics-across-dotCMS-and-CA-Event-Manager%2C%20in%20the%20isContentAnalytics%20method%2C%20Check%202%20(missing%20required%20env%20vars)%20incorrectly%20returns%20Health.NOT_CONFIGURED.name()%20when%20app%20secrets%20are%20present%20but%20env%20vars%20are%20missing.%20Change%20it%20to%20return%20Health.CONFIGURATION_ERROR.name()%20instead%2C%20since%20the%20app%20IS%20configured%20but%20env%20vars%20are%20missing.&repo=dotCMS%2Fcore)
> 
> #### 2. Misleading comment vs. actual behavior — `MonitorHelper.java:235`
> The comment says **"using customer credentials"** but `healthCheck()` sends **no `Authorization` header** — it is explicitly unauthenticated. The `healthCheck()` javadoc correctly says "unauthenticated GET", so the inline comment in `MonitorHelper` contradicts it.
> 
> ```java
> // Check 3: Service connectivity using customer credentials   ← misleading
> return EventAnalyticsProxyHelper.healthCheck()
> ```
> 
> #### 3. Log message omits one env var — `MonitorHelper.java:230-232`
> The error message lists three properties but four are checked. `DOT_ANALYTICS_PASSWORD` is silently omitted from the log, making it harder to diagnose why the check failed.
> 
> ```java
> Logger.error(this, "Content Analytics health check: missing required env vars " +
>         "(DOT_ANALYTICS_BASE_URL, DOT_ANALYTICS_CUSTOMER_ID, DOT_ANALYTICS_ENVIRONMENT)");
> //                                                            ↑ DOT_ANALYTICS_PASSWORD is missing from this list
> ```
> 
> #### 4. ERROR log for an expected/normal condition — `EventAnalyticsProxyHelper.java:179-181`
> When `DOT_ANALYTICS_BASE_URL` is not configured, `healthCheck()` logs at `ERROR`. But "not configured" is a normal operational state (feature not enabled), not an error. This will produce noisy logs for every customer who doesn't use CA. Should be `Logger.info` or `Logger.debug`.
> 
> ```java
> Logger.error(EventAnalyticsProxyHelper.class,
>         "Health check skipped: '" + DOT_ANALYTICS_BASE_URL + "' is not configured");
> ```
> 
> Note: In practice, this path is also unreachable via `isContentAnalytics()` (Check 2 already validates the URL), so this is a defensive path — but if `healthCheck()` is ever called directly, the log level matters.
> 
> #### 5. Stale Javadoc in `MonitorStats.java:62` (untouched file)
> `getContentAnalytics()` still references the old enum:
> 
> ```java
> // Return the content analytics status: see {@link com.dotcms.experiments.business.ExperimentsAPI.Health}
> ```
> 
> This should now reference `com.dotcms.rest.api.v1.system.monitor.Health` instead.
> 
> #### 6. No unit tests for new code paths
> `MonitorHelperTest.java` has no tests for `isContentAnalytics()` and `EventAnalyticsProxyHelper.healthCheck()`. The three-step conditional logic (secrets present → env vars set → HTTP probe) and the two `Health` return values for Check 2 (the wrong one noted above) are good candidates for unit coverage.
> 
> #### Minor: URL-stripping duplication — `EventAnalyticsProxyHelper.java:184-185`
> Trailing-slash stripping is copy-pasted from `buildUpstreamUrl()` into `healthCheck()`. Small duplication, but it means a future change to one won't automatically apply to the other. Could be a private `stripTrailingSlash(String url)` helper.

@claude regarding your feedback:

All six items are valid, but with a few cnosiderations:

  - 1–5: All genuine bugs or inaccuracies. Fixed. 
  - 6 (tests): There's currently no way to test dotCMS against the Dot Event Manager service. We'll leave this for later. However, some additional tests were added to the `MonitorHelperTest` class as they were easy to implement.
  - 7 (duplication): Minor cleanup. Fixed.

freddyDOTCMS (approved):

PR #35528, opened 2026-04-30
include style editor schemas in the GQL response

### LINKED ISSUE #35454: feat(sdk-client): gate fetchStyleEditorSchemas on UVE edit mode to avoid unnecessary network calls
## Description

`fetchStyleEditorSchemas` in `@dotcms/client` was firing an HTTP round-trip to `/api/v1/page/{pageId}/contenttype-schema` on every `client.page.get()` call - including live, preview, and production pages - even though the downstream consumer (`registerStyleEditorSchemas` in `@dotcms/uve`) discards the result in any non-edit mode. The only guard was an SSR check (`typeof window === 'undefined'`).

The issue originally proposed an inline iframe + mode guard inside `fetchStyleEditorSchemas` as a short-term fix, with the long-term fix being to embed schemas in the GraphQL page response.

**This PR implements the long-term fix directly, skipping the workaround entirely.**

---

### Backend

- **Added `styleEditorSchemas` to the GraphQL page response** via a new `StyleEditorSchemasDataFetcher`. The field is only populated when `FEATURE_FLAG_UVE_STYLE_EDITOR` is enabled and the page is rendered in `EDIT_MODE`. In all other modes the field is absent.
- **Added `styleEditorSchemas` to the REST page API response** (`/api/v1/page/render`) following the same gate: `EDIT_MODE` + feature flag. The schemas are computed once in `HTMLPageAssetRenderedBuilder` and reused for both the rendered HTML UVE script injection and the JSON response - no double computation.
- **Removed the `/api/v1/page/{pageId}/contenttype-schema` endpoint** entirely (`PageResource`, `ResponseEntityContentTypeSchemaView`, OpenAPI spec). The dedicated round-trip is no longer needed.

### Frontend

- **`@dotcms/client`**: `fetchStyleEditorSchemas` is no longer a standalone network call. `page.get()` now includes `styleEditorSchemas` directly in its response when available, with no separate HTTP request.
- **UVE Store (`@ngrx/signals`)**: `setPageAsset` seeds `editorStyleSchemas` from the page response on every page load. The `$styleSchema` computed signal is simplified to a single source of truth, no more two-source fallback between `editorStyleSchemas` and `pageAsset.page.styleEditorSchemas`.
- **`dot-uve.js`**: Updated to consume schemas from the page response instead of triggering a separate fetch.
- **Types (`@dotcms/types`)**: `styleEditorSchemas` added as an optional field on `DotCMSPage`.

---

## Result

| Before | After |
|--------|-------|
| 2 HTTP calls per page render (page + schemas) | 1 HTTP call — schemas arrive with the page |
| Schemas fetched in live/preview/SSR even when unused | Schemas only computed server-side in `EDIT_MODE` |
| Separate endpoint that had to be maintained | Endpoint deleted |
| Two-source fallback in the UVE store computed | Single source of truth seeded at page load |

---

## Acceptance Criteria

- [x] `styleEditorSchemas` is included in the GraphQL page response in `EDIT_MODE` with feature flag on
- [x] `styleEditorSchemas` is included in the REST page API response in `EDIT_MODE` with feature flag on
- [x] No `styleEditorSchemas` field is returned in `LIVE`, `PREVIEW_MODE`, or `ADMIN_MODE` via the REST/GQL response (UVE script injection still applies for all non-LIVE modes)
- [x] `/api/v1/page/{pageId}/contenttype-schema` endpoint removed
- [x] No separate network call for schemas in `@dotcms/client`
- [x] UVE store `$styleSchema` uses a single source of truth
- [x] Integration tests cover `EDIT_MODE` (schemas present) and non-edit modes (schemas absent)
- [x] GraphQL Postman tests cover the schema field
- [x] All existing tests pass

## Additional Context
Identified during review of PR #35355 (style editor schema unification). The `registerStyleEditorSchemas` downstream guard already prevents side effects, but the HTTP request itself is wasted. This is a performance optimization, not a correctness fix.

dario-daza:
### Proposed Changes
Removed the ‎`api/v1/page/{pageId}/contenttype-schema` endpoint previously used to retrieve schemas for a page. This endpoint was only used by Headless pages, so we reused its logic to retrieve Style Editor Schemas and now expose them through GraphQL.

### Checklist
- [ ] Remove `api/v1/page/{pageId}/contenttype-schema` endpoint 
- [ ] Refactor Style Editor Schema logic in order to re use it in the GQL and REST Implementation.
- [ ] Expose styleEditorSchemas via GQL
- [ ] Update consumption in the SDK to now retrieve the Schemas from GQL instead of the endpoint.
- [ ] Update integration / postman tests
- [ ] Update FE tests

This PR fixes: #35454

This PR fixes: #35454


nollymar (approved):

core-web/libs/sdk/client/src/lib/client/page/page-api.ts
```diff
@@ -268,12 +262,7 @@ export class PageClient extends BaseApiClient {
                 );
             }
 
-            const styleEditorSchemas = await fetchStyleEditorSchemas(
-                pageResponse.page.identifier,
-                this.config,
-                this.requestOptions,
-                this.httpClient
-            );
+            const styleEditorSchemas = response.data.page?.styleEditorSchemas;
```
KevinDavilaDotCMS: I think we can use pageResponse
Also, do we need to include this in the if(!pageResponse)?

If the page not found, style Editor schemas neither

core-web/libs/portlets/edit-ema/portlet/src/lib/store/features/editor/withEditor.ts
```diff
@@ -176,11 +176,15 @@ export function withEditor() {
...
                 $styleSchema: computed<StyleEditorFormSchema>(() => {
                     const activeContentlet = store.editorActiveContentlet();
-                    const styleSchemas = store.editorStyleSchemas();
-                    const contentSchema = styleSchemas.find(
+                    // Live schemas pushed mid-session by the iframe take precedence;
+                    // fall back to what the page load returned (REST or GQL).
+                    const schemas =
+                        store.editorStyleSchemas().length > 0
+                            ? store.editorStyleSchemas()
+                            : (store.pageAsset()?.page?.styleEditorSchemas ?? []);
```
KevinDavilaDotCMS: Maybe we can set the editorStyleSchemas when we fetch the pageAsset, to always use store.editorStyleSchemas, instead to check the pageAsset here. Is a bit difficult to read

KevinDavilaDotCMS (approved):

PR #35539, opened 2026-05-02
refactor(uve): explicit iframe sizing with DevTools-style canvas controls

### LINKED ISSUE #35324: UVE - Anchor links with internal IDs incorrectly handled as external routes in VTL pages
### Problem Statement

When opening a traditional VTL (Velocity Template Language) page within the UVE environment, anchor tags (<a>) containing href attributes referencing internal page IDs (e.g., #section-id) are being misinterpreted as external routes. Instead of scrolling to the target element, the system attempts to navigate and fetch a new page based on the ID. This results in a '404 Page not found' error, severely impacting user experience when interacting with common UI components like lists or tabs that rely on internal page referencing

https://github.com/user-attachments/assets/5cbe9827-4626-4290-8579-e983017af499

### Steps to Reproduce

1. Navigate to a traditional VTL-based page within the UVE environment.
2. Locate a UI element (such as a tab or list item) that uses an anchor tag with an ID reference (e.g., <a href="#tab-1">).
3. Click on the anchor link.
4. Observe that the system attempts to perform a fetch/navigate action to a non-existent route.
5. Verify that a '404 Page not found' error message is displayed to the user.

### Acceptance Criteria

The navigation logic must distinguish between internal ID fragments and external URL paths. Anchor links starting with '#' should trigger the default browser behavior for internal page scrolling/linking or be handled by the local component state (tabs/lists) without initiating a network fetch or a global route change. The '404 Page not found' error should no longer appear when clicking these links.

### dotCMS Version

main `a2f24da`

### Severity

### Links
https://dotcms.freshdesk.com/a/tickets/36261
-

### LINKED ISSUE #35448: Copy URL functionality to use the selected site instead of the URL
### Problem Statement

Currently when previewing a page the only URLs you get within the copy URL section are specifically based around the URL that the user logged in to. Having an option or a third link to copy that uses the currently selected site would be a good feature to have. 

### Steps to Reproduce

https://github.com/user-attachments/assets/cf280ac1-e9fd-41b4-b8d6-b7eb5f9cc361

### Acceptance Criteria

Have an option that allows us to copy the URL using the site selected instead of what is within the url used to access the backend.

### dotCMS Version

Current Evergreen 26.04.11-01

### Severity

### Links

https://dotcms.freshdesk.com/a/tickets/36416

### LINKED ISSUE #35514: refactor(uve): adopt DevTools-style iframe sizing to eliminate height feedback loop
### Description

The UVE editor sizes `.canvas-inner` and `.canvas-outer` from the iframe's measured document height so scrolling happens at the canvas level rather than inside the iframe.

This design has a fundamental flaw: **the iframe's height is both an input to and an output of layout**. When a customer's page uses viewport-relative units on a layout container — e.g. `min-height: 100dvh` on a top-level wrapper — the iframe enters an infinite growth loop:

1. Page loads, body height ≈ viewport (say 800px).
2. Store updates the canvas height → the `<iframe>` element grows.
3. Iframe's viewport is now taller → `100dvh` re-resolves larger → body height grows.
4. The observer fires → loop continues, monotonically.

We cannot fix this with measurement tricks (CSS overrides, content probes, runaway detection) without breaking legitimate uses of `100vh` / `100dvh` (e.g. hero banners sized to the viewport for design). The contract is broken: there is no stable viewport for the page to resolve viewport units against.

### Proposed direction

Adopt the Chrome DevTools device-toolbar model:

- The iframe has **explicit, user-controllable dimensions**. These are inputs, not derived values.
- Zoom is applied independently of iframe dimensions.
- The **canvas viewport** scrolls when zoomed iframe size exceeds the canvas size.
- Content taller than the iframe **scrolls inside the iframe**.
- Inside the iframe, `vh` / `dvh` / `%` resolve against a fixed iframe height. Stable. No measurement, no loop.

### Acceptance Criteria

- [ ] A page with `min-height: 100dvh` on a layout wrapper renders without runaway growth
- [ ] A page with a hero section using `min-height: 100vh` for design renders correctly
- [ ] Zoom (10%–300%) works with no flicker or loop
- [ ] Canvas scrolls when zoomed content exceeds the canvas viewport
- [ ] Iframe scrolls internally when content exceeds the iframe height
- [ ] Contentlet selection, drag-drop, palette overlays, and dropzone work correctly when the iframe is scrolled
- [ ] Device presets set explicit iframe dimensions
- [ ] User can resize the iframe in responsive mode
- [ ] Drag-drop from palette near the iframe edge auto-scrolls the iframe

fmontes:
## Summary

Closes #35514. Replaces the iframe height-observation pipeline with explicit, user-controllable iframe sizing, and rebuilds the overlay, selection, and bounds-sync machinery on top.

The original bug was an infinite growth loop on pages using viewport-relative units (e.g. `min-height: 100dvh`) — the iframe's height was both an input to and output of layout.

---

## Iframe sizing

- Iframe width and height are now editor state, not derived measurements.
- The user controls them via resize handles, numeric W × H inputs, or device presets.
- Zoom scales what the page inside renders via CSS transform; the iframe's CSS box stays put.
- Iframe size survives zoom, device switches, and side-panel toggling.
- Reset button restores 100% zoom and snaps the iframe back to fill the canvas.

## Bounds and overlay anchoring

- Editor used to *pull* bounds on demand; that races every kind of reflow.
- SDK now *pushes* fresh bounds whenever the iframe layout settles, debounced 100ms.
- One escape hatch (`UVE_FLUSH_BOUNDS`) for drag/drop's synchronous needs.
- New `$iframeLayoutLocked` predicate captures "iframe is mid-flux" in one place.
- Overlays hide while locked; SET_BOUNDS arrival delivers coords and releases the lock together.
- Re-anchor logic moved into a dedicated `withSelectionAnchor` slice.
- Match key is `(inode, container.identifier, container.uuid)` to disambiguate duplicate contentlets.

## Selection and the hover toolbar

- Two overlays: hover follows the pointer and owns all action buttons; selected is a persistent border + label, no tools.
- Visual hint between them: 1px outer line on hover, +1px inset shadow on selected — outer rectangle doesn't grow on click.
- Pencil button restored to its legacy behavior: opens the full-editor modal with the "all pages / this page" prompt.
- Pencil is stateless w.r.t. selection — opening the modal doesn't change what's selected or what the side panel shows.
- Side-panel quick edit moved to its own bolt button next to the pencil.
- Toolbar order: drag · VTL · bolt · style · pencil · delete, all tooltipped.

## Quick-edit form

- Cancel and Save are always enabled — no more confusing disabled states.
- Cancel resets the form (if dirty), clears the active contentlet, and closes the panel.
- Save with an invalid form calls `markAllAsTouched()` to surface validation errors instead of silently no-op'ing.
- "Edit in full editor" button moved from the footer to the form's top-right corner — small, with a pencil icon.
- Footer is now Cancel (left) + Save (right).

## Page navigation

- Clicking a link in the iframe (or switching persona/language) used to unmount the entire editor for the fetch.
- Cause: `pageAssetResponse: null` was being written inside the load reset.
- Fix: keep the previous asset until the new one resolves; chrome stays mounted, content updates in place.
- Selection state still clears on navigation so it doesn't leak into the new page.

## UI consistency

- The pencil's copy-decision modal and the quick-edit's copy-decision view now share one card UI.
- Built with plain `<button>` + Tailwind utilities — no `p-button` / `[dt]` / `::ng-deep` workarounds.
- Same icons, same i18n keys across both surfaces.

## Architecture doc

- New [`libs/portlets/edit-ema/ARCHITECTURE.md`](libs/portlets/edit-ema/ARCHITECTURE.md) covers the contracts above.
- Bounds sync, reflow lifecycle, three contentlet signals (hovered / selected / panel-target), navigation rule, parent/iframe seam.

---

## Test plan

- [x] `min-height: 100dvh` page renders without runaway growth
- [x] Zoom (10–300%) preserves on-screen iframe size
- [x] Resize handles clamp to canvas in responsive mode
- [x] Device presets fit canvas at auto-fit zoom; orientation refit works
- [x] User-set iframe size survives toggling left/right panels
- [x] Hover overlay lets wheel/click through; click selects without firing page links
- [x] Second click on a selected contentlet lets internal links fire
- [x] Selected overlay re-anchors smoothly on scroll, zoom, device switch, sidebar open/close
- [x] Pencil opens full-editor modal with copy-decision on multi-page; doesn't change selection
- [x] Bolt opens side-panel quick-edit; style icon opens style editor
- [x] Quick-edit Cancel resets form, clears active, closes panel
- [x] Quick-edit Save with required fields empty surfaces validation
- [x] Page navigation keeps chrome mounted and clears selection
- [x] Both side panels open on a 14" laptop without horizontal bleed
- [x] Palette section-jump scrolls inside the iframe
- [x] Contentlet tools and dropzone hide during resize and reappear on release

## Tests added

- `withView.spec.ts` (~30 cases): iframe sizing, device-fit math, orientation refit, exit-preset, zoom reset/snap
- `dot-uve-iframe-resize-handles.component.spec.ts`: pointerdown/move/up + destroy-mid-drag
- `dot-uve-zoom-controls.component.spec.ts`: option list, change handler
- `dot-uve-contentlet-tools.component.spec.ts`: hover overlay rendering + content-type label
- `edit-ema-editor.component.spec.ts`: `handleSectionOffset` regression

Some specs reference symbols renamed or removed during the bounds-channel and binary-selector refactors; cleanup deferred.

This PR fixes: #35514


core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-iframe-size-input/dot-uve-iframe-size-input.component.ts
```diff
@@ -0,0 +1,40 @@
...
+    readonly $width = this.store.viewIframeWidth;
+    readonly $height = this.store.viewIframeHeight;
+
+    onWidthChange(event: Event): void {
+        const value = Number((event.target as HTMLInputElement).value);
+        if (!Number.isFinite(value) || value <= 0) {
+            return;
+        }
+        this.store.viewExitDevicePreset();
+        this.store.viewSetIframeSize({ width: value });
```
fmontes: Why two calls here? Shouldn't the store update its own internal props for this use?

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-iframe-size-input/dot-uve-iframe-size-input.component.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    onHeightChange(event: Event): void {
+        const value = Number((event.target as HTMLInputElement).value);
+        if (!Number.isFinite(value) || value <= 0) {
+            return;
+        }
+        this.store.viewExitDevicePreset();
+        this.store.viewSetIframeSize({ height: value });
```
fmontes: Why two calls here? Shouldn't the store update its own internal props for this use?

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-contentlet-tools/dot-uve-contentlet-tools.component.spec.ts
```diff
@@ -79,12 +88,25 @@ describe('DotUveContentletToolsComponent', () => {
...
-            })
+            }),
+            {
+                provide: UVEStore,
+                useFactory: () => ({
+                    editorSelectedContentletArea: selectedContentletArea,
+                    $iframeLayoutLocked: () => false
+                })
+            }
         ],
```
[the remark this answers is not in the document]
fmontes: Already migrated in commit 9c3b26e962 — the spec's UVEStore mock now exposes `editorSelected` and `setSelected`, and the test ids were renamed to the `hover-*` form to match the template. Resolving.

core-web/libs/portlets/edit-ema/portlet/src/lib/services/iframe-messenger/uve-iframe-messenger.service.ts
```diff
@@ -65,11 +65,14 @@ export class UveIframeMessengerService {
...
+     * snapshot before the user moves another pixel (drag/drop dropzone);
+     * everything else should rely on the auto-bounds push channel.
      */
-    requestBounds(): void {
+    flushBounds(): void {
         this.sendPostMessage({
-            name: __DOTCMS_UVE_EVENT__.UVE_REQUEST_BOUNDS
+            name: __DOTCMS_UVE_EVENT__.UVE_FLUSH_BOUNDS
         });
     }
```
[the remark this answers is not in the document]
fmontes: The `requestBounds()` → `flushBounds()` rename is fully propagated. The messenger service spec was updated in commit 47bcd8b071 (now asserts `UVE_FLUSH_BOUNDS`). `grep -rn requestBounds core-web` returns nothing — no external callers of the old name remain. Keeping the rename without a backwards-compat alias since the API is internal to the editor + SDK pair.

core-web/libs/portlets/edit-ema/portlet/src/lib/store/models.ts
```diff
@@ -98,8 +98,22 @@ export interface UVEState {
...
+     * Set by the SDK's CONTENTLET_CLICKED event (via SET_SELECTED_CONTENTLET)
+     * and by the hover toolbar's bolt / palette buttons (via
+     * `promoteHoverToSelected`). Re-anchored on every iframe reflow by
+     * `withSelectionAnchor.applyBoundsForSelection`. Hides — but doesn't
+     * clear — during `$iframeLayoutLocked` phases.
+     *
+     * Distinct from `editorContentArea` (hovered) — selection persists
+     * while the user hovers other contentlets.
+     */
+    editorSelected: SelectedContentlet | null;
```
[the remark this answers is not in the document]
fmontes: All four specs were already migrated to `editorSelected`: `withEditor.spec.ts` (commit fdf0a33d88), `dot-uve.store.integration.spec.ts` (004d2aea3c), `withPageApi.spec.ts` (d013042ece), and `dot-uve-style-editor-form.component.spec.ts` (2dcac43950). Full edit-ema test suite is green (1351 passing). Resolving.

core-web/libs/ui/src/lib/components/dot-binary-option-selector/dot-binary-option-selector.component.html
```diff
@@ -1,40 +1,53 @@
...
-        <label data-testId="option1" for="option1">
-            <input
-                [(ngModel)]="value"
-                [value]="firstOption.value"
-                id="option1"
-                type="radio"
-                name="options" />
-            <div class="item">
+<section class="flex flex-col p-6" aria-labelledby="binary-option-selector-title">
+    <div class="flex flex-col gap-3">
```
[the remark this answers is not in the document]
fmontes: Out of scope for this PR — `dot-binary-option-selector` is a pre-existing component this branch doesn't touch. Tracking separately if it needs the `aria-labelledby` target added.

dotCMS/src/main/webapp/ext/uve/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "exports": {
+    "./package.json": "./package.json",
+    ".": "./src/index.ts",
+    "./internal": "./src/internal.ts"
+  },
+  "typesVersions": {
+    "*": {
+      ".": ["./src/index.d.ts"],
+      "internal": ["./src/internal.d.ts"]
+    }
```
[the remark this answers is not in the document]
fmontes: Removed from this PR in commit acfe79dade — the file is unused (the runtime loads `/ext/uve/dot-uve.js` directly via `<script>` tag, nothing reads the package.json) and its `exports`/`typesVersions` pointed at non-existent `./src/*` paths. Deleting rather than fixing it since there's no consumer that needs it.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-contentlet-tools/dot-uve-contentlet-tools.component.html
```diff
@@ -1,80 +1,144 @@
...
                         icon="pi pi-code" />
                     <p-menu
                         [model]="vtlMenuItems()"
                         [popup]="true"
-                        #menuVTL
+                        #menuHoverVTL
                         appendTo="body"
-                        data-testId="menu-vtl" />
+                        data-testId="hover-menu-vtl" />
                 }
```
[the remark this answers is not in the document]
fmontes: Fixed in commit ac5ea7c29f — `viewChild` queries renamed to `menuHover` / `menuHoverVTL` / `menuHoverActions`, and `hideMenus()` now closes all three (the collapsed-actions tiered menu was missing too). Hover transitions close menus again.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-contentlet-quick-edit/dot-uve-quick-edit-form/dot-uve-quick-edit-form.component.html
```diff
@@ -0,0 +1,160 @@
...
+                            <p-select
+                                [formControlName]="field.variable"
+                                [inputId]="field.variable"
+                                [options]="field.options"
+                                optionLabel="label"
+                                optionValue="value"
+                                [showClear]="true"
+                                [disabled]="field.readOnly"
+                                [placeholder]="'Select ' + field.name"></p-select>
+                        }
```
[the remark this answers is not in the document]
fmontes: Fixed in commit cc783adb34 — placeholder now uses the existing `Select` dm key: `('Select' | dm) + ' ' + field.name`. Same change applied to the MULTI_SELECT branch.

core-web/libs/sdk/uve/src/internal/events.ts
```diff
@@ -62,34 +62,133 @@ export function onPageReload(callback: UVEEventHandler) {
...
+            if (m.type !== 'childList') continue;
+            if (containsContainerNode(m.addedNodes) || containsContainerNode(m.removedNodes)) {
+                observeAll();
+                scheduleEmit();
+                return;
+            }
+        }
+    });
+    mutationObserver.observe(document.body, { childList: true, subtree: true });
+
```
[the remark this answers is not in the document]
fmontes: Fixed in commit cf6c5b7ff5 — `mutationObserver.observe(document.body ?? document.documentElement, ...)`. The fallback covers SDK loading from `<head>` before `<body>` exists; subtree:true on `<html>` still picks up containers when `<body>` mounts.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-contentlet-quick-edit/dot-uve-quick-edit-form/dot-uve-quick-edit-form.component.html
```diff
@@ -0,0 +1,160 @@
...
+                            <p-multiSelect
+                                [formControlName]="field.variable"
+                                [inputId]="field.variable"
+                                [options]="field.options"
+                                optionLabel="label"
+                                optionValue="value"
+                                [showClear]="true"
+                                [disabled]="field.readOnly"
+                                [placeholder]="'Select ' + field.name"></p-multiSelect>
+                        }
```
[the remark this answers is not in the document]
fmontes: Fixed in commit cc783adb34 (same as the SELECT case) — placeholder is now `('Select' | dm) + ' ' + field.name` so both fields stay localized.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-iframe-resize-handles/dot-uve-iframe-resize-handles.component.ts
```diff
@@ -0,0 +1,113 @@
...
+    #dragAbort: AbortController | null = null;
+
+    constructor() {
+        this.destroyRef.onDestroy(() => this.#endDrag(null));
+    }
+
+    onPointerDown(event: PointerEvent, axis: ResizeAxis): void {
+        event.preventDefault();
+        event.stopPropagation();
+
```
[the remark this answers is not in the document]
fmontes: Fixed in commit b4edfd7447 — added `#captureTarget` + `#capturePointerId` fields. `#endDrag` falls back to those when called with `null` (the destroy path), so pointer capture is released even when the component goes away mid-drag.

zJaaal:
We need to check if the default values are working on all supported fields on the quick edit form

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-contentlet-quick-edit/dot-uve-quick-edit-form/dot-uve-quick-edit-form.component.html
zJaaal, line 1: We need to make sure default values are being populated

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-iframe-resize-handles/dot-uve-iframe-resize-handles.component.html
zJaaal, line 1: Should we use p-buttons?

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-iframe-resize-handles/dot-uve-iframe-resize-handles.component.ts
```diff
@@ -0,0 +1,129 @@
...
+    #dragAbort: AbortController | null = null;
+
+    /**
+     * Capture target + pointerId, retained so the destroy path can release
+     * pointer capture even when no explicit target is passed in (some
+     * browsers leave capture stuck if the element goes away mid-drag).
+     */
+    #captureTarget: HTMLElement | null = null;
+    #capturePointerId: number | null = null;
+
```
zJaaal, line 33: Should this be signals?

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-iframe-size-input/dot-uve-iframe-size-input.component.html
zJaaal, line 1: Should we use primeNG Inputs?

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-palette/dot-uve-palette.component.html
```diff
@@ -47,9 +47,27 @@
         </p-tabpanel>
         <p-tabpanel [value]="TABS_MAP.LAYERS">
             @if ($activeTab() === TABS_MAP.LAYERS) {
-                <div class="w-full h-full overflow-y-auto">
-                    <dot-row-reorder (onRowSelect)="onNodeSelect.emit($event)" />
-                </div>
+                @if ($isStandardTemplate()) {
```
zJaaal, line 50: We should also disable the navigation to template builder on the nav bar

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/edit-ema-editor.component.scss
```diff
@@ -82,12 +100,12 @@ dot-results-seo-tool {
...
     &.open {
-        width: clamp(400px, 20vw, 500px);
+        width: clamp(360px, 20vw, 500px);
     }
 
     > * {
-        width: clamp(400px, 20vw, 500px);
-        min-width: clamp(400px, 20vw, 500px);
+        width: clamp(360px, 20vw, 500px);
+        min-width: clamp(360px, 20vw, 500px);
```
zJaaal, line 108: Should we use rem to let the UI adapt when the user changes the font size?

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/edit-ema-editor.component.ts
```diff
@@ -445,15 +434,71 @@ export class EditEmaEditorComponent implements OnInit, OnDestroy, AfterViewInit
...
 
+    /**
+     * Notify the iframe whenever the editor clears its contentlet selection
+     * (canvas resize, scroll, device switch, etc.). The SDK keeps a
+     * `lastSelectedInode` tracker that gates click→passthrough behavior; if
+     * the editor drops the selection without telling the SDK, a follow-up
+     * click on the same contentlet would be silently treated as a passthrough
+     * (page click) and the toolbar wouldn't reappear.
+     */
+    #lastSelectedAreaWasSet = false;
```
zJaaal, line 445: should this be a signal?

core-web/libs/portlets/edit-ema/portlet/src/lib/services/iframe-messenger/uve-iframe-messenger.service.ts
```diff
@@ -65,11 +65,14 @@ export class UveIframeMessengerService {
...
+     * Tell the SDK to emit page bounds NOW, bypassing the auto-bounds
+     * debounce. Use only for cases where the editor needs a synchronous
+     * snapshot before the user moves another pixel (drag/drop dropzone);
+     * everything else should rely on the auto-bounds push channel.
      */
-    requestBounds(): void {
+    flushBounds(): void {
         this.sendPostMessage({
-            name: __DOTCMS_UVE_EVENT__.UVE_REQUEST_BOUNDS
+            name: __DOTCMS_UVE_EVENT__.UVE_FLUSH_BOUNDS
```
zJaaal, line 75: beware of this changes, it can create breaking changes between versions

core-web/libs/portlets/edit-ema/portlet/src/lib/store/features/editor/toolbar/withView.ts
```diff
@@ -217,50 +308,165 @@ export function withView() {
...
+ */
+function computeDeviceFit(
+    deviceDims: { width: number; height: number },
+    canvasW: number,
+    canvasH: number
+): { width: number; height: number; zoom: number } {
+    if (canvasW <= 0 || canvasH <= 0 || deviceDims.width <= 0 || deviceDims.height <= 0) {
+        return { width: deviceDims.width, height: deviceDims.height, zoom: 100 };
+    }
+    const rawFit = Math.min(1, canvasW / deviceDims.width, canvasH / deviceDims.height);
```
zJaaal, line 463: Beware of divisions by zero

core-web/libs/portlets/edit-ema/portlet/src/lib/utils/index.ts
```diff
@@ -1004,21 +988,25 @@ export const getOrientation = (device: DotDevice): Orientation => {
...
+ */
+export const measureCanvasAvailableSize = (
+    el: HTMLElement
+): { width: number; height: number } | null => {
+    const styles = getComputedStyle(el);
+    const padX = parseFloat(styles.paddingLeft) + parseFloat(styles.paddingRight);
+    const padY = parseFloat(styles.paddingTop) + parseFloat(styles.paddingBottom);
+
+    const width = el.clientWidth - padX;
+    const height = el.clientHeight - padY;
```
zJaaal, line 1007: Check for negative values

core-web/libs/sdk/types/src/lib/events/internal.ts
```diff
@@ -10,9 +10,12 @@ export enum __DOTCMS_UVE_EVENT__ {
...
     UVE_RELOAD_PAGE = 'uve-reload-page',
     /**
-     * Request the bounds for the elements
+     * Tell the SDK to emit page bounds immediately, bypassing the
+     * auto-bounds debounce. Used when the editor needs a synchronous
+     * snapshot of bounds for an interaction (drag/drop dropzone) and
+     * cannot wait the ~100ms trailing edge.
      */
-    UVE_REQUEST_BOUNDS = 'uve-request-bounds',
+    UVE_FLUSH_BOUNDS = 'uve-flush-bounds',
```
zJaaal, line 18: same comment as above, beware of compatibility here

core-web/libs/sdk/uve/src/internal/events.ts
```diff
@@ -232,3 +361,106 @@ export function onContentletHovered(callback: UVEEventHandler) {
...
+            title: foundElement.dataset?.['dotTitle'],
+            inode: foundElement.dataset?.['dotInode'],
+            contentType: foundElement.dataset?.['dotType'],
+            baseType: foundElement.dataset?.['dotBasetype'],
+            widgetTitle: foundElement.dataset?.['dotWidgetTitle'],
+            onNumberOfPages: foundElement.dataset?.['dotOnNumberOfPages'],
+            ...(foundElement.dataset?.['dotStyleProperties'] && {
+                dotStyleProperties: JSON.parse(foundElement.dataset['dotStyleProperties'])
+            })
+        };
```
zJaaal: i think we have a function that does this

core-web/libs/ui/src/lib/components/dot-binary-option-selector/dot-binary-option-selector.component.html
```diff
@@ -1,40 +1,53 @@
...
-        pButton
-        pAutoFocus
-        data-testId="button"></button>
-</div>
+    <div class="flex justify-end mt-5">
+        <button
+            (click)="onSelect()"
+            [autofocus]="true"
+            [label]="btnLabel | dm"
+            pButton
```
zJaaal: I think we should prefer p-button instead of the directive

dotCMS/src/main/webapp/ext/uve/package.json
zJaaal, line 1: Not sure why we do the package.json, but we should avoid it

zJaaal (changes requested):

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-contentlet-quick-edit/dot-uve-quick-edit-form/dot-uve-quick-edit-form.component.html
fmontes: done

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-iframe-size-input/dot-uve-iframe-size-input.component.html
fmontes: No because I need them to be completely naked.

dotCMS/src/main/webapp/ext/uve/package.json
fmontes: deleted

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-iframe-size-input/dot-uve-iframe-size-input.component.ts
fmontes: Stale — the store already encapsulates the pair via `viewResizeIframe` (exits device preset, then `viewSetIframeSize`), which is the single call the component now makes. No double-call here. Resolving.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-iframe-size-input/dot-uve-iframe-size-input.component.ts
fmontes: Same as above — only `viewResizeIframe` is called now, and the store handles preset-exit + size-set internally. Resolving.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-iframe-resize-handles/dot-uve-iframe-resize-handles.component.html
fmontes: Keeping native `<button>` here intentionally — these aren't conventional buttons, they're DevTools-style drag bars (2px right edge, 2px bottom edge, 16px corner) that drive a `pointerdown` handler with custom cursors (`cursor-ew-resize`, `cursor-ns-resize`, `cursor-nwse-resize`). p-button wraps content in a themed shell with ripple/label/icon slots that would all need to be neutralized to keep the bare-bar look. Accessibility is covered with native semantics + `aria-label` per handle. Resolving.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-iframe-resize-handles/dot-uve-iframe-resize-handles.component.ts
fmontes: Keeping these as plain private fields — they're imperative bookkeeping for the in-flight pointer drag (capture target + pointerId), used only by `onPointerDown` (set) and `#endDrag` (read on cleanup). Nothing reactive consumes them: no template binding, no `computed`, no `effect`. Wrapping them in signals would add tracking overhead with zero readers. Resolving.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-palette/dot-uve-palette.component.html
fmontes: Already in place — `dot-ema-shell.component.ts` disables the Layout (template builder) nav-bar item for advanced templates and shows the `editema.editor.navbar.layout.tooltip.cannot.edit.advanced.template` tooltip explaining why. The palette empty-state added here is the in-tab counterpart of that nav-level guard. Resolving.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/edit-ema-editor.component.scss
fmontes: Keeping px for the panel rail width. The same `clamp(px, vw, px)` pattern is used for the palette (`clamp(280px, 15vw, 320px)`) and the edit panel (`clamp(360px, 20vw, 500px)`) — these are UI chrome rails where the px floor encodes a content-readability minimum at default font size, not a font-size-relative limit. Switching to rem would let the rail balloon to 700–1000px at 200% browser zoom, which fights the side-rail intent. Body text inside the panels still scales because that uses rem/em from the typography layer. Resolving.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/edit-ema-editor.component.ts
fmontes: Keeping as a plain field. `#lastSelectedAreaWasSet` is the previous-value tracker for the effect on line 446 — it's only read inside `untracked()` and immediately rewritten on the next run. The whole point is for it NOT to participate in the reactive graph; making it a signal would create a tracking node that nothing tracks. Plain mutable field is the right primitive for "compare to last run" inside an effect. Resolving.

core-web/libs/portlets/edit-ema/portlet/src/lib/services/iframe-messenger/uve-iframe-messenger.service.ts
fmontes: Good catch — added an SDK back-compat shim. The editor now dual-emits both `UVE_FLUSH_BOUNDS` (new) and `UVE_REQUEST_BOUNDS` (legacy, marked `@deprecated` in the enum) for one release. SDKs in the wild bundled with older `@dotcms/uve` keep receiving flushes; new SDKs only need to handle the new name. Drop the alias after the next minor release. Test added in `uve-iframe-messenger.service.spec.ts` to lock the dual-emit.

core-web/libs/sdk/types/src/lib/events/internal.ts
fmontes: Addressed in the same change as the messenger comment — `UVE_REQUEST_BOUNDS` is back in the enum, marked `@deprecated`, and the editor dual-emits both names for one release so old in-the-wild SDKs keep receiving flushes. Will drop after the next minor release of `@dotcms/uve`.

core-web/libs/portlets/edit-ema/portlet/src/lib/store/features/editor/toolbar/withView.ts
fmontes: Already guarded. Line 460 returns early when any of `canvasW / canvasH / deviceDims.width / deviceDims.height` is `<= 0` — falling back to native device dimensions at 100% zoom. By the time line 463 runs the divisions, all four are positive. Resolving.

core-web/libs/portlets/edit-ema/portlet/src/lib/utils/index.ts
fmontes: Already guarded — line 1009: `width > 0 && height > 0 ? { width, height } : null`. Negative values (and zero) hit the `null` branch, and callers early-return on `null` before pushing a degenerate size to the store. The doc comment above the function explicitly calls this out. Resolving.

core-web/libs/sdk/uve/src/internal/events.ts
fmontes: Good catch — the dataset-to-contentlet block was duplicated at `events.ts` lines 312–323 (auto-bounds path) and 413–424 (CONTENTLET_CLICKED path). Extracted to `readContentletDataset(element)` in `dom.utils.ts` and used at both call sites. SDK test suite green. Resolving.

core-web/libs/ui/src/lib/components/dot-binary-option-selector/dot-binary-option-selector.component.html
fmontes: Out of scope for this PR — `dot-binary-option-selector` is a pre-existing component that this branch doesn't touch (`git log origin/main..HEAD` shows no commits to that path). Worth tracking separately as a UI-library refactor (along with the related a11y `aria-labelledby` comment Copilot raised on the same file). Resolving here.

fmontes: Correction to my earlier reply on this thread — I was wrong, this PR DOES touch `dot-binary-option-selector` (I had been looking at a stale local branch). Switched the action button at the bottom from `pButton` directive on a native `<button>` to `<p-button>` element. The two option cards above remain native `<button>`s on purpose — they're large clickable cards with custom card styling, not standard buttons. UI test suite green.

core-web/libs/ui/src/lib/components/dot-binary-option-selector/dot-binary-option-selector.component.html
```diff
@@ -1,40 +1,53 @@
...
-    <div class="options">
-        <label data-testId="option1" for="option1">
-            <input
-                [(ngModel)]="value"
-                [value]="firstOption.value"
-                id="option1"
-                type="radio"
-                name="options" />
-            <div class="item">
+<section class="flex flex-col p-6" aria-labelledby="binary-option-selector-title">
```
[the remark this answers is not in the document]
fmontes: Fixed — removed the broken `aria-labelledby` reference. The component is rendered inside a `DynamicDialog` which already has its own labeled title bar, so a redundant section-level accessible name isn't needed. (Updating my earlier reply on this thread — I had incorrectly marked it out of scope; the file is in fact modified by this PR.)

core-web/libs/sdk/types/src/lib/editor/public.ts
```diff
@@ -208,11 +215,39 @@ export enum UVEEventType {
      */
     CONTENTLET_HOVERED = 'contentlet-hovered',
 
+    /**
+     * Triggered when a contentlet is clicked (pointerdown on its element).
+     * Used by the editor to promote the clicked contentlet to "selected"
+     * without the editor having to capture pointer events on its hover overlay.
+     */
+    CONTENTLET_CLICKED = 'contentlet-clicked',
```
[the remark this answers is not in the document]
fmontes: Fixed — doc now reads "capture-phase `click` event on its element" to match the actual listener registration in `sdk/uve/src/internal/events.ts:435` (`document.addEventListener('click', clickCallback, { capture: true })`).

zJaaal (approved):

rjvelazco (approved):

PR #35556, opened 2026-05-04
Modernization: feat(es-search): migrate Elasticsearch Search portlet to Angular

### LINKED ISSUE #34733: Dojo to Angular: ES Search Portlet
Migrate this portlet from Dojo to Angular (PrimeNG). Part of the UX Redesign pillar.

hmoreras:
## Summary

Migrates the legacy Dojo/JSP Elasticsearch Search portlet (`ESContentResourcePortlet`) to a fully modern Angular 19+ portlet built with signals, NgRx SignalStore, Monaco editor, and PrimeNG Tailwind theming. The legacy portlet is preserved as a JSP fallback during the transition.

**GitHub Issue:** #34733

https://github.com/user-attachments/assets/cefbb0be-920e-410d-85fc-78802d027f18

### What changed

**New Angular portlet (`libs/portlets/dot-es-search`)**
- Full NgRx SignalStore-based state management (`dot-es-search.store.ts`)
- Monaco editor for Elasticsearch JSON query authoring with real-time JSON validation
- Results table with client-side column sorting (title, type, identifier, modDate)
- Four response tabs: Results · Aggregations · Suggestions · Raw JSON
- Collapsible parameters panel: `live` toggle and `User ID / Email` filter (admin-only impersonation matching legacy JSP behavior)
- Export menu with **Copy as cURL** and **Copy as fetch (JS)** developer actions
- Help popover with 7 curated query examples (match-all, full-text, bool filters, aggregations, suggestions)
- 1000-row result cap enforced both client-side and server-side; warning banners shown when results are truncated
- Shareable link via URL query params

**Backend**
- Added `@QueryParam("userid")` to `ESContentResourcePortlet.search()` with admin-only `resolveUserForSearch()` helper (accepts user ID or email)
- Added `PortletID.ES_SEARCH` constant
- Registered the new portlet in `portlet.xml` and `app.routes.ts`
- Added JSP-based legacy fallback at `WEB-INF/jsp/es-search-legacy/render.jsp` (full feature parity with the original portlet)
- Extended `Language.properties` with all new i18n keys

**DotAlertConfirmService modernization**
- Converted `alertModel` / `confirmModel` from mutable properties to Angular signals so the component participates in signal-driven change detection — fixes the "error dialog not showing on first failure" issue

**Styling**
- Removed all `::ng-deep` SCSS overrides; replaced with PrimeNG `pt` (passthrough) API and Tailwind utility classes
- Deleted SCSS file entirely — zero custom styles

### Tests
- Unit tests for the SignalStore (`dot-es-search.store.spec.ts`)
- Component tests for the page component (`dot-es-search-page.component.spec.ts`)
- Service tests (`dot-es-search.service.spec.ts`)
- Updated `DotAlertConfirmService` specs to cover signal-based model
- Updated `ESContentResourcePortletTest` call sites for the new `userid` param

## Test plan

- [ ] Open the ES Search portlet from the dotCMS nav — editor loads empty
- [ ] Type a valid JSON query and click **Run** — results appear in the Results tab
- [ ] Type invalid JSON — Run button disables and a red error banner appears above the editor
- [ ] Run a query that returns 0 results — empty state message displayed, Share/Export buttons hidden
- [ ] Run a query returning >1000 rows — warning banner shown, table capped at 1000 rows
- [ ] Click **Export → Copy as cURL** and **Copy as fetch** — snippets copied to clipboard
- [ ] Toggle **Only show published results** and **User ID or Email** params — query updates accordingly
- [ ] Verify error dialog appears immediately on a bad query (e.g. wrong index name)
- [ ] Verify **Wrap code** toggle enables/disables Monaco word-wrap
- [ ] Open Help popover — all 7 examples render and are clickable to load into editor
- [ ] Verify column sorting works (click Title, Type, Identifier, modDate headers)
- [ ] Check Aggregations, Suggestions, and Raw tabs render correct sub-data
- [ ] Run existing unit tests: `nx test dot-es-search`

🤖 Generated with [Claude Code](https://claude.com/claude-code) — Modernization Team

This PR fixes: #34733


core-web/libs/data-access/src/lib/dot-es-search/dot-es-search.service.ts
```diff
@@ -0,0 +1,34 @@
...
+     */
+    search(query: string, params: ESSearchParams): Observable<ESSearchResponse> {
+        let body: unknown;
+        try {
+            body = JSON.parse(query);
+        } catch {
+            return throwError(() => new SyntaxError('Invalid JSON query'));
+        }
+
+        let httpParams = new HttpParams().set('depth', 1).set('live', params.live ?? true);
```
zJaaal, line 24: Should we let the user pass the depth?

core-web/libs/dotcms-models/src/lib/dot-es-search.model.ts
```diff
@@ -0,0 +1,39 @@
...
+    hits: {
+        total: number | { value: number; relation: string };
+        hits: ESHit[];
+    };
+    took: number;
+    timed_out: boolean;
+    aggregations?: Record<string, unknown>;
+    suggest?: Record<string, unknown>;
+    _shards: { total: number; successful: number; skipped: number; failed: number };
+}
```
zJaaal, line 53: lets type the nested objects too

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/store/dot-es-search.store.ts
```diff
@@ -0,0 +1,200 @@
...
+    DotMessageService
+} from '@dotcms/data-access';
+import { ComponentStatus, ESSearchParams, ESSearchResponse } from '@dotcms/dotcms-models';
+import { PrincipalConfiguration } from '@dotcms/ui';
+
+export const MAX_HITS = 1000;
+
+export type EsSearchActiveTab = 'results' | 'raw' | 'aggregations' | 'suggestions';
+
+export interface EsSearchState {
```
zJaaal: Lets preserve the `ES` to maintain consistency, check if we have this issue somewhere else

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/store/dot-es-search.store.ts
```diff
@@ -0,0 +1,200 @@
...
+            const totalCount = typeof total === 'object' ? total.value : total;
+            return totalCount > MAX_HITS;
+        }),
+        rawJson: computed(() => {
+            const response = store.response();
+            if (!response) return '';
+            return JSON.stringify(response, null, 2);
+        }),
+        isLoading: computed(() => store.status() === ComponentStatus.LOADING),
+        hasResults: computed(
```
zJaaal: this is checking more than just results, maybe we should change the name to something else

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/store/dot-es-search.store.ts
```diff
@@ -0,0 +1,200 @@
...
+                            .pipe(
+                                tapResponse({
+                                    next: (response) => {
+                                        const hits = response.esresponse[0]?.hits?.hits ?? [];
+                                        const aggs = response.esresponse[0]?.aggregations;
+                                        const hasAggs = !!aggs && Object.keys(aggs).length > 0;
+                                        const activeTab: EsSearchActiveTab =
+                                            hasAggs && hits.length === 0
+                                                ? 'aggregations'
+                                                : 'results';
```
zJaaal, line 167: Should we make const of this values? Like an enum or as const?

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/dot-es-search-page.component.html
```diff
@@ -0,0 +1,544 @@
...
+                (onClick)="downloadRawJson()"
+                data-testid="es-search-export-btn" />
+        }
+    </div>
+</p-toolbar>
+
+<!-- Main split pane -->
+<p-splitter [panelSizes]="[33, 67]" class="flex-1 min-h-0" [pt]="splitterPt">
+    <!-- Left: Query Editor + Parameters -->
+    <ng-template pTemplate>
```
zJaaal, line 32: do we need to add a name to the pTemplate?

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/dot-es-search-page.component.html
```diff
@@ -0,0 +1,544 @@
...
+                    class="flex items-center gap-1.5 px-4 py-4 bg-surface-50 border-b border-surface-200 text-sm text-color-secondary flex-shrink-0"
+                    data-testid="es-search-stats-bar">
+                    <i class="pi pi-circle-fill text-green-600" style="font-size: 0.5rem"></i>
+                    {{ 'esSearch.results.ready' | dm }}
+                    <span class="text-color-secondary mx-1">&middot;</span>
+                    <strong class="text-color">{{ store.hitCount() }}</strong>
+                    &nbsp;{{ 'esSearch.results.count' | dm: [store.hitCount().toString()] }}
+                    <span class="text-color-secondary mx-1">&middot;</span>
+                    <strong class="text-color">{{ store.queryTimeMs() }}</strong>
+                    &nbsp;{{
```
zJaaal: why &nbsp;?

zJaaal (changes requested):

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/dot-es-search-page.component.ts
```diff
@@ -0,0 +1,394 @@
...
+
+const RAW_EDITOR_OPTIONS = {
+    ...QUERY_EDITOR_OPTIONS,
+    readOnly: true,
+    lineNumbers: 'off'
+};
+
+@Component({
+    selector: 'dot-es-search-page',
+    standalone: true,
```
oidacra: Per `ANGULAR_STANDARDS.md`, `standalone: true` is implied by default in Angular 17+ and shouldn't be set explicitly inside `@Component`. Please remove this property from the decorator.

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-shell/dot-es-search-shell.component.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import { map } from 'rxjs/operators';
+
+import { DotEmptyContainerComponent, PrincipalConfiguration } from '@dotcms/ui';
+
+import { DotEsSearchPageComponent } from '../dot-es-search-page/dot-es-search-page.component';
+
+@Component({
+    selector: 'dot-es-search-shell',
+    standalone: true,
```
oidacra: Same as on the page component — `standalone: true` is the implicit default in Angular 17+ and shouldn't be set explicitly. Please remove it here as well.

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/dot-es-search-page.component.spec.ts
```diff
@@ -0,0 +1,346 @@
...
+        spectator.fixture.componentRef.changeDetectorRef.markForCheck();
+        spectator.detectChanges();
+        const btn = spectator.query(byTestId('es-search-run-btn'))?.querySelector('button');
+        if (btn) spectator.click(btn);
+        expect(store.runSearch).toHaveBeenCalled();
+    });
+
+    it('should disable the Run button when query is empty', () => {
+        const { hasEditorErrors, store } = spectator.component;
+        expect(hasEditorErrors() || !store.query()).toBe(true);
```
oidacra: This assertion is tautological. The store mock always returns `''` for `query()`, so `!store.query()` is permanently `true` and `||`-shorts the entire expression — the test passes regardless of what the button does. Please query the actual DOM, e.g.:

```ts
const btn = spectator.query(byTestId('es-search-run-btn'))?.querySelector('button');
expect(btn?.disabled).toBe(true);
```

mirroring the sibling describe block.

core-web/libs/data-access/src/lib/dot-es-search/dot-es-search.service.spec.ts
```diff
@@ -0,0 +1,65 @@
+import { HttpClientTestingModule, HttpTestingController } from '@angular/common/http/testing';
```
oidacra: This new spec bypasses Spectator entirely. The project standard (`TESTING_REVIEW_RULES.md`) requires Spectator (`@ngneat/spectator/jest`) for all tests; HTTP services should use `createHttpFactory`:

```ts
import { createHttpFactory, HttpMethod, SpectatorHttp } from '@ngneat/spectator/jest';

describe('DotEsSearchService', () => {
    let spectator: SpectatorHttp<DotEsSearchService>;
    const createHttp = createHttpFactory(DotEsSearchService);
    beforeEach(() => { spectator = createHttp(); });
    afterEach(() => { spectator.controller.verify(); });

    it('...', () => {
        spectator.service.search(MOCK, {}).subscribe(...);
        const req = spectator.expectOne('/api/es/search', HttpMethod.POST);
        req.flush(MOCK_RESPONSE);
    });
});
```

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/dot-es-search-page.component.html
```diff
@@ -0,0 +1,544 @@
...
+                    @for (_ of [1, 2, 3, 4, 5]; track $index) {
+                        <p-skeleton height="2.5rem" />
+                    }
+                </div>
+            } @else {
+                <!-- Stats bar -->
+                <div
+                    class="flex items-center gap-1.5 px-4 py-4 bg-surface-50 border-b border-surface-200 text-sm text-color-secondary flex-shrink-0"
+                    data-testid="es-search-stats-bar">
+                    <i class="pi pi-circle-fill text-green-600" style="font-size: 0.5rem"></i>
```
oidacra: Per `STYLING_STANDARDS.md`, hardcoded inline styles aren't allowed. Use a Tailwind utility (`text-[0.5rem]` for the exact size, or a closer standard like `text-xs`) instead of `style="font-size: 0.5rem"`.

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/dot-es-search-page.component.ts
```diff
@@ -0,0 +1,394 @@
...
+    readonly MAX_HITS = MAX_HITS;
+
+    readonly ComponentStatus = ComponentStatus;
+
+    readonly splitterPt = { root: { class: 'border-0! rounded-none!' } };
+    readonly tabPanelsPt = { root: { class: 'flex-1 min-h-0 overflow-auto p-0!' } };
+    readonly tabPanelPt = { root: { class: 'h-full p-0!' } };
+
+    readonly paramsOpen = signal(true);
+    readonly hasEditorErrors = signal(false);
```
oidacra: Per `ANGULAR_STANDARDS.md`, all signal-bearing values must use the `$` prefix. The signals declared in this component (`paramsOpen`, `hasEditorErrors`, `queryEditorOptions`, `parsedAggregations`, `parsedSuggestions`) should be renamed to `$paramsOpen`, `$hasEditorErrors`, etc., with template and test references updated accordingly. Plain non-signal helpers (`exportItems`, `helpExamples`, `noHitsConfig`, the `pt` objects) don't need the prefix.

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/dot-es-search-page.component.ts
```diff
@@ -0,0 +1,394 @@
...
+    ],
+    providers: [DotEsSearchStore, DotEsSearchService, DotCurrentUserService],
+    templateUrl: './dot-es-search-page.component.html',
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    host: { class: 'flex flex-col h-full min-h-0 bg-white' }
+})
+export class DotEsSearchPageComponent {
+    readonly store = inject(DotEsSearchStore);
+    private readonly messageService = inject(DotMessageService);
+    private readonly document = inject(DOCUMENT);
```
oidacra: Per `TYPESCRIPT_STANDARDS.md`, use ECMAScript private fields (`#`) instead of the `private` keyword for instance properties:

```ts
readonly #messageService = inject(DotMessageService);
readonly #document = inject(DOCUMENT);
```

and update usages accordingly (`this.#messageService`, `this.#document`).

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-shell/dot-es-search-shell.component.ts
```diff
@@ -0,0 +1,38 @@
...
+            <dot-es-search-page />
+        } @else {
+            <dot-empty-container [configuration]="unlicensedConfig" [hideContactUsLink]="true" />
+        }
+    `,
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    host: { class: 'flex flex-col h-full min-h-0 block' }
+})
+export class DotEsSearchShellComponent {
+    private readonly route = inject(ActivatedRoute);
```
oidacra: Same as on the page component — use ECMAScript private:

```ts
readonly #route = inject(ActivatedRoute);
```

and update usages to `this.#route`.

core-web/libs/data-access/src/lib/dot-es-search/dot-es-search.service.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+        }
+    ]
+};
+
+describe('DotEsSearchService', () => {
+    let service: DotEsSearchService;
+    let httpController: HttpTestingController;
+
+    beforeEach(() => {
+        TestBed.configureTestingModule({ imports: [HttpClientTestingModule] });
```
oidacra: The `search()` method has an explicit error path that returns `throwError(() => new SyntaxError('Invalid JSON query'))` on malformed JSON, but it isn't tested. Please add coverage for the failure case:

```ts
it('should emit a SyntaxError when query is not valid JSON', () => {
    spectator.service.search('not-valid-json', {}).subscribe({
        next: () => fail('should not emit'),
        error: (err) => {
            expect(err).toBeInstanceOf(SyntaxError);
            expect(err.message).toBe('Invalid JSON query');
        }
    });
    spectator.controller.expectNone('/api/es/search');
});
```

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/store/dot-es-search.store.ts
```diff
@@ -0,0 +1,200 @@
...
+            patchState(store, {
+                emptyStateConfig: {
+                    title: dotMessageService.get('esSearch.results.empty'),
+                    icon: 'pi-search',
+                    subtitle: dotMessageService.get('esSearch.results.empty.hint')
+                }
+            });
+
+            currentUserService
+                .getCurrentUser()
```
oidacra, line 202: The subscribe call here has no error handler. If `getCurrentUser()` errors (HTTP 401, 500, network timeout), `isAdmin` stays permanently `false` and admin users silently lose the `userid` filter UI — no toast, no log. Please add an error handler that surfaces the failure (e.g. via `httpErrorManager.handle(err)`):

```ts
currentUserService
    .getCurrentUser()
    .pipe(take(1))
    .subscribe({
        next: ({ admin }) => patchState(store, { isAdmin: admin }),
        error: (err: HttpErrorResponse) => httpErrorManager.handle(err)
    });
```

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/dot-es-search-page.component.ts
```diff
@@ -0,0 +1,394 @@
...
+    private copyAs(format: 'curl' | 'fetch'): void {
+        const snippet = format === 'curl' ? this.buildCurlSnippet() : this.buildFetchSnippet();
+        navigator.clipboard
+            .writeText(snippet)
+            .catch((err: unknown) => console.warn('Clipboard write failed', err));
+    }
+
+    private buildCurlSnippet(): string {
+        const qs = this.buildApiQueryString();
+        const url = `${window.location.origin}/api/es/search${qs ? '?' + qs : ''}`;
```
oidacra: `window` is being accessed as a global here. Since `DOCUMENT` is already injected, please go through it instead so the call is testable and SSR-safe:

```ts
const origin = this.#document.defaultView?.location.origin ?? '';
const url = `${origin}/api/es/search${qs ? '?' + qs : ''}`;
```

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/store/dot-es-search.store.spec.ts
```diff
@@ -0,0 +1,402 @@
...
+            (searchService.search as jest.Mock).mockReturnValueOnce(
+                throwError(() => new Error('fail'))
+            );
+
+            spectator.service.runSearch();
+            expect(spectator.service.status()).toBe(ComponentStatus.ERROR);
+            expect(errorManager.handle).toHaveBeenCalled();
+        });
+
+        it('should set activeTab to results when explicitly called', () => {
```
oidacra, line 189: The `queryWasCapped` branch (when the parsed query's `size` exceeds `MAX_HITS`, the store rewrites the query and flips `queryWasCapped`) is a real production safety guard but has no test coverage. Please add one:

```ts
it('should cap size to MAX_HITS and set queryWasCapped when query size exceeds the limit', () => {
    spectator.service.setQuery(`{"query":{"match_all":{}},"size":${MAX_HITS + 100}}`);
    spectator.service.runSearch();
    expect(spectator.service.queryWasCapped()).toBe(true);
    const sent = JSON.parse(spectator.service.query());
    expect(sent.size).toBe(MAX_HITS);
});
```

core-web/apps/dotcms-ui/src/app/app.routes.ts
```diff
@@ -156,6 +156,16 @@ const PORTLETS_ANGULAR: Route[] = [
         path: 'usage',
         loadChildren: () => import('@dotcms/portlets/dot-usage').then((m) => m.dotUsageRoutes)
     },
+    {
+        path: 'es-search',
+        canActivate: [MenuGuardService],
+        canActivateChild: [MenuGuardService],
+        providers: [DotEnterpriseLicenseResolver],
```
oidacra, line 163: The other recently-migrated portlet routes (`tags`, `plugins`, `categories`) only use `MenuGuardService` and don't attach `DotEnterpriseLicenseResolver`. If the resolved `isEnterprise` data is consumed somewhere (e.g. to gate the unlicensed-state UI inside the shell), please ignore. Otherwise, the resolver runs on every navigation, fires an extra HTTP call, and its result is never read — consider either consuming `route.snapshot.data.isEnterprise` (e.g. swap the `toSignal` source in the shell to use it) or dropping the resolver to match the surrounding pattern.

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-shell/dot-es-search-shell.component.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    readonly isLicensed = toSignal(
```
oidacra: `toSignal(...)` returns a Signal, so per `ANGULAR_STANDARDS.md` it should use the `$` prefix:

```ts
readonly $isLicensed = toSignal(...);
```

and the template needs to be updated to `$isLicensed()`.

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/dot-es-search-page.component.ts
```diff
@@ -0,0 +1,394 @@
...
+
+    useExample(query: string): void {
+        this.store.setQuery(query);
+        this.helpPopover().hide();
+    }
+
+    copyQuery(query: string): void {
+        navigator.clipboard
+            .writeText(query)
+            .catch((err: unknown) => console.warn('Clipboard write failed', err));
```
oidacra: The three clipboard handlers in this component (`copyQuery`, `copyToClipboard`, `copyAs`) only `console.warn` on failure, which is invisible in production. Common failure cases (`NotAllowedError`, page-not-focused `SecurityError`, Clipboard API missing in some embedded WebViews) leave the user with no signal that the action failed — they'll click again and assume it's broken. Please surface a brief toast via `MessageService` (or use the existing `DotCopyButtonComponent`, which already handles this).

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/store/dot-es-search.store.ts
hmoreras: Fixed in c8e88fb2d6 — renamed `EsSearchState` → `ESSearchState` and `EsSearchActiveTab` → `ESSearchActiveTab` throughout store, component, and spec.

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/store/dot-es-search.store.ts
hmoreras: Fixed in c8e88fb2d6 — renamed `hasResults` → `hasLoadedResults` to accurately reflect that it checks both loaded status and contentlets presence.

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/dot-es-search-page.component.ts
hmoreras: Fixed in c8e88fb2d6 — removed `standalone: true` from both page and shell components.

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-shell/dot-es-search-shell.component.ts
hmoreras: Fixed in c8e88fb2d6 — removed `standalone: true`.

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/dot-es-search-page.component.spec.ts
hmoreras: Fixed in c8e88fb2d6 — replaced tautological assertion with a DOM query on the button's `disabled` property.

core-web/libs/data-access/src/lib/dot-es-search/dot-es-search.service.spec.ts
hmoreras: Fixed in c8e88fb2d6 — rewrote spec to use Spectator's `createHttpFactory` / `SpectatorHttp` pattern.

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/dot-es-search-page.component.html
hmoreras: Fixed in c8e88fb2d6 — replaced `style="font-size: 0.5rem"` with Tailwind utility class `text-[0.5rem]`.

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/dot-es-search-page.component.ts
hmoreras: Fixed in c8e88fb2d6 — renamed all component signals to use the `$` prefix: `$paramsOpen`, `$hasEditorErrors`, `$queryEditorOptions`, `$parsedAggregations`, `$parsedSuggestions`. Updated template and spec accordingly.

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/dot-es-search-page.component.ts
hmoreras: Fixed in c8e88fb2d6 — replaced `private readonly` with ECMAScript `#` for `messageService` and `document`.

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-shell/dot-es-search-shell.component.ts
hmoreras: Fixed in c8e88fb2d6 — replaced `private readonly route` with `readonly #route`.

core-web/libs/data-access/src/lib/dot-es-search/dot-es-search.service.spec.ts
hmoreras: Fixed in c8e88fb2d6 — added test `should emit a SyntaxError when query is not valid JSON` in the Spectator-based spec.

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/store/dot-es-search.store.ts
hmoreras: Fixed in c8e88fb2d6 — added `catchError` to the `getCurrentUser()` pipe in `withHooks.onInit`, routing failures to `DotHttpErrorManagerService.handle`.

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/dot-es-search-page.component.ts
hmoreras: Fixed in c8e88fb2d6 — replaced `window.location.origin` with `this.#document.defaultView?.location.origin ?? ''`.

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/store/dot-es-search.store.spec.ts
hmoreras: Fixed in c8e88fb2d6 — added four tests for the `queryWasCapped` branch: initial false, cap applied when size exceeds MAX_HITS, no cap when within limit, and reset on next run.

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-shell/dot-es-search-shell.component.ts
hmoreras: Fixed in c8e88fb2d6 — renamed `isLicensed` → `$isLicensed` and updated the inline template reference.

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/dot-es-search-page.component.ts
hmoreras: Fixed in c8e88fb2d6 — replaced `console.warn` with `DotGlobalMessageService.error()` in all three clipboard handlers.

core-web/apps/dotcms-ui/src/app/app.routes.ts
hmoreras: ES Search portlet is intentionally enterprise-only following the same paterns as the jsp implementation.

core-web/libs/data-access/src/lib/dot-es-search/dot-es-search.service.ts
hmoreras: not really just keeping same functionality as before.

core-web/libs/dotcms-models/src/lib/dot-es-search.model.ts
hmoreras: Fixed in f57a634e39 — extracted the three inline object literals into named interfaces: `ESHitsTotal` (the `{ value, relation }` total shape), `ESHitsContainer` (replaces the anonymous `hits` type in both `ESSearchEsResponse` and `RawESSearchResponse`), and `ESShards` (replaces the inline `_shards` type).

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/store/dot-es-search.store.ts
hmoreras: EsSearchActiveTab union type already enforces correctness

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/dot-es-search-page.component.html
hmoreras: Unnamed pTemplate is valid for p-splitter panels

core-web/libs/dotcms-models/src/lib/dot-es-search.model.ts
hmoreras: done

core-web/libs/portlets/dot-es-search/src/lib/dot-es-search-page/dot-es-search-page.component.html
hmoreras: Fixed in 36e7bec8bc — replaced both `&nbsp;` usages with `inline-flex items-center gap-1` wrapper spans for spacing, and swapped `<strong class="text-color">` for `<span class="font-bold text-color">` to keep styling Tailwind-explicit.

zJaaal (approved):

dario-daza (approved):

oidacra (approved):

hassandotcms (approved):

PR #35569, opened 2026-05-05
fix(content-import): prevent duplicate tags on CSV import (#35124)

hassandotcms:
Align ImportUtil#processTagFields with ESContentletAPIImpl#relateTags so both paths resolve to the same Tag record. Previously the post-checkin path used a strict (name, siteId) lookup with searchInSystemHost=false and no tagStorage resolution, which on sites whose tagStorage differed from where the tag actually lived (e.g. tag at SYSTEM_HOST, site tagStorage = self) caused saveTag to create a parallel Tag and a second TagInode pointing at it; rendering as duplicate chips on the content.

Switch processTagFields to TagAPI#getTagsInText (smart lookup, prefers persona > site > global) and wipe TagInodes for (inode, fieldVar) before re-adding for strict-replace, idempotent semantics. 

Adds an integration test reproducing the duplicate-tag scenario.

### Checklist
- [ ] Tests
- [ ] Translations

### Additional Info

### Screenshots

This PR fixes: #35124

This PR fixes: #35124

### DIFF
--- a/dotCMS/src/main/java/com/dotmarketing/util/ImportUtil.java
+++ b/dotCMS/src/main/java/com/dotmarketing/util/ImportUtil.java
@@ -4028,8 +4028,15 @@ public class ImportUtil {
             final Field field = entry.getValue();
             final Object value = values.get(entry.getKey());
 
-            if (!field.getFieldType().equals(Field.FieldType.TAG.toString())
-                    || !(value instanceof String)) {
+            if (!field.getFieldType().equals(Field.FieldType.TAG.toString())) {
+                continue;
+            }
+            if (!(value instanceof String)) {
+                if (value != null) {
+                    Logger.warn(ImportUtil.class, String.format(
+                            "Tag field '%s' has unexpected value type '%s'; skipping",
+                            field.getVelocityVarName(), value.getClass().getName()));
+                }
                 continue;
             }
 
--- a/dotCMS/src/main/java/com/dotmarketing/util/ImportUtil.java
+++ b/dotCMS/src/main/java/com/dotmarketing/util/ImportUtil.java
@@ -4032,11 +4032,14 @@ public class ImportUtil {
                 continue;
             }
             if (!(value instanceof String)) {
-                if (value != null) {
-                    Logger.warn(ImportUtil.class, String.format(
-                            "Tag field '%s' has unexpected value type '%s'; skipping",
-                            field.getVelocityVarName(), value.getClass().getName()));
-                }
+                Logger.warn(
+                        ImportUtil.class,
+                        String.format(
+                                "Tag field '%s' has unexpected value type '%s'; skipping",
+                                field.getVelocityVarName(),
+                                value == null ? "null" : value.getClass().getName()
+                        )
+                );
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


dotCMS/src/main/java/com/dotmarketing/util/ImportUtil.java
```diff
@@ -4004,43 +4006,38 @@ private static Contentlet executeWorkflowAction(
...
-            Object value = values.get(entry.getKey());
+            final Field field = entry.getValue();
+            final Object value = values.get(entry.getKey());
 
-            if (field.getFieldType().equals(Field.FieldType.TAG.toString())
-                    && value instanceof String) {
-                String[] tags = ((String) value).split(",");
-                String hostId = getHostId(siteAndFolder);
+            if (!field.getFieldType().equals(Field.FieldType.TAG.toString())
+                    || !(value instanceof String)) {
```
fabrizzio-dotCMS: Let's add a log  WARN in case we encounter a field of type Tag with a non-String value

dotcms-integration/src/test/java/com/dotcms/util/ImportUtilTest.java
```diff
@@ -5336,4 +5339,165 @@ public void testImportMultilingualWithSiteNameAsKeyField() throws DotSecurityExc
...
+            hostField = fieldAPI.byContentTypeAndVar(contentType, hostField.variable());
+            tagsField = fieldAPI.byContentTypeAndVar(contentType, tagsField.variable());
+
+            workflowAPI.saveSchemesForStruct(new StructureTransformer(contentType).asStructure(),
+                    Arrays.asList(schemeStepActionResult1.getScheme()));
+
+            final String contentTitle = "tag-dup-repro-" + suffix;
+            final String csv = "title,hostFolder,tags\r\n" +
+                    contentTitle + "," + testSite.getIdentifier() + "," + tagName + "\r\n";
+
```
fabrizzio-dotCMS, line 5411: I can't recall, but how does it look when we import several tags? Can it be done in one line?
If we can I think its worth covering such a case.. and the new logic previously deletes existing tags. 
So it would be good if we could test such a scenario where a contentlet already has tags, and we replace them

fabrizzio-dotCMS (changes requested):

dotcms-integration/src/test/java/com/dotcms/util/ImportUtilTest.java
hassandotcms: added 2 more test cases for these scenarios

dotCMS/src/main/java/com/dotmarketing/util/ImportUtil.java
```diff
@@ -4019,28 +4021,36 @@ private static void processTagFields(
...
+                continue;
+            }
+            if (!(value instanceof String)) {
+                if (value != null) {
+                    Logger.warn(ImportUtil.class, String.format(
+                            "Tag field '%s' has unexpected value type '%s'; skipping",
+                            field.getVelocityVarName(), value.getClass().getName()));
+                }
+                continue;
+            }
```
fabrizzio-dotCMS, line 4044: if (!(value instanceof String)) {
    Logger.warn(
            ImportUtil.class,
            String.format(
                    "Tag field '%s' has unexpected value type '%s'; skipping",
                    field.getVelocityVarName(),
                    value == null ? "null" : value.getClass().getName()
            )
    );
    continue;
}

So that we know if somehow null was passed

fabrizzio-dotCMS (approved):

PR #35576, opened 2026-05-05
fix: keep the folder id when renaming it (#35399)

### LINKED ISSUE #35399: Folder identifiers not updated on rename, leaving stale folder on receiver after push publish
### Problem Statement

When a folder is renamed in dotCMS, the original identifier is deleted and a brand-new identifier is created for the renamed folder. This breaks push-publish workflows: since the receiver already has the folder under its original identifier, pushing the renamed folder creates a second folder on the receiver (under the new identifier) while the original folder remains — requiring manual deletion on every receiver environment.

The correct behavior is to **keep the same identifier** when renaming a folder and simply update the relevant fields (e.g. \`asset_name\`) on the existing identifier record, so the folder retains its identity across environments.

### Steps to Reproduce

1. Set up a push-publish environment (sender + receiver).
2. Create a folder (e.g. \`/my-folder/\`) and push-publish it — confirm it appears on the receiver.
3. Rename the folder on the sender (e.g. \`/my-folder/\` → \`/my-folder-renamed/\`).
4. Push-publish the renamed folder to the receiver.
5. Inspect the receiver's folder tree.

### Expected Result

- The folder on the receiver is updated in-place to \`/my-folder-renamed/\` — same identifier, new name.
- No orphaned \`/my-folder/\` remains.

### Actual Result

- \`/my-folder-renamed/\` is created on the receiver (new identifier).
- \`/my-folder/\` (original identifier) remains on the receiver as an orphaned folder.
- Manual deletion is required on each receiver environment after every rename.

### Root Cause (Hypothesis)

\`FolderAPI.rename()\` (or the underlying implementation) deletes the existing \`identifier\` record and inserts a new one instead of updating the existing row's \`asset_name\`. This causes the folder to lose its identity (identifier) on rename, breaking push-publish reconciliation which relies on the identifier to match sender and receiver objects.

### Acceptance Criteria

- [ ] Renaming a folder updates the existing \`identifier\` record in place (updating \`asset_name\` / \`parent_path\` as needed) — the identifier value must not change.
- [ ] Push-publishing a renamed folder updates the folder on the receiver in place — no duplicate or orphaned folder is left behind.
- [ ] No manual cleanup is required on receiver environments after a folder rename + push-publish.
- [ ] Regression test: rename a folder, push-publish, verify receiver has exactly one folder at the new path with the original identifier.

### dotCMS Version

Latest (main)

### Severity

High — every folder rename + push-publish leaves orphaned folders on all receiver environments, requiring manual intervention.

### Related Issues

- #35260 — NPE when push publishing a renamed folder (likely same root cause)

gortiz-dotcms:
## Summary

Fixes #35399

Before this change, renaming a folder deleted the old `identifier` row and inserted a new one with a different UUID. On push-publish receivers this appeared as a brand-new folder at the new path while the original folder (old UUID) stayed orphaned, requiring manual cleanup.

### What changed

**Stable identifier on rename (`FolderFactoryImpl`)**

The rename operation now updates `identifier.asset_name` in place instead of replacing the row. The folder UUID is unchanged after a rename. `full_path_lc` is a generated column so it updates automatically when `asset_name` or `parent_path` change.

**Child path updates (`updateChildPaths`)**

`parent_path` for all identifiers in the renamed subtree — including folder identifiers — is updated in bulk, sorted by ascending path depth to satisfy the `identifier_parent_path_trigger` ordering requirement.

**Push-publish detection**

- `version_ts` is bumped for every contentlet in the renamed subtree so push-publish picks them up.
- `mod_date` is bumped for every child sub-folder via a LIKE-based subquery so `DependencyModDateUtil.excludeByModDate` includes them in the bundle.

**Receiver reconciliation (`FolderHandler`)**

Added a fallback lookup: if the receiver cannot find the folder by path or by inode, it searches by the identifier's existing path. This handles independently-provisioned environments where the same identifier UUID exists at a different path after a previous rename that was not yet pushed.

**Cache eviction**

The root folder's `FolderCache` entry is now evicted before the `UPDATE identifier`, using the pre-rename `Identifier` object. This removes a race where a concurrent thread could clear the identifier cache between the UPDATE and `save()`, causing `save()` to evict the wrong (new-path) key.

### Tests updated

`FolderAPITest.renameFolder` and `FolderAPITest.renameFolder_updatesChildrenAndSubChildrenPaths` were rewritten to assert stable-UUID behaviour: identifier UUIDs unchanged after rename, `asset_name` / `parent_path` updated in place, sub-folder `mod_date` bumped, and old paths absent from the DB.

This PR fixes: #35399


dotCMS/src/enterprise/java/com/dotcms/enterprise/publishing/remote/handler/FolderHandler.java
```diff
@@ -202,6 +202,20 @@ private void handleFolders(Collection<File> folders) throws DotPublishingExcepti
...
+        			// the rename/move reconciliation below can update it rather than trying to
+        			// create a duplicate, which would fail with a constraint violation.
+        			if (localFolder == null || !UtilMethods.isSet(localFolder.getInode())) {
+        				if (id != null && UtilMethods.isSet(id.getId())) {
+        					final String existingPath = id.getPath();
+        					if (UtilMethods.isSet(existingPath)) {
+        						localFolder = fAPI.findFolderByPath(existingPath, localHost, systemUser, false);
+        					}
+        				}
+        			}
```
dsolistorres: My understanding is that the `inode` should be equals to the `identifier` for the folders, so not sure if we need this fallback mechanism here (maybe it works for legacy folders where inode is still different than identifier?)

dsolistorres (approved):

PR #35579, opened 2026-05-05
fix(dotAI): search threshold, image error feedback, per-site config routing

### LINKED ISSUE #36032: fix(dotAI): Embeddings search threshold too strict, image errors silent, create route missing
## Description

Three unrelated usability gaps in the dotAI LangChain4J integration:

### 1. Image generation errors silently disappear in Block Editor

When image generation fails, the Block Editor gallery showed nothing instead of an error state. The actual API error message was never propagated — a hardcoded missing-token key was always shown instead. The error should be surfaced as a failed image entry so the existing `dot-empty-container` error UI is triggered.

### 2. Per-site config create route missing

Non-configured sites navigate to `/apps/dotAI/create/:id`, which matched the generic `DotAppsConfigurationDetailComponent` instead of the custom `DotAiConfigDetailComponent`. A `dotAI/create/:id` route needs to be registered pointing to the same JSON textarea + example component used by the edit path.

**NOTE: The change in the threshold value to the 0.5 as default, mentioned on the PR, was rollbacked in another PR to the original 0.25. The issue was not the threshold being strict, but the model we were using for testing.**

## Acceptance criteria

- Dev Tools search threshold input shows the value from `providerConfig.settings.embeddingsSearchThreshold` when configured
- Forcing an image generation API error shows the actual error message in the Block Editor gallery with a Regenerate button
- Navigating to dotAI config for an unconfigured site shows `DotAiConfigDetailComponent` (not the generic form)

## Related PR

Implemented in #35579

ihoffmann-dot:
## Summary

- **Search threshold default raised to 0.5**: Previous default of `.25` (cosine distance ≤ 0.25, i.e. similarity ≥ 0.75) was too strict and returned no results in practice. Updated across all hardcoded locations: `AppKeys`, `SearchResource`, `CompletionsForm`, `EmbeddingsDTO`, `SearchTool`, and `openapi.yaml`. Config endpoint now also exposes `embeddingsSearchThreshold` so Dev Tools reads the configured value on load. **(NOTE: This was rollbacked in another PR to the original 0.25, the issue was not the threshold being strict, but the model we were using for testing)**
- **Image generation error feedback**: When image generation fails in the Block Editor, the error is now surfaced in the gallery instead of silently disappearing — adds a failed image entry to the array so the existing `dot-empty-container` error UI is triggered. The actual API error message is now propagated instead of always showing a hardcoded missing-token key.
- **Per-site config routing**: Non-configured sites navigated to `/apps/dotAI/create/:id` which matched the generic `DotAppsConfigurationDetailComponent`. Added a `dotAI/create/:id` route pointing to `DotAiConfigDetailComponent` (JSON textarea + example), same as the edit path.

## Test plan

- [ ] Dev Tools search — threshold input shows the value from `providerConfig.settings.embeddingsSearchThreshold` if configured
- [ ] Block Editor image generation — force an API error — actual error message appears in the gallery with a Regenerate button
- [ ] Navigate to dotAI config for an unconfigured site — `DotAiConfigDetailComponent` (JSON textarea + example) is shown, not the generic form

## Issue

Fixes: #35150


dotCMS/src/main/java/com/dotcms/ai/rest/ImageResource.java
```diff
@@ -119,9 +119,16 @@ public Response handleImageRequest(@Context final HttpServletRequest request,
...
-        final JSONObject resp = service.sendRequest(aiImageRequestDTO);
-
-        return Response.ok(Marshaller.marshal(resp)).type(MediaType.APPLICATION_JSON_TYPE).build();
+        try {
+            final JSONObject resp = service.sendRequest(aiImageRequestDTO);
+            return Response.ok(Marshaller.marshal(resp)).type(MediaType.APPLICATION_JSON_TYPE).build();
+        } catch (Exception e) {
+            final Throwable root = e.getCause() != null ? e.getCause() : e;
+            final String message = root.getMessage() != null ? root.getMessage() : "Error generating image";
+            return Response.status(Status.INTERNAL_SERVER_ERROR)
```
nollymar: should it be a BAD_REQUEST_ERROR instead?

dotCMS/src/main/webapp/WEB-INF/jsp/dotai/render.jsp
```diff
@@ -14,6 +14,10 @@
             if (!dotAiState.config["providerConfig"]) {
                 document.getElementById("openAIKeyWarn").style.display = "block";
             }
+            const configuredThreshold = dotAiState.config["com.dotcms.ai.embeddings.search.default.threshold"];
```
nollymar, line 17: this works with the old Edit Content, did you test with the new one?

nollymar (approved):

dotCMS/src/main/java/com/dotcms/ai/rest/ImageResource.java
ihoffmann-dot: Done

dotCMS/src/main/webapp/WEB-INF/jsp/dotai/render.jsp
ihoffmann-dot: Yes, the image generation error handling is working in both versions

dario-daza (approved):

core-web/libs/block-editor/src/lib/extensions/ai-image-prompt/ai-image-prompt.plugin.ts
```diff
@@ -94,7 +94,7 @@ export class AIImagePromptView {
             this.#dialogRef.onClose
                 .pipe(takeUntil(this.destroy$))
                 .subscribe((selectedImage: DotGeneratedAIImage) => {
-                    if (selectedImage) {
+                    if (selectedImage?.response) {
                         this.editor.chain().insertImage(selectedImage.response.contentlet).run();
                     }
```
KevinDavilaDotCMS, line 99: Do we need to add the else case?

ihoffmann-dot: We don't need it as the else case is already implicitly handled. closeImagePrompt() is called after the if, so the prompt always closes whether there's a valid response or not

KevinDavilaDotCMS (approved):

ihoffmann-dot:
Closes #36032

PR #35596, opened 2026-05-06
fix(edit-content): clear stale field state on manual locale translation

### LINKED ISSUE #35521: [BUG] Form fields retain stale values when switching between content items
### Problem
When editing content items in dotCMS, stale form field state is not properly cleared in two scenarios: (1) switching between content items of the same type, and (2) changing the locale using "Translate Manually". In both cases, a validation error is triggered even when all fields are correctly filled.

> **Note:** This bug is intermittent

### Scenario 1 — Switching between content items
1. Open the Blog content type editor
2. Edit Blog item A (e.g., set URL Title to 'item-a')
3. Save item A
4. Open Blog item B (item B may have no URL Title or a different value)
5. Observe: The form displays Blog A's URL Title value instead of Blog B's value
6. **Result**: Showing wrong url title

### Scenario 2 — Changing locale via "Translate Manually"
1. Open a content item (e.g., Blog)
2. Change the locale and select **Translate Manually**
3. Fill in all required fields for the new locale
4. Attempt to save
5. **Result**: A validation error appears even though all fields are correctly filled

https://github.com/user-attachments/assets/0a8bacad-847a-44a3-bf14-9f980b9d57a9

### Fields Affected
The issue is not limited to a single field. While URL Title is easily identifiable in Blog, other fields are also impacted:

- **Blog**: URL Title
- **Pages**: Title, URL (URL Title field does not exist in Pages)
- **Events**: Potentially Title, URL, and other fields
- **Scope**: All fields across all content types may be affected

## Expected Behavior
- The form displays the correct field values for the currently active item/locale
- The form clears any stale validation state when switching items or locales
- Saving validates against the current item's actual values
- No validation errors occur when all required fields are correctly filled

## Actual Behavior
- Form fields display stale values from the previously edited item or locale
- Validation fails even when all fields are visibly filled and correct
- The error occurs before any request is made to the backend; validation fails on the frontend only
- Users cannot save content after switching between items or after a locale change with "Translate Manually"

## Technical Details
- **Team suspicion**: The bridge layer may not be performing a proper cleanup/reset of field state when transitioning between content items or locales

## Acceptance Criteria
- [x] **AC1**: When switching to a different content item, all form fields are populated with the correct values from the newly selected item
- [x] **AC2**: When switching to a different content item, any previous validation error messages are cleared
- [x] **AC3**: Validation state is reset when opening a new item, ensuring validation rules apply to the current item's values, not stale values
- [x] **AC4**: A user can successfully edit and save Blog item A, then immediately edit and save Blog item B without encountering validation errors
- [x] **AC5**: A user can successfully edit and save Pages items in sequence without encountering validation errors
- [x] **AC6**: A user can successfully edit and save Events items in sequence without encountering validation errors
- [x] **AC7**: All content types (not just Blog, Pages, and Events) are verified to not retain stale field values when switching between items
- [x] **AC8**: Form field values are correctly displayed and validated across all field types (text, URL, title, and other relevant fields)
- [x] **AC9**: When changing locale and selecting "Translate Manually", the form validation state is fully reset and no false validation errors appear after filling all required fields

## Additional Notes
- No workaround currently exists; users must reload the page or navigate away and back to clear the stale state

adrianjm-dotCMS:
## Summary

Fixes the two scenarios from issue #35521 where stale form field state was not properly cleared:

- **Scenario 1** (`fix(angular-form-bridge)`): When switching between content items of the same type, stale values from the previous item were shown. Fixed by resetting the `AngularFormBridge` instance when called with a different `FormGroup`.
- **Scenario 2** (`fix(edit-content)`): When changing locale via **Translate Manually**, not all fields were cleared — fields with internal state (binary/image, date picker, relationship, custom fields) retained values from the original locale.

https://github.com/user-attachments/assets/e7ad2844-6c67-490a-8af4-ba87ba565880

## Root Cause (Scenario 2)

Two layers of state weren't being cleared on manual translation:

1. **FormControls** — resolution functions fell back to `field.defaultValue` when `contentlet` was `null`, so fields with a configured default showed that default instead of empty.
2. **Component internal state** — field components like binary, date, relationship, and custom fields manage their own visual state independently of the FormControl. Simply rebuilding the `FormGroup` with `null` values wasn't enough to clear what these components rendered.

## Changes

**`edit-content.store.ts`** — Adds `isManualTranslation: boolean` to the store state.

**`locales.feature.ts`** — Sets `isManualTranslation: true` in `patchState` when the user selects "Translate Manually".

**`dot-edit-content-form-resolutions.ts`** — Adds `isManualTranslation` as an optional 4th parameter to `FnResolutionValue` and updates all resolution functions that fell back to `field.defaultValue` when `contentlet` was `null`. When `isManualTranslation` is `true`, they return `null`/`[]` instead, ensuring FormControls start truly empty.

**`dot-edit-content-form.component.ts` / `.html`** — Adds `$shouldRenderFields` signal. When `isManualTranslation` is `true`, the copy-locale effect sets the signal to `false` (destroying all field components) then restores it to `true` via `setTimeout` (recreating them fresh). Components recreated with `contentlet = null` initialize to their empty/default visual state without any prior data.

## Checklist
- [ ] Tests
- [ ] Translations

## Additional Info

The `$shouldRenderFields` toggle causes a one-frame flash while field components are destroyed and recreated. This is intentional and expected — it happens only on the "Translate Manually" path after the user confirms the dialog.

This PR fixes: #35521

This PR fixes: #35521


core-web/libs/edit-content/src/lib/store/features/locales/locales.feature.ts
```diff
@@ -238,6 +238,7 @@ export function withLocales() {
                                             lastTask: null,
                                             state: ComponentStatus.LOADED,
                                             initialContentletState: 'copy',
+                                            isManualTranslation: copyType === 'manual',
```
oidacra, line 242: `isManualTranslation` is set to `true` here for the manual-copy path, but I don't see anywhere it gets reset to `false`. After the user saves, `workflow.feature.ts:289` transitions `initialContentletState` to `'reset'` but doesn't clear this flag, so subsequent `createFormControl` calls in the same store instance will keep suppressing `field.defaultValue`. Could we either reset `isManualTranslation: false` in the workflow/content transitions, or treat it as derived state that's only valid while `initialContentletState === 'copy'`?

core-web/libs/edit-content-bridge/src/lib/bridges/angular-form-bridge.ts
```diff
@@ -77,17 +77,29 @@ export class AngularFormBridge implements FormBridge {
...
+                    `AngularFormBridge: replacing instance while refCount=${AngularFormBridge.refCount}. ` +
+                        'Some custom fields may still hold a reference to the old bridge. ' +
+                        'Ensure all NativeFieldComponent instances are destroyed before the FormGroup changes.'
                 );
             }
-
+            AngularFormBridge.resetInstance();
+            AngularFormBridge.instance = new AngularFormBridge(
+                form,
+                zone,
```
oidacra, line 98: The warning rightly points out that custom fields may still hold a reference to the old bridge. The hazard is that `destroy()` (line 272) decrements the **static** `refCount`, so any leftover consumer of the old instance will decrement the new instance's refCount and can drive it to 0 — triggering `forceDestroy` while the new form is still in use. Two safer options: (a) make `refCount` an instance field so each bridge tracks its own consumers, or (b) early-return from `destroy()` when `this !== AngularFormBridge.instance` (orphan instances shouldn't affect the live one).

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form.component.ts
```diff
@@ -167,6 +169,8 @@ export class DotEditContentFormComponent implements OnInit {
      */
     form!: FormGroup;
 
+    $shouldRenderFields = signal(true);
```
oidacra: Other signals in this component are `readonly` (and most are `protected`). Suggest `protected readonly $shouldRenderFields = signal(true);` for consistency — the wrapper itself shouldn't be reassigned, only its value via `.set()`.

core-web/libs/edit-content/src/lib/store/features/locales/locales.feature.ts
oidacra, line 242: Could we add assertions on `store.isManualTranslation()` in `locales.feature.spec.ts`? The manual-copy test should assert it becomes `true` and the populate-copy test should assert it remains `false` — otherwise a regression in the `copyType === 'manual'` wiring here would pass CI silently.

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form.component.ts
oidacra: The `$shouldRenderFields` toggle is the central mechanism for resetting field components' internal visual state on manual translation, but there's no spec coverage. Could we add a `fakeAsync` test in `dot-edit-content-form.component.spec.ts` that mocks `isCopyingLocale=true` + `isManualTranslation=true`, asserts fields are torn down, then `tick()`s past the timeout and asserts they are recreated?

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form-resolutions.ts
```diff
@@ -120,14 +138,19 @@ const hostFolderResolutionFn: FnResolutionValue<string> = (contentlet, field, qu
...
+    isManualTranslation
+) => {
     const values = contentlet?.[field.variable];
 
     if (Array.isArray(values)) {
         return values.map((item) => Object.keys(item)[0]);
     }
 
-    return field.defaultValue ?? [];
+    return isManualTranslation ? [] : (field.defaultValue ?? []);
```
oidacra, line 155: Minor: when `contentlet[field.variable]` is a populated array, `isManualTranslation` is ignored. This is probably fine in practice because the manual-translation path always passes `contentlet: null`, but the asymmetry isn't obvious from this function alone. Consider either a one-line comment or a pinning test that documents "manual translation never reaches the array branch."

core-web/libs/edit-content/src/lib/store/features/locales/locales.feature.ts
adrianjm-dotCMS: Done. Added `isManualTranslation: false` to the `tapResponse.next` handler in the translated-locale branch so the flag resets whenever the user navigates to an existing translation after a manual copy.

core-web/libs/edit-content-bridge/src/lib/bridges/angular-form-bridge.ts
adrianjm-dotCMS: Done. Applied option (b): added an early return at the top of `destroy()` when `this !== AngularFormBridge.instance`, so orphan instances (replaced when the FormGroup changed) cannot decrement the live bridge's `refCount`.

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form.component.ts
adrianjm-dotCMS: Done. Changed to `protected readonly $shouldRenderFields = signal(true);` for consistency with the other signals in this component.

core-web/libs/edit-content/src/lib/store/features/locales/locales.feature.ts
adrianjm-dotCMS: Done. Added `expect(store.isManualTranslation()).toBe(false)` to the populate-copy test, `expect(store.isManualTranslation()).toBe(true)` to the manual-copy test, and a new `fakeAsync` test that verifies the flag resets to `false` when switching back to a translated locale.

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form.component.ts
adrianjm-dotCMS: Done. Added a `describe('Manual translation — $shouldRenderFields')` block in `dot-edit-content-form.component.spec.ts` with two `fakeAsync` tests: one verifying the toggle (false → true) when `isManualTranslation=true`, and one verifying no toggle when `isManualTranslation=false`.

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form-resolutions.ts
adrianjm-dotCMS: Done. Added an inline comment in the array branch documenting the asymmetry: manual translation always passes `contentlet=null`, so the array branch is never reached in that path.

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form-resolutions.ts
```diff
@@ -37,8 +38,17 @@ const emptyResolutionFn: FnResolutionValue<string> = () => '';
...
+    contentlet,
+    field,
+    _queryParams,
+    isManualTranslation
+) =>
+    contentlet
+        ? (contentlet[field.variable] ?? field.defaultValue)
+        : isManualTranslation
+          ? null
+          : field.defaultValue;
```
zJaaal: use if else, this is too hard to read

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form-resolutions.ts
```diff
@@ -47,10 +57,17 @@ const defaultResolutionFn: FnResolutionValue<string> = (contentlet, field) =>
...
+    field,
+    _queryParams,
+    isManualTranslation
+) => {
     const value = contentlet
         ? (contentlet[field.variable] ?? field.defaultValue)
-        : field.defaultValue;
+        : isManualTranslation
+          ? null
+          : field.defaultValue;
```
zJaaal: Same here, do not nest this if possible

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form-resolutions.ts
```diff
@@ -220,11 +245,15 @@ const relationshipResolutionFn: FnResolutionValue<string> = (contentlet, field)
...
+    field,
+    _queryParams,
+    isManualTranslation
 ) => {
     const value = contentlet
         ? (contentlet[field.variable] ?? field.defaultValue)
-        : field.defaultValue;
+        : isManualTranslation
+          ? null
+          : field.defaultValue;
```
zJaaal: Same here

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form.component.ts
```diff
@@ -297,21 +301,34 @@ export class DotEditContentFormComponent implements OnInit {
...
-                // Keep the revision key in sync so the main reinit effect
-                // doesn't rebuild the form again for the same contentlet.
-                const contentlet = this.$store.contentlet();
-                if (contentlet) {
-                    this.#lastContentletRevisionKey = `${contentlet.identifier}|${contentlet.inode}|${contentlet.modDate}`;
-                }
+                    if (isManualTranslation) {
+                        this.#flushFieldsForRerender();
+                    }
+                });
```
zJaaal: use the inverse of this to flatten the function

Early return when is not copying

`if (!isCopyingLocale) { return; }`

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form.component.ts
```diff
@@ -588,7 +609,8 @@ export class DotEditContentFormComponent implements OnInit {
      */
     private getInitialFieldValue(
         field: DotCMSContentTypeField,
-        contentlet: DotCMSContentlet | null
+        contentlet: DotCMSContentlet | null,
+        isManualTranslation = false
```
zJaaal, line 618: You can leave as it is, but sometimes when we have to add a boolean flag on a function is due to that function doing more than it should, is kinda considered a code smell

Apply this concept anywhere that is needed and evaluate case by case, it does not apply all the time

Also, now that you have more than 2 parameters, we should be using an object

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form.component.ts
```diff
@@ -729,6 +751,17 @@ export class DotEditContentFormComponent implements OnInit {
...
     }
 
+    /**
+     * Briefly removes all field components from the DOM (false → true) so Angular destroys
+     * and recreates them, ensuring each component's internal state (binary preview, date
+     * picker selection, etc.) is reset instead of inheriting stale values from the previous locale.
+     */
+    #flushFieldsForRerender(): void {
+        this.$shouldRenderFields.set(false);
+        const id = setTimeout(() => this.$shouldRenderFields.set(true));
```
zJaaal: im not sure if this `setTimeout` is running on the zone, it can lead to memory leak

Ask claude

zJaaal (approved):
Approving since all comments are related to readability

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form-resolutions.ts
adrianjm-dotCMS: Done. Refactored `defaultResolutionFn`, `textFieldResolutionFn`, and `blockEditorResolutionFn` to use early returns / if-else instead of nested ternaries.

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form-resolutions.ts
adrianjm-dotCMS: Done. Refactored to use an early return instead of the nested ternary.

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form-resolutions.ts
adrianjm-dotCMS: Done. Same refactor applied — early return for the no-contentlet path.

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form.component.ts
adrianjm-dotCMS: Done. Applied early return when `!isCopyingLocale`.

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form.component.ts
adrianjm-dotCMS: Done. Refactored `getInitialFieldValue` to use an object parameter.

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form.component.ts
adrianjm-dotCMS: Done. Fixed the accumulating `onDestroy` callbacks — timeout ID is now an instance field, cleaned up once on destroy.

dario-daza (approved):

PR #35628, opened 2026-05-08
feat(queue): add cloud-agnostic queue publisher abstraction

### LINKED ISSUE #35630: feat(queue): add cloud-agnostic queue publisher abstraction
## Summary

Introduce a `DotQueuePublisher` interface as a cloud-agnostic queue abstraction for dotCMS. This provides the infrastructure layer for publishing analytics events (and future use cases) to external message queues.

## Scope

- `DotQueuePublisher` interface with `publish()` and `isAvailable(queueName)`
- `DotQueuePublisherLocator` for config-driven provider selection (`DOT_QUEUE_PROVIDER`)
- `SqsQueuePublisher` — AWS SQS implementation using IAM/default credentials chain
- `NoOpQueuePublisher` — default no-op for environments without queue infrastructure
- Support for custom providers via fully-qualified class name in config

## Parent Epic

Part of #35048

swicken:
## Summary

Closes #35630

- Introduces a `DotQueuePublisher` interface with config-driven provider selection via `DotQueuePublisherLocator` (`DOT_QUEUE_PROVIDER`)
- Ships with two implementations: `SqsQueuePublisher` (AWS SQS via IAM/default credentials chain) and `NoOpQueuePublisher` (default, discards messages)
- Supports custom providers via fully-qualified class name in config, matching the existing `DotPubSubProviderLocator` extensibility pattern
- Logical queue names map to provider-specific destinations through config (`DOT_QUEUE_SQS_URL_<QUEUE_NAME>`)

This is a standalone infrastructure abstraction with no consumers wired up yet. Analytics event delivery will be connected in a follow-up PR.

### Configuration

| Property | Description | Default |
|----------|-------------|---------|
| `DOT_QUEUE_PROVIDER` | `sqs`, `noop`, or a fully-qualified class name | `noop` |
| `DOT_QUEUE_SQS_REGION` | AWS region | `us-east-1` |
| `DOT_QUEUE_SQS_ENDPOINT` | Custom endpoint (LocalStack/dev only) | _(omit for standard AWS)_ |
| `DOT_QUEUE_SQS_URL_<NAME>` | Maps logical queue name to SQS queue URL | _(required per queue)_ |

## Test plan

- [x] 8 unit tests passing (locator selection, null validation, no-op behavior, missing queue URL)
- [x] `NoOpQueuePublisher` is default when no provider configured
- [x] Bad provider config fails fast with `DotQueueException` (no silent fallback)
- [ ] Manual: verify SQS publish against LocalStack with `DOT_QUEUE_PROVIDER=sqs`

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/queue/DotQueuePublisherLocator.java
+++ b/dotCMS/src/main/java/com/dotcms/queue/DotQueuePublisherLocator.java
@@ -41,14 +41,14 @@ public final class DotQueuePublisherLocator {
             if (instance instanceof DotQueuePublisher) {
                 return (DotQueuePublisher) instance;
             }
-            Logger.error(DotQueuePublisherLocator.class,
-                    "Class '" + provider + "' does not implement DotQueuePublisher, falling back to no-op");
+            throw new DotQueueException(
+                    "Class '" + provider + "' does not implement DotQueuePublisher");
+        } catch (final DotQueueException e) {
+            throw e;
         } catch (final Exception e) {
-            Logger.error(DotQueuePublisherLocator.class,
+            throw new DotQueueException(
                     "Failed to instantiate queue provider '" + provider + "': " + e.getMessage(), e);
         }
-
-        return NoOpQueuePublisher.INSTANCE;
     });
 
     private DotQueuePublisherLocator() {
--- a/dotCMS/src/main/java/com/dotcms/queue/DotQueuePublisherLocator.java
+++ b/dotCMS/src/main/java/com/dotcms/queue/DotQueuePublisherLocator.java
@@ -28,6 +28,12 @@ public final class DotQueuePublisherLocator {
             return new SqsQueuePublisher();
         }
 
+        if (!"noop".equalsIgnoreCase(provider)) {
+            Logger.warn(DotQueuePublisherLocator.class,
+                    "Unrecognized queue provider '" + provider
+                            + "', falling back to no-op. Valid values: sqs, noop");
+        }
+
         return NoOpQueuePublisher.INSTANCE;
     });
 
--- a/dotCMS/src/main/java/com/dotcms/queue/DotQueuePublisherLocator.java
+++ b/dotCMS/src/main/java/com/dotcms/queue/DotQueuePublisherLocator.java
@@ -4,6 +4,7 @@ import com.dotcms.queue.provider.NoOpQueuePublisher;
 import com.dotcms.queue.provider.SqsQueuePublisher;
 import com.dotmarketing.util.Config;
 import com.dotmarketing.util.Logger;
+import com.google.common.annotations.VisibleForTesting;
 import io.vavr.Lazy;
 
 /**
@@ -29,7 +30,23 @@ public final class DotQueuePublisherLocator {
     public static final String DOT_QUEUE_PROVIDER = "DOT_QUEUE_PROVIDER";
 
     private static final Lazy<DotQueuePublisher> INSTANCE = Lazy.of(() -> {
-        final String provider = Config.getStringProperty(DOT_QUEUE_PROVIDER, "noop");
+        final String raw = Config.getStringProperty(DOT_QUEUE_PROVIDER, "noop");
+        return resolve(raw);
+    });
+
+    private DotQueuePublisherLocator() {
+    }
+
+    public static DotQueuePublisher get() {
+        return INSTANCE.get();
+    }
+
+    @VisibleForTesting
+    static DotQueuePublisher resolve(final String rawProvider) {
+        final String provider = (rawProvider == null || rawProvider.trim().isEmpty())
+                ? "noop"
+                : rawProvider.trim();
+
         Logger.info(DotQueuePublisherLocator.class,
                 "Initializing queue publisher with provider: " + provider);
 
@@ -41,7 +58,6 @@ public final class DotQueuePublisherLocator {
             return NoOpQueuePublisher.INSTANCE;
         }
 
-        // Treat as a fully-qualified class name for custom/plugin providers
         try {
             final Class<?> clazz = Class.forName(provider);
             final Object instance = clazz.getDeclaredConstructor().newInstance();
@@ -56,12 +72,5 @@ public final class DotQueuePublisherLocator {
             throw new DotQueueException(
                     "Failed to instantiate queue provider '" + provider + "': " + e.getMessage(), e);
         }
-    });
-
-    private DotQueuePublisherLocator() {
-    }
-
-    public static DotQueuePublisher get() {
-        return INSTANCE.get();
     }
 }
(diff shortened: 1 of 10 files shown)


jcastro-dotcms (approved):

dario-daza (approved):

dsolistorres (approved):

dotCMS/src/main/java/com/dotcms/queue/DotQueuePublisherLocator.java
```diff
@@ -0,0 +1,76 @@
...
+
+    private DotQueuePublisherLocator() {
+    }
+
+    public static DotQueuePublisher get() {
+        return INSTANCE.get();
+    }
+
+    @VisibleForTesting
+    static DotQueuePublisher resolve(final String rawProvider) {
```
freddyDOTCMS, line 45: I think rawProvider must be a enum

dotCMS/src/main/java/com/dotcms/queue/DotQueuePublisherLocator.java
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        try {
+            final Class<?> clazz = Class.forName(provider);
+            final Object instance = clazz.getDeclaredConstructor().newInstance();
+            if (instance instanceof DotQueuePublisher) {
+                return (DotQueuePublisher) instance;
+            }
+            throw new DotQueueException(
+                    "Class '" + provider + "' does not implement DotQueuePublisher");
```
freddyDOTCMS, line 68: I think we can handle this as a enum, and this enum must have a Class attribute to match there the name with the implementation

dotCMS/src/main/java/com/dotcms/queue/provider/NoOpQueuePublisher.java
```diff
@@ -0,0 +1,32 @@
...
+    public static final NoOpQueuePublisher INSTANCE = new NoOpQueuePublisher();
+
+    private NoOpQueuePublisher() {
+    }
+
+    @Override
+    public void publish(final String queueName,
+                        final String messageBody,
+                        @Nullable final Map<String, String> attributes) {
+        Logger.debug(NoOpQueuePublisher.class,
```
freddyDOTCMS, line 24: should not it sent the message directly to the event manager, or maybe this change is going to be in the next PR?

dotCMS/src/main/java/com/dotcms/queue/provider/SqsQueuePublisher.java
```diff
@@ -0,0 +1,130 @@
...
+            throw new DotQueueException("queueName must not be null or empty");
+        }
+        if (messageBody == null || messageBody.isEmpty()) {
+            throw new DotQueueException("messageBody must not be null or empty");
+        }
+
+        final String queueUrl = resolveQueueUrl(queueName);
+
+        final Map<String, MessageAttributeValue> sqsAttributes = new HashMap<>();
+        if (attributes != null) {
```
freddyDOTCMS, line 58: what are this attributes for?, we need to create them for each request, it means create MessageAttributeValue objects in each event batch that is created and are going to be a lot of event

there is a way that we can created this attribute just one and reused it?

dotCMS/src/main/java/com/dotcms/queue/provider/SqsQueuePublisher.java
```diff
@@ -0,0 +1,130 @@
...
+        }
+        return url;
+    }
+
+    private SqsClient getClient() {
+        if (client == null) {
+            synchronized (this) {
+                if (client == null) {
+                    final String region = Config.getStringProperty(REGION_PROP, "us-east-1");
+                    final String endpoint = Config.getStringProperty(ENDPOINT_PROP, "");
```
freddyDOTCMS, line 110: I think we should read these values just once maybe in the static block of the class

dotCMS/src/main/java/com/dotcms/queue/provider/SqsQueuePublisher.java
```diff
@@ -0,0 +1,130 @@
...
+        if (queueName == null || queueName.isEmpty()) {
+            return false;
+        }
+        final String configKey = QUEUE_URL_PREFIX + queueName.toUpperCase(Locale.ROOT);
+        return UtilMethods.isSet(Config.getStringProperty(configKey, ""));
+    }
+
+    private String resolveQueueUrl(final String queueName) {
+        final String configKey = QUEUE_URL_PREFIX + queueName.toUpperCase(Locale.ROOT);
+        final String url = Config.getStringProperty(configKey, "");
```
freddyDOTCMS, line 96: I think we should read it just once, I suppose we can not do it in a static method but maybe include a condition here if we already have the url for the queueName, we don;t need to read it again

dotCMS/src/main/java/com/dotcms/queue/DotQueuePublisherLocator.java
swicken: I left it as not an enum so it can be an extension point outside of core (if someone wants to build a plugin to extend providers), otherwise adding providers requires changes to core. That's the point of the reflection fallback.

dotCMS/src/main/java/com/dotcms/queue/DotQueuePublisherLocator.java
swicken: same answer as the first enum callout

dotCMS/src/main/java/com/dotcms/queue/provider/NoOpQueuePublisher.java
swicken: no-op is for if no one configured a queue provider, it drops the event and logs a debug message rather than causing an exception.

dotCMS/src/main/java/com/dotcms/queue/provider/SqsQueuePublisher.java
swicken: attributes are the metadata passed by the caller, they cannot be pre-created (things like event type, tenant ID etc), its not evident yet as this is not yet wired to any caller.

dotCMS/src/main/java/com/dotcms/queue/provider/SqsQueuePublisher.java
swicken: these ARE read once, if client = null then they are read in and client is created, otherwise it operates from the cached client. SqsQueuePublisher is a singleton.

dotCMS/src/main/java/com/dotcms/queue/provider/SqsQueuePublisher.java
swicken: we could, but config.getStringProperty is already a hashmap stored in memory. We'd be caching a cache.

freddyDOTCMS (approved):

PR #35632, opened 2026-05-08
fix(os-migration): harden startup validation, add emergency halt path, fix thread-safe formatter

### LINKED ISSUE #35631: fix(os-migration): harden startup validation, add emergency halt path, fix thread-safe formatter
## Problem Statement

Four interconnected defects in the OpenSearch migration startup path that can cause silent data loss, uncontrolled reindexing, and corrupted index names under concurrent load:

1. **`IndexStartupValidator` — connectivity failures were silently demoted to warnings**: When OpenSearch was unreachable at startup, the validator logged a `WARN` and allowed dotCMS to continue booting. Any node running in phases 1, 2, or 3 would start with no OS connectivity, routing all dual-writes to a black hole with no error surfaced to operators.

2. **No emergency halt path**: There was no way to revert the active migration phase at runtime without a full restart. If OS became unavailable mid-operation, an operator had no fast path to redirect all traffic back to ES.

3. **`hasEmptyIndices()` was phase-unaware**: In phases 1/2, a freshly-bootstrapped OS index is expectedly empty while the catchup migration is pending. The old logic treated an empty OS index as a signal to trigger a full ES reindex, destroying the existing ES index unnecessarily and causing a full re-crawl.

4. **`ESIndexResource` — `SimpleDateFormat` is not thread-safe**: `timestampFormatter` (a `SimpleDateFormat` instance) was used as a shared mutable field. Under concurrent REST requests for index creation it can produce corrupted or duplicate index names. This bug predates the migration work.

**Impact:** High — affects all dotCMS clusters running OpenSearch migration phases 1–3. Issues 1 and 3 can cause data loss or unintended full reindex events. Issue 4 causes incorrect index naming under concurrent load.

---

## Steps to Reproduce

**Issue 1 — Silent OS black hole:**
1. Configure `FEATURE_FLAG_OPEN_SEARCH_PHASE=1` (or 2/3).
2. Start dotCMS with OS_ENDPOINTS pointing to an unreachable host.
3. Observe: dotCMS starts without error. All OS dual-writes silently fail. ES continues to work, creating a diverged state.

**Issue 2 — No runtime halt:**
1. OS cluster goes down mid-operation with phase ≥ 1 active.
2. Operator has no mechanism to revert `FEATURE_FLAG_OPEN_SEARCH_PHASE` to `0` without restarting all cluster nodes.

**Issue 3 — Unintended full reindex:**
1. Configure `FEATURE_FLAG_OPEN_SEARCH_PHASE=1`.
2. Start dotCMS against a cluster that has a populated ES index but an empty newly-bootstrapped OS index.
3. Observe: `hasEmptyIndices()` returns `true` → `checkAndInitializeIndex()` drops and recreates the ES index → full reindex triggered.

**Issue 4 — Corrupted index name:**
1. Send two concurrent POST requests to `/api/v1/esindex/create`.
2. Observe: `SimpleDateFormat.format()` race condition produces malformed timestamps → index names are corrupted or identical.

---

## Acceptance Criteria

- [ ] `IndexStartupValidator.validateIndexingConfig()` returns `boolean` — `true` if valid, `false` on any failure (version mismatch, endpoint overlap, connectivity error).
- [ ] Connectivity failures in `validateOSVersion()` throw `DotRuntimeException` (same as version mismatches) instead of logging a warning.
- [ ] `IndexConfigHelper.haltMigration()` and `MigrationPhase.reset()` exist and write ordinal `0` back to `FEATURE_FLAG_OPEN_SEARCH_PHASE` at runtime with a WARN log.
- [ ] `ContentletIndexAPIImpl.checkAndInitializeIndex()` calls `haltMigration()` when `validateIndexingConfig()` returns `false`.
- [ ] `hasEmptyIndices()` only checks ES emptiness in phases 0/1/2; only checks OS emptiness in phase 3.
- [ ] `ESIndexResource` uses the thread-safe `threadSafeTimestampFormatter` (`DateTimeFormatter`) instead of `timestampFormatter` (`SimpleDateFormat`) for index name generation.
- [ ] No regression in phase 0 (ES-only) startup flow.

---

## dotCMS Version

`main` branch — affects all clusters with `FEATURE_FLAG_OPEN_SEARCH_PHASE ≥ 1`.

---

## Severity

---

## Links

NA

fabrizzio-dotCMS:
Closes #35631

## Summary

- **IndexStartupValidator**: connectivity failures are now hard errors (throws `DotRuntimeException`); `validateIndexingConfig()` returns `boolean` so callers can react without catching exceptions.
- **IndexConfigHelper**: added `MigrationPhase.reset()` and `haltMigration()` — runtime-only emergency rollback that writes ordinal `0` back to `FEATURE_FLAG_OPEN_SEARCH_PHASE` without a restart.
- **ContentletIndexAPIImpl**: calls `haltMigration()` when startup validation fails; `hasEmptyIndices()` is now phase-aware (empty OS index in phases 1/2 no longer triggers unintended full ES reindex).
- **ESIndexResource**: replaces non-thread-safe `SimpleDateFormat` with immutable `DateTimeFormatter` to prevent corrupted index names under concurrent requests.
- **docs**: adds `OPENSEARCH_MIGRATION_TEST_PLAN.md` — 530-line test plan covering all migration phases, rollback scenarios, and dual-write validation.

## Test plan

- [ ] Phase 0 (ES-only): startup unchanged, `checkAndInitializeIndex()` skips OS validator.
- [ ] Phase 1/2 with OS unreachable: `validateIndexingConfig()` returns `false` → `haltMigration()` resets phase to 0 → fatal log emitted → dotCMS falls back to ES-only.
- [ ] Phase 1/2 with empty OS index + populated ES: no full reindex triggered; ES index preserved.
- [ ] Phase 3 with empty OS index: full reindex is triggered (expected).
- [ ] Concurrent calls to `/api/v1/esindex/create`: index names are well-formed (no `SimpleDateFormat` race).
- [ ] See `docs/backend/OPENSEARCH_MIGRATION_TEST_PLAN.md` for full test matrix.


dotCMS/src/main/java/com/dotcms/content/index/IndexStartupValidator.java
```diff
@@ -48,13 +49,23 @@ public IndexStartupValidator(final OSClientProvider osClientProvider) {
...
+     *         configuration error was detected and dotCMS must not continue booting
      */
-    public static void validateIndexingConfig() {
-        new IndexStartupValidator(CDIUtils.getBeanThrows(OSClientProvider.class)).validate();
+    public static boolean validateIndexingConfig() {
+        try {
+            new IndexStartupValidator(CDIUtils.getBeanThrows(OSClientProvider.class)).validate();
+            return true;
+        } catch (DotRuntimeException e) {
+            Logger.fatal(IndexStartupValidator.class,
```
nollymar: if dotCMS will fallback to ES-only, why failing with a fatal error?

fabrizzio-dotCMS: The idea is that I want it to be clear that if we are in any migration phase — meaning DOT_FEATURE_FLAG_OPEN_SEARCH_PHASE has a value >= 1 — and we fail to connect to OpenSearch, then from my perspective, that is a fatal condition. The migration cannot continue.

What we are currently doing as a fallback is reverting to Elasticsearch by setting DOT_FEATURE_FLAG_OPEN_SEARCH_PHASE = 0, so the system does not attempt to create indices or perform any other OpenSearch-related operations. That is the fallback behavior. But in my opinion, this situation is still fatal.

Initially, I was going to halt the system completely. However, I thought that if I did that, the tester would not have the opportunity to correct the configuration values and restart dotCMS.

dario-daza (approved):

nollymar (approved):

PR #35684, opened 2026-05-12
Issue 35680 intermittent 401 on public pages after push publishing may be caused by empty permission list being cached during permission reference rebuild

### LINKED ISSUE #35680: fix(permissions): intermittent 401 on public pages after Push Publishing may be caused by empty permission list being cached during permission_reference rebuild
### Problem Statement

  After a Push Publishing operation completes on a receiver node, certain public pages intermittently respond with a 401, redirecting anonymous users to
  /dotCMS/login?referrer=/. The error does not resolve on its own — it persists until a content save or publish is performed directly on the production
  environment, which triggers resetPermissionReferences() for that asset, evicts the bad cache entry, and allows the next request to successfully rebuild the
  permission reference. The pages themselves have no configuration changes and their Host-level permissions remain intact throughout.
  
  This issue has been identified through customer report (support ticket #34942). A full reproduction has not yet been achieved. The description below represents
  our current best theory based on code analysis and database inspection of the affected environment.

  ---
  What We Know For Sure
  
  - Push Publishing calls permissionAPI.resetPermissionReferences(contentlet) inside internalCheckin() for every existing HTML page in the bundle. This deletes
  each page's row from permission_reference and broadcasts a cluster-wide cache invalidation.
  - The affected pages have no direct rows in the permission table. They rely entirely on permission_reference pointing to their Host, which correctly holds CMS 
  Anonymous / IHTMLPage / READ=1.
  - The Host's permissions are never touched by this process.
  - When permission_reference is cleared and a request arrives, PermissionBitFactoryImpl.loadPermissions() falls through to _loadParentPermissions() — a walk-up
  through the parent hierarchy that should always recover the Host's permissions.
  - The result of that walk-up — including an empty list — is written to cache unconditionally at line ~1723, before any validity check.

  ---
  The Theory
  
  The walk-up in _loadParentPermissions() calls page.getParentPermissionable(), which in Contentlet.java resolves the parent by calling fAPI.find(this.getFolder(),
   systemUser, false). If that call returns null at that moment — due to transient load, a folder cache miss during the high-demand push window, or any internal
  failure — the walk-up loop never executes and an empty permission list is returned.

  Because the cache write at line ~1723 is unconditional, that empty list is stored in the permission cache. Every subsequent request for the affected page hits
  the cache immediately, gets [], and CMSUrlUtil.isUnauthorizedAndHandleError() fires a 401 for CMS Anonymous. This persists until the cache entry is evicted and a
   subsequent walk-up succeeds.
  
  This is unconfirmed. We have not been able to reproduce it locally, and we have not ruled out other causes.

  ---
  Proposed Fix
  
  In PermissionBitFactoryImpl.loadPermissions() (~line 1723), guard the cache write so empty results are never persisted:

```
  // Before
  permissionCache.addToPermissionCache(permissionKey, permissionList);
```

```
  // After
  if (!permissionList.isEmpty()) {
      permissionCache.addToPermissionCache(permissionKey, permissionList);
  }
```

  If the theory is correct, this prevents a failed walk-up from being written to cache as a permanent empty result. If the theory is incorrect, the change is still
   a safe defensive improvement — caching an empty permission list after a walk-up is never semantically valid under normal conditions, as every asset in dotCMS
  resolves to the System Host which always carries inheritable permissions. The only downside is that a failed walk-up is retried on the next request instead of
  being served from cache.

### Steps to Reproduce

  This bug has not been reproduced locally and is expected to be very difficult to reproduce in a controlled environment. Here is why:

  The failure depends on a race condition between two concurrent events on the receiver node:

  1. Push Publishing clearing a page's permission_reference entry via resetPermissionReferences()
  2. An incoming HTTP request for that same page hitting loadPermissions() at the exact moment fAPI.find(folderId) fails to resolve the parent folder — likely due
  to high load, a cache miss, or a transient DB state during the push window

  Reproducing this locally would require:
  - A multi-node cluster setup with the receiver node under realistic production load
  - A bundle large enough to stress the async permission_reference rebuild thread pool (poolSize=1, maxPoolSize=5, queueCapacity=10000)
  - Precise timing of an anonymous request to a page whose folder lookup transiently fails during the push
  
  In practice, this combination appears to occur rarely — the customer first observed it approximately 2 months after the previous occurrence.

  The fix is proposed on defensive correctness grounds, not on a confirmed reproduction. Caching an empty permission list after a walk-up is never semantically
  valid: every asset in dotCMS resolves through the parent hierarchy to the System Host, which always carries inheritable permissions. Refusing to cache an empty
  result costs one extra walk-up on the next request and has no negative side effects.

### Acceptance Criteria

  - PermissionBitFactoryImpl.loadPermissions() does not write an empty permission list to the permission cache when _loadParentPermissions() returns no results
  - When a walk-up fails to find a parent (e.g., getParentPermissionable() returns null), the next request for that page retries the walk-up instead of being
  served a cached empty result
  - Pages with valid Host-level inheritable permissions for CMS Anonymous remain publicly accessible during and after a Push Publishing operation
  - A unit test covers the scenario where _loadParentPermissions() returns empty: the result must not be written to cache
  - An integration test or manual test verifies that a Push Publishing operation to a receiver node does not cause 401s on pages that inherit permissions from the
  Host
  - No regression in scenarios where an asset genuinely has no inheritable permissions — those cases should still be handled correctly (walk-up reruns on next
  request rather than being permanently cached)

### dotCMS Version

latest in main

### Severity

### Links

https://helpdesk.dotcms.com/a/tickets/34942

freddyDOTCMS:
## Summary
- Guard the `permissionCache.addToPermissionCache()` call in `PermissionBitFactoryImpl.loadPermissions()` to skip caching when the walk-up returns an empty permission list
- Add integration test to `PermissionBitFactoryImplTest` verifying that an empty walk-up result is not written to cache

## Context
During Push Publishing, `resetPermissionReferences()` is called for every existing HTML page in the bundle. This deletes each page's `permission_reference` row and broadcasts a cluster-wide cache invalidation. When a request arrives before the reference is rebuilt, `loadPermissions()` falls through to the walk-up (`_loadParentPermissions()`), which calls `page.getParentPermissionable()` to traverse the parent hierarchy.

If `getParentPermissionable()` returns `null` at that instant (transient DB state, folder cache miss under load), the walk-up loop never executes and returns an empty list. The previous code cached that empty result unconditionally — causing every subsequent request for the page to be served `[]` from cache, bypassing the walk-up entirely, and resulting in a persistent 401 for anonymous users until the next `resetPermissionReferences()` evicted the entry.

The fix refuses to cache empty walk-up results. If the walk-up fails transiently, the next request retries it rather than being served a stale empty result.

## Checklist
- [x] Tests
- [ ] Translations
- [x] Security Implications Contemplated — this fix prevents intermittent anonymous access denial (401) caused by a transient permission cache poisoning during Push Publishing

This PR fixes: #35680

This PR fixes: #35680


freddyDOTCMS:
> **1. There is already a negative-cache TTL mechanism that this PR bypasses rather than uses.**
> 
> `PermissionCacheImpl.addToPermissionCache` (lines 57‑67) does **not** store empty lists as `permissions = []`. It wraps them in an `EmptyPermissionsCacheEntry` with `EMPTY_PERMISSIONS_TTL` (default 300s), and `getPermissionsFromCache` evicts the entry once expired. So the pre-fix behavior was already "negative cache the empty result for 5 min, then re-evaluate" — not the "permanent 401" the new code comment claims. After this PR the negative cache is effectively dead for the walk-up path, and re-evaluation happens on every request.
> 
> That changes the trade-off:
> 
> * The intermittent 401 will clear on the next request instead of waiting up to 5 minutes — good.
> * Permissionables that _legitimately_ have no inheritable permissions of their type will now redo the full Step‑3 walk-up + async `upsertPermissionReferences` on every read. Under load that's a measurable regression for the negative-result path the TTL existed to absorb.
> 
> Two cheaper alternatives worth considering:
> 
> * Lower `EMPTY_PERMISSIONS_TTL` to a few seconds (or 0) only during/just-after PP rebuild, and invalidate affected entries in the rebuild path. This targets the root cause (stale cache after `permission_reference` rebuild) without giving up negative caching globally.
> * Push the `isEmpty()` skip _into_ `PermissionCacheImpl.addToPermissionCache` and delete the unused `EmptyPermissionsCacheEntry` — at least the intent is in one place and there's no dead code.

@claude take account in your review thatin dotCMS every asset resolves through the hierarchy to the System Host which always carries
  inheritable permissions — an empty result from _loadParentPermissions in a healthy environment is essentially theoretical

freddyDOTCMS:
> **Claude finished @freddyDOTCMS's task in 1m 16s** —— [View job](https://github.com/dotCMS/core/actions/runs/25760449707)
> 
> ### Re-review given the "System Host always has inheritable perms" invariant
> You're right — that invariant materially changes the analysis. Conceding where my prior review was off and tightening the parts that still stand.
> 
> #### Revised: Major #1 — negative-cache regression
> **Largely withdrawn.** If the walk-up loop in `_loadParentPermissions` (PermissionBitFactoryImpl.java:1813-1819) is guaranteed to bottom out at System Host and System Host always carries inheritable perms of the relevant type, then an empty return from the walk-up is not a "legitimate negative result we want to absorb." It's a signal of transient bad state. Negative-caching it for 5 minutes via `EmptyPermissionsCacheEntry` was always wrong for this path — it just turned a flicker into a 5‑minute outage. So "we're losing useful negative caching" doesn't really apply here; we're losing only the negative caching of _broken_ state, which is the whole point.
> 
> The one residual question is whether `_loadParentPermissions` can ever return empty in healthy code (not data): e.g., a contenttype/category/template whose `getParentPermissionable()` chain doesn't walk to a Host. If those exist, they'd now re-do Step‑3 + the async `upsertPermissionReferences` on every read. Quick grep of `getParentPermissionable()` implementations would confirm — if every concrete `Permissionable` either returns a parent or is System Host itself, point #1 is fully moot.
> 
> #### Still standing: Major #2 — line 814 inconsistency
> This one isn't resolved by the invariant. `addPermissionsToCache` (PermissionBitFactoryImpl.java:807-816) does:
> 
> ```java
> bitPermissionsList = loadPermissions( permissionable );
> permissionCache.addToPermissionCache( permissionable.getPermissionId(), bitPermissionsList );
> ```
> 
> If `loadPermissions` hits the same transient walk-up failure and returns `[]` (which it now can, since the PR removed the cache write but not the empty return), this caller will still negative-cache the empty list via `EmptyPermissionsCacheEntry` — re-introducing the exact same 5‑minute poison for any flow that warms via `addPermissionsToCache` (cache warmup, certain admin paths).
> 
> Given the invariant ("empty walk-up = broken state, never desired"), the cleanest expression is to push the guard _into_ the cache so it can't be reintroduced:
> 
> ```java
> // PermissionCacheImpl.addToPermissionCache
> protected List<Permission> addToPermissionCache(final String key, List<Permission> permissions) {
>     if (permissions == null || permissions.isEmpty()) {
>         return Collections.emptyList();   // refuse to poison cache with empty walk-up
>     }
>     cache.put(primaryGroup + key, permissions, primaryGroup);
>     return permissions;
> }
> ```
> 
> Then delete `EmptyPermissionsCacheEntry` + `EMPTY_PERMISSIONS_TTL` — they only exist to soften a misfeature that the invariant says shouldn't happen. The guard at line 1724 becomes redundant but harmless. [Fix this →](https://claude.ai/code?q=On%20branch%20issue-35680-intermittent-401-on-public-pages-after-Push-Publishing-may-be-caused-by-empty-permission-list-being-cached-during-permission_reference-rebuild%2C%20move%20the%20empty-list%20guard%20from%20PermissionBitFactoryImpl.java%3A1724%20into%20PermissionCacheImpl.addToPermissionCache%20(lines%2057-67).%20Refuse%20to%20cache%20null%20or%20empty%20permission%20lists%20at%20the%20cache%20layer%2C%20since%20in%20dotCMS%20every%20Permissionable%20resolves%20to%20System%20Host%20which%20always%20carries%20inheritable%20perms%2C%20so%20an%20empty%20walk-up%20result%20is%20always%20transient%20bad%20state.%20Then%20remove%20EmptyPermissionsCacheEntry%20and%20EMPTY_PERMISSIONS_TTL%20%E2%80%94%20they%20only%20exist%20to%20soften%20negative%20caching%20we%20don%27t%20want.%20Update%20PermissionBitFactoryImplTest%20to%20also%20assert%20the%20cache%20layer%20refuses%20empties.&repo=dotCMS/core)
> 
> #### Still standing: Major #3 — comment + root cause
> * The new comment at line 1723 ("must be retried, not persisted as a permanent 401") is now _more_ defensible given the invariant — but call it what it is: an empty walk-up is transient bad state, retry-on-next-read is the recovery path. Wording suggestion: `// Empty walk-up is always transient (System Host carries inheritable perms); don't cache the bad state — let the next read retry.`
> * The PR title still attributes the cause to "permission_reference rebuild." Under the invariant, this is just the most common trigger; a transient `getParentPermissionable() == null` from any cause (folder cache miss, DB hiccup mid-traversal) would produce the same poisoning. The fix correctly addresses the _cache write_, not the rebuild path — which is the right scope, but a one-line note in the commit that we deliberately did _not_ touch `resetPermissionReferences()` would help future readers.
> 
> #### Still standing minor points (unaffected by the invariant)
> * **Test cleanup**: anonymous `Permissionable` with random UUID will trigger `upsertPermissionReferences` async write to `permission_reference` — add a `finally` cleanup, or set `PERMISSION_REFERENCES_UPDATE_ASYNC=false` in the test.
> * **Test asymmetry**: assert both directions — non-empty walk-up _is_ cached, empty _is not_.
> * **Nit**: `assertEquals(Collections.emptyList(), perms)` over `assertTrue(... perms.isEmpty())` for a better failure message.
> 
> #### Net
> With the invariant in mind, the fix's _direction_ is unambiguously correct and the negative-cache concern largely dissolves. The one thing I'd still push to land is moving the guard into `PermissionCacheImpl.addToPermissionCache` so the line‑814 path can't quietly reintroduce the bug — and so the dead `EmptyPermissionsCacheEntry` machinery goes away with it.
> 
> --- · [branch](https://github.com/dotCMS/core/tree/issue-35680-intermittent-401-on-public-pages-after-Push-Publishing-may-be-caused-by-empty-permission-list-being-cached-during-permission_reference-rebuild)

@claude really we are not completly sure about what cause the problem so this is why is better no chance the return value for now

dario-daza (approved):

PR #35718, opened 2026-05-15
fix(ci): use matching Java base image for dotcms-dev variant builds

### LINKED ISSUE #35719: dotcms/dotcms-dev:java-25 image not working
### Problem Statement

It seems like the.dotCMS-dev Java 25 image is shipping with Java 21 inside of it. We need to ship this with Java 25 inside of it. 

### Steps to Reproduce

Try to run dotcms/dotcms-dev:java-25 - it fails to start with 

```
java.lang.UnsupportedClassVersionError: com/dotcms/cdi/business/EnterpriseFeatureInterceptor
has been compiled by a more recent version of the Java Runtime (class file version 69.0),
this version of the Java Runtime only recognizes class file versions up to 65.0
```

### Acceptance Criteria

Work

### dotCMS Version

26.04

### Severity

Critical - System unusable

### Links

https://dotcms.slack.com/archives/C029PPBDE/p1778247599318899

wezell:
Fixes: #35719

## Summary
- `docker/dev-env/Dockerfile` hardcoded `mcr.microsoft.com/openjdk/jdk:21-ubuntu` as its second-stage base, so the `dotcms/dotcms-dev:java-25` tag (and any future variant) was being built on top of Java 21 even though the workflow tagged it as `java-25`.
- Parameterize the base via a new `JAVA_BASE_IMAGE` build-arg (defaults to `jdk:21-ubuntu` for backward compatibility).
- In `cicd_comp_deployment-phase.yml`, emit `java_major` from the `Get SDKMan Version` step (covers both explicit `java-version` override and `.sdkmanrc` fallback) and pass `JAVA_BASE_IMAGE=mcr.microsoft.com/openjdk/jdk:${java_major}-ubuntu` to the dev-image build.

After this change, when `cicd_7-release-java-variant.yml` runs with `java_version=25.x.y-ms`, the dev image is built on `jdk:25-ubuntu`. Standard releases (no override) continue to use `jdk:21-ubuntu` via `.sdkmanrc`.

## Test plan
- [ ] Verify `mcr.microsoft.com/openjdk/jdk:25-ubuntu` manifest exists for both `amd64` and `arm64` (confirmed locally via `docker manifest inspect`)
- [ ] Local build smoke test:
  ```bash
  cd docker/dev-env
  docker buildx build --platform linux/amd64 \
    --build-arg DOTCMS_DOCKER_TAG=java-25 \
    --build-arg JAVA_BASE_IMAGE=mcr.microsoft.com/openjdk/jdk:25-ubuntu \
    -t dotcms-dev:java25-test --load .
  docker run --rm --entrypoint bash dotcms-dev:java25-test -c \
    'echo "=== system java ==="; java -version; echo "=== /java ==="; /java/bin/java -version'
  ```
  Both `java -version` outputs should report Java 25.
- [ ] Trigger `cicd_7-release-java-variant.yml` (workflow_dispatch) against a release branch and confirm the dev-image build step shows the new build-arg and the pushed image runs Java 25.
- [ ] Confirm a standard release build (no `java_version` override) still produces a Java-21 dev image (regression check).

### DIFF
--- a/.github/workflows/cicd_comp_deployment-phase.yml
+++ b/.github/workflows/cicd_comp_deployment-phase.yml
@@ -165,9 +165,12 @@ jobs:
           # This convention may be revisited in future.
 
           # Use provided java-version if set, otherwise read from .sdkmanrc
+          JAVA_MAJOR=""
           if [ -n "${{ inputs.java-version }}" ]; then
             SDKMAN_JAVA_VERSION="${{ inputs.java-version }}"
             echo "Using override Java version: ${SDKMAN_JAVA_VERSION}"
+
+            JAVA_MAJOR=$(echo "${{ inputs.java-version }}" | grep -oE '^[0-9]+')
 
             # Compute Java suffix for artifact isolation
             # Priority: explicit artifact-suffix > derived from java-version major
@@ -175,7 +178,6 @@ jobs:
               RAW_SUFFIX="${{ inputs.artifact-suffix }}"
               echo "Using explicit artifact suffix: ${RAW_SUFFIX}"
             else
-              JAVA_MAJOR=$(echo "${{ inputs.java-version }}" | grep -oE '^[0-9]+')
               RAW_SUFFIX="java-${JAVA_MAJOR}"
               echo "Using derived artifact suffix: ${RAW_SUFFIX}"
             fi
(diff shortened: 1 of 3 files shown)


swicken (approved):
minimal changes, good fix

wezell:
re: 4. Dev image size grows by the OpenSearch JDK (~hundreds of MB)

the issue is that opensearch 1.3 cannot run in java 25.  Hopefully the hundreds of mb we added by not removing OS's embedded java is offset by the fact that we are no longer using the `mcr.microsoft.com/openjdk/jdk:21-ubuntu` image as our base.

PR #35741, opened 2026-05-18
feat(maintenance): add REST endpoint for log file listing #35208

hassandotcms:
Adds GET /api/v1/maintenance/_logFiles which returns a sorted, regex-filtered list of names The names are used as the {fileName} path parameter for the existing tail and download endpoints, replacing the JSP scriptlet that populated the legacy log-files dropdown.

This PR fixes: #35208

This PR fixes: #35208

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/rest/api/v1/maintenance/MaintenanceResource.java
+++ b/dotCMS/src/main/java/com/dotcms/rest/api/v1/maintenance/MaintenanceResource.java
@@ -7,7 +7,6 @@ import com.dotcms.cluster.business.ServerAPI;
 import com.dotcms.concurrent.DotConcurrentFactory;
 import com.google.common.annotations.VisibleForTesting;
 import com.dotcms.rest.InitDataObject;
-import com.dotcms.rest.ResponseEntityListStringView;
 import com.dotcms.rest.ResponseEntityStringView;
 import com.dotcms.rest.ResponseEntityView;
 import com.dotcms.rest.WebResource;
@@ -79,7 +78,6 @@ import java.util.Date;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
-import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import com.dotcms.cdi.CDIUtils;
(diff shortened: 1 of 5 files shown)


dotCMS/src/main/java/com/dotcms/rest/api/v1/maintenance/MaintenanceResource.java
```diff
@@ -197,6 +199,84 @@ public final Response shutdownCluster(@Context final HttpServletRequest request,
...
+                    description = "Unauthorized - authentication required",
+                    content = @Content(mediaType = "application/json")),
+            @ApiResponse(responseCode = "403",
+                    description = "Forbidden - CMS Administrator role and Maintenance portlet access required",
+                    content = @Content(mediaType = "application/json"))
+    })
+    @GET
+    @Path("/_logFiles")
+    @NoCache
+    @Produces({MediaType.APPLICATION_JSON})
```
fabrizzio-dotCMS: we have a TailLogResource

@Path("/v1/logger")
@Tag(name = "System Logging", description = "System logging configuration and management")
public class LoggerResource

I think the new methods should probably live there

dotCMS/src/main/java/com/dotcms/rest/api/v1/maintenance/MaintenanceResource.java
```diff
@@ -197,6 +199,84 @@ public final Response shutdownCluster(@Context final HttpServletRequest request,
...
+    public final ResponseEntityListStringView listLogFiles(
+            @Parameter(hidden = true) @Context final HttpServletRequest request,
+            @Parameter(hidden = true) @Context final HttpServletResponse response) {
+
+        assertBackendUser(request, response);
+
```
fabrizzio-dotCMS: Look at this:

@GET
    @Path("/{fileName:.+}/_tail")
    @JSONP
    @NoCache
    @Produces(SseFeature.SERVER_SENT_EVENTS)
    public final EventOutput getLogs(@Context final HttpServletRequest request,
            @PathParam("fileName") final String fileName, @QueryParam("linesBack") final int linesBack) throws IOException {

        final InitDataObject initData =

fabrizzio-dotCMS (changes requested):

dotCMS/src/main/java/com/dotcms/rest/api/v1/maintenance/MaintenanceResource.java
hassandotcms: make more sense, moving the new endpoint here in TailLogResource

fabrizzio-dotCMS (approved):

PR #35747, opened 2026-05-19
feat(uve): Add validation of FF TOGGLE_LOCK on different pageViews

### LINKED ISSUE #35743: UVE lock overlay covers the iframe in Preview and Live modes when FEATURE_FLAG_UVE_TOGGLE_LOCK is enabled
### Problem Statement

When `FEATURE_FLAG_UVE_TOGGLE_LOCK` is enabled, switching the UVE editor from
**Edit** to **Preview** or **Live (Published)** mode renders the lock overlay on top of the page iframe. The overlay is fully opaque and captures pointer events on its inner element, so:

- The preview/published page is not visible.
- The user cannot scroll the page.

This happens whether or not the page is actually locked, and whether or not it is locked by the current user.

## Expected Behavior
- In Preview and Live modes the user should see the rendered page and be
  able to scroll it freely.
- The lock overlay is an *edit-mode* affordance — it should never appear in
  Preview or Live, regardless of the page's lock state.

## Current Behavior
- A full-viewport dark overlay (the "lock the page to edit" / "page is
  locked" message) is shown on top of the iframe in Preview and Live.
- The page is unreadable and unscrollable.

## Root Cause (from code inspection)
Two computed signals interact incorrectly:

`core-web/libs/portlets/edit-ema/portlet/src/lib/store/features/workflow/withWorkflow.ts`
(approx. lines 123–149) — `$lockOptions` deliberately returns `null` outside
of EDIT mode

Combined with the overlay template
(`dot-uve-lock-overlay.component.html` + `.ts`), which renders a full-size
absolute div with an opaque background and `pointer-events-auto` on the
inner element, the page becomes invisible and unscrollable in non-EDIT modes.

### Steps to Reproduce

1.  Enable the `FEATURE_FLAG_UVE_TOGGLE_LOCK` feature flag.
2. Open any page in the new UVE in **Edit** mode.
3. Toggle the mode selector to **Preview** (or **Published / Live**).
4. Try to view and scroll the page.

### Acceptance Criteria

- Toggling from Edit to Preview never shows the lock overlay, regardless of the page's lock state.
-  Toggling from Edit to Live (Published) never shows the lock overlay,regardless of the page's lock state.
- Preview and Live pages are visible and scrollable.
-  Edit mode behavior is unchanged: overlay still shown when the page isnot yet locked (with feature flag on), or when it's locked by another user / current user lacks lock permission (without feature flag).

### dotCMS Version

26.05.18-01

### Severity

### Links

https://helpdesk.dotcms.com/a/tickets/37379

KevinDavilaDotCMS:
https://github.com/user-attachments/assets/e577f03c-f8e2-44de-a8b7-6163fe8a0c6b

### Screenshots

This PR fixes: #35743

This PR fixes: #35743


dario-daza (approved):

adrianjm-dotCMS (approved):

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/edit-ema-editor.component.spec.ts
```diff
@@ -878,6 +878,19 @@ describe('EditEmaEditorComponent', () => {
...
+                it('should return false when view mode is not EDIT', () => {
+                    patchState(store, {
+                        pageParams: {
+                            ...(store.pageParams() ?? {}),
+                            mode: UVE_MODE.PREVIEW
+                        },
+                        flags: { FEATURE_FLAG_UVE_TOGGLE_LOCK: false },
+                        pageAssetResponse: { pageAsset: lockedByAnotherUser }
+                    });
+
```
oidacra, line 890: This test passes, but it conflates the flag-disabled path with the new mode guard, so it wouldn't necessarily catch a regression of just the guard. To isolate the new behavior, set `flags: { FEATURE_FLAG_UVE_TOGGLE_LOCK: true }` and use a non-locked page asset (e.g. `MOCK_RESPONSE_HEADLESS`). Without the new guard, `flag=true` + non-locked would return `!isLocked = true`; with it, the guard short-circuits to `false` — that's the decisive assertion.

oidacra, line 890: The PR title says "different pageViews" but only `UVE_MODE.PREVIEW` is covered here. The guard `mode !== UVE_MODE.EDIT` also fires for `UVE_MODE.LIVE`, so a future refactor that only special-cases PREVIEW would slip through this suite. Could you add a second `it` block asserting `$showLockOverlay()` is `false` when `mode: UVE_MODE.LIVE`?

PR #35749, opened 2026-05-19
feat(request-cost): publish aggregate token telemetry to external endpoint

### LINKED ISSUE #35750: Publish aggregate request-cost (token) telemetry to external endpoint
## Parent Issue

NA — internal observability enhancement.

## Background

`com.dotcms.cost.RequestCostApiImpl` already aggregates per-request token counts (window + lifetime) and a `ScheduledExecutorService` ticks every `REQUEST_COST_TIME_WINDOW_SECONDS` (default 60s) to emit a local log line. These aggregates aren't queryable across a cluster — they only live in the local log.

## Proposal

Wire the same aggregates into an external time-series sink so per-cluster, per-environment token consumption can be tracked centrally.

- Reuse the existing monitor tick — no new scheduler.
- Dispatch the HTTP POST through `DotConcurrentFactory.getInstance().getSubmitter()` so a slow or dead collector never blocks the monitor thread.
- Opt-in. Activate implicitly when both `REQUEST_COST_PUSH_URL` and `REQUEST_COST_PUSH_TOKEN` are configured.
- Drop-on-failure (observational telemetry, not durable accounting). Rate-limited warn logs.

## Payload

10 fields, exact:

```json
{
  "clusterId": "...",
  "environmentId": "...",
  "timestamp": "2026-05-19T18:48:00Z",
  "windowSeconds": 60,
  "windowRequests": 1234,
  "windowTokens": 5678.5,
  "windowAvgTokensPerRequest": 4.6,
  "lifetimeRequests": 999999,
  "lifetimeTokens": 12345678.25,
  "lifetimeAvgTokensPerRequest": 12.35
}
```

- `clusterId` ← `ClusterFactory.getClusterId()`
- `environmentId` ← `ConfigUtils.getServerId()`
- Token values pre-divided by `REQUEST_COST_DENOMINATOR`.

## Config

| Key | Default | Notes |
| --- | --- | --- |
| `REQUEST_COST_PUSH_URL` | — | URL + token activates the publisher |
| `REQUEST_COST_PUSH_TOKEN` | — | Bearer token |
| `REQUEST_COST_PUSH_TIMEOUT_MS` | 5000 | Per-request timeout |

## Acceptance criteria

- [x] Publisher activates only when both URL and token are set.
- [x] HTTP POST never blocks the monitor scheduler.
- [x] Transport failures are rate-limited warns, snapshot dropped.
- [x] Unit tests lock the JSON shape (exactly 10 fields) and the enable gate.
- [ ] Manual smoke against a request bin shows a snapshot every window tick.

## Implementation

See PR #35749.

wezell:
Fixes #35750

## Summary

- Adds scheduled publishing of the per-window + lifetime request-cost (a.k.a. token) counters to an external REST collector.
- Reuses the existing `RequestCostApiImpl` monitor scheduler tick (default 60s, controlled by `REQUEST_COST_TIME_WINDOW_SECONDS`) — no new threads, no new schedulers.
- HTTP POST runs on `DotConcurrentFactory.getInstance().getSubmitter()` so a slow or dead collector never blocks the monitor scheduler.
- Off by default. Activates implicitly when both `REQUEST_COST_PUSH_URL` and `REQUEST_COST_PUSH_TOKEN` are set.

## Why

`RequestCostApiImpl` already aggregates request counts and token totals (window + lifetime). Today those numbers are only emitted to the local log — they aren't queryable across a cluster. This wires the same aggregates into an external time-series sink so per-cluster, per-environment token consumption can be tracked centrally.

## Wire format

One snapshot per tick (10 fields, exact):

```json
{
  "clusterId": "...",
  "serverId": "...",
  "timestamp": "2026-05-19T18:48:00Z"  // ISO-8601, truncated to seconds,
  "windowSeconds": 60,
  "windowRequests": 1234,
  "windowTokens": 5678.5,
  "windowAvgTokensPerRequest": 4.6,
  "lifetimeRequests": 999999,
  "lifetimeTokens": 12345678.25,
  "lifetimeAvgTokensPerRequest": 12.35
}
```

- `clusterId` ← `ClusterFactory.getClusterId()`
- `serverId` ← `ConfigUtils.getServerId()`
- Both wrapped in `Try` with `"unknown"` fallback so a transient lookup failure doesn't drop the snapshot.
- `timestamp` is `Instant.now().toString()` (ISO-8601, UTC).
- Token values are pre-divided by `REQUEST_COST_DENOMINATOR` (consistent with what gets logged).

## Config keys

| Key | Default | Notes |
| --- | --- | --- |
| `REQUEST_COST_PUSH_URL` | — | Presence + token activates the publisher |
| `REQUEST_COST_PUSH_TOKEN` | — | Bearer token; presence + url activates the publisher |
| `REQUEST_COST_PUSH_TIMEOUT_MS` | `5000` | Per-request timeout |

> **Operability:** `CircuitBreakerUrl` blocks private-subnet URLs unless `ALLOW_ACCESS_TO_PRIVATE_SUBNETS=true` is set on the same JVM. Most internal collectors live on a private network, so set that flag too if your collector hostname resolves to a private address. The flag is captured at first access via a static `Lazy<Boolean>`, so set it before traffic warms up.

Existing knobs (unchanged):

- `REQUEST_COST_ACCOUNTING_ENABLED` — overall request-cost gating
- `REQUEST_COST_TIME_WINDOW_SECONDS` — scheduler cadence (drives publish cadence too)
- `REQUEST_COST_DENOMINATOR` — scaling factor

## Behavior

- Publisher emits **every tick** while enabled, even when the window had zero requests — keeps the time series continuous so idle vs. dead is distinguishable.
- Log line is suppressed on consecutive idle windows (unchanged from prior behavior).

## Failure handling

- Transport errors and non-2xx responses are logged via `Logger.warnEvery(..., 10 min)` and the snapshot is dropped. This is observational telemetry, not durable accounting — buffering on disk would be over-engineered.
- `publish()` returns immediately; the actual POST happens on the shared submitter.

## Test plan

- [x] `RequestCostSnapshotTest` (3 tests) — locks JSON shape: every expected field present, all values round-trip correctly, exactly 10 fields on the wire (guards against accidental field leakage when someone adds a private helper later).
- [x] `RequestCostPublisherTest` (9 tests) — enable gate: disabled by default, URL-alone is not enough, token-alone is not enough, both-set activates, `publish()` is a no-op when disabled.
- [x] `./mvnw compile -pl :dotcms-core -DskipTests` — BUILD SUCCESS.
- [x] `./mvnw test -pl :dotcms-core -Dtest='RequestCostSnapshotTest,RequestCostPublisherTest'` — 12/12 passing.
- [ ] Manual smoke: point `REQUEST_COST_PUSH_URL` at a local request bin, verify a snapshot lands within the window.

## Out of scope

- A localhost-HttpServer integration test of the actual POST. `CircuitBreakerUrl` blocks private subnets via a static `Lazy<Boolean>` that captures `ALLOW_ACCESS_TO_PRIVATE_SUBNETS` on first access — that makes the test fragile to JVM/test ordering. The post path is mostly orchestration; serialization is fully covered by the snapshot test.
- Per-`Price` breakdown in the payload (e.g. `{DB_QUERY: N, ES_QUERY: M}`). Would require additional counters in `incrementCost`; can land as a follow-up if the collector needs it.
- Naming. The package and existing classes still say `RequestCost`; the new field names use "tokens" since that's the better term, but a package rename is a separate, larger refactor (24+ callers).

## Files

- `dotCMS/src/main/java/com/dotcms/cost/RequestCostSnapshot.java` (new)
- `dotCMS/src/main/java/com/dotcms/cost/RequestCostPublisher.java` (new)
- `dotCMS/src/main/java/com/dotcms/cost/RequestCostApiImpl.java` (modified — adds snapshot construction + publisher call in `logRequestCost`)
- `dotCMS/src/test/java/com/dotcms/cost/RequestCostSnapshotTest.java` (new)
- `dotCMS/src/test/java/com/dotcms/cost/RequestCostPublisherTest.java` (new)


dotCMS/src/main/java/com/dotcms/cost/RequestCostPublisher.java
```diff
@@ -0,0 +1,169 @@
...
+        if (!UtilMethods.isSet(url) || !UtilMethods.isSet(token)) {
+            return;
+        }
+        warnOncePlainHttp(url);
+        try {
+            // CircuitBreakerUrl sniffs the rawData and applies Content-Type automatically when
+            // the payload starts with '{', so we don't set it explicitly (would risk a duplicate
+            // header on some Apache HttpClient versions).
+            final Map<String, String> headers = new HashMap<>();
+            headers.put("Authorization", "Bearer " + token);
```
fabrizzio-dotCMS, line 82: Prefer Map.of("Authorization", "Bearer " + token)

dotCMS/src/main/java/com/dotcms/cost/RequestCostPublisher.java
```diff
@@ -0,0 +1,169 @@
...
+ *   <li>{@code REQUEST_COST_PUSH_TOKEN} (bearer token; presence with url activates the publisher)</li>
+ *   <li>{@code REQUEST_COST_PUSH_TIMEOUT_MS} (default 5000)</li>
+ * </ul>
+ * </p>
+ */
+@ApplicationScoped
+public class RequestCostPublisher {
+
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+    private static final int FAIL_LOG_INTERVAL_MS = 10 * 60 * 1000;
```
fabrizzio-dotCMS, line 35: private static final long FAIL_LOG_INTERVAL_MS = TimeUnit.MINUTES.toMillis(10);
and Configurable ?  
just a nit

dotCMS/src/main/java/com/dotcms/cost/RequestCostPublisher.java
```diff
@@ -0,0 +1,169 @@
...
+     * scheme — the bearer token would otherwise traverse the wire in cleartext. Stays a warning
+     * rather than a refusal so a misconfiguration doesn't silently drop telemetry.
+     */
+    private void warnOncePlainHttp(final String url) {
+        if (httpSchemeWarned.get()) {
+            return;
+        }
+        try {
+            final String scheme = URI.create(url).getScheme();
+            if (scheme != null && scheme.equalsIgnoreCase("http")
```
fabrizzio-dotCMS, line 130: f ("http".equalsIgnoreCase(scheme)
                && httpSchemeWarned.compareAndSet(false, true)) {
            ...
        }

dotCMS/src/main/java/com/dotcms/cost/RequestCostSnapshot.java
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * Immutable payload shipped to the external request-cost (a.k.a. request token) collection
+ * endpoint on each scheduled tick. One snapshot = one point in the time series.
+ */
+@JsonAutoDetect(
+        fieldVisibility = Visibility.PUBLIC_ONLY,
+        getterVisibility = Visibility.NONE,
+        isGetterVisibility = Visibility.NONE)
+public final class RequestCostSnapshot {
+
```
fabrizzio-dotCMS, line 15: Immutables

fabrizzio-dotCMS (approved):

dotCMS/src/main/java/com/dotcms/cost/RequestCostSnapshot.java
wezell: I'm not touching Immutables as we are going to java 25 and will use Records - Immutables are tech debt to me at this point.

dotCMS/src/main/java/com/dotcms/cost/RequestCostPublisher.java
wezell: tell that to claude 😄

dotCMS/src/main/java/com/dotcms/cost/RequestCostPublisher.java
wezell: nice, won't change here but good to see this pattern.

dotCMS/src/main/java/com/dotcms/cost/RequestCostPublisher.java
```diff
@@ -0,0 +1,169 @@
...
+                        FAIL_LOG_INTERVAL_MS);
+                return;
+            }
+            final int response = call.response();
+            if (!CircuitBreakerUrl.isSuccessResponse(response)) {
+                Logger.warnEvery(this.getClass(),
+                        "REQUEST_COST_PUSH_FAIL_HTTP",
+                        "Request cost push to " + sanitizeUrlForLog(url) + " returned HTTP " + response,
+                        FAIL_LOG_INTERVAL_MS);
+            }
```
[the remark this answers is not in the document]
wezell: wont' fix here but good comment.

dotCMS/src/main/java/com/dotcms/cost/RequestCostPublisher.java
wezell: should be fine - nothing should be using http at this point.

PR #35762, opened 2026-05-19
feat(release): warn in Slack about PRs in a release that lack QA verdict

nollymar:
## Summary

Adds a QA-coverage warning to the Slack notification we send when a release is generated. For every PR in the release, the new tool inspects its closing-issue references and reports whether the linked issue carries a recognized QA label.

Driven by the question: *"which changes shipped in this release haven't been tested?"*

## How it works

New standalone tool at `.github/scripts/release-qa-status/` (TypeScript, mirrors the shape of `gather-release-data` so the two can be consolidated later).

For each PR between the previous and current release tag:
- **excluded** — bot/dependency-bump/release-machinery → skipped
- otherwise look up closing-issue references via GitHub's GraphQL `closingIssuesReferences` (catches body keywords *and* Development-panel links) and bucket:
  - **failed** — any linked issue carries `QA : Failed`
  - **missing** — linked but no recognized QA label
  - **unlinked** — no closing-issue reference at all
  - **external** — only cross-repo closing refs (e.g. `dotCMS/private-issues#N`) — cannot verify QA from here
  - **passed** — every linked issue is `QA : Passed` or `QA : Not Needed`

The release workflow (`.github/workflows/cicd_6-release.yml`) Report job now installs and runs the tool with `--format slack`, capturing the output into `SLACK_MESSAGE`. The step is `continue-on-error: true` — any failure leaves the announcement intact without a QA section, never blocking a release.

## Slack message (when something is flagged)

Healthy releases keep the current short message. Otherwise the announcement gains a section like:

```
:warning: *QA Coverage* — 6 PRs need review
  failed QA: 0  |  missing QA: 5  |  Orphan PRs: 1  |  Not in the core repo: 0

:warning: *Missing QA (5)*
  Linked issue exists but has no `QA : Passed` / `QA : Not Needed` / `QA : Failed`.
  • <pr-url|#35463> feat(maintenance): thread dump endpoints — @hassandotcms → <issue-url|#35205>
  …

:grey_question: *Orphan PRs (1)*
  PR has no closing-issue reference — QA cannot be verified.
  • <pr-url|#35609> feat(opensearch): SearchAPI router — @fabrizzio-dotCMS
```

## Validation

Ran the tool locally against recent releases:

| Release | failed | missing | unlinked | external | passed | excluded |
|---|---|---|---|---|---|---|
| `v26.05.19-01` | 0 | 5 | 1 | 0 | 0 | 0 |
| `v26.05.18-01` | 0 | 5 | 0 | 1 | 4 | 0 |
| `v26.05.11-01` | 0 | 1 | 1 | 0 | 1 | 0 |

Spot-checked every verdict against `gh issue view` — labels and statuses line up.

Dry-ran the new bash step locally with the same logic the workflow will execute; the multi-line `GITHUB_OUTPUT` capture round-trips correctly into the Slack action's env var.

## Follow-up

Filed #35763 to migrate `gather-release-data` to the same GraphQL approach so release-notes bullets stop missing their issue cross-links. Scoped separately to keep this PR focused.

## Test plan

- [ ] Run workflow_dispatch on `cicd_6-release.yml` with `notify_slack=false` against a recent release tag — verifies the new steps are correctly gated.
- [ ] Run workflow_dispatch with `notify_slack=true` against a release whose previous tag is auto-resolvable — verify the QA section renders in #release Slack channel.
- [ ] Simulate a failure (e.g. invalid token) and confirm the announcement still goes out without a QA section.

### DIFF
--- a/.github/scripts/release-qa-status/.gitignore
+++ b/.github/scripts/release-qa-status/.gitignore
@@ -1,0 +1,3 @@
+node_modules/
+dist/
+*.log
--- a/.github/scripts/release-qa-status/jest.config.js
+++ b/.github/scripts/release-qa-status/jest.config.js
@@ -1,0 +1,14 @@
+/** @type {import('ts-jest').JestConfigWithTsJest} */
+module.exports = {
+  preset: 'ts-jest',
+  testEnvironment: 'node',
+  roots: ['<rootDir>/src'],
+  testMatch: ['**/*.test.ts'],
+  transformIgnorePatterns: [
+    'node_modules/(?!(@octokit|universal-user-agent|before-after-hook)/)',
+  ],
+  transform: {
+    '^.+\\.tsx?$': 'ts-jest',
+    '^.+\\.jsx?$': ['ts-jest', { tsconfig: { allowJs: true } }],
+  },
+};
--- a/.github/scripts/release-qa-status/package-lock.json
+++ b/.github/scripts/release-qa-status/package-lock.json
@@ -1,0 +1,453 @@
+{
+  "name": "@dotcms/release-qa-status",
+  "version": "0.1.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "@dotcms/release-qa-status",
+      "version": "0.1.0",
+      "dependencies": {
+        "@octokit/rest": "^21.1.1"
+      },
+      "bin": {
+        "release-qa-status": "dist/index.js"
+      },
+      "devDependencies": {
+        "@types/node": "^22.15.3",
+        "ts-node": "^10.9.2",
+        "typescript": "^5.8.3"
+      }
+    },
+    "node_modules/@cspotcode/source-map-support": {
+      "version": "0.8.1",
+      "resolved": "https://registry.npmjs.org/@cspotcode/source-map-support/-/source-map-support-0.8.1.tgz",
+      "integrity": "sha512-IchNf6dN4tHoMFIn/7OE8LWZ19Y6q/67Bmf6vnGREv8RSbBVb9LPJxEcnwrcwX6ixSvaiGoomAUvu4YSxXrVgw==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@jridgewell/trace-mapping": "0.3.9"
+      },
+      "engines": {
+        "node": ">=12"
+      }
+    },
+    "node_modules/@jridgewell/resolve-uri": {
+      "version": "3.1.2",
+      "resolved": "https://registry.npmjs.org/@jridgewell/resolve-uri/-/resolve-uri-3.1.2.tgz",
+      "integrity": "sha512-bRISgCIjP20/tbWSPWMEi54QVPRZExkuD9lJL+UIxUKtwVJA8wW1Trb1jMs1RFXo1CBTNZ/5hpC9QvmKWdopKw==",
+      "dev": true,
+      "license": "MIT",
+      "engines": {
+        "node": ">=6.0.0"
+      }
+    },
+    "node_modules/@jridgewell/sourcemap-codec": {
+      "version": "1.5.5",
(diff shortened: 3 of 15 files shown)


nollymar:
Thanks for the thorough review. Addressed in 475074239f:

**Fixed**
- **#1 Failure isolation** — every QA step (Setup Node, npm ci, compute) now has `continue-on-error: true`, so a transient npm-registry hiccup can no longer block the announcement.
- **#2 GraphQL batch swallowing** — re-throws now instead of returning empty refs; the outer `continue-on-error` still drops the QA section cleanly. Prevents the "bogus orphan-PR Slack spam" failure mode.
- **#3 LTS / CLI tags** — the QA steps now skip non-standard tags (`_lts_`, `dotcms-cli-`, non-`v`-prefixed), matching the existing AI release-notes phase. No more silent exit-1.
- **#4 Squash-merge assumption** — added a comment on `extractPRNumbers` noting the dependency on dotCMS's squash-merge policy.
- **#5 Tests** — added jest + ts-jest with 26 unit cases covering `classifyExclusion`, `computePRQA`, and `renderSlack` (aggregation rules, bot edge cases, Slack escaping/truncation, emoji escalation).
- **#6 Bot-login detection** — tightened from `.includes()` to `.startsWith()` on a known-prefix list. `fan-of-github-actions` is no longer treated as a bot. Added a test that locks this in.
- **#7 :rotating_light: when failed > 0** — header emoji escalates so an actual QA failure pops in the channel.
- **#8 Excluded reason in text output** — `renderText` now lists the excluded bucket with the `bot-author` / `dependency-bump` / `release-machinery` reason.
- **#9 Trailing blank line** — bash now skips `printf` when output is empty, and `SLACK_MESSAGE` reverted to the original single-line quoted form so empty `qa_warning` round-trips cleanly.
- **#10 Ellipsis on truncated titles** — `…` appended when title > 140 chars.
- **#11 GraphQL alias safety** — added a comment confirming PR numbers come from a strict integer regex (`extractPRNumbers`), so the alias/argument interpolation is safe by construction. GraphQL aliases can't be parameterized.

Tests all green (26/26).

nollymar:
Second round addressed in 37c1f6950a:

- **#1 tag-list race** — if `toTag` isn't yet in the releases API response, fall back to `tags[0]` (newest indexed) as the predecessor. `listReleases` returns newest-first, so this preserves the QA section's signal on the first release post-merge instead of silently dropping it.
- **#3 Promise.allSettled in `fetchIssueInfos`** — one transient 5xx on `issues.get` can no longer take down the whole batch. Failed entries are treated like `notFound`; `qa.ts` already ignores both when aggregating.
- **#2 page size** — bumped `first: 50` → `first: 100`. The `hasNextPage` warning still fires if a PR ever exceeds that.
- **#4 squash-merge regression detection** — explicit stderr warning when `commits.length > 0 && prNumbers.length === 0`, naming the likely cause (merge-strategy change). The QA section silently vanishing would now have a logged smoking gun.
- **#5 `Number.isInteger` guard** — added in the alias builder so the trust boundary is enforced *at* `fetchClosingIssueRefs` rather than living in a comment.
- **#6 backtick escape** — `escapeSlack` now replaces backticks with apostrophes; PR titles like `bump deps to \`v1.2.3\`` no longer render as Slack monospace.
- **#9 blurb alignment** — `renderText` excluded blurb now lists all four reasons (bot / dependency-bump / version-bump / release-machinery).
- **#10 missing-vs-external coexistence** — added a test asserting that same-repo `missing` wins over coexistent `externalRefs`, with the external ref still surfaced for context.

**Not addressed:**
- **#7 empty-result race** — review confirmed clean; no change needed.
- **#8 wasted API calls for excluded PRs** — flagged as Minor / Harmless. Skipping the GraphQL lookup for excluded PRs would require moving `classifyExclusion` into the fetch path, which couples modules. Acknowledging the trade-off.

Tests: **28/28** pass.

erickgonzalez (approved):

PR #35767, opened 2026-05-20
feat(security): periodic job to encrypt plaintext passwords in user_ table

### LINKED ISSUE #35766: Periodic encryption of plaintext passwords in user_ table
## Description

Add a scheduled job that periodically scans the `user_` table for rows whose `passwordEncrypted` boolean is `false` (i.e., the password is sitting in `password_` as plaintext) and rewrites them with a secure hash via `PasswordFactoryProxy.generateHash`, then flips `passwordEncrypted = true`.

### Why this is needed

Plaintext passwords can land in the `user_` table from several paths:

- Legacy migration data and seed inserts (the bootstrap SQL contains `passwordEncrypted = false` rows by design).
- Bulk user imports that bypass the standard `UserAPI.save(..., validatePassword=true)` path.
- Manual SQL inserts during recovery or environment provisioning.
- Older code paths that set the password but never set the encrypted flag.

A periodic sweep is a defense-in-depth control: it guarantees that no matter how a plaintext password lands in the column, it gets hashed on the next tick rather than persisting forever.

### Behaviour

- Runs every minute by default (configurable cron).
- Implemented as a Quartz `StatefulJob` so concurrent firings cannot overlap.
- Uses `PasswordFactoryProxy.generateHash` — the same hashing utility used everywhere else in the platform (authentication continues to work because `authPassword` accepts the resulting hash).
- Skips rows where `password_ is null`.
- Logs per-cycle counts (only when there is work to do — silent when the table is clean).

### Configuration

| Property | Default | Effect |
| --- | --- | --- |
| `ENABLE_ENCRYPT_PLAIN_PASSWORDS_JOB` | `true` | Kill switch. Checked at startup (job is not scheduled if `false`) **and** at every firing (so it can be flipped at runtime without a restart). |
| `ENCRYPT_PLAIN_PASSWORDS_CRON_EXPRESSION` | `0 0/1 * * * ?` | Standard Quartz cron expression. |

## Acceptance Criteria

- [ ] `EncryptPlainPasswordsJob` exists in `com.dotmarketing.quartz.job` implementing `StatefulJob`.
- [ ] Job is registered in `DotInitScheduler` and runs every minute by default.
- [ ] Toggle via `ENABLE_ENCRYPT_PLAIN_PASSWORDS_JOB` works at both startup and runtime.
- [ ] Integration tests cover: happy path hash + auth roundtrip, already-encrypted rows untouched, null password skipped, multiple rows in one pass, disabled-flag no-op.
- [ ] No regression in existing user auth flows.

## Additional Context

This is a backend-only change. No API contract, schema, or frontend touched.

wezell:
## Proposed Changes

Adds **`EncryptPlainPasswordsJob`** — a Quartz `StatefulJob` that runs every 5 minutes, scans the `user_` table for rows whose `passwordEncrypted` flag is `false`, hashes the cleartext value via `PasswordFactoryProxy.generateHash`, and flips the flag to `true`.

Defense-in-depth against any code path that lands a plaintext password in `user_.password_` — migrations, bulk imports, manual SQL recovery, or older code that set the password without the encrypted flag. Once the job ticks, the row is hashed using the same utility as the rest of the platform, so existing login (`authPassword`) continues to work transparently.

## Files Touched

| File | Change |
| --- | --- |
| `dotCMS/src/main/java/com/dotmarketing/quartz/job/EncryptPlainPasswordsJob.java` | **New.** `StatefulJob` implementation. |
| `dotCMS/src/main/java/com/dotmarketing/init/DotInitScheduler.java` | Registers the new job (mirrors `FreeServerFromClusterJob` pattern). |
| `dotcms-integration/src/test/java/com/dotmarketing/quartz/job/EncryptPlainPasswordsJobTest.java` | **New.** Five integration tests. |

## Configuration

| Property | Default | Effect |
| --- | --- | --- |
| `ENABLE_ENCRYPT_PLAIN_PASSWORDS_JOB` | `true` | Kill switch. Checked at startup (job not scheduled if `false`) **and** at every firing — flip at runtime without a restart. |
| `ENCRYPT_PLAIN_PASSWORDS_CRON_EXPRESSION` | `0 0/5 * * * ?` | Standard Quartz cron. |

## Why a periodic job rather than a one-off migration

The risk of a fresh plaintext row landing in `user_` is non-zero on a live system (admin tooling, recovery scripts, bulk imports that bypass `UserAPI`). A standing sweep catches them within one minute regardless of how they got there.

The query is cheap: `passwordEncrypted = false` is an extremely selective predicate, so in steady state the job does an index/sequential check that finds zero rows and returns immediately. A partial index `CREATE INDEX ... ON user_ (userId) WHERE passwordEncrypted = false` would be the right escalation if perf ever becomes a concern, but is unnecessary today.

## Test Plan

Integration tests (`EncryptPlainPasswordsJobTest`):

- [x] **Happy path** — plaintext row gets hashed; `authPassword(plaintext, storedHash)` returns `AUTHENTICATED`.
- [x] **Already encrypted row** — left untouched (no double-hashing).
- [x] **Null password** — skipped (no UPDATE issued).
- [x] **Multiple rows** — all hashed in a single pass.
- [x] **Disabled flag** — `ENABLE_ENCRYPT_PLAIN_PASSWORDS_JOB=false` makes the firing a no-op.

Run locally:
```bash
./mvnw verify -pl :dotcms-integration -Dcoreit.test.skip=false -Dit.test=EncryptPlainPasswordsJobTest
```

## Rollback safety

Pure additive — new class, new scheduler registration, new test. No schema change, no API contract change, no frontend touched. If anything misbehaves in production, flip `ENABLE_ENCRYPT_PLAIN_PASSWORDS_JOB=false` and the job becomes a no-op immediately; revert the commit for a full backout.

Refs #35766

This PR fixes: #35766

### DIFF
--- a/dotCMS/src/main/java/com/dotmarketing/init/DotInitScheduler.java
+++ b/dotCMS/src/main/java/com/dotmarketing/init/DotInitScheduler.java
@@ -16,6 +16,7 @@ import com.dotmarketing.quartz.job.CleanUnDeletedUsersJob;
 import com.dotmarketing.quartz.job.DeleteInactiveLiveWorkingIndicesJob;
 import com.dotmarketing.quartz.job.DeleteSiteSearchIndicesJob;
 import com.dotmarketing.quartz.job.DropOldContentVersionsJob;
+import com.dotmarketing.quartz.job.EncryptPlainPasswordsJob;
 import com.dotmarketing.quartz.job.FreeServerFromClusterJob;
 import com.dotmarketing.quartz.job.PruneTimeMachineBackupJob;
 import com.dotmarketing.quartz.job.ServerHeartbeatJob;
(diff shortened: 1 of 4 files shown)


wezell:
Thanks for the review — all four substantive findings addressed in `bc1d54d`:

**1 · UPDATE race** → `UPDATE_HASHED_PASSWORD_SQL` now ends with `... and passwordEncrypted = ? and password_ = ?` (binding `false, plaintext`). If anyone changed the row between SELECT and UPDATE, the affected-rows count is 0, the row is logged at debug as skipped, and the concurrent writer's value is preserved. New test `test_concurrent_change_is_not_clobbered` exercises this.

**2 · No batch limit** → added `ENCRYPT_PLAIN_PASSWORDS_BATCH_SIZE` (default **500**). SELECT now ends with `LIMIT ?`. Remaining rows are caught on the next tick — a bulk-import of 50k rows is bounded to ~100 ticks rather than one 2.5-hour firing.

**3 · Test/scheduler race** → `@BeforeClass` now calls `QuartzUtils.getScheduler().deleteJob(JOB_NAME, DOTCMS_JOB_GROUP_NAME)` so the real cron is unscheduled for the duration of the test class. The in-test `TestJobExecutor.execute()` calls are the only invocations.

**5 · Cron frequency** → default lowered from `0 0/1 * * * ?` to `0 0/5 * * * ?`. Still overridable via `ENCRYPT_PLAIN_PASSWORDS_CRON_EXPRESSION`.

**Minor nits** → all three applied: `this.getClass()` replaced with `EncryptPlainPasswordsJob.class`, FQN `java.util.*` cleaned up, hardcoded `"dotcms.org"` replaced with `PublicCompanyFactory.getDefaultCompany().getCompanyId()` (resolved once in `@BeforeClass`).

dotCMS-Machine-User (approved):

Issue #35602, opened 2026-05-06
Add VelocityHealthCheck for Kubernetes readiness probe (follow-up to spike #35329)


dsolistorres:
## Description

Spike #35329 identified that the Velocity engine can enter a broken state where the global macro library fails to load while `engine.init()` reports success (see companion ticket #35601 for the underlying fix). There is currently no health check that would catch this state before K8s starts routing traffic to the pod.

dotCMS already has a health-check framework with readiness/liveness separation — see `dotCMS/src/main/java/com/dotcms/health/api/HealthCheck.java` and the existing CDI checks at `dotCMS/src/main/java/com/dotcms/health/checks/cdi/{DatabaseHealthCheck, ElasticsearchHealthCheck, CacheHealthCheck}.java`. A `VelocityHealthCheck` slots in alongside these.

The probe is simple — evaluate a tiny VTL string that invokes `#renderMarks($null)` and check whether the output contains the literal string `#renderMarks` (which means the macro was not resolved). The proof-of-concept logic already exists in the spike repro at `VelocityUtil.verifyMacroLibraryLoaded()` on branch `spike-35329-velocity-repro` and should be moved into the new health check.

## Acceptance Criteria

- [ ] New class `com.dotcms.health.checks.cdi.VelocityHealthCheck` extends `HealthCheckBase`, registered via CDI alongside the other CDI health checks.
- [ ] `isReadinessCheck()` returns `true`; `isLivenessCheck()` returns `false` (one-time init failures should not cause pod restart loops).
- [ ] `getName()` returns `"velocity"`.
- [ ] `getOrder()` runs after database and cache checks (they are prerequisites for Velocity).
- [ ] `performCheck()` evaluates the probe template `#renderMarks($null)` through `VelocityUtil.getEngine()`. Returns `DOWN` when output contains literal `#renderMarks`; `UP` otherwise.
- [ ] Probe completes in under 50ms per call (acceptable for K8s probe cadence at 5–10s intervals).
- [ ] `/readyz` endpoint (already wired up in `HealthProbeServlet`) reflects `DOWN` when the macro library failed to load.
- [ ] Unit test: with a mock engine where `#renderMarks` is registered, returns `UP`.
- [ ] Unit test: with a mock engine that renders the probe input verbatim, returns `DOWN`.

## Additional Context

[...]

Issue #35329, opened 2026-04-15
Spike: Velocity macro library (VM_global_library.vm) not loaded before request serving on pod restart — #renderMarks renders as literal text


dotCMS-Machine-User:
## Summary

After a pod restart (triggered by JVM memory pressure, a Kubernetes node change, or a dotEvergreen update), the Velocity macro library (`VM_global_library.vm`) is sometimes not fully loaded before the application begins serving HTTP requests. When this happens, macros defined in that library — primarily `#renderMarks($element)` and `#editContentlet` — render as literal text on all public-facing pages. The issue self-resolves once the engine fully initializes, typically within 20–30 minutes, or after a targeted node restart.

This is a recurring platform-level bug that has affected multiple cloud customers across at least five support tickets spanning February–April 2026.

---

## Observed Behavior

- Velocity macros (`#renderMarks($element)`, `#editContentlet`) appear as raw, unrendered text in LIVE/rendering mode
- EDIT mode is unaffected (uses a different rendering path)
- Issue is intermittent and self-resolves after ~20–30 minutes, consistent with engine initialization completing
- In multi-node environments, the issue may affect only one node at a time, making it appear random to end users
- Cache flush and manual page republishing do not resolve it — the issue is in the engine state, not the content cache
- No errors visible in dotCMS application logs in all cases (at least one case had no log errors at all)

---

### VelocityEngine is lazily initialized

`VelocityUtil.java:59` uses `Lazy.of()` to initialize the `VelocityEngine`. The engine loads `VM_global_library.vm`, `dotCMS_library.vm`, and `dotCMS_library_ext.vm` at init time via the `velocimacro.library` config property.

If the first rendering request arrives before `Lazy.of()` has resolved (i.e., the engine is not yet initialized), the macro library is not available. With `velocimacro.library.autoreload=false`, there is no self-correction mechanism — macros remain unresolvable until the engine is fully re-initialized or the process restarts.

**Relevant config:**
```
velocimacro.library=VM_global_library.vm,dotCMS_library.vm,dotCMS_library_ext.vm
velocimacro.library.autoreload=false
```

### No readiness probe gates traffic on Velocity engine readiness

The Kubernetes deployment spec does not appear to include a readiness probe that verifies Velocity engine initialization before routing traffic to a pod. This means a pod can receive live traffic while still in the startup phase, before the macro library is loaded.

### Apache Velocity upstream behavior

[...]

PR #35768, opened 2026-05-20
fix(velocity): fail loud when global macro library load fails (#35601)

### LINKED ISSUE #35601: Velocity: fail loud when global macro library load fails (follow-up to spike #35329)
## Description

Spike #35329 confirmed the root cause of the recurring "macros render as literal text after pod restart" issue: dotCMS's patched `VelocimacroFactory` silently swallows `ResourceNotFoundException` when loading the configured `velocimacro.library` files (`VM_global_library.vm`, `dotCMS_library.vm`, `dotCMS_library_ext.vm`) at engine init time.

See `dotCMS/src/main/java/org/apache/velocity/runtime/VelocimacroFactory.java:194-217`:

```java
try {
    Template template = rsvc.getTemplate(lib);
    ...
} catch(ResourceNotFoundException rnse){
    Logger.warn(this.getClass(), rnse.getMessage());   // ← silent
} catch (Exception e) {
    throw new VelocityException(msg, e);                // ← Apache's original behavior
}
```

When the resource lookup fails (a transient I/O error during early startup, e.g. EFS-backed volumes in K8s), `engine.init()` returns successfully and the engine reports ready, but the global macro library is never registered. `#renderMarks($element)` and `#editContentlet` then render as literal text on every public page until the JVM is restarted.

## Acceptance Criteria

- [ ] New config flag `velocimacro.library.fail-on-missing` (default `false`) added to `dotCMS/src/main/resources/system.properties`.
- [ ] When the flag is `true` and any configured `velocimacro.library` file throws `ResourceNotFoundException` during init, `engine.init()` throws `VelocityException` with the failed library file name in the message (matches Apache's original fail-fast behavior).
- [ ] When the flag is `false`, the current silent-warn behavior is preserved (backward-compat for self-hosted operators that may rely on optional libraries).
- [ ] The `Logger.warn` at `VelocimacroFactory.java:209` is upgraded to `Logger.error` regardless of the flag value, with the library file name in the message.
- [ ] After the macro-library for-loop completes, an `INFO` log line summarizes load results — greppable in startup logs (e.g. `Velocimacro libraries loaded: VM_global_library.vm=OK, dotCMS_library.vm=OK, ...`).
- [ ] Unit test (in `dotCMS/src/test/java/com/dotcms/rendering/velocity/util/VelocityUtilTest.java` or new test class): simulating `ResourceNotFoundException` for `VM_global_library.vm` throws `VelocityException` when flag is `true`.
- [ ] Unit test: same simulation with flag=`false` logs an error and returns normally.
- [ ] No regression in `dotcms-integration/src/test/java/com/dotcms/rendering/velocity/VelocityUtilTest.java`.

## Additional Context

- **Companion ticket**: VelocityHealthCheck readiness probe — that ticket prevents K8s from routing traffic to a pod in this broken state. This ticket prevents the broken state from being silent in the first place.
- **Files changed (expected)**: `VelocimacroFactory.java`, `system.properties`, plus tests.

## Additional Note
- We should discuss if the default value for the new config flag `velocimacro.library.fail-on-missing` should be `true` or `false`. If we set to false, we still have the warning in the log and also the new VelocityHealthCheck will alert about the problem. We can confirm with the cloud team if this behavior will work fine for cloud environments or if a better solution can be implemented.

dsolistorres:
## Summary

Closes #35601 (follow-up implementation from spike #35329).

`VelocimacroFactory` silently swallowed `ResourceNotFoundException` when loading the configured `velocimacro.library` files at engine init, allowing `engine.init()` to return successfully with no global macros registered. After K8s pod restarts with transient I/O errors (notably on EFS-backed volumes), `#renderMarks($element)` and `#editContentlet` rendered as literal text on every public page until the JVM was restarted.

## Changes

- **`VelocimacroFactory.initVelocimacro()`** — track loaded/failed libraries through the load loop. The `ResourceNotFoundException` catch now logs at `ERROR` with the failed library file name. After the loop, an `INFO` summary line lists loaded and failed libraries (greppable for ops). When any library failed, throw `VelocityException` with an actionable message naming the failed library and the opt-out flag.
- **New flag** `velocimacro.library.fail-on-missing` (default `true`) preserves the legacy silent-warn behavior for self-hosted operators who intentionally configure optional libraries.
- **`RuntimeConstants.VM_LIBRARY_FAIL_ON_MISSING`** — new constant.
- **`system.properties`** — explicit default with a one-line comment.
- **`VelocimacroFactoryTest`** *(new)* — three unit tests covering `fail-on-missing=true`, `fail-on-missing=false`, and the default.

## Why this is enough

`engine.init()` throws → `VelocityUtil.getEngine()` (`VelocityUtil.java:65`) rethrows as `DotRuntimeException` → propagates out of `InitServlet.init()` at line 199 → `dotcms.started.up` is never set → existing `ServletContainerHealthCheck` already gates `/readyz` on that flag → K8s readiness probe fails → no traffic routed → eventual pod restart gives EFS time to settle.

Companion ticket #35602 (`VelocityHealthCheck`) is still valuable as defense-in-depth — it catches the opt-out case (`fail-on-missing=false`) and any future post-init macro corruption that this flag wouldn't see.

## Test plan

- [x] `VelocimacroFactoryTest` — 3/3 pass (`fail-on-missing=true` throws, `false` succeeds, default throws). Runtime ERROR/INFO log lines verified in test output.
- [x] `VelocityUtilTest` — 20/20 pass (no regression).
- [x] **Local smoke test on real dotCMS startup**: temporarily added `bogus-smoke-test.vm` to `velocimacro.library`, ran `just dev-run`. Observed:
  - `ERROR runtime.VelocimacroFactory - Velocimacro : VM library not found : bogus-smoke-test.vm : ...`
  - `INFO  runtime.VelocimacroFactory - Velocimacro libraries loaded: [VM_global_library.vm, dotCMS_library.vm, dotCMS_library_ext.vm] ; failed: [bogus-smoke-test.vm]`
  - `VelocityException: Velocimacro : required VM libraries failed to load: [bogus-smoke-test.vm]. Set velocimacro.library.fail-on-missing=false to allow startup with missing libraries.`
  - Stack trace propagation through `InitServlet.init(InitServlet.java:199)` and rethrow as `DotRuntimeException`
  - `InitServlet init Completed` **absent** from log (init aborted before `dotcms.started.up=true`)
  - `/dotmgt/readyz` returns **HTTP 503**
  - `/dotmgt/livez` returns **HTTP 503**

This PR fixes: #35601

### DIFF
--- a/dotCMS/src/main/java/org/apache/velocity/runtime/VelocimacroFactory.java
+++ b/dotCMS/src/main/java/org/apache/velocity/runtime/VelocimacroFactory.java
@@ -20,6 +20,7 @@ package org.apache.velocity.runtime;
  */
 
 import com.dotcms.rendering.velocity.services.DotResourceLoader;
+import com.dotmarketing.util.Config;
 import com.dotmarketing.util.Logger;
 import java.io.StringReader;
 import java.util.ArrayList;
@@ -172,8 +173,12 @@ public class VelocimacroFactory
                      macroLibVec.add(libfiles);
                  }
 
-                 final boolean failOnMissing = rsvc.getBoolean(
-                         RuntimeConstants.VM_LIBRARY_FAIL_ON_MISSING, true);
+                 // Feature flag: when true, engine init throws VelocityException if any
+                 // configured velocimacro.library file fails to load. Default is false to
+                 // preserve historical behavior; operators opt in once they've verified
+                 // their library files load cleanly (see issue #35601).
+                 final boolean failOnMissing = Config.getBooleanProperty(
+                         "VELOCITY_LIBRARY_FAIL_ON_MISSING", false);
                  final List<String> loadedLibraries = new ArrayList<>();
                  final List<String> failedLibraries = new ArrayList<>();
 
@@ -243,8 +248,8 @@ public class VelocimacroFactory
                      throw new VelocityException(
                              "Velocimacro : required VM libraries failed to load: "
                                      + failedLibraries
-                                     + ". Set " + RuntimeConstants.VM_LIBRARY_FAIL_ON_MISSING
-                                     + "=false to allow startup with missing libraries.");
+                                     + ". Set VELOCITY_LIBRARY_FAIL_ON_MISSING=false (or unset)"
+                                     + " to allow startup with missing libraries.");
                  }
              }
 
--- a/dotCMS/src/main/java/org/apache/velocity/runtime/VelocimacroFactory.java
+++ b/dotCMS/src/main/java/org/apache/velocity/runtime/VelocimacroFactory.java
@@ -172,6 +172,11 @@ public class VelocimacroFactory
                      macroLibVec.add(libfiles);
                  }
 
+                 final boolean failOnMissing = rsvc.getBoolean(
+                         RuntimeConstants.VM_LIBRARY_FAIL_ON_MISSING, true);
+                 final List<String> loadedLibraries = new ArrayList<String>();
+                 final List<String> failedLibraries = new ArrayList<String>();
+
                  for(int i = 0, is = macroLibVec.size(); i < is; i++)
                  {
                      String lib = (String) macroLibVec.get(i);
@@ -205,9 +210,16 @@ public class VelocimacroFactory
                              twonk.template = template;
                              twonk.modificationTime = template.getLastModified();
                              libModMap.put(lib, twonk);
+                             loadedLibraries.add(lib);
                          }
                          catch(ResourceNotFoundException rnse){
-                        	 Logger.warn(this.getClass(),rnse.getMessage());
+                             // Surfaces the failed library by name so operators can grep startup logs.
+                             // See dotCMS issue #35601 — previously this was a generic WARN that
+                             // produced silent macro-rendering failures after pod restarts.
+                             Logger.error(this.getClass(),
+                                     "Velocimacro : VM library not found : " + lib
+                                             + " : " + rnse.getMessage());
+                             failedLibraries.add(lib);
                          }
                          catch (Exception e)
                          {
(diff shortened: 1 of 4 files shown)


dsilvam (approved):

dotCMS/src/main/resources/system.properties
```diff
@@ -118,6 +118,10 @@ output.encoding=UTF-8
 
 velocimacro.library.autoreload=false
 velocimacro.library=VM_global_library.vm,dotCMS_library.vm,dotCMS_library_ext.vm
+# When true (default), engine init throws VelocityException if any velocimacro.library
+# file fails to load — prevents silent macro-rendering failures on pod restart (issue #35601).
+# Set to false only if you intentionally configure optional/missing library files.
+velocimacro.library.fail-on-missing=true
```
wezell: This is a feature flag and should use `Config.getBooleanProperty("XXXX")``

dotCMS/src/main/java/org/apache/velocity/runtime/VelocimacroFactory.java
```diff
@@ -221,6 +233,19 @@ else if (libfiles instanceof String)
                          vmManager.setRegisterFromLib(false);
                      }
                  }
+
+                 Logger.info(this,
+                         "Velocimacro libraries loaded: " + loadedLibraries
+                                 + " ; failed: " + failedLibraries);
+
+                 if (failOnMissing && !failedLibraries.isEmpty())
```
wezell, line 246: This is a breaking change - who knows how many dotCMS customers would fail this and their sites stop rendering when we updated them? Before we change default behavior, we need to make sure we understand what we're doing, if it could result in breakage and and gate it properly.  This needs to default to off.

wezell (changes requested):

dotCMS/src/main/resources/system.properties
dsolistorres: Done in 0e8d6888 — replaced `rsvc.getBoolean(...)` with `Config.getBooleanProperty("VELOCITY_LIBRARY_FAIL_ON_MISSING", false)`. Removed the corresponding `velocimacro.library.fail-on-missing` entry from `system.properties` and the now-unused `RuntimeConstants.VM_LIBRARY_FAIL_ON_MISSING` constant. Flag is now read via the standard dotCMS Config infrastructure (env var `DOT_VELOCITY_LIBRARY_FAIL_ON_MISSING` or system property).

dotCMS/src/main/java/org/apache/velocity/runtime/VelocimacroFactory.java
dsolistorres: Done in 0e8d6888 — default flipped to `false`. Existing customers see no behavior change (legacy silent-warn preserved). Operators opt in with `VELOCITY_LIBRARY_FAIL_ON_MISSING=true` once they have verified their library files load cleanly. The companion VelocityHealthCheck PR (#35771) still surfaces the broken state via `/readyz` either way, so K8s still gates traffic.

wezell (approved):
Better, thanks Daniel!

PR #35771, opened 2026-05-20
feat(health): add VelocityHealthCheck for K8s readiness probe (#35602)

### LINKED ISSUE #35602: Add VelocityHealthCheck for Kubernetes readiness probe (follow-up to spike #35329)

dsolistorres:
## Summary

Closes #35602 (follow-up implementation from spike #35329, companion to #35601 / #35768).

Adds a readiness-only health check that verifies the Velocity engine's global macro library is registered. When the library failed to load (see spike #35329), `engine.init()` can return successfully while every public page renders `#renderMarks` / `#editContentlet` as literal text. This check surfaces that state through `/readyz` so K8s can stop routing traffic to the affected pod.

Defense-in-depth alongside the fail-loud flag in #35601: that flag handles the common case (init throws → existing `ServletContainerHealthCheck` gates traffic via `dotcms.started.up`). This check covers the opt-out case (`velocimacro.library.fail-on-missing=false`) and any future post-init macro-registry corruption.

## Changes

- **`VelocityHealthCheck`** *(new)* — `com.dotcms.health.checks.cdi.VelocityHealthCheck` extends `HealthCheckBase`. Evaluates the probe `#renderMarks($null)` through `VelocityUtil.getEngine()` and returns DOWN when the rendered output contains the literal `#renderMarks(` marker.
- **`CoreHealthCheckProvider`** — register the new check alongside the existing CDI checks (database, cache, elasticsearch).
- **`VelocityHealthCheckTest`** *(new)* — three unit tests covering the UP path, the literal-rendering DOWN path, and a propagated-exception DOWN path.

## Design notes

- `isLivenessCheck()` returns **false** — a one-time init failure should remove the pod from the load balancer, not trigger a restart loop.
- `isReadinessCheck()` returns **true**.
- `getOrder()` returns 110 — runs after database (default 100), cache (30), and elasticsearch (40).
- Default mode is PRODUCTION (no monitor-mode downgrade) — when this is DOWN the system genuinely cannot render correctly.
- Skips during shutdown.
- The check is small and cheap — 0ms in practice on a healthy engine.

## Test plan

- [x] `VelocityHealthCheckTest` — 3/3 pass:
  - UP when probe renders macro body output
  - DOWN when probe renders the literal directive (macro not registered)
  - DOWN when `VelocityUtil.getEngine()` throws
- [x] **Local smoke test on a real dotCMS startup**: `just dev-run`, then `GET /dotmgt/health`:
  ```json
  {
    "name": "velocity",
    "status": "UP",
    "durationMs": 0,
    "message": "Velocity macro library registered (#renderMarks resolved) [0ms]",
    "monitorModeApplied": false
  }
  ```
  Confirmed alongside the other 10 health checks (database, cache, elasticsearch, etc.).

This PR fixes: #35602

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/health/checks/cdi/VelocityHealthCheck.java
+++ b/dotCMS/src/main/java/com/dotcms/health/checks/cdi/VelocityHealthCheck.java
@@ -84,7 +84,8 @@ public class VelocityHealthCheck extends HealthCheckBase {
             engine.evaluate(new VelocityContext(), writer, PROBE_LOG_TAG, PROBE_TEMPLATE);
             final String rendered = writer.toString();
             if (rendered.contains(LITERAL_MARKER)) {
-                throw new Exception("Velocity global macro library not registered: "
+                throw new IllegalStateException(
+                        "Velocity global macro library not registered: "
                         + "probe template rendered as literal text. See issues #35329 / #35601.");
             }
             return "Velocity macro library registered (#renderMarks resolved)";
--- a/dotCMS/src/main/java/com/dotcms/health/checks/cdi/VelocityHealthCheck.java
+++ b/dotCMS/src/main/java/com/dotcms/health/checks/cdi/VelocityHealthCheck.java
@@ -1,0 +1,93 @@
+package com.dotcms.health.checks.cdi;
+
+import com.dotcms.health.util.HealthCheckBase;
+import com.dotcms.rendering.velocity.util.VelocityUtil;
+import com.dotmarketing.util.Logger;
+
+import org.apache.velocity.VelocityContext;
+import org.apache.velocity.app.VelocityEngine;
+
+import javax.enterprise.context.ApplicationScoped;
+import java.io.StringWriter;
+
+/**
+ * CDI-based health check that verifies the Velocity engine's global macro
+ * library is loaded and resolvable.
+ *
+ * <p>Background: {@code VelocimacroFactory} has historically been able to silently
+ * swallow a {@code ResourceNotFoundException} at engine init, leaving
+ * {@code #renderMarks} and {@code #editContentlet} unregistered while
+ * {@code engine.init()} still returns successfully. In that state every public
+ * page renders the macro source as literal text. See spike #35329 and the
+ * fail-loud companion fix (#35601).
+ *
+ * <p>The probe evaluates {@value #PROBE_TEMPLATE} through
+ * {@link VelocityUtil#getEngine()}. When the macro is registered the rendered
+ * output is whatever the macro body produces; when it is not, the engine
+ * renders the directive source verbatim, which we detect via the literal
+ * {@value #LITERAL_MARKER} marker.
+ *
+ * <p>Excluded from liveness probes: a missing macro library is a one-time
+ * startup concern that should remove the pod from the load balancer, not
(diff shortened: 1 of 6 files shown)


dsilvam (approved):

dsolistorres:
Heads up — the intermittent Postman failures in `Template_Resource` and `AI` collections (with `language_id = -1` FK violation) are unrelated to this PR. Tracked separately in #35780, which includes the root-cause analysis and three suggested fixes (preferred: fail-fast 400 in `WorkflowResource` instead of silent `-1L` default).

`Integration Tests - Junit5 Suite 1` passed in the same run (it needs dotCMS up and serving requests), so startup isn't impacted by the new `VelocityHealthCheck`. A re-run should clear the matrix.

PR #35776, opened 2026-05-20
feat(analytics): HMAC bearer auth 

### LINKED ISSUE #35775: Wire HMAC bearer auth + JIT tenant provisioning save flow into dotCMS analytics proxy
## Description

Harden the dotCMS-side analytics proxy and credential flow so the system can talk to the dot-ca-event-manager microservice with per-tenant HMAC bearer tokens instead of plain Basic auth, and so backend authorization is consistent with the rest of dotCMS.

Parent epic: #35048 — Phase 3 (Security) work, with some Phase 4 (Reliability) overlap (proxy/health timeouts, structured failure modes).

### What changes

**Auth model**
- Replace Basic auth on the proxy with bearer tokens; remove `DOT_ANALYTICS_PASSWORD` entirely.
- New `ContentAnalyticsAppListener`: when an admin saves the Content Analytics App with `adminPassword` populated, the listener exchanges the password for a bearer token via `POST {DOT_ANALYTICS_BASE_URL}/v1/admin/token`, persists the returned token as a hidden secret on the app config, and atomically clears the user-entered password. The password is never persisted.
- App YAML grows new fields (`adminPassword`, `bearerToken`) with hints explaining the save-flow exchange.

**Proxy hardening**
- Path allowlist: only `event/*` may be proxied. Traversal segments, encoded `?`/`#`, and non-`event` prefixes are rejected with 400.
- Hard timeouts: 10s proxy / 2s health probe — prevents a slow upstream from exhausting the Jersey thread pool.
- Blank-body / `null` JSON guard on the ingest POST.
- `project=` query param deduped: caller-supplied wins over `DOT_ANALYTICS_PROJECT`.
- RFC-6750-compliant: when no token is available, the `Authorization` header is omitted entirely instead of emitting a malformed `Bearer ` (no token).
- Empty-body POSTs drop `Content-Type` (per RFC 9110 §8.3).

**Per-site permission gating on the GET catch-all**
- Site is resolved as the logged-in user via `HostAPI.find(siteId, user, …)` and `HostWebAPI.getCurrentHost(request, user)`, so `PermissionAPI` gates access.
- A backend user without READ on the requested site gets `403 SITE_ACCESS_DENIED` even when crafting `?siteId=` by hand to bypass the (already-permissioned) site picker.
- `SecurityLogger.logInfo` is called on the 403 path to mirror what the global `DotForbiddenExceptionMapper` would have logged — keeps denied-site accesses in the audit trail.

**Health monitor**
- Drops `DOT_ANALYTICS_PASSWORD` from the configuration check.
- Null-safe site probe (monitor pings landing on `System Host` no longer NPE).
- Tighter `/v1/health` timeout so monitor probes fail fast.

**Tests**
- `ContentAnalyticsAppListenerTest`: failure paths (missing tenant / missing base URL clear the password and notify the user) plus a direct happy-path test of `persistTokenAndClearCredentials` that asserts the atomic invariant — `adminPassword` absent, `bearerToken` replaced with the new value, unrelated secrets (`siteAuth`, etc.) preserved verbatim.
- `EventAnalyticsProxyHelperTest`: path allowlist (accept/reject matrix including traversal, smuggled `?`/`#`, non-`event` prefixes), `buildUpstreamUrl` (trailing-slash strip, project dedup, query forwarding), and `buildAuthHeader` covering the four security-critical branches (per-site wins, env fallback with trim, null host, both absent → empty Optional).

### Out of scope on this PR

These are intentionally deferred and tracked in `.scratch/reviews/feat-analytics-hmac-auth.md`:
- `EventAnalyticsProxyResourceTest` — resource-level test for the auth-decision branches (admin bypass, READ-permitted 200, READ-denied 403, session fallback). Auth model was in flux during the branch; folded into a follow-up once it settled.
- Promoting `DOT_ANALYTICS_TENANT` to a per-site App field — multi-tenant-per-JVM is a real product question, not a refactor.
- Standardizing on `siteId` vs `host_id` query param across analytics endpoints — cross-cutting product decision.
- `wrapUpstreamResponse` branch test coverage — correctness branches, not security; lower priority than `buildAuthHeader`.

## Acceptance Criteria

- [x] HMAC bearer auth wired into the proxy; `DOT_ANALYTICS_PASSWORD` no longer read anywhere
- [x] Save flow on Content Analytics App exchanges admin password for bearer token, clears password atomically
- [x] Proxy path allowlist + hard timeouts + blank-body guard + project= dedup
- [x] GET catch-all enforces per-site READ via `PermissionAPI`; non-permitted requests return 403 `SITE_ACCESS_DENIED`
- [x] 403 path logs to `SecurityLogger` for audit-trail parity with the global exception mapper
- [x] Health monitor no longer requires `DOT_ANALYTICS_PASSWORD`; null-safe site probe
- [x] Tests: listener failure paths + persist-token invariant; helper allowlist, URL building, `buildAuthHeader` security branches
- [x] Generated `openapi.yaml` reflects the new endpoint shapes (rebuild on merge via swagger-maven-plugin)

## Priority

Medium

## Additional Context

- Branch: `feat/analytics-hmac-auth` (~16 commits, ~1016 LOC across 10 files)

swicken:
Closes #35775
Sub-issue of #35048 (Phase 3 — Security)

## Summary

Replaces Basic auth on the dotCMS-side analytics proxy with per-tenant HMAC bearer tokens, adds a save-flow that exchanges admin credentials for a token (clearing the password from the app config after the exchange), tightens the proxy/health surface, and enforces per-site permissions on the GET read endpoint.

### Auth model
- Drop `DOT_ANALYTICS_PASSWORD` entirely. The proxy now uses bearer tokens sourced from the per-site Content Analytics App secret, with `DOT_ANALYTICS_BEARER_TOKEN` as a bootstrap fallback for non-site-bound contexts.
- New `ContentAnalyticsAppListener`: when an admin saves the Content Analytics App with `adminPassword` populated, the listener calls `POST {DOT_ANALYTICS_BASE_URL}/v1/admin/token` (Basic auth: `tenant:password`), stores the returned `bearerToken` as a hidden secret on the app config, and clears the `adminPassword` field. The user-entered password is written to encrypted app-secret storage by the original save and then removed automatically once the token exchange completes; on a failed exchange (or a failed token-persist) the listener still attempts to clear the password as a best-effort cleanup.
- The save flow re-fires `AppSecretSavedEvent`; the listener short-circuits via the "admin password not set" guard.
- App YAML grows new fields (`adminPassword`, `bearerToken`) with operator hints explaining the one-time exchange.

### Proxy hardening (`EventAnalyticsProxyHelper`)
- Path allowlist: only `event` / `event/...` are forwarded. Path segments exactly equal to `..` or `.`, encoded `?`/`#`, backslashes, and non-`event` prefixes are rejected with 400.
- Hard timeouts: 10s proxy / 2s health probe — prevents a slow event manager from exhausting the Jersey thread pool.
- Caller `project=` wins over `DOT_ANALYTICS_PROJECT`; the configured value is no longer appended when the URL already has one.
- RFC-6750-compliant `buildAuthHeader` returns `Optional.empty()` when no token is available; the proxy then omits `Authorization` instead of emitting a malformed `Bearer ` (no token).
- POST ingest: blank-body / literal-`null` JSON guard returns 400 with a clear error envelope. Empty-body POSTs drop `Content-Type` per RFC 9110 §8.3.

### Per-site permission gating (`EventAnalyticsProxyResource.proxyGetRequest`)
- Site is resolved as the logged-in user via `HostAPI.find(siteId, user, DONT_RESPECT_FRONT_END_ROLES)` for the explicit `?siteId=` branch and `HostWebAPI.getCurrentHost(request, user)` for the session fallback, so `PermissionAPI.READ` gates access.
- A backend user without READ on the resolved site gets `403 SITE_ACCESS_DENIED`, even if they hand-craft `?siteId=` to bypass the (already-permissioned) site picker.
- The 403 path calls `SecurityLogger.logInfo(...)` to mirror what the global `DotForbiddenExceptionMapper` would log — keeps denied-site accesses visible in the audit trail.

### Health monitor (`MonitorHelper.isContentAnalytics`)
- Drops the `DOT_ANALYTICS_PASSWORD` requirement.
- Null-safe site probe: monitor pings that land on the System Host (no current host in session) no longer NPE inside `getAppSecrets`.
- Tightened `/v1/health` timeout via the same 2s ceiling so monitor probes fail fast.

### Tests
- `ContentAnalyticsAppListenerTest`: failure paths (missing tenant / missing base URL clear the password and notify the user) plus a direct happy-path test of `persistTokenAndClearCredentials` that asserts the atomic invariant — `adminPassword` absent, `bearerToken` replaced with the new value, unrelated secrets (`siteAuth`) preserved verbatim. Method relaxed to package-private specifically for this test.
- `EventAnalyticsProxyHelperTest`: path allowlist matrix (traversal, smuggled `?`/`#`, non-`event` prefixes), `buildUpstreamUrl` (trailing-slash strip, project dedup, query forwarding), and `buildAuthHeader` covering the four security-critical branches (per-site wins, env fallback with trim, null host → fallback, both absent → empty Optional — the RFC 6750 branch).

## Files changed

- `dotCMS/src/main/java/com/dotcms/analytics/listener/ContentAnalyticsAppListener.java` *(new)*
- `dotCMS/src/main/java/com/dotcms/rest/api/v1/analytics/event/EventAnalyticsProxyHelper.java`
- `dotCMS/src/main/java/com/dotcms/rest/api/v1/analytics/event/EventAnalyticsProxyResource.java`
- `dotCMS/src/main/java/com/dotcms/rest/api/v1/analytics/content/util/ContentAnalyticsUtil.java`
- `dotCMS/src/main/java/com/dotcms/rest/api/v1/system/monitor/MonitorHelper.java`
- `dotCMS/src/main/java/com/dotcms/system/event/local/business/LocalSystemEventSubscribersInitializer.java`
- `dotCMS/src/main/resources/apps/dotContentAnalytics-config.yml`
- `dotCMS/src/test/java/com/dotcms/analytics/listener/ContentAnalyticsAppListenerTest.java` *(new)*
- `dotCMS/src/test/java/com/dotcms/rest/api/v1/analytics/event/EventAnalyticsProxyHelperTest.java` *(new)*
- `justfile` *(import optional `justfile.local`)*

## Operator notes

- `DOT_ANALYTICS_PASSWORD` is no longer read. Remove it from secrets/Terraform/k8s configs after merge.
- `DOT_ANALYTICS_TENANT` is still required. Tenant must already exist on the upstream event manager; this PR exchanges credentials, it does not provision the tenant.
- `DOT_ANALYTICS_BEARER_TOKEN` (optional) is a bootstrap fallback for non-site-bound contexts; production deployments should use the per-site App save flow instead.

## Behavioral changes worth flagging to consumers

- GET catch-all now returns `403 SITE_ACCESS_DENIED` for backend users without READ on the resolved site. The dashboard already passes `?siteId=` and the global site picker is already permissioned, so this only affects callers that hand-craft a `siteId`.
- Empty-body POSTs to `/api/v1/analytics/content/event` now return `400` with `"Request body is required"`. Postman tests that asserted on status alone still pass; tests that asserted on message text may need updates.
- Health check semantics: `NOT_CONFIGURED` requires either global env vars OR per-site App secrets, no longer both.
- Proxy hard timeout is 10s; previously inherited the `CircuitBreakerUrl` default.

## Test plan

- [ ] `./mvnw test -pl :dotcms-core -Dtest=ContentAnalyticsAppListenerTest,EventAnalyticsProxyHelperTest`
- [ ] Regenerate `openapi.yaml`: `./mvnw compile -pl :dotcms-core -DskipTests` — verify the GET catch-all gains 400/403 entries and the POST endpoint description matches reality
- [ ] Manual: save the Content Analytics App on a test site with a valid `adminPassword`; confirm `bearerToken` is populated and `adminPassword` is cleared in a single round-trip
- [ ] Manual: save the App with a wrong password; confirm the password is still cleared (no silent retention) and a system message surfaces to the user
- [ ] Manual: as a non-admin backend user with READ on Site A only, hit `/api/v1/analytics/event/total-events?siteId=<Site B id>` and confirm 403 `SITE_ACCESS_DENIED`
- [ ] Manual: confirm `/api/v1/analytics/event/...` against an unconfigured deployment fails fast (timeout) rather than hanging


dotCMS/src/main/java/com/dotcms/rest/api/v1/analytics/event/EventAnalyticsProxyHelper.java
```diff
@@ -44,16 +48,26 @@ public class EventAnalyticsProxyHelper {
...
     public static final String DOT_ANALYTICS_PROJECT = "DOT_ANALYTICS_PROJECT";
+    public static final String DOT_ANALYTICS_BEARER_TOKEN = "DOT_ANALYTICS_BEARER_TOKEN";
+
+    /** Hard upstream-call timeout. Prevents a slow event manager from exhausting the
+     *  Jersey thread pool and degrading the rest of the admin portal. */
+    static final int PROXY_TIMEOUT_MS = 10_000;
+
+    /** Relative paths allowed by the catch-all proxy. Anything outside this list — including
+     *  attempts to escape via {@code ..} segments — is rejected with 400. */
+    private static final List<String> ALLOWED_PATH_PREFIXES = List.of("event");
```
freddyDOTCMS: really we support more than just event url path in the event manager

dotCMS/src/main/java/com/dotcms/rest/api/v1/analytics/event/EventAnalyticsProxyResource.java
```diff
@@ -235,13 +270,59 @@ public Response proxyGetRequest(
...
+                                    "User does not have access to the requested site"))))
+                    .build();
+        } catch (DotDataException e) {
+            Logger.warn(this, "Failed to resolve site for analytics proxy: " + e.getMessage());
+            return Response.status(Response.Status.BAD_REQUEST)
+                    .entity(new ResponseEntityView<>(
+                            List.of(new ErrorEntity("INVALID_SITE_ID",
+                                    "Could not resolve a site for the request"))))
+                    .build();
+        }
```
freddyDOTCMS: I think it should go inside a local method

dotCMS/src/main/java/com/dotcms/rest/api/v1/analytics/event/EventAnalyticsProxyHelper.java
swicken: Yes sorry, meant to expand the list, this is done. 

`List.of("event", "conversion", "session", "health");`

Am I missing any @freddyDOTCMS ? I want to keep the allowlist otherwise the proxy means any authenticated backend user can tunnel through the auth boundary.

dotCMS/src/main/java/com/dotcms/rest/api/v1/analytics/event/EventAnalyticsProxyResource.java
swicken: Good call! This is done.

justfile
```diff
@@ -1,4 +1,5 @@
 set positional-arguments := true
+import? 'justfile.local'
```
jcastro-dotcms, line 2: Is this change intended?

swicken: Yes, I have a local justfile with things to ease my personal dev process and its an optional import so it won't impact others. Just committing it to main now so I don't have to re-add it on every branch. There are some things I plan to push into the main justfile but I am fleshing them out locally.

jcastro-dotcms (approved):

freddyDOTCMS (approved):

PR #35847, opened 2026-05-27
fix(postman): set languageId=1 on PagesResourceTests content bodies (#35780)

dsolistorres:
## Summary

Corrects the approach from #35838. After that merged, `PR Test / Postman Tests - Page` **still failed** (run [26517996164](https://github.com/dotCMS/core/actions/runs/26517996164), test `Live page cache invalidation / Create Test Rich Text Contentlet 1`).

## Why #35838's `_makedefault` prerequest didn't work

Two independent reasons, both confirmed in the logs:

1. **It 401'd.** `PUT /api/v2/languages/1/_makedefault → 401 Unauthorized` (`Authentication credentials are required`). This collection runs with a pre-populated JWT, so the `user` / `password` env vars are unset — the Basic-auth call had no real credentials (server logged `Email 'undefined' is not valid`).

2. **Even with valid auth, it cannot help these tests.** The failing tests use `POST /api/v1/workflow/actions/default/fire/...`. `WorkflowResource.fireActionDefaultSinglePart` defaults a missing `languageId` to a **hardcoded `-1L`** (`WorkflowResource.java:3476`) — it does *not* call `getDefaultLanguage()`. So changing the default language has zero effect on workflow-fire content creation.

`_makedefault` only helps endpoints that resolve via `getDefaultLanguage()` (page lookups, AI search, site create). For workflow-fire content creation, the **body must carry `languageId`**.

## Fix

- Add `"languageId": 1` to all **15** content-creating bodies in the collection that omitted it (the same fix #35818 applied to `Create Test Page`).
- Remove the dead `_makedefault` prerequest block added in #35838 (it 401s and does nothing useful here).

## Patched tests

```
Live page cache invalidation / Create Test Rich Text Contentlet 1, 2, 3
Render / Check related content in depth 2 / Create Content (+ relationship)
TestHostVariables / Create Page using the template
In-Place Content Copy / Create Test Page (+ Rich Text Contentlet)
Number of Content References in Pages / Create First/Second Test Page (+ Contentlet)
Get Page Types / Create new UrlMap Page
Permissions / Create Test Page Permission
Check Languages for a Page / Create Test Page in English
Put Create Page With CacheTTL 100
```

## Note on durability

[...]

PR #35818, opened 2026-05-22
fix(postman): hardcode language_id=1 on page-related Postman tests (#35780)

dsolistorres:
## Summary

Same `language=-1` class of failure as #35780, surfacing in `Experiments_Resource.postman_collection.json` this time. CI run [26312707306](https://github.com/dotCMS/core/actions/runs/26312707306/job/77472990761?pr=35771) hit:

```
expected response to have status code 200 but got 404
inside "Variants / Keep Content after change Layout to a Specific Variant / Add contetlet into DEFAULT variant"
```

## Server-side diagnosis

```
ERROR page.PageResource - HTMLPageAssetNotFoundException on PageResource.addContent,
      pageId: f41ac5d4f5a2e854e1d79ea6ca39df27
```

`PageResourceHelper.getPage()` looks the page up via:

```java
final Language currentLanguage = WebAPILocator.getLanguageWebAPI().getLanguage(request);
...
APILocator.getVersionableAPI().getContentletVersionInfo(pageId,
        renderContext.getCurrentLanguageId(), ...)
```

`getLanguage(request)` falls back to `getDefaultLanguage()` when no language is on the request — and in this test env that returns the `LANG__404` sentinel (`id=-1`). The page was created in the previous step with `languageId=1` (verified in the test bodies), so the `(id, -1)` lookup misses and the server returns 404.

## Fix

Add `language_id=1` to the 5 URLs that target page-content endpoints without specifying language. All affected pages are created earlier in the flow with `languageId=1` (English), so hardcoding `1` matches the test data exactly — no need for the `_getdefault` resolver dance here.

Patched URLs:

| Section | Test |
|---|---|
| Variants / Keep Content after change Layout to a Specific Variant | `Add contetlet into DEFAULT variant` |
| Variants / Keep Content after change Layout to a Specific Variant | `Add Contentlet into a specific variant` *(already had `variantName`, now also `language_id`)* |
| Variants / Add Contentlet on Default Variant after A Specific Variant exists | `Add contetlet into DEFAULT variant` |
| Variants / Add Contentlet on Default Variant after A Specific Variant exists | `Add Another Contentlet into a DEFAULTvariant` |
| Add Spanish Contentlet inside Variant | `Add contetlet into DEFAULT variant` |

## Diff

[...]

PR #35816, opened 2026-05-22
fix(postman): pass explicit language in AI Search Related by identifier (#35780)

dsolistorres:
## Summary

Refs #35780. Follow-up to PR #35795 which fixed the same `language=-1` class of failure in `Workflow_Resource_Tests` and `GraphQLTests`. AI was intentionally out of scope in that PR; this surfaces in the latest CI run as the only remaining Postman-side blocker.

## What was diagnosed

CI run [26312707306](https://github.com/dotCMS/core/actions/runs/26312707306/job/77467087126?pr=35771) — `PR Test / Postman Tests - AI` failed on `Search / Search Related by identifier`:

```
expected response to have status code 200 but got 404
```

Server-side log during the failing request:

```
WARN  rest.SearchResource - unable to find matching contentlet for id:8d37011d-... inode: language:-1
```

Same `language:-1` symptom as the rest of #35780. The test body is:

```json
{"identifier": "{{identifier}}", "fieldVar": "seo"}
```

…with no `language` field. The endpoint at `dotCMS/src/main/java/com/dotcms/ai/rest/SearchResource.java`:

```java
final long language = json.optLong(AiKeys.LANGUAGE,
        APILocator.getLanguageAPI().getDefaultLanguage().getId());
```

…falls back to `getDefaultLanguage().getId()` — which returns `-1` (LANG__404 sentinel) in this test env. Then `findContentletByIdentifier(id, -1)` returns 404.

The CLI `_makedefault` fix in #35813 affects the CLI test dotCMS instance only — Postman jobs run their own separate Docker containers, so the test env state is unchanged here.

## Changes

`AI.postman_collection.json`:

1. **Top-level prerequest** — adds the same `GET /api/v2/languages/_getdefault` resolver pattern shipped in #35795 v2. Idempotent via `pm.environment.get('defaultLanguageId')` guard, throws on network error / non-200 / missing `entity.id` (no silent fallback to invalid JSON request bodies).

2. **Failing test body** — adds `"language": {{defaultLanguageId}}`. Note the key is `"language"` (per `AiKeys.LANGUAGE`), not `"languageId"`.

## Diff

`+32 / −2` — surgical text patch that preserves the file's 2-space formatting (no full reserialize).

## Test plan

- [x] JSON validity verified.
- [x] Pattern identical to #35795 v2 (already approved in review).
- [ ] CI verifies — `PR Test / Postman Tests - AI` should pass on this branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

Issue #35780, opened 2026-05-20
Flaky Postman tests: WorkflowResource silently defaults missing languageId to -1L causing FK violation


dsolistorres:
## Problem Statement

Multiple Postman test collections (`Template_Resource`, `AI`) intermittently fail in PR check runs with the same server-side signature:

```
ERROR: insert or update on table "contentlet" violates foreign key constraint "fk_contentlet_lang"
Detail: Key (language_id)=(-1) is not present in table "language".
```

Affected tests observed in run [26167537035](https://github.com/dotCMS/core/actions/runs/26167537035) (PR #35771):

- `Template_Resource.postman_collection` → `Not anonymous layout / PUT Create Page Using NOT anonymous Template` — HTTP 500 from `PUT /api/v1/workflow/actions/default/fire/NEW`
- `AI.postman_collection` → `Search / Search Related by identifier` — HTTP 404 from `POST /api/v1/ai/search/related`

**Impact**: blocks PR merges; affects multiple unrelated PRs over the last week; not customer-facing.

## Root Cause

The endpoint silently defaults a missing `languageId` to `-1L` and then attempts a DB INSERT with it.

`dotCMS/src/main/java/com/dotcms/rest/api/v1/workflow/WorkflowResource.java:3476`:

```java
final long languageId = ConversionUtils.toLong(contentMap.get("languageId"), -1l);
```

When the Postman request body omits `languageId` (or its pre-request script's lookup of the default language returns nothing — e.g., cold cache), the request progresses with `languageId = -1L`. The call chain:

```
WorkflowResource.fireActionDefaultSinglePart → fireAction
  → WorkflowAPIImpl.fireContentWorkflow
  → SaveContentAsDraftActionlet.executeAction
  → ESContentletAPIImpl.checkin → internalCheckin
  → ESContentFactoryImpl.upsertContentlet
  → PostgreUpsertCommand.execute
  → Postgres rejects with fk_contentlet_lang violation
```

A second site of the same pattern exists in `com.liferay.portal.language.LanguageUtil.internalGetLanguageId` (also returns `-1` on missing input). Both sites have existed for years — the _intermittency_ is what changed, likely from recent timing/load shifts (large OpenSearch refactor work over the past two weeks). The defaulting itself is the structural defect.

## Steps to Reproduce

1. Open any PR that runs the `PR Test / Postman Tests - Template` job
2. Observe intermittent failure on `Not anonymous layout / PUT Create Page Using NOT anonymous Template`
[...]

PR #35849, opened 2026-05-27
fix(language): never surface LANG_404 sentinel from getDefaultLanguage (#35780)

dsolistorres:
## Summary

The central, server-side fix for the #35780 class of failures. `getDefaultLanguage()` could return the `LANG_404` sentinel (`id=-1`), which then propagated into every downstream consumer and produced the symptoms we've been patching test-by-test:

| Consumer | Symptom when default = -1 |
|---|---|
| Contentlet save (`MapToContentletPopulator`) | `fk_contentlet_lang` FK violation |
| Page lookup (`PageResourceHelper.getPage`) | `HTMLPageAssetNotFound` → 404 |
| Variant fallback (`VariantWebAPIImpl.getContentletVersionInfoByFallback`) | `ResourceNotFoundException "lang:-1"` |
| AI search / Site create | 404 / "Language cannot be null" |

## Root cause

`LanguageFactoryImpl.getDefaultLanguage()` resolves through an `AtomicReference` lock. A thread that **loses** `createDefaultLanguageLock.compareAndSet(null, …)` falls straight to `return createDefaultLanguageLock.get()` while the winner is still inside `transactionalGetDefaultLanguage()` — at that instant the lock still holds the `LANG_404` sentinel. A cleared/uninitialized cache surfaces it too.

## Fix

Guard the return value: when the resolved default is null or has a non-positive id, resolve synchronously via `transactionalGetDefaultLanguage()` (which reads the company default or creates English) and cache that — never hand back the sentinel. One place, covers all consumers.

```java
final Language resolved = createDefaultLanguageLock.get();
if (resolved == null || resolved.getId() <= 0) {
    final Language fresh = transactionalGetDefaultLanguage();
    if (fresh != null && fresh.getId() > 0) {
        languageCache.setDefaultLanguage(fresh);
        createDefaultLanguageLock.set(fresh);
        return fresh;
    }
    Logger.warn(this, "...");
}
return resolved;
```

## Test

`LanguageFactoryIntegrationTest#test_getDefaultLanguage_neverReturnsSentinel` reproduces the cache+lock sentinel precondition (via reflection on the lock) and asserts a positive id is always returned.

## Follow-up

Once this lands, the per-collection Postman/CLI workarounds (#35795, #35816, #35818, #35847, the SiteAPIIT warmup in #35813) become redundant and can be reverted — they were all compensating for this single root cause.

## Test plan

- [x] `./mvnw compile -pl :dotcms-core` → BUILD SUCCESS
- [x] Integration test compiles
- [ ] CI runs `LanguageFactoryIntegrationTest` + the previously-failing Postman suites (Page/Template/AI/Experiments) to confirm the central fix clears them

### DIFF
--- a/dotCMS/src/main/java/com/dotmarketing/portlets/languagesmanager/business/LanguageFactoryImpl.java
+++ b/dotCMS/src/main/java/com/dotmarketing/portlets/languagesmanager/business/LanguageFactoryImpl.java
@@ -215,7 +215,27 @@ public class LanguageFactoryImpl extends LanguageFactory {
 			createDefaultLanguageLock.set(defaultLanguage);
 		}
 
-		return createDefaultLanguageLock.get();
+		final Language resolved = createDefaultLanguageLock.get();
+
+		// Guard (issue #35780): the lock can transiently hold the LANG_404 sentinel
+		// (id=-1) while a concurrent caller is still resolving the default, and a
+		// cleared/uninitialized cache can surface the same sentinel. A non-positive
+		// language id must never reach callers: it causes fk_contentlet_lang FK
+		// violations on contentlet save and HTMLPageAssetNotFound (404) on page
+		// lookups. When that happens, resolve the default synchronously instead of
+		// handing back the sentinel.
+		if (resolved == null || resolved.getId() <= 0) {
+			final Language fresh = transactionalGetDefaultLanguage();
+			if (fresh != null && fresh.getId() > 0) {
+				languageCache.setDefaultLanguage(fresh);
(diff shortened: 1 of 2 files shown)


dsolistorres:
Closing as **superseded by #35771** (merged).

The central `getDefaultLanguage()` guard from this PR was folded into #35771 (commit `c2e0b9e944`) and merged there alongside the `VelocityHealthCheck` work and the Postman `language_id` hardening. This standalone PR is therefore redundant — no code is lost. See #35771 and issue #35780 (closed) for the full resolution.

PR #35860, opened 2026-05-28
feat(apps): env var / Kubernetes secret provisioning for app secrets (#35859)

### LINKED ISSUE #35859: Provide app secrets via environment variables / Kubernetes secrets (DOT_{APP_KEY}_{HOSTNAME}_{APP_VALUE_KEY})
## Parent

User Story / Enhancement

## Problem

dotCMS app secrets (the Apps framework, `com.dotcms.security.apps`) can today only be fully configured through the UI / stored secrets store. There is an existing env-var hook (`AppsUtil.guessEnvVar` → `APP_{APP_KEY}_PARAM_{APP_VALUE_KEY}`), but it:

- is **not host-aware** (one var overrides a param for every site),
- only **enriches an existing stored blob** — it cannot provision an app on a host that has no stored config,
- has **override** semantics that aren't expressed as a clear precedence model.

This blocks headless / IaC / Kubernetes deployments where secrets should be injected at the platform level.

## Proposal

Add per-host environment-variable / Kubernetes-secret provisioning for **all** apps using the pattern:

```
DOT_{APP_KEY}_{HOSTNAME}_{APP_VALUE_KEY}
```

plus a global System Host form `DOT_{APP_KEY}_SYSTEM_HOST_{APP_VALUE_KEY}`.

Resolution is **per-param, specificity-first** (highest → lowest):

1. host-specific env — `DOT_{APP_KEY}_{HOSTNAME}_{APP_VALUE_KEY}` (locks the UI field)
2. host-specific stored / UI value
3. System Host env — `DOT_{APP_KEY}_SYSTEM_HOST_{APP_VALUE_KEY}`
4. legacy global env — `APP_{APP_KEY}_PARAM_{APP_VALUE_KEY}` (deprecated, throttled warning)
5. System Host stored value

### Behavior

- **Full provisioning**: an app with no stored blob is usable purely from env vars (partial provisioning returns only env-backed params).
- **Normalization**: env var names use `Config.envKey()` only (uppercase, dots/dashes → underscores, `DOT_` prefix) — no parallel normalization; construct-and-lookup from `AppDescriptor.getParams()` (never parses arbitrary env vars). A registered descriptor is required.
- **Editability**: only tier-1 (host-specific env) locks the field read-only; global tiers stay editable so a deliberate per-host stored value still wins.
- **`fromEnv`**: exposed on the Java `Secret` object and the REST `SecretView` (boolean; no tier enum).
- **Listing/presence**: `listAppKeys`, `appKeysByHost`, `filterSitesForAppKey` report env-only apps as configured (lazy detection; System Host/legacy tiers mirror the stored System Host cascade).
- **`fallbackOnSystemHost=false`** suppresses both System Host tiers (3 and 5).
- **Masking** follows the descriptor param type regardless of source; env values are returned raw (no coercion).

### Out of scope

Live/hot secret rotation without restart (env vars are process-static), file-mounted-secret watching, type coercion, startup index, tier-enum REST field.

### Backward-compatibility note

The legacy `APP_..._PARAM_...` pattern keeps working as tier-4 (with a deprecation warning), but precedence becomes **specificity-first**: a host-specific stored value now wins over the legacy global env var (previously legacy env overrode stored). Existing integration tests asserting the old override semantics will need to be updated.

wezell:
## Summary

Adds per-host environment-variable / Kubernetes-secret provisioning for **all** dotCMS apps, resolving #35859.

App secret params are resolved per-param via a **specificity-first 5-tier** model:

1. host-specific env — `DOT_{APP_KEY}_{HOSTNAME}_{APP_VALUE_KEY}` (locks the UI field)
2. host-specific stored / UI value
3. System Host env — `DOT_{APP_KEY}_SYSTEM_HOST_{APP_VALUE_KEY}`
4. legacy global env — `APP_{APP_KEY}_PARAM_{APP_VALUE_KEY}` (deprecated, throttled warning)
5. System Host stored value

## Changes

- `AppsUtil`: `envVarName()` (`Config.envKey()` normalization), tier helpers `hostEnvSecret` (tier-1, locks), `systemHostEnvSecret` (tier-3), `legacyEnvSecret` (tier-4, throttled deprecation warning); `SYSTEM_HOST_ENV_SEGMENT`; overlay-free `readJson(chars, false)` variant.
- `AppsAPIImpl.getSecrets`: rewritten to `resolveSecrets()` — per-param tier walk, full provisioning with no stored blob, `fallbackOnSystemHost=false` suppresses System Host tiers (3 & 5).
- `AppsAPIImpl` listing/presence (`hasAnySecrets`, `listAppKeys`, `appKeysByHost`): lazily detect env-backed config so env-only apps appear configured (System Host/legacy tiers mirror the stored cascade).
- `AbstractProperty`/`Secret`: `fromEnv` signal.
- `SecretView` REST serializer: boolean `fromEnv` (derived from the Java signal; no tier enum).

## Behavior notes

- Construct-and-lookup from `AppDescriptor.getParams()` — never parses arbitrary env vars; a registered descriptor is required for env tiers.
- Only tier-1 locks the field (`hasEnvVarValue`); global tiers stay editable so a host-specific stored value wins.
- Env values returned raw (no coercion); SECRET masking follows the descriptor type regardless of source.
- Writes to env-locked params are accepted silently (enforced at read time).
- Out of scope: live hot-reload / file-mount watching, type coercion, startup index, tier-enum REST field.

## ⚠️ Backward-compatibility / review focus

Precedence is now **specificity-first**: a host-specific stored value beats the legacy global `APP_..._PARAM_...` env var (previously legacy env overrode stored). **Existing integration tests in `dotcms-integration` (`AppsAPIImplTest`, `AppsUtilTest`) that assert the old legacy-override semantics will likely need updating** — please validate under CI.

## Test plan

- [x] New unit tests (12) green: `AppsUtilEnvKeyTest`, `AppsUtilHostEnvSecretTest`, `AppsUtilGlobalEnvTest`, `AppsAPIImplHostEnvProvisioningTest`
- [x] `dotcms-core` module compiles
- [ ] CI: run `dotcms-integration` app-secrets suite; update any tests asserting legacy override
- [ ] Add integration coverage for tier-3 (System Host env cascade) and `fallbackOnSystemHost=false` suppression (require DB/APILocator infra)
- [ ] Manual: provision an app on a site purely via `DOT_{APP_KEY}_{HOSTNAME}_{APP_VALUE_KEY}`, verify configured badge + read-only field + `fromEnv` in REST

This PR fixes: #35859


wezell:
Thanks for the thorough review. Addressed in `cd737dfce4`:

**🔴 Fixed**
- **#1 env-value persistence leak** — `fromEnv` secrets are now filtered out in `deleteSecret`, `saveSecret`, and `maintainHiddenValues` (unchanged-mask case), so an env-backed value is never serialized into the stored blob. Added `AppsAPIImplEnvWriteGuardTest` (env-backed not promoted; normal stored value still retained).
- **#2 `Secret.destroy()` NPE** — null-guarded both arrays. Added `SecretDestroyTest` (tier-1 envValue-only + value-only).

**🟠 Fixed**
- **#3 perf** — `appKeysByHost`/`listAppKeys` now resolve the hostname once per site via a `hasEnvBackedSecretsForHostName` overload instead of `HostAPI.find` per app.
- **#4 collision** — `hostEnvSecret` now skips tier-1 (with a warning) when a site name normalizes to the reserved `SYSTEM_HOST` segment.

**🟠 Documented (per design decisions in the originating interview)**
- **#5 tier-4 gating** — intentionally left always-on (the precedence decision only suppresses tiers 3 & 5 under `fallbackOnSystemHost=false`); the #1 leak is fixed independently by the `fromEnv` write-filter, so tier-4 no longer feeds it. Added a clarifying comment.
- **#6 comma values** — env reads go through `Config.getStringProperty` by design (single normalization path). Added a characterization test documenting that simple comma values round-trip but whitespace-adjacent commas are normalized; called out as a known constraint of env provisioning.

**🟡 Nits** — moved `SYSTEM_HOST_ENV_SEGMENT`/throttle set to the constants block; `envSecret` now uses `isHidden()`.

**Deferred / follow-up**
- **#7 BOOL coercion guardrail** — out of scope per the spec (no coercion); a startup-time warn for non-coercible env-bound BOOL/SELECT params is a reasonable follow-up.
- **Integration suite** — the precedence flip (host-stored beats legacy global env) will require updating `dotcms-integration` `AppsAPIImplTest`/`AppsUtilTest`; planned as part of CI validation on this PR.

All app-secrets unit tests green (17).

wezell:
Thanks for the re-review — addressed in `e680751b`:

**🔴 Save-time mask leak for tier-3 hidden params** — Fixed. `maintainHiddenValues` now **never persists the literal `HIDDEN_SECRET_MASK`**, including when there is no existing value (removed the `existingAppSecrets.isEmpty()` early-return that bypassed the guard). So a tier-3-only hidden param submitted as the mask is dropped and stays env-resolved. Took the defensive option (also fixes the latent pre-existing mask-persist bug); left the public `getSecrets` tier-3 gating intact per the spec decision.

**🟠 Non-hidden env value frozen on re-submit** — Fixed for tier-1/tier-4: `maintainHiddenValues` now drops a non-hidden submitted value that exactly equals an env-backed `existing` value (unchanged form re-submit), so it isn't snapshotted; a genuinely changed value is still persisted (host-specific stored legitimately wins per specificity). Note: tier-3 non-hidden isn't visible in the `fallbackOnSystemHost=false` existing read, so an *unchanged* non-hidden tier-3 value re-saved on a child host can still snapshot — narrow edge (non-hidden + System Host env + untouched re-save); flagging as a follow-up rather than widening the existing-read to `true` (which would risk copying tier-5 System-Host-stored values into host blobs).

**🟠 Unthrottled collision warning** — Fixed. Added a `ConcurrentHashMap.newKeySet()` throttle (`SYSTEM_HOST_COLLISION_LOGGED`), once per JVM per site.

**🟠 `appKeysByHost` redundant global lookups** — Fixed. Split into `hasHostEnvBackedSecrets` (tier-1, per-site) and `hasGlobalEnvBackedSecrets` (tier-3/4, host-independent). Global-env apps are now computed once and applied to every site; only tier-1 runs in the per-site loop. `listAppKeys` unchanged (single site).

**🟡 `getHidden()` → `isHidden()`** — Done (now null-safe in `maintainHiddenValues`).

**🟡 Acknowledged / documented (no code change):**
- Global env makes every site appear configured — intended (mirrors the stored System Host cascade, per the design decision); will add a release note.
- `Config.envKey` hostname normalization collisions — operator-docs note; once provisioning via env, hostnames must stay unique post-normalization.
- `fromEnv` excluded from `equals/hashCode` — left as-is (no current map-key/set-diff usage).
- `envSecret` not copying `paramDescriptor.getEnvVar()/getEnvShow()` — the REST view still surfaces the descriptor's `envVar` via `buildParam`; leaving as-is.

**Integration tests** — I ran the `dotcms-integration` app-secrets suite locally. `AppsUtilTest` 32/32 green (the override tests needed `SOME_APP_ENV_VAR`/`DOT_APP_TEST_APP_PARAM_TEST_PARAM` in the runtime). Found + fixed two real bugs there too (`e722504`): a null-key NPE in `resolveSecrets` and an immutable-keySet `removeAll` in `deleteSecret`. The precedence-flip assertions did **not** need changes. `Test_AppKeyByHost_On_Index_Deactivation` errored only in its `finally` (null index from `loadIndicies()` in an isolated single-class run) — its body/assertion passed.

Unit tests: 20 green.

erickgonzalez (approved):

jcastro-dotcms (approved):

PR #35863, opened 2026-05-28
fix(os-index): replace os:: prefix with .os suffix as DB uniqueness artifact

### LINKED ISSUE #35820: [DEFECT] Confusing empty OS index appears in indices portlet on Phase 1 migration bootstrap

fabrizzio-dotCMS:
## Summary

Closes #35820

### Why this change is safe

This code is still behind the OpenSearch feature flag and **has never been rolled out to production**. The OS index layer is entirely in active development, which means there is room for this kind of intentional renaming experiment to get the naming strategy right before any production data is involved.

### `.os` suffix — why it matters for correctness

OpenSearch indices are registered under the same logical names as their ES counterparts (e.g. \`working_20260528\`). When APIs return a \`Map<String, ...>\` or a \`Set<String>\` keyed by index name, ES and OS entries would silently collide — the OS entry would overwrite the ES one and data would be lost. Appending a \`.os\` suffix to every OS index name guarantees that both providers can coexist in the same map/set without any information loss.

**Collision-free guarantee:** logical index names always end in \`_YYYYMMDDHHMMSS\` (numeric timestamp) — they can never naturally end in \`.os\`.

Changes:
- **`IndexTag.OS`** now appends `.os` (suffix) instead of prepending `os::` (prefix)
- `isTagged()` checks `endsWith(".os")`; `untag()` strips the last 3 chars
- `IndexTag` gains a `suffix` field alongside the existing `prefix` field — all existing callers unchanged
- `VersionedIndicesAPIImpl` comments updated to reflect the new storage format
- `OPENSEARCH_MIGRATION.md` — "tag ownership rule" section rewritten with the suffix rationale

> **No DB migration task:** the OS index layer has never shipped to production, so there are no `os::name` rows to convert. Any dev/QA instance that ran an earlier build of this feature must drop its stale `indicies` OS rows (or let bootstrap recreate the OS indices) — there is no automatic conversion.

### Improved bootstrap sync strategy

`initOSCatchup()` in `ContentletIndexAPIImpl` now follows a three-case priority order when OS indices are absent from the cluster:

1. **OS DB record exists, cluster index missing** — recreate the OS index using the name already registered in `VersionedIndicesAPI`. This is the authoritative source in Phase 3.
2. **OS DB empty, ES DB has a record** (migration catch-up, Phases 1/2) — mirror the ES logical name so both indices share the same base name. This prevents timestamp divergence — e.g. the ES index `working_20240101` produces `working_20240101.os`, not a freshly-stamped `working_20260528.os`.
3. **Both empty** (fresh install or data-loss) — generate a new timestamp and emit a warning for the operator.

### Tests — `OSIndexAPIImplIntegrationTest`

Two stub tests that only asserted "returns non-null" were upgraded to real integration assertions:

- **`test_getClusterHealth_shouldReturnHealthForCreatedIndex`** — creates a live index, then verifies the health map contains an entry for it with a non-null/non-empty status and positive shard count.
- **`test_getIndicesStats_shouldReturnStatsForCreatedIndices`** — creates live and working indices, then verifies both appear in the stats map with non-negative document count, non-negative raw size, and a non-empty human-readable size string.

Three new phase-routing tests were also added to `ContentletIndexAPIImplMigrationIT` to verify that `getIndicesStats` and `getClusterHealth` correctly merge ES + OS results in Phase 1 and return ES-only results in Phase 0.

## Test plan

- [ ] Verify `IndexTag.OS.tag("cluster_abc.working_20260528")` returns `"cluster_abc.working_20260528.os"`
- [ ] Verify `IndexTag.strip("cluster_abc.working_20260528.os")` returns `"cluster_abc.working_20260528"`
- [ ] Verify `IndexTag.resolve("cluster_abc.working_20260528.os") == IndexTag.OS`
- [ ] Verify `IndexTag.resolve("cluster_abc.working_20260528") == IndexTag.ES` (untagged default)
- [ ] On a fresh instance: confirm Phase 1 bootstrap stores `name.os` in DB and portlet shows correct indices
- [ ] Confirm `getIndicesStats` / `getClusterHealth` maps in Phase 1 contain both ES and OS entries (no collision/overwrite)
- [ ] No regression on ES portlet behavior
- [ ] `OSIndexAPIImplIntegrationTest` — `test_getClusterHealth_shouldReturnHealthForCreatedIndex` and `test_getIndicesStats_shouldReturnStatsForCreatedIndices` pass against a live OS cluster


dotcms-integration/src/test/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImplMigrationIntegrationTest.java
```diff
@@ -106,6 +108,20 @@ public class ContentletIndexAPIImplMigrationIT extends IntegrationTestBase {
     @Inject
     private OSIndexAPIImpl osIndexAPI;
 
+    // ── Used to resolve the OS physical name for indices created through the phase router ──
```
nollymar, line 111: This class and the new test don't follow the naming convention. They end with "IT" instead of "Test". Also, the new test class is not in a MainSuite

dotcms-integration/src/test/java/com/dotcms/OpenSearchUpgradeSuite.java
```diff
@@ -42,6 +43,7 @@
         OSClientProviderIntegrationTest.class,
         OSClientConfigTest.class,
         ContentletIndexAPIImplMigrationIT.class,
+        ContentletIndexAPIImplPhaseSwitchIT.class,
```
nollymar: I see what happened. Both tests are part of the new OS upgrade suite. In this case, only the renaming observation is valid

dotCMS/src/main/java/com/dotcms/content/index/IndexTag.java
```diff
@@ -41,73 +70,95 @@
...
      * no exception is thrown. Use {@link #isTagged(String)} to guard first
      * when strict matching is required.</p>
      *
      * @param taggedName tagged or untagged name; {@code null} returns {@code null}
-     * @return the name without the vendor prefix
+     * @return the name without the vendor marker
      */
     public String untag(final String taggedName) {
-        if (taggedName != null && taggedName.startsWith(prefix)) {
+        if (taggedName == null) {
```
nollymar, line 139: is this logic tested

dotCMS/src/main/java/com/dotcms/content/index/IndexTag.java
```diff
@@ -41,73 +70,95 @@
...
 
     /**
-     * Returns {@code true} when {@code name} carries this vendor's prefix.
+     * Returns {@code true} when {@code name} carries this vendor's marker.
      *
      * @param name any string; {@code null} returns {@code false}
      */
     public boolean isTagged(final String name) {
-        return name != null && name.startsWith(prefix);
+        if (name == null) {
```
nollymar, line 157: is there a test for this?

dotCMS/src/test/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImplPhaseTest.java
```diff
@@ -495,6 +495,12 @@ public String getNameWithClusterIDPrefix(final String name) {
             return name.startsWith(CLUSTER_PREFIX) ? name : CLUSTER_PREFIX + name;
         }
 
+        @Override
+        public String removeClusterIdFromName(final String name) {
```
nollymar: why a new removeClusterIdFromName method is required here? why isn't it reusing the one in the IndexAPI?

dotCMS/src/main/java/com/dotcms/content/index/IndexTag.java
fabrizzio-dotCMS: Added `IndexTagTest` (unit, `dotCMS/src/test/.../content/index/IndexTagTest.java`) covering `untag` — round-trip vs `tag`, no-op on untagged input, ES no-op (never strips another vendor’s marker), and null-safety. (8a8ada8)

dotCMS/src/main/java/com/dotcms/content/index/IndexTag.java
fabrizzio-dotCMS: Yes — same `IndexTagTest` pins `isTagged`: detects only its own marker, false on legacy/untagged names, ES is always false (empty marker), and null-safe. (8a8ada8)

dotCMS/src/test/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImplPhaseTest.java
fabrizzio-dotCMS: Good catch — this exposed real duplication. The ES and OS `removeClusterIdFromName` impls were byte-identical and differed only by the cluster prefix, which is a JVM-wide value. I consolidated the whole prefix concern into `IndexAPI` as `default` methods over a single `getClusterPrefix()`: `removeClusterIdFromName`, `getNameWithClusterIDPrefix` and `hasClusterPrefix` now derive from it and are inherited. `ESIndexAPI`/`OSIndexAPIImpl` dropped their `clusterPrefix` field + the three methods, and the router dropped its delegating overrides. The fake (and the two ITs) now override only `getClusterPrefix()` to supply a deterministic/dotted prefix — no re-implemented logic. A deterministic value is still needed in the pure unit test because the real `getClusterPrefix()` reads the cluster id from the DB. New `IndexAPIPrefixTest` exercises the real defaults. Bonus: restoring the defaults also closes the M-4 rollback note. (8a8ada8)

dotcms-integration/src/test/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImplMigrationIntegrationTest.java
fabrizzio-dotCMS: Renamed to `ContentletIndexAPIImplMigrationIntegrationTest` (and the sibling to `ContentletIndexAPIImplPhaseSwitchIntegrationTest`) to match the suite convention. Re: MainSuite — as you noted below, both are already wired into `OpenSearchUpgradeSuite` (`@RunWith(MainBaseSuite.class)`), so only the rename was needed. (8a8ada8)

dotcms-integration/src/test/java/com/dotcms/OpenSearchUpgradeSuite.java
fabrizzio-dotCMS: Both classes renamed to `*IntegrationTest` and `OpenSearchUpgradeSuite` updated (import + `@SuiteClasses`). No `*IT` classes remain. (8a8ada8)

nollymar (approved):

PR #35866, opened 2026-05-28
Issue 35861 tag field not cleared

### LINKED ISSUE #35861: Bug: Tag field values are not cleared when all tags are removed from a contentlet
## Problem Statement

When a contentlet has tags and the user removes all tags (saves with an empty tags field), the tags are **not cleared** from the API response. This affects:

- Single-language contentlets: clearing all tags on an existing contentlet does not wipe the displayed tag value.

**Possible Root cause:** `Contentlet.setTags()` (`Contentlet.java` ~line 1572) only writes tag values to the contentlet's property map when entries exist in the `tag_inode` table. When `tag_inode` is empty for that inode, the method is a no-op — leaving any previously-copied tag string intact in the map.

## Steps to Reproduce

**Via API (minimal reproduction):**

1. Create a Content Type with a Text field and a Tag field (e.g., variable `tags`).
2. Create a contentlet of that type, set a tag value (e.g., `"new edit content"`), and save/publish.
3. Fire a workflow save action with `tags: ""` in the payload:
   ```
   PUT /api/v1/workflow/actions/{actionId}/fire?indexPolicy=WAIT_FOR&inode={inode}
   { "contentlet": { ..., "tags": "", "languageId": 1, ... } }
   ```
4. Observe the response — `"tags"` still shows the old value.

**Via UI:**

1. Open any contentlet that has tags set.
2. Remove all tags from the Tag field.
3. Save.
4. Re-open the contentlet — the tags reappear.

## Acceptance Criteria

- [ ] Saving a contentlet with an empty Tag field clears all tags — `tag_inode` has no rows for that inode and the API response returns an empty/absent `tags` field.
- [ ] Additional QA: confirm that all other field types (Text, TextArea, Date, DateTime, Time, Binary, HostFolder, etc.) can also be fully cleared and the cleared value is correctly persisted and returned.

## dotCMS Version

Latest `main` branch.

## Severity

## Links

NA

erickgonzalez:
This PR fixes: #35861

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ESContentletAPIImpl.java
+++ b/dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ESContentletAPIImpl.java
@@ -5807,6 +5807,10 @@ public class ESContentletAPIImpl implements ContentletAPI {
             contentlet.setIndexPolicyDependencies(indexPolicyDependencies);
 
             contentlet = relateTags(contentlet, tagsValues, tagsHost);
+            // tag_inode is now authoritative for this contentlet; clear the loadedTags guard so any
+            // later setTags() (e.g. the response/display transform) re-reads the reconciled state
+            // instead of a snapshot taken before checkin (issue #35861, bulk workflow path).
+            contentlet.resetLoadedTags();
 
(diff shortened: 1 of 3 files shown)


erickgonzalez:
@claude generate a PR description

PR #35729, opened 2026-05-16
create new endpoint to retrieve workflow schemas by multiple Content Types

dario-daza:
### Proposed Changes

* Added `GET /api/v1/workflow/contenttypes/schemes` endpoint (`findAllSchemesByContentTypeList`) that accepts a comma-separated `contentTypeIds` query parameter and returns workflow schemes grouped by content type
* Created `AbstractContentTypeWorkflowSchemesView` (Immutables) with `contentTypeId` + `contentTypeSchemes` fields, and `ResponseEntityContentTypeWorkflowSchemesView` as the typed response wrapper
* The endpoint accepts `contentTypeIds` comma-separated values (`?contentTypeIds=id1,id2,id3`) rather than repeated params — trims whitespace and filters empty tokens
* Added integration tests in `WorkflowResourceResponseCodeIntegrationTest` covering: null input, empty string, whitespace-only tokens, IDs with surrounding whitespace, invalid ID (expects 404), happy path with comma-separated valid IDs, and documented repeated-param limitation
* Fix and add postman test to test the Workflow Schemas

### Additional Info

The endpoint returns a structured `List<{contentTypeId, contentTypeSchemes[]}>` rather than a flat deduplicated list - this keeps the association between each content type and its schemes intact and lets the frontend handle deduplication if needed.

This PR fixes: #35471

This PR fixes: #35471

PR #35871, opened 2026-05-29
fix(workflow) #35471: QA findings on batch content-type schemes endpoint

### LINKED ISSUE #35471: [TASK] Content Drive: endpoint to return workflow schemes for a list of content types
## Problem Statement

The chip-filter UI for "workflow" in content drive needs to know which workflow schemes are valid options given the user's currently-selected content types. The existing endpoint ([`GET /api/v1/workflow/schemes/schemescontenttypes/{contentTypeId}`](dotCMS/src/main/java/com/dotcms/rest/api/v1/workflow/WorkflowResource.java)) only accepts a single content type ID. Calling it N times client-side is slow and creates ordering/race conditions in the UI.

## Acceptance Criteria

- New endpoint that accepts a list of content type identifiers (variable names or IDs) and returns the **union** of workflow schemes assigned to those content types.
- Response shape: `{ schemes: WorkflowSchemeView[] }` (a content type → schemes breakdown is not needed for the current UI; can be added later if required).
- Schemes are deduplicated by scheme ID before being returned.
- Empty input list returns `{ schemes: [] }` (200 OK), not 400.
- Unknown content type IDs are silently ignored (do not 404 the whole request).
- Auth/permissions: same model as the existing single-content-type endpoint.

## Implementation Pointers

- Reuse `WorkflowAPI.findSchemesForContentType(ContentType)` per content type and combine results.
- Deduplicate by scheme ID before returning.
- The existing single-content-type endpoint stays for backward compatibility.

## Discussion Point — Endpoint Shape

The recommended shape is a **new** `POST /api/v1/workflow/schemes/schemescontenttypes` with a JSON body: `{ "contentTypes": ["Blog", "News"] }`. This matches the dotCMS multi-value convention used in `DriveRequestForm` and avoids touching the existing path-param endpoint.

Alternative: extend the existing endpoint to also accept a CSV path parameter or query param. Cheaper to ship but mixes two URL contracts on one path.

**The implementer should confirm with the team before choosing.** The default recommendation is the new POST.

## Tests

- Unit: `WorkflowResourceTest` covering empty input, single content type, multiple content types with overlapping schemes, multiple content types with disjoint schemes, unknown content type ID mixed with valid ones.
- Integration: `dotcms-integration` covering permission denial.
- Postman: new entry under the workflow collection.

## Out of Scope

- Filtering by archived schemes, languages, or workflow actions.
- Returning workflow steps (steps come from a separate API; this endpoint only returns schemes).
- Per-content-type breakdown of which scheme came from which type.

ihoffmann-dot:
## Summary

QA follow-up on [#35729](https://github.com/dotCMS/core/pull/35729). Addresses findings F-2, F-4, F-5, and F-6 from review of the `GET /api/v1/workflow/contenttypes/schemes` endpoint. F-1, F-3, and F-7 are excluded pending product alignment or confirmed as environment-side issues.

- **F-4** — `contentTypeId` in the response now always contains the real UUID, regardless of whether the caller passed a UUID or a variable name
- **F-2** — Inputs are now resolved to `ContentType` objects before deduplication; a UUID and a variable name referencing the same content type produce a single entry instead of two
- **F-5** — Added `contentTypeVariable` field to the response so callers can correlate results without an extra lookup
- **F-6** — Renamed `contentTypeSchemes` → `schemes` in the response shape, matching the contract defined in the original acceptance criteria

## Changes

- `AbstractContentTypeWorkflowSchemesView`: added `contentTypeVariable()`, renamed `contentTypeSchemes()` → `schemes()`
- `WorkflowResource.findAllSchemesByContentTypeList`: replaced raw-string stream with a resolve-then-dedup loop using `ContentTypeAPI.find()` + `LinkedHashMap<UUID, ContentType>`
- `openapi.yaml`: regenerated
- Integration test and Postman collections updated to reference `schemes` instead of `contentTypeSchemes`

## Test plan

- [ ] Existing integration tests in `WorkflowResourceResponseCodeIntegrationTest` still pass
- [ ] Call endpoint with variable name — `contentTypeId` in response is the UUID, `contentTypeVariable` is the variable name
- [ ] Call with both UUID and variable name of the same content type — single entry returned
- [ ] Call with an unknown ID mixed with valid ones — unknown is skipped, valid entries present with correct UUID/variable

Closes #35471

This PR fixes: #35729


dotCMS/src/main/java/com/dotcms/rest/api/v1/workflow/WorkflowResource.java
```diff
@@ -615,33 +617,49 @@ public final Response findAllSchemesByContentTypeList(
...
+            final List<ContentTypeWorkflowSchemesView> result = new ArrayList<>();
+            for (final ContentType ct : resolvedTypes.values()) {
+                try {
+                    result.add(ContentTypeWorkflowSchemesView.builder()
+                            .contentTypeId(ct.id())
+                            .contentTypeVariable(ct.variable())
+                            .schemes(this.workflowAPI.findSchemesForContentType(ct))
+                            .build());
+                } catch (DotDataException e) {
+                    skippedIds.add(ct.id());
```
fabrizzio-dotCMS: Add a log warn

dotCMS/src/main/java/com/dotcms/rest/api/v1/workflow/WorkflowResource.java
```diff
@@ -615,33 +617,49 @@ public final Response findAllSchemesByContentTypeList(
...
-                        } catch (DotWorkflowException e) {
-                            if (ExceptionUtil.causedBy(e, NotFoundInDbException.class)) {
-                                skippedIds.add(contentTypeId);
-                                return null;
-                            }
-                            throw e; // permission errors and unexpected causes propagate
-                        }
-                    })
-                    .filter(Objects::nonNull)
+            for (final String token : ids) {
```
fabrizzio-dotCMS, line 632: Verify this  pattern in other resources, and see if they need to get sanitized

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/workflow/WorkflowResourceResponseCodeIntegrationTest.java
```diff
@@ -451,49 +451,41 @@ public void Find_Actions_By_Scheme_Invalid_SchemeId()  {
...
                 request, new EmptyHttpResponse(), null);
-        assertEquals(Status.OK.getStatusCode(), response.getStatus());
-        @SuppressWarnings("unchecked")
-        final List<?> result = (List<?>) ResponseEntityView.class.cast(response.getEntity()).getEntity();
-        assertTrue("Null contentTypeIds should return an empty list", result.isEmpty());
+        assertEquals(Status.BAD_REQUEST.getStatusCode(), response.getStatus());
     }
 
     @Test
-    public void Find_Schemes_By_Content_Type_List_Empty_List() {
```
fabrizzio-dotCMS, line 465: doc these tests please

fabrizzio-dotCMS (commented):
Just a few minor things to check before approving

dotCMS/src/main/java/com/dotcms/rest/api/v1/workflow/WorkflowResource.java
ihoffmann-dot: Sure, `ContentTypeResource` for instance passes strings directly to `contentTypeAPI.find()` without sanitization. We are applying a trim and filter(UtilMethods::isSet) before the loop in this case

dotCMS/src/main/java/com/dotcms/rest/api/v1/workflow/WorkflowResource.java
ihoffmann-dot: Done!

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/workflow/WorkflowResourceResponseCodeIntegrationTest.java
ihoffmann-dot: Done!

fabrizzio-dotCMS (approved):

PR #35886, opened 2026-06-01
fix(block-editor): apply custom contentlet renders inside grid blocks

### LINKED ISSUE #35884: Custom render (VTL) not applied to contentlets inside a Block Editor grid
## Problem Statement

When a contentlet that has a **custom render (VTL)** is placed **inside a Block Editor grid (columns)**, the backend does **not** apply the custom render. Instead, only the default rendering (the contentlet's *title*) is output on the page.

The exact same contentlet renders correctly — with its custom VTL applied — when it is placed **outside** the grid in the Block Editor. This indicates the custom-render resolution is skipped specifically for contentlets nested inside a grid/column block.

- **Impact:** Pages using custom renders for contentlets inside Block Editor grids show only the contentlet title instead of the intended custom markup, breaking the page layout/content.
- **Environment type:** Traditional (not headless) — VTL rendering.
- **Custom render in use:** `dotContent-FaqList.vtl`

### Root cause summary (from initial investigation)

- By default, only the **title** of a contentlet inside the Block Editor is rendered on the page.
- This can be overridden with a custom render — one is configured here: `dotContent-FaqList.vtl`.
- The backend correctly applies the custom render for contentlets placed directly in the Block Editor, but **fails to apply it when the contentlet is nested inside a grid**.

## Steps to Reproduce

1. In a traditional (VTL) instance, create a custom contentlet render (e.g., `dotContent-FaqList.vtl`).
2. Edit a page that uses a Block Editor field.
3. Add the contentlet **directly** in the Block Editor (outside any grid) → custom render is applied correctly. ✅
4. Add the **same** contentlet **inside a grid (column)** in the Block Editor.
5. View/render the page.
6. **Observed:** Only the contentlet title is rendered for the contentlet inside the grid — the custom render is ignored. ❌

**Reference page:** https://skeleton-economy.proserve-fis-sandbox.dotcms.cloud/faq-block-editor

## Acceptance Criteria

- [ ] A contentlet with a custom render placed **inside** a Block Editor grid applies the custom render (`dotContent-{ContentType}.vtl`) identically to one placed outside the grid.
- [ ] Rendering behavior for contentlets **outside** a grid is unchanged (no regression).
- [ ] Custom render resolution works for contentlets nested at any supported grid/column depth.
- [ ] When no custom render exists, the default title rendering is still used (both inside and outside the grid).
- [ ] Verified on a traditional (VTL) instance.

## dotCMS Version

Cloud — `skeleton-economy.proserve-fis-sandbox.dotcms.cloud` (ProServ FIS sandbox). Traditional rendering. Confirm whether the issue also reproduces on the latest `main`.

## Severity

## Links

- [Slack thread (#proserv / FIS)](https://dotcms.slack.com/archives/C08MDK1BYLF/p1780082692925109)
- Block Editor custom rendering docs: https://dev.dotcms.com/docs/block-editor#RenderingContentCustom
- NA (no Freshdesk ticket)

rjvelazco:
## Proposed Changes

Fixes #35884

Contentlets (and other content blocks) that have a custom render template were rendered with the generic **default** template when nested inside a Block Editor **gridBlock**, instead of their custom render. The same block renders correctly when placed directly in the Block Editor (outside a grid).

### Root cause

There are two parallel render paths and only one was custom-render-aware:

| | Top-level block | Block inside a grid |
|---|---|---|
| Path | Java: `StoryBlockMap` → `GenericRenderableImpl.toHtml(baseTemplatePath)` | Velocity: `gridBlock.vtl` → `#renderContentBlock` macro |
| Custom VTL lookup | ✅ `existsFile` resolves `dotContent-{Type}.vtl` / `{type}.vtl` | ❌ macro hardcoded the generic `{type}.vtl` |
| Output | Custom render | Default (e.g. title only) |

The grid recursion happens purely in Velocity (`render.vtl` `#renderContentBlock`) and never invoked the Java-side custom-template lookup. Additionally, `GenericRenderableImpl.toHtml(baseTemplatePath)` dropped the `baseTemplatePath` when it fell back to the no-arg `toHtml()` for container blocks like `gridBlock`.

### Fix

- **`GenericRenderableImpl.toHtml(baseTemplatePath)`** now resolves the render context once and exposes a `StoryBlockRenderHelper` (carrying the `baseTemplatePath`) in the Velocity context — in **both** the custom-template and default-template branches — so nested blocks keep access to the custom-render resolution.
- **`StoryBlockRenderHelper`** (new) renders a single nested node back through the same `GenericRenderableImpl` path, so the custom-template lookup runs for grid-nested blocks too.
- **`render.vtl`** routes the leaf content branches — **`dotContent`, `dotImage`, and `dotVideo`** — through the helper when present, falling back to the generic template otherwise (**backward compatible** for any caller rendering without a base path). Structural branches (paragraph, heading, lists, table…) already recurse via the macro, and `gridBlock` is itself a recursing container, so no change is needed there.

## How to test

1. On a traditional instance, create a custom render (`dotContent-{ContentType}.vtl`, or a type-level `dotImage.vtl` / `dotVideo.vtl`) and reference it via a base path.
2. In a Block Editor, add the corresponding block **inside a grid column**.
3. Render the page → the custom template is applied (previously only the default showed).
4. Confirm a block **outside** a grid still renders correctly (no regression).

### Automated tests

- `test_overridden_render_inside_grid_block_to_html` — `dotContent` nested in a grid uses its `dotContent-{ContentType}.vtl` custom render.
- `test_overridden_dotimage_render_inside_grid_block_to_html` — type-level `dotImage.vtl` override applies to a `dotImage` nested in a grid (also guards the identical `dotVideo` branch).

## Checklist

- [x] Tests added (integration)
- [x] `dotcms-core` compiles; `dotcms-integration` test-compiles
- [x] Reviewer to run the new integration tests in CI

## Videos

https://github.com/user-attachments/assets/b645d476-3708-44ea-b986-d11666c39b93

https://github.com/user-attachments/assets/a10e0aa1-7a05-467f-9eb0-19ebd9f71d08

This PR fixes: #35884


rjvelazco:
@claude Review these changes and propose a clean architecture if neede

rjvelazco:
Thanks for the thorough architecture review — really useful. Addressed in `fe1cd22`. Here's the disposition of each point:

### Done (the stated pre-merge minimum + cheap wins)

| # | Smell | Fix |
|---|---|---|
| 2 | `processType` duplicated in `StoryBlockMap` and `StoryBlockRenderHelper` | Extracted to `NodeTypes.typeKey(JSONObject)` — single source of truth; both callers delegate. |
| 3 | Shared-context `$item` clobbering | `GenericRenderableImpl.toHtml(baseTemplatePath)` now snapshots `$item` and restores it in a `finally`, so nested rendering can't leak the last node's value back to the caller. |
| 4/6 | New helper + factory per recursion level | Helper is created once per render tree (create-if-absent in the context); `StoryBlockRenderHelper#renderableFactory` is now `static final`. |
| 8 | Inverted ternary NPE at the old `GenericRenderableImpl.java:63` (`null == host ? host.getHostname() : …`) | Fixed to `null != host ? host.getHostname() : "dotcms.com"`. It was in a method I was already touching, so corrected here rather than filed. |
| 9 | No `dotVideo` test | Added `test_overridden_dotvideo_render_inside_grid_block_to_html`, exercising the `dotVideo` macro branch independently of `dotImage`. |

### Deferred (you flagged these as out of scope for this PR)

- **The `StoryBlockRenderer` / `RenderContext` / `NodeTypes` / `TemplateResolver` redesign (#1).** Agreed this is the right destination — unify the two dispatch tables and collapse `toHtml()` / `toHtml(baseTemplatePath)` into one context-driven path. It's a larger structural change than a customer-facing bugfix should carry, so I'd rather land it separately.
- **`resolveHost` per node (#5).** Sharing one helper trims the allocations, but `existsFile` still resolves the host per node. Resolving it once belongs with the `RenderContext` work above, since that's where the per-tree state would live.
- **`baseTemplatePath` overload (#4 in smells list).** Same — naturally resolved by `RenderContext.baseTemplatePath: Optional<String>` in the refactor.

I can open a follow-up tech-debt issue capturing the `RenderContext` design (with your sketch) and link it here — want me to?

ihoffmann-dot (approved):

dotCMS/src/main/java/com/dotcms/rendering/velocity/viewtools/content/util/GenericRenderableImpl.java
```diff
@@ -151,25 +151,50 @@ public String toHtml(final String baseTemplatePath) {
...
-                Context context = externalContext;
-                if (externalContext == null) {
+            Context context = externalContext;
+            if (externalContext == null) {
 
-                    final HttpServletRequest requestProxy   = new FakeHttpRequest(host.getHostname(), null).request();
-                    final HttpServletResponse responseProxy = new BaseResponse().response();
-                    context = VelocityUtil.getInstance().getContext(requestProxy, responseProxy);
-                }
+                final HttpServletRequest requestProxy   = new FakeHttpRequest(host.getHostname(), null).request();
```
wezell: Is there a reason we are using a FakeRequest here or should we use the one stored in the threadlocal?  E.g.
```
HttpServletRequest request = HttpServletRequestThreadLocal.getRequest() != null 
? HttpServletRequestThreadLocal.getRequest()
: new FakeHttpRequest(host.getHostname(), null).request();
```

dotCMS/src/main/java/com/dotcms/rendering/velocity/viewtools/content/util/GenericRenderableImpl.java
```diff
@@ -60,7 +60,7 @@ public String toHtml() {
 
             final Host   host     = Try.of(()->APILocator.getHostAPI().findDefaultHost(
                     APILocator.systemUser(), false)).getOrNull();
-            final String hostname = null == host? host.getHostname():"dotcms.com"; // fake host
+            final String hostname = null != host ? host.getHostname() : "dotcms.com"; // fallback to a fake host
```
wezell: Same thing here - why the Default host?  I think you need real host resolution here, no?  Does this work in a multi-host environment?

Use the `HostWebAPI.getCurrentHost()` to get the host of a real request.  It will fall back to the default host if not found.

wezell (dismissed):
See comments.  Not 100% sure they are valid but should be explored.

dotCMS/src/main/java/com/dotcms/rendering/velocity/viewtools/content/util/GenericRenderableImpl.java
rjvelazco: Good call — done in `b24c852cf`. Both `toHtml()` and `toHtml(baseTemplatePath)` now go through a shared `resolveRenderContext(...)` that prefers the real request from `HttpServletRequestThreadLocal.INSTANCE.getRequest()` (and `HttpServletResponseThreadLocal` for the response), matching the existing `ActionletUtil` idiom. The `FakeHttpRequest` is only built as a fallback when nothing is bound to the thread (programmatic/background rendering). Note this whole branch only runs when no external context was supplied — on a real page render `externalContext` is non-null and is used as-is.

dotCMS/src/main/java/com/dotcms/rendering/velocity/viewtools/content/util/GenericRenderableImpl.java
rjvelazco: Agreed the default host is wrong for multi-host. Rather than extract a host name via `HostWebAPI.getCurrentHost()` and still fabricate a request, the fix now prefers the **real** thread-local request, so the true current host (plus session/attributes) flows into the render context naturally. The default-host lookup is now only the fallback when no request is bound to the thread — and it's behind a `Supplier`, so it's skipped entirely whenever a real request exists. When there genuinely is no request, the default host is the only sensible value (`getCurrentHost(null)` would resolve to it anyway). The path-derived host used by `existsFile()` for the custom-template lookup is unchanged.

rjvelazco:
@wezell thanks — both were valid. Addressed in `b24c852cfe`:

- Replaced the unconditional `FakeHttpRequest` + default-host construction with a shared `resolveRenderContext(...)` that **prefers the real request/response from the thread-local** (`HttpServletRequestThreadLocal` / `HttpServletResponseThreadLocal`), so the true current host is used in multi-host environments. The mock request is now a fallback only, used when no request is bound to the thread, and the default-host lookup is deferred behind a `Supplier` so it's skipped when a real request exists.
- This path only executes when no external Velocity context was passed in; normal page renders pass a non-null context and are unaffected.
- The path-derived host used for the custom-template file lookup (`existsFile`) is intentionally unchanged.

Replied inline on both threads with the details. Re-requesting review when you get a chance.

dotCMS/src/main/java/com/dotcms/rendering/velocity/viewtools/content/util/GenericRenderableImpl.java
```diff
@@ -86,6 +78,44 @@ private boolean existsInFileSystem(final String filePath) {
...
+    private Context resolveRenderContext(final Supplier<String> fallbackHostNameSupplier) {
+
+        if (null != externalContext) {
+            return externalContext;
+        }
+
+        final HttpServletRequest request = null != HttpServletRequestThreadLocal.INSTANCE.getRequest()
+                ? HttpServletRequestThreadLocal.INSTANCE.getRequest()
+                : new FakeHttpRequest(fallbackHostNameSupplier.get(), null).request();
+        final HttpServletResponse response = null != HttpServletResponseThreadLocal.INSTANCE.getResponse()
```
wezell, line 103: Can we ask Claude whether or not it’s safe to use the real responses here? Are we capturing (or trying to capture) the string output from the response, or is writing to the responsewriter okay here?  It is probably ok, but just to be sure.

wezell (changes requested):

PR #35904, opened 2026-06-02
feat(velocity-playground): Angular portlet migration (#35810)

### LINKED ISSUE #35810: [Task] Velocity Playground: Angular implementation + backend wiring
## Goal

Ship the Angular Velocity Playground portlet end-to-end — UI, `portlet.xml` registration, and rollback toggle — in a single PR.

## Frontend scope

New Nx library: `core-web/libs/portlets/dot-velocity-playground/`

- `dot-velocity-playground-shell.component.ts`
- `dot-velocity-playground-page.component.ts`
- `lib.routes.ts` exported as `dotVelocityPlaygroundRoutes`
- SignalStore: `DotVelocityPlaygroundStore`
- Data-access service: `core-web/libs/data-access/src/lib/dot-velocity-playground/dot-velocity-playground.service.ts` calling `POST /api/vtl/dynamic/`
- PrimeNG: `p-splitter`, Monaco editor (Velocity / VTL language mode), output panel that renders JSON, XML, and plain text outputs distinctly
- History affordance equivalent to the legacy dropdown (scope and storage location finalized by the spike — default: last 10 entries in `localStorage`)
- Tests: Jest + Spectator for shell / page / store with `data-testid` selectors on all interactive elements

## Backend wiring

`dotCMS/src/main/webapp/WEB-INF/portlet.xml`:

- Replace the JSP-backed `velocity_playground` portlet with the new Angular shell.
- Add a `velocity_playground-legacy` entry pointing at the existing JSP as the rollback escape hatch.
- Add the corresponding `PortletID` enum entry and i18n key (or reuse the existing `velocity_playground` key).
- Live toggle so admins can revert to the JSP without redeploy.

## Inputs

- Spike output from the consolidated audit issue (parity matrix + API audit + UI sketch).
- Portlet conventions: `core-web/libs/portlets/CLAUDE.md`.

## Definition of Done

Every Gherkin scenario in epic #34737's Definition of Done passes manually against the local build:

- Admin runs valid Velocity code
- Output renders JSON, XML, and plain text
- Velocity error surfaces inline
- History recall
- Resizable split-pane
- Non-admin role is blocked
- Unlicensed instance blocks the portlet
- Rollback to legacy JSP without redeploy

hmoreras:
### Proposed Changes

Closes #35810. Ships the Velocity Playground as an Angular portlet with parity for the legacy JSP and a rollback toggle.

https://github.com/user-attachments/assets/2bf549dd-5a1a-41a6-bd7d-a9315b31bdde

**Frontend — new Nx library `libs/portlets/dot-velocity-playground/`**

* Shell + page components, lazy route registered in `app.routes.ts` and gated by `MenuGuardService` (URL slug `velocity-playground` mapped to backend portlet id `velocity_playground`).
* `DotVelocityPlaygroundStore` (NgRx SignalStore) — owns `code`, `output`, `outputContentType`, `elapsedMs`, `history`, `wrapCode`, `splitterRatio`. History capped at 10 entries, persisted in `localStorage`.
* `DotVelocityPlaygroundService` — `POST /api/vtl/dynamic/` with `responseType: 'text'`; maps `content-type` header to `json | xml | plaintext`.
* Monaco editor with a custom Velocity language + dark theme registered before the editor mounts (no flash to default light theme).
* Layout aligned with `es-search`: resizable `p-splitter`, Run button always enabled, share/export actions for the output (cURL + fetch snippets, file download), help popover with example snippets.
* Loading uses `dot-spinner` (no skeleton bars). Empty state via `DotEmptyContainerComponent`.
* Stats bar header shows the response content type (left-aligned, font-semibold, non-uppercase) and elapsed run time.
* JSON responses are pretty-printed with 2-space indentation in both the editor view and the exported file; malformed payloads fall back to the raw body. XML / plaintext are passed through untouched.

**Backend wiring**

* `portlet.xml`: Angular shell registered under `velocity_playground`; legacy JSP kept as `velocity_playground-legacy` for rollback without redeploy.
* `PortletID.java` updated for the legacy entry.
* `Language.properties` keys for the new portlet labels.

### Checklist
- [x] Tests — Jest + Spectator for shell, page, store, service (`yarn nx test portlets-dot-velocity-playground-portlet` → 55 passed)
- [x] Translations — new keys in `Language.properties`
- [x] Security Implications Contemplated — endpoint already enforces admin role and license; `MenuGuardService` blocks the route for non-authorized users; rollback toggle leaves the legacy JSP available without redeploy

### Additional Info

History entries store the original (un-wrapped) Velocity code, not the timer-instrumented payload submitted to the backend.

To roll back to the legacy JSP at runtime, switch the `velocity_playground-legacy` portlet entry in `portlet.xml` — no redeploy needed.

### Screenshots
_legacy JSP playground_  |  _Angular portlet with split editor, dark Monaco theme, JSON pretty-print, stats bar_

This PR fixes: #35810


core-web/libs/portlets/dot-velocity-playground/src/lib/dot-velocity-playground-page/store/dot-velocity-playground.store.ts
```diff
@@ -0,0 +1,176 @@
...
+                                        outputContentType: response.contentType,
+                                        elapsedMs: response.elapsedMs,
+                                        history: nextHistory
+                                    });
+                                },
+                                error: (error: HttpErrorResponse) => {
+                                    patchState(store, {
+                                        status: ComponentStatus.LOADED,
+                                        errorMessage:
+                                            error?.error?.message ??
```
oidacra, line 148: With `responseType: 'text'`, Angular sets `HttpErrorResponse.error` to the raw response body as a **string**, so `error.error.message` is always `undefined` here. The `??` chain then falls through to `error.message` — Angular's generic `"Http failure response for /api/vtl/dynamic/: 500…"` — and the actual server-side VTL error text never reaches the inline error panel. (`httpErrorManager.handle` still surfaces it in a toast, so it's degraded rather than fully lost.)

Suggested fix — prefer the string body when present:
```ts
const serverMessage =
    (typeof error.error === 'string' && error.error.trim())
        ? error.error
        : error?.error?.message ?? error?.message ?? null;

patchState(store, {
    status: ComponentStatus.LOADED,
    errorMessage: serverMessage ?? 'velocityPlayground.error.unknown'
});
```

core-web/libs/portlets/dot-velocity-playground/src/lib/dot-velocity-playground-page/dot-velocity-playground-page.component.ts
```diff
@@ -0,0 +1,213 @@
...
+        PanelModule,
+        PopoverModule,
+        DotEmptyContainerComponent,
+        DotSpinnerComponent,
+        DotMessagePipe
+    ],
+    providers: [DotVelocityPlaygroundStore, DotClipboardUtil],
+    templateUrl: './dot-velocity-playground-page.component.html',
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    host: { class: 'flex flex-col h-full min-h-0 bg-white' }
```
oidacra: The host class uses `bg-white` here, and the help-popover description paragraph (`...component.html:246`) does too — but the rest of this template uses surface tokens (`bg-surface-0` on the run footer, `bg-surface-50` on the stats bar and help-example header). Since `bg-white` is a literal color, it won't follow the theme on a dark/alternate surface. Should these be `bg-surface-0` to stay consistent and theme-adaptive?

core-web/libs/portlets/dot-velocity-playground/src/lib/dot-velocity-playground-page/dot-velocity-playground-page.component.spec.ts
```diff
@@ -0,0 +1,383 @@
...
+
+            appendSpy.mockRestore();
+            URL.createObjectURL = createObjectUrlOriginal;
+        });
+
+        it('toggleExportMenu delegates to the menu when the viewChild resolves', () => {
+            loadedSetup();
+            const menu = spectator.component.$exportMenu();
+            if (!menu) {
+                // viewChild may not resolve in JSDOM with popup overlays — skip gracefully
```
oidacra: This test can pass without running any assertion: if `$exportMenu()` returns null (which happens under JSDOM when the `<p-menu popup>` viewChild doesn't resolve), it hits the early `return` before any `expect()`, so it reports green while covering nothing. Could we make it fail loudly instead — e.g. `expect(menu).toBeTruthy()` — or stub `$exportMenu` via `Object.defineProperty` (like the `useExample` test already does for `$helpPopover`) so the `toggle` delegation is actually asserted?

core-web/libs/portlets/dot-velocity-playground/src/lib/dot-velocity-playground-page/store/dot-velocity-playground.store.spec.ts
```diff
@@ -0,0 +1,235 @@
...
+        });
+
+        it('setSplitterRatio ignores invalid ratios', () => {
+            spectator.service.setSplitterRatio([NaN, 30] as [number, number]);
+            expect(spectator.service.splitterRatio()).toEqual([...DEFAULT_SPLITTER_RATIO]);
+        });
+    });
+
+    describe('history', () => {
+        it('selectHistoryEntry sets the code when the entry exists', () => {
```
oidacra, line 96: This covers the happy path (entry present → `code` is set). `selectHistoryEntry` also has a guard — `if (!store.history().includes(entry)) return;` — whose branch isn't exercised here. Could we add a sibling test that calls `selectHistoryEntry` with a value not in history and asserts `code` stays unchanged, so the early-return is covered?

oidacra (approved):

core-web/libs/portlets/dot-velocity-playground/src/lib/dot-velocity-playground-page/store/dot-velocity-playground.store.ts
hmoreras: Good catch — fixed in f0c2d65e65. The error handler now prefers `error.error` when it's a non-empty string (the raw VTL body) before falling back to `error.error.message` and `error.message`. Added `uses the raw text response body as errorMessage when responseType is text` to lock the behavior in.

core-web/libs/portlets/dot-velocity-playground/src/lib/dot-velocity-playground-page/dot-velocity-playground-page.component.ts
hmoreras: Agreed — fixed in f0c2d65e65. Both `bg-white` literals (host class and help-popover description) now use `bg-surface-0` so they follow the theme.

core-web/libs/portlets/dot-velocity-playground/src/lib/dot-velocity-playground-page/dot-velocity-playground-page.component.spec.ts
hmoreras: Fixed in f0c2d65e65. Replaced the silent-skip with an `Object.defineProperty` stub on `$exportMenu` (same pattern as the existing `$helpPopover` test) and now assert `toggle` is called with the click event — no more JSDOM-dependent no-op.

core-web/libs/portlets/dot-velocity-playground/src/lib/dot-velocity-playground-page/store/dot-velocity-playground.store.spec.ts
hmoreras: Done in f0c2d65e65. Added `selectHistoryEntry leaves code unchanged when the entry is not in history` covering the `!history.includes(entry)` early-return.

Issue #35470, opened 2026-04-27
[TASK] Content Drive: filter drive search by workflow (scheme with optional step)


zJaaal:
## Problem Statement

The content drive search endpoint (`POST /api/v1/drive/search`, [`ContentDriveResource.search()`](dotCMS/src/main/java/com/dotcms/rest/api/v1/drive/ContentDriveResource.java)) exposes no workflow filter. The content-drive chip-filter UI (#35880) lets the user pick **multiple workflow schemes, each optionally pinned to one step**, and needs the search to honor that selection.

## Approach

**DB-first** — filter in the SQL layer (`BrowserAPIImpl.selectQuery`), keeping the drive's DB-first hybrid intact (no forced Elasticsearch path). Added to the shared SQL builder, so it covers both `drive/search` and `/v1/browser`; only the drive REST form is wired.

## Request contract (FE → BE)

A single optional `workflow` field — a list of `{ scheme, step? }`. Each entry is one selected scheme, optionally pinned to one step (step omitted = "all steps of that scheme").

```json
"workflow": [
  { "scheme": "d61a59e1-…", "step": "dc3c9cd0-…" },
  { "scheme": "2a4e1d2e-…" }
]
```

## Semantics

- **Scheme-only entry** → content whose **content type is assigned that scheme** (`workflow_scheme_x_structure`), so **imported / push-published / never-actioned content** (no `workflow_task`) still appears under its scheme.
- **Step-pinned entry** → content whose **current task** is at that step (`workflow_task.status`).
- Entries combine with **OR**. Task-less content matches at the scheme level but not a specific step.
- Trade-off: a multi-scheme content type matches under every assigned scheme regardless of its current task — consistent with "content governed by this workflow".

## Acceptance Criteria

- `DotContentDriveSearchRequest` + the Java form accept the optional `workflow: [{ scheme, step? }]` field.
- `[{scheme:A}]` → all content of A-assigned types, including imported/never-actioned content with no task.
- `[{scheme:A, step:X}]` → content whose current task is at step X.
- `[{scheme:A}, {scheme:C, step:Y}]` → (A-assigned types) ∪ (current task at Y).
- Multi-language filtering is per-language.
- Empty/omitted `workflow` → no regression.
- Folders suppressed when a workflow filter is active.

## Tests

[...]

PR #35906, opened 2026-06-02
feat(content-drive): add workflow filter (scheme → step) to toolbar #35880

### LINKED ISSUE #35880: [TASK] Content Drive: Add Workflow filter (Scheme → Step) to the toolbar
## Description

Add a **Workflow filter** to the Content Drive toolbar, allowing users to narrow the content listing by **workflow scheme** and **workflow step**.

This filter should mirror the existing **content-type filter** (`dot-content-drive-content-type-filter`), which uses a two-column popover: the **left column** lists the parent dimension and the **right column** lists the children narrowed by the focused parent. For this filter:

- **Left column** → Workflow **Schemes**
- **Right column** → **Steps** belonging to the focused scheme

Selecting a scheme with no step narrowing filters by "all steps of that scheme"; selecting specific steps narrows further (same cascade-up / cascade-down behavior the content-type filter uses between base types and content types).

### Reference / pattern to follow

| Piece | Reference file |
|-------|----------------|
| Filter component | `libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-content-type-filter/` |
| Toolbar wiring | `dot-content-drive-toolbar/dot-content-drive-toolbar.component.html` |
| Store filters API | `store/dot-content-drive.store.ts` (`patchFilters`, `removeFilter`, `getFilterValue`) |
| Query builder | `utils/functions.ts` → `buildContentDriveQuery` |
| Filter types | `shared/models.ts` → `DotKnownContentDriveFilters` |
| Shared UI | `DotChipFilterComponent`, `DotFilterListItemComponent` (`@dotcms/portlets/content-drive/ui`) |

### Technical notes

- Add the new filter key(s) to `DotKnownContentDriveFilters` (e.g. `workflowScheme`, `workflowStep` / `wfStep`). The query builder already handles `string[]` filter values generically (`+key:(a OR b)`), so wiring a new multi-select key requires no changes to `buildContentDriveQuery` beyond confirming the correct Lucene field names for the index (e.g. `wfScheme`, `wfStep`).
- Persist selection through the store's `patchFilters` / `removeFilter` like the content-type filter does, so URL encode/decode round-trips for free via `encodeFilters` / `decodeFilters`.
- Source data: workflow schemes and their steps come from the workflow service/API — confirm the available endpoint(s) for schemes → steps before building.
- Register the component in the toolbar's filter row alongside `dot-content-drive-content-type-filter` and `dot-content-drive-language-field`, and ensure it participates in the existing **Clear all filters** action.
- Follow project conventions: `dot-` prefix, standalone, `@if`/`@for`, `input()`/`output()`, i18n keys via `DotMessagePipe`, `data-testid` on interactive elements, Jest + Spectator tests.

### ⚠ Open question — multi-select vs single-select

It is **not yet decided** whether scheme/step selection should be **multi-select** (matching the content-type filter, which allows multiple base types and content types) or **single-select**. This needs to be resolved before/while implementing:

- **Multi-select** → reuse the content-type filter's chip + cascade behavior verbatim.
- **Single-select** → simpler popover, single chip, no cascade.

The acceptance criteria below assume the multi-select pattern for parity with the existing filter; **adjust if single-select is chosen.**

## Acceptance Criteria

- [ ] A Workflow filter control is added to the Content Drive toolbar's filter row, next to the content-type and language filters.
- [ ] Opening the filter shows a two-column popover: workflow **schemes** on the left, **steps** of the focused scheme on the right.
- [ ] Selecting a scheme (no steps narrowed) filters content by all steps of that scheme; selecting specific steps narrows the result further.
- [ ] Selection cascades like the content-type filter: selecting a step selects its scheme (cascade up); deselecting the last step of a scheme drops the scheme (cascade down). *(If single-select is chosen, this criterion is replaced by single-selection behavior.)*
- [ ] Active selections render as chips with human-readable labels (scheme/step names), with an "all" suffix when a scheme is selected without step narrowing.
- [ ] The selection is persisted to the store via `patchFilters` and produces a correct Lucene query through `buildContentDriveQuery` (verified against the index field names).
- [ ] Selections round-trip through the URL (`encodeFilters` / `decodeFilters`) and are restored on reload.
- [ ] The filter is cleared by the toolbar's **Clear all filters** action and contributes to `$hasFilters()`.
- [ ] All user-facing text uses i18n keys; interactive elements have `data-testid`; icon-only controls have `aria-label`.
- [ ] Component is covered by Jest + Spectator tests following the content-type filter's spec as a model.
- [ ] **Decision recorded** in the issue/PR on whether the filter is multi-select or single-select before merge.

## Additional Context

- Existing related component (not a filter): `dot-content-drive-workflow-actions` handles bulk workflow **actions** — this new work is a **filter**, distinct from that.
- The content-type filter handles pagination, lazy loading, debounced search, and in-flight request cancellation. Evaluate which of these are needed for workflow schemes/steps (the scheme/step catalogs are typically small, so virtual scrolling / pagination may be unnecessary — keep it as simple as the data allows).

zJaaal:
### Proposed Changes

Frontend **and** backend for the Content Drive **Workflow filter** (scheme → optional step), implemented together in this PR. Closes #35880 (frontend) and #35470 (backend — consolidated here).

**Frontend (#35880)**
* Workflow filter in the Content Drive toolbar — a two-column popover (schemes → steps) mirroring the content-type filter.
* **Multi-select schemes, single step per scheme**: each selected scheme can pin one step; pinning a step cascades to select its scheme, deselecting a scheme drops its pinned step.
* Selection persists as a single structured filter key `workflow: string[]` (entries are `schemeId` or `schemeId:stepId`), so it round-trips through the URL via the existing `encode`/`decode` machinery with no step lookups.
* Scheme source follows the content-type selection (all schemes / schemes for the selected content types); steps load on demand and are cached per scheme.
* Empty-schemes state shows a dedicated, i18n'd message (singular/plural by selected content-type count); empty/loading states are centered.
* New `DotWorkflowService` methods: `getSchemesByContentTypes()` and `getSteps()`.

**Backend (#35470) — DB-first**
* `POST /api/v1/drive/search` now honors an optional `workflow: [{ scheme, step? }]` field.
* Filter lives in the **shared SQL builder** (`BrowserAPIImpl.selectQuery` → `appendWorkflowQuery`), keeping the drive's DB-first hybrid intact — **no forced Elasticsearch path**.
* **Scheme-only entry** → matches by content-type assignment (`workflow_scheme_x_structure`), so **imported / push-published / never-actioned** content (no `workflow_task`) still appears under its scheme.
* **Step-pinned entry** → matches the contentlet's **current task** (`workflow_task.status`). Entries combine with **OR**. Folders are suppressed when a workflow filter is active.
* New immutable `WorkflowFilterForm` on `DriveRequestForm`; `BrowserQuery.withWorkflowSchemeIds` / `withWorkflowStepIds`. All ids are bound as `?` parameters (no string concatenation).
* FE `$request` wires the stored `workflow` tokens into the drive-search request.

**Chore**
* `chore(vscode)`: add MCP server config (primeng, nx, angular-cli).

### Checklist
- [x] Tests
  - FE: workflow-filter component spec + `workflow` encode/decode round-trip in `functions.spec.ts` + store empty-result pagination test
  - BE: `ContentDriveWorkflowFilterTest` integration test — scheme filter (incl. task-less content), step filter (current task), OR semantics, no-filter regression, folder suppression
- [x] Translations — `content-drive.workflow-filter.*` keys added to `Language.properties`
- [x] Security Implications Contemplated — read-only filter; all workflow ids bound as SQL parameters (no concatenation); MCP config contains no secrets

### Additional Info
The backend ticket #35470 has been **consolidated into this PR** (and its design folded into #35880); it is closed in favor of #35906. The request contract there matches the structured `workflow` field used here.

This PR fixes: #35880


core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-workflow-filter/dot-content-drive-workflow-filter.component.ts
```diff
@@ -0,0 +1,375 @@
...
+ *   scheme keeps the pairing explicit (no need to infer a step's scheme), so
+ *   reconcile and URL restore need no step lookups.
+ * - Focus (which scheme's steps show on the right) is separate from selection,
+ *   exactly like base-type focus vs its checkbox in the content-type filter.
+ * - Cascade: pinning a step selects its (focused) scheme; deselecting a scheme
+ *   removes its pinned step.
+ *
+ * NOTE: backend search support is tracked in dotCMS/core#35470. Until it ships,
+ * this filter renders, persists its selection, drives the chip + Clear-all, and
+ * round-trips in the URL, but does NOT change the listing results yet.
```
oidacra: This NOTE is now stale: this PR consolidates the #35470 backend (store.ts wires `parseWorkflowFilter` into the drive-search request, and `BrowserAPIImpl.appendWorkflowQuery` + `ContentDriveHelper` apply it server-side), so the filter does change listing results. Suggest removing this paragraph or replacing it with an accurate description (tokens serialize as `schemeId[:stepId]`, applied as SQL, OR-combined).

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-workflow-filter/dot-content-drive-workflow-filter.component.ts
```diff
@@ -0,0 +1,375 @@
...
+        patchState(this.$state, { loadingSchemes: true });
+
+        const source$ = contentTypes.length
+            ? this.#workflowService.getSchemesByContentTypes(contentTypes)
+            : this.#workflowService.get();
+
+        source$
+            .pipe(
+                take(1),
+                catchError(() => of([] as DotCMSWorkflow[])),
```
oidacra: `#loadSchemes` swallows the error with `catchError(() => of([]))` and still calls `#reconcileSelection(schemes)`. On error `schemes` is `[]`, so reconcile sets `$selection` to `[]` and `#syncStore()` calls `store.removeFilter('workflow')` — a transient backend failure here (this runs from the constructor effect on load, including URL-restored filters) silently drops the user's persisted filter and reloads unfiltered, with no logging or feedback. Consider splitting next/error: on error set `loadingSchemes: false` and leave the existing selection intact (skip reconcile with the empty fallback), and surface/log the error per the `catchError -> httpErrorManager.handle` convention.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/utils/functions.ts
```diff
@@ -78,6 +78,23 @@ const multiSelector: DotContentDriveDecodeFunction = (value = ''): string[] =>
...
 
+/**
+ * Parses the `workflow` filter tokens (`schemeId` or `schemeId:stepId`) into the
+ * `{ scheme, step? }` entries the drive-search request expects. Splits on the FIRST
+ * colon only, so the inner step id is preserved.
+ *
+ * @param {string[]} tokens
+ * @return {*}  {{ scheme: string; step?: string }[]}
+ */
+export function parseWorkflowFilter(tokens: string[] = []): { scheme: string; step?: string }[] {
```
oidacra: `parseWorkflowFilter` here and `parseSelection` in dot-content-drive-workflow-filter.component.ts implement the same `schemeId[:stepId]` first-colon split. Consider exporting one canonical `parseWorkflowToken(token)` from this file and having both `parseWorkflowFilter` and the component use it (its `toToken` inverse could live here too), so the encoding lives in one place.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-workflow-filter/dot-content-drive-workflow-filter.component.ts
zJaaal: Nice catch! Updating the comment

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-workflow-filter/dot-content-drive-workflow-filter.component.ts
zJaaal: Yes that is a complete bug, working on it!

core-web/libs/portlets/dot-content-drive/portlet/src/lib/utils/functions.ts
zJaaal: Removing duplicated code!

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-workflow-filter/dot-content-drive-workflow-filter.component.ts
```diff
@@ -0,0 +1,382 @@
...
+        const map = new Map<string, string>();
+        Object.values(this.#stepCache()).forEach((steps) =>
+            steps.forEach((step) => map.set(step.id, step.name))
+        );
+        return map;
+    });
+
+    /** Current selection, parsed from the single `workflow` filter key. */
+    readonly $selection = linkedSignal<WorkflowSelection[]>(() =>
+        ((this.#store.getFilterValue('workflow') as string[]) ?? []).map(parseWorkflowToken)
```
oidacra: `getFilterValue` returns `string | string[]` (from the `[key: string]: string | string[]` index signature on `DotContentDriveFilters`), so the bare `as string[]` casts here (and at :155, :239) aren't type-safe — if a value is ever a single `string`, the cast succeeds silently and `.map(parseWorkflowToken)` / `.length` would operate on the string character-by-character.

Suggest narrowing instead:
```ts
const raw = this.#store.getFilterValue('workflow');
const tokens = Array.isArray(raw) ? raw : [];
return tokens.map(parseWorkflowToken);
```

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/dot-content-drive.store.ts
```diff
@@ -74,6 +74,9 @@ export const DotContentDriveStore = signalStore(
                     baseTypes: filters()?.baseType?.map(
                         (baseType) => MAP_NUMBERS_TO_BASE_TYPES[Number(baseType)]
                     ),
+                    workflow: filters()?.workflow?.length
```
oidacra, line 77: `$request` now maps `workflow` through `parseWorkflowFilter` and adds `!filters()?.workflow?.length` to the `showFolders` guard, but `dot-content-drive.store.spec.ts` has no test for either — there are zero references to `workflow` / `showFolders` in the store spec.

Worth adding two store tests, mirroring the existing filter patterns:
1. `workflow: ['a:a2', 'b']` → `request.workflow` equals `[{ scheme: 'a', step: 'a2' }, { scheme: 'b' }]`
2. `request.showFolders` is `false` when the workflow filter is non-empty

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-workflow-filter/dot-content-drive-workflow-filter.component.ts
zJaaal: Noted!

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/dot-content-drive.store.ts
zJaaal: Added!

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
```diff
@@ -1769,6 +1771,51 @@ private void appendFilterQuery(StringBuilder sqlQuery, String filter,
...
+                    + ")) ");
+            parameters.addAll(workflowSchemeIds);
+        }
+
+        if (hasSteps) {
+            final String placeholders = workflowStepIds.stream()
+                    .map(id -> "?").collect(Collectors.joining(","));
+            orClauses.add(" exists (select 1 from workflow_task wt "
+                    + " where wt.webasset = cvi.identifier and wt.language_id = cvi.lang "
+                    + " and wt.status in (" + placeholders + ")) ");
```
fabrizzio-dotCMS, line 1816: We need to sanitize this with SQLUtils. sanitize

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
```diff
@@ -1769,6 +1771,51 @@ private void appendFilterQuery(StringBuilder sqlQuery, String filter,
...
+            parameters.addAll(workflowStepIds);
+        }
+
+        sqlQuery.append(" and (").append(String.join(" or ", orClauses)).append(") ");
+    }
+
     /**
```
fabrizzio-dotCMS, line 1823: **HINT**

There is this method: 
```
private SelectQuery selectQuery(final BrowserQuery browserQuery) {
  ... 
  At the end of this method, there is this:
    Logger.debug(this, "Select Query: " + selectQuery);
}
```
Enable that logger line temporarily to print out the SQL statement and evaluate the resulting SQL using AI feed it the postgresql.sql file for more context 
and also test this functionality with a fair amount of data to see how it impacts performance

fabrizzio-dotCMS (commented):
Just a couple of things to review: 

There are a couple of important things to keep in mind.

First, this code is shared with Site Browser, so any changes introduced here must not break Site Browser functionality. Whenever a new field is added that is not sent by Site Browser, make sure to guard its processing with a validation check, for example:

```
if (isSet(newField)) {     
 // Process parameters in the query builder 
} 
```
This ensures that existing behavior remains unaffected and that new parameters do not inadvertently impact current functionality.

The second thing to pay close attention to is the generated query itself. I always print the final query and review it carefully to make sure there are no potential performance issues. I left a hint for you to review

core-web/libs/portlets/dot-content-drive/portlet/src/lib/utils/functions.ts
```diff
@@ -78,6 +34,52 @@ const multiSelector: DotContentDriveDecodeFunction = (value = ''): string[] =>
...
+
+/**
+ * Canonical parse for one `workflow` token. Splits on the FIRST separator only,
+ * so any separator inside the step id is preserved.
+ * `'A:X'` → `{ scheme: 'A', step: 'X' }`; `'B'` → `{ scheme: 'B' }`.
+ *
+ * @param {string} token
+ * @return {*}  {WorkflowFilterEntry}
+ */
+export function parseWorkflowToken(token: string): WorkflowFilterEntry {
```
oidacra, line 54: `parseWorkflowToken`, `workflowEntryToToken` and `parseWorkflowFilter` are exported but only covered indirectly via the URL encode/decode round-trip. Since they're pure functions and the canonical token (de)serializers, consider a direct `describe('parseWorkflowToken')` covering the edge cases: empty token `''`, leading separator `:stepId`, and a step id containing colons (`'scheme:step:with:colons'` → `{ scheme: 'scheme', step: 'step:with:colons' }`), plus a `workflowEntryToToken` scheme-only vs scheme+step pair. Cheap insurance against a future refactor that keeps the round-trip green while breaking an edge case.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-workflow-filter/dot-content-drive-workflow-filter.component.spec.ts
```diff
@@ -0,0 +1,368 @@
...
+
+    /** Plain attribute selector — `click`/`triggerEventHandler` need a string, not a DOMSelector. */
+    const testId = (id: string) => `[data-testid="${id}"]`;
+
+    /** Default store: no active filters. Override `workflow`/`contentType` per test. */
+    const stubFilters = (values: Record<string, string[]> = {}) =>
+        (store.getFilterValue as jest.Mock).mockImplementation((key: string) => values[key]);
+
+    const openPanel = () => {
+        spectator.click(testId('workflow-filter-chip'));
```
oidacra: `openPanel()` clicks via the raw CSS string from `testId(...)`. Since `spectator.click()` accepts a `byTestId(...)` result (already imported here), prefer `spectator.click(byTestId('workflow-filter-chip'))` for consistency with the project's Spectator idiom — keep the `testId()` string helper only for the `triggerEventHandler` calls that genuinely need a CSS string.

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
zJaaal: Checked! Performance seems to be on point

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
zJaaal: Done!

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-workflow-filter/dot-content-drive-workflow-filter.component.spec.ts
```diff
@@ -0,0 +1,368 @@
...
+            stepsA$.next(STEPS_A);
+
+            expect(spectator.component.$state().steps).toEqual(STEPS_B);
+        });
+    });
+
+    describe('empty schemes', () => {
+        const schemeEmptyText = () =>
+            spectator.fixture.debugElement
+                .query(By.css('[data-testid="workflow-scheme-empty"]'))
```
oidacra, line 346: `schemeEmptyText()` queries via `By.css('[data-testid="workflow-scheme-empty"]')`. Prefer `spectator.query(byTestId('workflow-scheme-empty'))` — it's the project's Spectator idiom and would let you drop the `By` import from `@angular/platform-browser` entirely.

oidacra:
This PR removes `$query` / `buildContentDriveQuery` / `escapeLuceneSpecialChars` / `BASE_QUERY`. Per git history (`2d7eb3e6`, #35503) that `$query` computed was intentionally kept with an inline note tying it to **#33338** ("Enable select all checkbox"), which is still open. The removal looks like reasonable cleanup if select-all is now going server-side via the new `$request` pipeline — could you confirm that's the plan (and ideally note it on #33338, or close it), so whoever picks that issue up later doesn't have to reconstruct the Lucene-query infra from history?

core-web/libs/data-access/src/lib/dot-workflow/dot-workflow.service.ts
```diff
@@ -65,6 +70,52 @@ export class DotWorkflowService {
...
+     * Get the steps that belong to a workflow scheme.
+     *
+     * @param {string} schemeId
+     * @return {*}  {Observable<WorkflowStep[]>}
+     * @memberof DotWorkflowService
+     */
+    getSteps(schemeId: string): Observable<WorkflowStep[]> {
+        return this.httpClient
+            .get<{ entity: WorkflowStep[] }>(`${this.WORKFLOW_URL}/schemes/${schemeId}/steps`)
+            .pipe(map((x) => x?.entity));
```
oidacra: Minor consistency nit: `getSteps` maps `(x) => x?.entity` without the `?? []` fallback that the sibling `getSchemesByContentTypes` (added in this same PR) uses. The declared return is `Observable<WorkflowStep[]>`, but on a 200 with an empty/`null` body this emits `undefined`, which then lands in the step signal state and `#stepCache` and would break the `!$state.steps().length` empty-state check (blank panel, no spinner). Unlikely against the real backend, but aligning to `.pipe(map((x) => x?.entity ?? []))` matches the sibling method and keeps the runtime value consistent with the type.

zJaaal:
> This PR removes `$query` / `buildContentDriveQuery` / `escapeLuceneSpecialChars` / `BASE_QUERY`. Per git history (`2d7eb3e6`, #35503) that `$query` computed was intentionally kept with an inline note tying it to **#33338** ("Enable select all checkbox"), which is still open. The removal looks like reasonable cleanup if select-all is now going server-side via the new `$request` pipeline — could you confirm that's the plan (and ideally note it on #33338, or close it), so whoever picks that issue up later doesn't have to reconstruct the Lucene-query infra from history?

The removal was intentional

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-workflow-filter/dot-content-drive-workflow-filter.component.spec.ts
zJaaal: done!

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-workflow-filter/dot-content-drive-workflow-filter.component.spec.ts
zJaaal: After testing, we need to use the By.css for this case

core-web/libs/data-access/src/lib/dot-workflow/dot-workflow.service.ts
zJaaal: Fixed!

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/drive/ContentDriveWorkflowFilterTest.java
```diff
@@ -0,0 +1,323 @@
...
+ *   (imported / push-published / never-actioned) still appears under its scheme.</li>
+ *   <li><b>Step-pinned entry</b> — matches the contentlet's <i>current</i> task
+ *   ({@code workflow_task.status}); task-less content is therefore excluded.</li>
+ *   <li><b>OR semantics</b> — multiple entries union their matches.</li>
+ *   <li><b>No filter</b> — result set is unchanged (regression guard).</li>
+ *   <li><b>Folder suppression</b> — folders carry no workflow state, so they are dropped when a
+ *   workflow filter is active.</li>
+ * </ul>
+ */
+@ApplicationScoped
```
fabrizzio-dotCMS, line 59: This test must be registered in one of our TestSuites; otherwise, it'll never run in our pipeline

fabrizzio-dotCMS (changes requested):

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/drive/ContentDriveWorkflowFilterTest.java
fabrizzio-dotCMS: Find the Suite where BrowserAPITest is registered and use that one

zJaaal: Added!

hmoreras (approved):

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/drive/ContentDriveWorkflowFilterTest.java
zJaaal: Good catch 🙏 — registered `ContentDriveWorkflowFilterTest` in `MainSuite3a` (alongside the sibling `ContentDriveHelperContentletAPIComparisonTest`) in adb928b661, so it now runs in the pipeline.

fabrizzio-dotCMS (approved):

PR #35911, opened 2026-06-03
build: make Java 25 the default, demote Java 21 to the variant

sfreudenthaler:
## What

Make **Java 25 the default** for all builds and for the published `dotcms/dotcms:latest` image, and demote **Java 21 to the parallel variant** (`…_java-21`).

This is a single source-of-truth change:

| File | Change |
|------|--------|
| `.sdkmanrc` | `java=21.0.8-ms` → `java=25.0.2-ms` |
| `docker/java-base/Dockerfile` | default `ARG SDKMAN_JAVA_VERSION` → `25.0.2-ms` (cosmetic; callers always pass the arg) |
| `cicd_8-manual-deploy.yml` | fix stale "default Java 21" doc string → "default Java 25" |

## Why this is all that's needed (no workflow tag-logic changes)

The `latest` tag is gated in `cicd_comp_deployment-phase.yml` on `inputs.latest && docker_suffix == ''`. The **primary** release build (`cicd_6-release.yml`) reads its Java version from `.sdkmanrc` and has an **empty suffix**, so once `.sdkmanrc` says Java 25, the primary build ships Java 25 and automatically wins `latest` (and, via `cicd_weekly-rolling-tags.yml`, `weekly`). The Java 21 build moves to the variant pathway (`cicd_7-release-java-variant.yml`) and ships under `…_java-21` + the floating `java-21` tag.

`dotcms/java-base:25.0.2-ms` already exists on Docker Hub (the Java 25 variant has been building against it), so the base-image path (`FROM dotcms/java-base:${SDKMAN_JAVA_VERSION}`) is already proven.

## ⚠️ Required companion change at merge time (repo variables, not in this PR)

Repo variables are GitHub settings, so they can't live in the PR. **At merge, flip them** so the variant builds Java 21 instead of Java 25:

```
RELEASE_JAVA_VARIANT_VERSION:  25.0.2-ms  →  21.0.8-ms
RELEASE_JAVA_VARIANT_SUFFIX:   java-25    →  java-21
```

If `.sdkmanrc` is merged but the vars are **not** flipped: the variant would also build Java 25 (duplicate of `latest`), and **no Java 21 image would be produced** — breaking consumers still pinned to Java 21. They must change together.

## Resulting tags after merge + var flip

- `dotcms/dotcms:latest`, `:weekly`, `:{version}` → **Java 25**
- `dotcms/dotcms:java-21`, `:{version}_java-21` → **Java 21** (transition support)

## Rollout

[...]

PR #35914, opened 2026-06-03
feat(java): move to Java 25 (runtime 25.0.2-ms, default bytecode 25)

wezell:
## What

Moves dotCMS to **Java 25** — both runtime and default compiler bytecode target. This is a clean re-application of the `issue-33865-java-25-part2` work onto current `main` (the original branch was ~500 commits behind and tangled with merge commits).

## Changes

**Runtime / build JDK → Java 25.0.2-ms**
- `.sdkmanrc`: `java=25.0.2-ms`
- `docker/java-base/Dockerfile`: `SDKMAN_JAVA_VERSION=25.0.2-ms`
- `parent/pom.xml`: `jdk.min.version` 21 → 25 (enforcer minimum)

**Compiler bytecode target → 25 by default**
- `parent/pom.xml`: kept main's configurable `dotcms.core.compiler.release` property but flipped the default from **11 → 25**. Override with `-Ddotcms.core.compiler.release=11` for older bytecode compatibility.
- `test-jmeter/pom.xml`, `test-karate/pom.xml`: compiler source/target/release 21 → 25

**Profiling**
- `parent/pom.xml`: `glowroot.version` → `0.14.5-beta.3-java25` (Java 25-compatible build)

**Docs**
- `CLAUDE.md`: runtime references 21 → 25
- Added `JAVA25_PROBLEMS_AND_SOLUTIONS.md` and `JAVA25_SMOKE_TEST.md`

## Notes / Review focus
- `docker/dev-env/Dockerfile` intentionally **not** changed — main restructured it to inherit Java from the base image (`COPY --from=dotcms /java /java`).
- ⚠️ Verify `glowroot 0.14.5-beta.3-java25` resolves from the configured Maven repositories.

This PR fixes: #33865

This PR fixes: #33865


wezell:
Thanks for the review — addressed in 57d2ce969c:

**1. CLAUDE.md "Java 11 syntax" vs default bytecode 25** ✅ Fixed. The Java 11-compatibility constraint is intentionally being lifted for core — this PR makes Java 25 the default compile target. Updated the docs (Critical Rules + Tech Stack) to state core compiles to Java 25 by default (override-able via `-Ddotcms.core.compiler.release`), removing the contradiction.

**2. Glowroot `0.14.5-beta.3-java25` resolution** ✅ Verified present on Artifactory — `HTTP 200` on both:
- `https://repo.dotcms.com/artifactory/libs-release/org/glowroot/glowroot-agent/0.14.5-beta.3-java25/glowroot-agent-0.14.5-beta.3-java25.jar`
- `https://repo.dotcms.com/artifactory/libs-release-local/...`

It's a dotCMS-built Java 25-compatible agent (standard glowroot releases don't yet support JDK 25), deployed to Artifactory, so CI and other devs resolve it fine. No change needed.

**3. test-jmeter / test-karate hardcoded 25** ✅ Fixed. Both now use `${dotcms.core.compiler.release}` so the documented `-Ddotcms.core.compiler.release=11` escape hatch applies to them too.

**4. Bytecode 25 breaks JVM < 25 deploys** ✅ Acknowledged — this is the intended outcome of the "Run on Java 25" spike, not an oversight. The same PR moves the runtime/base images to `25.0.2-ms`, so build and deploy targets move to JVM 25 together. Rollout is gated on all environments being on JVM 25.

**Minor (950 lines of docs):** intentional for now — `JAVA25_PROBLEMS_AND_SOLUTIONS.md` / `JAVA25_SMOKE_TEST.md` capture migration findings for the spike. Can relocate to the wiki before this leaves draft if preferred.

wezell:
Thanks — all three are pre-existing state rather than changes introduced by this PR, so I've left them out of scope to keep this focused on the core Java 25 default:

**1. `tools/dotcms-cli/pom.xml` hardcoded at 11** — Unchanged by this PR. The CLI is a Quarkus/native-image module under the `dotcms-cli-parent` chain and intentionally targets Java 11 for distribution portability. It deliberately does *not* track the core compiler release. The CLAUDE.md note ("CLI may target lower for portability") documents exactly this. Agree a pom comment would help — happy to add it, but as a separate cleanup so this PR stays scoped to core.

**2. `tika-plugin` / `tika-api`** — These are OSGi plugins under `dotcms-core-plugins-parent`, a different parent from the `parent/pom.xml` this PR modifies, so they're unaffected and intentionally stay at Java 11 for OSGi container compatibility. Not in scope for "core moves to 25"; they'd be a deliberate, separate decision if ever revisited.

**3. `glowroot.version` location** — This property already lived in `parent/pom.xml` on `main` (`0.14.4`); `bom/application/pom.xml` consumes it via `${glowroot.version}`. This PR only bumps the value in its existing location, so it doesn't introduce a new BOM-rule violation. The "versions live in the BOM" rule targets new dependency version numbers; relocating this long-standing property is a separate refactor.

On the beta-fork concern (good call): `0.14.5-beta.3-java25` is a dotCMS-built agent because upstream Glowroot doesn't yet support JDK 25. I can file a tracking issue to "replace with official Glowroot JDK 25 support when available" so it isn't forgotten — let me know and I'll open it.

sfreudenthaler (commented):
Possible duplicate of #35911 

Different approach though.  Where 35911 keeps building both (requiring a removal later in another PR), @wezell's approach one shots it for automatic build and leaves it available for manual trigger.

Honestly that's probably enough because rollback of java runtime version should come with a plugin check-gate and user coordination. 


The rest of the PR looks good so it's a timing and approach issue to hash out.

wezell:
@sfreudenthaler I think this change was missing from your pr which unlocks our ability to use all the cool new java features:

https://github.com/dotCMS/core/pull/35914/changes#diff-b5a06276719e759fe07dfe6f75d781be5f83d2215179d82bdb195ad035348214R103

sfreudenthaler:
> @sfreudenthaler I think this change was missing from your pr which unlocks our ability to use all the cool new java features:
> 
> https://github.com/dotCMS/core/pull/35914/changes#diff-b5a06276719e759fe07dfe6f75d781be5f83d2215179d82bdb195ad035348214R103

sure is.  i can fold it in, just wanted to decide the approach first.  i'm fine with yours, it's less cautious but i think it's the same end result so "cleaner".  

i'll close mine out as duplicate and you can have the internet points 😄

sfreudenthaler (approved):
after waffling a bit on approach.  I've come around to this one.  My potential alternate approach looked more cautious but in practice was just theatre.  Resulting in the same outcome. 

let's go with this one and i'll close out #35911

PR #35915, opened 2026-06-03
ci(java): stop auto-building the java-25 release variant

wezell:
## What

Now that Java 25 is the default build (#33865 / companion PR #35914), the parallel **java-25 variant** image that `cicd_7-release-java-variant.yml` builds on every release is redundant with the primary release image (`dotcms/dotcms:latest`, `:trunk`, `:<version>`).

This removes the automatic `push: release-*` trigger so the java-25 variant is no longer built on release.

## Details

- `cicd_7` is the **only** workflow that auto-builds a Java variant (verified: trunk/nightly/release/manual-deploy only build a variant when a human passes `java-version` explicitly).
- The workflow is **kept as manual-only** (`workflow_dispatch`) rather than deleted, so it can be repurposed for a future **java-11 variant** (the planned follow-up to preserve the old default for users not yet on JVM 25).
- Header comment updated to reflect manual-only operation.

## Reviewer notes

- The repository variables `RELEASE_JAVA_VARIANT_VERSION=25.0.2-ms` and `RELEASE_JAVA_VARIANT_SUFFIX=java-25` are **left untouched** (they're live repo settings, not in-tree). They simply go dormant. ⚠️ Consequence: a *manual* `workflow_dispatch` of `cicd_7` with no inputs would still build a **java-25** image from those vars. That's acceptable for a deliberate manual run; the java-11 follow-up PR will repoint these vars.
- Depends on / pairs with #35914 (which makes Java 25 the default).

Refs #33865

This PR fixes: #33865

### DIFF
--- a/.github/workflows/cicd_7-release-java-variant.yml
+++ b/.github/workflows/cicd_7-release-java-variant.yml
@@ -24,7 +24,7 @@
 # Java version for the same release are prevented via concurrency groups.
 #
 # Configuration:
-# Set repository variables for automatic parallel builds:
+# Repository variables provide defaults when the dispatch inputs are left empty:
 #   RELEASE_JAVA_VARIANT_VERSION: Java version in SDKMAN format (e.g., 25.0.2-ms)
 #   RELEASE_JAVA_VARIANT_SUFFIX: Artifact suffix without separator (e.g., java-25-ms)
 #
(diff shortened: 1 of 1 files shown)


.github/workflows/cicd_7-release-java-variant.yml
```diff
@@ -31,17 +33,19 @@ name: '-7 Release Java Variant'
 run-name: "Release Java Variant - ${{ github.event_name == 'workflow_dispatch' && inputs.release_branch || github.ref_name }}${{ inputs.java_version && format(' [{0}]', inputs.java_version) || '' }}"
 
 on:
-  push:
-    branches:
-      - 'release-*'
+  # NOTE: Automatic trigger on release-* branches has been removed.
+  # Java 25 is now the default build (see #33865), so the previously
+  # auto-built java-25 variant is redundant with the primary release image.
```
sfreudenthaler: remove this extra doc fluff.  the manual-only is enough
```suggestion
```

sfreudenthaler (approved):
small change non-blocking

PR #35918, opened 2026-06-03
35910 add sort options on dropdownsradiocheckboxes on style editor schema generator

### LINKED ISSUE #35910: Add sort options on dropdowns/radio/checkboxes on Style Editor Schema Generator
### Description

Add sort options on dropdowns/radio/checkboxes on Style Editor Schema Generator

### Acceptance Criteria

- [ ] All options can be sorted on dropdowns/radios/checkboxes on schema generator

### Priority

Medium

### Additional Context

_No response_

KevinDavilaDotCMS:
https://github.com/user-attachments/assets/62f4809c-c2de-48b5-8b25-fdec89e84b64

fixes #35910

This PR fixes: #35910


core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-palette/components/dot-uve-style-editor-form/dot-uve-style-editor-form.component.ts
```diff
@@ -191,65 +225,105 @@ export class DotUveStyleEditorFormComponent {
...
-                        // Capture activeContentlet at the time of form change (before debounce)
-                        // and identify the editor mode for this update.
-                        map((formValues) => ({
-                            formValues,
+                        debounceTime(STYLE_EDITOR_INPUT_IDLE_SAVE_TIME)
+                    );
+
+                    // A commit is: a discrete field change, an input idle pause,
+                    // or an input signalling it finished editing (blur/Enter).
+                    return merge(discreteCommits$, idleInputCommits$, this.#inputCommit$).pipe(
```
oidacra: `#listenToFormChanges` uses `mergeMap`, so each rebuilt form's inner subscription stays alive for the component's lifetime, and `#inputCommit$` (a component-lifetime `Subject`) is merged into all of them. Because the post-`merge` `map` reads `editorSelected()` live, a stale value held by a *previous* form (e.g. an input edited on contentlet A, then selection moves to B without the input blurring) can be persisted against the **current** contentlet via either a later `#inputCommit$` or the old form's pending `debounceTime(2000)` idle timer — per-inner `distinctUntilChanged` does not stop it (it is the first emission for that inner). Note `#restoreFormFromRollback`'s JSDoc already assumes `switchMap` cancels old-form saves, but the operator here is `mergeMap`. Consider creating `#inputCommit$` *inside* the `mergeMap` projection (per-form, dies with its inner) or moving the outer chain to `switchMap`. The simpler "idle overwrites rollback" case is already safe thanks to the per-inner `distinctUntilChanged`.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-palette/components/dot-uve-style-editor-form/dot-uve-style-editor-form.component.ts
```diff
@@ -191,65 +225,105 @@ export class DotUveStyleEditorFormComponent {
...
+                        // Capture the full value and the active contentlet at commit
+                        // time (before the coalescing window), not at save time.
+                        map(() => ({
+                            formValues: form.getRawValue(),
                             activeContentlet: this.#uveStore.editorSelected()?.payload,
```
oidacra: On the idle branch, `activeContentlet` is read here — after `idleInputCommits$`'s `debounceTime(2000)` — so doesn't this reflect whatever is selected ~2s after the keystroke rather than the contentlet that was being edited? The pre-PR code captured `activeContentlet` at change time (before the debounce) on purpose. Would it make sense to move the `editorSelected()`/`pageType()` capture up into `changes$`, so all three commit branches freeze the contentlet at change time and the post-`merge` `map` becomes a pass-through? That would also close the wrong-contentlet path noted on the `merge(...)` line above.

core-web/libs/portlets/edit-ema/portlet/src/lib/shared/consts.ts
```diff
@@ -21,8 +21,25 @@ export const IFRAME_SCROLL_ZONE = 100;
...
-export const STYLE_EDITOR_TRADITIONAL_DEBOUNCE_TIME = 500;
+/**
+ * Coalescing window (ms) for persisting style-editor changes.
+ *
+ * This is intentionally small: it only batches a rapid burst of commits (e.g.
+ * several quick radio clicks, or a blur landing right after a change) into a
+ * single save. It is NOT used to "wait for typing to finish" — that is handled
+ * by blur/Enter and by STYLE_EDITOR_INPUT_IDLE_SAVE_TIME.
+ */
+export const STYLE_EDITOR_SAVE_DEBOUNCE_TIME = 250;
```
oidacra, line 32: This collapses the previous traditional-page window (500ms) and headless window (2s) into a single 250ms coalescing window for both page types. On traditional pages each surviving commit reaches `#performSave` and, on success, the handler calls `#uveStore.pageReload()` (a full iframe reload). Since `debounceTime` only coalesces commits closer than its window, halving it from 500ms→250ms means a burst of discrete-field changes (radio/dropdown) spaced 250–500ms apart that previously collapsed into one save+reload can now produce two. Was the traditional path considered here — should it keep a longer coalescing window (e.g. branch the `debounceTime` on `isTraditionalPage`)?

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-palette/components/dot-uve-style-editor-form/dot-uve-style-editor-form.component.ts
KevinDavilaDotCMS: Fixed in 495b17e8ad. Kept `mergeMap` (so pending saves from an old form still complete — the reason it was chosen), but two changes close the stale-form path:

- blur/Enter commits now flow through a per-form `blurCommits$` that gates on `this.$form() === form`, so a stale inner never reacts to a blur that happened on a newer form. `withLatestFrom(changes$)` also means a blur with no prior change in that form emits nothing at all.
- the save target is frozen at change time (see the thread below), so the only thing a stale inner can still do is complete its own pending idle save against the contentlet that was actually edited.

Also fixed the `#restoreFormFromRollback` JSDoc that incorrectly claimed `switchMap` semantics. Added a spec covering this exact scenario: `should ignore a blur on the new form in the stale (pre-rebuild) form stream`.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-palette/components/dot-uve-style-editor-form/dot-uve-style-editor-form.component.ts
KevinDavilaDotCMS: Fixed in 495b17e8ad — exactly as suggested: the `editorSelected()`/`pageType()` capture moved up into `changes$`, every commit branch (discrete, idle, blur) carries the context frozen at change time, and the post-`merge` `map` is gone.

One deliberate nuance: `formValues` are still read at **emit** time from the closed-over form, not frozen at change time. Freezing them too would let a late idle commit resurrect a stale snapshot over a newer discrete commit (the existing `should persist an in-progress (uncommitted) input value when a discrete field commits` spec catches this). Reading them live is safe because they always belong to that form — they can never cross contentlets; only the contentlet/page-type context could.

Added a regression spec: an idle commit firing after the selection moved to another contentlet saves against the original one.

core-web/libs/portlets/edit-ema/portlet/src/lib/shared/consts.ts
KevinDavilaDotCMS: Considered, and keeping the single 250ms window for now. The old 500ms traditional window did double duty (typing inactivity + burst coalescing); typing is now handled by blur/Enter + the 2s idle timer, so this constant only coalesces bursts of discrete commits.

The trade-off you describe is real: two radio clicks spaced 250–500ms apart can now produce two save+reload cycles on a traditional page where the old window collapsed them into one (`switchMap` still cancels the in-flight save, so the worst case is one extra reload). We took the simpler single constant; if it proves noisy in practice, branching the `debounceTime` on `isTraditionalPage` is a small, contained follow-up.

oidacra (approved):

PR #35924, opened 2026-06-03
fix(edit-content): reduce Content Status and locale chip height to 1.75rem

### LINKED ISSUE #35891: Reduce Content Status and locale chip height to 1.75 rem
### Description

The **Content Status** and **locale** chips are too big. Reduce the chip **height to 1.75 rem**, adjusting font-size and padding as needed so the content remains vertically centered and legible.

These chips are shared across multiple surfaces — the change must apply consistently everywhere they appear:

| Surface | Location |
|---|---|
| **Content Drive** | Content listing |
| **New Edit Contentlet** | Command bar |
| **New Edit Contentlet** | Side panel |
| **New Edit Contentlet** | Relationship field |

### Acceptance Criteria

Share with Design how it looks in each of the places, so they can confirm it's what they expect.

- [ ] The Content Status chip renders at a height of **1.75 rem**
- [ ] The locale chip renders at a height of **1.75 rem**
- [ ] Chip label text and any icon remain vertically centered within the resized chip
- [ ] The new height is applied in **Content Drive**
- [ ] The new height is applied in the New Edit Contentlet **command bar**
- [ ] The new height is applied in the New Edit Contentlet **side panel**
- [ ] The new height is applied in the New Edit Contentlet **relationship field**

### Priority

Medium

### Additional Context

The chips are a shared component, so a single styling change should propagate to all four surfaces. Verify each surface after the change to confirm consistency. `1,75 rem` (comma decimal) = **1.75 rem**.

oidacra:
## Summary

Makes dotCMS chips **compact by default** — `1.75 rem` tall — by setting the height on the base `.p-chip` style in the custom Lara preset (`theme.config.ts`), via `components.chip.css` (the same mechanism already used for `card` and `confirmpopup`).

This covers the **Content Status** and **locale** chips (and every other chip) with **no per-template sizing classes** — the size is centralized in the theme.

Closes #35891

## Approach

The chip styling lives in the **JS theme preset** (`CustomLaraPreset`), not in SCSS. In PrimeNG v21 the theme is driven by the preset; the legacy `dotcms-theme.scss` is disabled (commented out of the active `styles.scss`), so a rule placed in an SCSS partial there never compiles.

```ts
chip: {
    css: `
        .p-chip {
            height: 1.75rem;
            padding: 0 0.5rem;
            font-size: 0.75rem;
        }
    `
}
```

**Why only these three declarations:** `display: inline-flex` and `align-items: center` are already on PrimeNG's base `.p-chip` (text stays centered without re-declaring them), and `min-height` is redundant with the fixed `height`. `padding` zeroes PrimeNG's default vertical padding (sized for taller chips); `font-size` keeps labels compact app-wide. Verified empirically in the browser.

## Changed files

- `theme.config.ts` — the actual change: compact `.p-chip` as the design-system default.
- **Dead-code cleanup** (from PR review): the `p-chip-sm` class no longer has any CSS behind it now that compact is the default, so it was stripped from the relationship-field / select-existing-content / search templates and the `Chip.stories.ts` size knob. The redundant `py-1!` (no-op against the fixed border-box height) was removed from the search filter chip. No behavior change.

## Note on "1.75 rem (28 px)"

dotCMS sets `html { font-size: 14px }`, so **`1.75 rem = 24.5 px`**, not 28 px. The issue's "(28 px)" assumed a 16 px root. Per decision, we honor the **`rem`** value (consistent with the rem-based design system), so chips render at **24.5 px**. If 28 px is required, change `height` to `2rem`.

## Acceptance criteria

- [x] Content Status chip height = 1.75 rem — verified 24.5 px in browser
- [x] Locale chip height = 1.75 rem — verified 24.5 px in browser
- [x] Label text vertically centered, not clipped (PrimeNG base `.p-chip` is `inline-flex` / `align-items: center`)
- [x] Applied in **Content Drive** (status + locale)
- [x] Applied in Edit Contentlet **side panel** (status chip)
- [x] Applied in Edit Contentlet **relationship field**
- [ ] Applied in Edit Contentlet **command bar** — surface not present in codebase (see below)

## Notes for Design / reviewers

- **Global default:** this sizes **every** `<p-chip>` in the app (13 files), not just status/locale — intentional (compact chips as the design-system default). None of the existing chips use image/avatar templates, so the fixed height does not clip. Please confirm the compact size is desired app-wide.
- **rem vs px:** chips are 24.5 px (1.75 rem @ 14 px root), not 28 px — see note above.
- **Command bar:** no command-bar surface rendering status/locale chips exists in the current codebase. Flagging in case it is a future surface to track separately.
- **Side-panel locale:** rendered as `p-button` (a switcher), not a chip — unaffected by chip styling. The side-panel status chip is covered.

## Test plan

- `yarn nx lint ui` / `yarn nx lint edit-content` — pass
- Browser-verified on Content Drive: status + locale chips at 24.5 px, text centered, consistent across rows

This PR fixes: #35891


core-web/libs/ui/src/lib/theme/theme.config.ts
```diff
@@ -42,6 +42,20 @@ export const CustomLaraPreset = definePreset(Lara, {
...
         },
+        chip: {
+            // dotCMS chips are compact by default: 1.75rem (24.5px at the 14px root)
+            // tall, vertically centered, with a small label. Applied to the base
+            // `.p-chip` so every chip (Content Status, locale, relationship, etc.)
+            // gets the size without per-template classes. PrimeNG has no chip size
+            // token, so this is expressed as CSS — same mechanism as card/confirmpopup.
+            css: `
+                .p-chip {
+                    height: 1.75rem;
```
nicobytes: Could we use the tailwind css vars for spacing? The official vs is: var(--spacing)   example:

```css
.rule {
   (4 * 0.25rem = 1rem)
  padding: calc(var(--spacing) * 4); 

  //  (6 * 0.25rem = 1.5rem)
  margin-top: calc(var(--spacing) * 6); 

  //  (3 * 0.25rem = 0.75rem)
  gap: calc(var(--spacing) * 3); 
}

```

nicobytes (dismissed):

adrianjm-dotCMS (approved):

core-web/libs/ui/src/lib/theme/theme.config.ts
oidacra: Done — switched to Tailwind spacing tokens: `calc(var(--spacing) * 7)` for height, `calc(var(--spacing) * 2)` for padding, and `var(--text-xs)` for font-size. Resolves to the same 24.5px, verified in the browser. (accc31360c)

core-web/libs/ui/src/lib/theme/theme.config.ts
```diff
@@ -42,6 +42,20 @@ export const CustomLaraPreset = definePreset(Lara, {
...
         },
+        chip: {
+            // dotCMS chips are compact by default: 1.75rem (24.5px at the 14px root)
+            // tall, vertically centered, with a small label. Applied to the base
+            // `.p-chip` so every chip (Content Status, locale, relationship, etc.)
+            // gets the size without per-template classes. PrimeNG has no chip size
+            // token, so this is expressed as CSS — same mechanism as card/confirmpopup.
+            css: `
+                .p-chip {
+                    height: calc(var(--spacing) * 7); /* 1.75rem */
```
fmontes, line 53: why not just do 1.75rem? if we already dont have the var... all that math whats the win?

fmontes (dismissed):

core-web/libs/ui/src/lib/theme/theme.config.ts
nicobytes: Consistency and avoid magic values. It is the same story with colors; we could put the hex color, but we use vars. And --spacing is the Tailwind CSS vars to achieve that.

Even I noticed that there is a way to avoid the verbose 

--spacing()
Use the --spacing() function to generate a spacing value based on your theme:

```css
/* Input CSS */
.my-element {
  margin: --spacing(4);
}
/*  Compiled CSS */
.my-element {
  margin: calc(var(--spacing) * 4);
}
```

PR #34173, opened 2025-12-26
Refactor(UVE): Real-Time Canvas

fmontes:
## Scope
This summary is based on the branch artifacts available locally in this workspace (not live GitHub PR metadata, which is not accessible from this environment). The changes are centered on **UVE store state refactoring** in `edit-ema`.

## What Changed
The branch refactors the UVE store architecture from a duplicated state model to a single-source-of-truth model.

### 1. New computed-signal access layer (`withPageAsset`)
A new store feature was introduced to expose page asset data through computed signals:
- `$page()`, `$site()`, `$containers()`, `$template()`, `$layout()`
- `$viewAs()`, `$vanityUrl()`, `$urlContentMap()`, `$numberContents()`
- `$clientResponse()`

This centralizes derived state access and reduces direct dependence on internal response structure.

### 2. State/API renaming for cleaner semantics
Several store properties and methods were renamed to remove GraphQL-specific wording and better describe intent:
- `graphqlRequest` -> `requestMetadata`
- `graphqlResponse` -> `pageAssetResponse`
- `legacyGraphqlResponse` -> `legacyResponseFormat`
- `setGraphqlResponse()` -> `setPageAssetResponse()`
- `setGraphqlResponseOptimistic()` -> `setPageAssetResponseOptimistic()`
- `rollbackGraphqlResponse()` -> `rollbackPageAssetResponse()`
- `$graphqlWithParams` -> `$requestWithParams`
- `$customGraphqlResponse` -> `$clientResponse`

### 3. Duplication removed from load/save flows
`withLoad` and `withSave` were simplified by removing repeated manual extraction/sync of `pageAsset` sub-properties.

Instead of repeatedly patching multiple top-level fields, flows now rely on updating `pageAssetResponse` once and reading via computed signals.

### 4. Consumer updates
Related consumers were updated to use the new API naming and signals, including:
- Store features (`withLayout`, etc.)
- Services (for example, action handling)
- Components (for example, editor/style-editor integrations)
- Tests across affected store/features/components

### Architectural improvements
- Moves from **multiple duplicated state fields** to **one authoritative source**.
- Removes manual synchronization points that could drift out of sync.
- Makes API naming implementation-agnostic and easier to maintain.
- Improves type-safety and readability via explicit computed accessors.

### Engineering impact
- Lower maintenance overhead when `pageAsset` shape changes.
- Fewer fragile updates across load/save/consumer code paths.
- Simpler mocking in tests (less setup of duplicated top-level state).

[...]

PR #35927, opened 2026-06-04
fix(uve) #35926: don't reuse the previous page's client GraphQL request on headless navigation

### LINKED ISSUE #35926: UVE dont load the right page data in headless
### Problem Statement

The Universal Visual Editor (UVE) fails to load the correct page content when a user attempts to edit different pages consecutively within the same browser tab. Specifically, if a user opens UVE for 'Page A', then exits UVE, and subsequently tries to open UVE for 'Page B', UVE erroneously displays the content of 'Page A' instead of 'Page B'. This behavior suggests that the content of the first page loaded into UVE persists or is cached incorrectly, preventing subsequent pages from loading properly. This issue is critical as it completely prevents users from editing any page other than the very first one they open in UVE in a given tab session. It impacts all users who need to edit multiple pages on the platform. This is a front-end issue, likely affecting all common web browsers (e.g., Chrome, Firefox, Safari, Edge) across any operating system.

### Steps to Reproduce

1.  Open your web browser and navigate to the platform where UVE is integrated.
2.  Access the UVE to edit 'Page A' (e.g., by clicking an 'Edit with UVE' button associated with Page A).
3.  Verify that 'Page A' content loads and is displayed correctly within UVE.
4.  Exit or close UVE (e.g., by clicking a 'Save & Exit' or 'Close Editor' button, or navigating away from the UVE interface).
5.  In the *same browser tab*, navigate to access UVE for a different 'Page B'.
6.  **Observe:** UVE displays the content of 'Page A' instead of 'Page B'.

**Does NOT occur:**
*   If UVE is accessed directly for a page via its URL.
*   If internal navigation occurs (e.g., clicking a link within the page loaded inside UVE's iframe).

### Acceptance Criteria

When a user opens UVE for any specific page, the content of that exact page must be displayed correctly and reflect the current page being edited, even if the user has previously opened and closed UVE for a different page within the same browser tab. The editor must not display stale or cached content from a prior UVE session.

### dotCMS Version

latest

### Severity

### Links

-

KevinDavilaDotCMS:
https://github.com/user-attachments/assets/f1d6a520-53a4-4c7b-9dc1-a94e78383794

This PR fixes #35926

### Problem

In headless mode, opening "Page One" in UVE and then navigating to "Page Two" could fail: the fetch for Page Two errored out and the editor bounced back to Page One, getting stuck on it. Traditional pages were not affected.

### Root cause

When a headless app boots inside UVE, the SDK sends `CLIENT_READY` with the page's GraphQL request (query + **variables**: `url`, `languageId`, `publishDate`, `siteId`, custom vars, etc.). UVE stores it as `requestMetadata` and prefers it for every subsequent page fetch.

Two recent refactors broke the lifecycle of that stored request:

1. **#34173 (Real-Time Canvas)** renamed `graphql` → `requestMetadata`. The old `resetClientConfiguration()` reset the whole client state (including the stored query); the new implementation no longer cleared it.
2. **#35539 (explicit iframe sizing)** changed `pageLoad` to call `markPageLoading()` instead of `resetClientConfiguration()` (intentionally, to keep the editor chrome mounted during navigation). `markPageLoading()` only resets `isClientReady` + history — `requestMetadata` survives every navigation.

On top of that, the `CLIENT_READY` handler early-returned when `isClientReady` was already `true`. During a client-side route change inside the headless app, the SDK sends `CLIENT_READY` (for the new page) **before** `NAVIGATION_UPDATE` — so the new page's request was **discarded** while the editor was still marked ready for the previous page.

**Failure chain:** navigate to Page Two → `pageLoad` fetches it via GraphQL using **Page One's stored query/variables** → the query can resolve to `page: null` (stale variables) → `graphqlToPageEntity` returns `null` → `const { vanityUrl } = pageAsset` throws → `uveStatus: ERROR` with no HTTP status (no error UI). Since `markPageLoading()` keeps the previous `pageAssetResponse`, `$iframeURL` still derives from **Page One** → the iframe reloads Page One → the app re-announces `updateNavigation('/page-one')` → UVE loads Page One again. That's the "falls back to the last page" behavior.

### Proposed Changes

* **`dot-uve-actions-handler.service.ts` (`CLIENT_READY`)** — the stored client request is now refreshed on every `CLIENT_READY` (it is page-scoped data). When the client is already ready, only the config is updated — no `pageReload` is fired, so no extra request happens. This means that by the time a client-side navigation's `NAVIGATION_UPDATE` triggers `pageLoad`, the store already holds the **target page's** query/variables and the fetch goes through GraphQL with the correct config.
* **`withPageApi.ts` (`pageLoad`)** — before fetching, the store now checks whether the stored request **belongs to the target page** by comparing the request's own `variables.url` against the navigation target (`compareUrlPaths`, so leading-slash variants match). Requests without a `url` variable are assumed to belong to the currently loaded page (preserves legacy-client behavior).
  * Belongs to the target → GraphQL fetch, same as today.
  * Belongs to another page (e.g. navigation from the toolbar / Pages portlet, where no `CLIENT_READY` for the target exists yet) → the stale request is dropped and the page loads through the standard Page API — which is exactly the first-load flow — until the page's own `CLIENT_READY` installs fresh metadata.
* **`withPage.ts`** — new `resetRequestMetadata()` method (drops only the stored client request), wired into `withPageApi` deps in `dot-uve.store.ts`.

### What deliberately did NOT change

| Flow | Behavior |
|---|---|
| Same-page param changes (language / persona / mode) | Unchanged — GraphQL with the stored request, as today |
| Page reload / save / style editor | Unchanged |
| Traditional pages | Unchanged (they never use `requestMetadata`) |
| Request count on navigation | Unchanged — same number of editor-side fetches as today, only the *config* used is corrected |
| `resetClientConfiguration()` | Untouched (currently has no production callers; left out of scope on purpose) |

**Known constraint (legacy clients):** a custom GraphQL client that omits the `url` variable from its `CLIENT_READY` request is assumed to belong to the page currently loaded in the editor. On cross-page navigation its stored request is dropped and the page loads through the standard Page API until the client re-announces itself — this matches the pre-#34173 behavior, where the stored request was reset on every `pageLoad`. The SDK (`@dotcms/client`) always sends `url`, so this only affects hand-rolled clients.

### Checklist
- [x] Tests
- [ ] Translations (N/A — no user-facing text)
- [x] Security Implications Contemplated (no new attack surface: `CLIENT_READY` could already overwrite the stored request before the editor was ready; messages still come from the same UVE iframe channel)

### Additional Info

**Tests added:**
* `withPageApi.spec.ts`: navigating to a different page drops the stale request and uses the Page API; a request captured *for* the target page is kept (client-side navigation); same-page param changes keep the request; a fresh load with no previous `pageParams` drops stale metadata surviving from a prior visit (the store lives at the route level and outlives the shell).
* `dot-uve-actions-handler.service.spec.ts`: new `CLIENT_READY` suite — installs config + reloads when not ready; refreshes config **without** reloading when already ready; no-ops without a graphql query.
* `withPage.spec.ts`: `resetRequestMetadata()` drops only the stored request.

Full `portlets-edit-ema-portlet` suite: 70 suites / 1418 passing. Lint clean.

**Assumption worth knowing:** the client-side navigation path relies on the SDK sending `CLIENT_READY` before `NAVIGATION_UPDATE` (both are emitted in that order from the same effect in `useEditableDotCMSPage` / `DotCMSEditablePageService.listen`). If that order ever changes, behavior degrades safely to the Page API path (first-load flow), not to a failure.

**Follow-ups (separate issues recommended):**
1. Make the `CLIENT_READY` payload carry an explicit page identity (typed in `@dotcms/types`) instead of inferring ownership from `variables.url` — removes the implicit contract between the editor store and the SDK's GraphQL variable naming.
2. `@dotcms/client` regression in `page.get()`: `UVE_MODE[mode]` returns `undefined` when apps forward the raw `mode=EDIT_MODE` URL param (string enums have no reverse mapping), silently falling back to LIVE — unpublished pages 404 inside the editor.
3. No E2E covers "navigate between two headless pages in UVE" — both refactors that introduced this regression passed CI.

This PR fixes: #35926

### DIFF
--- a/core-web/libs/portlets/edit-ema/portlet/src/lib/services/dot-uve-actions-handler/dot-uve-actions-handler.service.spec.ts
+++ b/core-web/libs/portlets/edit-ema/portlet/src/lib/services/dot-uve-actions-handler/dot-uve-actions-handler.service.spec.ts
@@ -536,3 +536,99 @@ describe('DotUveActionsHandlerService – REGISTER_STYLE_SCHEMAS', () => {
         expect(setStyleSchemas).toHaveBeenCalledWith(mockSchemas);
     });
 });
+
+describe('DotUveActionsHandlerService – CLIENT_READY', () => {
+    let spectator: SpectatorService<DotUveActionsHandlerService>;
+    let service: DotUveActionsHandlerService;
+
+    const createService = createServiceFactory({
+        service: DotUveActionsHandlerService,
+        providers: [
+            mockProvider(DotWorkflowActionsFireService),
+            mockProvider(DotMessageService),
+            mockProvider(MessageService),
+            mockProvider(DotCopyContentModalService),
+            {
+                provide: UVEStore,
+                useValue: buildMockStore()
+            }
+        ]
+    });
+
+    const buildClientReadyStore = (isClientReady: boolean) => ({
+        ...buildMockStore(),
+        isClientReady: jest.fn().mockReturnValue(isClientReady),
+        setCustomClient: jest.fn(),
+        setIsClientReady: jest.fn()
+    });
+
+    const CLIENT_READY_PAYLOAD = {
+        graphql: {
+            query: '{ page { url } }',
+            variables: { url: '/page-two', depth: '1' }
+        }
+    };
+
+    const buildDeps = (mockStore: ReturnType<typeof buildClientReadyStore>) => ({
+        uveStore: mockStore as unknown as InstanceType<typeof UVEStore>,
+        dialog: null,
+        inlineEditingService: null,
+        contentWindow: null,
+        host: 'http://localhost',
+        onCopyContent: jest.fn()
+    });
+
+    beforeEach(() => {
+        jest.clearAllMocks();
+        spectator = createService();
+        service = spectator.service;
+    });
+
+    it('should install the client request and reload when the client is not ready', () => {
+        const mockStore = buildClientReadyStore(false);
+
+        service.handleAction(
+            { action: DotCMSUVEAction.CLIENT_READY, payload: CLIENT_READY_PAYLOAD },
+            buildDeps(mockStore)
+        );
+
+        expect(mockStore.setCustomClient).toHaveBeenCalledWith({
+            query: CLIENT_READY_PAYLOAD.graphql.query,
+            variables: CLIENT_READY_PAYLOAD.graphql.variables
+        });
+        expect(mockStore.pageReload).toHaveBeenCalled();
+        expect(mockStore.setIsClientReady).toHaveBeenCalledWith(true);
+    });
+
+    it('should refresh the stored client request without reloading when already ready', () => {
+        // Client-side navigation: the new page's CLIENT_READY arrives while the
+        // editor is still initialized for the previous page. The config must be
+        // refreshed (so the upcoming NAVIGATION_UPDATE pageLoad uses the new
+        // page's query/variables) but no reload should fire.
+        const mockStore = buildClientReadyStore(true);
+
+        service.handleAction(
+            { action: DotCMSUVEAction.CLIENT_READY, payload: CLIENT_READY_PAYLOAD },
+            buildDeps(mockStore)
+        );
+
+        expect(mockStore.setCustomClient).toHaveBeenCalledWith({
+            query: CLIENT_READY_PAYLOAD.graphql.query,
+            variables: CLIENT_READY_PAYLOAD.graphql.variables
+        });
+        expect(mockStore.pageReload).not.toHaveBeenCalled();
+        expect(mockStore.setIsClientReady).not.toHaveBeenCalled();
+    });
+
+    it('should not install a client request when the payload has no graphql query', () => {
+        const mockStore = buildClientReadyStore(true);
+
+        service.handleAction(
+            { action: DotCMSUVEAction.CLIENT_READY, payload: {} },
+            buildDeps(mockStore)
+        );
+
+        expect(mockStore.setCustomClient).not.toHaveBeenCalled();
+        expect(mockStore.pageReload).not.toHaveBeenCalled();
+    });
+});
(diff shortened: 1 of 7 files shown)


zJaaal (approved):

KevinDavilaDotCMS:
Triaged the automated review findings — here's the disposition:

**Addressed in `7273a59` + PR description update:**

- **#5 (missing test for `NAVIGATION_UPDATE` before `CLIENT_READY`)** — Partially valid: the fallback branch was already covered, but the branch where the stored request *explicitly* declares another page's `url` wasn't. Added `should drop a stored request explicitly captured for another page (NAVIGATION_UPDATE before CLIENT_READY)`, asserting the request is dropped and the page loads through the standard Page API (first-load flow).
- **#6 (fixture intent unclear)** — Renamed the shared fixture to `graphqlRequestWithoutUrl` with a comment stating it exercises the legacy-client fallback branch.
- **#3 (behavior change for legacy clients without a `url` variable)** — Intentional, now documented as a **Known constraint** in the PR description: clients omitting `url` get their stored request dropped on cross-page navigation and fall back to the Page API until they re-announce — which matches the pre-#34173 behavior. The SDK always sends `url`, so only hand-rolled clients are affected.

**Acknowledged, deliberately not changed here:**

- **#1 (message-ordering contract)** — Already documented in the description ("Assumption worth knowing"); degradation is safe (Page API path). A runtime warning would fire on legitimate flows (toolbar / Pages-portlet navigation), so it would be noise. The real fix is the typed page identity in the `CLIENT_READY` payload — already listed as follow-up 1.
- **#2 (`compareUrlPaths` uses `window.origin`)** — Pre-existing and not reachable outside the browser: this portlet only runs inside dotAdmin, and the `NAVIGATION_UPDATE` handler already called `compareUrlPaths` before this PR.
- **#4 (`setIsClientReady(true)` after `pageReload`)** — Pre-existing pattern this PR didn't introduce (the lines only moved in the diff). Reordering it touches a validated flow with no observed defect; better as a follow-up if we harden the handler.

Full portlet suite: 70 suites / 1419 passing.

dario-daza (approved):

PR #35990, opened 2026-06-04
feat(image): optional libvips (vips-ffm) image engine behind a feature flag (#35989)

### LINKED ISSUE #35989: feat(image): optional libvips (vips-ffm) image processing engine behind a feature flag
## Problem / Opportunity

dotCMS image processing (`com.dotmarketing.image.filter`) is a pure-JVM stack (TwelveMonkeys, PDFBox, Batik, webp-imageio, jhlabs). It works, but:

- The `ImageFilterAPI` Javadoc itself warns that resizing large images "can cause garbage collections and OOM exceptions" because the whole image is decompressed into heap.
- No modern output formats (AVIF/JXL), no content-aware crop, per-filter temp-file round-trips in multi-filter chains.

[libvips](https://www.libvips.org/) is a streaming, demand-driven image library that is dramatically faster and lower-memory, and — via the [vips-ffm](https://github.com/lopcode/vips-ffm) Panama FFM bindings — is a natural fit now that dotCMS runs on **Java 25**.

## Proposal

Add a **feature-flagged, drop-in alternative** image engine backed by libvips, alongside the existing pure-JVM engine (which stays the default and is untouched).

- Flag `IMAGE_API_USE_LIBVIPS` (default `false`) selects the engine; `IMAGE_API_LIBVIPS_FALLBACK` (default `true`) degrades to the legacy engine if a native op fails (corrupt input, missing delegate).
- Every libvips filter subclasses the legacy `ImageFilter`, reusing its caching, unique-naming and `overwrite()` logic, so the **URL parameter contract is identical** (`crop_w`, `resize_w`, `filter=resize,jpeg`, …). Only the pixels change.

## Scope

- [x] 16 libvips filters at parity (crop, resize, scale, thumbnail, rotate, flip, gamma, grayscale, hsb, exposure, subsample, jpeg, png, webp, gif, pdf)
- [x] `VipsImageFilterApiImpl implements ImageFilterAPI`; engine selected via `ImageEngine.resolve()` in the exporter, metadata generator and binary view tool
- [x] Graceful fallback to the legacy engine on libvips failure
- [x] New capabilities: content-aware **smartcrop**, **AVIF** output
- [x] `libvips42` added to the Ubuntu 24.04 runtime Dockerfile (bundles pdf/heif/svg/webp/jxl delegates — no custom build needed)
- [x] CI installs libvips on the JVM unit-test runner so parity tests run
- [x] Tests: `VipsParityTest` + `ImageFilterExporterEngineSelectionTest` (30 green; legacy suite unchanged)

## Parity definition

Not pixel-identical (different engines). Parity = deterministic geometry exact (shared `ResizeCalc`), encoding correct (container/alpha/animation preserved), and visual delta under threshold vs the legacy output. `exposure` and `hsb` are documented approximations.

## Out of scope / follow-ups

- Pixel-exact `exposure`/`hsb` curves
- Single-pass `VipsPipeline` to collapse multi-filter chains
- BlurHash / dominant-colour derived metadata
- Animated WebP/AVIF transcode, ICC handling

See `dotCMS/src/main/java/com/dotmarketing/image/vips/README.md` and `NEW-FEATURES.md` for details.

wezell:
## What

Adds an optional, **feature-flagged** image processing engine backed by [libvips](https://www.libvips.org/) via the [vips-ffm](https://github.com/lopcode/vips-ffm) (Panama FFM) bindings, alongside the existing pure-JVM engine. The legacy engine remains the default and is untouched.

Fixes #35989

## Why

- `ImageFilterAPI`'s own Javadoc warns that resizing large images "can cause garbage collections and OOM exceptions" — libvips is streaming/demand-driven and sidesteps this.
- Unlocks modern formats (AVIF) and content-aware crop the AWT stack can't do.
- Natural fit now that dotCMS runs on Java 25 (FFM, no JNI).

## How it works

- **Feature flag** `IMAGE_API_USE_LIBVIPS` (default `false`). `ImageEngine.resolve()` is the single selection point, used by `ImageFilterExporter`, `MetadataGeneratorImpl` and `BinaryMap`.
- **Identical URL contract:** every libvips filter subclasses the legacy `ImageFilter`, reusing its caching, unique-naming and `overwrite()` logic. Only the pixels differ (libvips lanczos3 vs jhlabs/TwelveMonkeys), so cached renditions regenerate once when the flag flips.
- **Graceful fallback** (`IMAGE_API_LIBVIPS_FALLBACK`, default `true`): if a native op fails (corrupt input, missing delegate) the equivalent legacy filter runs instead.
- **Native dep:** `libvips42` added to the Ubuntu 24.04 runtime Dockerfile — the stock package bundles the pdf/heif(avif)/svg/webp/jxl delegates (verified in a clean `ubuntu:24.04` container), so no custom libvips build is required.

## Filters

16 at parity (crop, resize, scale, thumbnail, rotate, flip, gamma, grayscale, hsb, exposure, subsample, jpeg, png, webp, gif, pdf) + two libvips-only capabilities: **smartcrop** (attention-based) and **AVIF** output.

## Tests

- `VipsParityTest` (18) — geometry exact (shared `ResizeCalc`), encoders valid, animated GIF preserved through the production `runFilter` path, resize visual luma-delta < 12 vs legacy, forced-failure fallback, smartcrop, AVIF magic bytes.
- `ImageFilterExporterEngineSelectionTest` (3) — flag off → legacy, flag on → libvips, identical filter-key contract.
- Legacy `ImageFilterAPIImplTest` unchanged (9). **30 green total.**
- CI installs `libvips42` on the JVM unit-test runner so the parity tests run rather than self-skip.

## Parity definition

Not pixel-identical (different engines). Parity = deterministic geometry exact + encoding correct (container/alpha/animation) + visual delta under threshold. `exposure` and `hsb` are documented approximations.

## Risk / rollout

Off by default; zero behavioural change until `IMAGE_API_USE_LIBVIPS=true`. When on, requests fall back to the legacy engine on any libvips failure. Docs: `dotCMS/.../image/vips/README.md` and `NEW-FEATURES.md`.

This PR fixes: #35989

### DIFF
--- a/.github/workflows/cicd_comp_test-phase.yml
+++ b/.github/workflows/cicd_comp_test-phase.yml
@@ -189,7 +189,14 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
-          
+
+      # The libvips image engine (IMAGE_API_USE_LIBVIPS) is exercised by VipsParityTest.
+      # Install native libvips on the JVM unit-test runner so those tests run instead of
+      # self-skipping. Ubuntu 24.04 libvips42 bundles the pdf/heif/svg/webp/jxl delegates.
+      - name: Install native libvips (image engine parity tests)
+        if: matrix.test_type == 'jvm'
+        run: sudo apt-get update && sudo apt-get install -y --no-install-recommends libvips42
+
       - name: Run ${{ matrix.name }}
         uses: ./.github/actions/core-cicd/maven-job
         with:
--- a/.github/workflows/cicd_comp_test-phase.yml
+++ b/.github/workflows/cicd_comp_test-phase.yml
@@ -195,7 +195,7 @@ jobs:
       # self-skipping. Ubuntu 24.04 libvips42 bundles the pdf/heif/svg/webp/jxl delegates.
       - name: Install native libvips (image engine parity tests)
         if: matrix.test_type == 'jvm'
-        run: sudo apt-get update && sudo apt-get install -y --no-install-recommends libvips42
+        run: sudo apt-get update && sudo apt-get install -y --no-install-recommends libvips42 libheif-plugin-aomenc
 
       - name: Run ${{ matrix.name }}
         uses: ./.github/actions/core-cicd/maven-job
--- a/bom/application/pom.xml
+++ b/bom/application/pom.xml
@@ -17,6 +17,7 @@
         <aws-java-sdk.version>1.12.488</aws-java-sdk.version>
         <aws-java-sdk-v2.version>2.34.0</aws-java-sdk-v2.version>
         <twelvemonkeys.version>3.13.1-fixed</twelvemonkeys.version>
+        <vips-ffm.version>1.9.8</vips-ffm.version>
         <swagger.version>2.2.34</swagger.version>
         <bytebuddy.version>1.18.1</bytebuddy.version>
         <batik.version>1.17</batik.version>
@@ -1101,6 +1102,12 @@
                 <groupId>com.github.gotson</groupId>
                 <artifactId>webp-imageio</artifactId>
                 <version>0.2.2</version>
+            </dependency>
+            <!-- libvips bindings (FFM, JDK 23+) — alternative image engine, requires native libvips on host -->
+            <dependency>
+                <groupId>app.photofox.vips-ffm</groupId>
+                <artifactId>vips-ffm-core</artifactId>
+                <version>${vips-ffm.version}</version>
             </dependency>
             <dependency>
                 <groupId>com.twelvemonkeys.imageio</groupId>
--- a/dotCMS/pom.xml
+++ b/dotCMS/pom.xml
@@ -931,6 +931,11 @@
             <groupId>com.github.gotson</groupId>
             <artifactId>webp-imageio</artifactId>
         </dependency>
+        <!-- libvips bindings (FFM) — alternative image engine; native libvips must be present on host -->
+        <dependency>
+            <groupId>app.photofox.vips-ffm</groupId>
+            <artifactId>vips-ffm-core</artifactId>
+        </dependency>
         <dependency>
             <groupId>com.twelvemonkeys.imageio</groupId>
             <artifactId>imageio-batik</artifactId>
--- a/dotCMS/src/main/docker/original/Dockerfile
+++ b/dotCMS/src/main/docker/original/Dockerfile
@@ -59,13 +59,16 @@ RUN apt update && \
         libarchive-tools \
         libpq-dev \
         libvips42 \
+        libheif-plugin-aomenc \
         apt-transport-https && \
     rm -rf /var/lib/apt/lists/*
 
 # libvips42 (Ubuntu 24.04 = libvips 8.15.1) backs the optional libvips image engine
-# (IMAGE_API_USE_LIBVIPS=true). The stock package bundles the PDF (poppler), AVIF/HEIC
-# (libheif), SVG (rsvg), WebP and JPEG-XL delegates, so no custom build is needed.
-# The engine is off by default; when off this library is simply unused.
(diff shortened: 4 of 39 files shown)


wezell:
Thanks for the review — addressed in `a8313c3f5`:

1. **Crop out-of-bounds early-exit** — intentional. Writing `src` to the cache populates a valid (full-image) rendition that matches the requested-but-degenerate crop; the visual result is identical and it keeps the cache consistent with the rest of the libvips path. Left as-is.
2. **Thumbnail `bg` parse** — now validates exactly 9 digits and defaults to white on bad input.
3. **SubSample zero-dimension** — guarded; with no target it re-encodes the source instead of asking libvips for a 0-px thumbnail.
4. **`resizeImage(File)` no fallback** — now wrapped with the same `IMAGE_API_LIBVIPS_FALLBACK` degrade-to-legacy as `getWidthHeight()`.
5. **Hsb unbounded** — `h/s/b` clamped to the documented `-1.0..1.0`.
6. **Rotate background** — documented the default (transparent w/ alpha, else black) and added an optional `rotate_bg` 9-digit fill.
7. **`isAvailable()` memoizes `false`** — intentional: the probe runs on first image request (well after native libs are ready), libvips availability doesn't change at runtime, and the alternative (re-probing on every miss) would repeatedly pay the failure cost. Left memoized.

Separately, CI surfaced that AVIF *encode* needs libheif's AV1 encoder, which `libvips42 --no-install-recommends` drops — added `libheif-plugin-aomenc` to the Dockerfile and the CI step (verified in a clean `ubuntu:24.04` container), and the AVIF test now self-skips where that encoder is absent.

wezell:
Round 2 — addressed in `e1aba9fb1`:

**Fixed (real issues):**
- **#6 HSB crash on RGBA** — split alpha off before `sRGB2HSV` and re-`bandjoin` it after, so the 3-band `linear()` no longer throws on 4-band input. Added a test that feeds a genuine RGBA PNG and asserts alpha is preserved.
- **#10 `subsampleImage` no fallback** — now degrades to legacy on libvips failure, consistent with `getWidthHeight`/`resizeImage`.
- **#9 PDF output format** — added an explicit `getResultsFile → png` override. (Note: the base `FILE_EXT` is already `"png"`, so output was PNG, identical to the legacy `PDFImageFilter` which also has no source-derived extension — the override just makes it unmistakable.) Added a PDF-render test that asserts the `runFilter` result is `.png`, not `.pdf`.

**Intentional / parity (left as-is):**
- **#1, #2 Crop edge clamp + focal-point math** — these are verbatim ports of `CropImageFilter` (`width = current.width - x - 1` and the focal half-width handling are exactly the legacy logic). Changing them would diverge from the parity target; I'd rather fix both engines together in a follow-up if the off-by-one is real.
- **#3 `toBufferedImage` temp leak** — the temp is already deleted in a `finally`, so it's cleaned up on `ImageIO.read` failure.
- **#4 cross-filesystem rename** — the temp file is created in the *same directory* as the result (`resultFile + "_" + nanos + ext`), so the rename is always intra-filesystem; this mirrors the legacy filters.
- **#5 exposure negative** — exposure is a documented linear-gain approximation; `<= 0` is treated as a no-op rather than the legacy's degenerate all-black at `0`. Noted in the Javadoc.
- **#7, #8, nits** — acknowledged; no behavioural change needed.

Tests now 32 green locally (incl. the new RGBA-HSB and PDF-render cases).

sfreudenthaler:
@wezell  nit. I dont like the feature flag name IMAGE_API_USE_LIBVIPS because it's too literal about the tech. The feature flag name should be more about the feature unlocked. 

In this case maybe image processor v2?  It will make life easier for the outbound marketing and support teams. 

Trade off is we lose the search ability through the envvars but it's be easy for the agents to trace it and humans could find the functional code based on the name.

dsilvam (approved):

PR #36036, opened 2026-06-06
fix(cache): harden lettuce Redis cache — cluster-scope, cache-stats, and codec defects (#36035)

### LINKED ISSUE #36035: Lettuce Redis cache: cross-cluster cache wipe, missing/blank cache-stats region, and related defects
I had RedisCache set up by default in the temu cluster and found:

- there was no cache reporting in the backend
- A full cache flush did not work - this is because it was actually trying to run a full flushall on redis without prefixing which our per-tenant namespaces did not allow. Yea! Redis permissions and isolation worked!

I asked Claudio to fix (and simplify all the unnecessary abstractions) and here is what we got:

----

## Problem

A deep review of the `com.dotcms.cache.lettuce` package surfaced several defects in the Lettuce-based Redis cache (`RedisCache` / `MasterReplicaLettuceClient`). The most serious break the package's own stated design goal (per-cluster key isolation in a shared Redis space) and the cache-stats maintenance screen.

### Defects

1. **Cross-cluster cache wipe (HIGH).** `RedisCache.remove(group)` and `removeAll()` build `*PREFIX*` match patterns **without the cluster id**. In a shared Redis space (which the clusterId prefix exists specifically to support), these patterns match and delete **other clusters'/tenants' keys**. `keyCount`/`scanKeys` are correctly cluster-scoped, so the codebase is internally inconsistent.

2. **`deleteFromPattern` is fragile (HIGH).** It runs `redis.call('del', unpack(redis.call('keys', '<pattern>')))` via `eval`:
   - errors out (`wrong number of arguments for 'del'`) when the pattern matches nothing — and the error is swallowed (async, result discarded), so the delete silently no-ops;
   - uses blocking `KEYS` (O(N), blocks the Redis event loop);
   - is vulnerable to Lua string injection via the interpolated pattern.

3. **Redis region missing / blank on the cache-stats screen (HIGH).** `RedisCache.getStats()` / `keyCount()` can throw on a null pooled connection (`conn.isOpen()` on a `null` borrow). `CacheProviderAPIImpl.getStats()` **drops any provider whose `getStats()` throws**, so the Redis region disappears from the cache-stats maintenance screen entirely. When the group set is empty it also produced an empty/blank table (no fallback row like `CaffineCache` has).

4. **`delete(K...)` never matches stored keys (MED).** The varargs `delete` converts keys with `keyToStringConverter` instead of `wrapKey`, so it omits the cluster prefix and never matches what was stored (contrast the correct single-key `delete` and `deleteNonBlocking`).

5. **Pub/sub connection leak + non-thread-safe subscriber lists (MED).** The pub/sub connection is leaked when it cannot be opened; subscriber lists are plain `ArrayList`s mutated/iterated across concurrent subscribe/unsubscribe.

6. **Misc:** `DotObjectCodec.decodeValue` uses `bytes.array()` (ignores buffer position/limit/offset); `unwrapKey`/`getKeys` use a global `String.replace` instead of a leading-anchored strip; sync `set(k,v,ttl)` doesn't guard `ttl == -1`; `RedisClientFactory` swallows reflection failures and uses deprecated `newInstance()`; duplicate/dead code (`removeKeysRaw`, dead `getStats` INFO block).

## Scope / constraint

The `RedisClient` interface is **kept intact** (generics + all methods); genuinely-unused methods are `@Deprecated` rather than removed.

## Notes

- After the fix, `remove(group)` / `removeAll()` become **synchronous** (SCAN + UNLINK) instead of async fire-and-forget, matching the Caffeine/H22 providers. The one production caller is `RedisStoragePersistenceAPI.remove(...)`.

wezell:
Fixes #36035

I had RedisCache set up by default in the a cluster and found:
1) there was no Redis cache reporting in the backend
2) A full cache flush did not work - this is because it was actually trying to run a full flushall on redis without prefixing which our per-tenant namespaces did not allow.  Yea! Redis permissions and isolation worked!

I asked Claudio to fix (and simplify all the unnecessary abstractions) and here is what we got:

----------------------

## What

Hardens the Lettuce Redis cache (`com.dotcms.cache.lettuce`). Bug fixes only — the `RedisClient` interface is kept intact (generics + all methods); genuinely-unused methods are `@Deprecated` rather than removed.

## Fixes

- **Cross-cluster cache wipe (HIGH):** `RedisCache.remove(group)`/`removeAll()` built `*PREFIX*` patterns without the cluster id, matching/deleting other clusters' keys in a shared Redis space. `deleteFromPattern` now prepends `keyPrefix()` (consistent with `scanKeys`/`keyCount`); callers pass un-anchored patterns.
- **`deleteFromPattern` (HIGH):** replaced the Lua `KEYS`+`DEL` eval (blocking `KEYS`, errors on empty match, Lua-injectable) with `SCAN` + `UNLINK`.
- **Cache-stats region missing/blank (HIGH):** `getStats()`/`keyCount()` could throw on a null pooled connection; `CacheProviderAPIImpl.getStats()` drops any provider whose `getStats()` throws, so the Redis region vanished from the cache-stats screen. Both are now exception-proof, plus a `CaffineCache`-style empty-groups "n/a" fallback row.
- **`delete(K...)` (MED):** used `keyToStringConverter` instead of `wrapKey`, so it never matched stored keys. Now wraps like `deleteNonBlocking`.
- **Pub/sub (MED):** fixed connection leak when the connection can't be opened; subscriber lists are now `CopyOnWriteArrayList` for safe concurrent subscribe/unsubscribe.
- **Misc:** `DotObjectCodec.decodeValue` now reads `remaining()` into a copy instead of `bytes.array()` (which ignores position/limit/offset); `Charset.forName` → `StandardCharsets.UTF_8`; `unwrapKey`/`getKeys` strip only the leading prefix; sync `set(k,v,ttl)` guards `ttl == -1`; `RedisClientFactory` de-duplicated, `getDeclaredConstructor().newInstance()`, logs reflection failures; removed duplicate `removeKeysRaw` and the dead `getStats` INFO/`getRedisProperties` block; `NullLettuceClient` returns safe non-null defaults.

## Behavior notes for reviewers

- `remove(group)`/`removeAll()` are now **synchronous** (`SCAN`+`UNLINK`) rather than async fire-and-forget — this matches the Caffeine/H22 providers and avoids stale-read-after-remove. The one production caller is `RedisStoragePersistenceAPI.remove(...)`.
- `getClient(name, codec)` still ignores `codec` when reflection succeeds — pre-existing behavior, deliberately preserved.

## Testing

- `dotcms-core` compiles clean (`BUILD SUCCESS`, no warnings from the changed files).
- `DotObjectCodec` round-trip validated standalone against the built classes (encode/decode of key & value, `null` → `DecodeException`, and a **sliced-buffer** case that exercises the `bytes.array()` fix). The integration `DotObjectCodecTest` is Redis/Docker-gated and was not run in this environment.

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/cache/lettuce/DotObjectCodec.java
+++ b/dotCMS/src/main/java/com/dotcms/cache/lettuce/DotObjectCodec.java
@@ -6,7 +6,7 @@ import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
 import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
-import com.dotmarketing.exception.DotRuntimeException;
+import java.nio.charset.StandardCharsets;
 import io.lettuce.core.codec.RedisCodec;
 
 /**
@@ -17,7 +17,7 @@ import io.lettuce.core.codec.RedisCodec;
  */
 public class DotObjectCodec<K,V> implements RedisCodec<K, V> {
 
-    private final Charset charset = Charset.forName("UTF-8");
+    private final Charset charset = StandardCharsets.UTF_8;
 
     @Override
     public K decodeKey(final ByteBuffer bytes) {
@@ -28,11 +28,18 @@ public class DotObjectCodec<K,V> implements RedisCodec<K, V> {
     @Override
     public V decodeValue(final ByteBuffer bytes) {
 
-        try (ObjectInputStream in = new ObjectInputStream(new ByteArrayInputStream(bytes.array()))){
-        
-            final Object o = in.readObject();
-            return  (V)o;
-        } catch(Exception e) {
+        try {
+            // copy exactly the readable region: bytes.array() ignores position/limit/arrayOffset and can
+            // read stale bytes when the buffer is a slice. (null buffer -> NPE -> DecodeException, as before.)
+            final byte[] data = new byte[bytes.remaining()];
+            bytes.get(data);
+
+            try (ObjectInputStream in = new ObjectInputStream(new ByteArrayInputStream(data))) {
+
+                final Object o = in.readObject();
+                return (V) o;
+            }
+        } catch (Exception e) {
 
             throw new DecodeException(e);
         }
--- a/dotCMS/src/main/java/com/dotcms/cache/lettuce/MasterReplicaLettuceClient.java
+++ b/dotCMS/src/main/java/com/dotcms/cache/lettuce/MasterReplicaLettuceClient.java
@@ -935,6 +935,12 @@ public class MasterReplicaLettuceClient<K, V> implements RedisClient<K, V> {
 
         // Cache invalidation must not throw into callers (matches the other CacheProviders); a mid-scan
         // connection failure is logged and swallowed rather than propagated through remove()/removeAll().
+        //
+        // NOTE: a single pooled connection is held for the whole SCAN+UNLINK iteration. For a very large
+        // match (e.g. removeAll() over millions of keys) this occupies one pool slot for the duration. We
+        // deliberately do NOT release/reborrow between batches: under master-replica a SCAN cursor is bound
+        // to the node it started on and is not portable across connections/nodes. removeAll() is a rare
+        // full-flush operation and the default pool leaves other slots free for concurrent get/put/stats.
         try (StatefulRedisConnection<String, V> conn = this.getConn()) {
 
             if (this.isOpen(conn)) {
--- a/dotCMS/src/main/java/com/dotcms/cache/lettuce/MasterReplicaLettuceClient.java
+++ b/dotCMS/src/main/java/com/dotcms/cache/lettuce/MasterReplicaLettuceClient.java
@@ -913,12 +913,16 @@ public class MasterReplicaLettuceClient<K, V> implements RedisClient<K, V> {
 
         if (this.channelReferenceMap.containsKey(channel)) {
 
-            final List<String> channelUUIDList = channelReferenceMap.get(channel);
+            final List<String> channelUUIDList =
+                    channelReferenceMap.getOrDefault(channel, Collections.emptyList());
             for (final String channelUUID : channelUUIDList) {
                 final Tuple2<DotPubSubListener, StatefulRedisPubSubConnection<String, V>> channelRedisPubSubAsyncCommandsTuple =
                         channelStatefulRedisPubSubConnectionMap.get(channelUUID);
 
-                subscribers.add(channelRedisPubSubAsyncCommandsTuple._1());
+                // can be null if a concurrent unsubscribe removed the entry between the list snapshot and lookup
+                if (null != channelRedisPubSubAsyncCommandsTuple) {
+                    subscribers.add(channelRedisPubSubAsyncCommandsTuple._1());
+                }
             }
         }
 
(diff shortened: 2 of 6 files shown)


wezell:
Thanks for the review — addressed the substantive findings in `0e2e04455e`:

- **#1 / #2 / #3 (`keyCount` blocking `KEYS`, Lua injection, unchecked cast):** `keyCount()` now counts via the cluster-scoped `scanKeys` (SCAN) instead of the Lua `KEYS` eval. This finishes the `deleteFromPattern` fix's intent and removes both the `MasterReplicaLettuceClient` downcast and the Lua string-injection point in one go.
- **#4 (N× blocking on stats refresh):** resolved transitively — each per-group count is now an iterative SCAN rather than a server-blocking `KEYS`.
- **#6 (synchronous `deleteFromPattern` propagates):** wrapped in try/catch (log + swallow) so cache `remove()`/`removeAll()` never throw into callers, matching the other `CacheProvider`s.
- **#5 (codec ignored when `LETTUCE_CLIENT_CLASS` set):** added a clarifying comment in `RedisClientFactory.getOrCreate`. Behavior unchanged (pre-existing).
- **Minor — redundant `"0"` eval arg:** moot now that `keyCount` no longer uses `eval`.
- **Minor — `instanceof` pattern:** applied to `extractObject`.

Deferred (pre-existing, out of scope for this defect PR): **getGroups() can list groups with 0 live keys** once all their keys expire (the group SET is never pruned). Tracking this separately rather than expanding scope here.

wezell:
Round 2 — addressed in `b4278e8adb`:

- **#1 (deleteFromPattern holds a pooled connection across the whole SCAN):** documented the pool-hold and the throughput ceiling. I deliberately did **not** release/reborrow between batches: under master-replica a SCAN cursor is bound to the node it started on and isn't portable across connections/nodes, so reborrowing could corrupt the iteration. `removeAll()` is a rare full-flush and the default pool leaves other slots free for concurrent get/put/stats.
- **#5 (getGroups `.toString()` NPE):** added a `null` filter before `toString()`.

Verified, no change needed:
- **#2 (keyCount double-wraps):** correct as-is. `scanKeys` prepends `keyPrefix()`, stored keys carry that prefix (`clusterid_PREFIX_KEY.group.key.`), so the match is right; `keyCount` only counts `keys.size()` (no prefix-stripping), so there's no zero-count mismatch. The `getKeys` prefix-strip quirk is pre-existing and unaffected by this change.

Acknowledged as low-risk / intentional (no change):
- **#3** `isOpen()` is a null-check + `isOpen()` and won't throw in practice; the leak-guard covers the realistic path.
- **#4** the per-name client cache (and codec ignore when `LETTUCE_CLIENT_CLASS` is set) is existing by-design behavior, now documented.
- **#6** `NullLettuceClient` intentionally relies on the interface's default pub/sub no-ops (null/empty/-1L) rather than restating them.

wezell:
Round 3 — addressed in `1d777f1c4d`:

- **#1 (HIGH — SCAN→replica / UNLINK→master split):** good catch, this was a regression from the SCAN switch (the old Lua `eval` ran entirely on master). `deleteFromPattern` now pins `ReadFrom.UPSTREAM` on the borrowed master-replica connection for the duration of the scan+unlink and restores `REPLICA_PREFERRED` afterward, so both target the master and unreplicated keys aren't missed. The single-URL path is unaffected (plain connection, one node).
  - For `keyCount()`/`getStats()` I deliberately left the scan on the replica: that's an approximate display count where replication lag is acceptable, and pinning it would add master load on every stats refresh.
- **#2 (MED — getSubscribers NPE):** null-guarded the per-UUID map lookup.
- **#4 (MINOR — inaccurate comment):** corrected — the cursor-portability rationale was a Cluster concern, not master-replica; replaced with the actual reason (read/write routing split).

Deferred (pre-existing, out of scope for this defect PR):
- **#3 (LOW — glob metacharacters in group names):** cache group names are internal region identifiers (code constants), not user/API-derived, so `*?[\` injection isn't reachable in practice. Worth a follow-up if group names ever become externally-derived; not expanding scope here.
- **#5 (LOW — unsubscribe remove+size+map.remove not atomic):** pre-existing check-then-act race in the pub/sub bookkeeping. The `CopyOnWriteArrayList` change is strictly safer than the prior plain `ArrayList`; a full synchronized redesign of the subscribe/unsubscribe bookkeeping is broader than this defect PR and is better done as a focused pub/sub change.

wezell:
Round 4 — addressed the nits in `f9e6706689`:

- **MED (raw cast):** parameterized both `StatefulRedisMasterReplicaConnection` casts as `<String, V>` to drop the unchecked-cast warnings.
- **MED (`getClient(name, codec)` misleading API):** added Javadoc spelling out that the codec is ignored for an already-cached name or when `LETTUCE_CLIENT_CLASS` is set.
- **LOW (`size() == 0`):** switched to `isEmpty()`.

Deferred (pre-existing, follow-up):
- **LOW (`getKeys` trailing dot):** real but pre-existing — `cacheKey` always appends a trailing `.`, so `getKeys` returns `key.` and a naive `get(group, "key.")` round-trip would double the dot. The Redis-gated tests don't exercise this and fixing it means touching `cacheKey`'s key shape (broader blast radius), so I'm leaving it for a dedicated follow-up rather than widening this defect PR.

I believe the substantive items are all resolved now; remaining note is the deferred pre-existing `getKeys` round-trip, tracked for follow-up.

wezell:
Round 5 — addressed in `7ce929c66a`:

- **MED (partial-flush visibility):** `deleteFromPattern` now tracks keys removed and reports the count in the warn log, so a partial flush (e.g. a per-batch `RedisCommandTimeoutException`) is distinguishable from a clean completion. Note the connection timeout is per round-trip, not per whole loop, so each SCAN/UNLINK batch (≤1000 keys) has the full budget — cumulative `removeAll()` size doesn't itself trip the timeout.
- **LOW (buildPool raw cast):** parameterized to `<String, V>` to match the `deleteFromPattern` casts.

Acknowledged / deferred:
- **LOW (`getStats()` pool contention at high group count):** per-group SCAN borrows can serialize against the default pool of 5 during heavy activity. Not a regression in behavior, but the per-group borrow pattern makes it more visible. A shared single-connection stats sweep is a reasonable optimization but is broader than this defect PR — leaving as a noted follow-up.

I think the actionable items are now resolved; remaining notes (`getStats` pool sizing, `getKeys` trailing-dot round-trip) are pre-existing and tracked for follow-up rather than expanded here.

wezell:
Round 6 — addressed the remaining doc/robustness items in `4ec9f339c6`:

- **`deleteFromPattern` interface contract:** added Javadoc on `RedisClient.deleteFromPattern` documenting that the pattern must NOT include the cluster key prefix (implementations prepend it) — prevents the double-prefix-no-op / cross-cluster-match footgun for future callers and overrides.
- **Silent success path:** added a `Logger.debug` on successful completion reporting the keys-removed count, symmetric with the partial-flush warn.
- **`getIntProperty` NFE:** now tolerates a non-numeric config value via `Try` + default (with a warn) instead of throwing `NumberFormatException` at startup from `buildPool()`.

Thanks for the thorough multi-round review — I believe all actionable findings are now resolved. The two remaining notes (`getKeys` trailing-dot round-trip, `getGroups` stale zero-key groups) are genuinely pre-existing and out of scope; I've left them as tracked follow-ups rather than widening this defect PR.

wezell:
Round 7 — addressed the MED in `feb53f9146`:

- **MED (`removeAll()` leaves group registry → unbounded `getGroups()`/`getStats()` cost):** good catch — `removeAll()` now also clears the `GROUP_KEY` registry SET via `deleteNonBlocking(REDIS_GROUP_KEY)`. After a full flush, `getGroups()` returns empty (so stats shows the `n/a` fallback) and the registry is repopulated by subsequent puts. This bounds the growth and resolves the previously-deferred "stale zero-key groups" note too.

Deferred (LOW):
- **`subscribe()` doesn't catch a throw from `getPubSubConn()`:** real, but the right behavior is a judgment call — silently swallowing a `connectPubSub` failure would mask a subscription that never receives messages, which is arguably worse than surfacing the error to the caller. I'd rather not change pub/sub failure semantics inside this cache-defect PR; flagging it for the pub/sub owners as a focused follow-up.

That should be the last actionable item — appreciate the thorough pass.

wezell:
Round 8 — these are refinements rather than defects, so I'm holding the code here (the substantive fixes are in and the full test matrix is green). Disposition:

- **MED (ReadFrom restore hardcodes `REPLICA_PREFERRED`):** not a bug in practice — that's the only value `buildPool` ever sets on these connections, and Lettuce 6.1 doesn't expose a `getReadFrom()` to capture/restore the prior value dynamically. Restoring the known pool default is correct.
- **MED (`remove(group)` doesn't prune the group registry):** real, but **bounded** — cache groups are a finite set of region names and are re-added on the next put, so a removed group lingers only as a transient 0-count row in stats. Unlike the `removeAll` case (which accumulated *every* group across *every* flush, i.e. unbounded — now fixed), this needs a new `SREM`/`removeMembers` on the `RedisClient` interface to fix cleanly, which is broader than this defect PR. Tracking as a small follow-up.
- **LOW (`removeAll` sync-delete vs async registry-delete race):** the registry clear is fire-and-forget by design; a concurrent put during a full flush is inherently racy regardless, and the effect is a transient stale stats row, not data loss.
- **LOW (`deleted` count is keys-targeted, not confirmed):** fair wording point; `UNLINK` does remove them (async memory reclaim). Not worth re-spinning the full CI matrix for a log-string tweak.
- **LOW (`subscribe()` throws if `getPubSubConn()` fails):** deferred as in round 7 — swallowing it would silently mask a dead subscription; the failure-semantics call belongs with the pub/sub owners.

Net: all serious/actionable items across 7 rounds are resolved; the remaining notes are bounded/cosmetic and tracked as follow-ups. Ready for human review.

wezell:
Tested with the docker image, working:

Screenshot 2026-06-08 at 10 55 27 AM

Screenshot 2026-06-08 at 10 58 26 AM

dsantos-dcms (approved):

PR #36042, opened 2026-06-08
fix(block-editor): table header toggle, scroll-jump, and table-properties popover cleanup

### LINKED ISSUE #35980: Block Editor bug hunting: editability after save & scroll-jump on empty-line delete
## Description

Tracking task for **bug hunting in the Block Editor (TipTap)**. We currently have three confirmed defects (listed below) and this ticket is intended as a living checklist — additional Block Editor bugs found during QA/exploration should be appended to the Acceptance Criteria as they are discovered.

### Confirmed bugs

**1. Block Editor becomes non-editable after saving content (new Edit Content)**
- In the new Edit Content screen, after saving a contentlet the Block Editor field is no longer editable — the user cannot continue typing/editing the content.
- Expected: the Block Editor should remain editable after a save so the user can keep editing without reloading.

**2. Editor "jumps" / scrolls when deleting an empty line**
- When deleting an empty line, the editor view jumps. This appears to be caused by TipTap's `scrollIntoView` behavior firing on the change, making the viewport jump as if the content shifted.
- Expected: deleting an empty line should not cause the viewport to scroll/jump unexpectedly.

**3. Table "toggle header" only affects the first row / first column**
- In the Block Editor table, the "toggle header row" action only works on the first row, and "toggle header column" only works on the first column — toggling the header on other rows/columns has no effect.
- Expected: toggling the header should work consistently for any selected row or column, not just the first one.

## Acceptance Criteria

- [x] **Bug 1** — After saving content in the new Edit Content screen, the Block Editor field remains editable (user can continue editing without reloading).
- [x] **Bug 2** — Deleting an empty line in the Block Editor no longer causes the viewport to jump/scroll (review TipTap `scrollIntoView` usage).
- [x] **Bug 3** — Table "toggle header row" / "toggle header column" works for any selected row or column, not only the first one.
- [x] Any additional Block Editor bugs found during this hunt are documented here (append new checkboxes) and addressed or split into their own issues.
- [ ] Fixes verified manually in the new Edit Content / UVE Block Editor.

## Priority

Medium

## Additional Context

- Component: Block Editor field (TipTap-based), frontend (`core-web`).
- Bug 2 is suspected to stem from TipTap's `scrollIntoView` being triggered on content changes.
- Bug 3 concerns the TipTap table extension header-toggle behavior.
- This is an exploratory bug-hunting task; expect the checklist above to grow as more defects are uncovered.

rjvelazco:
Block Editor bug-hunting (#35980). Three confirmed defects from the ticket plus two UX cleanups in the new TipTap-based block editor (`libs/new-block-editor`). Each is an isolated commit.

### Proposed Changes

* **Bug 3 — Table header toggle only affected the first row/column.** TipTap's `toggleHeaderRow`/`toggleHeaderColumn` delegate to prosemirror-tables' `toggleHeader('row'|'column')`, whose modern logic is hard-coded to row 0 / column 0 (`top:0,bottom:1` / `left:0,right:1`). Since dotCMS exposes a *per-row / per-column* toggle via the cell handle popovers, every toggle landed on the first row/column. Overrode both commands on `DotTable` to use prosemirror-tables' deprecated variants, which act on the **currently selected** cell's row/column (the popover places the selection in the target cell first).

* **Bug 2 — Editor "jumped" (page scroll) when deleting an empty line.** ProseMirror's default scroll-into-view (fired on edits like deleting an empty line) walks up the DOM and scrolls **every** scrollable ancestor — including the host page's main scroll. Inside dotCMS the editor lives in a taller scrollable page (new Edit Content / UVE), so the whole viewport jumped even though the caret never left the editor. Added `editorProps.handleScrollToSelection` that scrolls only the editor's own `.editor-scroll-container` (minimally) and returns `true` so ProseMirror leaves the page scroll alone.

* **Table properties popover cleanup.** Removed the buggy "Add caption" checkbox (it gated the caption input behind a boolean that drifted out of sync). The caption is now always editable — a non-empty value adds the `<caption>`, an empty value clears it. Converted the three text fields to PrimeNG `pInputText` at `small` size with placeholders (new i18n keys).

* **cursor-pointer on popover action buttons.** Tailwind v4's preflight no longer applies `cursor:pointer` to native `<button>`, so the hand-rolled Cancel/Apply/Insert buttons looked non-clickable. Added `cursor-pointer` across the table-properties, table, link, image, and asset-by-url popovers.

> **Bug 1 (Block Editor non-editable after save)** is intentionally **not** in this PR. Runtime investigation showed the editor should self-heal via Angular 21's default `setDisabledState` re-push on the post-save form rebind; the reporter is manually re-testing before we decide whether a fix is needed (the only fragile spot is the now-stale `this.form.disabled` guard in `dot-edit-content-form`).

### Checklist
- [x] Translations — added 3 placeholder keys to `Language.properties` (`table-properties.{caption,aria-label,aria-labelledby}.placeholder`); removed the unused `add-caption` key.
- [x] Security Implications Contemplated — none; UI-only changes, no new inputs reach the backend.

### Additional Info

Runtime-verified with Playwright against the standalone editor harness (`nx serve dotcms-block-editor`) and the real new Edit Content screen:
- **Bug 2:** deleting an empty line near the container edge moved the main page scroll ~26px before the fix and **0px** after, while the editor's own container still scrolls to follow the caret (1637px) when it leaves the visible area.
- **Popover cleanup:** confirmed the checkbox is gone, the caption input is always enabled, and all three fields render as `p-inputtext p-inputtext-sm` with placeholders bound.

### Video

https://github.com/user-attachments/assets/ee411e8b-733a-47e4-b24e-f586d68eff6f

This PR fixes: #35980


rjvelazco:
Thanks for the thorough review. Addressed the actionable items in 12bb1059bd:

- **#1 (deprecated table API):** added a `TODO` on the `toggleHeaderRow`/`toggleHeaderColumn` overrides noting they rely on prosemirror-tables' deprecated single-arg exports (re-exported via `@tiptap/pm`), so an upstream bump could drop them without a compile-time break.
- **#2 (scroll-container contract):** documented `.editor-scroll-container` as a required contract in the `scrollCaretIntoEditorContainer` docstring — without that wrapper the handler returns `false` and the page-level jump returns.
- **Formatting:** fixed a Prettier issue on the `aria-label` placeholder binding.

### On #3 (caption never rendered) — confirmed, and split into a follow-up

I verified this live in the standalone editor: setting a caption produces **no `<caption>` element** (`<table><colgroup><tbody>…` only). The root cause is that the table runs with `resizable: true`, so prosemirror's `TableView` NodeView drives the in-editor display — neither `renderHTML: () => ({})` nor the JSON attribute ever becomes a visible caption.

Important nuance: dotCMS stores the **ProseMirror JSON**, not rendered HTML, and the caption **is** persisted on the `table` node's `attrs`. So this isn't a data-loss bug — it's a **UX gap**: the author sets caption / aria-label / aria-labelledby and gets *no visual hint* that the table carries that metadata (the caption isn't shown, and the aria attributes are invisible by nature).

Making the caption actually render means customizing the resizable `TableView` — a real, riskier change that's **pre-existing** (not introduced by this PR) and deserves its own focused work + verification. So I've:
- kept this PR scoped to the toggle/scroll/popover fixes,
- corrected the docstrings/comments that overclaimed the caption renders today, and
- opened a follow-up to make the metadata visible: **#36050**.

core-web/libs/new-block-editor/src/lib/editor/extensions/table-extensions.ts
```diff
@@ -9,6 +10,36 @@ import type { EditorPopoverService } from '../services/editor-popover.service';
...
+     * FIRST row / FIRST column regardless of the cursor position (it builds the cell rect with
+     * `top: 0, bottom: 1` for rows and `left: 0, right: 1` for columns). dotCMS exposes a
+     * per-row / per-column header toggle through the cell handle popovers, so the default
+     * commands make every toggle land on the first row/column only.
+     *
+     * We override them with prosemirror-tables' deprecated variants, which operate on the row /
+     * column of the *currently selected* cell (the popover places the selection in the target
+     * cell before running the command). See issue #35980 (bug 3).
+     *
+     * TODO: `toggleHeaderRow` / `toggleHeaderColumn` are prosemirror-tables' deprecated
```
dario-daza: I this TODO comment a future fix in another task?

dario-daza (approved):

core-web/libs/new-block-editor/src/lib/editor/extensions/table-extensions.ts
rjvelazco: Done! It was more a comment like we are using a custom toggle command to achieve what we want not like a todo per se

PR #36052, opened 2026-06-08
fix(block-editor): scope image/link popover anchor to the owning editor (#35908)

### LINKED ISSUE #35908: [BUG] Multiple Block Editor fields in a content type share state, causing menus to misbehave
## Problem Statement

When a content type contains **two or more Block Editor fields**, the editor instances appear to share the same internal ID or state. This causes menus triggered from one Block Editor (e.g., Image Properties, Add Link, Add Hyperlink) to either fail to open or open in the wrong editor instance.

With a single Block Editor in a content type, everything works correctly. The issue only surfaces when multiple Block Editor fields coexist in the same content type.

**Impact:** Editors who need multiple rich-text Block Editor fields in a single content type cannot reliably insert images or links, making the affected fields partially unusable.

https://github.com/user-attachments/assets/55a34f0d-f0fe-4c38-aa4d-adc5f2bb6e7d

---

## Steps to Reproduce

1. Create a content type with **two or more Block Editor fields**.
2. Open a new or existing contentlet of that type.
3. In any of the Block Editor fields, type some text and place the cursor inside it.
4. Try to insert an image: click the image tool → select **Image Properties**.
5. Alternatively, try to insert a link via **Add Link** or **Add Hyperlink**.

**Expected:** The corresponding menu/dialog opens within the Block Editor field currently being edited.

**Actual (inconsistent — one of two behaviors):**
- The menu/dialog does **not open at all**, or
- The menu/dialog opens in a **different Block Editor** instance than the one being edited.

---

## Acceptance Criteria

- [ ] Each Block Editor field instance in a content type has its own unique, isolated ID and state — no shared state between instances.
- [ ] Clicking **Image Properties** inside Block Editor A opens the image menu within Block Editor A, not Block Editor B.
- [ ] Clicking **Add Link** or **Add Hyperlink** inside any Block Editor opens the link dialog within that same editor instance.
- [ ] The behavior is consistent regardless of which editor (first, second, third, etc.) the user interacts with.
- [ ] Content types with 2, 3, or more Block Editor fields all work independently without cross-instance interference.
- [ ] No regression for content types that contain only a single Block Editor field.

---

## dotCMS Version

Latest 

---

## Severity

---

## Links

NA

rjvelazco:
Fixes #35908.

### Proposed Changes
* Scope `EditorModalDirective.getReferenceClientRect()`'s bubble-menu lookup to the **owning editor instance** instead of the whole document, so the image-properties and link popovers anchor to the correct Block Editor when a content type has multiple Block Editor fields.

### Root cause
When a node (image, etc.) is selected, the legacy editor positions the Image Properties / Add Link popover from the bubble menu's rect:

```ts
// editor-modal.directive.ts (before)
const bubbleMenu = document.querySelector('[tiptapbubblemenu]'); // global, first match
```

With **two or more Block Editor fields** in a content type there are multiple `[tiptapbubblemenu]` elements on the page, and `document.querySelector` always returns the **first** one. So opening Image Properties / Add Link in the *second* (or later) editor anchored the popover to the **wrong instance** — the misbehavior in the issue video. With a single Block Editor there's only one match, which is why it only broke with multiple fields.

### The fix
The directive's host (`<form dotEditorModal>`) lives inside the same `<dot-bubble-menu>` host as the `[tiptapbubblemenu]` element, so we scope the lookup to that host:

```ts
// after
const bubbleMenu = this.elRef.nativeElement
    .closest('dot-bubble-menu')
    ?.querySelector('[tiptapbubblemenu]');
```

Each editor now anchors to its own bubble menu. This mirrors the already-correct `target.closest('[tiptapbubblemenu]')` pattern in `dot-link-editor-popover.component.ts`.

### New editor (libs/new-block-editor) is unaffected
I checked the new TipTap editor for the same class of bug and **runtime-verified** it with two instances on one page: each editor opens its own Link popover in the correct instance. The new editor provides all per-editor services at component scope and anchors popovers via per-instance callbacks, so it doesn't share state.

### Checklist
- [x] Translations — none.
- [x] Security Implications Contemplated — none; DOM-scoping change only.

### Additional Info / Verification
- The fix's correctness rests on the DOM structure (verified from templates): the `dotEditorModal` `<form>` and the `[tiptapbubblemenu]` element are siblings inside one `<dot-bubble-menu>` host, so `closest('dot-bubble-menu')` resolves to the directive's own editor instance.
- The legacy editor could not be driven in the isolated dev harness (it requires backend APIs), and the new Edit Content screen renders the *new* editor (no bubble menu), so the bubble-menu interaction is best confirmed with the issue's original repro: a content type with 2+ Block Editor fields in the classic content editor, then Image Properties / Add Link in the second editor.
- **Related, out of scope:** `AiContentPromptStore` is `@Injectable({ providedIn: 'root' })`, so AI-content state is also shared across instances — a separate shared-state issue worth a follow-up, not addressed here.

### Video

https://github.com/user-attachments/assets/a45b35cf-9247-481d-a622-299165b17264

https://github.com/user-attachments/assets/e59d98b6-f6f2-4703-b0a1-05bacaa16583

This PR fixes: #35908


core-web/libs/block-editor/src/lib/directive/editor-modal.directive.ts
hmoreras, line 1: test needed ?

hmoreras (approved):

core-web/libs/block-editor/src/lib/directive/editor-modal.directive.ts
rjvelazco: Don't think so. This lib should be deprecated

dario-daza (approved):

PR #36053, opened 2026-06-08
feat(edit-content): enhanced locale selector for content editing sidebar

### LINKED ISSUE #35889: Build locale-management component for Edit Contentlet (isolated: Storybook + tests)
### Description

Create a new **locale-management component** for the edit content screen of the new Edit Contentlet. Build and validate it **in isolation first** (Storybook + unit tests) before injecting it into the screen — the goal is to surface edge cases and missed scenarios early, and to verify whether the endpoint response provides everything the component needs.

The component renders two layouts driven by the **number of locales**:

#### Few locales (≤ 5)
- Inline card under the accordion.
- Locales grouped under **Translated** (first) and **Pending** titles, each with a count.
- Single-line rows: `Language en-XX`, with a **Default** chip at the right when applicable.
- Rows are **selectable** — selecting a locale switches to it (same behavior as the many-locales view).
- **Manage locales** link pinned in the card footer.

#### Many locales (> 5)
- Inline card under the accordion.
- Card has **search** + **segmented filter** (All / Translated / Pending) + the full list.
- Selecting a locale updates the **selected option shown at the top** of the locale list.
- **Manage locales** link pinned at the bottom of the list.

#### Manage locales link
- Hidden when the user lacks **Languages portlet** access.
- When visible and clicked, **redirects to the Languages/Locales portlet**.

### Acceptance Criteria

**Deliverable (isolation)**
- [ ] Component is implemented as a standalone Angular component, not yet wired into the edit content screen
- [ ] Storybook stories cover every state: few (≤5) and many (>5), Translated/Pending groupings, Default chip present/absent, and empty/edge states
- [ ] Unit tests (Spectator/Jest) cover rendering logic, the ≤5 vs >5 branch, grouping/counts, selection, and the Manage-locales permission gate
- [ ] A short note documents any gaps found in the endpoint response (data the component needs but the API does not return)

**Few locales (≤ 5)**
- [ ] With ≤5 locales, the component renders as an inline card under the accordion
- [ ] Locales are grouped under **Translated** (shown first) and **Pending**, each with a count
- [ ] Each row is a single line showing `Language en-XX`
- [ ] A **Default** chip appears at the right of the row for the default locale only
- [ ] Selecting a row switches to that locale (consistent with the many-locales view)
- [ ] A **Manage locales** link is pinned in the card footer

**Many locales (> 5)**
- [ ] With >5 locales, the component renders as an inline card under the accordion
- [ ] The card shows a **search** input, a **segmented filter** (All / Translated / Pending), and the full list
- [ ] The segmented filter narrows the list to All / Translated / Pending respectively
- [ ] Search filters the list by language
- [ ] Selecting a locale updates the **selected option shown at the top** of the locale list
- [ ] A **Manage locales** link is pinned at the bottom of the list

**Manage locales — permission & navigation**
- [ ] The Manage locales link is **not rendered** when the user lacks Languages portlet access
- [ ] When the user has Languages portlet access, the link is shown
- [ ] Clicking the link **redirects to the Languages/Locales portlet**

**Edge cases**
- [ ] Exactly 5 locales renders the **few** layout; exactly 6 renders the **many** layout
- [ ] When all locales are translated, the **Pending** group/count is empty (or hidden) and Translated shows the full count
- [ ] When all locales are pending, the **Translated** group/count is empty (or hidden)
- [ ] A single locale renders correctly with the Default chip
- [ ] Counts in the group titles match the actual number of rows in each group

### Priority

Medium

### Additional Context

https://github.com/user-attachments/assets/97c2a89e-38e8-4b42-93ff-e0efcd8b2860

adrianjm-dotCMS:
## Summary

Closes #35889

Adds a new `dot-edit-content-sidebar-locales-selector` component that replaces the existing locale list in the edit-content sidebar with an improved UX, gated behind `FEATURE_FLAG_LOCALE_SELECTOR_V2`.

https://github.com/user-attachments/assets/f64c9970-a67e-4f9a-b87b-b7e3bb54c45d

### Two display modes

- **Simple view** (≤5 locales): groups locales into **Translated** / **Pending** sections with full-width selection highlight and a divider between sections
- **Enhanced view** (>5 locales): tabbed interface (All / Translated / Pending) with count indicators, live search with a clear button, and a sticky "Selected:" row always visible regardless of active tab

### Additional features
- "Manage locales" link rendered only for users with locales portlet access (`DotCurrentUserService.hasAccessToPortlet`)
- `p-button [link]` for the manage locales action
- Feature flag registered in `FeatureFlagName.java`, exposed via `ConfigurationResource`, and seeded in `dotmarketing-config.properties`
- 8 i18n keys added to `Language.properties`
- 30 unit tests (loading, simple view, enhanced view, search, manage locales access)

> **Note:** `$isLocaleSelectorV2` in `dot-edit-content-sidebar.component.ts` is currently hardcoded to `signal(true)` for development. Before merging, it must be wired to the real feature flag via `DotPropertiesService`.

## Test plan

- [ ] Open edit-content with a content type that has ≤5 languages — simple view renders Translated / Pending sections
- [ ] Open edit-content with a content type that has >5 languages — enhanced view with tabs and search renders
- [ ] Search filters locales by name and ISO code; clear button resets the list
- [ ] Switching tabs always keeps the sticky "Selected:" row visible
- [ ] Clicking a locale emits `switchLocale` and updates the form language
- [ ] "Manage locales" link is visible only when user has locales portlet access; clicking navigates to `/locales`
- [ ] Run unit tests: `yarn nx test edit-content --testPathPattern=dot-edit-content-sidebar-locales-selector`
- [ ] Disable feature flag → old `dot-edit-content-sidebar-locales` renders as before

This PR fixes: #35889


core-web/libs/edit-content/src/lib/components/dot-edit-content-sidebar/components/dot-edit-content-sidebar-locales/dot-edit-content-sidebar-locales-selector/dot-edit-content-sidebar-locales-selector.component.spec.ts
```diff
@@ -0,0 +1,388 @@
...
+    'edit.content.sidebar.locales.selector.default': 'DEFAULT',
+    'edit.content.sidebar.locales.selector.tab.all': 'All',
+    'edit.content.sidebar.locales.selector.search': 'Search locales',
+    'edit.content.sidebar.locales.selector.selected': 'Selected:',
+    'edit.content.sidebar.locales.selector.empty': 'All caught up',
+    'edit.content.sidebar.locales.selector.manage': 'Manage locales'
+};
+
+const SIMPLE_LOCALES: DotLanguage[] = [
+    { id: 1, language: 'English', isoCode: 'en-us', translated: true },
```
oidacra: Prefer `createFakeLanguage()` from `@dotcms/utils-testing` over hand-built `DotLanguage` literals here (and for the duplicated `mockLocale` in the sidebar spec) so the mocks stay in sync with the model shape.

core-web/libs/edit-content/src/lib/components/dot-edit-content-sidebar/dot-edit-content-sidebar.component.spec.ts
```diff
@@ -210,7 +218,13 @@ describe('DotEditContentSidebarComponent', () => {
                 expect(workflowComponent).toBeTruthy();
             });
 
-            it('should render DotEditContentSidebarLocalesComponent', () => {
+            it('should render DotEditContentSidebarLocalesComponent when locale data is available', () => {
+                const mockLocale: DotLanguage = { id: 1, languageCode: 'en', language: 'English' };
+                patchState(store as never, {
```
oidacra: Could the `as never` cast be dropped here in favor of the already-typed `store`? As written it fully suppresses store-shape checking, so a future rename of `currentLocale`/`systemDefaultLocale` wouldn't fail these tests — is that intentional, or would passing the typed `store` directly work?

core-web/libs/edit-content/src/lib/components/dot-edit-content-sidebar/components/dot-edit-content-sidebar-locales/dot-edit-content-sidebar-locales-selector/dot-edit-content-sidebar-locales-selector.component.ts
```diff
@@ -0,0 +1,132 @@
...
+
+type LocaleTab = 'all' | 'translated' | 'pending';
+
+type TabDef = Readonly<{ value: LocaleTab; label: string }>;
+
+const SIMPLE_VIEW_MAX_LOCALES = 5;
+
+@Component({
+    selector: 'dot-edit-content-sidebar-locales-selector',
+    standalone: true,
```
oidacra: `standalone: true` is implied by default in Angular 19+ and is omitted by the other components in this folder — safe to remove for consistency.

core-web/libs/edit-content/src/lib/components/dot-edit-content-sidebar/components/dot-edit-content-sidebar-locales/dot-edit-content-sidebar-locales-selector/dot-edit-content-sidebar-locales-selector.component.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    onSearchChange(event: Event): void {
+        this.$searchQuery.set((event.target as HTMLInputElement).value);
+    }
+
+    clearSearch(): void {
+        this.$searchQuery.set('');
+    }
+
+    onLocaleClick(locale: DotLanguage): void {
```
oidacra: `onLocaleClick()` isn't referenced from the template (the listbox uses `(ngModelChange)` directly), so its `locale.id !== currentLocale` guard is never applied. Either wire the listbox to this method or remove it, to avoid the impression the guard is active.

core-web/libs/edit-content/src/lib/models/dot-edit-content.model.ts
```diff
@@ -51,6 +51,7 @@ export interface UIState {
     isSidebarOpen: boolean;
     activeSidebarTab: number;
     isBetaMessageVisible: boolean;
+    localeSelectorTab: 'all' | 'translated' | 'pending';
```
oidacra: `'all' | 'translated' | 'pending'` is repeated here and in `ui.feature.ts`, `functions.util.ts`, and the component. Exporting the `LocaleTab` type (ideally from this model) and reusing it everywhere would keep them in sync.

core-web/libs/edit-content/src/lib/utils/functions.util.ts
```diff
@@ -480,7 +480,8 @@ export const getStoredUIState = (): UIState => {
         activeTab: 0,
         isSidebarOpen: true,
         activeSidebarTab: 0,
-        isBetaMessageVisible: true
+        isBetaMessageVisible: true,
+        localeSelectorTab: 'all'
```
oidacra: Worth double-checking: `getStoredUIState` returns the parsed stored object directly, so a session persisted before this field existed would deserialize without `localeSelectorTab` even though `UIState` types it as required. The `?? 'all'` guard in `ui.feature.ts` covers the read, but spreading defaults here (`{ ...defaults, ...parsed }`) would close the type/runtime gap at the source — could you confirm whether a pre-existing stored state can reach this path?

core-web/libs/edit-content/src/lib/components/dot-edit-content-sidebar/components/dot-edit-content-sidebar-locales/dot-edit-content-sidebar-locales-selector/dot-edit-content-sidebar-locales-selector.component.spec.ts
```diff
@@ -0,0 +1,388 @@
...
+        });
+
+        it('should render the tab bar with All, Translated and Pending tabs', () => {
+            expect(spectator.query(byTestId('tab-all'))).toBeTruthy();
+            expect(spectator.query(byTestId('tab-translated'))).toBeTruthy();
+            expect(spectator.query(byTestId('tab-pending'))).toBeTruthy();
+        });
+
+        it('should default to the All tab', () => {
+            expect(spectator.component.$activeTab()).toBe('all');
```
oidacra: A few assertions read internal signals (`$activeTab()`, `$searchQuery()`) rather than the rendered DOM. Asserting item counts / active-tab state via `byTestId` keeps these behavioral — several here are already redundant with the adjacent DOM checks.

core-web/libs/edit-content/src/lib/components/dot-edit-content-sidebar/dot-edit-content-sidebar.component.html
```diff
@@ -88,13 +88,29 @@
...
-                        <dot-edit-content-sidebar-locales
-                            [locales]="$store.locales()"
-                            [defaultLocale]="$store.systemDefaultLocale()"
-                            [currentLocale]="$store.currentLocale()"
-                            [isLoading]="$store.isLoadingLocales()"
-                            (switchLocale)="$store.switchLocale($event)"
-                            data-testId="locales" />
+                        @let defaultLocale = $store.systemDefaultLocale();
+                        @let currentLocale = $store.currentLocale();
+                        @if (defaultLocale && currentLocale) {
```
oidacra: This gate now wraps the legacy path too; previously it rendered unconditionally and showed its own loading skeleton via `isLoading`. If `currentLocale` can be null while `isLoadingLocales()` is true, the skeleton is now hidden. Worth confirming `currentLocale` is always set during load, or guarding with `@if (isLoadingLocales() || (defaultLocale && currentLocale))`.

core-web/libs/edit-content/src/lib/components/dot-edit-content-sidebar/components/dot-edit-content-sidebar-locales/dot-edit-content-sidebar-locales-selector/dot-edit-content-sidebar-locales-selector.component.spec.ts
adrianjm-dotCMS: Done. Replaced all hand-built `DotLanguage` literals in the selector spec with `createFakeLanguage()` from `@dotcms/utils-testing`, and updated the `mockLocale` in the sidebar spec as well. Removed the `as unknown as DotLanguage[]` casts.

core-web/libs/edit-content/src/lib/components/dot-edit-content-sidebar/components/dot-edit-content-sidebar-locales/dot-edit-content-sidebar-locales-selector/dot-edit-content-sidebar-locales-selector.component.ts
adrianjm-dotCMS: Done. Removed `standalone: true` — it's the default in Angular 19+ and the other components in this folder omit it.

core-web/libs/edit-content/src/lib/components/dot-edit-content-sidebar/components/dot-edit-content-sidebar-locales/dot-edit-content-sidebar-locales-selector/dot-edit-content-sidebar-locales-selector.component.ts
adrianjm-dotCMS: Done. Removed `onLocaleClick()` — it was dead code since the template uses `(ngModelChange)` directly on all three listboxes. PrimeNG's listbox with `dataKey` already suppresses `ngModelChange` when the same item is re-selected, so the guard wasn't needed in practice.

core-web/libs/edit-content/src/lib/models/dot-edit-content.model.ts
adrianjm-dotCMS: Done. Exported `LocaleTab` from `dot-edit-content.model.ts` and replaced the inline union type in `UIState` and the local `type LocaleTab` definition in the component.

core-web/libs/edit-content/src/lib/utils/functions.util.ts
adrianjm-dotCMS: Good catch. Refactored `getStoredUIState` to use `{ ...defaults, ...JSON.parse(storedState) }` so any session stored before `localeSelectorTab` existed will get the default filled in at the source, removing the need for downstream fallbacks.

core-web/libs/edit-content/src/lib/components/dot-edit-content-sidebar/components/dot-edit-content-sidebar-locales/dot-edit-content-sidebar-locales-selector/dot-edit-content-sidebar-locales-selector.component.spec.ts
adrianjm-dotCMS: Done. Removed the redundant `$activeTab()` signal assertions where the adjacent item-count DOM check already covers the behavior. Also replaced the `$searchQuery()` assertions with checks on the input element's `.value` property.

core-web/libs/edit-content/src/lib/components/dot-edit-content-sidebar/dot-edit-content-sidebar.component.html
adrianjm-dotCMS: Good point. Changed the guard to `@if ($store.isLoadingLocales() || (defaultLocale && currentLocale))` so the skeleton is visible during the initial load when both locale signals are still `null`. Bindings use `?? ({} as DotLanguage)` to satisfy strict template checking — the component never reads those inputs when `isLoading` is true.

core-web/libs/edit-content/src/lib/components/dot-edit-content-sidebar/dot-edit-content-sidebar.component.spec.ts
adrianjm-dotCMS: The cast can't be fully removed — `SpyObject<T>` has `StateSource<S>` but `patchState` requires `WritableStateSource<S>`, so TypeScript rejects a direct call. Replaced `as never` with the more explicit `as unknown as WritableStateSource<object>` to communicate the intent while preserving the same runtime behavior. Also updated both `mockLocale` usages to `createFakeLanguage()`.

oidacra (approved):

PR #36070, opened 2026-06-09
Create new endpoint to save style editor schema per content type

### LINKED ISSUE #35781: [Style Editor] - Updating/Creating Style Editor Schema for Content Type Silently Deletes Assigned Workflow
### Problem Statement

When a user creates or updates the Style Editor schema for a Content Type, the workflow previously assigned to that Content Type is silently deleted. This issue severely impacts Content Types because the deletion occurs without any user notification. Consequently, users are unaware of the change and may continue creating Contentlets of that type incorrectly, which disrupts their entire workflow. This is a critical issue due to its silent nature and the significant downstream impact on content creation processes. The original description does not specify if this is a front-end specific issue, therefore browser and OS details are not included.

### Steps to Reproduce

1.  **Prerequisite:** Ensure you have an existing Content Type with a workflow successfully assigned to it.
2.  Navigate to the configuration settings for the aforementioned Content Type.
3.  Access the 'Style Editor Schema' section for this Content Type.
4.  Perform an action to either create a new Style Editor schema or update an existing one for this Content Type (e.g., save changes after modifying the schema).
5.  After saving the Style Editor schema, navigate back to the Content Type's general settings or workflow assignment section.
6.  **Expected Result:** The previously assigned workflow should still be correctly associated with the Content Type.
7.  **Actual Result:** The assigned workflow is found to be silently deleted or unassigned from the Content Type.

[Please provide a video screencast demonstrating the issue here, if possible.]

### Acceptance Criteria

When the Style Editor schema of a Content Type is created or updated, the action must strictly affect only the Style Editor schema. The assigned workflow for the Content Type must remain intact and **not** be deleted or unassigned. No other unintended side effects should occur on any other Content Type configurations, specifically workflows.

### dotCMS Version

latest

### Severity

### Links

https://dotcms.slack.com/archives/C089DJ48C01/p1778859252464499

### LINKED ISSUE #35979: Fields are deleted when saving Style Editor schema on a Content Type
## Problem Statement

When a user creates a Content Type, adds fields to it, and then saves a Style Editor schema from the **Style Editor** tab, all previously added fields are unexpectedly removed from the Content Type.

Multiple toast notifications appear during the save: _"CleanUpFieldReferencesJob finished for field: [fieldName]"_ — one per deleted field. After refreshing the page, the Content Type has no fields left.

This makes it impossible to define a Style Editor schema on a Content Type that already has fields without losing all of them.

## Steps to Reproduce

1. Create a new Content Type
2. Add one or more fields to it (e.g. Text, Custom Field)
3. Navigate to the **Style Editor** tab of the same Content Type
4. Define a Style Editor schema and click **Save**
5. Refresh the page

## Acceptance Criteria

- [ ] Saving a Style Editor schema from the Style Editor tab does not delete any fields previously added to the Content Type
- [ ] No "CleanUpFieldReferencesJob" toasts appear when saving only the Style Editor schema
- [ ] After saving and refreshing, all fields added before the Style Editor save remain intact

## dotCMS Version

Latest from `main` branch

## Severity

## Links

NA

dario-daza:
### Proposed Changes
Create the new `PATCH /v1/contenttype/id/{idOrVar}/metadata` endpoint to avoid editing other Content Type information when creating and saving a Style Editor Schema for a Content Type.

#### BE
* Create the new `PATCH /v1/contenttype/id/{idOrVar}/metadata` endpoint to handle the metadata of the Content Types, specifically the `DOT_STYLE_EDITOR_SCHEMA` that is related to the Style Editor Schema.
* Fix Postman tests using the new endpoint.
* Create an integration test for the new endpoint.

#### FE
* Call the new endpoint when saving Style Editor Schemas per Content Type.
* Replace the use of `DotCrudService` with `http`.
* Fix tests.

This PR fixes: #35781
This PR fixes: #35979

This PR fixes: #35781


dotCMS/src/main/java/com/dotcms/rest/api/v1/contenttype/ContentTypeResource.java
```diff
@@ -2018,6 +2022,155 @@ public final ResponseEntityPaginatedDataView getPagesContentTypes(@Context final
...
+                .rejectWhenNoUser(true)
+                .init()
+                .getUser();
+
+        final ContentTypeAPI contentTypeAPI = APILocator.getContentTypeAPI(user, true);
+        try {
+            if (!UtilMethods.isSet(idOrVar)) {
+                return Response.status(Response.Status.NOT_FOUND).build();
+            }
+            final ContentType existing = contentTypeAPI.find(idOrVar);
```
fabrizzio-dotCMS: **Race condition: lost update on metadata PATCH**

Unguarded read-modify-write: reads `existing.metadata()`, merges in memory, and calls `save()`. Metadata is a single JSON column that `save()` rewrites whole — no optimistic locking. Two concurrent PATCHes on the same Content Type clobber each other (last write wins, a key is silently lost).

**Recommendation:** move the read-merge-save (and the `DOT_STYLE_EDITOR_SCHEMA` normalization) out of the REST layer into `contentTypeHelper`, and serialize it with the striped lock already used in the codebase (`DotConcurrentFactory.getInstance().getIdentifierStripedLock()`, same as `ESContentletAPIImpl` checkin), re-reading **inside** the lock:

```java
return lockManager.tryLock("ct-metadata-" + existing.id(), () -> {
    final ContentType current = contentTypeAPI.find(idOrVar);   // re-read inside the lock
    final Map<String, Object> merged = new HashMap<>(
            current.metadata() != null ? current.metadata() : Map.of());
    patch.forEach((k, v) -> { if (v == null) merged.remove(k); else merged.put(k, v); });
    return contentTypeAPI.save(ContentTypeBuilder.builder(current).metadata(merged).build());
});
```

Note: the striped lock is JVM-local. Fine for low-frequency metadata writes; if multi-node safety is needed, follow up (separate PR) with optimistic locking via `modDate` or a DB-level atomic JSON merge.

dario-daza: Solved

core-web/apps/dotcms-ui/src/app/portlets/shared/dot-content-types-edit/components/style-editor/dot-style-editor-builder.component.ts
```diff
@@ -253,8 +241,11 @@ export class DotStyleEditorBuilderComponent {
...
         const raw = contentType.metadata?.[STYLE_EDITOR_SCHEMA_KEY];
-        if (!raw || typeof raw !== 'string') {
-            console.warn('[StyleEditorBuilder] Invalid schema in metadata');
+        if (!raw) {
+            return;
+        }
+        if (typeof raw !== 'string') {
+            console.warn('[StyleEditorBuilder] DOT_STYLE_EDITOR_SCHEMA is not a string; ignoring');
             return;
         }
```
KevinDavilaDotCMS: Are different error cases?
Maybe can add console.warn in the first too

KevinDavilaDotCMS (approved):

KevinDavilaDotCMS (approved):

fabrizzio-dotCMS:
**Question: is `metadata` preserved when a Content Type is saved through the regular full PUT?**

This new endpoint (`PATCH /api/v1/contenttype/id/{idOrVar}/metadata`) merges into the existing `metadata` and serializes `DOT_STYLE_EDITOR_SCHEMA` correctly — that part looks good.

My concern is the interaction with the regular update path (`putContentTypeUpdate` → `PUT /api/v1/contenttype/id/{idOrVar}`), whose Swagger description states:

> ⚠️ **Destructive semantics.** This endpoint treats the request body as the full desired state. Any editable property (including items in `fields[]` and `metadata` keys) absent from the body will be removed.

So if any client does a normal full-CT PUT **without** carrying `DOT_STYLE_EDITOR_SCHEMA` in `metadata`, the schema written by this new endpoint would be silently wiped.

1. Does the CT edit screen (or any existing client) round-trip the **full** `metadata` map — including `DOT_STYLE_EDITOR_SCHEMA` — on a regular save? If not, a normal save would clobber the style editor schema.
2. Should the style editor schema be protected against the full-PUT's destructive merge (e.g. preserved server-side unless explicitly cleared), or is the contract strictly "GET → mutate → PUT the whole object"?

Mainly want to make sure the schema persisted here survives a subsequent normal CT save.

core-web/apps/dotcms-ui/src/app/portlets/shared/dot-content-types-edit/components/style-editor/dot-style-editor-builder.component.ts
dario-daza: When the value is `undefined` or `null`, the return in line 245 will be executed. Otherwise, the warn log in the catch statement will be thrown.

dario-daza:
> **Question: is `metadata` preserved when a Content Type is saved through the regular full PUT?**
> 
> This new endpoint (`PATCH /api/v1/contenttype/id/{idOrVar}/metadata`) merges into the existing `metadata` and serializes `DOT_STYLE_EDITOR_SCHEMA` correctly — that part looks good.
> 
> My concern is the interaction with the regular update path (`putContentTypeUpdate` → `PUT /api/v1/contenttype/id/{idOrVar}`), whose Swagger description states:
> 
> > ⚠️ **Destructive semantics.** This endpoint treats the request body as the full desired state. Any editable property (including items in `fields[]` and `metadata` keys) absent from the body will be removed.
> 
> So if any client does a normal full-CT PUT **without** carrying `DOT_STYLE_EDITOR_SCHEMA` in `metadata`, the schema written by this new endpoint would be silently wiped.
> 
> 1. Does the CT edit screen (or any existing client) round-trip the **full** `metadata` map — including `DOT_STYLE_EDITOR_SCHEMA` — on a regular save? If not, a normal save would clobber the style editor schema.
> 2. Should the style editor schema be protected against the full-PUT's destructive merge (e.g. preserved server-side unless explicitly cleared), or is the contract strictly "GET → mutate → PUT the whole object"?
> 
> Mainly want to make sure the schema persisted here survives a subsequent normal CT save.

The way we handled it in the frontend, it never happens, but it is possible that any REST API caller (Postman, a migration script, or a third-party integration) that makes a raw PUT and omits metadata from the body would silently wipe the schema. I added a validation that prevents this and preserves the metadata if a user omits it in a PUT request. If you want to wipe the metadata, you'll need to send metadata: null.

dotCMS/src/main/java/com/dotcms/rest/api/v1/contenttype/ContentTypeResource.java
```diff
@@ -2018,6 +2033,106 @@ public final ResponseEntityPaginatedDataView getPagesContentTypes(@Context final
...
+                Logger.warn(this, "No metadata patch found for " + idOrVar);
+                final ContentType existing = contentTypeAPI.find(idOrVar);
+                return Response.ok(new ResponseEntityContentTypeDetailView(
+                        new HashMap<>(contentTypeHelper.contentTypeToMap(existing, user)))).build();
+            }
+
+            final ContentType saved = contentTypeHelper.mergeAndSaveMetadata(idOrVar, metadataPatch, contentTypeAPI);
+            return Response.ok(new ResponseEntityContentTypeDetailView(
+                    new HashMap<>(contentTypeHelper.contentTypeToMap(saved, user)))).build();
+        } catch (final NotFoundInDbException e) {
```
ihoffmann-dot, line 2107: Can we use ResponseUtil.mapExceptionResponse(e) with a single catch block instead of the different blocks we have right now? That util method delagates all the responses creation

ihoffmann-dot (changes requested):

dotCMS/src/main/java/com/dotcms/rest/api/v1/contenttype/ContentTypeResource.java
dario-daza: I used *ErrorEntity* to return responses with `errorCode` and `fieldName` to make it easier to read the cause, use `mapExceptionResponse` would reply plain message strings. Happy to simplify if those fields aren't consumed, but keeping them for now to preserve the API contract

ihoffmann-dot: Sound good!

ihoffmann-dot (approved):

Issue #34067, opened 2025-12-09
JVM crash on startup due to Tomcat Native APR incompatibility with OpenSSL 3.x


mbiuki:
## Problem Description

dotCMS crashes on startup with a JVM segmentation fault when the Tomcat Native APR library attempts to initialize OpenSSL 3.x. This prevents dotCMS from starting successfully.

## Error Details

```
SIGSEGV (0xb) at pc=0x00007fbce915d234
Problematic frame: C [libcrypto.so.3+0x1fd234] EVP_MD_get0_provider+0x4
Java frames:
  org.apache.tomcat.jni.SSL.fipsModeGet()I+0
  org.apache.catalina.core.AprLifecycleListener.initializeSSL()V+198
```

## Root Cause

The Tomcat Native APR library (tcnative 1.2.35) is incompatible with OpenSSL 3.x. This causes a native library crash during SSL initialization when the APR Lifecycle Listener attempts to load OpenSSL.

**System Details:**
- OS: Ubuntu 24.04.3 LTS
- Java: OpenJDK 21.0.4+7-LTS
- Tomcat: 9.0.108
- tcnative: 1.2.35 (outdated - Tomcat recommends minimum 1.3.0)
- OpenSSL: 3.x (libcrypto.so.3)

## Proposed Solution

**Remove libtcnative-1.so from container image** to disable Tomcat Native/OpenSSL entirely and let Tomcat use pure Java JSSE for SSL/TLS operations.

This approach:
- Eliminates the OpenSSL 3.x compatibility issue
- Removes the need to maintain native library versions
- Uses Java's built-in SSL implementation (JSSE), which is fully functional
- Avoids FIPS mode complications with native OpenSSL
- Simplifies the container image

## Configuration Location

`dotCMS/src/main/resources/container/tomcat9/conf/server.xml:5`
```xml
<Listener className="org.apache.catalina.core.AprLifecycleListener" SSLEngine="${CMS_SSL_ENGINE:-on}" />
```

## Alternative Workarounds

1. **Temporary workaround**: Set environment variable `CMS_SSL_ENGINE=off` to disable APR SSL
2. **Upgrade tcnative**: Update to version 1.3.0+ (but this still requires maintaining native library compatibility)

## Recommendation

The cleanest solution is to remove libtcnative-1.so from the container build entirely, eliminating this entire class of compatibility issues and simplifying the deployment.

## Impact

- **Severity**: High - prevents dotCMS startup
[...]

PR #35777, opened 2026-05-20
fix(docker): remove libtcnative-1 at runtime when FIPS mode is detected

dsilvam:
## Summary

Fixes the JVM crash (`SIGSEGV` in `libcrypto.so.3 EVP_MD_get0_provider+0x4`) that occurs when running dotCMS on FIPS-enabled hosts (e.g. RHEL with CIS Level 2 hardening).

The fix shipped in PR #34213 (`SSLEngine=off`) was insufficient. Setting `SSLEngine=off` only prevents OpenSSL from being used for SSL connections, but `libtcnative-1` still loads `libcrypto.so.3` and calls OpenSSL for non-SSL operations (e.g. random number generation). On a FIPS-enabled kernel without the OpenSSL FIPS provider (`fips.so`, not shipped in Ubuntu 24.04), this triggers the same crash regardless of the SSL engine setting.

This PR extends `15-detect-fips-and-set-ssl-engine.sh` to also remove `libtcnative-1.so` at container startup when FIPS mode is detected, before Tomcat starts. Tomcat then falls back to pure Java NIO/JSSE automatically. `libtcnative-1` remains installed in the image by default for all non-FIPS environments — no performance impact for the majority of installations.

## Changes

- `15-detect-fips-and-set-ssl-engine.sh`: when `FIPS_ENABLED=true`, remove `libtcnative-1.so.0*` before Tomcat starts, in addition to setting `CMS_SSL_ENGINE=off`
- Added comment explaining why `SSLEngine=off` alone is not sufficient

## Test plan

- [ ] Start container on a **non-FIPS host** — `libtcnative-1` must still load and `CMS_SSL_ENGINE=on` (no regression)
- [ ] Start container on a **FIPS-enabled host** (`/proc/sys/crypto/fips_enabled=1`) — container must start without SIGSEGV, log must show `[FIPS Detection] Removing libtcnative-1 to prevent OpenSSL initialization in FIPS mode`
- [ ] Verify `CMS_DISABLE_APR_SSL=true` path still works on non-FIPS host
- [ ] Verify `CMS_SSL_ENGINE=off` manual override still works

## Related

Closes #34212
Related: #34067, PR #34213, Freshdesk ticket #34489

Issue #34489, opened 2026-02-03
Investigate try-with-resources closing ThreadLocal connections in FolderFactoryImpl.findSystemFolder()


spbolton:
## Description

FolderFactoryImpl.findSystemFolder() uses try-with-resources with DbConnectionFactory.getConnection() which may prematurely close parent ThreadLocal connections when called from within transactions. This could cause intermittent Connection closed errors in parent methods.

## Acceptance Criteria

- [ ] [Define what needs to be accomplished]

## Additional Notes

[Any additional context or notes]

PR #34213, opened 2026-01-06
fix: Add FIPS mode detection and auto-disable APR SSL Engine

mbiuki:
## Summary

This PR implements automatic FIPS mode detection to prevent JVM crashes with OpenSSL 3.x while maintaining the performance benefits of the Tomcat Native APR library by default.

This addresses the reviewer feedback from @wezell on PR #34068:
> "We need to maintain the `libtcnative` functionality by default. It brings performance benefits to a majority of dotCMS installations. Instead of this PR, add a flag that either checks for FIPS enabled environments and disables it or just a configuration flag."

### 1. New FIPS Detection Script
**File**: `dotCMS/src/main/docker/original/ROOT/srv/15-detect-fips-and-set-ssl-engine.sh`

- Automatically detects FIPS mode by checking `/proc/sys/crypto/fips_enabled`
- Sets `CMS_SSL_ENGINE=off` when FIPS mode is detected
- Provides `CMS_DISABLE_APR_SSL` flag for manual control
- Defaults to APR SSL enabled for optimal performance
- Respects explicit `CMS_SSL_ENGINE` settings (user override)

### 2. Entrypoint Integration
**File**: `dotCMS/src/main/docker/original/ROOT/srv/entrypoint.sh`

- Sources FIPS detection script early in startup process
- Ensures environment variables are set before Tomcat starts

### 3. Documentation Updates
**File**: `dotCMS/src/main/resources/container/tomcat9/conf/server.xml`

- Added comprehensive documentation about FIPS auto-detection
- Explains APR SSL Engine functionality and benefits
- Documents all configuration options
- Clarifies performance implications

### 4. User Guide
**File**: `FIPS_APR_SSL_FIX.md`

- Complete configuration guide with examples
- Testing scenarios and verification steps
- Troubleshooting guide
- Migration guide from PR #34068

## Configuration Options

Users have three ways to control APR SSL behavior:

### Option 1: Automatic FIPS Detection (Default)
```bash
docker run -p 8080:8080 dotcms/dotcms:latest
```
- FIPS mode automatically detected at startup
- APR SSL disabled if FIPS enabled, otherwise enabled for performance

### Option 2: Manual Disable Flag
```bash
docker run -e CMS_DISABLE_APR_SSL=true -p 8080:8080 dotcms/dotcms:latest
```
- Explicitly disable APR SSL without FIPS mode
- Useful for OpenSSL 3.x systems not in FIPS mode

### Option 3: Direct Control
```bash
[...]

PR #36071, opened 2026-06-09
fix(docker): relocate libtcnative-1 to /srv/native-libs for correct FIPS removal

### LINKED ISSUE #34212: fix: Add FIPS mode detection and auto-disable APR SSL Engine
## Problem

The Tomcat Native APR library (libtcnative-1) version 1.2.35 is incompatible with OpenSSL 3.x when running in FIPS mode, causing JVM segmentation faults during startup on modern systems (Ubuntu 24.04+, RHEL 9+).

PR #34068 attempted to fix this by removing the native library entirely, but the reviewer (@wezell) requested that we maintain the native library by default for performance benefits and instead add FIPS detection or a configuration flag.

## Solution

This fix implements the reviewer's recommendation:

1. **Keeps native library installed by default** - No changes to Dockerfiles; libtcnative-1 remains for performance
2. **Automatic FIPS detection** - Checks `/proc/sys/crypto/fips_enabled` at container startup
3. **Configuration flags** - Provides `CMS_DISABLE_APR_SSL` and `CMS_SSL_ENGINE` for manual control
4. **SSL endpoint stays enabled** - Only disables APR library when needed, not SSL itself

### New Files
- `dotCMS/src/main/docker/original/ROOT/srv/15-detect-fips-and-set-ssl-engine.sh` - FIPS detection script that runs at container startup
- `FIPS_APR_SSL_FIX.md` - Comprehensive documentation with configuration options and troubleshooting

### Modified Files
- `dotCMS/src/main/docker/original/ROOT/srv/entrypoint.sh` - Sources the FIPS detection script
- `dotCMS/src/main/resources/container/tomcat9/conf/server.xml` - Added documentation about FIPS auto-detection

## Configuration Options

Users have three ways to control APR SSL behavior:

```bash
# Option 1: Automatic (default) - FIPS auto-detection
docker run -p 8080:8080 dotcms/dotcms:latest

# Option 2: Manual disable flag
docker run -e CMS_DISABLE_APR_SSL=true -p 8080:8080 dotcms/dotcms:latest

# Option 3: Direct control
docker run -e CMS_SSL_ENGINE=off -p 8080:8080 dotcms/dotcms:latest
```

## Testing Plan

- [ ] Build Docker image successfully
- [ ] Test container startup in non-FIPS environment (APR SSL should be enabled)
- [ ] Test container startup with `CMS_DISABLE_APR_SSL=true` (APR SSL should be disabled)
- [ ] Test container startup with explicit `CMS_SSL_ENGINE=off` (APR SSL should be disabled)
- [ ] Verify startup logs show correct FIPS detection messages
- [ ] Test SSL/TLS connectivity on port 8443

## Impact

- **User Impact**: None for non-FIPS environments - APR SSL remains enabled by default
- **Performance**: No change - APR SSL still used by default for optimal performance
- **Security**: Improved - FIPS environments now work automatically without crashes
- **Compatibility**: Improved - Eliminates OpenSSL 3.x + FIPS crashes

## Related Issues

- Addresses reviewer feedback on PR #34068
- Fixes #34067 (if that issue exists)

dsilvam:
## Summary

Supersedes PR #35777 which was merged but proved insufficient in customer testing.

The previous fix (removing `AprLifecycleListener` from `server.xml` at runtime) did not prevent the crash because `setenv.sh` sets `-Djava.library.path=/usr/lib/<arch>-linux-gnu/` and Tomcat auto-detects and loads `libtcnative-1` from that path regardless of `server.xml` configuration.

**Root cause confirmed:** On a FIPS-enabled kernel (`fips_enabled=1`), OpenSSL 3.x requires the FIPS provider module (`fips.so`) before allowing any cryptographic operation. Ubuntu 24.04 does not ship `fips.so`. The first OpenSSL call — `EVP_MD_get0_provider` for random number generation inside `libtcnative-1` — segfaults with `SIGSEGV` in `libcrypto.so.3`. This occurs even with `SSLEngine=off` and even without `AprLifecycleListener` in `server.xml`.

## Changes

- **`Dockerfile`**: During image build (as root), move `libtcnative-1.so.0*` from `/usr/lib/<arch>-linux-gnu/` to `/srv/native-libs/` (owned by the `dotcms` user) and leave symlinks in `/usr/lib`. Works on both `amd64` and `arm64` via `$(uname -m)`.
- **`15-detect-fips-and-set-ssl-engine.sh`**: When FIPS detected, `rm /srv/native-libs/libtcnative-1.so.0*` (writable by `dotcms` user). The symlinks in `/usr/lib` become dangling so `dlopen()` cannot load the library regardless of `java.library.path` or `server.xml`. Includes a post-condition check that logs loudly on failure.

Non-FIPS containers are unaffected: symlinks resolve to the real files normally and the library loads, preserving performance.

## Test plan

- [ ] Build image and start container on a **non-FIPS host** — `libtcnative-1` must still load (`AprLifecycleListener - Loaded Apache Tomcat Native library`), `CMS_SSL_ENGINE=on`
- [ ] Start container on a **FIPS-enabled host** (`/proc/sys/crypto/fips_enabled=1`, e.g. RHEL 8 with CIS Level 2) — container must start without `SIGSEGV`, log must show `libtcnative-1 removed from /srv/native-libs`
- [ ] Verify `ls /usr/lib/<arch>-linux-gnu/libtcnative-1.so.0*` on FIPS container shows dangling symlinks
- [ ] Verify `ls /srv/native-libs/` on FIPS container is empty after startup
- [ ] Verify `arm64` image build succeeds (symlink path uses `aarch64-linux-gnu`)

## Related

Closes #34212
Related: #34067, PR #34213, PR #35777, Freshdesk ticket #34489

### DIFF
--- a/dotCMS/src/main/docker/original/Dockerfile
+++ b/dotCMS/src/main/docker/original/Dockerfile
@@ -92,6 +92,19 @@ COPY --from=container-base /java /java
 COPY --from=container-base /srv /srv
 COPY --from=container-base /data /data
 
+# Move libtcnative-1 to /srv/native-libs (owned by dotcms user) and leave symlinks
+# in /usr/lib. This lets the FIPS detection script remove the library at runtime
+# without root access — making the symlinks dangling so dlopen() fails to load it.
+RUN set -e; \
+    ARCH=$(uname -m); \
+    mkdir -p /srv/native-libs; \
+    for f in /usr/lib/${ARCH}-linux-gnu/libtcnative-1.so.0*; do \
+        [ -e "$f" ] || continue; \
+        mv "$f" /srv/native-libs/; \
+        ln -sf "/srv/native-libs/$(basename "$f")" "$f"; \
+    done; \
+    chown -R dotcms:dotcms /srv/native-libs
+
 USER $USER_UID:$USER_GID
 ENV JAVA_HOME="/java"
 ENV PATH=$PATH:/java/bin
--- a/dotCMS/src/main/docker/original/ROOT/srv/15-detect-fips-and-set-ssl-engine.sh
+++ b/dotCMS/src/main/docker/original/ROOT/srv/15-detect-fips-and-set-ssl-engine.sh
@@ -2,25 +2,27 @@
 
 # FIPS Mode Detection and APR SSL Engine Configuration
 # =====================================================
-# This script automatically detects FIPS-enabled environments and disables the
-# Tomcat Native APR SSL Engine to prevent JVM crashes with OpenSSL 3.x.
+# This script automatically detects FIPS-enabled environments and prevents
+# libtcnative-1 from loading to avoid JVM crashes with OpenSSL 3.x in FIPS mode.
 #
-# The Tomcat Native APR library (libtcnative-1) version 1.2.35 is incompatible
-# with OpenSSL 3.x when running in FIPS mode, causing segmentation faults.
-# Setting SSLEngine=off alone is insufficient: libtcnative-1 still loads
-# libcrypto.so.3 and calls OpenSSL for non-SSL operations (e.g. random number
-# generation), which triggers the same FIPS provider crash.
+# Root cause: libtcnative-1 links against libcrypto.so.3. On a FIPS-enabled kernel,
+# OpenSSL 3.x requires the FIPS provider (fips.so) to be present before allowing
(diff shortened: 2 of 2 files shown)


wezell (approved):
Ok, hate this complexity but looks like it could work.

Issue #36072, opened 2026-06-09
[TASK] Content type list endpoint — support excluding system types + global sort across base types


zJaaal:
## Description

Content Drive's new "New" button menu reuses the UVE palette content-type list against `GET /api/v1/contenttype` (`DotPageContentTypeService.getAllContentTypes`). Building it surfaced two gaps in the global content-type list endpoint that currently force client-side workarounds in the frontend.

### Issue 1 — No way to exclude system content types
The global list returns non-creatable **system** content types (e.g. **Favorite Page** `dotFavoritePage`, **Host**) that share a base type with real content (Favorite Page is baseType `CONTENT`), so the `type` base-type filter cannot exclude them.

Frontends work around this client-side: `dot-content-drive-content-type-filter` strips `!ct.system && ct.baseType !== FORM`, and the new Content Drive picker drops Favorite Page with a `-1` count adjustment. Both break/approximate pagination counts.

**Ask:** an optional `system` query param (e.g. `system=false`) on `getContentTypes`, defaulting to current behavior (backward compatible — no param ⇒ unchanged).

The SQL layer already supports a boolean `system` safe-condition: `isBooleanField()` in `ContentTypeFactoryImpl` includes `"system"`, the `SafeCondition` machinery parameterizes it, and the count path applies the same condition (so totals stay exact). The main work is threading a dedicated param through `ContentTypeResource.getContentTypes` → `ContentTypesPaginator.getItems` → `searchMultipleTypes`/`search` + their count loops, because `SearchCondition` currently treats its input as EITHER a name-search term OR a condition (not both), and the picker needs name-search + system filter together.

### Issue 2 — Multi-type results are not globally sorted
When multiple `type` filters are passed, results are grouped per base type (each group sorted by name) and paginated per group — so content types reappear on later pages (e.g. a "B…" type shows up again on page 2 after the first group's "B…" were exhausted). `orderby=name` should sort the entire result set across base types.

Source: the multi-type UNION path in `ContentTypesPaginator.getItems` / `ContentTypeFactoryImpl.searchMultipleTypes` and its per-type count.

### Impact
Content Drive's content-type picker currently filters Favorite Page client-side (with a `-1` count workaround) and shows grouped (non-global) ordering on "All Content Types". Fixing these server-side lets the frontend drop the workarounds and keeps pagination counts exact.

## Acceptance Criteria

**Issue 1 — system exclusion**
[...]

Issue #36028, opened 2026-06-05
[TASK] Content Drive: Surface archived content at archive/resolved workflow steps in Browser/Drive search (DB-first, per-step)


zJaaal:
### Problem
When content is filtered by a **workflow step**, content that has been **archived** and parked at an **archive / resolved step** is **invisible**. The shared SQL builder [`BrowserAPIImpl.selectQuery`](dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java) appends a global `cvi.deleted = 'false'`, which removes archived rows *before* the workflow filter is evaluated — so a step that holds archived content (e.g. an "Archive" step with `workflow_step.resolved = true`, reached by an action running `ArchiveContentActionlet`) can never list that content.

This is a **shared-layer** limitation affecting **both** browsing endpoints that go through `selectQuery`:
- `POST /api/v1/drive/search` — Content Drive workflow filter (#35880 / #35906)
- `/v1/browser` — Site Browser

It is **pre-existing and platform-wide**, not introduced by the workflow filter PR: the legacy workflow task view (the JSP skips rows whose contentlet won't hydrate) and Site Browser both hide archived-at-step content. Net effect: an "Archive" workflow step is effectively unbrowsable.

### Proposed approach — per-step, DB-first
When a pinned step is an **archive-target step**, include archived content for **that step's branch only**; keep scheme-only and normal-step matches **live-only**. Restructure `appendWorkflowQuery` to own the `deleted` condition per branch and **suppress the global `cvi.deleted='false'`** only when an archive step is present:

```sql
and (
   ( cvi.deleted = 'false' and (
        exists (… workflow_scheme_x_structure … scheme_id in (?))     -- scheme-only: live
        or exists (… workflow_task … status in (?,?))                 -- normal steps: live
   ) )
   or
   exists (… workflow_task … status in (?))                           -- archive steps: archived allowed
)
```

**Detect archive-target steps** from the filter's `{scheme, step}` pairs: a step is archive-target if an action whose `nextStep` = that step carries `ArchiveContentActionlet`. Resolve **once per request** via the cached `WorkflowAPI` (`findActions(scheme)` → `findActionClasses(action)`), never per row — config tables are tiny and cached, so the cost is negligible.

### Why NOT a global `showArchived` flip
[...]

PR #36079, opened 2026-06-09
feat(content-drive): New button menu with content type selection (#35574)

### LINKED ISSUE #35574: [TASK] Content Drive: "New" button menu with type selection
### Description

Implement the new **"New"** button menu in Content Drive's toolbar, plus split **Upload** out as its own separate button on the side (as shown in the prototype). The Upload workflow itself is preserved as-is in this ticket — full Upload changes are tracked in epic #35436.

[Prototype video](https://dotcms.slack.com/archives/C08MTB0RH6D/p1778000582312239)

#### New button

Clicking **New** opens a menu with three sections:

1. **All Content Types** — opens a content type browser dialog.
2. **Base Types list** — every base type **except Form**.
3. **Folder** — keeps the current behavior.

Image

Image

Behavior:

- Selecting **All Content Types** or any **base type** opens a dialog containing a content type list. The list must **reuse the UVE Palette List** (`libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-palette/components/dot-uve-palette-list`) — same design, same functionality. Share what's needed instead of duplicating.

Image

- When a base type is preselected, the list opens filtered by that base type.
- After the user selects a content type and clicks **Create**, navigate to **Edit Content** the same way Content Drive currently handles editing a contentlet.
- Selecting **Folder** preserves the current folder creation behavior — no changes.

#### Upload button placement

- **Upload must be its own button on the side**, not an option inside the **New** menu (per prototype).
- The current upload workflow is **preserved as-is** in this ticket. Behavior changes (Asset/File prompt, endpoint param, folder preferences, drag-and-drop, etc.) are tracked under epic #35436.

### Acceptance Criteria

- [ ] **New** button opens a menu with: "All Content Types", base type list (excluding Form), and "Folder".
- [ ] Selecting "All Content Types" opens a dialog with the content type list reused from the UVE Palette.
- [ ] Selecting a base type opens the same dialog filtered by that base type.
- [ ] Selecting a content type and clicking "Create" navigates to Edit Content using Content Drive's existing edit flow.
- [ ] Selecting "Folder" preserves the current folder creation behavior.
- [ ] **Upload** is rendered as a separate button on the side and preserves the current upload workflow.
- [ ] UVE Palette content type list is shared/reused — no copy/paste duplication.

#### Nice to Have

- Open **New Edit Content** from the Create flow.
- Full reuse of the UVE Palette (filtering, search, presentation), not just visual replication.

#### Out of Scope

- Any change to the Upload behavior itself (tracked in epic #35436).

### Priority

Medium

### Additional Context

- Toolbar: `libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar`
- UVE Palette List to reuse: `libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-palette/components/dot-uve-palette-list`
- Existing folder dialog reference: `libs/portlets/dot-content-drive/portlet/src/lib/components/dialogs/dot-content-drive-dialog-folder`

Parent epic: #35438

zJaaal:
### Proposed Changes

Implements the Content Drive **"New"** button menu with content-type selection (#35574), reusing the UVE palette list.

* **Reusable palette list** — moved the UVE palette list family (list, content-type/contentlet cards, favorite-selector, models, utils) into `@dotcms/portlets/dot-ema/ui` and decoupled its store from `UVEStore` (allowed content types now passed in as data). UVE behavior is unchanged.
* **Selection mode** — the palette list gains an opt-in `selectionMode`: cards become clickable (no drag handle/chevron, not draggable), single-select highlight, favorites/grid-list toggle hidden, and it emits the picked content type instead of drilling into contentlets.
* **"New" menu** — toolbar menu with *All Content Types*, one entry per base type (all except deprecated Form, with Material Symbols icons), and *Folder* (unchanged). **Upload** is now its own standalone button.
* **Selector dialog** — opening a menu option shows the reused palette list in a dialog (3 cards/row, padded content, full-width pagination/footer separators); choosing a type + **Create** navigates to the content editor.
* **Navigation** — `createContent` mirrors the existing edit flow: new editor when `CONTENT_EDITOR2` is enabled for the type, else the legacy editor. The legacy create route now resolves from Content Drive, and (like editing) carries `CD_`-prefixed params so closing returns to Content Drive with filters preserved.
* **Card name tooltip** — content-type cards show the full name on hover (parity with the contentlet card from #36028).
* i18n keys added to `Language.properties`.

https://github.com/user-attachments/assets/9bddb198-6802-447f-8e85-55ea63185cb5

### Checklist
- [ ] Tests
- [ ] Translations

### Additional Info

* Favorite Page (a system content type that shares the CONTENT base type) is intentionally **not** filtered on the frontend — it's handled server-side via backend issue #36072 (also covers global sort across base types for the global content-type endpoint).
* `dotContentletEditorRoutes` now provides `DotCreateContentletResolver` so the legacy create route resolves from any entry point (its only dep is already `providedIn: 'root'`).

Closes #35574

This PR fixes: #35574


core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.ts
```diff
@@ -106,10 +116,64 @@ export class DotContentDriveShellComponent {
...
+
+    /**
+     * Syncs the dialog open/close state from the store. Opening sets the body and visibility
+     * together (no blank-frame flash); closing flips visibility off but leaves the body mounted
+     * so PrimeNG can animate it out — the body is cleared later in {@link onDialogHidden}.
+     */
+    readonly syncDialogEffect = effect(() => {
+        const dialog = this.#store.dialog();
+
+        untracked(() => {
```
nicobytes: I suggest using the `signalMethod` to avoid the manual untrack

core-web/libs/portlets/edit-ema/ui/src/lib/palette/components/dot-uve-palette-contenttype/dot-uve-palette-contenttype.component.html
```diff
@@ -0,0 +1,58 @@
...
+        <i class="pi pi-ellipsis-v -ml-2 text-[14px]"></i>
+    </div>
+}
+
+<!-- Content type icon and name -->
+<div
+    class="flex min-w-0 flex-1 items-center justify-center"
+    [pTooltip]="'uve.palette.item.disabled.tooltip' | dm"
+    [tooltipDisabled]="!contentType.disabled"
+    [ngClass]="$isListView() ? 'flex-row justify-start gap-3' : 'flex-col gap-3'">
```
nicobytes: ngClass will be deprecated; use [class]

core-web/apps/dotcms-ui/src/app/view/components/dot-contentlet-editor/components/dot-create-contentlet/dot-create-contentlet.component.ts
```diff
@@ -52,8 +53,27 @@ export class DotCreateContentletComponent implements OnInit {
...
+            );
+            const contentDriveParams = mapParamsFromEditContentlet(searchParams);
+
+            if (Object.keys(contentDriveParams).length) {
+                this.dotRouterService.gotoPortlet('content-drive', {
+                    queryParams: contentDriveParams
+                });
+            } else {
+                this.dotRouterService.goToContent();
+            }
```
fmontes, line 76: is there any way that we can do this smarter? like using the browser history or something?

core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.ts
```diff
@@ -106,10 +116,64 @@ export class DotContentDriveShellComponent {
...
+
+    /**
+     * Content-type selector: sized to fit ~4 UVE-width cards per row. No horizontal padding so
+     * the paginator/footer separators span edge-to-edge; the list and footer add their own inset.
+     */
+    readonly $dialogContentClass = computed(() =>
+        this.$activeDialog()?.type === DIALOG_TYPE.CONTENT_TYPE_SELECTOR
+            ? 'w-[min(92vw,38rem)] px-0! pt-0 pb-4'
+            : 'w-[43.75rem] pt-0 p-4'
+    );
```
fmontes, line 158: https://primeng.org/classnames

core-web/libs/portlets/edit-ema/ui/src/lib/palette/components/dot-uve-palette-contenttype/dot-uve-palette-contenttype.component.html
```diff
@@ -0,0 +1,58 @@
+@let contentType = $contentType();
+
+<!-- Drag handle (hidden in selection mode) -->
+@if (!$selectable()) {
+    <div
+        class="drag-handle flex size-8 cursor-grab items-center justify-center rounded-sm text-gray-400 opacity-0 transition-opacity select-none group-hover:text-(--color-palette-primary-600) hover:bg-black/5 hover:text-gray-500 active:cursor-grabbing"
+        [ngClass]="
+            contentType.disabled
+                ? 'pointer-events-none cursor-not-allowed opacity-0'
+                : 'group-hover:opacity-100'
```
fmontes: https://primeng.org/classnames

fmontes (changes requested):

core-web/libs/portlets/edit-ema/ui/src/lib/palette/components/dot-uve-palette-contenttype/dot-uve-palette-contenttype.component.html
```diff
@@ -0,0 +1,58 @@
+@let contentType = $contentType();
+
+<!-- Drag handle (hidden in selection mode) -->
+@if (!$selectable()) {
+    <div
+        class="drag-handle flex size-8 cursor-grab items-center justify-center rounded-sm text-gray-400 opacity-0 transition-opacity select-none group-hover:text-(--color-palette-primary-600) hover:bg-black/5 hover:text-gray-500 active:cursor-grabbing"
+        [ngClass]="
```
oidacra: This template uses `[ngClass]` in four places (L7, L22, L35, L47). Per `ANGULAR_STANDARDS.md` ("Do NOT use ngClass — use class bindings"), and since this component is being rewritten/moved here, please switch to `[class]` object bindings or `[class.x]`, e.g. `[class]="{ 'flex-row justify-start gap-3': $isListView(), 'flex-col gap-3': !$isListView() }"`.

core-web/libs/portlets/edit-ema/ui/src/lib/palette/components/dot-uve-palette-contenttype/dot-uve-palette-contenttype.component.ts
```diff
@@ -72,6 +86,14 @@ export class DotUVEPaletteContenttypeComponent {
         this.onSelectContentType.emit(contentType.variable);
     }
 
+    @HostListener('click')
```
oidacra: The newly added `@HostListener('click')` (and the existing `@HostListener('contextmenu')`) should move into the `host` object of `@Component`, per `ANGULAR_STANDARDS.md` ("Do NOT use `@HostListener` — use the `host` object"):

```ts
host: {
  '(click)': 'onHostClick()',
  '(contextmenu)': 'onContextMenu($event)'
}
```

Then drop the `HostListener` import. See Angular's own guidance: https://angular.dev/api/core/HostListener

core-web/libs/portlets/edit-ema/ui/src/lib/palette/components/dot-uve-palette-contenttype/dot-uve-palette-contenttype.component.ts
```diff
@@ -46,12 +55,17 @@ export class DotUVEPaletteContenttypeComponent {
...
 
         const disabled = isDisabled ? 'cursor-not-allowed opacity-60 pointer-events-none' : '';
+        const selectableClass = selectable && !isDisabled ? 'cursor-pointer' : '';
+        const selectedClass =
+            selectable && this.$selected()
+                ? 'border-primary-500 bg-primary-100 shadow-sm'
+                : 'border-gray-200';
 
-        return `${base} ${this.$isListView() ? list : grid} ${disabled}`;
+        return `${base} ${this.$isListView() ? list : grid} ${disabled} ${selectableClass} ${selectedClass}`;
```
oidacra: `$hostClass` builds the Tailwind classes as a concatenated string in TS. Dynamically assembled class strings can sometimes escape Tailwind's static detection (purge risk in prod builds). Would it be worth returning a class-map object and binding `[class]="$hostClass()"` instead — e.g.

```ts
readonly $hostClass = computed(() => ({
  'border-primary-500 bg-primary-100 shadow-sm': this.$selectable() && this.$selected(),
  'border-gray-200': !(this.$selectable() && this.$selected()),
  'cursor-pointer': this.$selectable() && !this.$isDisabled(),
}));
```

Open to your call if you've already confirmed these utilities are safelisted.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/dot-content-drive-toolbar.component.ts
```diff
@@ -134,12 +208,19 @@ export class DotContentDriveToolbarComponent {
         minWidth: this.$treeExpanded() ? '0' : undefined
     }));
 
+    /** Pending animation-sequencing timer; cleared on each transition so rapid toggles don't race. */
+    #animationTimeout: ReturnType<typeof setTimeout>;
```
oidacra: `#animationTimeout` is declared `ReturnType<typeof setTimeout>` with no initializer, so it's `undefined` until first assigned. It compiles today because `strict` isn't enabled in the inherited tsconfig — but if/when `strict` (→ `strictPropertyInitialization`) is turned on, this would fail with TS2564 ("Property has no initializer and is not definitely assigned in the constructor"). Suggest declaring it `ReturnType<typeof setTimeout> | undefined` now to be future-proof. Minor / preventive.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/shared/services/dot-content-drive-navigation.service.ts
```diff
@@ -46,6 +54,51 @@ export class DotContentDriveNavigationService {
...
 
+    /**
+     * Navigates to the content editor to CREATE a new content of the given type.
+     * Mirrors the edit flow ({@link editContent}): the new content editor is only used when it
+     * is enabled for the selected content type (CONTENT_EDITOR2 flag); otherwise it falls back
+     * to the legacy create editor.
+     *
+     * @param contentTypeVariable - The variable name of the content type to create
+     */
+    createContent(contentTypeVariable: string) {
```
oidacra: Should `createContent` declare an explicit return type here? As a public service method (an API boundary), adding `: void` — `createContent(contentTypeVariable: string): void` — would align with `TYPESCRIPT_STANDARDS.md` (explicit types at public boundaries).

core-web/apps/dotcms-ui/src/app/view/components/dot-contentlet-editor/components/dot-create-contentlet/dot-create-contentlet.component.spec.ts
```diff
@@ -127,6 +127,21 @@ describe('DotCreateContentletComponent', () => {
...
     });
 
+    it('should emit shutdown and redirect to Content Drive with un-prefixed params when coming from Content Drive', () => {
+        jest.spyOn(routerService, 'currentSavedURL', 'get').mockReturnValue('/c/content/new/');
+        jest.spyOn(routerService, 'currentPortlet', 'get').mockReturnValue({
+            url: 'c/content/new/blog?CD_path=/foo&CD_filters=bar',
+            id: 'content'
+        } as any);
+        spectator.detectChanges();
+        spectator.component.onClose({});
```
oidacra: This calls `spectator.component.onClose({})` directly instead of triggering the `close` output on the wrapper child. Per `TESTING_REVIEW_RULES.md` ("trigger via DOM/output, not `component.method()`"), prefer `spectator.triggerEventHandler(DotContentletWrapperComponent, 'close', {})`.

Note the risk: as written, if the DOM wiring that actually fires `close` were removed, this test would **still pass** — so it doesn't guard the integration, only that the method runs.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/shared/services/dot-content-drive-navigation.service.spec.ts
```diff
@@ -199,6 +209,79 @@ describe('DotContentDriveNavigationService', () => {
...
+                metadata: { [FeaturedFlags.FEATURE_FLAG_CONTENT_EDITOR2_ENABLED]: false }
+            });
+
+            location.path.mockReturnValue('/content-drive?path=/foo&filters=bar');
+
+            contentTypeService.getContentType.mockReturnValue(of(mockContentType));
+
+            service.createContent('news');
+
+            expect(contentTypeService.getContentType).toHaveBeenCalledWith('news');
```
oidacra, line 256: The PR adds symmetric `catchError` + `httpErrorManager.handle` error handling to both `createContent` and `#editContentlet`, but only `createContent`'s error path is covered by tests. Could we add an `editContent` test where `getContentType` throws, asserting `httpErrorManager.handle` is called and `router.navigate` is not? Otherwise the edit-path error handling is unguarded against regressions.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-palette/dot-uve-palette.component.spec.ts
```diff
@@ -38,6 +37,7 @@ const mockUVEStore = {
     pageURI: signal('/test/page/path'),
     pageLanguageId: signal(1),
     pageVariantId: signal('DEFAULT'),
+    $allowedContentTypes: signal<Record<string, true>>({}),
```
oidacra, line 40: This PR adds `[allowedContentTypes]="$allowedContentTypes()"` to the Favorites tab's palette-list — the one behavioral change in this component — but the "inputs passed to Favorites tab" test still asserts only `listType/languageId/pagePath/variantId`. Could we assert `allowedContentTypes` is threaded through (set a non-empty map on the mock and verify it reaches the palette-list input)? Right now the new binding is unguarded against being dropped or wired to the wrong store property.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.ts
zJaaal: Nice catch!

core-web/libs/portlets/edit-ema/ui/src/lib/palette/components/dot-uve-palette-contenttype/dot-uve-palette-contenttype.component.html
zJaaal: nice catch!

core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.ts
zJaaal: TIL

core-web/apps/dotcms-ui/src/app/view/components/dot-contentlet-editor/components/dot-create-contentlet/dot-create-contentlet.component.ts
zJaaal: This is a temporal solution, I want to implement the dialog with the new edit content, using the browser history can leave us on unwanted states if they hard reload or enter the editor from somewhere else

zJaaal: Good question. We kept the `CD_`-prefixed params instead of `Location.back()`: the legacy editor is reached via a router navigation so `back()` works for the happy path, but the params also survive a **reload/deep-link** (where the SPA history stack is empty) and keep this in sync with the edit flow (`DotContentletWrapperComponent.onClose`), which already uses the same mechanism. Left as-is for robustness + parity; happy to revisit simplifying both flows together as a follow-up.

— 🤖 _Claude (via Claude Code)_

core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.ts
zJaaal: Applied the PrimeNG classnames approach where it earns its keep — the contenttype card `$hostClass` now uses `cn()` (tailwind-merge) and the icon container uses `[pClass]`. `$dialogContentClass` here is a passthrough class **string** with two mutually-exclusive alternatives (no base+conditional conflict to merge), so I left it as a plain ternary.

— 🤖 _Claude (via Claude Code)_

core-web/libs/portlets/edit-ema/ui/src/lib/palette/components/dot-uve-palette-contenttype/dot-uve-palette-contenttype.component.html
zJaaal: Done where it adds value: the icon container now uses `[pClass]` so the list-view `justify-start` merges over the base `justify-center`. The drag handle is `opacity-0 pointer-events-none` when disabled, so its cursor conflict is never visible — left as `[class]`.

— 🤖 _Claude (via Claude Code)_

core-web/libs/portlets/edit-ema/ui/src/lib/palette/components/dot-uve-palette-contenttype/dot-uve-palette-contenttype.component.html
zJaaal: Done — all four `[ngClass]` replaced with `[class]` (and `[pClass]` on the icon container, where the base/conditional `justify` utilities actually conflict).

— 🤖 _Claude (via Claude Code)_

core-web/libs/portlets/edit-ema/ui/src/lib/palette/components/dot-uve-palette-contenttype/dot-uve-palette-contenttype.component.ts
zJaaal: Done — moved `click`/`contextmenu` into the `host` object and dropped the `HostListener` import.

— 🤖 _Claude (via Claude Code)_

core-web/libs/portlets/edit-ema/ui/src/lib/palette/components/dot-uve-palette-contenttype/dot-uve-palette-contenttype.component.ts
zJaaal: Done — `$hostClass` now builds with `cn()` (tailwind-merge): base + state classes combine into one string and conflicts resolve (selected `bg-primary-100`/`border-primary-500` override base `bg-white`/`border`).

— 🤖 _Claude (via Claude Code)_

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/dot-content-drive-toolbar.component.ts
zJaaal: Done — typed as `ReturnType<typeof setTimeout> | undefined`.

— 🤖 _Claude (via Claude Code)_

core-web/libs/portlets/dot-content-drive/portlet/src/lib/shared/services/dot-content-drive-navigation.service.ts
zJaaal: Done — added explicit `: void`.

— 🤖 _Claude (via Claude Code)_

core-web/apps/dotcms-ui/src/app/view/components/dot-contentlet-editor/components/dot-create-contentlet/dot-create-contentlet.component.spec.ts
zJaaal: Done — the test now drives the wrapper's `(shutdown)` output via `triggerEventHandler` instead of calling `onClose()` directly, so it guards the wiring.

— 🤖 _Claude (via Claude Code)_

core-web/libs/portlets/dot-content-drive/portlet/src/lib/shared/services/dot-content-drive-navigation.service.spec.ts
zJaaal: Done — added an `editContent` error-path test: `getContentType` throws → asserts `httpErrorManager.handle` is called and `router.navigate` is not.

— 🤖 _Claude (via Claude Code)_

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-palette/dot-uve-palette.component.spec.ts
zJaaal: Done — added a test asserting `allowedContentTypes` is threaded from the store to the Favorites palette list (non-empty map reaches the `allowedContentTypes` input).

— 🤖 _Claude (via Claude Code)_

nicobytes (approved):

oidacra (approved):

fmontes (approved):

zJaaal:
Thanks — triaged the final pass:

**Fixed (`13434538d3`):**
- ✅ **Debug artifact** — removed the `Default dialog` `@default` from the shell dialog `@switch`.
- ✅ **Static `[class]` literal** — folded the `grid-cols-[…]` literal into the toolbar div's static `class`.

**Confirmed already resolved** (as the review notes): `#animationTimeout | undefined`, `createContent(): void`, and the `editContent` error-path test.

**Intentionally not changed (with reasoning):**
- **`onClose` fall-through to `reloadData`** — keeping `onClose` as-is per a prior call. The naive early-`return` the bot suggests would also skip `shutdown.emit()`, which the edit counterpart (`DotContentletWrapperComponent.onClose`) always emits — so it'd diverge. The extra `reloadData` runs against a portlet the router is already tearing down (benign). A proper fix (emit `shutdown` first, then return) is a follow-up touching both create + edit flows.
- **`DotPaletteListStore` `console.error`** — this is pre-existing **UVE** store behavior (the store was moved, not rewritten). Routing it through `DotHttpErrorManagerService` changes UVE's error UX and should be a focused change with UVE owners, not bundled into this Content Drive PR.
- **`$items` "dead signal"** — labels resolve via `DotMessageService.get()` at construction; dotCMS loads i18n at bootstrap before portlets render (same as the existing folder menu). Note neither a plain field nor `computed()` would restore language-change reactivity here (`get()` isn't reactive — the real fix is the `| dm` pipe), so the change would be cosmetic. Left as-is.
- **`@ViewChild` → `viewChild()`** — pre-existing UVE internals in the moved palette list; the modernization is desirable but carries real regression risk to palette behavior (`menu`/`contextMenu`/`favoritesPanel` call sites). Better as a dedicated follow-up.

— 🤖 _Claude (via Claude Code)_

PR #36081, opened 2026-06-09
ci: build CLI only when it changes, reuse prior release otherwise

### LINKED ISSUE #36080: CI: only build the CLI when it changes; reuse prior release otherwise
## Problem

The dotCMS CLI is rebuilt and re-released on every trunk push and every dotCMS release, even when `tools/dotcms-cli/**` is unchanged. The native build runs an expensive 3-OS GraalVM matrix (including two macOS runners), so we burn that cost repeatedly to ship byte-identical binaries.

## Proposed change

Gate the CLI build on actual CLI source changes, and reuse the prior release when unchanged.

**Trunk (`cicd_3-trunk.yml`)**
- Run `build-cli` only when the `cli` path filter matches; gate `deploy-cli` on the same predicate.

**Release (`cicd_release-cli.yml`)**
- Detect whether `tools/dotcms-cli/**` changed since the previous release tag (fail-safe: build when uncertain).
- When unchanged, skip the native build + JReleaser and republish the previous release's binaries under the new version across all channels — npm, the `dotcms-cli-<version>` GitHub release, and Artifactory.
- Remove the Slack announcement from the CLI release cycle.

The CLI is self-contained under `tools/dotcms-cli/**` (its only `com.dotcms` dependency lives in that subtree), so the path filter is a correct change-key.

wezell:
## What

Performance Optimization: Build/release the dotCMS CLI only when it actually changes, and reuse the prior release otherwise.
Trying to make our automated tests and release actions as fast as possible.

Fixes #36080

## Why

The CLI is rebuilt and re-released on every trunk push and every dotCMS release, even when `tools/dotcms-cli/**` is byte-identical to the prior release. The native build runs an expensive 3-OS GraalVM matrix (including two macOS runners), so we repeatedly pay that cost to ship identical binaries.

The CLI is self-contained under `tools/dotcms-cli/**` (its only `com.dotcms` dependency, `dotcms-api-data-model`, lives in that same subtree), so a path/diff change-key is correct.

## Changes

**Trunk — `cicd_3-trunk.yml`**
- `build-cli` now also requires `fromJSON(needs.initialize.outputs.filters).cli == 'true'`.
- `deployment` passes `deploy-cli` gated on the same predicate, so it never tries to deploy artifacts that were not built.

**Release — `cicd_release-cli.yml`**
- New detection in `precheck`: `git diff <prev-release-tag>..<this-tag> -- tools/dotcms-cli/` sets `CLI_CHANGED`. **Fail-safe:** any uncertainty (manual dispatch, no previous tag) → `CLI_CHANGED=true` (full build), so a wrong "unchanged" verdict can never ship stale binaries.
- `build`, `build-cli` (native), and `release` (JReleaser) are gated on `CLI_CHANGED == 'true'`.
- **Unchanged path republishes the prior release's binaries under the new version across all three channels** JReleaser/the pipeline feed:
  - **npm** (`publish-npm-package`): repackage the published `latest` tarball under the new version.
  - **GitHub release + Artifactory** (new `reuse-release` job): download `dotcms-cli-<prev>` assets, re-stamp to the new version, create the `dotcms-cli-<new>` release, and upload to `libs-release-local/com/dotcms/dotcms-cli/<new>/`.
- Removed the Slack announcement from the CLI release cycle.

## Notes / caveats

- The `reuse-release` job mirrors the channels in `tools/dotcms-cli/jreleaser.yml` in bash — keep the two in sync if platforms or the Artifactory path change (noted in-code).
- On the reuse path, JReleaser-generated checksums/signatures are not reproduced (only the `.zip`/`.jar` artifacts), and the reused binaries internally report the prior version (inherent to reusing them).
- These are workflow-logic changes not exercised locally. The `reuse-release` and npm-reuse steps publish irreversibly and have no dry-run guard; recommend a `workflow_dispatch` rehearsal before relying on the unchanged path in a real release (can add a `dry-run` short-circuit if wanted).


sfreudenthaler (commented):
**Acknowledged — won't fix** (re: `npm view "${PACKAGE_FULL_NAME}" version` in the "Reuse previous CLI release and publish" step)

The feedback notes that `npm view ... version` returns the `latest` tag version rather than a version derived from `PREV_RELEASE_VERSION`, and suggests deriving it from the git-based output instead.

This was a deliberate choice, documented in the inline comment above that step:

> *"We read the source version from npm ('latest') rather than reconstructing it from the git tag, sidestepping the leading-zero normalization applied to release versions above."*

Deriving from `PREV_RELEASE_VERSION` would require replicating the same leading-zero stripping logic from the "Dynamic configuration" step — adding complexity with no practical benefit. This workflow is the sole publisher to the `latest` npm tag (LTS releases are explicitly excluded, and no other code path publishes there), so the two values are guaranteed to agree under normal operation. The only divergence scenario is out-of-band manual intervention on npm, which is outside the scope of this workflow to defend against.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #36092, opened 2026-06-09
perf(relationships): bulk tree persistence, lean hydration and delta-only reindex when saving related content

### LINKED ISSUE #36147: Saving contentlets with many-to-many relationship fields and large related-content volumes is extremely slow
## Problem Statement

Saving a contentlet whose Content Type has several **relationship fields with MANY_TO_MANY cardinality** becomes extremely slow once the volume of related content grows. On a real-world dataset (a parent type with ~1,500 contentlets, each related to ~20 children of another type, which means each child is related back to ~350 parents on average), a single **Save/Publish takes from 40 seconds up to several minutes**, generating tens of thousands of JDBC queries per save. The slowdown compounds over time: the more related content exists, the slower every save gets.

This blocks day-to-day content authoring for any installation that models catalog-style data (products ↔ categories, offers ↔ cards, articles ↔ tags, etc.) through many-to-many relationship fields.

### Root causes identified

1. **Per-row relationship persistence** — `ESContentletAPIImpl.deleteRelatedContent`/`relateContent` delete and re-insert `tree` rows one at a time (a `getTree` + `saveTree` pair per related record: 2–3 queries each).
2. **Unnecessary full hydration** — the save path loads every related contentlet (with all its fields, permissions, and version info) when only identifiers or counts are needed: `dbRelatedContent` for the delete scope, `getRelatedContentFromIndex` just to compute a tree-order base, and `getRelatedContent` inside the reindex-dependency calculation.
3. **Over-broad dependency reindexing** — `ESMappingAPIImpl.dependenciesLeftToReindex` marks **every currently-related contentlet** for reindexing on every save (the old-state/new-state comparison degenerates because both sides read the new DB state), instead of only the relationships that were actually added or removed. Saving one child re-queues hundreds of unchanged parents, each of which is fully hydrated and remapped.

## Steps to Reproduce

1. Create two Content Types, `ParentType` and `ChildType`, related through a relationship field with **MANY_TO_MANY** cardinality.
2. Bulk-create ~1,500 `ParentType` contentlets and ~90 `ChildType` contentlets, relating each parent to ~20 children (so each child ends up related to hundreds of parents).
3. Edit any `ChildType` contentlet (or a `ParentType` one related to several children) and Save/Publish.
4. Observe the save time and the number of JDBC queries (e.g., via Glowroot on port 4000).

**Expected:** the save completes in a bounded time, proportional to the relationships actually changed.
**Actual:** the save takes from ~40 seconds (3 related items) to 2–10 minutes (10+ related items on a large dataset), with tens of thousands of queries against `tree`, `contentlet`, and `contentlet_version_info`.

## Acceptance Criteria

- [ ] Saving a contentlet with relationship fields performs bulk/batched `tree` operations — no per-row delete/select/insert per related record.
- [ ] Related-content identifiers are read without hydrating full contentlets when only identifiers are needed; only contentlets that actually require reindexing are loaded.
- [ ] `dependenciesLeftToReindex` returns only the **delta** (relationships added or removed by the save); contentlets whose relationships did not change are not re-queued for reindexing.
- [ ] Permission semantics are preserved: relationships to content the saving user cannot READ are not deleted by the save.
- [ ] Duplicate related records in the payload (same identifier twice, or multiple language versions) do not fail the save with a primary-key violation.
- [ ] Relationship ordering (`tree_order`) is preserved for both changed and preserved rows.
- [ ] Multilingual related content (multiple index documents per identifier) does not trigger spurious reindexing.
- [ ] Covered by integration tests (TreeFactory operations, permission preservation, duplicate handling, reindex delta, multilingual dedup).
- [ ] Fix is backported to the 24.12.27 LTS.

## dotCMS Version

Reproduced on **24.12.27 LTS** and current `main`. Likely affects all currently supported versions.

## Severity

## Links

- Internal support ticket: [#36004](https://helpdesk.dotcms.com/helpdesk/tickets/36004)
- Proposed fix: PR #36149

jcastro-dotcms:
This PR fixes: #36147

### 🎯 Problem

Saving a contentlet whose Content Type has several **MANY_TO_MANY relationship fields** became unusably slow as related-content volume grew: on a real-world dataset (~1,500 parents × ~20 children each, so each child relates back to ~350 parents), a single Save/Publish took **from 40 seconds up to several minutes** and fired **tens of thousands of JDBC queries**. Three compounding root causes:

1. **Per-row relationship persistence** — every related record cost a `getTree` + `saveTree` pair (2–3 queries each) plus one `DELETE` per row.
2. **Unnecessary full hydration** — the save path loaded complete contentlets (fields, permissions, version info) when it only needed identifiers or a count.
3. **Over-broad dependency reindexing** — `dependenciesLeftToReindex()` re-queued **every currently-related contentlet** for reindexing on every save, because the old-state/new-state comparison read the new DB state on both sides. Saving one child re-indexed hundreds of unchanged parents, each fully hydrated and remapped.

### ✨ Proposed Changes

**Relationship persistence — bulk, batched, and permission-aware (`ESContentletAPIImpl`, `TreeFactory`)**
* `deleteRelatedContent` now splits into two paths: a fast identifier-only path for the system user / CMS Admins (single bulk `DELETE`, zero hydration, only *removed* parents loaded for reindex), and a permission-scoped path for limited users that preserves the legacy contract — **relationships to content the user cannot READ are never deleted** (bulk delete when the filter removes nothing; chunked `IN` deletes otherwise).
* `relateContent` collects all rows and persists them through a new `TreeFactory.insertTrees`: a duplicate-safe batched **upsert** (dedup by the `(child, parent, relation_type)` PK, batched `DELETE`+`INSERT` wrapped in `LocalTransaction` so it is atomic for any caller). Duplicate records in the payload (same identifier twice, multiple languages) no longer risk a PK violation.
* The tree-order base comes from a single `MAX(tree_order)+1` lookup over surviving rows (replacing a full `getRelatedContentFromIndex` hydration), queried only on the parent branch where it is actually used.

**Hydration elimination (`ESContentletAPIImpl`, `VersionableFactory/Impl`)**
* Related identifiers are read straight from the `tree` table via new `TreeFactory` identifier lookups — no contentlet hydration to compute delete scopes or counts.
* Removed parents that genuinely need reindexing are loaded through a new **batched, uncached** `VersionableFactory.findAllContentletVersionInfos(Collection)` (chunked `IN`, 500 ids/statement) + the existing batch `findContentlets`, instead of one lookup per identifier. All SQL lives in factory classes per house architecture.

**Delta-only dependency reindexing (`ESMappingAPIImpl`)**
* `dependenciesLeftToReindex()` now compares the **search index** (pre-save state) against the **`tree` table** (post-save state) and returns their true **symmetric difference**: only relationships actually added or removed. Unchanged parents are never re-queued — this is the biggest win for heavily related datasets.
* Identifiers are deduped (the index holds one document per language/variant), so multilingual content no longer triggers spurious reindexing; a scroll-search fallback removes the silent 10,000-document truncation.

**Hardening (review feedback)**
* `assertTreeColumn` allowlist guard on every `TreeFactory` helper that interpolates a column name into SQL (only `parent`/`child` accepted) — enforces what the Semgrep findings flagged conventionally.
* Javadoc pass: class-level doc for `TreeFactory`, contracts on the bulk deletes (caller-enforced permissions), and a full explanation of the new reindex-delta model on `dependenciesLeftToReindex`.

### ✅ Checklist
- [x] Tests
- [ ] Translations (N/A — backend only)

**Security notes:** permission semantics are preserved and now covered by an integration test (a limited user's save cannot delete relationships to content they cannot READ — this also fixes a data-loss risk present in the first draft of this PR). The Semgrep `CUSTOM_INJECTION-2` findings are false positives (concatenated fragments are compile-time literals or generated `?,?,...` placeholders; all user data is bound parameters) and are additionally guarded by the new column-name allowlist.

### 🧪 Testing

11 new integration tests, all green:
* `TreeFactoryTest` (new class, 7 tests): batch upsert dedup, pre-existing row replacement, scoped deletes, >500-id chunking, ordered id lookups, next-tree-order.
* `ESContentletAPIImplTest` (+3): limited-user permission preservation, duplicate-records save, append-after-preserved ordering. Full class: **90/90 passing**.
* `ESMappingAPITest` (+2): reindex delta returns only added+removed parents; multilingual parents are not spuriously re-indexed.

### ℹ️ Additional Info

* Verified against a production-scale dataset: re-saving heavily related content drops from minutes to seconds, with query counts reduced by orders of magnitude (bulk statements + delta-only reindex instead of per-row queries + full-graph reindex).
* A manual backport to **24.12.27 LTS** is prepared in parallel (adapted for `-source 11`), including the test suite.
* Behavioral notes for reviewers: orphan `tree` rows (pointing to fully-deleted content) are now cleaned up by the bulk delete; `tree_order` values are rebased per save (relative ordering preserved — consumers only `ORDER BY tree_order`).

### Screenshots
N/A — backend-only performance change.


jcastro-dotcms:
Regarding Semgrep's analysis, all three are false positives:

### Finding analysis

- Findings 842408061 + 842408062 — TreeFactory L510–511 (both point at the same two lines, in getNextTreeOrder):
```java
  .setSQL("SELECT COALESCE(MAX(tree_order), 0) + 1 AS next_position FROM tree"
          + " WHERE " + whereColumn + " = ? AND relation_type = ?")
```
The rule fires on the string concatenation, but the only concatenated variable is whereColumn, which is a private method parameter whose only two callers pass the hardcoded literals "parent" and "child". Everything user-controllable (id, relationType) goes through addParam bound parameters. No injection path exists.

- Finding 842408064 — VersionableFactoryImpl L476–477:
```java
  .setSQL("SELECT * FROM contentlet_version_info WHERE identifier IN ("
          + DotConnect.createParametersPlaceholder(chunk.size()) + ")");
```
The concatenated value is the output of createParametersPlaceholder(int) — a generated ?,?,?... string derived from a list size, never from content. All identifiers are bound via addParam. This is also the established house pattern: ContentletIndexAPIImpl.SELECT_CONTENTLET_VERSION_INFO and ESContentFactoryImpl build chunked IN clauses with the exact same helper.

jcastro-dotcms:
@claude regarding your feedback:

**1. SQL injection via dynamic column names — ✅ VALID (as hardening) — FIXED**
Agreed there's no current vector (both private methods are only called with the literals `"parent"`/`"child"`, and all user-controlled values are bound parameters), but the contract was conventional rather than enforced. Added a
private `assertTreeColumn(String)` allowlist guard that throws `IllegalArgumentException` for anything other than `parent`/`child`, called at the top of all three SQL-building helpers (`getRelatedIds`, `getNextTreeOrder`,
`deleteTreesScopedByRelationType`).

  **2. Non-atomic delete-then-insert in `insertTrees` — ⚠️ PARTIALLY VALID — FIXED**
For the actual production path the concern doesn't apply: `relateContent` opens an explicit local transaction before `deleteRelatedContent` and commits after `insertTrees`, and `@WrapInTransaction`/`LocalTransaction` *joins* an
existing transaction (see `LocalTransaction.wrapReturn` javadoc), so both batches were already atomic on the same thread-local connection. However, the point is fair for hypothetical standalone callers with no open transaction.
`insertTrees` now wraps both batches in `LocalTransaction.wrap(...)`, making it self-contained for any caller. Verified by `TreeFactoryTest`, which calls it outside any transaction.

  **3. `isExemptFromPermissionFiltering` null-user contract — ❌ NOT VALID**
Returning `false` for a `null` user routes into `deleteReadableRelatedContent`, which is *exactly the legacy code path* — the same `permissionAPI.filterCollection(…, null)` call the pre-refactor code made. `PermissionBitAPIImpl` explicitly handles `null` users (treated as anonymous), so behavior is unchanged. Additionally, a `null` user fails the `PERMISSION_EDIT` check at the top of `deleteRelatedContent` before any filtering is reached. Throwing `IllegalArgumentException` would *change* legacy behavior, which is the opposite of this PR's intent.

  **4. Fast path skips non-relationship-field cache eviction — ❌ NOT VALID**
The audit requested was done when the identifier-only overload was written: the *entire body* of the `Contentlet`-based `invalidateRelatedContentCache` is inside `if (relationship.isRelationshipField())` — for legacy (non-relationship-field) relationships the old method is already a no-op, identical to the new overload's early return. The only behavioral delta is the skipped per-instance `setRelated(null)`, which is intentional and documented in the overload's javadoc: readers always resolve related content through the API, which consults the identifier-keyed `RelationshipCache` that *is* evicted.

  **5. Scroll fallback triggered incorrectly — ❌ NOT VALID (one fair nit)**
The fallback fires only when `totalResults > results.size()`, i.e., only when truncation actually occurred — never when the total is ≤ MAX_LIMIT. When truncation has occurred, a complete re-fetch is mandatory regardless of whether the overage is 1 document or 100k, so there is no "unconditional scroll for edge cases". The fair nit is that `limit > MAX_LIMIT → scroll` is internal branching of `searchIndex`; that dependence is documented with an inline comment at the call site. Exposing the scroll path through a dedicated public API would be a nice follow-up but is out of scope for this PR.

  **6. Wasted `getNextTreeOrder` DB read in the child branch — ✅ VALID — FIXED**
Correct: when the contentlet is the child side, trees are positioned via `positionInParent` and `treePosition` was fetched but never used. The query is now gated to `related.isHasParent()`; the child branch uses a constant. As a result, `getNextTreeOrderByChildAndRelationType` lost its only caller and was removed (introduced in this PR, never released). Added a `TreeFactoryTest` case covering the remaining `getNextTreeOrderByParentAndRelationType` (returns 1 when empty, max+1 otherwise).

  **7. No transaction boundary around delete-then-insert in `relateContent` — ❌ NOT VALID**
The `relateContent` explicitly wraps the whole sequence: `HibernateUtil.startLocalTransactionIfNeeded()` before `deleteRelatedContent`, `commitTransaction()` after `TreeFactory.insertTrees`, and `rollbackTransaction()` in the catch block. `deleteRelatedContent`'s `@WrapInTransaction` joins that transaction rather than creating its own, so a failure in `insertTrees` rolls back the deletes. The pair is atomic both in the standalone-API case and inside the `checkin` flow.

jcastro-dotcms:
This PR was automatically closed by GitHub when its head branch was renamed to match issue #36147. It is superseded by **#36149**, which contains the exact same commits and description. All review feedback from this PR (Semgrep findings and bot review) has been addressed in the code — see #36149.

PR #36098, opened 2026-06-10
fix(multitree): reject save when empty payload would wipe all page content

### LINKED ISSUE #36097: Page content silently wiped when saving with a stale edit session (concurrent editing, TRADITIONAL pages)
## Summary

When multiple users edit the same TRADITIONAL (legacy-template) page simultaneously, a user whose Edit Mode session was opened while the page was empty — or before other users added content — can silently delete all contentlets added by the other users with a single save. The victim users see their content disappear when they next interact with or refresh the page.

## Root Cause

The Edit Mode editor's save model (`$pageData` computed signal in `withEditor.ts`) is built from `store.pageAsset()?.containers` at the time the page is loaded. For TRADITIONAL pages, `isClientReady=true` blocks `pageReload()` from handling subsequent `CLIENT_READY` events, so `$pageData` is never refreshed for the lifetime of that edit session.

When User A opens the page while it is empty and stays in the session while User B adds contentlets, User A's `$pageData` remains empty. If User A saves, `POST /api/v1/page/{id}/content` is called with an empty `containers` payload.

On the backend, `MultiTreeAPIImpl.overridesMultitreesByPersonalization()` does a page-wide **DELETE all rows** then **INSERT only what was posted**. An empty payload deletes every `multi_tree` row for that page with no warning and no recovery path.

## Affected Area

- `MultiTreeAPIImpl.java` — `overridesMultitreesByPersonalization()`: unconditional page-wide delete with no guard against empty payloads
- `withEditor.ts` — `$pageData` computed signal, `isClientReady` guard that prevents `pageAsset` refresh
- `withPageApi.ts` — `editorSave` rxMethod that posts whatever `$pageData` currently holds

## Steps to Reproduce

1. User A opens a TRADITIONAL page in Edit Mode. The page is empty (or has few contentlets).
2. User B (and/or other users) adds a significant number of contentlets to the page and saves. The page now has content.
3. User A, without refreshing their Edit Mode session, performs any action that triggers a save (adds a new contentlet, moves one, publishes, etc.).
4. User A's save posts an empty (or near-empty) `containers` payload.
5. All contentlets added by User B are permanently deleted from `multi_tree`. User B's work is lost.

## Fix

**Short-term (backend guard 
Add a net-loss guard in `overridesMultitreesByPersonalization()`: if the incoming `multiTrees` list is empty and the page already has existing rows, throw `DotDataException` with a message instructing the user to refresh and retry. This is trigger-agnostic and prevents this and any future "empty save wipes the page" regression.

**Long-term (frontend):**
Before executing a save on a TRADITIONAL page, re-fetch `pageAsset` to refresh `$pageData` and reconcile any concurrent changes made since the session was opened.

## Links

- FD-36897: https://dotcms.freshdesk.com/a/tickets/36897

dsilvam:
## Summary

Adds a net-loss guard in \`MultiTreeAPIImpl.overridesMultitreesByPersonalization()\` to prevent a stale Edit Mode session from silently deleting all contentlets on a page.

- Guard is **off by default** — enabled by setting \`MULTITREE_EMPTY_SAVE_GUARD_ENABLED=true\` in dotmarketing-config.properties.
- When enabled: if the incoming \`multiTrees\` payload is empty **and** the page already has existing \`multi_tree\` rows, throw \`DotDataException\` with a message instructing the user to refresh and try again.
- No behaviour change when the page is genuinely empty (0 existing rows) — those saves proceed normally.

## Root Cause (see #36097)

For TRADITIONAL pages, the editor's \`\$pageData\` signal is frozen at session-open time and never refreshed. A user who opened the page while empty will post an empty \`containers\` payload after other users have built up content. \`overridesMultitreesByPersonalization()\` does a page-wide DELETE + reinsert, so an empty post wipes all concurrent work with no warning.

## Test Plan

- [ ] Set \`MULTITREE_EMPTY_SAVE_GUARD_ENABLED=true\`
- [ ] Open a TRADITIONAL page in Edit Mode while it has no content
- [ ] Have a second user add contentlets and save
- [ ] From the first session, trigger a save — confirm an error is returned instead of a silent wipe
- [ ] Verify pages with genuinely empty content (0 existing rows) still save normally
- [ ] Verify guard is inactive when property is not set (default behaviour unchanged)
- [ ] Run \`MultiTreeAPIImplTest\` integration tests

## Links

- Fixes #36097
- FD-36897: https://dotcms.freshdesk.com/a/tickets/36897

This PR fixes: #36097

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/rest/api/v1/page/PageResource.java
+++ b/dotCMS/src/main/java/com/dotcms/rest/api/v1/page/PageResource.java
@@ -854,7 +854,7 @@ public class PageResource {
                                     + "  }\n"
                                     + "]")))
             final PageContainerForm pageContainerForm)
-            throws DotSecurityException, DotDataException {
+            throws DotSecurityException {
 
         final String variantName = UtilMethods.isSet(variantNameParam) ? variantNameParam :
                 VariantAPI.DEFAULT_VARIANT.name();
@@ -879,8 +879,16 @@ public class PageResource {
             this.validateContainerEntries(pageContainerForm.getContainerEntries());
 
             // Save content and Get the saved contentlets
-            final List<ContentView> savedContent = pageResourceHelper.saveContent(
-                    pageId, this.reduce(pageContainerForm.getContainerEntries()), language, variantName, user);
+            final List<ContentView> savedContent;
+            try {
+                savedContent = pageResourceHelper.saveContent(
+                        pageId, this.reduce(pageContainerForm.getContainerEntries()), language, variantName, user);
+            } catch (DotDataException e) {
+                Logger.warn(this, String.format("Page content save rejected for pageId '%s': %s", pageId, e.getMessage()));
+                return ExceptionMapperUtil.createResponse(
+                        new DotDataException("Page content may have been modified by another user — please refresh and try again."),
+                        Response.Status.CONFLICT);
+            }
 
             return Response.ok(new ResponseEntityContentView(savedContent)).build();
         } catch(HTMLPageAssetNotFoundException e) {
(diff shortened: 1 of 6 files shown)


gortiz-dotcms (approved):

dsolistorres (approved):

swicken (approved):
Looks good!

PR #36112, opened 2026-06-10
feat(rest): add /api/v2/assets endpoint for reading and writing file assets

### LINKED ISSUE #35928: feat(rest): add /api/v2/assets endpoint for reading and writing file assets
## Description

Add a clean v2 REST surface for reading and writing dotCMS file assets (`.vtl`, `.dotsass`, `.css`, `.js`, …) in place, addressable by host-qualified path (`//host/application/...`) or identifier.

The current v1 path forces clients through a multipart workflow endpoint plus a separate search to resolve the content-type inode — a search that fails for some extensions (e.g. `.dotsass`) — and raw reads hit content-negotiation and file-size edges. A purpose-built v2 endpoint removes that friction.

`/api/v1/assets` stays untouched: it backs the dotCMS CLI, so its contract is a rollback-unsafe API surface we will not change. v2 reuses the same underlying helper logic — no business logic is duplicated or moved.

## Acceptance Criteria

- [ ] Read a file asset's raw content by host-qualified path or identifier via a single GET
- [ ] Write (create or update) a file asset's content, with separate working-save and publish operations
- [ ] Path/identifier and publish intent expressed in the URL, not request-body flags
- [ ] Works for all file-asset extensions, including ones that fail today (e.g. `.dotsass`)
- [ ] Handles large files without truncation or content-negotiation errors
- [ ] `/api/v1/assets` API contract is unchanged

### Contract guarantees (public, must hold)

- [ ] **Working-save preserves the live version** — saving a draft over a published asset creates/updates the working version and leaves the currently-live version live and serving its bytes; only publish promotes the new version
- [ ] **Identifier reads enforce READ permission** — a user without permission on the asset is rejected (403), never served the bytes; identifier and path addressing select the same version with identical defaults (working, default language)
- [ ] **Zero-byte writes are rejected** — both a *missing* `file` part and a *present, zero-byte* `file` part return 400 with distinct, clear messages. v2 surfaces core's existing 0-length checkin reject (`DotContentletStateException`, `ESContentletAPIImpl#handleBinaries`, default `CONTENT_ALLOW_ZERO_LENGTH_FILES=false`) as a clean 400 — never a 500. Empty-file scaffolding is explicitly **out of scope** (see Decisions locked).
- [ ] **Writes return stored `fileSize`** — the write response includes the persisted byte length (`fileSize` / `metaData.length`) so a caller can assert what actually landed without a second read. Eliminates the silent-success failure mode (write reports success while writing wrong/empty bytes).
- [ ] **Language handling** — omitted/blank language defaults to the site default; a non-blank unknown language is rejected (400)

### Error handling

Every failure returns a correct, meaningful HTTP status (not a generic 500):

- [ ] Unknown host → 404
- [ ] Path or identifier not found / not a file asset → 404
- [ ] Missing asset, or requested language/version unavailable on read → 404
- [ ] Invalid request (missing-or-conflicting path/id, missing file part, **zero-byte file part**, unknown language, path points at a folder) → 400
- [ ] File exceeds max upload size → 400
- [ ] Insufficient permission on target folder → 403

### Testing & docs

- [ ] Java integration tests cover the contract guarantees above (v1 unchanged, working-save preserves live, identifier READ enforcement, default/version selection, filename mismatch, temp cleanup + rollback, multipart validation, **missing-file-part and zero-byte-file-part both → 400**, **`fileSize` returned in the write response**)
- [ ] OpenAPI spec and Postman coverage added, including the error-status cases above

## Decisions locked

Resolved during planning/brainstorm; recorded so implementation doesn't re-litigate or revert them:

- **Empty-file writes are out of scope.** Earlier drafts treated zero-byte files as valid (scaffold-then-fill). Reversed: it is not a real agent workflow (an agent emits complete files; a `fileSize: 0` result is indistinguishable from the silent-failure mode that caused an outage). v2 honors core's existing 0-length reject on **both** `/save` and `/publish`, surfaced as a 400. If a *human* draft-empty workflow is ever needed, the path is a route-scoped `Contentlet.VALIDATE_EMPTY_FILE=false` on `/save` only (never `/publish`) — not built now.
- **v2 checkin always creates a new version.** Verified: `WebAssetHelper` checkin hardcodes `createNewVersion=true`; there is no instance env var that defeats it (the only lever is the programmatic `DONT_VALIDATE_ME`, which v2 does not set). So writes are always rollback-able.
- **The real cause of the 0-byte outage was a *null* binary, not core writing empty.** Core rejects a true 0-length checkin; the prior `template.vtl → 0 bytes` corruption came from a malformed multipart where *no* binary part attached (the 0-length check is skipped when the file is null). The "missing `file` part → 400" validation is what closes that specific hole.
- **Verify the exception mapping before build:** confirm `DotContentletStateException` maps to **400** via an existing JAX-RS `ExceptionMapper`. If none exists, the v2 resource catches it and rethrows `BadRequestException`.

## Additional Context

Motivated by the dotCMS MCP server using these endpoints for AI-assisted editing of file assets — an agent driving read → edit → publish entirely through REST. The MCP tooling itself is a separate effort; this issue covers only the endpoint.

Evidence behind the design: a full a11y fix session on `demo.dotcms.com` run entirely through the MCP surfaced v1's undocumented multipart contract (≈6 trial attempts to find the working shape), a silent-success write that published `template.vtl` as 0 bytes and took the home page down, and read-path friction (POST-to-read, `stInode` Lucene lookup failing for `.dotsass`, 406s, large-file overflow). v2 removes the *cause* (un-mis-buildable flat multipart, intent-in-route, GET reads, `fileSize` in the response), not just the symptom.

_Priority: High_

fmontes:
### Proposed Changes

Closes #35928

Adds a clean v2 REST surface for reading and writing dotCMS file assets (`.vtl`, `.dotsass`, `.css`, `.js`, …) in place, addressable by host-qualified path or identifier. All business logic stays in v1's `WebAssetHelper`; `/api/v1/assets` (which backs the CLI) is contract-unchanged.

**New endpoints** (`com.dotcms.rest.api.v2.asset.WebAssetResourceV2`):

| Endpoint | Purpose |
|---|---|
| `GET /api/v2/assets?path=//host/...` | Raw content read by host-qualified path (`language`, `version=working\|live` query params) |
| `GET /api/v2/assets/{identifier}` | Raw content read by identifier — same defaults as path reads (working, default language), READ permission enforced before bytes are served |
| `PUT /api/v2/assets/save` | Create/update the **working** version only — live keeps serving its bytes |
| `PUT /api/v2/assets/publish` | Create/update **and publish** |

Writes use a flat multipart (`file`, `path`, optional `language`) — no nested JSON `detail` blob, publish intent lives in the URL. Reads stream raw bytes with explicit content type and `Content-Disposition` (no content-negotiation 406s, no large-file truncation).

**Contract guarantees**
- Working-save preserves the live version; only publish promotes (checkin always creates a new version, so writes are rollback-able)
- Write responses include the persisted `fileSize` so callers can assert what actually landed (closes the silent-success failure mode that once published a 0-byte `template.vtl`)
- Zero-byte writes rejected with 400 — distinct messages for a *missing* `file` part vs a *present-but-empty* one
- Omitted/blank language → site default; unknown language → 400
- Clean statuses throughout: unknown host / path / version 404, invalid request 400, oversize upload 400, insufficient permission 403 — no generic 500s, no stacktraces in 404 bodies

### Checklist
- [x] Java integration tests: `WebAssetResourceV2IntegrationTest` (round-trip reads, save-preserves-live/publish-promotes, both zero-byte 400s, unknown language, permission 403, folder-path 400, unknown host 404, `fileSize` assertion, `.dotsass`)
- [x] Postman: new `v2` folder in `WebAssets.postman_collection.json` (8 requests incl. error-status cases)
- [x] OpenAPI annotations + regenerated `openapi.yaml`
- [x] Validated live against a local instance: full curl pass of every contract item, including v1 sanity check (both versions listed, contract untouched)

### Notes for the reviewer
- Only v1 change: `WebAssetHelper.parseLang` widened from package-private to `public` (additive, no behavior change)
- `DotContentletStateException` (core's 0-length checkin guard) already maps to 400 via the existing `DotStateExceptionMapper`; the resource-level zero-byte check makes the message explicit and distinct
- Known cosmetic edge: a path pointing at a folder returns the correct 400, but the `IllegalArgumentException` body includes a stacktrace (same generic mapper used elsewhere) — can be tightened in a follow-up


dotCMS/src/main/java/com/dotcms/rest/api/v2/asset/WebAssetResourceV2.java
```diff
@@ -0,0 +1,718 @@
...
+                .init();
+
+        final User user = initData.getUser();
+        Logger.debug(this, () -> String.format(
+                "User [%s] reading asset by path [%s] lang=[%s] version=[%s]",
+                user.getUserId(), path, language, version));
+
+        final boolean live = "live".equalsIgnoreCase(version);
+        final String resolvedLang = resolveLanguageParam(language);
+        final FileAsset fileAsset;
```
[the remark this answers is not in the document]
fmontes: Fixed in 82153fec3a — `version` is now validated by a shared `resolveVersionParam()` helper: blank/`working` → working, `live` → live, anything else → 400 with "Unknown version [...] — accepted values are 'working' (default) or 'live'." Integration test `testGetByPath_invalidVersion_returns400` covers it.

dotCMS/src/main/java/com/dotcms/rest/api/v2/asset/WebAssetResourceV2.java
```diff
@@ -0,0 +1,718 @@
...
+                .init();
+
+        final User user = initData.getUser();
+        Logger.debug(this, () -> String.format(
+                "User [%s] reading asset by identifier [%s] lang=[%s] version=[%s]",
+                user.getUserId(), identifier, language, version));
+
+        final boolean live = "live".equalsIgnoreCase(version);
+        final long languageId = resolveLanguageId(language);
+
```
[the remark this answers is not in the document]
fmontes: Fixed in 82153fec3a — same shared `resolveVersionParam()` validation applies to the identifier read path: unknown version values now return 400 instead of silently selecting working.

dotCMS/src/main/java/com/dotcms/rest/api/v2/asset/WebAssetResourceV2.java
```diff
@@ -0,0 +1,718 @@
...
+        // Validate required form fields.
+        if (!UtilMethods.isSet(path)) {
+            throw new BadRequestException("Form parameter 'path' is required.");
+        }
+        if (fileInputStream == null) {
+            throw new BadRequestException(
+                    "Missing 'file' part — a multipart field named 'file' with binary content is required.");
+        }
+
```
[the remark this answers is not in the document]
fmontes: Fixed in 82153fec3a — `writeAsset` now rejects a multipart where the `file` stream is present but the `FormDataContentDisposition` is null or carries no filename: 400 "Malformed 'file' part — the multipart field must carry a Content-Disposition header with a filename." Covered by `testSave_missingContentDisposition_returns400`.

dotCMS/src/main/java/com/dotcms/rest/api/v2/asset/WebAssetResourceV2.java
```diff
@@ -0,0 +1,718 @@
...
+            final boolean live,
+            final long languageId,
+            final User user) throws DotDataException, DotSecurityException {
+
+        // Use system user for the lookup — permission check is explicit below.
+        Contentlet contentlet = Try.of(() ->
+                contentletAPI.findContentletByIdentifier(
+                        identifier, live, languageId, APILocator.systemUser(), false))
+                .getOrNull();
+
```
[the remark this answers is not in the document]
fmontes: Fixed in 82153fec3a — dropped the `Try.of(...).getOrNull()` wrapper. `DotDataException`/`DotSecurityException` now propagate to their existing JAX-RS mappers (500/403) so real backend failures are no longer masked as 404; only a null return is treated as not-found.

dotCMS/src/main/java/com/dotcms/rest/api/v2/asset/WebAssetResourceV2.java
```diff
@@ -0,0 +1,718 @@
...
+                while ((read = fis.read(buffer)) != -1) {
+                    outputStream.write(buffer, 0, read);
+                }
+            }
+        };
+
+        return Response.ok(streamingOutput, mimeType)
+                .header("Content-Disposition", "attachment; filename=\"" + file.getName() + "\"")
+                .header("Content-Length", file.length())
+                .build();
```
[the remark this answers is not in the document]
fmontes: Fixed in 82153fec3a — the filename is now encoded with `UtilMethods.encodeURL(...)` before being embedded in the `Content-Disposition` header, matching the existing pattern in `BinaryExporterServlet`.

dotCMS/src/main/java/com/dotcms/rest/api/v2/asset/WebAssetResourceV2.java
```diff
@@ -0,0 +1,734 @@
...
+        if (!permissionAPI.doesUserHavePermission(contentlet, PermissionAPI.PERMISSION_READ, user, false)) {
+            throw new DotSecurityException(
+                    String.format("User [%s] does not have READ permission on asset [%s]",
+                            user.getUserId(), identifier));
+        }
+
+        if (!contentlet.isFileAsset()) {
+            throw new NotFoundInDbException(
+                    String.format("Content [%s] is not a file asset.", identifier));
+        }
```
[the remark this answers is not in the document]
fmontes: Fixed in ee6b787b20 — reordered `getById` so the `isFileAsset()` check runs before the READ-permission check. A non-file identifier now always returns 404 (contract-required) regardless of the caller's permission; READ is still enforced before any bytes are served for genuine file assets.

dotCMS/src/main/java/com/dotcms/rest/api/v2/asset/WebAssetResourceV2.java
```diff
@@ -0,0 +1,734 @@
...
+import javax.ws.rs.core.StreamingOutput;
+import com.dotcms.rest.exception.NotFoundException;
+import java.io.BufferedInputStream;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.Serializable;
+import java.util.Map;
+import java.util.Optional;
+
```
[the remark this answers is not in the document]
fmontes: Fixed in ee6b787b20 — removed the unused `java.io.Serializable` import.

dotcms-integration/src/test/java/com/dotcms/rest/api/v2/asset/WebAssetResourceV2IntegrationTest.java
```diff
@@ -0,0 +1,702 @@
...
+                .replaceAll("/$", "")   // trim trailing slash added by template
+                .replace("/" + fileName + "/", "/" + fileName);
+    }
+
+    /**
+     * Builds the asset path correctly — folder is one level, file is the last segment.
+     */
+    private String assetPath2(final String fileName) {
+        return String.format("//%s/%s/%s", host.getHostname(), folder.getName(), fileName);
+    }
```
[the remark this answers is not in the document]
fmontes: Fixed in ee6b787b20 — removed the dead `assetPath2()` helper; tests use the single `assetPath()` builder.

dotcms-integration/src/test/java/com/dotcms/rest/api/v2/asset/WebAssetResourceV2IntegrationTest.java
```diff
@@ -0,0 +1,695 @@
+package com.dotcms.rest.api.v2.asset;
```
hassandotcms, line 1: `WebAssetResourceV2IntegrationTest`
this file needs to be added to MainSuite, otherwise it will not run in CI.

dotCMS/src/main/java/com/dotcms/rest/api/v2/asset/FileAssetView.java
```diff
@@ -0,0 +1,66 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import io.swagger.v3.oas.annotations.media.Schema;
+
+/**
+ * Lightweight view returned by write operations (save/publish) in the v2 file-asset endpoint.
+ * Contains the persisted asset's key metadata so callers can verify what was stored.
+ */
+@Schema(description = "File asset view returned after a save or publish operation")
+public class FileAssetView {
```
hassandotcms: Small consistency thought: the v1 asset views (AssetView, AssetsRequestForm) use the @Value.Immutable Abstract* pattern, can we make this view to follow the same approach so the v2 package stays aligned with our REST conventions?

dotCMS/src/main/java/com/dotcms/rest/api/v2/asset/WebAssetResourceV2.java
```diff
@@ -0,0 +1,735 @@
...
+        // Verify the contentlet is a file asset first so a non-file identifier always returns 404
+        // (the documented contract), regardless of the caller's permission on it.
+        if (!contentlet.isFileAsset()) {
+            throw new NotFoundInDbException(
+                    String.format("Content [%s] is not a file asset.", identifier));
+        }
+
+        // Enforce READ permission explicitly — never serve bytes to unauthorized callers.
+        if (!permissionAPI.doesUserHavePermission(contentlet, PermissionAPI.PERMISSION_READ, user, false)) {
+            throw new DotSecurityException(
```
hassandotcms, line 306: minor: Just a thought for keeping the resource thin, like: v1's WebAssetResource delegates all of this to WebAssetHelper (no direct API calls in the resource), would it be worth doing the same here for the lookup/type-check/permission logic in getById and the form adaptation in writeAsset, so v2 stays consistent

hassandotcms (changes requested):

dotCMS/src/main/java/com/dotcms/rest/api/v2/asset/WebAssetResourceV2.java
```diff
@@ -0,0 +1,735 @@
...
+        final User user = initData.getUser();
+        Logger.debug(this, () -> String.format(
+                "User [%s] reading asset by path [%s] lang=[%s] version=[%s]",
+                user.getUserId(), path, language, version));
+
+        final boolean live = resolveVersionParam(version);
+        final String resolvedLang = resolveLanguage(language).toString();
+        final FileAsset fileAsset;
+        try {
+            fileAsset = helper.getAsset(buildAssetsRequestForm(path, resolvedLang, live), user);
```
fabrizzio-dotCMS, line 203: **Permission asymmetry between the two GET endpoints (403 vs 404).**

This path-based read delegates to `helper.getAsset(...)`, whose `BrowserQuery.withUser(user)` filters by permission — so an asset the caller can't read is simply absent from the result → `NotFoundInDbException` → **404** (existence hidden).

`getById`, by contrast, looks up with the system user and does an explicit `PERMISSION_READ` check, returning **403** (existence confirmed).

Net effect: the *same asset*, for a user without permission, returns **403 by identifier but 404 by path**. Both stances are defensible on their own, but mixing them in one resource looks accidental rather than designed.

Since the resource requires `requiredBackendUser(true)` (authenticated backend/admin users, not anonymous public), 403 (clarity) is probably the better policy — but then `getByPath` should distinguish it too. Worth picking one policy and applying it to both. Not a blocker; flagging for a deliberate decision.

fabrizzio-dotCMS:
Nice set of improvements here — reviewing this taught me a lot about REST done right and where v1 fell short.

A few things that clicked for me:
- **GET for reads instead of `POST /_download`.** v1 used POST with the path in the body for what is a pure read. Moving to `GET ?path=...` recovers caching, idempotency, and linkability that POST silently throws away. The verb should match the *operation type*, not just be a habit.
- **Query param over path segment for `path`.** Subtle but smart — a host-qualified path put in a `@PathParam` would get mangled by Tomcat/proxy `%2F` normalization. As a query param it survives intact. Right call.
- **`PUT /save` vs `PUT /publish`** expressing intent through the verb + sub-path instead of a flag buried in a JSON `detail` blob. Much clearer and `PUT` correctly signals the idempotent create-or-replace semantics.
- Flat multipart fields and the sharper 400/401/403/404 error semantics are a real step up over v1's "delegate everything and return 200/401".

Left one inline note on the 403-vs-404 permission asymmetry between `getById` and `getByPath` — that's the only design decision I'd want made deliberately rather than inherited.

dotcms-integration/src/test/java/com/dotcms/rest/api/v2/asset/WebAssetResourceV2IntegrationTest.java
fmontes: Good catch — fixed in eef55d467b. Registered `WebAssetResourceV2IntegrationTest` in `MainSuite2b` (next to `WebAssetHelperIntegrationTest`), both the import and the `@SuiteClasses` entry, so it actually runs in CI now. It was compiling but silently never executing. Also documented the gate in `docs/testing/INTEGRATION_TESTS.md` so it's harder to miss next time.

dotCMS/src/main/java/com/dotcms/rest/api/v2/asset/FileAssetView.java
fmontes: Done in eef55d467b. Replaced the hand-written `FileAssetView` with an `AbstractFileAssetView` `@Value.Immutable` interface using the same `@Value.Style(typeAbstract = "Abstract*", typeImmutable = "*")` as the v1 views, so the generated type stays `FileAssetView` and construction moves to `FileAssetView.builder()`.

One gotcha worth flagging: Immutables doesn't copy *type-level* annotations to the generated class by default, so the class-level `@Schema(description=…)` was getting dropped from `openapi.yaml`. Added `passAnnotations = Schema.class` to the style to carry it through — verified the generated spec is byte-identical to before. Captured both points in `docs/backend/REST_API_PATTERNS.md`.

dotCMS/src/main/java/com/dotcms/rest/api/v2/asset/WebAssetResourceV2.java
fmontes: Agree this is the right direction — v2 should stay as thin as v1 and push the lookup/type-check/permission and the form adaptation into `WebAssetHelper`. I'd like to handle it as a separate follow-up rather than in this PR, since doing it faithfully means adding new methods to the shared (CLI-stable) v1 `WebAssetHelper`, and I'd rather isolate that change from the v2 surface work here. I'll open a follow-up issue and link it. Flagging so it's a deliberate defer, not dropped.

dotCMS/src/main/java/com/dotcms/rest/api/v2/asset/AbstractFileAssetView.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * Lightweight view returned by write operations (save/publish) in the v2 file-asset endpoint.
+ * Contains the persisted asset's key metadata so callers can verify what was stored.
+ *
+ * <p>Follows the same {@code @Value.Immutable} {@code Abstract*} convention as the v1 asset views
+ * (e.g. {@link com.dotcms.rest.api.v1.asset.view.AbstractAssetView}) so the v2 package stays
+ * aligned with the REST view conventions.
+ */
+@Value.Style(typeImmutable = "*", typeAbstract = "Abstract*", passAnnotations = Schema.class)
+@Value.Immutable
```
wezell: Should not use Immutables any longer - prefer Java's new `record` for value classes

wezell (changes requested):

hassandotcms (approved):

dotCMS/src/main/java/com/dotcms/rest/api/v2/asset/AbstractFileAssetView.java
fmontes: Done — converted to a Java `record` in 936d0eb. `AbstractFileAssetView` (the `@Value.Immutable` interface) is removed and replaced by a `FileAssetView` record, following the existing record pattern in `com.dotcms.content.index.domain` (e.g. `ContentSearchResponse`). A hand-written nested `Builder` preserves the fluent `FileAssetView.builder()…build()` construction, so the resource call site and the record accessors are unchanged, and the generated `openapi.yaml` is unaffected.

fmontes:
@wezell thanks for the review — addressed. Changes since your review:

- **Immutables → record** (936d0eb): replaced `AbstractFileAssetView` with a `FileAssetView` Java `record`, following the existing record pattern in `com.dotcms.content.index.domain`. Builder + accessors preserved → no call-site churn; `openapi.yaml` unchanged.
- **Folder-path error shape** (9455936784): a GET path ending in `/` (a folder) previously fell through to the generic mapper as `{"message":"Unspecified Asset name.","stacktrace":…}`. It now returns a clean `400 {"error":"dotcms.api.error.bad_request: Path […] does not point at a file asset …"}` matching the rest of the endpoint.
- **403 IT fix** (827f262636): `testGetById_noReadPermission_throws403` was flaky because the shared Chris Publisher user carried an inherited host-level CONTENTLET READ grant. Now uses a freshly-created isolated user with no inherited READ, plus a fixture sanity assertion.

Validated end-to-end against a local instance with curl — 23/23 cases pass (GET by path/id happy + 400/401/403/404 error paths, and PUT save/publish). The `FileAssetView` record serializes correctly on the write paths.

Ready for another look when you have a moment 🙏

wezell (approved):
good

hassandotcms:
PR #36112: QA Findings.

  Result: ~50 requests, ~70 response checks, 0 blockers, 4 non-blocking findings.

  Environment

  - Build: main @ 429540f , includes the PR #36112 merge (/api/v2/assets)
  - Server: http://localhost (Docker-mapped to 8080). Runner: curl. Date: 2026-06-17
  - Auth: admin@dotcms.com. Site: default

  Endpoints exercised

  GET /api/v2/assets?path=//host/..., GET /api/v2/assets/{identifier}, PUT /api/v2/assets/save, PUT /api/v2/assets/publish

  - Happy path: save, read, update, publish, read. Round-trips are byte-identical and fileSize is accurate, including a 200 KB file (no silent truncation).
  - Core contract: a working-save preserves the live version (working returns the new bytes, live returns the old ones), and publish promotes to live. Both confirmed.
  - Auth: 401 unauthenticated, 401 bad credentials, 200 valid, on all four endpoints.
  - Validation: missing path or file returns 400, zero-byte returns 400 with distinct messages for missing vs empty, unknown language returns 400, unknown version returns 400 on
  both read endpoints, and version is case-insensitive.
  - Errors: missing path/id returns 404, unknown host returns 404 on read and write, a folder path on read returns a clean 400, POST to a PUT endpoint returns 405, empty multipart
  returns 400.
  - Binary: dog.avif read by identifier returns image/avif, exact 103543 bytes, with correct Content-Disposition and Content-Length.
  - Idempotency: identifier is stable across saves, inode is versioned. Schema: all 8 FileAssetView keys are present.

  Non-blocking findings

  🟠 Issue 1 (High): save/publish with a folder-style path (trailing /) returns HTTP 500 with a leaked stacktrace (raw PostgreSQL constraint or temp-dir error). The write path doesn't validate that the path has a filename, though the read path does and returns a clean 400. Not a regression; PUT /api/v1/assets/ produces the same 500. Fix: add the filename guard getByPath already uses to writeAsset.

  🟡 Issue 2 (Low): special-character/space filenames on write return 400 with a stacktrace in the body. Status is correct; only the body leaks an IllegalArgumentException. Fix: remap the parse error like getByPath does.

  🟡 Issue 3 (Medium): GET /{identifier} 404s leak full stacktraces (missing id, non-file-asset, missing version). The sibling getByPath returns a clean {"error":...} for the same cases. Not a regression, but an inconsistency within v2. Fix: catch NotFoundInDbException and rethrow NotFoundException.

  🟡 Issue 4 (Low): write-response path is folder-only but the schema says host-qualified. Returns the bare parent folder (e.g. /v2test-.../), no host or filename, while FileAssetView.path documents //demo.dotcms.com/.../banner.vtl. Correct the schema description or do code fix.

fmontes:
Stacktrace is because it rans in dev mode.

PR #36126, opened 2026-06-11
fix(content-types): lazy-load field variables to eliminate cold-cache N+1 hydration

### LINKED ISSUE #36125: Cold-cache N+1: DbFieldTransformer eagerly loads field variables and double-transforms each field
## Problem Statement

On a **cold cache**, hydrating content fires a large N+1 burst of field-variable queries:

```sql
select id, field_id, variable_name, variable_key, variable_value, user_id, last_mod_date
from field_variable where field_id = ?
```

A single cold `POST /api/content/_search` for Blogs with `render:true` on the demo (Costa Rica) starter issues **~298** of these. The render path never actually reads field variables (warm is already `0`), so every one of these cold queries is wasted work.

**Root cause** — `com.dotcms.contenttype.transform.field.DbFieldTransformer` has two defects in the field-loading path used by `FieldFactoryImpl.selectByContentTypeInDb(...)` → `ContentType.fields()`:

1. **Eager hydration:** `fromMap(...)` force-calls `newField.fieldVariables()` on every field as it is transformed from the DB. `Field#fieldVariables()` is a `@Value.Lazy` getter that loads variables via `FieldFactoryImpl.loadVariables(...)` (raw `field_variable` SELECT). Forcing it loads variables for *every* field even when nothing in the consumer ever reads them.
2. **Double transform:** `asList()` runs `fromMap(map)` **twice** per field — the first result is assigned to an unused local `Field f` (dead code) and discarded; the second is added to the list. This doubles the per-field load (the line was clearly meant to be `list.add(f)`).

Combined, these turn one field-set load into `2 × fieldCount` `field_variable` queries. This is why the rejected side-cache spike landed at exactly 149 (≈ 298 / 2): it was masking the double-call, not addressing the eager load.

**Impact:** every cold content hydration across the platform (content search, page render, API) pays a redundant per-field `field_variable` query burst — unnecessary DB load and cold-start latency on a hot, widely-used legacy path. Functionally correct (content renders), purely an efficiency defect.

| Cold Blog `_search` (render:true), demo starter | Released baseline | This fix |
|---|---|---|
| `field_variable` queries | **298** | **0** |
| total JDBC | 632 | 374 |
| warm `field_variable` / JDBC | 0 / 46 | 0 / 46 |

## Steps to Reproduce

Measured with the local OpenTelemetry → Grafana Tempo rig (JDBC spans per request).

1. Start dotCMS on the demo (Costa Rica) starter.
2. **Cold:** restart the dotCMS container to clear in-memory caches (data persists).
3. Fire **one** request:
   ```bash
   curl -s -o /dev/null -X POST http://localhost:8080/api/content/_search \
     -H 'Content-Type: application/json' \
     -d '{"query":"+contentType:Blog +live:true","limit":10,"render":true}'
   ```
4. Count `field_variable` JDBC spans in the resulting trace → **~298** on a cold cache.
5. Repeat the search (warm) → **0** `field_variable` queries (the existing Field/ContentType cache absorbs it).

The cold count scales with `2 × (fields per distinct content type touched)`, confirming the double-transform + eager-load behavior in `DbFieldTransformer`.

## Acceptance Criteria

- [ ] A cold `POST /api/content/_search` with `render:true` on the demo Blogs issues **0** `field_variable` queries (down from ~298); warm remains **0**.
- [ ] Field variables are loaded **lazily** — only when a field's `fieldVariables()` is actually read — and are served from the existing Field/ContentType cache thereafter (no new side-cache introduced).
- [ ] `DbFieldTransformer.asList()` transforms each field **exactly once** (no double `fromMap`).
- [ ] Field-variable reads remain correct: adding, editing, or deleting a field variable is reflected on the next read (no staleness); behavior under transaction rollback and in a cluster is unchanged.
- [ ] An integration test asserts: loading a content type's fields triggers **0** eager `field_variable` loads, lazy access loads exactly once per field, and warm reads are served from cache.
- [ ] No regression in existing field / content-type behavior (`FieldFactoryImplTest` and related suites pass).

## dotCMS Version

Reproducible on the latest `main` and on released `26.04.22-02`. Longstanding — the eager `fieldVariables()` call and the `asList()` double-transform predate 2023 (the 2023 OpenRewrite commit only reformatted the latter).

## Severity

(No functional breakage — content renders correctly — but a platform-wide cold-cache N+1 causing redundant DB load and cold-start latency on a hot legacy path.)

## Links

NA

swicken:
Fixes #36125

### Proposed Changes
* **Lazy field-variable hydration** — `DbFieldTransformer.fromMap()` no longer force-calls `newField.fieldVariables()` on every field. Variables now load lazily on first access to `Field#fieldVariables()` and are memoized on the cached Field/ContentType instance. Consumers that actually need them (content-type save/copy/diff, field-variable REST, import/export & push-publish, indexing, unique-per-site/charLimit validation, AI auto-tag, and templates that explicitly read `fieldVariablesMap()`) still get identical data through the same method — the query just runs when the data is needed instead of for every field on every type load.
* **Remove double-transform** — `DbFieldTransformer.asList()` ran `fromMap(map)` twice per field (the first result was assigned to an unused local and discarded). Combined with the eager load, every field issued **two** `field_variable` queries, one entirely wasted. Now each field is transformed once.

### Why
On a cold cache, a single `POST /api/content/_search` for Blogs with `render:true` on the demo (Costa Rica) starter fired **~298** `select ... from field_variable where field_id = ?` queries. The render path never reads field variables (warm was already `0`), so the eager hydration — doubled by the duplicate `fromMap` — was pure waste. This is a hot, widely-used legacy path: every cold content hydration across the platform paid the burst.

The change is **strictly subtractive**: it only removes queries (and the 2× transform). No new cache, no invalidation logic, no API/DB/ES-mapping changes — existing Field/ContentType cache invalidation is untouched. Rollback-safe.

### Measurements (local OTel → Tempo rig, cold = first search after restart)

| Cold Blog `_search` (render:true) | Released baseline | This PR |
|---|---|---|
| `field_variable` queries | **298** | **0** |
| total JDBC | 632 | 374 |
| warm `field_variable` / JDBC | 0 / 46 | 0 / 46 |

Beats the previously-rejected side-cache spike (which only reached 149 — it was masking the double-transform) and needs none of its cluster/commit/TTL machinery.

### Checklist
- [x] Tests — new `FieldVariableLazyHydrationTest` (asserts 0 eager loads on field load, one lazy load per field on access, warm reads served from cache); existing `FieldFactoryImplTest` regression suite passes (32 tests, 0 failures).
- [ ] Translations — N/A (backend, no user-facing strings)
- [x] Security Implications Contemplated — none; no auth/permission/data-exposure surface changes. `Field#fieldVariables()` remains `@JsonIgnore` and `@CloseDBIfOpened`, so serialization and connection handling are unchanged.

### Additional Info
* **Behavior of the lazy getter is safe:** `@Value.Lazy` uses synchronized double-checked locking; on failure the memo bit isn't set, so there's no poisoned-empty-list state (it retries next access). `@CloseDBIfOpened` means the lazy load works on any thread and closes the connection it opens.
* **One timing nuance:** if a `field_variable` read fails, the `DotStateException` now surfaces at first access (e.g. render/serialization) rather than at content-type load — same exception type, narrower blast radius (only the consumer that needs that field's variables is affected, not every type load).
* Verified end-to-end on the rig: all demo Blogs still render with full body content; warm behavior unchanged.

### Screenshots
N/A — backend change.

This PR fixes: #36125

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/contenttype/transform/field/DbFieldTransformer.java
+++ b/dotCMS/src/main/java/com/dotcms/contenttype/transform/field/DbFieldTransformer.java
@@ -176,10 +176,11 @@ public class DbFieldTransformer implements FieldTransformer {
 			}
 
 		};
-		Field newField = new ImplClassFieldTransformer(field).from();
-		//hydrate field variables
-		newField.fieldVariables();
-		return newField;
+		// Field variables are loaded lazily on first access to Field#fieldVariables() and memoized
+		// on the cached Field/ContentType instance. Do NOT force-load them here: during cold content
+		// hydration this transform runs once per field, so eager loading produced an N+1 burst of
+		// "select ... from field_variable where field_id = ?" for fields whose variables are never read.
+		return new ImplClassFieldTransformer(field).from();
 
 	}
 
@@ -187,7 +188,6 @@ public class DbFieldTransformer implements FieldTransformer {
 	public List<Field> asList() throws DotStateException {
 		List<Field> list = new ArrayList<>();
 		for (Map<String, Object> map : results) {
-		    Field f  = fromMap(map);
 			list.add(fromMap(map));
 		}
 
--- a/dotcms-integration/src/test/java/com/dotcms/contenttype/business/FieldVariableLazyHydrationTest.java
+++ b/dotcms-integration/src/test/java/com/dotcms/contenttype/business/FieldVariableLazyHydrationTest.java
@@ -1,6 +1,7 @@
 package com.dotcms.contenttype.business;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertSame;
 
 import com.dotcms.IntegrationTestBase;
 import com.dotcms.contenttype.model.field.Field;
@@ -126,16 +127,21 @@ public class FieldVariableLazyHydrationTest extends IntegrationTestBase {
             // memoized across reads. It holds for the in-memory (Caffeine) provider used today. If a
             // serialized/remote content-type cache is ever introduced, memoization becomes
             // per-round-trip and this assertion is the intended canary, not a flaky test.
-            ContentType cached = APILocator.getContentTypeAPI(APILocator.systemUser()).find(typeId);
-            cached.fields().forEach(Field::fieldVariables);
+            final ContentType warmed = APILocator.getContentTypeAPI(APILocator.systemUser()).find(typeId);
+            warmed.fields().forEach(Field::fieldVariables);
 
             final AtomicInteger fieldVarQueries = new AtomicInteger();
             final Connection current = DbConnectionFactory.getConnection();
             DbConnectionFactory.setConnection(
                     new CountingConnection(current, "field_variable", fieldVarQueries));
             try {
-                // Same cached instance — reading variables again must hit the cache, not the DB.
-                cached = APILocator.getContentTypeAPI(APILocator.systemUser()).find(typeId);
+                final ContentType cached = APILocator.getContentTypeAPI(APILocator.systemUser()).find(typeId);
+                // The warm guarantee depends on the cache returning the SAME instance, so the
+                // @Value.Lazy fieldVariables() memoization survives. Assert identity explicitly:
+                // if a future cache provider returns copies, this fails loudly instead of the
+                // query-count assertion below silently going green on a per-round-trip reload.
+                assertSame("Content-type cache must return the same instance (lazy memoization depends on it)",
+                        warmed, cached);
                 cached.fields().forEach(Field::fieldVariables);
                 assertEquals("Warm reads must be served from cache (0 field_variable queries)",
                         0, fieldVarQueries.get());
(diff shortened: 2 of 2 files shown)


swicken:
RE: Claude review

Validated all five against the code:

**`@CloseDBIfOpened` (interception skipped) — doesn't hold.** It's not proxy-based; it's ByteBuddy bytecode instrumentation (`ByteBuddyFactory`, `RETRANSFORMATION`, scope `com.dotcms*`) on the annotated base method `Field.fieldVariables()`. The generated `ImmutableTextField.fieldVariables()` calls `super.fieldVariables()` (invokespecial into the instrumented body), so the advice fires on the lazy path. And `loadVariables` reuses the open thread connection anyway, so there's nothing to leak — confirmed by the multi-fork IT run (0 errors).

**`@Value.Lazy` / `innerFieldVariables` dual-guard — real but pre-existing and not triggered.** The only `constructFieldVariables()` caller is `JsonFieldTransformer:143`, which pre-seeds a freshly deserialized field before any getter access (`fieldVariables()` is `@JsonIgnore`), i.e. the safe order. This PR doesn't touch `Field.java` or that path — it removes an eager getter, so exposure goes down, not up. Out of scope here.

**`assertSame` on the cached refs — taking it.** Turns the identity-cache canary from a comment into a real assertion.

`asList`/anonymous-class notes: agreed.

wezell (approved):
I love this type of work and it is good proof that the otel agent can pinpoint real issues in code.

PR #36150, opened 2026-06-12
perf(core): eliminate avoidable per-call allocation in hot logging/config paths

### LINKED ISSUE #36145: Eliminate avoidable per-call allocation in hot logging and config paths
### Description

Allocation profiling of a running dotCMS instance (Pyroscope, TLAB allocation profiles + flame-graph analysis, demo-site under mixed REST/GraphQL/page-render load) attributed several allocation hotspots — `Throwable.fillInStackTrace`, `Method.copy`/`Field.copy`, and `StringLatin1.toLowerCase` leaf frames — to four small, mechanical issues in core utility code. In a 13-minute mixed window, `Config.get*` subtrees alone accounted for **19.3% of all sampled allocation** (and 2.6% of CPU). The system is allocation/GC-bound rather than logic-bound, so reducing avoidable allocation lowers ZGC pressure; no single-endpoint latency change is expected.

The four fixes:

**1. `Logger` lazy-`Supplier` overloads are eager** — `Logger.java:106-118`: all three `Logger.debug(…, Supplier<String>)` overloads call `message.get()` unconditionally, before any level check. ~900 call sites across `dotCMS/src/main/java` use `Logger.debug(x, () -> …)` specifically to avoid message-building cost when debug is off, and all of them evaluate eagerly today. Profiled consequence: `PermissionBitFactoryImpl.loadPermissions` debug suppliers invoke `Contentlet.toString()` → `ToStringBuilder.reflectionToString` → `Class.getDeclaredFields`/`Field.copy` per contentlet on the reindex path, with debug logging disabled.

**2. `Config` int/long/float getters throw an exception on every read of a non-env-overridden key** — `Config.java`: `getIntProperty(name, default)`, `getLongProperty(name, default)`, `getFloatProperty(name, default)` (and the deprecated no-default variants) probe the `DOT_*` env-var alias via commons-configuration's throwing 1-arg getter wrapped in `Try.of`. When the env override is not set — the common case — every call allocates a `NoSuchElementException` with full stack trace and discards it. `getBooleanProperty(name, default)` already has the fix (a `props.containsKey(envKey)` guard); the sibling getters were never updated. Profiled consequences include `ReindexThread`'s paused loop throwing ~4 exceptions/second and `ClickstreamListener.sessionDestroyed` doing a full Config walk per expired session (15% of window allocation under session churn) to check a disabled flag.

**3. `Logger.isVelocityMessage` allocates a lowercased class name on every Class-keyed log call** — `Logger.java:535`: `clazz.getName().toLowerCase()` runs for every `Logger.debug/info(Class|Object, …)` call regardless of log level. The result is a pure function of the `Class` and should be memoized (the logger instances themselves are already memoized the same way).

**4. `LeakyTokenBucketImpl` re-reads Config per request-cost increment** — `LeakyTokenBucketImpl.java:62,67`: `getMaximumBucketSize()`/`getRefillPerSecond()` call `Config.getLongProperty` on every `drainFromBucket`/refill (i.e. per `RequestCostHandler.incrementCost` invocation, which runs per story-block node per rendered contentlet), even though both values are already read into fields in the constructor. Combined with (2), this throws and discards an exception on the request path.

### Acceptance Criteria

- [ ] `Logger.debug(Object, Supplier)`, `Logger.debug(String, Supplier)`, and `Logger.debug(Object, Throwable, Supplier)` do not invoke `Supplier.get()` when debug logging is disabled for the resolved logger, and still log correctly (including the velocity-logger branch) when it is enabled
- [ ] `Config.getIntProperty(name, default)`, `getLongProperty(name, default)`, `getFloatProperty(name, default)` and the deprecated `getIntProperty(name)` / `getFloatProperty(name)` / `getBooleanProperty(name)` no longer throw/catch `NoSuchElementException` internally when the `DOT_*` env alias is absent; returned values are unchanged for all existing cases (env override set, base property set, neither set → default)
- [ ] `Logger.isVelocityMessage(Class)` result is computed at most once per Class (memoized); velocity/viewtool classes still route to the velocity loggers
- [ ] `LeakyTokenBucketImpl` does not call `Config.get*Property` per `drainFromBucket`/`incrementCost` invocation; rate-limit values still honor configuration at construction time
- [ ] Unit tests cover: supplier not evaluated when debug disabled; Config getters return correct values without internal exceptions for the missing-key case; existing Config/Logger/rate-limit test suites pass

### Priority

Medium

### Additional Context

Found while validating the Pyroscope/Tempo observability PoC (related: #598). Flame-graph evidence: each fix maps to a specific allocation stack — `Field.copy` via eager debug suppliers in `PermissionBitFactoryImpl` (reindex path), `fillInStackTrace` via `Config` getters under `ReindexThread`, `LeakyTokenBucketImpl`, and `ClickstreamListener.sessionDestroyed`, and `toLowerCase` via `Logger.isVelocityMessage` on background-thread debug calls.

Out of scope (identified but deferred): caching `TikaUtils` in `MetadataGeneratorImpl` (parser-registry rebuild per binary on reindex), replacing `BeanUtils.copyProperties` in `WorkflowFactoryImpl.convert*`, replacing `Contentlet.toString()` reflectionToString, and cache-key lowercase normalization churn.

swicken:
### Proposed Changes
* **Logger lazy `Supplier` overloads are now actually lazy** — the three `debug(…, Supplier)` and two `info(…, Supplier)` overloads called `message.get()` unconditionally; they now check the level first. ~900 call sites use suppliers specifically to skip message building, including ones that did `reflectionToString` on Contentlets via `PermissionBitFactoryImpl` debug logging on the reindex path.
* **`Config` int/long/float/boolean getters no longer throw per read** — the `DOT_*` env-alias probe used commons-configuration's throwing 1-arg getter wrapped in vavr `Try`, so every read of a non-env-overridden key allocated and discarded a `NoSuchElementException` (with full stack trace). Applied the `containsKey` guard that `getBooleanProperty(name, default)` already had.
* **`Logger.isVelocityMessage` memoized via `ClassValue`** — previously allocated a lowercased class name on every Class-keyed log call; `ClassValue` lookups are allocation-free, keyed per actual Class (no FQN conflation across OSGi classloaders), and entries are released with their class.
* **`LeakyTokenBucketImpl` no longer walks Config per request-cost increment** — the four config reads (`enabled`, `refill`, `max bucket`, `add-header-info`) are TTL-memoized (30s). Runtime overrides via system table / config refresh still take effect within the TTL, preserved deliberately so rate limiting can be toggled during an incident; a test with an injectable TTL pins that contract.

### Why
Allocation profiling of a running instance (Pyroscope TLAB profiles under mixed REST/GraphQL/page-render load) attributed the `Throwable.fillInStackTrace`, `Method.copy`/`Field.copy`, and `StringLatin1.toLowerCase` allocation hotspots to these four spots; `Config.get*` subtrees alone were ~19% of sampled allocation in a mixed window (e.g. one full Config walk — exception included — per expired session via `ClickstreamListener`, per 250ms reindex-pause iteration, per request-cost increment). Expected effect is reduced allocation → less GC pressure, not a single-endpoint latency win.

### Checklist
- [x] Tests
- [ ] Translations
- [x] Security Implications Contemplated (none — no input handling or auth paths touched; rate-limit values remain config-driven)

### Additional Info
New/extended unit tests: `LoggerTest` (suppliers not evaluated when level disabled), `LeakyTokenBucketImplTest` (runtime config override reaches a running bucket; reads within TTL stay cached), `ConfigTest` (`getLongProperty` missing-key default; env-alias-present-but-unparseable falls back to default for int/long/float). The deprecated no-default `Config` getters still throw `NoSuchElementException` for missing base keys — that documented contract is unchanged (and pinned by existing tests).

Closes #36145

This PR fixes: #36145

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/cost/LeakyTokenBucketImpl.java
+++ b/dotCMS/src/main/java/com/dotcms/cost/LeakyTokenBucketImpl.java
@@ -52,19 +52,22 @@ public class LeakyTokenBucketImpl implements LeakyTokenBucket {
     }
 
 
+    // These run on every request-cost increment, so they return the values resolved at
+    // construction time rather than walking Config on each call. Config changes are picked
+    // up when the application-scoped bean is recreated.
     @Override
     public boolean isEnabled() {
-        return Config.getBooleanProperty("RATE_LIMIT_ENABLED", enabled);
+        return enabled;
     }
 
     @Override
     public long getMaximumBucketSize() {
-        return Config.getLongProperty("RATE_LIMIT_MAX_BUCKET_SIZE", maximumBucketSize);
+        return maximumBucketSize;
     }
 
     @Override
     public long getRefillPerSecond() {
-        return Config.getLongProperty("RATE_LIMIT_REFILL_PER_SECOND", refillPerSecond);
+        return refillPerSecond;
     }
 
     @Override
--- a/dotCMS/src/main/java/com/dotcms/cost/LeakyTokenBucketImpl.java
+++ b/dotCMS/src/main/java/com/dotcms/cost/LeakyTokenBucketImpl.java
@@ -3,10 +3,13 @@ package com.dotcms.cost;
 import com.dotmarketing.util.Config;
 import com.dotmarketing.util.Logger;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Suppliers;
 import io.vavr.Tuple;
 import io.vavr.Tuple2;
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Supplier;
 import javax.enterprise.context.ApplicationScoped;
 
 @ApplicationScoped
@@ -17,6 +20,14 @@ public class LeakyTokenBucketImpl implements LeakyTokenBucket {
     final long maximumBucketSize;
     final boolean enabled;
 
+    // These values are read on every request-cost increment, but a Config walk per call is an
+    // allocation hotspot. The TTL-memoized suppliers keep runtime overrides (system table /
+    // config refresh) working with at most CONFIG_REFRESH_SECONDS of staleness.
+    private static final long CONFIG_REFRESH_SECONDS = 30;
+    private final Supplier<Boolean> enabledSupplier;
+    private final Supplier<Long> refillPerSecondSupplier;
+    private final Supplier<Long> maximumBucketSizeSupplier;
+
     private final AtomicLong lastRefill = new AtomicLong(0);
     private final AtomicLong tokenCount = new AtomicLong(0);
     String REQUEST_COST_HEADER_TOKEN_MAX = "x-dotratelimit-toks-max";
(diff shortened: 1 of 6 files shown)


swicken:
> ## 🤖 Codex Review — `openai.gpt-5.5`
> > **[> **[🔴 Critical]** `dotCMS/src/main/java/com/dotcms/cost/LeakyTokenBucketImpl.java:121` — The new debug lambda captures `currentCount`, but `currentCount` is reassigned inside the retry loop, so it is not effectively final and this class will not compile. Copy it to a final local before the lambda or keep the eager string logging.
> 
> Run: [#27548822095](https://github.com/dotCMS/core/actions/runs/27548822095) · tokens: in: 4536 · out: 4535 (reasoning: 4447) · total: 9071

Not even sure how it came up with this :) There is no retry loop, it must be hallucinating.

jcastro-dotcms (approved):

PR #36163, opened 2026-06-15
fix(uve): scan correct site in Page Health a11y/GEO reports on multisite

### LINKED ISSUE #36162: Page Health scanner (a11y / GEO) resolves the wrong site on multisite instances
## Problem Statement

The Page Health a11y and GEO reports in UVE scan the wrong site on multisite instances. Only the page path is sent to the scanner, so when multiple sites share a path (e.g. `/index`), dotCMS resolves the page against the request-derived site instead of the site open in the editor. Worse in headless/UVE, where the request host maps to no dotCMS site.

## Steps to Reproduce

1. Two sites with a page at the same path (e.g. `/index`), with different content.
2. Open Site B's page in UVE.
3. Page Tools → Accessibility (or GEO).
4. **Observed:** report is for the wrong site.
5. **Expected:** report is for Site B's page.

## Acceptance Criteria

- [ ] a11y and GEO reports scan the site currently open in the editor.
- [ ] Works for traditional and headless/UVE pages.
- [ ] Two sites sharing a path each produce their own report.

## dotCMS Version

Latest `main` (introduced in #35212 / #35255).

## Severity

## Links

NA

fmontes:
https://github.com/user-attachments/assets/48b7f092-a64b-4409-b87e-2cfc9dab5397

## What

The Page Health a11y and GEO reports in UVE scanned the wrong site on multisite instances. Only the page path was sent to the scanner, so when multiple sites share a path (e.g. `/index`) dotCMS resolved the page against the request-derived site instead of the site open in the editor — and in headless/UVE the request host maps to no dotCMS site at all.

This appends `host_id=<siteId>` (the page's own `pageAsset.site.identifier`, already computed in `$seoParams()`) to the URL passed to the scanner. dotCMS honors `host_id` to resolve the current site for backend users (`HostWebApiImpl.getCurrentHost`), so the scanner now re-renders the correct page.

## Scope

- `dot-ema-shell.component.ts` — `handleScannerToolClick()` builds the scanner URL with `host_id` via the `URL` API (preserves existing query params, omits when no siteId).
- `dot-ema-shell.component.spec.ts` — updated expectation + added a multisite assertion.

No backend changes needed: the scanner fetches the URL verbatim, so the param survives the round-trip.

Closes #36162

### DIFF
--- a/core-web/libs/portlets/edit-ema/portlet/src/lib/dot-ema-shell/dot-ema-shell.component.spec.ts
+++ b/core-web/libs/portlets/edit-ema/portlet/src/lib/dot-ema-shell/dot-ema-shell.component.spec.ts
@@ -1500,7 +1500,9 @@ describe('DotEmaShellComponent', () => {
 
                 const { currentUrl, siteId } = spectator.component['$seoParams']();
                 const expectedUrl = new URL(currentUrl, window.location.origin);
-                expectedUrl.searchParams.set('host_id', siteId);
+                if (siteId) {
+                    expectedUrl.searchParams.set('host_id', siteId);
+                }
                 expect(openSpy).toHaveBeenCalledWith('a11y', expectedUrl.toString());
             });
 
@@ -1521,12 +1523,35 @@ describe('DotEmaShellComponent', () => {
                 spectator.component['pageScanner'] = {
                     open: openSpy
                 } as unknown as DotPageScannerReportComponent;
+                jest.spyOn(spectator.component, '$seoParams' as never).mockReturnValue({
+                    currentUrl: '/my-page',
+                    requestHostName: 'https://content-site.example.com',
+                    siteId: 'site-b-identifier',
+                    languageId: 1
+                } as never);
 
                 spectator.component.handleScannerToolClick('a11y');
 
-                const { siteId } = spectator.component['$seoParams']();
                 const calledUrl = openSpy.mock.calls[0][1] as string;
-                expect(new URL(calledUrl).searchParams.get('host_id')).toBe(siteId);
+                expect(new URL(calledUrl).searchParams.get('host_id')).toBe('site-b-identifier');
+            });
+
+            it('should omit host_id when the page has no site identifier', () => {
+                const openSpy = jest.fn();
+                spectator.component['pageScanner'] = {
+                    open: openSpy
+                } as unknown as DotPageScannerReportComponent;
+                jest.spyOn(spectator.component, '$seoParams' as never).mockReturnValue({
+                    currentUrl: '/my-page',
+                    requestHostName: 'https://content-site.example.com',
+                    siteId: undefined,
+                    languageId: 1
+                } as never);
+
+                spectator.component.handleScannerToolClick('a11y');
+
+                const calledUrl = openSpy.mock.calls[0][1] as string;
+                expect(new URL(calledUrl).searchParams.has('host_id')).toBe(false);
             });
 
             it('should pass geo type to the page scanner', () => {
--- a/core-web/libs/portlets/edit-ema/portlet/src/lib/dot-ema-shell/dot-ema-shell.component.spec.ts
+++ b/core-web/libs/portlets/edit-ema/portlet/src/lib/dot-ema-shell/dot-ema-shell.component.spec.ts
@@ -1498,8 +1498,23 @@ describe('DotEmaShellComponent', () => {
 
                 spectator.component.handleScannerToolClick('a11y');
 
-                const { requestHostName, currentUrl } = spectator.component['$seoParams']();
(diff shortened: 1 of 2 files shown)


oidacra (approved):

core-web/libs/portlets/edit-ema/portlet/src/lib/dot-ema-shell/dot-ema-shell.component.ts
```diff
@@ -354,13 +354,26 @@ export class DotEmaShellComponent implements OnInit, OnDestroy {
...
      * @memberof DotEmaShellComponent
      */
     handleScannerToolClick(type: PageScannerToolType): void {
-        const { currentUrl, requestHostName } = this.$seoParams();
-        const pageUrl = `${requestHostName}${currentUrl ?? '/'}`;
-        this.pageScanner.open(type, pageUrl);
+        const { currentUrl, siteId } = this.$seoParams();
+        const url = new URL(currentUrl ?? '/', window.location.origin);
+
+        if (siteId) {
```
oidacra, line 372: `DotPageToolUrlParams.siteId` is typed as `string`, but here it's sourced from `pageAsset()?.site?.identifier`, which can be `undefined` before a page loads. The `if (siteId)` guard handles that correctly at runtime, but the type doesn't reflect it. Consider typing the field as `siteId: string | undefined` so the optionality is visible to other consumers and the guard is type-checked rather than defensive.

core-web/libs/portlets/edit-ema/portlet/src/lib/dot-ema-shell/dot-ema-shell.component.spec.ts
```diff
@@ -1498,8 +1498,60 @@ describe('DotEmaShellComponent', () => {
...
-                expect(openSpy).toHaveBeenCalledWith('a11y', `${requestHostName}${currentUrl}`);
+                const { currentUrl, siteId } = spectator.component['$seoParams']();
+                const expectedUrl = new URL(currentUrl, window.location.origin);
+                if (siteId) {
+                    expectedUrl.searchParams.set('host_id', siteId);
+                }
+                expect(openSpy).toHaveBeenCalledWith('a11y', expectedUrl.toString());
+            });
+
+            it('should scan the authoring instance origin, not the page content host', () => {
```
oidacra, line 1509: Unlike the host_id tests below, this one doesn't mock `$seoParams()`, so it asserts against whatever the outer `beforeEach` leaves in the store — if that setup changes, the test silently checks a different URL. It also doesn't assert the URL is *not* the previous `requestHostName`-based form, so it wouldn't catch a regression to the old behavior. Consider mocking `$seoParams()` explicitly here and asserting the origin is the authoring origin rather than the content host.

core-web/libs/portlets/edit-ema/portlet/src/lib/dot-ema-shell/dot-ema-shell.component.spec.ts
```diff
@@ -1498,8 +1498,60 @@ describe('DotEmaShellComponent', () => {
...
+                const openSpy = jest.fn();
+                spectator.component['pageScanner'] = {
```
oidacra, line 1511: `openSpy` and this `pageScanner` assignment are repeated across all four tests in this block. Consider hoisting them into a `beforeEach` for the `handleScannerToolClick` describe block to remove the duplication.

core-web/libs/portlets/edit-ema/portlet/src/lib/dot-ema-shell/dot-ema-shell.component.ts
fmontes: Good catch — `siteId` is genuinely optional. Leaving it as-is here since the type lives in `dot-page-tool.model.ts` (outside this PR's scope) and other consumers depend on it; will tighten to `string | undefined` in a follow-up.

core-web/libs/portlets/edit-ema/portlet/src/lib/dot-ema-shell/dot-ema-shell.component.spec.ts
fmontes: Fair — this one intentionally runs against the real fixture to assert the origin is the authoring instance under default setup. The explicit-mock + regression coverage lives in the two host_id tests right below. Will leave as the fixture-based smoke check.

core-web/libs/portlets/edit-ema/portlet/src/lib/dot-ema-shell/dot-ema-shell.component.spec.ts
fmontes: Agreed, it's repetitive. Keeping it inline for now so each test reads standalone; can hoist to a `beforeEach` in a cleanup pass if the block grows.

PR #36180, opened 2026-06-16
fix(apps): export env-backed app secrets without aborting (#36179)

### LINKED ISSUE #36179: App secrets export fails for System Host env-backed apps
## Problem Statement

Exporting App secrets fails with an `IllegalArgumentException` when any registered App is provisioned **only** through **System Host environment variables** (tier‑3, `DOT_{APP_KEY}_SYSTEM_HOST_{PARAM}`).

`AppsAPIImpl.collectSecretsForExport(...)` aborts the entire export with:

```
Unable to find secret identified by key `<appKey>` under site `<siteId>`
```

This impacts any environment that provisions Apps via System‑Host env vars (a supported provisioning tier) and then attempts an **Export All** of secrets. A single env‑backed App poisons the export for *every* site, so the feature becomes unusable rather than degrading gracefully.

### Root cause

There is an asymmetry between how Apps are **enumerated** for export and how their secrets are **resolved**:

| Step | Path | Behavior |
|------|------|----------|
| Enumerate | `appKeysByHost()` → `hasGlobalEnvBackedSecrets()` | Counts an App as present on **every** valid site if it resolves from **tier‑3 (System Host env)** *or* tier‑4 (legacy env). |
| Resolve | `collectSecretsForExport()` → `getSecrets(appKey, site, user)` | Calls the overload with `fallbackOnSystemHost = false`, which **skips tier‑3 (System Host env) and tier‑5 (System Host stored)**. |

So a System‑Host‑env‑only App is listed for every site, but `getSecrets(...)` returns `Optional.empty()` for those sites because the System Host tier is suppressed → the `else` branch throws and the whole export aborts.

> Legacy global env (tier‑4) does **not** trigger the bug, because tier‑4 is consulted even when `fallbackOnSystemHost = false`. The defect is specific to the **System Host env tier**.

## Steps to Reproduce

1. Register an App descriptor (e.g. `myapp`) with at least one declared param.
2. Provision it **only** via a System Host env var, e.g. `DOT_MYAPP_SYSTEM_HOST_TOKEN=abc` — do **not** save any stored value and do **not** set a host‑specific env var.
3. Have at least one regular (non‑System) site in the instance.
4. Trigger an **Export All** of App secrets (`AppsAPIImpl.exportSecrets` with `exportAll = true`, via the Apps export endpoint).
5. **Observe:** the export fails with `IllegalArgumentException: Unable to find secret identified by key 'myapp' under site '<siteId>'`.

**Expected:** the export completes; env‑backed Apps that have nothing persistable are simply skipped (env‑provided values are not portable and are not persisted anyway).

## Acceptance Criteria

- [ ] Exporting App secrets no longer throws `IllegalArgumentException` when an App is backed only by System Host env vars.
- [ ] `collectSecretsForExport` resolves secrets with the System Host tiers enabled, matching the enumeration done by `appKeysByHost()`.
- [ ] An App key listed for a site but with no exportable (persistable) secret is skipped (logged at debug), not treated as a fatal error.
- [ ] A genuinely empty export (nothing resolvable across all params/sites) still fails fast via the existing empty‑result guard.
- [ ] The explicit (non‑`exportAll`) export path still rejects a non‑existent **site** with a clear error.

## dotCMS Version

Latest from `main` branch.

## Severity

## Links

NA

fabrizzio-dotCMS:
## Proposed Changes

Fixes #36179.

Exporting App secrets (`AppsAPIImpl.exportSecrets` with `exportAll = true`) aborted with `IllegalArgumentException: Unable to find secret identified by key '<appKey>' under site '<siteId>'` whenever an App was provisioned **only** through **System Host environment variables** (tier‑3, `DOT_{APP_KEY}_SYSTEM_HOST_{PARAM}`).

### Root cause

An asymmetry between enumeration and resolution in `collectSecretsForExport`:

| | Path | Behavior |
|---|---|---|
| **Enumerate** | `appKeysByHost()` → `hasGlobalEnvBackedSecrets()` | Counts the app as present on **every** site if it resolves from tier‑3 (System Host env) **or** tier‑4 (legacy env). |
| **Resolve** | `getSecrets(appKey, site, user)` | Uses `fallbackOnSystemHost = false`, which **suppresses tier‑3 and tier‑5** (System Host tiers). |

A System‑Host‑env‑only app is therefore listed for every site, but resolves to `Optional.empty()` for each — and the `else` branch threw, aborting the whole export. (Legacy tier‑4 was unaffected, since it is consulted even with `fallbackOnSystemHost = false`.)

### Fix

In `collectSecretsForExport`:

1. **Resolve with the System Host tiers enabled** — `getSecrets(appKey, true, site, user)` — so resolution matches what `appKeysByHost()` enumerated.
2. **Skip instead of throw** when an app key has no exportable blob: presence in the key listing does not guarantee a persistable secret (env‑backed values are not persisted and are not portable), so they are debug‑logged and skipped.

The existing empty‑result guard still fails a fully empty export, and the unknown‑**site** error is preserved.

## Checklist

- [x] Tests (Unit/Integration)? — see note below
- [x] Translations: NA
- [x] Security implications: handles secrets export; no secret values logged (debug log references key/site only).
- [x] Backwards compatibility: behavior change is strictly more permissive (no longer aborts on env‑backed apps); fully empty exports still error.

## Testing notes

Manual: register an app provisioned only via `DOT_{APP_KEY}_SYSTEM_HOST_{PARAM}`, add a non‑System site, and run Export All — export now completes instead of throwing.

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/security/apps/AppsAPIImpl.java
+++ b/dotCMS/src/main/java/com/dotcms/security/apps/AppsAPIImpl.java
@@ -974,14 +974,24 @@ public class AppsAPIImpl implements AppsAPI {
                 final Set<String> appKeysBySiteId = paramAppKeysBySite.get(siteId);
                 if (isSet(appKeysBySiteId)) {
                     for (final String appKey : appKeysBySiteId) {
-                        final Optional<AppSecrets> optional = getSecrets(appKey, site, user);
+                        // Resolve with the System Host tiers enabled so an app that is only counted
+                        // by `appKeysByHost` through a global env tier (e.g. System Host env) still
+                        // resolves here, instead of coming back empty and aborting the whole export.
+                        final Optional<AppSecrets> optional = getSecrets(appKey, true, site, user);
                         if (optional.isPresent()) {
                             final AppSecrets appSecrets = optional.get();
                             exportedSecrets
                                     .computeIfAbsent(siteId, list -> new LinkedList<>())
                                     .add(appSecrets);
                         } else {
-                            throw new IllegalArgumentException(String.format("Unable to find secret identified by key `%s` under site `%s` ",appKey, site.getIdentifier()));
+                            // Presence in the key listing does not guarantee an exportable blob:
+                            // env-backed apps are provisioned from the environment and are not
+                            // persisted, so they may legitimately resolve to nothing here. Skip
+                            // them rather than failing the export; the empty-result guard below
+                            // still catches the case where nothing at all could be collected.
+                            Logger.debug(AppsAPIImpl.class, () -> String.format(
+                                    "No exportable secret resolved for key `%s` under site `%s` ; skipping.",
+                                    appKey, site.getIdentifier()));
                         }
                     }
                 }
--- a/dotCMS/src/main/java/com/dotcms/security/apps/AppsAPIImpl.java
+++ b/dotCMS/src/main/java/com/dotcms/security/apps/AppsAPIImpl.java
@@ -989,7 +989,7 @@ public class AppsAPIImpl implements AppsAPI {
                             // persisted, so they may legitimately resolve to nothing here. Skip
                             // them rather than failing the export; the empty-result guard below
                             // still catches the case where nothing at all could be collected.
-                            Logger.debug(AppsAPIImpl.class, () -> String.format(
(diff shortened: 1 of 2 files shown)


fabrizzio-dotCMS:
closed in favor of https://github.com/dotCMS/core/pull/36187/changes

PR #36185, opened 2026-06-16
fix-vanity-url-unpublish-canonical-page

riccardoruocco:
## Summary

This PR fixes S3 static Vanity URL cleanup when the canonical page/resource is unpublished.

When `STATIC_PUSH_S3_VANITY_ALIAS_ENABLED` is enabled, unpublishing a canonical resource now looks up any generated Vanity URL aliases in `static_s3_vanity_mapping`, removes the related S3 objects, and clears the mapping rows.

If a vanity alias was shadowing an existing live resource at the same path, dotCMS restores that live resource on S3 instead of deleting the key.

The PR also updates the startup task number and `postgres.sql` definition for `static_s3_vanity_mapping`, so both upgrades and fresh installs create the table correctly.


dsilvam:
Closing in favor of https://github.com/dotCMS/core/pull/36205 so the checks run

PR #36187, opened 2026-06-16
fix(apps): exclude env-var overrides from app secrets export (#36186)

### LINKED ISSUE #36186: App secrets export fails / leaks env values when an env-var override is defined (regression from #35859)
## Problem
PR #35860 (issue #35859) added env-var / Kubernetes-secret overrides for app secrets: `AppsAPIImpl.getSecrets()` now overlays env-var tiers onto the resolved `AppSecrets`.

The export path (`AppsAPIImpl.collectSecretsForExport`) builds the export from `getSecrets()`, so when an env-var override is defined for an app:
- the **export fails / produces a corrupt file** — a locked (tier-1) env secret carries its value in `envVarValue` with a `null` stored `value`, which breaks the export round-trip; and
- env-sourced values **leak into the export file**, even though they belong to the target environment, not the exported config.

## Expected
Exporting app config must **never** include env-var overrides. The export should contain only persisted (stored) secrets. A value that is *masked* by an env override should still export its real **stored** value.

## Fix
`collectSecretsForExport` reads the persisted secrets blob directly (`readStoredSecrets`) instead of `getSecrets()`, so no env value can enter an export. Apps provisioned purely from env (no stored blob — note `appKeysByHost()` lists env-backed apps) are skipped; the existing empty-result guard still errors when nothing remains to export.

## Tests (integration)
- success: stored secret masked by a host env override → export succeeds, round-trip restores the **stored** value, nothing is `fromEnv`.
- failure: env-only app (no stored blob) → export throws `IllegalArgumentException` instead of leaking the env value.

wezell:
## Summary
Fixes #36186.

PR #35860 (#35859) made `AppsAPIImpl.getSecrets()` overlay env-var tiers onto resolved `AppSecrets`. The export path built the file from `getSecrets()`, so a defined env-var override either **corrupted the export** (a locked tier-1 secret has its value in `envVarValue` with a `null` stored `value`) or **leaked the env value** into the export file.

## Fix
`collectSecretsForExport` now reads the persisted secrets blob directly via `readStoredSecrets()` instead of `getSecrets()`:
- env values can never enter an export;
- a stored value *masked* by an env override still exports its real **stored** value;
- apps provisioned purely from env (no stored blob — `appKeysByHost()` lists env-backed apps) are **skipped**, so export-all keeps working in env-provisioned deployments; the existing empty-result guard still errors when nothing remains.

Admin + license are already enforced at the `exportSecrets()` entry point, so bypassing `getSecrets()` does not weaken access control.

## Tests (integration — `AppsAPIImplTest`, in `MainSuite3a`)
- `Test_Export_Excludes_Env_Overridden_Secret` (**success**): stored `p1` masked by a host env override → export succeeds; round-trip restores the **stored** value (not the env value); nothing restored is `fromEnv`.
- `Test_Export_Env_Only_App_Has_Nothing_To_Export` (**failure**): app provisioned only from env → `exportSecrets` throws `IllegalArgumentException` instead of leaking the env value.

## Verification
Core installs; integration module test-compiles.

This PR fixes: #36186

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/security/apps/AppsAPIImpl.java
+++ b/dotCMS/src/main/java/com/dotcms/security/apps/AppsAPIImpl.java
@@ -974,15 +974,27 @@ public class AppsAPIImpl implements AppsAPI {
                 final Set<String> appKeysBySiteId = paramAppKeysBySite.get(siteId);
                 if (isSet(appKeysBySiteId)) {
                     for (final String appKey : appKeysBySiteId) {
-                        final Optional<AppSecrets> optional = getSecrets(appKey, site, user);
-                        if (optional.isPresent()) {
-                            final AppSecrets appSecrets = optional.get();
-                            exportedSecrets
-                                    .computeIfAbsent(siteId, list -> new LinkedList<>())
-                                    .add(appSecrets);
-                        } else {
-                            throw new IllegalArgumentException(String.format("Unable to find secret identified by key `%s` under site `%s` ",appKey, site.getIdentifier()));
+                        // Export persisted secrets only. Read the stored blob directly rather than
+                        // through getSecrets(), which (since #35859) overlays env-var tiers onto the
+                        // result. Env overrides belong to the target environment, not the export, and a
+                        // locked (tier-1) env value masks the stored value with a null, which would
+                        // corrupt the round-trip. Reading the raw blob preserves the real stored value.
+                        final Map<String, Secret> stored = readStoredSecrets(internalKey(appKey, site));
+                        if (stored.isEmpty()) {
+                            // Nothing persisted for this app on this site — it's provisioned purely from
+                            // env vars (appKeysByHost() lists env-backed apps too). There is nothing to
+                            // export; skip it. If every selection is skipped the empty-result guard below
+                            // raises a clear error.
+                            Logger.debug(AppsAPIImpl.class, () -> String.format(
+                                    "No stored secrets for app `%s` on site `%s`; skipping export (env-only).",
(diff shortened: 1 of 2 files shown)


wezell:
Thanks for the review. Evaluated both findings:

**[🟠 High] `readStoredSecrets` NPE — false positive.** `readStoredSecrets` cannot return `null`: it returns `Map.of()` when the blob is absent (`AppsAPIImpl.java:320`), otherwise `AppSecrets.getSecrets()`, whose backing map is constructor-guaranteed non-null (`AppSecrets.java:25` coerces a null `secrets` to `new HashMap<>()`). The existing callers in the same class (`resolveSecrets` at lines 254/272) already use the result without a null check for this reason. No change needed.

**[🟡 Medium] Test mutates global `Config`.** Acknowledged. This is the established pattern for env-override tests in this codebase (e.g. `AppsAPIImplHostEnvProvisioningTest` uses the same `Config.setProperty(envVarName(), …)` + null-reset approach), and the override is reset in a `finally` so it's cleaned up even on exception. Keeping it consistent with the existing convention.

fabrizzio-dotCMS (approved):

wezell:
Re the re-run's [🟡 Medium] on `AppsAPIImpl.java:989`: `readStoredSecrets(String)` is a pre-existing private method in this same class (`AppsAPIImpl.java:317`), unchanged by this PR — it's just not shown in the diff context. Verified: `dotcms-core` installs and `dotcms-integration` test-compiles cleanly, so there's no compilation/runtime concern here. (Same line as the earlier finding; non-null behavior already addressed above.)

Issue #36054, opened 2026-06-08
[OpenSearch Migration] Phase 3: inferIndexToHit throws when VersionedIndices is unpopulated; write vs read/count paths resolve the index from divergent sources


fabrizzio-dotCMS:
## Summary

Surfaced while working on PR #36027 (#36026 VTL aggregation regression). In **Phase 3 (OS-only, `FEATURE_FLAG_OPEN_SEARCH_PHASE=3`)** the OpenSearch read/count path throws when the new `VersionedIndices` (`OPENSEARCH_3X`) rows are not populated, even though content was written successfully — because the **write path and the read/count path resolve the target index from different sources**.

## Evidence

`ContentFactoryIndexOperationsOS.inferIndexToHit` (`dotCMS/src/main/java/com/dotcms/content/index/opensearch/ContentFactoryIndexOperationsOS.java:360-385`):

```java
final Optional<VersionedIndices> optional = Try.of(() ->
    APILocator.getVersionedIndicesAPI().loadDefaultVersionedIndices()).getOrElse(Optional.empty());
if (optional.isEmpty()){
    throw new DotRuntimeException("Unable to load default versioned indices, falling back to default index");
}
```

Two problems:

1. **Misleading message / no fallback.** The message says *"falling back to default index"* but the code **throws** — it never falls back.
2. **Divergent index resolution in Phase 3.** Observed in an integration run (Phase 3):
   - **Write path** (`addContentToIndex` → publish) resolved and wrote to the bootstrap content index `live_<timestamp>` (via the legacy `IndiciesInfo`/`IndiciesAPI`) — succeeded.
   - **Read/count path** (`isInodeIndexed` → `ESContentletAPIImpl.indexCount` → `ContentFactoryIndexOperationsOS.inferIndexToHit`) used `VersionedIndicesAPI.loadDefaultVersionedIndices()` (`OPENSEARCH_3X`), which was **empty** → threw `DotRuntimeException: Unable to load default versioned indices, falling back to default index`.

   Stack:
   ```
   ContentFactoryIndexOperationsOS.inferIndexToHit(ContentFactoryIndexOperationsOS.java:365)
   ContentFactoryIndexOperationsOS.getCountRequest(...:400)
   ContentFactoryIndexOperationsOS.indexCount(...:392)
   ESContentFactoryImpl.indexCount(ESContentFactoryImpl.java:1580)
   ESContentletAPIImpl.isInodeIndexed(ESContentletAPIImpl.java:9829)
[...]

PR #36191, opened 2026-06-16
fix(opensearch): stop Phase 3 reindex from orphaning ES rows in indicies table (#36077)

### LINKED ISSUE #36077: Phase 3 reindex leaves orphan ES entries in the indicies table
## Problem Statement

During OpenSearch migration **Phase 3 (OS is primary)**, running a full **reindex** leaves orphan/garbage entries in the dotCMS `indicies` table (and the corresponding orphaned physical indices). After the OpenSearch reindex completes and the `.os` (v3.X) indices are promoted to `live`/`working`, the cleanup step fails to remove:

- the previous ES `live`/`working` pointers (NULL version), and
- the ES-side `reindex_live` / `reindex_working` pointers (NULL version) created during the same reindex.

The expected end state of a successful reindex is exactly **two** rows (`live` + `working`). Instead the table accumulates stale ES-side rows alongside the promoted OS rows.

This matches the previously-documented **delete-orphan bug class** for the OS migration (bare index name → OS provider resolves without the `.os` tag, so the OS cleanup/swap targets a non-existent name and the real index/row is left behind). See related work on branch `fix/issue-35820-os-index-naming-suffix` / PR #35863.

**Impact:** orphan rows and orphan physical indices accumulate on every Phase-3 reindex, polluting cluster/index state, wasting storage, and potentially breaking later index cleanup/swap operations. No data loss on the OS primary side.

### Observed state (after one Phase-3 reindex)

`select * from indicies`:

| # | index_name | index_type | index_version |
|---|------------|------------|---------------|
| 1 | `cluster_409d7b8c31.working_20260604152231` | working | NULL (ES) |
| 2 | `cluster_409d7b8c31.live_20260604152231` | live | NULL (ES) |
| 3 | `cluster_409d7b8c31.working_20260604192544` | reindex_working | NULL (ES) |
| 4 | `cluster_409d7b8c31.live_20260604192544` | reindex_live | NULL (ES) |
| 5 | `cluster_409d7b8c31.live_20260604192544.os` | live | 3.X (OS) |
| 6 | `cluster_409d7b8c31.working_20260604192544.os` | working | 3.X (OS) |

> Rows 1–4 (NULL version = ES) are the leftover garbage. Only rows 5–6 (the promoted `.os` / v3.X pair) should remain.

_Screenshot evidence (`indicies` table) to be attached to this issue._

## Steps to Reproduce

1. Stand up a dotCMS environment in OpenSearch migration **Phase 3** (OS primary), e.g. the `opensearch-upgrade` env.
2. Confirm the `indicies` table holds the expected `live`/`working` pair.
3. Trigger a full **reindex** (Maintenance → Reindex, or the reindex API).
4. Wait for the reindex to complete and the new `.os` (v3.X) indices to be promoted to `live`/`working`.
5. Query `select * from indicies`.
6. **Observe:** stale ES-side rows (`live`/`working` NULL version and `reindex_live`/`reindex_working` NULL version) remain alongside the promoted OS rows — 6 rows instead of 2.

## Acceptance Criteria

- [ ] After a successful Phase-3 reindex, the `indicies` table contains exactly the active `live` + `working` pair (the promoted `.os` / v3.X rows) — no leftover ES-side `live`/`working` (NULL version) rows.
- [ ] The transient `reindex_live` / `reindex_working` rows are removed (or promoted) once the reindex completes — none remain as orphans.
- [ ] The physical indices backing the removed rows are deleted from the cluster (no orphaned `.os` or bare physical indices left behind).
- [ ] Re-running reindex multiple times in Phase 3 does not accumulate additional orphan rows/indices.
- [ ] Integration coverage exercises a full Phase-3 reindex lifecycle and asserts the final `indicies` table state (2 rows) and absence of orphan physical indices.

## dotCMS Version

main branch (latest) — OpenSearch migration **Phase 3 (OS primary)**, observed in the `opensearch-upgrade` environment.

## Severity

## Links

NA

fabrizzio-dotCMS:
## Proposed Changes

Fixes #36077.

During OpenSearch migration **Phase 3** (OS primary, ES decommissioned), a full reindex left stale ES-side rows in the `indicies` table — the old `live`/`working` pair plus the transient `reindex_live`/`reindex_working` pair (all NULL version) — alongside the correctly-promoted `.os` (v3.X) pair. The table accumulated **6 rows instead of 2** on every Phase-3 reindex.

### Root cause
An asymmetry in `ContentletIndexAPIImpl`:
- `initAndPointReindex` wrote the legacy ES store **unconditionally** (`pointES` had no phase gate), so a Phase-3 reindex still inserted ES `reindex_*` rows and re-persisted the old ES `live`/`working` rows.
- The Phase-3 switchover (`fullReindexSwitchoverOS`) only touches the OS store (`versionedIndicesAPI`) — it never promotes or clears those ES rows. So they orphaned.

### The fix (DB-only — never contacts the ES cluster, which may be down in Phase 3)
1. **Gate `pointES(...)`** in `initAndPointReindex` on `!isMigrationComplete()` so a Phase-3 reindex no longer writes ES pointers (mirrors the existing `isMigrationStarted()` gate on the OS write).
2. **Add `VersionedIndicesAPI.removeLegacyContentIndices()`** (delegating to `IndicesFactory`): deletes the NULL-version `live`/`working`/`reindex_live`/`reindex_working` rows, **preserves** the unmigrated `site_search` pointer and all OS (non-NULL version) rows, and flushes the index caches. Invoked best-effort by `fullReindexSwitchoverOS` after promoting OS (a housekeeping failure must not undo a successful switchover).

**Physical index deletion is intentionally out of scope** — in Phase 3 ES may not be running, so touching the cluster is fragile. Orphaned physical indices are cleaned by the existing scheduled `DeleteInactiveLiveWorkingIndicesJob`.

### Why an explicit cleanup — and why earlier phases never needed one

The legacy ES `reindex_*` rows were never purged by a dedicated call. In the classic ES path (phases 0–2) the cleanup is an **emergent property of `IndiciesFactory.point()`**: `fullReindexSwitchover()` rebuilds the entire `indicies` table from a *fresh* `IndiciesInfo.Builder` that sets only `working`/`live`/`site_search` and leaves `reindex_*` null, so `point()`'s delete-then-conditional-insert drops the reindex rows for free — and overwrites the old `working`/`live` rows in the same pass.

The Phase-3 switchover (`fullReindexSwitchoverOS`) does **not** go through that full-table rebuild — it writes only the OS/versioned rows via `versionedIndicesAPI`, so the implicit cleanup that every earlier phase gets for free simply never runs and the legacy ES rows orphan. `removeLegacyContentIndices()` is the explicit equivalent of that implicit `point()` sweep, scoped to the OS-only path. It is therefore *restoring symmetry of outcome* (a clean 2-row table) across two structurally different switchover paths — not introducing a new, riskier deletion.

**Scope guard:** the delete is intentionally narrow (`index_version IS NULL` + content-index types only), so it can never touch OS (non-NULL version) rows nor the still-unmigrated `site_search` pointer. The phase protection currently lives at the *call site* (`removeLegacyContentIndices()` is only invoked from `fullReindexSwitchoverOS`, which runs only in Phase 3); the method itself has no internal phase guard.

## Acceptance criteria
- [x] After a Phase-3 reindex, no leftover ES `live`/`working` (NULL version) rows.
- [x] Transient `reindex_live`/`reindex_working` rows removed.
- [ ] Physical indices deleted — deferred to `DeleteInactiveLiveWorkingIndicesJob` (see scope note above).
- [x] Re-running reindex in Phase 3 does not accumulate orphan rows.
- [x] Integration coverage asserts the final `indicies` table state.

## Testing
- `ContentletIndexAPIImplMigrationIntegrationTest#test_phase3_removeLegacyContentIndices_purgesEsRowsPreservesSiteSearchAndOs` — seeds the issue's exact 6-row orphan state and asserts the legacy ES content rows are purged while `site_search` and OS rows survive.
- `ContentletIndexAPIImplPhaseTest` — fake updated for the new API method.
- `./mvnw test-compile -pl :dotcms-core,:dotcms-integration` passes on JDK 25.

> Note: the new IT still needs a run against a live OpenSearch Phase-3 environment (`opensearch-upgrade` profile) — a full-reindex E2E was deliberately avoided due to the known `inferIndexToHit` gap (#36054); the test validates the cleanup guarantee deterministically at the DB level.


fabrizzio-dotCMS:
Applied the `optimize()` null/empty guard from the DeepSeek review (🟠 High finding on `IndexAPIImpl.java`).

`IndexAPIImpl.optimize()` now returns early when `indexNames` is `null` or empty, instead of dereferencing it via `stream()`. This removes the NPE risk if a future caller passes `null`, and short-circuits the vendor grouping when there's nothing to optimize.

Commit: 584138996a

Note on the other review findings — after evaluating each against the code, no further changes were warranted:
- 🔴 `newActiveOs` staying empty on OS-mirror failure is **intended** (Phase 1/2 OS writes are best-effort; skipping OS optimize when promotion failed is correct).
- 🟡 The Phase-3 guard for `removeLegacyContentIndices()` correctly lives at the API layer (`VersionedIndicesAPIImpl.isMigrationComplete()`); the factory is an internal DAO.
- 🟡 `WARN` level on the migration-incomplete guard is appropriate — reaching that branch means a misuse outside Phase 3.
- 🟠 The rollback-safety H-1 flag is an accepted property of the Phase 3 cutover (ES is decommissioned by then), not an inline-fixable bug.

dotCMS/src/main/java/com/dotcms/content/index/VersionedIndicesAPIImpl.java
```diff
@@ -228,6 +228,33 @@ public Optional<VersionedIndices> loadDefaultVersionedIndices() throws DotDataEx
...
+        if (!IndexConfigHelper.isMigrationComplete()) {
+            Logger.warn(this, "removeLegacyIndices() called outside Phase 3 "
+                    + "(PHASE_3_OPENSEARCH_ONLY); skipping — the legacy ES live/working rows are "
+                    + "still active in this phase.");
+            return 0;
+        }
+        final int removed = indicesFactory.removeLegacyIndices();
+        // Flush all index-related caches so no stale legacy names survive the deletion:
+        // 1. VersionedIndicesCache — our own versioned-index cache
+        cache.clearCache();
```
jcastro-dotcms, line 250: Is it expected to flush absolutely all cache entries in here?

jcastro-dotcms (approved):

dotCMS/src/main/java/com/dotcms/content/index/VersionedIndicesAPIImpl.java
fabrizzio-dotCMS: Yeah, there's room for improvement there.  I'll work on it later

PR #36202, opened 2026-06-16
perf(velocity): remove avoidable per-render allocations in template engine hot path

### LINKED ISSUE #36201: perf(velocity): remove avoidable per-render allocations in template engine hot path
## Problem

Two spots in the Velocity rendering hot path do avoidable work on **every** template render. Both are pure CPU/allocation overhead with no behavioral effect.

### 1. `VelocityResourceKey` recompiles a regex on every cache lookup

`VelocityResourceKey` is constructed on every Velocity resource cache `get()` — i.e. once per page/template/container/contentlet/field lookup, dozens to hundreds of times per page render. Its constructor calls:

```java
final String[] pathArry = path.split("[/\\.]", 0);
```

`String.split` only fast-paths single literal characters; a **character class** (`[/\.]`) falls through to `Pattern.compile(...)` on **every call**. That's regex compilation on the hottest path in the rendering engine.

### 2. `ASTMethod` allocates `new Object[0]` for every zero-arg method call

`org.apache.velocity.runtime.parser.node.ASTMethod.execute()` does:

```java
Object [] params = new Object[paramCount];
```

For zero-arg calls (`$foo.size()`, `$x.toUpperCase()`, `$date.getTime()` — extremely common), this allocates a throwaway `new Object[0]` every render. The sibling `paramClasses` array already uses the shared `ArrayUtils.EMPTY_CLASS_ARRAY` for the zero-arg case; `params` was simply left inconsistent.

## Fix

1. Hoist the split regex to a precompiled `static final Pattern` (compile once at class load).
2. Reuse `ArrayUtils.EMPTY_OBJECT_ARRAY` for the zero-arg case in `ASTMethod`, mirroring the existing `paramClasses` optimization.

Both changes are behavior-identical.

## Explicitly out of scope (and why)

- **Arg-array pooling** (`new Object[n]`/`new Class[n]` for n>0): unsafe — method calls are re-entrant (`$a.foo($b.bar())`), so a ThreadLocal buffer corrupts. The arrays are tiny, eden-allocated, and die young.
- **`StringWriter` per interpolated string literal**: speculative; the buffer is unavoidable and the synchronized `StringBuffer` is usually lock-elided by the JIT. Touching the fragile interpolation kludge isn't worth an unmeasured gain.
- **MethodHandle caching for invocation**: redundant on Java 25 — core reflection has been MethodHandle-backed since JDK 18 (JEP 416), and the resolved executor is already cached per node/class. No reflection-lookup churn to remove.

## Verification

- Existing `VelocityResourceKeyTest` covers parsing behavior; a variant-with-underscores case is added.
- New `ASTMethodTest` renders zero-arg, single/multi-arg, repeated, and mixed/nested method calls through Velocity to lock in behavior.

wezell:
Fixes #36201

## What

Two behavior-identical optimizations in the Velocity rendering hot path, surfaced while evaluating whether the [reflex "ast.walk 220x" optimizations](https://reflex.dev/blog/why-ast-walk-when-you-can-ast-sprint/) applied to our forked Velocity engine. The tree-walk itself (`SimpleNode.render`) is already optimal; these are the two genuine wins.

### 1. `VelocityResourceKey` — precompile the path-split regex
`VelocityResourceKey` is built on every Velocity resource-cache `get()` (per page/template/container/contentlet/field lookup — dozens to hundreds per page render). Its constructor used `path.split("[/\\.]", 0)`; a character-class regex does **not** hit `String.split`'s single-char fast path, so it recompiled the `Pattern` every call. Hoisted to a `static final Pattern`.

### 2. `ASTMethod` — reuse the shared empty array for zero-arg calls
`ASTMethod.execute()` allocated `new Object[0]` for every zero-arg method call (`$x.toUpperCase()`, `$foo.size()`, …). Now reuses `ArrayUtils.EMPTY_OBJECT_ARRAY`, mirroring the existing `paramClasses` zero-arg optimization one line below.

## Tests
- **New** `ASTMethodTest` — renders zero-arg, single/multi-arg, repeated, and mixed/nested method calls through Velocity (locks in behavior; guards the empty-array re-entrancy concern). Registered in `MainSuite1b`.
- **Extended** `VelocityResourceKeyTest` — adds a variant-name-with-underscores case covering the path split + variant join.

## Deliberately out of scope
- **Arg-array pooling** — unsafe; method calls are re-entrant (`$a.foo($b.bar())`).
- **`StringWriter` per interpolated literal** — speculative; JIT typically lock-elides the synchronized buffer, and the interpolation kludge is fragile.
- **MethodHandle invocation caching** — redundant on Java 25; core reflection is MethodHandle-backed since JDK 18 (JEP 416), and executors are already cached per node/class.

## Risk
Low. No schema, API, or behavioral changes — pure hot-path micro-optimization. Safe to rollback.

This PR fixes: #36201

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/rendering/velocity/services/VelocityResourceKey.java
+++ b/dotCMS/src/main/java/com/dotcms/rendering/velocity/services/VelocityResourceKey.java
@@ -20,12 +20,16 @@ import org.apache.velocity.runtime.resource.ResourceManager;
 import java.io.Serializable;
 import java.io.StringWriter;
 import java.util.Optional;
+import java.util.regex.Pattern;
 
 public class VelocityResourceKey implements Serializable {
 
     private static final char RESOURCE_TEMPLATE = ResourceManager.RESOURCE_TEMPLATE + '0';
     private static final String HOST_INDICATOR = "///";
     private static final long serialVersionUID = 1L;
+    // Compiled once: String.split("[/\\.]") would compile this character-class regex on every
+    // construction, and a VelocityResourceKey is built on every Velocity cache get (per render).
+    private static final Pattern PATH_SPLIT = Pattern.compile("[/.]");
 
     public final String path, language, id1, id2, cacheKey, variant;
     public final VelocityType type;
@@ -76,7 +80,7 @@ public class VelocityResourceKey implements Serializable {
 
         path = cleanKey(filePath);
 
-        final String[] pathArry = path.split("[/\\.]", 0);
+        final String[] pathArry = PATH_SPLIT.split(path, 0);
 
         this.mode = PageMode.get(pathArry[1]);
 
--- a/dotCMS/src/main/java/org/apache/velocity/runtime/parser/node/ASTMethod.java
+++ b/dotCMS/src/main/java/org/apache/velocity/runtime/parser/node/ASTMethod.java
@@ -140,7 +140,7 @@ public class ASTMethod extends SimpleNode
          *  at execution time.  There can be no in-node caching,
          *  but if we are careful, we can do it in the context.
          */
-        Object [] params = new Object[paramCount];
+        Object [] params = paramCount > 0 ? new Object[paramCount] : ArrayUtils.EMPTY_OBJECT_ARRAY;
 
           /*
            * sadly, we do need recalc the values of the args, as this can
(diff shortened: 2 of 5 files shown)


swicken (approved):
well-tested, low risk. SHIP IT

jcastro-dotcms (approved):

PR #36215, opened 2026-06-17
fix(page): exclude archived content from page render and content save (#35993)

### LINKED ISSUE #35993: Archived content still renders in page Edit/Preview mode when used in a container
### Problem Statement

When a contentlet that is placed in a container on a page is **archived**, it continues to
render in the page's **Edit** and **Preview** modes (UVE). It is correctly hidden in **Live**
mode. Archived content should not appear on a page in any mode.

A secondary symptom: while an archived contentlet remains in a container, attempting to
**remove a different (non-archived) contentlet** from the page hangs ("spinning forever").

### Steps to Reproduce

1. Create/identify a contentlet that is added to a container on a page (e.g. Content Type `C01Textbox`).
2. View the page in Edit/Preview mode — confirm the contentlet renders.
3. Archive the contentlet (from the Search/Content portlet, or directly on the page).
4. Reload the page in **Edit** or **Preview** mode.

In edit/preview rendering, `PageRenderUtil` resolves each container's contentlets via the working
version and never filters out archived content:

- `PageRenderUtil.populateContainers()` → `getContentlet()` → `getSpecificContentlet()` calls
  `contentletAPI.findContentletByIdentifier(id, mode.showLive, …)`.
- `PageMode.EDIT_MODE` and `PageMode.PREVIEW_MODE` have `showLive = false`, so the **working**
  version is returned. An archived contentlet retains its working version (archiving sets
  `deleted = true` on the version info but does not drop the working inode), so it is rendered.
- `PageMode.LIVE` has `showLive = true`; there is no live version for archived content, so it is
  correctly excluded — which is why the bug only appears in Edit/Preview.
- There is no `isArchived()` / `deleted` check anywhere in the render path.

### Acceptance Criteria

The archived contentlet should no longer render on the page in any mode (Edit, Preview, or Live),
consistent with Live-mode behavior.

- The archived contentlet still renders in **Edit** and **Preview** mode.
- It is correctly excluded in **Live** mode.
- While the archived contentlet is present, removing another (non-archived) contentlet from the
  page hangs indefinitely.

In edit/preview mode, skip contentlets where `contentlet.isArchived()`
is true after resolution in `populateContainers()` (or filter in `getContentlet`). Edge cases to
weigh: content added from the palette, and the layout-save/MultiTree path behind the "remove spins
forever" symptom.

### dotCMS Version

- Reproduced on **Evergreen (latest)**

### Severity

### Links

https://dotcms.freshdesk.com/a/tickets/37535
https://dotcms.freshdesk.com/a/tickets/37562
https://helpdesk.dotcms.com/a/tickets/37539

dsolistorres:
## Problem

When a contentlet placed in a container on a page is **archived**, it continues to render in the page's **Edit** and **Preview** modes (UVE). It is correctly hidden in **Live** mode. Archived content should not appear on a page in any mode.

A secondary symptom: while an archived contentlet remains in a container, removing a different (non-archived) contentlet from the page hangs ("spinning forever"). The browser POST to `/api/v1/page/{id}/content` returns **400 "Can't find contentlet: \<archived-id\>"**.

## Root cause

Archiving sets `deleted = true` on the version info but keeps the working inode. In Edit/Preview `PageMode.showLive = false`, so `PageRenderUtil.getSpecificContentlet()` resolves `find(workingInode)` and returns the archived working version — and `populateContainers()` only guarded against `null`. Live mode has no live version, so it was already excluded.

Because the archived content kept rendering, it stayed in the editor's page model. On save (a **full replacement**), `PageResource.validateContainerEntries()` looked the id up via `findContentletByIdentifierAnyLanguageAnyVariant()`, which filters out deleted versions, found nothing, and threw `DotContentletStateException` — a `RuntimeException` not caught by the method's `catch (DotDataException)` — surfacing as HTTP 400 and hanging the remove.

## Changes

- **`PageRenderUtil.populateContainers()`** — skip `contentlet.isArchived()` in all modes, consistent with Live-mode behavior. (Also corrected a stray `DotStateException` import.)
- **`PageResource.validateContainerEntries()`** — skip archived content instead of throwing the 400, so a page that still references archived content remains saveable (the content gets removed on save).
- **`PageResourceTest`** — new self-validating integration test `testArchivedContentNotRenderedInEditAndPreviewMode`: renders the page **before** archiving (asserts the content shows, count == 1), archives, then asserts count == 0 in PREVIEW and EDIT modes.

## Testing

- Bug reproduced on unfixed core: the new test **failed** `expected:<0> but was:<1>`.
- With the fix: the test **passes**.
- Full `PageResourceTest` (31 tests) passes against the fixed core with the build cache disabled.

## ⚠️ Reviewer note — intentional behavior change in page-save semantics

After this fix, archived content no longer renders on a page in any mode. Because the page-save endpoint (`POST /api/v1/page/{id}/content`) is a **full replacement** of each container's content, the next time a user saves the page the archived content — no longer in the editor's model — is **removed from the page's `multi_tree` association**.

**Implication:** if that content is later **unarchived**, it will **not** automatically reappear on the page; it must be re-added manually.

We consider this acceptable (archived content shouldn't be bound to live pages), but flagging it explicitly. An alternative — preserving the `multi_tree` association so unarchive restores placement — was considered and rejected as more complex and less intuitive. **Please confirm you're comfortable with the chosen semantics.**

Refs: #35993

This PR fixes: #35993

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/rendering/velocity/services/PageRenderUtil.java
+++ b/dotCMS/src/main/java/com/dotcms/rendering/velocity/services/PageRenderUtil.java
@@ -299,13 +299,15 @@ public class PageRenderUtil implements Serializable {
                     // Archived (deleted) content keeps its working version, so a showLive=false
                     // lookup (EDIT/PREVIEW modes) still resolves it. Skip it in every mode so that
                     // archived content never renders on the page, consistent with LIVE-mode behavior.
+                    // A DotSecurityException here is a genuine access-control failure and must NOT be
+                    // swallowed as "probably archived" -- let it propagate to the caller.
                     try {
                         if (nonHydratedContentlet.isArchived()) {
                             Logger.debug(this, () -> "Skipping archived contentlet: "
                                     + nonHydratedContentlet.getIdentifier());
                             continue;
                         }
-                    } catch (final DotStateException | DotDataException | DotSecurityException e) {
+                    } catch (final DotStateException | DotDataException e) {
                         Logger.warn(this, "Could not determine archived state for contentlet '"
                                 + nonHydratedContentlet.getIdentifier() + "'; skipping it", e);
                         continue;
--- a/dotCMS/src/main/java/com/dotcms/rendering/velocity/services/PageRenderUtil.java
+++ b/dotCMS/src/main/java/com/dotcms/rendering/velocity/services/PageRenderUtil.java
@@ -299,8 +299,11 @@ public class PageRenderUtil implements Serializable {
                     // Archived (deleted) content keeps its working version, so a showLive=false
                     // lookup (EDIT/PREVIEW modes) still resolves it. Skip it in every mode so that
                     // archived content never renders on the page, consistent with LIVE-mode behavior.
-                    // A DotSecurityException here is a genuine access-control failure and must NOT be
-                    // swallowed as "probably archived" -- let it propagate to the caller.
+                    // isArchived() declares DotSecurityException, but VersionableAPI.isDeleted() does
+                    // not throw it via this path (it is declared for forward-compatibility). Should a
+                    // DotSecurityException ever surface, it is a genuine access-control failure and
+                    // must NOT be swallowed as "probably archived" -- so it is intentionally left
+                    // uncaught and propagates to the caller.
                     try {
                         if (nonHydratedContentlet.isArchived()) {
                             Logger.debug(this, () -> "Skipping archived contentlet: "
--- a/dotCMS/src/main/java/com/dotcms/rendering/velocity/services/PageRenderUtil.java
+++ b/dotCMS/src/main/java/com/dotcms/rendering/velocity/services/PageRenderUtil.java
@@ -296,6 +296,21 @@ public class PageRenderUtil implements Serializable {
                         continue;
                     }
 
+                    // Archived (deleted) content keeps its working version, so a showLive=false
+                    // lookup (EDIT/PREVIEW modes) still resolves it. Skip it in every mode so that
+                    // archived content never renders on the page, consistent with LIVE-mode behavior.
+                    try {
+                        if (nonHydratedContentlet.isArchived()) {
+                            Logger.debug(this, () -> "Skipping archived contentlet: "
+                                    + nonHydratedContentlet.getIdentifier());
+                            continue;
+                        }
+                    } catch (final DotStateException | DotDataException | DotSecurityException e) {
+                        Logger.warn(this, "Could not determine archived state for contentlet '"
+                                + nonHydratedContentlet.getIdentifier() + "'; skipping it", e);
+                        continue;
+                    }
+
                     final DotContentletTransformer transformer = new DotTransformerBuilder()
                             .defaultOptions().content(nonHydratedContentlet).build();
                     final Contentlet contentlet = transformer.hydrate().get(0);
--- a/dotCMS/src/main/java/com/dotcms/rest/api/v1/page/PageResource.java
+++ b/dotCMS/src/main/java/com/dotcms/rest/api/v1/page/PageResource.java
@@ -87,6 +87,7 @@ import com.dotmarketing.exception.DotSecurityException;
 import com.dotmarketing.exception.StalePageSaveException;
 import com.dotmarketing.portlets.containers.model.Container;
 import com.dotmarketing.portlets.contentlet.business.ContentletAPI;
+import com.dotmarketing.portlets.contentlet.business.DotContentletStateException;
 import com.dotmarketing.portlets.contentlet.model.Contentlet;
 import com.dotmarketing.portlets.contentlet.transform.DotTransformerBuilder;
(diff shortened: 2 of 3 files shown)


dsolistorres:
### Review findings addressed

All findings from the automated backend-review passes have been resolved across the latest commits. Summary:

**🔴 Critical**
- `validateContainerEntries` no longer wraps `DotSecurityException` as a 400 — it propagates to `DotSecurityExceptionMapper` (403 + security audit log).

**🟠 High**
- Exception messages are no longer forwarded to the client; full detail is logged server-side and a generic message is returned (no PII / SQL / identifiers).
- `PageRenderUtil` archived-state check no longer swallows `DotSecurityException` — only `DotStateException`/`DotDataException` warn-and-skip.
- Single contentlet lookup (was up to 3 DB calls); the `DotContentletStateException` cause is inspected so a real DB error is surfaced rather than silently treated as "archived/not found".
- Removed the misleading `throws DotDataException` from `validateContainerEntries` (it created a raw-message leak path via the JAX-RS mapper); the checked exception is handled internally.

**🟡 Medium**
- Added `isSet` guards for `containerId` and `contentletId`.
- Archived/not-found are treated identically (silent skip) to avoid leaking content existence.
- Benign "archived or not found" case logs message-only (no stack trace) on every save; the exception is kept only on the error path.
- `getContainerContentTypes` returns a generic message and logs detail server-side — including the unchecked `DoesNotExistException` path that previously leaked `containerId` via the 404 mapper.
- Test uses `APILocator.getContentletAPI()` (not the `Impl`) and reuses the `systemUser` local.

**Verification:** each round, `dotcms-core` builds and the full `PageResourceTest` suite passes 31/31 (build cache disabled). The regression test `testArchivedContentNotRenderedInEditAndPreviewMode` was confirmed to fail against the unfixed core and pass with the fix.

> **Reviewer note (unchanged decision):** archived content is de-associated from a page on the next save because the page-content endpoint is a full replacement. If content is later unarchived it will not automatically reappear on the page. This is intentional — see the PR description.

dsolistorres:
Thanks — addressing both observations:

**🟡 `StructureDataGen` → `ContentTypeDataGen` (test):** Applied in `f24972fc99`. The new test now builds the container via `ContainerDataGen.withContentType(new ContentTypeDataGen()...)` instead of the legacy `Structure`. Behavior is unchanged (rendering is driven by the MultiTree); the test still passes.

**🟡 `containerId` in `DoesNotExistException` message — intentionally NOT changed.** Heads-up for the next review pass so this isn't re-flagged:

- I reverted catching `DoesNotExistException` in `getContainerContentTypes` (commit `412c5c1975`). A missing container must surface as **HTTP 404** with the message `"Container with ID :<id> not found"` — this is an established API contract asserted by the `Define_Contentlets_StyleProperties` postman collection ("Unexistd container error" test). Catching it (or removing the id from the message) turns the 404 into a 400 / changes the body and **breaks that test**.
- The id in that message is the **caller's own request input** (e.g. `fake-123abc-1`), not sensitive data or an internal identifier — echoing it back in a not-found error is standard REST behavior, not an information leak.

The genuine leak vectors in this method — `DotDataException` / `DotSecurityException` (which can carry SQL fragments or other users' PII) — are still sanitized: full detail is logged server-side and a generic message is returned to the client. Only the caller-supplied container id in the 404 is intentionally preserved.

dsolistorres:
### Review status — addressed vs. intentionally rejected

Pushed `3da7152550`. Summary of the latest review items:

**✅ Addressed**
- 🟠 **`StructureDataGen` in the new test** → switched to `ContentTypeDataGen` + `ContainerDataGen.withContentType(...)` (`f24972fc99`).
- 🟠 **`NotFoundInDbException` matched only on the direct cause** → now uses `ExceptionUtil.causedBy(e, NotFoundInDbException.class)`, which walks the full cause chain, so an added intermediate wrapper can't silently turn the archived/not-found path into a client error.
- 🟡 **`DotSecurityException` collapsed into a 400 in `getContainerContentTypes`** → split the catch: `DotSecurityException` is re-thrown as `ForbiddenException` (→ 403 via the mapper, generic message), `DotDataException` stays a generic 400.
- 🟡 **Content-type validation message too generic** → now includes the content-type variable (`"Content type '<var>' is not allowed in this container"`); the variable is public schema metadata, not sensitive.

**🚫 Intentionally not changed (reiterating, with rationale)**
- 🟡 **`containerId` embedded in `DoesNotExistException` message** — a missing container must surface as **HTTP 404** with `"Container with ID :<id> not found"`. This is a tested API contract (`Define_Contentlets_StyleProperties` → "Unexistd container error"); changing it (or removing the id) **breaks that postman test**. The id is the caller's own request input, not sensitive data. The genuine leak vectors (`DotDataException` / `DotSecurityException`) are sanitized.

**Verification:** `dotcms-core` builds, full `PageResourceTest` passes 31/31 (cache disabled), and the postman regression that broke earlier is restored by keeping the 404 contract.

dsilvam (approved):

PR #36234, opened 2026-06-18
feat(opensearch-migration): verbose, easy-to-locate startup logging with masked secrets

### LINKED ISSUE #36233: Improve ES→OpenSearch migration startup logging (verbose, easy-to-locate, masked secrets)
## Description

During the ES → OpenSearch migration, testers cannot reliably follow what the system did at startup with the OpenSearch connection parameters. Two concrete gaps cause confusion:

1. **Silent anonymous connect.** When only an OS endpoint URL is provided (no user/pass), credentials are never applied (they are only set when *both* username and password are present). If the OpenSearch cluster has security disabled, dotCMS connects anonymously. A tester expecting an auth error instead saw a silent success, because the log said `OpenSearch client initialized successfully` without stating that no credentials were resolved.
2. **"Initialized" ≠ "connected".** Building the `OpenSearchClient` does not open a connection — reachability and the OS version are only verified later in `IndexStartupValidator`. The old log made "configured" look like "connected".

The migration startup logging should be verbose and easy to locate so testers can follow the flow, see which parameters were actually resolved, and immediately tell whether OS was reached or the system fell back to ES.

## Acceptance Criteria

- [ ] A single, easy-to-grep log block is emitted when the OpenSearch client is built, containing: migration phase, OS endpoints, effective authentication mode (`BASIC` / `JWT` / `CERT` / `NONE`), and TLS flags (enabled / cert required / trust self-signed / CA cert).
- [ ] When no credentials are resolved, the auth mode is reported explicitly as `NONE — connecting ANONYMOUSLY (no username/password/token resolved)`.
- [ ] Sensitive values (basic-auth password, JWT token) are masked using `****` plus the final character (e.g. `s3cret` → `****t`); a single-character secret is fully masked; an unset value shows `(not set)`. The mask does not reveal the secret's real length.
- [ ] The configuration block clarifies that connectivity and OS version are verified separately at startup (building the client does not connect).
- [ ] A result banner at startup states clearly whether OS connected successfully (with the active phase) or validation failed and dotCMS falls back to ES-only (`PHASE_0_MIGRATION_NOT_STARTED`).
- [ ] No changes to existing Elasticsearch classes (migration design rule).

## Priority

Medium

## Additional Context

Affected files:
- `dotCMS/src/main/java/com/dotcms/content/index/opensearch/ConfigurableOpenSearchProvider.java` — configuration summary banner + secret masking.
- `dotCMS/src/main/java/com/dotcms/content/index/IndexStartupValidator.java` — success / fallback result banner.

Reported by QA while exercising the `os-migration` Docker stack (OpenSearch with the security plugin disabled).

fabrizzio-dotCMS:
## What & why

Make the ES→OpenSearch migration startup logging verbose and easy to locate so QA can follow the flow and immediately tell whether OS connected or the system fell back to ES.

### The problem
Testers couldn't follow the connection flow:
- A URL-only config (no user/pass) connected **silently** — credentials are only applied when *both* username and password are present, so against a security-disabled OS cluster dotCMS connected **anonymously** while the log just said `initialized successfully`.
- Building the `OpenSearchClient` doesn't open a connection; reachability/version are only checked later in `IndexStartupValidator`. "Configured" looked like "connected".

### 1. One coherent startup banner, one logger
The configuration summary **and** the connection outcome are now logged together by `IndexStartupValidator`, as a single startup story:

```
========== OpenSearch Migration — client configuration ==========
  Migration phase   : PHASE_1_DUAL_WRITE_ES_READS
  OS endpoints      : [http://localhost:9201]
  Authentication    : NONE — connecting ANONYMOUSLY (no username/password/token resolved)
  TLS enabled       : false
  TLS cert required : false
  TLS trust selfsign: false
  TLS CA cert       : (not set)
  (connectivity + OS version are verified by the checks that follow)
=================================================================
OpenSearch startup validation PASSED — connected to OS successfully; migration phase PHASE_1_DUAL_WRITE_ES_READS is active.
```

- Effective auth mode: `BASIC` / `JWT` / `CERT` / `NONE` — with an explicit **anonymous** warning when no credentials resolve (the exact gap that confused QA).
- On failure: `OpenSearch startup validation FAILED — halting OS migration; dotCMS falls back to ES-only (PHASE_0_MIGRATION_NOT_STARTED): <reason>`.

### 2. Secret masking moved to `StringUtils`
`StringUtils.maskSecret(String)` — reusable, general-purpose: `****` + last char (`s3cret` → `****t`); a 1-char secret is fully masked; unset → `(not set)`; the mask never reveals the real length.

### 3. `IndexStartupValidator` moved into `com.dotcms.content.index.opensearch`
So the validator can re-resolve the exact same config the client is built from via a **package-private** `ConfigurableOpenSearchProvider.configFromProperties()` — **without** exposing any config accessor outside the package or widening the `OSClientProvider` CDI contract. The provider no longer logs the INFO banner; it keeps a quiet `DEBUG` line so runtime (re)builds stay observable without duplicating the banner.

### 4. Endpoint resolution unified + honest docs
- The endpoint-separation check now consumes the already-resolved `OSClientConfig.endpoints()` instead of its own `resolveOSEndpoints()` helper (removed) — one fewer independent config-read path. Behaviour preserved (same fallback chain, same `host:port` normalisation).
- Javadoc corrected: the banner is a **deterministic re-resolution from the same properties**, not a readback of the live client's stored config (documented the one case they could differ: runtime property change without a client rebuild).

## Design notes / conscious trade-offs
- **Re-resolution, not live readback**: the banner re-resolves config rather than reading the instance the live client holds. Deterministic at startup; documented. If a diagnostics endpoint ever needs the *live* client config, the provider should then store and expose the resolved `OSClientConfig`.
- **No ES classes changed** (migration design rule).

## Files
| File | Change |
|---|---|
| `IndexStartupValidator.java` | Moved to `…/opensearch`; owns the config + outcome banner; consumes resolved config for endpoint separation |
| `ConfigurableOpenSearchProvider.java` | `configFromProperties()` package-private static; DEBUG-only build line; banner removed |
| `StringUtils.java` | New reusable `maskSecret(String)` |
| `ContentletIndexAPIImpl.java` | Import update for the moved validator |

## Verification
`./mvnw compile -pl :dotcms-core -DskipTests` → **BUILD SUCCESS**.

Fixes #36233

This PR fixes: #36233

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImpl.java
+++ b/dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImpl.java
@@ -24,7 +24,7 @@ import com.dotcms.content.elasticsearch.util.MappingHelper;
 import com.dotcms.content.index.ContentletIndexOperations;
 import com.dotcms.content.index.IndexAPI;
 import com.dotcms.content.index.IndexAPIImpl;
-import com.dotcms.content.index.IndexStartupValidator;
+import com.dotcms.content.index.opensearch.IndexStartupValidator;
 import com.dotcms.content.index.IndexTag;
 import com.dotcms.content.index.PhaseRouter;
 import com.dotcms.content.index.VersionedIndices;
--- a/dotCMS/src/main/java/com/dotcms/content/index/IndexStartupValidator.java
+++ b/dotCMS/src/main/java/com/dotcms/content/index/IndexStartupValidator.java
@@ -59,11 +59,14 @@ public class IndexStartupValidator {
     public static boolean validateIndexingConfig() {
         try {
             new IndexStartupValidator(CDIUtils.getBeanThrows(OSClientProvider.class)).validate();
+            Logger.info(IndexStartupValidator.class,
+                    "OpenSearch startup validation PASSED — connected to OS successfully; migration phase "
+                    + IndexConfigHelper.MigrationPhase.current().name() + " is active.");
             return true;
         } catch (DotRuntimeException e) {
             Logger.error(IndexStartupValidator.class,
-                    "OpenSearch configuration error — halting OS migration, dotCMS will fall back to ES-only: "
-                    + e.getMessage(), e);
+                    "OpenSearch startup validation FAILED — halting OS migration; dotCMS falls back to"
+                    + " ES-only (PHASE_0_MIGRATION_NOT_STARTED): " + e.getMessage(), e);
             return false;
         }
     }
--- a/dotCMS/src/main/java/com/dotcms/content/index/opensearch/ConfigurableOpenSearchProvider.java
+++ b/dotCMS/src/main/java/com/dotcms/content/index/opensearch/ConfigurableOpenSearchProvider.java
@@ -94,9 +94,13 @@ class ConfigurableOpenSearchProvider {
 
     /**
      * Resolves an {@link OSClientConfig} from dotCMS properties. Package-private and {@code static}
-     * so {@link IndexStartupValidator} (same package) can re-resolve the exact configuration the
-     * client is built from — for the startup banner — without exposing any configuration accessor
-     * outside this package or widening the {@link OSClientProvider} contract.
+     * so {@link IndexStartupValidator} (same package) can re-resolve the configuration from the same
+     * properties — for the startup banner and the endpoint-separation check — without exposing any
+     * configuration accessor outside this package or widening the {@link OSClientProvider} contract.
+     *
+     * <p>Resolution is deterministic from properties: the validator's re-resolution matches what the
+     * client is built with at startup. It is a re-resolution, not a readback of the live client's
+     * stored config; the two could only differ if properties changed without a client rebuild.</p>
      */
     static OSClientConfig configFromProperties() {
         Builder builder = OSClientConfig.builder();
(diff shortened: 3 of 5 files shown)


fabrizzio-dotCMS:
### Review feedback addressed (Bedrock `deepseek.v3.2`)

**Applied — commit `8bec844`:**
- **[🟠 High] `IndexStartupValidator.java:59`** — `validateIndexingConfig()` caught only `DotRuntimeException`. Broadened to `catch (Exception e)` so any failure escaping `validate()` funnels through the single FAILED-banner + `haltMigration()`/Phase-3-abort path, instead of surfacing as a generic FATAL that skips the ES-only fallback.

**Reviewed, not actioned (rationale):**
| Finding | Why not changed |
|---|---|
| [🟠 High] `:106` missing null check on `jwtToken().get()` | False positive — the `.get()` is inside `if (config.jwtToken().isPresent())`. |
| [🟡] `:106` token exposure if `maskSecret` bypassed | Value always goes through `maskSecret`; comment self-resolves ("it does"). |
| [🟡] `:109-110` cert/key **paths** logged | Paths (not key material) are non-secret and useful for debugging; masking them defeats the banner's purpose. |
| [🟡] `:113` `password().orElse(null)` null handling | Already robust via `UtilMethods.isSet` (comment self-resolves). |
| [🟡] `:80` banner not logged if `resolveConfig()` fails | By design; with the High fix above the FAILED banner is now always logged. |
| [🟡] `StringUtils.maskSecret` 1-char → `****` reveals length=1 | Suggested fix (always `****`) removes the last-char sanity-check, which is the feature. Full-masking a 1-char secret is the intended security tradeoff. |

The `claude[bot]` rollback analysis returned **Safe To Rollback** — no action needed.

nollymar (approved):

Issue #36067, opened 2026-06-08
[IMAGE EDITOR] Wire image editor Save, Download, and Cancel to FileFieldStore temp-file flow


nicobytes:
## Description

Complete the save/cancel/download flow from the Angular image editor back to the unified file field:

- **Save** → `POST /api/v1/image/edit` → `DotCMSTempFile` → launcher `onClose` → `FileFieldStore.setPreviewFile`
- **Download** → preview URL + `force_download=true`
- **Cancel** → close modal, no temp file

Field availability (Image always, Binary only when `metadata.isImage`, File never) is covered by #36058.

## Acceptance Criteria

- [ ] Save calls REST endpoint with current filter chain + focal point; closes modal with temp file.
- [ ] Field value updated to temp file id; preview refreshes in file-field component.
- [ ] Download opens filtered URL with `force_download=true` in new tab.
- [ ] Cancel closes modal and emits `null` — field value unchanged.
- [ ] Enterprise license gate respected (match Dojo behavior).

## Priority

High

## Additional Context

- Epic: #35846
- Depends on: #36061 (REST), #36062 (launcher), #36058 (edit action + availability)
- Never overwrite source binary — temp-only until workflow check-in

oidacra:
**Progress note (from PR #36236 / #36063)**

**Download and Cancel are already implemented** in the Angular editor:
- Download triggers the current preview/filter URL from the editor footer.
- Cancel closes the modal returning `null` (field value unchanged), behind an unsaved-changes guard.

So those two acceptance criteria are effectively covered already — the remaining work here is the **Save** temp-file flow.

**Focal point — must be verified to actually save as part of this issue.** The focal point is now tracked as editor state in `ImageEditorStore` (`focalPoint`, set via the focal tool), but it is intentionally **not** persisted on its own. We removed the per-move `/filter/FocalPoint/fp/…?overwrite=` call because that write only stages: `FocalPointImageFilter` prepends `TMP::` to the inode "leaving that work to the check-in process" (`BinaryExporterServlet.copyMetadata`) — without a save/check-in it never commits to the content.

When implementing Save here:
- Commit the focal point as part of the save/check-in (the temp-file `copyMetadata` path that copies custom metadata → the saved asset, or include `focalPoint` in the save payload), and **verify it persists**: open → set focal → save → reopen/reload → marker restored.
[...]

Issue #36062, opened 2026-06-08
[IMAGE EDITOR] Implement AngularImageEditorLauncher full-screen modal for new edit content


nicobytes:
## Description

Implement `AngularImageEditorLauncher` (`DotImageEditorLauncher`) that opens the new Angular image editor as a full-screen PrimeNG `DialogService` modal in the **new** edit-content UI. Replace `NoopImageEditorLauncher` in the file-field provider chain. Keep `LegacyDojoImageEditorLauncher` untouched for JSP / `dotcms-binary-field` web component.

## Acceptance Criteria

- [ ] `AngularImageEditorLauncher.isAvailable()` returns `true` when enterprise license allows (match legacy Dojo gate).
- [ ] `open(context)` opens `DotImageEditorComponent` full-screen (Lara dark-canvas layout).
- [ ] Returns `Observable<DotCMSTempFile | null>` — temp file on Save, `null` on Cancel.
- [ ] Provided via `IMAGE_EDITOR_LAUNCHER` in `dot-edit-content-file-field` (not in web-component builder).
- [ ] `LegacyDojoImageEditorLauncher` remains the provider for `dotcms-binary-field-builder`.

## Priority

High

## Additional Context

- Epic: #35846
- Depends on: #36056 (launcher seam), #36058 (edit action wiring)
- Path: `core-web/libs/edit-content/src/lib/fields/shared/image-editor/`

erickgonzalez:
QA Passed, tested pre-merged in falconteam-image-editor.dotcms.dev

PR #36236, opened 2026-06-18
feat(image-editor): add Angular image editor

### LINKED ISSUE #36063: [IMAGE EDITOR] Build DotImageEditorComponent modal shell and image-editor signalStore
## Description

Create the full-screen image editor modal shell matching the Lara design: dark canvas, left tool rail (move / crop / focal), right-side panels container, bottom Cancel / Download / Save bar, preview address field, and zoom/fit controls. State lives in an NgRx `signalStore` (`image-editor.store.ts`).

## Acceptance Criteria

- [ ] `DotImageEditorComponent` standalone, OnPush, PrimeNG + Tailwind (Lara).
- [ ] `ImageEditorStore` holds asset context, adjust/transform/crop/fileInfo/focalPoint/zoom state.
- [ ] Header, canvas area, tool rail, panel column, and footer action bar match design spec.
- [ ] `data-testid` on primary controls.
- [ ] i18n keys added to `Language.properties` for editor labels.

## Priority

High

## Additional Context

- Epic: #35846
- Design: linked from #35846
- Depends on: #36056, AngularImageEditorLauncher task

### LINKED ISSUE #36064: [IMAGE EDITOR] Image editor server-side preview pipeline via /contentAsset/image filter URLs
## Description

Build the live preview pipeline for the Angular image editor using existing server-side filter URLs (`/contentAsset/image/{id}/{field}/filter/...`). Same filter vocabulary as the legacy Dojo `ImageEditor.js` (Resize, Crop, Rotate, Flip, Hsb, Grayscale, Quality/Jpeg/WebP, FocalPoint).

## Acceptance Criteria

- [ ] `image-editor-url.builder.ts` builds filter-chain URLs from editor state.
- [ ] Live `` preview bound to computed preview URL in the store.
- [ ] Zoom in/out/fit controls adjust canvas scale (UI only; filters unchanged).
- [ ] File size fetched via HTTP HEAD on preview URL and shown in File info panel.
- [ ] Cache-bust query param on preview URLs so changes reflect immediately.
- [ ] Unit tests for URL builder (filters, params, HSB scale, flip, compression).

## Priority

High

## Additional Context

- Epic: #35846
- Depends on: modal shell / store (#36063)
- Reference: `dotCMS/src/main/webapp/html/js/dotcms/dijit/image/ImageEditor.js`

### LINKED ISSUE #36065: [IMAGE EDITOR] Image editor panels: Adjust, Transform, Flip, Output dimensions, File info
## Description

Implement the right-side editor panels with parity to the Dojo editor operations:

| Panel | Filters |
|---|---|
| Adjust | Brightness/Hue/Saturation → `Hsb`; Grayscale → `Grayscale` |
| Transform | Scale → `Resize`; Rotate → `Rotate` |
| Flip | Horizontal → `Flip` (`flip_flip/1`) — **no vertical flip** (backend unsupported) |
| Output | Dimensions reflecting resize/crop result |
| File info | Compression none/auto/jpeg/webp + quality slider |

## Acceptance Criteria

- [ ] Adjust panel: sliders for brightness/hue/saturation (-100..100) and grayscale checkbox; updates preview.
- [ ] Transform panel: scale % and rotate degrees sliders.
- [ ] Horizontal flip toggle only (vertical flip omitted — document in open-questions spike).
- [ ] Output dimensions inputs synced with crop/resize state.
- [ ] File info: compression select + quality slider (when not "none"); live file size display.

## Priority

High

## Additional Context

- Epic: #35846
- Depends on: preview pipeline issue

### LINKED ISSUE #36066: [IMAGE EDITOR] Image editor interactive crop tool and focal point picker with seeding
## Description

Add interactive canvas tools for crop and focal point, matching Dojo editor behavior:

- **Crop tool:** draggable/resizable box on canvas → `crop_w`, `crop_h`, `crop_x`, `crop_y`
- **Focal point tool:** click on canvas → `fp/{x},{y}` normalized [0,1]
- **Seeding:** read existing focal point from metadata on open (fixes JSP bridge hardcoding `0.0,0.0`)

## Acceptance Criteria

- [ ] Left rail tools: move, crop, focal — active tool highlighted.
- [ ] Crop overlay draggable/resizable; updates store crop state and preview URL.
- [ ] Focal point marker visible in focal mode; click sets coordinates.
- [ ] Existing focal point loaded on open (REST read or metadata); not defaulted to 0,0 when stored value exists.
- [ ] Focal point persisted on Save via save endpoint / `FocalPointAPI`.

## Priority

High

## Additional Context

- Epic: #35846
- Depends on: preview pipeline, backend save endpoint (#36061)
- JSP gap: `edit_field.jsp` passes `focalPoint: "0.0,0.0"` hardcoded

oidacra:
## Summary

https://github.com/user-attachments/assets/0b74c648-88fe-4a79-92df-04ea75a28d29

Replaces the legacy Dojo `ImageEditor.js` with a new standalone Angular library
`@dotcms/image-editor`, wired into Edit Content's binary field through an
`IMAGE_EDITOR_LAUNCHER` seam (Angular-only path).

The editor is a **viewer of a dotCMS endpoint**: the `` `src` is a computed
`/contentAsset/image/{id}/{field}/filter/...` URL, so every control change rebuilds
the URL and the server renders the adjusted image — no client-side pixel work.

State is an **events-based NgRx Signal Store** (`@ngrx/signals/events`), composed of
**one vertical `signalStoreFeature` per area of functionality** (adjust, transform,
crop, focal point, file info, tools, history, asset, preview, download) — each
bundling its own reducers, selectors and effects.

## Scope note

This PR delivers the editor shell, the server-side preview pipeline, the panels and
the crop/focal tools, plus **Download**. **Saving the edited image back to the field
is intentionally deferred to #36067** — the editor currently previews and downloads.

## What's included

- **Library `@dotcms/image-editor`**: root dialog, header, canvas (crossfade, zoom +
  pan, crop/focal overlays), side panels (Adjust / Transform / File info / History),
  footer (Cancel + Download).
- **Server-side preview** via the filter-URL builder (single source of truth) and Download.
- **Command history**: removable applied-edits list + undo/redo, keyboard shortcuts
  (Ctrl/Cmd+Z, Ctrl/Cmd+Shift+Z, Ctrl+Y), ignored while a text field is focused.
- **Launcher seam** in `libs/edit-content` (token + Angular launcher) and binary-field wiring.
- **Canvas interactions**: zoom with drag-to-pan, fit-to-screen, crop-to-current-view,
  focal-centered aspect crop.

## Preview robustness

- Preview fetched as a **complete, verified blob** (local object URL), so a
  partially-generated server response can't paint a truncated frame; incomplete
  responses are detected and retried silently.
- `decode()` + natural-dimension gate before promoting a frame.

## Code organization

- Constants in one `image-editor.constants.ts`; types/models in `models/image-editor.models.ts`.
- Store split into `store/features/with-*.feature.ts` + `store-utils.ts`; `store.ts` is a thin composition.

## Testing

- `nx lint image-editor` / `nx lint edit-content` — clean
- `nx test image-editor` — **258 passing** (per-feature unit specs + store-utils + dimensions.util + integration; store-feature branch coverage ~100%)
- `nx test edit-content` — **1961 passing** (28 skipped)
- Verified live (local dotCMS, demo content): preview, crop, focal, zoom/pan, fit, download, undo/redo (incl. Cmd+Z), panel persistence.

## Issues

- Closes #36063 — modal shell + image-editor signalStore
- Closes #36064 — server-side preview pipeline via `/contentAsset/image` filter URLs
- Closes #36065 — panels (Adjust, Transform, Flip, Output dimensions, File info)
- Closes #36066 — interactive crop tool + focal point picker with seeding
- Partially addresses #36067 — Download and Cancel are wired; **Save** to the FileFieldStore temp-file flow is deferred to that issue
- Related #36062 — the `AngularImageEditorLauncher` modal is implemented, but true **full-screen sizing is not working yet** (follow-up there)

This PR fixes: #36063


wezell:
Love this so much and am happy to ship as is but here are some of my thoughts:

- love the cropper and presets but do we need a vertical "phone" crop as well?, e.g. 9x16 or whatever the kids use nowadays?  Copy instagram or something....

- the image is 2752 x 1536 but my zoom says I am viewing at 100%?  We need to do the math I think?

- Transform and file info should be open be default (I think).  Adjust is less used.  
- The header accordians might take up too much vertical space as headers?
- Clicking on the url should drop it in your clipboard. copying the url is the # 1 use case.


- We now have avif for compression as well.  Also have jpxl but jpxl is not well supported.   avif is and generates smaller images than webp.  e.g. `/filter/avif/avif_q/75`
   - https://falconteam-image-editor.dotcms.dev/contentAsset/image/0050282e-2393-4203-bbea-739629da58a3/binaryField/filter/avif/avif_q/60
or lossless:
   - https://falconteam-image-editor.dotcms.dev/contentAsset/image/0050282e-2393-4203-bbea-739629da58a3/binaryField/filter/avif/avif_lossless/true

- We now have different crop options - width and height `/filter/smartcrop/smartcrop_w/500/smartcrop_h/400`
and modes: attention , entropy, centre.  `attention` is the default:
   - https://falconteam-image-editor.dotcms.dev/contentAsset/image/0050282e-2393-4203-bbea-739629da58a3/binaryField/filter/smartcrop/smartcrop_w/500/smartcrop_h/400/smartcrop_mode/attention
   - https://falconteam-image-editor.dotcms.dev/contentAsset/image/0050282e-2393-4203-bbea-739629da58a3/binaryField/filter/smartcrop/smartcrop_w/500/smartcrop_h/400/smartcrop_mode/entropy
   - https://falconteam-image-editor.dotcms.dev/contentAsset/image/0050282e-2393-4203-bbea-739629da58a3/binaryField/filter/smartcrop/smartcrop_w/500/smartcrop_h/400/smartcrop_mode/centre
    
- I don't see focalpoint picker anywhere, am I missing it?
- Where is the save button to save the new rendition to the content?

oidacra:
> Love this so much and am happy to ship as is but here are some of my thoughts:
> 
> * love the cropper and presets but do we need a vertical "phone" crop as well?, e.g. 9x16 or whatever the kids use nowadays?  Copy instagram or something....
> * the image is 2752 x 1536 but my zoom says I am viewing at 100%?  We need to do the math I think?
> * Transform and file info should be open be default (I think).  Adjust is less used.
> * The header accordians might take up too much vertical space as headers?
> * Clicking on the url should drop it in your clipboard. copying the url is the # 1 use case.
> * We now have avif for compression as well.  Also have jpxl but jpxl is not well supported.   avif is and generates smaller images than webp.  e.g. `/filter/avif/avif_q/75`
>   
>   * https://falconteam-image-editor.dotcms.dev/contentAsset/image/0050282e-2393-4203-bbea-739629da58a3/binaryField/filter/avif/avif_q/60
>     or lossless:
>   * https://falconteam-image-editor.dotcms.dev/contentAsset/image/0050282e-2393-4203-bbea-739629da58a3/binaryField/filter/avif/avif_lossless/true
> * We now have different crop options - width and height `/filter/smartcrop/smartcrop_w/500/smartcrop_h/400`
>   and modes: attention , entropy, centre.  `attention` is the default:
>   
>   * https://falconteam-image-editor.dotcms.dev/contentAsset/image/0050282e-2393-4203-bbea-739629da58a3/binaryField/filter/smartcrop/smartcrop_w/500/smartcrop_h/400/smartcrop_mode/attention
>   * https://falconteam-image-editor.dotcms.dev/contentAsset/image/0050282e-2393-4203-bbea-739629da58a3/binaryField/filter/smartcrop/smartcrop_w/500/smartcrop_h/400/smartcrop_mode/entropy
>   * https://falconteam-image-editor.dotcms.dev/contentAsset/image/0050282e-2393-4203-bbea-739629da58a3/binaryField/filter/smartcrop/smartcrop_w/500/smartcrop_h/400/smartcrop_mode/centre
> * I don't see focalpoint picker anywhere, am I missing it?
> * Where is the save button to save the new rendition to the content?
@wezell @melissarojas-dotcms 
- `vertical "phone" crop as well` ✅ fixed
- `the image is 2752 x 1536 but my zoom says I am viewing at 100%` ✅ fixed
- `Transform and file info should be open be default` ✅ now opened by default
- ` We now have avif for compression as well. ` ✅ added AVIF.
- `We now have different crop options` I will create an issue for the smartcrop only, we need add a button for that (design)
- `I don't see focalpoint picker anywhere, am I missing it` same as preview, I will create a issue for that one
- `Where is the save button to save the new rendition to the content` all the flow of saving, is an issue  https://github.com/dotCMS/core/issues/36067

core-web/libs/image-editor/src/lib/store/features/with-transform.feature.ts
```diff
@@ -0,0 +1,72 @@
...
+    return signalStoreFeature(
+        type<{ state: ImageEditorState }>(),
+        withReducer(
+            on(imageEditorTransformEvents.scaleChanged, ({ payload }, state) => {
+                const value = clamp(payload, RANGES.scale.min, RANGES.scale.max);
+                // Resizing supersedes crop, mirroring the filter-chain rule.
+                const transform: TransformState = { ...state.transform, scale: value };
+                const crop = value !== 100 ? initialCropState : state.crop;
+
+                return transformPatch(state, transform, crop, 'adjust', `Scale ${value}%`);
```
oidacra: `scaleChanged` (and the Resize handler below) pass the `'adjust'` history category, the same bucket used by brightness/hue/saturation, while `rotate` and `flip` get their own categories. Since history coalesces consecutive same-category edits into one entry, a color adjustment followed by a scale collapses into a single undo step — undoing the scale also reverts the adjustment. Consider giving scale/resize their own category so they form distinct history entries.

core-web/libs/image-editor/src/lib/components/dot-image-editor/dot-image-editor.component.ts
```diff
@@ -0,0 +1,202 @@
...
+
+        if (!el) {
+            return false;
+        }
+
+        return (
+            el.tagName === 'INPUT' ||
+            el.tagName === 'TEXTAREA' ||
+            el.tagName === 'SELECT' ||
+            el.isContentEditable
```
oidacra: The editable-target guard only matches `INPUT`/`TEXTAREA`/`SELECT` and `isContentEditable`. PrimeNG composite controls (`p-select`/`p-dropdown`/`p-autocomplete`) expose a focusable element with `role=combobox` and `contenteditable="false"`, so they pass this guard. As a result, `Ctrl/Cmd+Z` or `Ctrl+Y` typed while one of those controls is focused triggers the editor's undo/redo instead of the control's native behavior. Consider also blocking when `target.closest('p-select, p-dropdown, p-autocomplete, p-inputnumber, [contenteditable="true"]')` matches.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-binary-field/dot-edit-content-binary-field.component.ts
```diff
@@ -389,14 +395,47 @@ export class DotEditContentBinaryFieldComponent
...
+                fieldName: this.$field()?.name,
+                byInode: !!inode,
+                fileName: this.contentlet?.fileName ?? metadata?.name,
+                mimeType: metadata?.contentType
+            })
+            .pipe(
+                filter((tempFile): tempFile is DotCMSTempFile => !!tempFile),
+                takeUntilDestroyed(this.#destroyRef)
+            )
+            .subscribe((tempFile) => this.#dotBinaryFieldStore.setFileFromTemp(tempFile));
```
oidacra: This `subscribe` has only a `next` callback. If the launcher open / dialog `onClose` stream errors, it propagates to the global error handler uncaught and the edited image is silently never applied to the field, with no feedback to the user. Consider adding an `error` callback that logs and optionally surfaces a toast.

core-web/libs/image-editor/src/lib/services/dot-image-editor.service.ts
```diff
@@ -0,0 +1,138 @@
...
+            dimensions: this.#resolveNaturalDimensions(ctx.originalUrl),
+            originalBytes: this.getFileSize(ctx.originalUrl)
+        }).pipe(
+            map(({ dimensions, originalBytes }) => ({
+                naturalWidth: dimensions.naturalWidth,
+                naturalHeight: dimensions.naturalHeight,
+                originalBytes
+            })),
+            catchError(() =>
+                of<AssetMeta>({ naturalWidth: 0, naturalHeight: 0, originalBytes: null })
```
oidacra: This `catchError` returns a 0×0 / null-bytes fallback silently. Because `resolveNaturalDimensions` already soft-catches its own `onerror`, this outer catch only fires on a genuinely unexpected failure (e.g. CORS/403/404 on the asset), and when it does the editor opens in a degraded 0×0 state with nothing logged to diagnose it. Consider logging the error before returning the fallback.

core-web/libs/image-editor/src/lib/components/dot-image-editor-crop-overlay/dot-image-editor-crop-overlay.component.ts
```diff
@@ -0,0 +1,530 @@
...
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    imports: [DotMessagePipe],
+    templateUrl: './dot-image-editor-crop-overlay.component.html',
+    styleUrl: './dot-image-editor-crop-overlay.component.scss',
+    animations: [imageEditorOverlayEnterLeave()],
+    host: { '(keydown.escape)': 'onEscape($event)' }
+})
+export class DotImageEditorCropOverlayComponent {
+    /** Bounds of the rendered image within the canvas, in CSS px. */
+    imageRect = input<ImageRect>();
```
oidacra: Signal members throughout the new library omit the `$` prefix required by ANGULAR_STANDARDS.md ("For signals, use the `$` prefix"). The neighboring edit-content sidebar components already follow this convention. This is a library-wide rename: `input()`/`output()`/`signal()`/`computed()` members in the canvas, overlays, address-bar, header, footer, and panel components should be `$name`, with templates updated to match. Flagging here as a representative location.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-binary-field/dot-edit-content-binary-field.component.ts
```diff
@@ -389,14 +395,47 @@ export class DotEditContentBinaryFieldComponent
...
+                tempId: this.tempId,
+                variable: this.variable
+            });
+
+            return;
+        }
+
+        const inode = this.contentlet?.inode;
+        const metadata = this.contentlet
+            ? (getFileMetadata(this.contentlet) as Partial<DotFileMetadata>)
```
oidacra, line 425: `as Partial<DotFileMetadata>` is an unvalidated assertion. The same pattern appears elsewhere in the PR (`as HTMLElement | undefined` on the dialog container in dot-image-editor.component.ts, and `as unknown as Record<string, unknown>` / `as SlicePatch` in image-editor.store-utils.ts). Prefer `satisfies`, an `instanceof` guard, or making the helper generic over `keyof EditableSlices` so the cast isn't needed.

core-web/libs/edit-content/src/lib/fields/shared/image-editor-launcher/image-editor-launcher.token.ts
```diff
@@ -0,0 +1,16 @@
...
+import { DotImageEditorLauncher } from '@dotcms/image-editor';
+
+export type { DotImageEditorLauncher, ImageEditorOpenParams } from '@dotcms/image-editor';
+
+/**
+ * DI seam for launching the image editor from the binary field.
+ *
+ * The Angular edit-content shell provides the dialog-based launcher. When the
+ * token is left unprovided, the binary field injects it as optional and hides
+ * the "edit image" action, so no fallback implementation is needed.
```
oidacra: This comment states the action is hidden and no fallback is needed when the token is unprovided, but `onEditImage()` actually falls back to the legacy Dojo editor via `DotBinaryFieldEditImageService.openImageEditor(...)` when `isAvailable()` is false. The comment block in dot-edit-content-binary-field.component.ts ("safely no-ops") has the same inaccuracy. Consider rewording both to "falls back to the legacy Dojo image editor."

core-web/libs/image-editor/src/lib/components/dot-image-editor-canvas/dot-image-editor-canvas.component.ts
```diff
@@ -0,0 +1,596 @@
...
+import { DotImageEditorService } from '../../services/dot-image-editor.service';
+import { imageEditorLifecycleEvents } from '../../store/image-editor.events';
+import { ImageEditorStore } from '../../store/image-editor.store';
+import { clamp } from '../../utils/dimensions.util';
+import { DotImageEditorAddressBarComponent } from '../dot-image-editor-address-bar/dot-image-editor-address-bar.component';
+import { DotImageEditorCropOverlayComponent } from '../dot-image-editor-crop-overlay/dot-image-editor-crop-overlay.component';
+import { DotImageEditorFocalOverlayComponent } from '../dot-image-editor-focal-overlay/dot-image-editor-focal-overlay.component';
+
+/**
+ * Dark stage that renders the live image preview at the center of the editor.
```
oidacra: The JSDoc describes a "Dark stage," but the canvas background is `var(--surface-0, #ffffff)` (white). Consider dropping "Dark" so the comment matches the styling.

dotCMS/src/main/java/com/dotcms/rest/api/v1/system/ConfigurationResource.java
```diff
@@ -92,7 +92,8 @@ public class ConfigurationResource implements Serializable {
             FeatureFlagName.FEATURE_FLAG_UVE_LEGACY_SCRIPT_INJECTION,
             FeatureFlagName.FEATURE_FLAG_NEW_BLOCK_EDITOR,
             FeatureFlagName.FEATURE_FLAG_CONTENT_EDITOR2_ENABLED,
-            FeatureFlagName.FEATURE_FLAG_LOCALE_SELECTOR_V2);
+            FeatureFlagName.FEATURE_FLAG_LOCALE_SELECTOR_V2,
+            FeatureFlagName.FEATURE_FLAG_NEW_IMAGE_EDITOR);
```
oidacra: The flag is correctly added to both `BOOLEAN_FEATURE_FLAGS` and `WHITE_LIST`, but these two hand-maintained lists have no enforcement that they stay in sync (and `FEATURE_FLAG_REPORT_ISSUE_ENABLED` has already drifted). If a future flag lands in only one list, the frontend receives the raw string `"false"`, which is truthy, silently enabling the feature. A guard test such as `assertTrue(WHITE_LIST.containsAll(BOOLEAN_FEATURE_FLAGS))` was requested on the PR that introduced this structure (#35769) and would be a cheap safeguard.

core-web/libs/image-editor/src/lib/services/dot-image-editor.service.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+
+            httpMock
+                .expectOne('/dA/preview.png')
+                .flush(blob, { headers: { 'Content-Type': 'text/html' } });
+
+            expect(errored).toBe(true);
+        });
+    });
+
+    describe('triggerDownload', () => {
```
oidacra, line 152: This test asserts `href`, `download`, and `click()` but not the DOM side-effects the method actually performs: `document.body.appendChild(anchor)` and `anchor.remove()`. Without those assertions a regression that drops the append/remove would go undetected. Consider spying on `document.body.appendChild` and `anchor.remove` and asserting both are called.

core-web/libs/image-editor/src/lib/store/features/with-crop.feature.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const CropStore = signalStore(withState(RESIZED), withCrop());
+const CropStoreActive = signalStore(
+    withState({ ...initialImageEditorState, crop: ACTIVE_CROP, activeTool: 'crop' as const }),
+    withCrop()
+);
+
+describe('withCrop', () => {
+    it('applies a crop, clears resize, returns to move and adds a crop entry', () => {
+        TestBed.configureTestingModule({ providers: [CropStore, Dispatcher] });
```
oidacra: `TestBed.configureTestingModule` is called inside the `it` bodies with no `beforeEach`/reset, and the second test configures a different store class. Without a reset the module configuration can leak between tests, creating an inter-test isolation gap. Consider moving each setup into its own nested `describe` with a `beforeEach`. (Several other feature specs also configure TestBed without `resetTestingModule` — worth confirming the Jest preset resets between tests.)

core-web/libs/image-editor/src/lib/services/dot-image-editor.service.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+
+import { provideHttpClient } from '@angular/common/http';
+import { HttpTestingController, provideHttpClientTesting } from '@angular/common/http/testing';
+
+import { DotImageEditorService } from './dot-image-editor.service';
+
+// jsdom has no object-URL API; the service creates one for each verified blob.
+URL.createObjectURL = jest.fn(() => 'blob:mock-object-url');
+
+describe('DotImageEditorService', () => {
```
oidacra, line 8: `loadAssetMeta` has no test coverage. It is the primary asset-boot integration point (composes `resolveNaturalDimensions` with `getFileSize` via `forkJoin`) and has a non-trivial error fallback. Consider adding a describe block covering both the happy path (correct dimensions + byte size) and the `catchError` fallback (`{0, 0, null}`).

core-web/libs/image-editor/src/lib/components/dot-image-editor-panels/dot-image-editor-panels.component.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+        expect(spectator.query(byTestId('image-editor-panel-adjust'))).toExist();
+        expect(spectator.query(byTestId('image-editor-panel-transform'))).toExist();
+        expect(spectator.query(byTestId('image-editor-panel-fileinfo'))).toExist();
+        expect(spectator.query(byTestId('image-editor-panel-history'))).toExist();
+    });
+
+    it('should start with every section open by default', () => {
+        spectator = createComponent();
+
+        expect(spectator.component['openPanels']()).toEqual([
```
oidacra: These assertions reach into `protected` members via bracket notation (`spectator.component['openPanels']()`, `['onOpenPanelsChange'](...)`), which couples the test to implementation details and bypasses the template — a broken accordion binding would still pass. The transform-panel and canvas specs do the same via `as unknown as` casts. Consider driving these through the template (trigger the accordion `valueChange`, assert the `localStorage` side-effect / rendered DOM).

core-web/libs/image-editor/src/lib/components/dot-image-editor-canvas/dot-image-editor-canvas.component.scss
```diff
@@ -0,0 +1,299 @@
...
+    gap: 0.5rem;
+    // Size to content (capped to the viewport) so the bar is as wide as it needs to
+    // be. Without an explicit width an absolutely-positioned box is capped at the
+    // space right of `left: 50%`, which squeezed "Apply crop" onto two lines.
+    width: max-content;
+    max-width: calc(100% - 2rem);
+    padding: 0.375rem 0.625rem;
+    border-radius: 0.75rem;
+    background-color: var(--surface-0, #ffffff);
+    box-shadow: 0 4px 16px rgba(0, 0, 0, 0.18);
```
oidacra, line 60: Hardcoded color/shadow values are used here and in several SCSS files in this PR (raw `rgba(...)` and hex literals, plus `var(--token, #hex)` fallbacks in canvas, crop-overlay, focal-overlay, adjust-panel, transform-panel, panels, and address-bar). STYLING_STANDARDS.md asks never to hardcode colors/shadows — prefer Tailwind utilities (e.g. `shadow-md`, `bg-black/50`) or design-token variables, and drop the hex fallbacks. Flagging here as a representative location.

core-web/libs/image-editor/src/lib/components/dot-image-editor-panels/dot-image-editor-panels.component.scss
```diff
@@ -0,0 +1,177 @@
+:host {
+    display: block;
+}
+
+// Accordion section styling, aligned to the Claude Design "Edit image" spec:
+// header title 600/13px, subtitle 400/11px muted, a 26px rounded icon chip that
+// turns primary when the section is open, and the design's 14/18px density.
+// Scoped to this component's subtree via :host; pierces PrimeNG internals with
+// ::ng-deep since the header content is projected into <p-accordion-header>.
```
oidacra, line 9: Two points on this file's styling: the panels container reaches into child-component internals (`label`, `.p-slider`, `.p-inputnumber*`, `.ie-transform__*`) through `::ng-deep`, which breaks encapsulation — those rules belong in each sub-panel's own SCSS. Several rules here also re-implement layout/spacing that Tailwind covers (`display:flex`, `gap`, etc.). Consider moving child styles down and replacing layout-only SCSS with utility classes.

core-web/libs/dotcms-models/src/lib/shared-models.ts
```diff
@@ -36,7 +36,8 @@ export const enum FeaturedFlags {
     FEATURE_FLAG_UVE_LEGACY_SCRIPT_INJECTION = 'FEATURE_FLAG_UVE_LEGACY_SCRIPT_INJECTION',
     FEATURE_FLAG_NEW_BLOCK_EDITOR = 'FEATURE_FLAG_NEW_BLOCK_EDITOR',
     FEATURE_FLAG_REPORT_ISSUE_ENABLED = 'FEATURE_FLAG_REPORT_ISSUE_ENABLED',
-    FEATURE_FLAG_LOCALE_SELECTOR_V2 = 'FEATURE_FLAG_LOCALE_SELECTOR_V2'
+    FEATURE_FLAG_LOCALE_SELECTOR_V2 = 'FEATURE_FLAG_LOCALE_SELECTOR_V2',
+    FEATURE_FLAG_NEW_IMAGE_EDITOR = 'FEATURE_FLAG_NEW_IMAGE_EDITOR'
```
oidacra, line 40: This adds a new member to a `const enum`. TYPESCRIPT_STANDARDS.md prohibits enums in favor of `as const` objects. The enum is pre-existing, so this isn't a new violation on its own, but it extends a prohibited pattern — worth tracking the migration to an `as const` map.

core-web/libs/image-editor/src/lib/store/image-editor.store-utils.ts
```diff
@@ -0,0 +1,266 @@
...
+        const history = [
+            ...state.history.slice(0, state.historyIndex),
+            { ...head, label, snapshot }
+        ];
+
+        return { history, historyIndex: state.historyIndex };
+    }
+
+    const entry: ImageEditorHistoryEntry = {
+        id: `${category}-${Date.now()}-${state.cacheBust}`,
```
oidacra: Using `Date.now()` inside a history-entry id makes id generation non-deterministic and can collide under fake timers (two entries in the same millisecond share an id). Consider `crypto.randomUUID()` or a monotonic counter carried in state to keep this pure and test-stable.

core-web/libs/image-editor/src/lib/utils/panel-state.storage.ts
```diff
@@ -0,0 +1,49 @@
...
+    try {
+        const stored = localStorage.getItem(IMAGE_EDITOR_PANEL_STATE_KEY);
+        if (stored) {
+            const parsed = JSON.parse(stored);
+            if (Array.isArray(parsed) && parsed.every((value) => typeof value === 'string')) {
+                return parsed;
+            }
+        }
+    } catch (e) {
+        console.warn('Error reading image editor panel state from localStorage:', e);
```
oidacra: `console.warn` in production library code. The `try/catch` already handles the localStorage failure gracefully; consider removing the console call or routing it through the application logger.

core-web/libs/image-editor/src/lib/image-editor.constants.ts
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Times a failed preview is silently retried before the error UI is shown. The
+ * server renders filter chains on the fly, so the first request for a fresh URL
+ * can race that generation and return an incomplete response; a couple of silent
+ * re-attempts (mirroring a manual refresh) almost always lands the finished image.
+ */
+export const AUTO_PREVIEW_RETRY_LIMIT = 3;
+
+/** Human-readable labels per compression mode, used in history entries. */
+export const COMPRESSION_LABELS: Record<CompressionMode, string> = {
```
oidacra, line 61: `COMPRESSION_LABELS` maps to hardcoded English strings, but these labels are user-visible (they appear in history entries). Other user-facing text in this PR goes through the `dm` message pipe. Consider using i18n keys and resolving them at the point of display.

core-web/libs/image-editor/src/lib/components/dot-image-editor-panels/dot-image-editor-transform-panel/dot-image-editor-transform-panel.component.ts
```diff
@@ -0,0 +1,159 @@
...
+        this.dispatch.flipHToggled();
+    }
+
+    /** Dispatches a vertical flip toggle. */
+    protected flipVToggled(): void {
+        this.dispatch.flipVToggled();
+    }
+
+    /** Dispatches a change to the explicit output width, guarding the minimum. */
+    protected outputWidthChanged(value: number | null): void {
```
oidacra, line 114: `outputWidthChanged` dispatches `height: this.store.transform().outputHeight`, which is the raw (possibly `null`) stored value, while the template displays `store.outputDimensions().height` (a computed fallback to the natural height). After a width-only change the displayed height looks set, but the stored `outputHeight` stays `null`. If preserving the displayed height is intended, read from `outputDimensions()` instead so the dispatched value matches what the user sees.

core-web/libs/image-editor/src/lib/store/features/with-preview.feature.ts
```diff
@@ -0,0 +1,112 @@
...
+                isBusy: computed(() => store.previewStatus() === 'loading')
+            };
+        }),
+        withEventHandlers((store) => {
+            const dispatcher = inject(Dispatcher);
+            const service = inject(DotImageEditorService);
+
+            return {
+                // Resolve the edited preview size, debounced against rapid edits.
+                resolveSize$: toObservable(store.previewUrl).pipe(
```
oidacra, line 117: The `resolveSize$` `tap` pipeline has no `catchError`: if the inner dispatch throws, the stream terminates and the file-size readout silently freezes for the rest of the session. (The `getFileSize` service call self-catches, so this is defense-in-depth.) Separately, the preview error path stores the hardcoded string `'Failed to render preview'`, discarding the root cause — consider threading the underlying error through the `previewErrored` payload.

core-web/libs/ui/src/lib/theme/theme.config.ts
```diff
@@ -24,6 +24,45 @@ export const CustomLaraPreset = definePreset(Lara, {
         primary: DotUiColorsService.getDefaultPrimeNGPalette()
     },
     components: {
+        accordion: {
```
oidacra, line 27: These `accordion` token overrides live in the global `CustomLaraPreset`, so they affect every `p-accordion` across the admin UI, not just the image editor. This file has a history of post-migration regressions with broad blast radius. If the flat/borderless style is only needed inside the editor modal, consider scoping it to the component rather than the global preset.

core-web/libs/image-editor/src/lib/store/features/with-file-info.feature.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    it('sets the compression mode under the compression category', () => {
+        fileInfo.compressionChanged('jpeg');
+        expect(store.fileInfo().compression).toBe('jpeg');
+        expect(store.history()[0].category).toBe('compression');
+        expect(store.history()[0].label).toBe('Compression JPEG');
+    });
+
+    it('sets and labels quality', () => {
+        fileInfo.qualityChanged(50);
```
oidacra, line 35: Quality clamping is tested at the upper bound but not the lower bound. The reducer uses `clamp(payload, min, max)` and a negative value is reachable. Consider adding a case like `qualityChanged(-50)` asserting it clamps to the minimum.

core-web/libs/image-editor/src/lib/components/dot-image-editor-canvas/dot-image-editor-canvas.component.spec.ts
```diff
@@ -0,0 +1,500 @@
...
+    configurable: true,
+    get: () => 800
+});
+Object.defineProperty(HTMLImageElement.prototype, 'naturalHeight', {
+    configurable: true,
+    get: () => 600
+});
+
+/** Flushes the microtask queue so a resolved/rejected `decode()` promise settles. */
+const flushDecode = () => new Promise<void>((resolve) => setTimeout(resolve));
```
oidacra, line 65: `flushDecode` uses a real `setTimeout` to flush the microtask queue, which relies on real timers. A plain `Promise.resolve()` (in an async test) or `fakeAsync` + `flushMicrotasks()` would be more deterministic.

adrianjm-dotCMS (approved):

PR #36238, opened 2026-06-18
fix(opensearch): make index bootstrap idempotent for orphaned cluster indices (#36237)

### LINKED ISSUE #36237: OpenSearch: index bootstrap not idempotent — orphaned cluster index aborts startup
## Problem

   ```
   ```

   ```
   ```

## Root cause

## Fix

fabrizzio-dotCMS:
## Proposed Changes

Fixes #36237.

Makes the OpenSearch/Elasticsearch index **bootstrap idempotent** against an *orphaned cluster index* — one that physically exists in the cluster but is missing from the dotCMS index store.

### Root cause

`ContentletIndexAPIImpl.createContentIndex(name, shards, tag)` (used only by the ES/OS bootstrap paths) always issues a create. Readiness is decided by `indexReadyOS()`/`indexReadyES()`, which key off the **store** (`VersionedIndices` / `indicies`), not the physical cluster state. When a prior startup created the physical index but never committed its store pointer (partial/interrupted init, failed `pointOS`, etc.), the two diverge: on restart the catchup path re-derives the same logical name and the create fails with `resource_already_exists`.

That exception bubbles to `checkAndInitializeIndex()` and is swallowed, leaving the instance **half-initialised** — and the read provider then queries an unfinished index, surfacing as `search_phase_execution_exception / all shards failed`.

### The fix

In `createContentIndex(name, shards, tag)`, before creating, check `providerApi.indexExists(physicalName)`:
- **Already exists** → reuse it, re-assert the custom mapping (`putMapping` is additive/idempotent, so a previously unmapped orphan is repaired), and return `true`. The caller's `point()` re-registers it in the store.
- **Does not exist** → create as before.

This is scoped to the two callers (`bootstrapAndPointES` / `bootstrapAndPointOS`); no public/interface signatures change. The `(IndexAPIImpl) indexAPI` cast mirrors the existing `indexReadyES()/indexReadyOS()` usage.

## Testing

- `./mvnw compile -pl :dotcms-core` (JDK 25) — compiles clean.
- No signature changes; `ContentletIndexAPIImplPhaseTest` fakes never exercise the bootstrap path, so existing unit coverage is unaffected.
- A meaningful idempotency IT requires a live cluster (`opensearch-upgrade` profile) since the bootstrap path casts to `IndexAPIImpl` — can be added on request.

## Out of scope (tracked in #36237)

- `checkAndInitializeIndex()` swallowing bootstrap failures instead of failing loudly.
- `ContentFactoryIndexOperationsES` hiding the per-shard `ShardSearchFailure` root cause behind the generic "all shards failed" wrapper.

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImpl.java
+++ b/dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImpl.java
@@ -640,22 +640,50 @@ public class ContentletIndexAPIImpl implements ContentletIndexAPI {
 
 
     /**
-     * Create an index exclusively in one of the SE Providers
-     * @param indexName
-     * @param shards
-     * @param tag
-     * @return
-     * @throws IOException
+     * Create an index exclusively in one of the SE Providers.
+     *
+     * <p><b>Idempotent bootstrap.</b> If the physical index already exists in the target
+     * cluster it is reused instead of issuing a create. This guards against an orphaned
+     * cluster index — present in the cluster but missing from the index store — left behind
+     * when a previous bootstrap created the index but never committed its store pointer
+     * (e.g. the OS {@code VersionedIndices} row, or after a partial/interrupted startup).
+     * Without this guard the restart re-derives the same logical name, the create fails with
+     * {@code resource_already_exists}, and {@code checkAndInitializeIndex()} aborts — leaving
+     * the instance half-initialised. The custom mapping is (re)applied either way
+     * ({@code putMapping} is additive/idempotent), so a previously unmapped orphan is repaired,
+     * and the caller's {@code point()} re-registers the index in the store.</p>
+     *
+     * @param indexName logical index name (no cluster prefix, no vendor tag)
+     * @param shards    number of shards to create with (ignored when the index already exists)
+     * @param tag       target provider ({@link IndexTag#ES} or {@link IndexTag#OS})
+     * @return {@code true} when the index exists (reused) or was created successfully
+     * @throws IOException on a hard creation failure
      */
     private boolean createContentIndex(final String indexName, final int shards, IndexTag tag)
             throws IOException {
         final MappingHelper helper = MappingHelper.getInstance();
+        final IndexAPIImpl impl = (IndexAPIImpl) indexAPI;
 
         ContentletIndexOperations ops = router.esImpl();
+        IndexAPI providerApi = impl.esImpl();
         if(tag == IndexTag.OS) {
            ops = router.osImpl();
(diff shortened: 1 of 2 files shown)


ihoffmann-dot (approved):

nollymar (commented):
Could some tests be implemented to validate this logic?

fabrizzio-dotCMS:
@nollymar good call — added unit coverage in `ContentletIndexAPIImplBootstrapTest` (commit f3329fa791).

To make the orphan-reuse logic testable without a running ES/OS cluster or the `MappingHelper` singleton, I extracted the idempotent-bootstrap decision in `createContentIndex(name, shards, tag)` into a collaborator-injected package-private seam (`ops` / `providerApi` / `helper` passed in). The public method now just resolves those and delegates — behaviour is unchanged.

The 5 cases cover:
- orphaned cluster index already exists → reused, mapping re-asserted, **no** create issued
- index missing + create succeeds → created with the resolved physical name/shards + mapping applied
- index missing + create fails → no mapping, returns `false` (failure not masked)
- existence probe throws → treated as "does not exist", falls through to the create path (no bootstrap abort)
- `IndexTag.OS` is propagated unchanged to the mapping helper

All 19 tests in the suite pass (5 new + 14 existing `ContentletIndexAPIImplPhaseTest`).

fabrizzio-dotCMS:
Thanks for the review @github-actions (Bedrock). Resolution of the three 🟡 Medium points:

1. **`Try.of` swallows the `indexExists` probe failure (L667)** — ✅ **Applied** in `1fc4c342c7`. Kept the best-effort "treat as does-not-exist" semantics (a transient probe error must not abort bootstrap), but added `.onFailure(...)` to log the cause at **DEBUG**, so a real connectivity/config problem stays traceable instead of being silently masked.

2. **Parameterized logging `{}` instead of `String.format` (L670)** — ⏭️ **Not applicable.** dotCMS's `com.dotmarketing.util.Logger` does not support SLF4J-style `{}` placeholders (signatures are `(Object, String)` / `(Object, Supplier<String>)`); using `{}` would print the literal braces. `String.format`/concatenation is the existing convention throughout this class.

3. **`mock(IndiciesAPI.class)` is misspelled, should be `IndicesAPI` (test L54)** — ⏭️ **False positive.** `IndiciesAPI` (with that spelling) is the real legacy ES index-store interface at `com.dotcms.content.elasticsearch.business.IndiciesAPI`; the OpenSearch counterpart is `VersionedIndicesAPI`. There is no `IndicesAPI` type — the current name is correct and compiles.

@nollymar's request for tests was already addressed in `f3329fa791` (`ContentletIndexAPIImplBootstrapTest`).

PR #36240, opened 2026-06-18
fix(opensearch): fail loudly on migration init failure + surface ES shard failures (#36237)

### LINKED ISSUE #36237: OpenSearch: index bootstrap not idempotent — orphaned cluster index aborts startup

fabrizzio-dotCMS:
## Proposed Changes

The two diagnostics/robustness follow-ups tracked in #36237 (the bootstrap-idempotency fix is #36238). Both target the same incident class where a migration startup problem silently degraded into a confusing downstream symptom.

### 1. `ContentletIndexAPIImpl.checkAndInitializeIndex()` — fail loudly during migration

The catch-all swallowed **every** bootstrap exception — including the Phase 3 `DotRuntimeException` that is thrown a few lines above specifically to "abort loudly". The instance then came up **half-initialised**, and the read provider queried an unfinished/missing index, surfacing downstream as `all shards failed`.

Now, when a migration is active (`isMigrationStarted() || isReadEnabled() || isMigrationComplete()`), the failure is re-thrown to abort startup so the operator restores connectivity/config and restarts. **Legacy ES-only installs (migration not started) keep the historical best-effort boot** — no behavior change for non-migration deployments.

### 2. `ContentFactoryIndexOperationsES` — surface the per-shard root cause

Both the search and count catch blocks logged only `e.getCause().getMessage()`, hiding the actionable detail behind the generic `search_phase_execution_exception / all shards failed` wrapper.

Added `describeEsFailure(ElasticsearchException)` which renders:
- `getDetailedMessage()` (full ES type/reason chain),
- the per-shard `ShardSearchFailure` reasons (capped at 5, with an "... and N more" summary),
- any suppressed server-side causes.

The enriched message still contains the substrings `shouldQueryCache(...)` keys off, so the error-caching behavior is unchanged.

## Testing

- `./mvnw compile -pl :dotcms-core` (JDK 25) — compiles clean.
- No public/interface signatures changed.

## Relationship to other PRs

- #36238 — makes index bootstrap idempotent (removes the most common trigger of the swallowed failure). This PR is the defense-in-depth + observability layer on top; the two are independent and can merge in any order.

Part of #36237 (does not close it — the parent issue is also addressed by #36238).

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImpl.java
+++ b/dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImpl.java
@@ -554,6 +554,18 @@ public class ContentletIndexAPIImpl implements ContentletIndexAPI {
         } catch (Exception e) {
             Logger.fatal(this.getClass(), "Failed to create new indexes:" + e.getMessage(),e);
 
+            // During an active OpenSearch migration a half-initialised index store is worse
+            // than a hard stop: the read provider would serve an unfinished or missing index
+            // (surfacing downstream as "all shards failed"), and a Phase 3 validation failure
+            // intentionally throws above (line ~530) only to be swallowed here. Fail loudly so
+            // the operator restores connectivity/config and restarts, rather than silently
+            // coming up broken. Legacy ES-only installs (migration not started) keep the
+            // historical best-effort behaviour and still boot.
+            if (isMigrationStarted() || isReadEnabled() || isMigrationComplete()) {
+                throw new DotRuntimeException(
+                        "Index initialization failed during OpenSearch migration; aborting startup"
+                        + " to avoid serving an unfinished index store. Cause: " + e.getMessage(), e);
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


ihoffmann-dot (approved):

dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImpl.java
```diff
@@ -554,6 +554,18 @@ public synchronized void checkAndInitializeIndex() {
...
             Logger.fatal(this.getClass(), "Failed to create new indexes:" + e.getMessage(),e);
 
+            // During an active OpenSearch migration a half-initialised index store is worse
+            // than a hard stop: the read provider would serve an unfinished or missing index
+            // (surfacing downstream as "all shards failed"), and a Phase 3 validation failure
+            // intentionally throws above (line ~530) only to be swallowed here. Fail loudly so
+            // the operator restores connectivity/config and restarts, rather than silently
+            // coming up broken. Legacy ES-only installs (migration not started) keep the
+            // historical best-effort behaviour and still boot.
+            if (isMigrationStarted() || isReadEnabled() || isMigrationComplete()) {
```
nollymar: It seems like the `isMigrationStarted()` is enough here phase != PHASE_0_MIGRATION_NOT_STARTED

fabrizzio-dotCMS: Good catch — `isReadEnabled()` (phases 2/3) and `isMigrationComplete()` (phase 3) are both subsets of `isMigrationStarted()` (phase != PHASE_0), so the OR was redundant. Collapsed both occurrences in `checkAndInitializeIndex()` to the single predicate in e081b171d7.

nollymar (approved):

Issue #35578, opened 2026-05-05
[TASK] Content Drive: Folder Upload Behavior tab — defaultBaseType preference + consolidate upload settings


zJaaal:
### Description

Implements the folder upload-behavior UX from epic #35436: let users control whether a folder uploads as **Assets** or **Files**, and manage that preference — together with the other upload-related folder settings — from a dedicated **Upload Behavior** tab in the folder settings dialog. Backed by the `defaultBaseType` folder field from #35577.

> **Design change (supersedes the original prototype):** the "remember this choice" checkbox is **gone**. Choosing the base type on upload is now a lightweight **menu** with no persistence; the *only* place a folder's upload preference is set/changed is the new **Upload Behavior** tab in folder settings.

#### 1. Upload menu — no persistence, just a pointer to settings

When the user starts an upload on a folder with no preference (Upload button or OS drag-and-drop), the existing Asset / File selector menu (`DotContentDriveDialogUploadSelectorComponent`, shipped in #36243) stays a **transient menu**:

- **Asset** (marked *Recommended*, default) and **File** options — unchanged.
- **No** "Always upload to this folder as {Assets|Files}" checkbox. (This is the part that changed — the checkbox from the original prototype is dropped.)
- Add a footer line: **"Set your default upload type in the Folder Settings."** — a pointer to where the preference is actually configured. (Final wording → copywriting.)

The menu does **not** call any PUT/persist the choice; it only decides the base type for the current upload.

<img width="414" height="290" alt="Image" src="https://github.com/user-attachments/assets/1926212d-3d42-4af0-bf15-9822313b7518" />

#### 2. Folder settings — new "Upload Behavior" tab

Add an **Upload Behavior** tab to `DotContentDriveDialogFolderComponent` (used for both Create and Edit modes), positioned **between GENERAL and ADVANCED**. It consolidates all upload-related folder configuration:

**a) Upload behavior** — radio group (title + description per option, same visual design as the upload menu), bound to the new `defaultBaseType` folder field:

- **Ask each time** — no preference (`defaultBaseType = null`). Uploads show the Asset/File menu.
- **Always use Assets** — `defaultBaseType = dotAsset`. Content type routed automatically by file type.
- **Always use Files** — `defaultBaseType = fileAsset`. Content type routed automatically by file type.

On Edit, pre-select from the folder's current `defaultBaseType`. Saving persists it via the folder POST/PUT endpoint.

[...]

Issue #35577, opened 2026-05-05
[TASK] Content Drive: Folder schema — add defaultBaseType field


zJaaal:
### Description

To support folder-level upload preferences described in epic #35436, the folder needs a single new field that records which base type Content Drive should use when uploading to that folder.

#### Field

- **`defaultBaseType`** — the default base type for uploads on the folder.
  - Values: `null`/absent (no preference — "ask each time"), `dotAsset`, or `fileAsset`.
  - Purely records the folder's upload-mode preference; it does not change existing FileAsset behavior.

> Note: this is intentionally scoped to a single field. The previously-discussed `defaultAssetType` field and any migration/deprecation of `defaultFileAssetType` are **not** part of this work. `defaultFileAssetType` stays exactly as it is today.

#### API

- Folder **POST** (create) and **PUT** (update) endpoints must accept `defaultBaseType`.
- Folder retrieval endpoints must expose `defaultBaseType`.

### Acceptance Criteria

- [ ] `defaultBaseType` added to the folder model (`null` / `dotAsset` / `fileAsset`).
- [ ] Folder create (POST) endpoint accepts `defaultBaseType`.
- [ ] Folder update (PUT) endpoint accepts `defaultBaseType`.
- [ ] Folder retrieval endpoints expose `defaultBaseType`.
- [ ] Omitting the field leaves the folder with no preference (null), and existing folders are unaffected.

#### Out of Scope

- `defaultAssetType` field.
- Any migration or deprecation of `defaultFileAssetType`.
- UI for editing this field — covered in companion ticket #35578.
- Inheritance of preferences across the folder hierarchy (tracked in epic #35436).

### Priority

Medium

### Additional Context

Parent: #35575
Epic: #35436
Companion (frontend): #35578

dario-daza:
## ❌ QA FAILED
When you have defined the folder Upload behavior as a `FileAsset/DotAsset` and you want to change it to ask every time (`defaultBaseType = null`) it doesn't change the `FileAsset/DotAsset`

https://github.com/user-attachments/assets/b5f594f4-4bda-4faf-af3a-1c0856b7592f

dario-daza:
## ✅ QA PASSED:

https://github.com/user-attachments/assets/b684b311-cbb9-4bd1-b79b-fb2d0102ca6d

Issue #35436, opened 2026-04-23
Content Drive: Upload (EPICID-194)


fmontes:
## Overview

Folder-level upload preferences that control whether uploads create dotAssets ("Assets") or FileAssets ("Files"). The upload button label communicates the folder's effective type — no separate indicator needed. Includes drag-and-drop, MIME-type routing for Assets mode, preference inheritance, and duplicate file handling.

---

## Problem

Content Drive is folder-and-file oriented, but uploads currently create dotAssets by default, which breaks the mental model for customers using FileAssets (~75% of existing customers). There is no compelling product reason to force one type over the other.

---

## Design Principle

The UI never surfaces "dotAsset" or "FileAsset." Only **"Assets"** and **"Files."**

---

## Folder States & Upload Behavior

| Folder state | Button label | Drag-and-drop | File creation |
|---|---|---|---|
| Explicit or inherited "Assets" | **Upload Assets** | Direct | Routed by MIME type |
| Explicit or inherited "Files" | **Upload Files** | Direct | All created as FileAssets |
| No preference (default) | **Upload** | Triggers choice dialog | Depends on user choice |

**No preference state:** Drag-and-drop triggers a choice dialog. If user has edit permission, a "Remember my preference" checkbox is shown.

---

## Preference Inheritance

Child folders inherit parent's preference unless explicitly overridden. Inherited folders behave identically to explicitly set folders.

---

## Duplicate File Handling

Replace, Keep both (appends suffix), or Cancel — following macOS / Google Drive pattern.

---

## Out of Scope (v1)

- Site-level or user-level upload preferences
- Predictable URL support for dotAssets
- Deprecating FileAssets

PR #36243, opened 2026-06-19
feat(content-drive): prompt Asset vs File on upload (#35575)

### LINKED ISSUE #35575: [TASK] Content Drive: Upload button — Asset vs File type selection
### Description

When uploading to a folder with **no preference set** (the default state described in epic #35436), prompt the user to choose between **Asset** and **File**, and send the corresponding base type to the upload endpoint instead of the current hardcoded `dotAsset`.

This applies to both upload entry points:
- Clicking the **Upload** button.
- **Drag-and-dropping** files from the OS into Content Drive.

[Prototype video](https://dotcms.slack.com/archives/C08MTB0RH6D/p1778000582312239)

#### Behavior

1. Triggering an upload (button click **or** OS drag-and-drop) prompts the user to choose between **Asset** (`dotAsset`) and **File** (`fileAsset`).

Image

2. For the button flow, after choosing, the OS file picker opens (existing flow).
3. For the drag-and-drop flow, after choosing, the dropped files are uploaded.
4. The rest of the upload flow stays as-is — the only change is sending the selected base type (`dotAsset` or `fileAsset`) to the upload endpoint instead of the current hardcoded `dotAsset`.

### Acceptance Criteria

- [ ] Clicking **Upload** in the default (no preference) state prompts the user to choose Asset or File.
- [ ] Drag-and-dropping files from the OS in the default (no preference) state prompts the user to choose Asset or File.
- [ ] Selecting **Asset** sends `dotAsset` to the upload endpoint.
- [ ] Selecting **File** sends `fileAsset` to the upload endpoint.
- [ ] The existing upload flow (file picker, drag-and-drop handling, success/error handling) is unchanged beyond the base type parameter.

#### Out of Scope — first iteration

Tracked in epic #35436:
- Folder-level upload preferences (explicit/inherited "Assets" or "Files" states).
- Dynamic button label (`Upload Assets` / `Upload Files` / `Upload`).
- MIME-type routing for Assets mode.
- Preference inheritance between folders.
- Duplicate file handling (Replace / Keep both / Cancel).
- Multiple files upload.

Tracked in follow-up sub-issues (out of scope for the first iteration of this ticket):
- #35577 — folder schema additions: `defaultBaseType`, `defaultAssetType`, migrate `defaultFileAssetType`, and expose via folder POST/PUT/GET endpoints.
- #35578 — "Remember my preference" checkbox + new folder fields in the Create/Edit Folder dialogs.

### Priority

Medium

### Additional Context

- Current upload code hardcodes `dotAsset`. The change is making the base type a parameter driven by user selection.

Parent epic: #35436

zJaaal:
## Proposed Changes

Closes #35575.

When uploading to a folder with no preference set, Content Drive now prompts the user to choose between **Asset** (`dotAsset`) and **File** (`FileAsset`) before the upload runs, and sends the selected base type to the upload endpoint instead of the hardcoded `dotAsset`. This applies to both upload entry points: the **Upload** button and **OS drag-and-drop**.

### Behavior
- **Upload button**: choose a type → the OS file picker opens → the file uploads as the chosen type.
- **Drag-and-drop**: drop files → choose a type → the dropped files upload as the chosen type.
- The selector defaults to **Assets** (marked *Recommended*); **Files** maps to `FileAsset`.
- Backend resolves the type from the contentlet `contentType` field; `FileAsset` is the File Asset content-type variable (matches `DEFAULT_FILE_ASSET_TYPES`).

https://github.com/user-attachments/assets/6136a0c8-afe6-4e89-a8de-4c736b3bef42

### Changes
- New `dot-content-drive-dialog-upload-selector` dialog, wired through the shell's `UPLOAD_SELECTOR` dialog. It emits the full selection (target folder + content type + files) so a future per-folder "remember" preference (#35578) can reuse it without rework.
- `DotUploadFileService.uploadDotAsset` takes an optional `contentType` (defaults to `dotAsset`; backward compatible with existing callers).
- Renamed `onAddNewDotAsset` / `addNewDotAsset` → `onUpload` / `upload` (no more hardcoded `dotAsset` in handler/event names).
- Fixed two pre-existing layout bugs noticed during review: dropzone overlay icon centering, and the locale chip wrapping at narrow widths.

### Out of scope (tracked in epic #35436)
Folder-level preferences/inheritance (#35577), the "Remember for this folder" checkbox (#35578), dynamic button label, MIME-type routing, and multi-file upload.

## Testing
- New spec for the upload selector dialog (option rendering, Recommended chip, default selection, emitted selection for Asset/File, Cancel).
- Shell spec reworked to the dialog flow (driven via UI events): opening the selector from button/drag-drop/sidebar, drag-drop upload as `dotAsset`/`FileAsset`, root upload, message handling, multi-file warning, and the button flow (pick → picker → upload).
- Upload service spec covers the default and explicit `contentType`.
- `portlets-content-drive`: 666 passing; `data-access`: 690 passing. Lint clean.

> Manual UI verification of the live upload + i18n copy is pending a backend message-cache reload.

This PR fixes: #35575


zJaaal:
Thanks for the review — went through each item.

**Addressed (e097bfaa68):**
- 🟡 `models.ts` — `DotContentDriveUploadSelection.contentType` is now a union (`DotContentDriveUploadContentType`) derived from `UPLOAD_SELECTOR_OPTIONS`, so it can't accept arbitrary strings and stays in sync with the rendered choices. This also covers the "validate the `contentType` param" point — the only producer is the typed selector.
- 🟡 `shell.component.ts` — renamed the shadowed `contentType` in the upload success handler (`contentType: uploadedContentType`) for clarity.

**Intentional / no change (with rationale):**
- 🟡 `hostFolder ?? ''` — empty string is the established "site root" convention here (same as `DotUploadFileService.publishContent`); the backend resolves it to the current host.
- 🟡 `$canContinue` — kept as a defensive guard so `onContinue` and the disabled state stay correct if the default selection ever changes; harmless given a radio can't be deselected today.
- 🟡 `onFileChange` reset ordering — the reset and `input.value = ''` are adjacent synchronous statements (no `await`/async between them), so a `finally` block isn't needed; resetting before the guard is deliberate so a re-opened picker can't reuse a stale selection.
- 🟡 duplicated `!files?.length` checks — each guard serves a distinct purpose: `resolveFilesUpload` needs it before reading `files.length`, and `uploadFile` needs it to narrow `files` before `files[0]`. They're type-narrowing guards, not redundant validation.
- 🟡 hardcoded `dotAsset`/`FileAsset` in `UPLOAD_SELECTOR_OPTIONS` — these are dotCMS's canonical base content-type variables (matches `DEFAULT_FILE_ASSET_TYPES`); they're the single source the union type now derives from.
- 🟠 `createFileList` test helper — test-only; it replicates the `length`/indexed-access/`item()` surface the implementation actually uses. Fine for these specs.
- 🟡 `jest` not imported in the dialog spec — `jest`/`afterEach` are ambient globals here, consistent with the sibling `dot-content-drive-dialog-content-type-selector.component.spec.ts`. Tests pass.

_— Reply drafted by Claude (Claude Code) on behalf of @zJaaal._

core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.ts
```diff
@@ -147,15 +151,37 @@ export class DotContentDriveShellComponent implements OnInit {
...
      */
-    readonly $dialogContentClass = computed(() =>
-        this.$activeDialog()?.type === DIALOG_TYPE.CONTENT_TYPE_SELECTOR
-            ? 'w-[min(92vw,38rem)] px-0! pt-0 pb-4'
-            : 'w-[43.75rem] pt-0 p-4'
-    );
+    readonly $dialogContentClass = computed(() => {
+        switch (this.$activeDialog()?.type) {
+            case DIALOG_TYPE.CONTENT_TYPE_SELECTOR:
+                return 'w-[min(92vw,38rem)] px-0! pt-0 pb-4';
```
adrianjm-dotCMS: Couldn't we use tailwind values for this?  like `w-152` for example

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-dropzone/dot-content-drive-dropzone.component.html
```diff
@@ -4,9 +4,12 @@
     class="message pointer-events-none visible absolute inset-0 z-1001 flex size-full items-center justify-center opacity-0 transition-all duration-300 ease-in-out"
     data-testid="message">
     <div class="flex flex-col items-center justify-center gap-3" data-testid="message-content">
-        <i
-            class="pi pi-upload h-[calc(48px+1.5rem)] w-[calc(48px+1.5rem)] rounded-full bg-primary-200 text-4xl text-primary-500"
-            data-testid="message-content-icon"></i>
+        <div
+            class="flex h-[calc(48px+1.5rem)] w-[calc(48px+1.5rem)] items-center justify-center rounded-full bg-primary-200">
+            <i
+                class="pi pi-upload text-[2.25rem]! leading-none text-primary-500"
```
adrianjm-dotCMS: Same question here, could we use like `text-4xl` for this case?

adrianjm-dotCMS (approved):

core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.ts
zJaaal: Good call — switched to the spacing scale in c9363abb8b: `w-152` / `w-125` / `w-175` (the repo already uses tokens like `w-100`/`w-160`, and Tailwind v4 generates these dynamically). Kept `max-w-[92vw]` so the dialog still caps on small screens — that viewport bound has no scale token, so it stays arbitrary.

_— Reply by Claude (Claude Code) on behalf of @zJaaal._

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-dropzone/dot-content-drive-dropzone.component.html
zJaaal: Done in c9363abb8b — now `text-4xl!`. The catch is the `!`: PrimeNG's `.pi` sets its own `font-size`, so a plain `text-4xl` gets overridden and the glyph renders tiny. The important flag is what makes the scale token win.

_— Reply by Claude (Claude Code) on behalf of @zJaaal._

zJaaal:
Re: the second automated pass — most points repeat the first review (already addressed above). One new claim worth clarifying:

- 🟡 **"removing `baseType` breaks FileAsset uploads"** — it doesn't. The backend resolves the type from `contentType` first; `baseType` is only consulted when `contentType` is absent (`ContentHelper.checkOrSetContentType`). Since we always send `contentType` (`dotAsset` or `FileAsset`), the old `baseType: 'dotAsset'` was dead. The binary auto-maps to each type's binary field, so the same multipart works for both.
- 🟡 **`uploadedContentType` could render "undefined"** — that's the *success* toast, and it reads the created contentlet's `contentType` from the response (same as before this PR); the NEW workflow action always returns it. Pre-existing behavior, not introduced here.

The remaining items (`hostFolder ?? ''` root convention, `$canContinue` guard, `onFileChange` reset ordering, file-input cleanup) are unchanged from my earlier reply.

_— Reply by Claude (Claude Code) on behalf of @zJaaal._

core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.ts
zJaaal: Done!

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-dropzone/dot-content-drive-dropzone.component.html
zJaaal: Done!

zJaaal:
**Design note — why we send `contentType`, not `baseType`:**

The upload dialog asks for a *base type* (Asset / File) semantically, so sending the base type would be the cleaner model. But it only works for `DOTASSET` today: `BaseTypeToContentTypeStrategyResolver` registers a strategy only for `DOTASSET`, and `MapToContentletPopulator.getContentTypeInode` has no `baseType` fallback — so firing `NEW` with `baseType: FILEASSET` (no `contentType`) fails with *"The content type or base type, is not set or is invalid."*

So this PR sends the **content type** (`dotAsset` / `FileAsset`), which works for both. Switching to base type for both requires a backend `FILEASSET` resolution strategy (and fixing the `subscribe()` bug that hardcodes `DOTASSET`) — tracked in #36279. Once that lands we can flip the frontend to base type.

_— Note by Claude (Claude Code) on behalf of @zJaaal._

PR #36264, opened 2026-06-22
Expose page render source metadata for a11y agent

### LINKED ISSUE #36265: Add backend render-source metadata for a11y agent
## Description

Add the backend/API support needed by the a11y agent to identify editable render sources for a page. This includes exposing theme files through `/_render-sources`, adding file extension metadata, supporting opt-in Sass source maps, and ensuring theme asset rendering respects the requested page mode.

Tracked PR: https://github.com/dotCMS/core/pull/36264

## Todo

- [x] Expand `/_render-sources` to return theme files beyond VTL-only refs.
- [x] Replace `VtlFileRefView` with generic `FileRefView`.
- [x] Add extension metadata to render-source file refs.
- [x] Add opt-in inline Sass source maps with `?sourcemap=true`.
- [x] Keep normal stylesheet responses unchanged when source maps are not requested.
- [x] Make theme CSS/template asset rendering respect the requested page mode.
- [x] Update OpenAPI metadata and language entries.
- [x] Add integration test coverage for render-source and source-map behavior.

## Acceptance Criteria

- [ ] `/_render-sources` returns theme file metadata the a11y agent can filter by extension.
- [ ] Compiled Sass requests can opt into inline source maps with `?sourcemap=true` without changing normal stylesheet responses.
- [ ] Theme CSS/template assets respect the requested page mode for backend-user scans.
- [ ] OpenAPI and integration test coverage are updated.

## Priority

Medium

## Additional Context

This is the backend slice of the broader a11y-agent work, separated for easier review and QA.

fmontes:
## Summary

This is the backend slice of the a11y-agent work. It gives the agent enough render-source metadata to understand which theme/template files participate in a page render, and enough stylesheet source information to trace compiled CSS back to editable source files.

## What changed

- Expanded `GET /api/v1/page/_render-sources` so theme sources are returned as a generic `files[]` list instead of only VTL-oriented refs. Each file now includes its path, identifier, and lowercased extension, so consumers can decide which files are relevant.
- Replaced the VTL-specific file ref view with a generic `FileRefView`, since the same shape is now used for VTL, CSS, SCSS/SASS/LESS, JS, and widget refs.
- Added opt-in inline Sass source maps using `?sourcemap=true` for compiled stylesheet requests. This keeps normal stylesheet responses unchanged, while allowing the agent to map CSS rules back to editable SCSS sources when needed.
- Fixed preview/live mode handling for theme stylesheet and template file requests so backend users scanning a page in `LIVE`, `PREVIEW`, or edit-related modes get assets that match the requested page mode instead of accidentally mixing in working content.
  -  Before: when you do `?mode=LIVE` it leaks `vtl` and or `css` files in working state. 
- Updated OpenAPI metadata, language entries, and integration tests around render-source and source-map behavior.

## Why

The a11y agent needs to propose or apply fixes against the files editors can actually change. For contrast fixes and template attribution, looking only at rendered HTML/CSS is not enough: the agent needs to know which theme files exist, which ones are editable stylesheet/template sources, and how compiled CSS maps back to source SCSS.

This PR keeps that backend/API support separate from the frontend and agent runtime changes so it is easier to review and QA independently.

## Verification

- `git diff --check` before commit
- Not run locally: full `dotcms-integration` suite; it uses the heavier Maven/Failsafe integration setup.

This PR fixes: #36265


fmontes:
## Review triage — addressed in `8e647ff`

Went through the DeepSeek code review and the Claude rollback analysis. Summary of what was fixed vs. intentionally left.

### ✅ Fixed

- **`CSSPreProcessServlet.java:389` — `Content-Disposition` filename** (DeepSeek 🟡 Medium). `fileUri.substring(fileUri.lastIndexOf('/'))` throws `StringIndexOutOfBoundsException` if the URI has no slash. Changed to `lastIndexOf('/') + 1`, which returns the bare file name and safely falls back to the full URI when there's no slash. (In practice CSS asset URIs are always rooted, so this was latent — but cheap to harden.)

### ⏭️ Not changed — and why

**DeepSeek 🟠 High / 🔴 Critical — all self-resolved in the same comment, verified against HEAD:**
- `:129` (Critical) old `DotLibSassCompiler` ctor — confirmed it delegates `this(host, uri, live, false, req)`. No issue.
- `:80` `sourceMap` null check — `equalsIgnoreCase` is null-safe; no functional bug.
- `:172` compiler source-map config — ctor passes the `sourceMap` flag; correct at the call site.
- `DotParse.java:173` flag cleanup without try-finally — `dontShowThemeTemplateIcon` is a one-shot flag; the exception window is negligible and the flag is re-set per render. Not worth a try-finally here.
- `SpeedyAssetServlet.java:114` Referer mode spoofing — `getWithReferer` only honors the referer when `user.isBackendUser()`, so a spoofed Referer from an anon user can't expose working content. Secure as written.

**Claude rollback — H-8 `templatePathIsFileAsset` VTL key (🟠 High):** the generated fragment is a guarded `#if($dotTheme.templatePathIsFileAsset) #dotParse #else #parse #end`. On rollback to N-1 the key resolves falsy and takes the `#else #parse` branch — which is exactly N-1's original behavior — so the generated VTL degrades safely. `PageLoader` also regenerates the velocity files itself, so N-1 rewrites the fragment regardless. No customer theme consumes the new key. Leaving as-is.

**Claude rollback — M-3 `vtls` → `files` REST shape (🟡 Medium):** this is the intentional, documented breaking change in the PR description (generic `FileRefView` replacing the VTL-specific view). Deliberate, not a regression.

**Remaining DeepSeek 🟡 Medium items** (389 rethrow, 1254/1460 file-ref handling, PageMode 147/164, FileRefView:18 lowercasing, CompilerOptions:156): cosmetic or behaviorally correct as-is; not blocking.

fmontes:
## Copilot review — both findings addressed in `ba5abdf`

Both inline comments were valid and are now fixed.

### ✅ `CSSPreProcessServlet.java` — source-map exposure (the important one)
Correct catch. `?sourcemap=true` enables `--embed-sources`, which inlines the raw SCSS — including `@import`ed partials that are never served on their own — into the response. As written, any caller who could fetch the compiled CSS (including anonymous users for a published stylesheet) could append the flag and pull back editable theme source.

Fixed by gating the source-map path behind **EDIT permission** on the asset; otherwise it returns `403`. `userHasEditPerms` was only computed for `!live` requests, so for `LIVE` requests (the a11y agent scans pages in `?mode=LIVE`) edit permission is now resolved explicitly before embedding sources.

### ✅ `SpeedyAssetServlet.java` — Referer in debug log
Fair point. Dropped the raw `Referer` header from the `Logger.debug` line. The resolved `PageMode` already captures what the log needs for diagnostics, and the explicit `?mode=` param is kept; the header (which can carry sensitive query-string data) is no longer logged.

Both verified with `./mvnw compile -pl :dotcms-core`. No OpenAPI/annotation changes, so `openapi.yaml` is unaffected.

dotCMS/src/main/java/com/dotcms/csspreproc/CSSPreProcessServlet.java
```diff
@@ -157,8 +169,16 @@ public void doGet(final HttpServletRequest req, final HttpServletResponse resp)
                 }
             }
             
+            // Source map requests are served fresh and never touch the shared CSS cache: caching is keyed by
+            // (site, uri, live) only, so storing the heavier map-embedded output would leak it to normal requests
+            // (and vice versa). This keeps the default path byte-identical while letting callers opt into the map.
+            if (sourceMap) {
+                serveWithSourceMap(req, resp, compiler, currentSite, fileUri, actualUri, userHasEditPerms);
+                return;
+            }
```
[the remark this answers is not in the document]
fmontes: Fixed in ba5abdf. Gated the source-map path behind EDIT permission on the asset — returns 403 otherwise. Since `userHasEditPerms` was only computed for `!live` requests, edit permission is now resolved explicitly for LIVE requests too (the a11y agent scans pages in `?mode=LIVE`), so `--embed-sources` output is never served to callers who can only read the compiled CSS.

dotCMS/src/main/java/com/dotmarketing/servlets/SpeedyAssetServlet.java
```diff
@@ -106,8 +106,14 @@ protected void service(HttpServletRequest request, HttpServletResponse response)
...
-        PageMode mode = PageMode.get(request);
-        Logger.debug(this, "SpeedyAssetServlet_service PageMode: " + mode);
+        // File assets (theme stylesheets, images, etc.) are fetched by the browser as standalone requests
+        // that carry no ?mode= of their own and do not update the session mode, so PageMode.get() alone
+        // would serve the working version to a backend user even when the page was requested with
+        // ?mode=LIVE. getWithReferer() honors the mode declared on the page's Referer in that case;
+        // anonymous users remain forced to LIVE.
+        PageMode mode = PageMode.getWithReferer(request);
+        Logger.debug(this, () -> "SpeedyAssetServlet_service PageMode: " + mode + " (referer="
+                + request.getHeader("Referer") + ", ?mode=" + request.getParameter("mode") + ")");
```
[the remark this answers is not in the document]
fmontes: Fixed in ba5abdf. Dropped the raw `Referer` header from the debug log; it now logs only the resolved `PageMode` and the explicit `?mode=` param, so sensitive query-string data in the referer is no longer written to logs.

dotcms-integration/src/test/java/com/dotcms/csspreproc/CSSPreProcessServletIT.java
```diff
@@ -129,6 +129,53 @@ public void testCompileInvalidScssFile() {
         executeTest(inputInvalidScssFile, INVALID_SCSS_FILE_RESPONSE, true);
```
fabrizzio-dotCMS, line 129: The convention for Integration-Tests is CSSPreProcessServletIT.java‎
The test must be added to a Test Suite

dotCMS/src/main/java/com/dotcms/rendering/velocity/viewtools/DotTemplateTool.java
```diff
@@ -362,18 +362,27 @@ private static Map<String, Object> setThemeData ( Folder themeFolder, String hos
 
         //Getting the template.vtl file path
         String themeTemplatePath;
+        boolean themeTemplateIsFileAsset;
         if ( UtilMethods.isSet( themeTemplate ) && InodeUtils.isSet( themeTemplate.getInode() ) ) {
-            themeTemplatePath = themeTemplate.getFileAsset().getPath();
+            // Logical, host-relative file-asset path (NOT the physical working-inode disk path) so the
```
fabrizzio-dotCMS: This change most likely needs doc and also an IT

dotCMS/src/main/java/com/dotcms/rest/api/v1/page/FileRefView.java
```diff
@@ -0,0 +1,38 @@
+package com.dotcms.rest.api.v1.page;
+
+import io.swagger.v3.oas.annotations.media.Schema;
+
+/**
+ * Reference to a file asset — path, identifier and extension only, no content.
+ * The {@code extension} lets consumers filter by type (e.g. vtl, css, scss, sass, js)
+ * without the API having to whitelist which extensions a theme may contain.
+ */
+public class FileRefView {
```
fabrizzio-dotCMS: Prefer records now that we are in Java 25

dotCMS/src/main/java/com/dotcms/rest/api/v1/page/PageResourceHelper.java
```diff
@@ -1251,24 +1251,26 @@ private ThemeSourceView buildThemeView(final Template template, final Host host,
...
-                    .filter(f -> UtilMethods.isSet(f.getFileName())
-                            && f.getFileName().endsWith(Constants.VELOCITY_FILE_EXTENSION))
-                    .map(f -> new VtlFileRefView(
-                            buildHostQualifiedPath(f.getPath() + f.getFileName(), host),
-                            f.getIdentifier()))
+                    .filter(f -> UtilMethods.isSet(f.getFileName()))
+                    .map(f -> toFileRef(f, host))
                     .collect(Collectors.toList());
 
-            return new ThemeSourceView(theme.getIdentifier(), theme.getName(), folderPath, vtls);
```
fabrizzio-dotCMS, line 1265: A test would be nice here

dotCMS/src/main/java/com/dotmarketing/util/PageMode.java
```diff
@@ -103,6 +103,60 @@ public static PageMode get(final String modeStr) {
...
+     * backend user, the {@code mode} declared on the {@code Referer} (the page URL that requested this
+     * asset) is honored.
+     * <p>
+     * The backend-user gate preserves the security boundary: an anonymous request can never escape the
+     * forced-LIVE behavior via a spoofed {@code Referer}, because it never reaches the fallback.
+     *
+     * @param request the current sub-resource request
+     * @return the resolved {@link PageMode}
+     */
+    public static PageMode getWithReferer(final HttpServletRequest request) {
```
fabrizzio-dotCMS, line 124: Add cobertura to PageModeTest

dotCMS/src/main/java/com/dotcms/rendering/velocity/viewtools/DotTemplateTool.java
```diff
@@ -362,18 +362,27 @@ private static Map<String, Object> setThemeData ( Folder themeFolder, String hos
...
+            // page render can include it with #dotParse and resolve live vs working from the PageMode.
+            themeTemplatePath = themePath + Template.THEME_TEMPLATE; // themePath already ends with "/"
```
fabrizzio-dotCMS: **Cache-key scoping risk with the new host-relative `templatePath`** 🟠

Going from `themeTemplate.getFileAsset().getPath()` (an absolute, host-independent path to the working binary on disk) to `themePath + THEME_TEMPLATE` is correct for version-aware live/working resolution via `#dotParse`, and it improves invalidation. But it interacts badly with the `themeMap` cache key.

The cache (`DotTemplateTool.java:55` — 5s `expireAfterWrite`, no event invalidation) is keyed by **theme folder identifier only** (`"themeMap" + themeFolder.getIdentifier()`, line 336), while `themePath` is **host-relative** whenever `themeFolder.getHostId().equals(hostId)`. And `#dotParse` resolves a relative path against `params.currentHost` (`DotParse.resolveFileAsset:295`) — the *rendering page's* host, not the theme's.

So for a theme shared across hosts (the `//host` branch right above exists precisely because that's supported):
- A page on the theme's own host renders first → relative `templatePath` gets cached.
- Within the 5s window, a page on a *different* host whose template points at this same theme reads the cached relative path → `#dotParse` resolves `/application/themes/<x>/template.vtl` against the **wrong** host → `ResourceNotFound`, or worse, a different theme that happens to sit at that path.

The old absolute disk path didn't have this failure mode for the include. Two clean fixes:
1. Always store the host-qualified (`//<themeHost>/...`) path for file-asset themes so it stays host-independent like before (the theme host is already resolved in the `else` branch), **or**
2. Add the rendering host to the cache key (`"themeMap" + folderId + hostId`).

I'd lean toward (1) — it keeps `#dotParse` resolution unambiguous regardless of which host renders.

dotCMS/src/main/java/com/dotcms/rendering/velocity/services/PageLoader.java
```diff
@@ -187,8 +187,18 @@ public InputStream buildStream(IHTMLPage htmlPage, PageMode mode, final String f
...
-            sb.append("#parse(\"$dotTheme.templatePath\")");
+            // Merging our template. File-asset themes are included with #dotParse so the theme
+            // template.vtl is resolved version-aware (live vs working) from the page's PageMode;
+            // the bundled static fallback templates are plain disk files and stay on #parse.
+            // dontShowThemeTemplateIcon is a one-shot flag consumed by #dotParse so the theme shell
+            // itself is not wrapped in an EDIT_MODE edit-control icon; nested #dotParse includes
+            // (html_head/header/footer) inside the theme still emit their icons.
+            sb.append("#if($dotTheme.templatePathIsFileAsset)")
+                .append("#set($dontShowThemeTemplateIcon = true)")
+                .append("#dotParse(\"$dotTheme.templatePath\")")
```
fabrizzio-dotCMS, line 198: **High-blast-radius behavior change with no direct render test** 🟠

This swaps the theme-shell include from `#parse` (filesystem, always the working binary) to `#dotParse` (version-aware) for *every* page that uses a file-asset theme. Net effect: a LIVE render now uses the **published** `template.vtl` instead of the latest working copy — that's the intended fix, but it's a semantic change on the hottest render path and nothing in the PR exercises it end-to-end.

Worth an integration test that renders a file-asset-theme page:
- **EDIT mode**: assert no stray `data-dot-object='vtl-file'` control is emitted for the shell (the one-shot `dontShowThemeTemplateIcon` suppression), while nested `#dotParse` includes (`html_head`/`header`/`footer`) still emit theirs.
- **LIVE vs WORKING**: publish `template.vtl`, then edit the working copy, and assert LIVE renders the published version while PREVIEW/EDIT renders the working one.

Static-fallback themes correctly stay on `#parse` — good. See also the cross-host `themeMap` cache-key concern on `DotTemplateTool` line 369, which this include now depends on.

fabrizzio-dotCMS (commented):
Two notes on the theme-template `#parse` → `#dotParse` switch. The change is good and actually *improves* content invalidation (it moves `template.vtl` onto the same version-aware file-asset velocity path that `header.vtl`/`footer.vtl`/`html_head.vtl` already use, and removes the old up-to-5s stale-binary pin). But it's a sensitive, high-blast-radius path, so flagging one concrete cross-host cache-key risk and one test gap inline.

dotcms-integration/src/test/java/com/dotcms/csspreproc/CSSPreProcessServletIT.java
fmontes: Done in d24909c — renamed to `CSSPreProcessServletIT` and updated the `MainSuite2b` registration to match.

dotCMS/src/main/java/com/dotcms/rendering/velocity/viewtools/DotTemplateTool.java
fmontes: Doc added in d24909c (`templatePathIsFileAsset` + the host-qualified template path on `setThemeData`). The end-to-end render IT is tracked in #36269.

dotCMS/src/main/java/com/dotcms/rest/api/v1/page/FileRefView.java
fmontes: Done in 000a59b — `FileRefView` is now a record. Regenerated `openapi.yaml` is byte-identical, so the REST contract is unchanged.

dotCMS/src/main/java/com/dotcms/rest/api/v1/page/PageResourceHelper.java
fmontes: Tracked in #36269 (test for the `toFileRef`/widget file-ref path).

dotCMS/src/main/java/com/dotmarketing/util/PageMode.java
fmontes: Done in d24909c — added `getWithReferer` coverage to `PageModeTest`: backend honors referer mode, anon/front-end ignore it (forced LIVE), explicit `?mode=` wins, no-referer falls back to PREVIEW.

dotCMS/src/main/java/com/dotcms/rendering/velocity/viewtools/DotTemplateTool.java
fmontes: Good catch — fixed in 7f27ee8. The file-asset `templatePath` is now always host-qualified (`//<themeHost>/...`, your option 1), so `#dotParse` resolves it against the theme's own host regardless of the rendering host. `$dotTheme.path` stays host-relative so existing consumers are unaffected.

dotCMS/src/main/java/com/dotcms/rendering/velocity/services/PageLoader.java
fmontes: Tracked in #36269 — the full publish/edit/render IT (EDIT-icon suppression + LIVE-vs-WORKING + cross-host) is too heavy for this PR. The cache-key concern it depends on is fixed here in 7f27ee8.

fabrizzio-dotCMS (approved):

hassandotcms:
QA Report | PR #36264

  Tested (all via live instance):
  - ✅ /_render-sources returns files[] with path/identifier/lowercased extension for all file types (vtl, css, scss, js) ; old code returned vtl-only
  - ✅ render-sources edge cases: invalid mode→400, no-auth→401, unknown page→404, LIVE/EDIT→200
  - ✅ ?sourcemap=true returns inline sourceMappingURL with embedded SCSS sources, correct headers (no-store, Content Disposition)
  - ✅ Default stylesheet response unchanged + cache isolated from sourcemap path
  - ✅ Sourcemap EDIT-permission gate returns clean 403 for logged-in non-editor (no source leak)
  - ✅ Mode-aware serving — backend user ?mode=LIVE→live, ?mode=EDIT→working, for both DOTSASS stylesheets and SpeedyAssetServlet file assets
  - ✅ Anonymous cannot escalate mode via spoofed Referer (forced LIVE) — both servlets
  - ✅ Theme template.vtl included version-aware via #dotParse; shell not edit-icon-wrapped, nested includes still emit their own icons
  - ✅ Never-published template.vtl still renders in LIVE (working fallback)
  - ✅ Cross-host theme renders correctly on a page on another host
  - ✅ FILE-backed widget resolved in widgets[] with correct path + identifier
  - ➖ LESS source maps — out of scope (SASS-only by implementation)

  Issues found:
  - Anonymous GET /DOTSASS/*.scss?sourcemap=true → NPE at CSSPreProcessServlet.java:185 (user.getUserId() on null user before the intended 403) → empty 200 + ERROR log
  per request. Not a leak; fix: null-guard the log / send 403 first. Add a deny-path test.
  - For a cross-host theme, render-sources theme.folderPath / files[].path are qualified to the page's host instead of the theme's host → those paths 404. identifier is correct (host-independent), so low impact. Optional one-line consistency fix to match this PR's DotTemplateTool behavior.
  
@fmontes

PR #36268, opened 2026-06-22
test(opensearch): phase-aware test bootstrap to run the indexing flow under a migration phase (#36266)

### LINKED ISSUE #36266: Add migration-phase / store control to ContentletDataGen for ES→OS dual-write tests
## Description

Tests cannot currently control **which search store** (ElasticSearch, OpenSearch, or both) a contentlet is indexed into when created through `ContentletDataGen`. This blocks straightforward verification of the ES→OS dual-write pipeline from the data-generation layer.

### Current state (confirmed in code)

- **Indexing seam:** `ContentletDataGen.persist()` → `checkin(contentlet, policy)` (`dotcms-integration/src/test/java/com/dotcms/datagen/ContentletDataGen.java:227`).
- `ContentletDataGen.setPolicy(IndexPolicy)` controls *when* indexing happens (`FORCE` / `WAIT_FOR` / `NONE`) — but **not which store** receives the write.
- The migration phase is **global only**: tests flip `Config.setProperty("FEATURE_FLAG_OPEN_SEARCH_PHASE", "1")` by hand.
- Routing lives in `PhaseRouter` (`ContentletIndexAPIImpl.java:173`); phase → write-providers is `0:[ES]  1:[ES,OS]  2:[ES,OS]  3:[OS]`.
- The existing migration ITs (`ContentletIndexAPIImplMigrationIntegrationTest`, `ContentletIndexAPIImplPhaseSwitchIntegrationTest`) create indices **directly** via `esImpl().createIndex()` / `osIndexAPI.createIndex()` — **none drive `ContentletDataGen` inside a phase context**. That is the missing capability.

### Proposed approach

Add a phase override to `ContentletDataGen`, applied **per-checkin** around the `checkin()` call in `persist()`, restoring the prior flag value in a `finally`:

```java
private MigrationPhase overridePhase = null;          // phase-oriented (primary)

public ContentletDataGen migrationPhase(final MigrationPhase phase) {
    this.overridePhase = phase;
    return this;
}

// store-oriented convenience, implemented on top of the phase override
public enum TargetStore { ES, OS, BOTH }
public ContentletDataGen targetStore(final TargetStore store) {
    this.overridePhase = switch (store) {
        case ES   -> MigrationPhase.PHASE_0_MIGRATION_NOT_STARTED; // ES only
        case OS   -> MigrationPhase.PHASE_3_OPENSEARCH_ONLY;       // OS only
        case BOTH -> MigrationPhase.PHASE_1_DUAL_WRITE_ES_READS;   // dual write
    };
    return this;
}

@Override
public Contentlet persist(final Contentlet contentlet) {
    final String prior = Config.getStringProperty(MigrationPhase.FLAG_KEY, null); // save prior
    try {
        if (overridePhase != null) {
            Config.setProperty(MigrationPhase.FLAG_KEY, String.valueOf(overridePhase.ordinal()));
        }
        return checkin(contentlet, null != policy ? policy : IndexPolicy.FORCE);
    } finally {
        if (overridePhase != null) {
            Config.setProperty(MigrationPhase.FLAG_KEY, prior); // restore prior, NOT null
        }
    }
}
```

Usage:

```java
new ContentletDataGen(contentType)
    .targetStore(TargetStore.BOTH)   // or .migrationPhase(PHASE_1_DUAL_WRITE_ES_READS)
    .setPolicy(IndexPolicy.WAIT_FOR)
    .nextPersisted();                // → written to both ES and OS
```

## Acceptance Criteria

- [ ] `ContentletDataGen` exposes a **phase-oriented** builder method (`migrationPhase(MigrationPhase)`) that overrides the active migration phase only for the duration of the persist/checkin.
- [ ] `ContentletDataGen` exposes a **store-oriented** convenience method (`targetStore(ES | OS | BOTH)`) implemented on top of the phase override (ES→Phase 0, OS→Phase 3, BOTH→dual-write Phase 1).
- [ ] Override scope is **per-checkin**: the flag is set immediately before the `checkin()` call in `persist()` and restored afterward — sequential `nextPersisted()` calls on different gens do not leak phase state.
- [ ] The override **restores the previous flag value** (captured before the call), never blindly clears it, so a test that set a phase globally is not clobbered.
- [ ] The flag is restored **even when `checkin()` throws** (restore lives in a `finally`).
- [ ] **Backward compatible:** with no override set, the gen respects the globally-active phase exactly as today — no behavior change for existing tests.
- [ ] Override composes with `setPolicy(...)` (e.g. `.migrationPhase(...).setPolicy(WAIT_FOR)`), and `WAIT_FOR` still allows immediate assertion of index contents.
- [ ] Publish/archive/delete/unpublish variants (`nextPersistedAndPublish`, etc.) honor the same phase override.
- [ ] A new **integration test** lives in `dotcms-integration`, ends in the literal `IT` suffix, and is registered in `OpenSearchUpgradeSuite`. It asserts:
  - [ ] `BOTH` (dual-write) → contentlet present in **both** ES and OS indices.
  - [ ] `OS` (Phase 3) → present in OS, **absent** from ES.
  - [ ] `ES` (Phase 0) → present in ES, **absent** from OS.

## Out of Scope

- `IndexAPI<F>` generic parameterization (tracked separately, must not be bundled here).
- Any change to production routing / `PhaseRouter` behavior — this is test-infrastructure only.

## Additional Context

- Architecture reference: `docs/backend/OPENSEARCH_MIGRATION.md`.
- `MigrationPhase` enum + `FLAG_KEY`: `dotCMS/src/main/java/com/dotcms/content/index/IndexConfigHelper.java`.
- All migration tests must be isolated in `OpenSearchUpgradeSuite` (project convention).

**Priority:** Medium

---

### Progress

- Branch: `issue-36266-contentletdatagen-phase-control`
- ✅ `ContentletDataGen` capability implemented (`migrationPhase(...)` + `targetStore(...)`, per-checkin scope, restore-prior + restore-on-throw, applied to both `persist(...)` overloads and `nextPersistedAndPublish()`) — compiles (`BUILD SUCCESS`).
- ⏳ Store-placement IT in `OpenSearchUpgradeSuite` — pending; requires a live two-store (`opensearch-upgrade`) environment to author/verify and works around the known content-lifecycle bootstrap gap (`inferIndexToHit` on un-bootstrapped `VersionedIndices`).

fabrizzio-dotCMS:
## Proposed Changes

Closes #36266.

Integration tests don't go through the production startup path
(`InitServlet.init() → ContentletIndexAPI.checkAndInitializeIndex() → IndexStartupValidator`), so
an OpenSearch index set is never created — and an unreachable OS service is never detected — unless
a test bootstraps it explicitly. This blocked running the existing test flow under an ES→OS
migration phase and having content (and **everything else that gets indexed**, not just contentlets)
land in OpenSearch the way it lands in ElasticSearch.

This PR adds a small, **separate** phase-aware bootstrap helper, wires it into the test init, and
adds a Maven profile + `just` recipe so the **whole indexing flow** can run under a chosen migration
phase with **no per-data-gen instrumentation**. The migration phase is already global config that the
phase router reads on every index operation — setting it and bootstrapping the stores is all that's
needed.

### `MigrationPhaseStoreBootstrap` (new test util)
- **`ensureStoresForCurrentPhase()`** — production-style bootstrap: validates OS reachability and
  **fails hard** if unreachable, then creates/registers the index set via `checkAndInitializeIndex()`.
  Used by `init()` (gated to phases ≥ 1). Stricter than production on purpose — production is lenient
  in phases 1–2 (`haltMigration()` → ES fallback), but a migration test that silently degraded to ES
  would mask the very problem it exists to catch.
- **`bootstrapOpenSearchStoreDirect()`** — creates + registers a fresh OS content index set directly
  (mirroring the router's OS bootstrap, with correct physical/`.os` naming), **bypassing the startup
  validator**. The validator's endpoint-separation check rejects ES==OS by design, which fails in the
  single-cluster `opensearch-upgrade` profile; this lets a test bootstrap OS there.
- **`openForPhase(phase)` → `OpenSearchPhaseStore` (AutoCloseable)** — one call that asserts OS
  reachable, captures + restores the prior phase and `VersionedIndices`, bootstraps OS, switches the
  phase, and exposes the placement helpers (`osHasDoc` / `awaitDoc` / `isIndexed` / `workingIndex`).
  Tests run under a phase with no per-test boilerplate and no injected OS client.
- **`assertOpenSearchReachable(phase)`** — standalone fail-fast reachability check (plain
  `client.info()`, not the full validator, to avoid the endpoint-separation check in single cluster).

### `IntegrationTestInitService.init()`
- Phase-gated bootstrap: when `FEATURE_FLAG_OPEN_SEARCH_PHASE` selects a phase ≥ 1, `init()` calls
  `ensureStoresForCurrentPhase()`. **Phase 0 (the default) is a no-op**, so existing single-store
  runs are unchanged.

### `opensearch-phase-suite` Maven profile + `just` recipe
- Activated by a single property `-Dopensearch.phase=N`; runs the MainSuite battery under phase `N`
  against **two separate clusters** (ES on `es.port`, OS 3.x on `os.upgrade.port`) so the
  endpoint-separation check passes and all phases can bootstrap. Injects
  `DOT_FEATURE_FLAG_OPEN_SEARCH_PHASE` into the failsafe fork.
- `just test-integration-phase [PHASE] [TEST]` wraps it.

### `MigrationPhaseStoreBootstrapIT` (new IT, registered in `OpenSearchUpgradeSuite`)
Opens an OS store under `PHASE_3_OPENSEARCH_ONLY` via `openForPhase(...)`, creates a contentlet with
the **ordinary** `ContentletDataGen` (no instrumentation), and asserts it (1) lands in the OS working
index — checked by document `_id` (`identifier_languageId_variantId`), noise-free, sidesteps the
`inferIndexToHit` gap — and (2) is retrievable via the high-level `ContentletAPI.search` path, which
reads OpenSearch under phase 3. Runs in the single-cluster profile; all bootstrap / teardown /
verification lives in `OpenSearchPhaseStore`.

## How to test

```bash
# the IT in its suite (single cluster)
./mvnw verify -pl :dotcms-integration \
  -Dcoreit.test.skip=false -Dopensearch.upgrade.test=true \
  -Dit.test=MigrationPhaseStoreBootstrapIT
```

```bash
# run other tests under a migration phase (two clusters, via the new profile)
./mvnw verify -pl :dotcms-integration -Dcoreit.test.skip=false -Dopensearch.phase=3
# or: just test-integration-phase 3 <TestClass>
```

## Notes

The OpenSearch-only (phase 3) path never writes through the legacy ES client, so it runs fine in the
single-cluster `opensearch-upgrade` profile. Dual-write phases (1/2) also exercise the ES leg and
therefore need two **separate** ES and OS clusters — the legacy ES `RestHighLevelClient` cannot parse
an OpenSearch 3.x content bulk-write response (`NullPointerException` in `DocWriteResponse`). The
`opensearch-phase-suite` profile keeps ES and OS on separate clusters precisely so all phases can run.

Follow-ups (separate PRs, already tracked): decouple `ContentletIndexAPIImplTest` from ES internals
(#36320) and the product gaps surfaced under phase 3 — optimize `.os` naming (#36323), sitesearch
index-settings parse (#36324), and `ES_TRACK_TOTAL_HITS` (#36325).

This PR fixes: #36266


fabrizzio-dotCMS:
Thanks for the automated reviews. These are AI (Bedrock) findings posted per-commit, and most point at intermediate commits that were later reverted/rewritten, so the line numbers are stale. Evaluated against the **latest commit**, none are actionable — details below.

### 🔴 `setPhase()` writes `phase.ordinal()` — "config must be enum names, not ordinals / corrupts config"
Not a bug — this is the **established convention**. The feature flag `FEATURE_FLAG_OPEN_SEARCH_PHASE` is an **integer ordinal by design**: `IndexConfigHelper.MigrationPhase.current()` reads it with `Config.getIntProperty(FLAG_KEY, 0)`, and the sibling suite `ContentletIndexAPIImplMigrationIntegrationTest.setPhase(int ordinal)` writes it exactly the same way (`Config.setProperty(FLAG_KEY, String.valueOf(ordinal))`). The value is also restored after each test in `OpenSearchPhaseStore.close()` (`@After`), and it's an in-memory flag in the test JVM — no cross-suite corruption.

### 🔴 `MigrationPhase.current()` may return `null` → reachability check silently skipped
`current()` **cannot return null**: it defaults via `Config.getIntProperty(FLAG_KEY, 0)` and falls back to `PHASE_0_MIGRATION_NOT_STARTED` for unset/invalid values. Additionally, `openForPhase(...)` is always called with an explicit, non-null phase (`PHASE_3_OPENSEARCH_ONLY`) and calls `assertOpenSearchReachable(phase)`, which **hard-fails** (throws) if OS is unreachable. There is no silent-skip path.

### 🟠 `esHasDoc()` / `osHasDoc()` swallow exceptions → "silent pass on a broken backend"
- `esHasDoc()` **no longer exists** (the line numbers reference a removed/rewritten file).
- `osHasDoc()` is used under `assertTrue(store.isIndexed(...))`: if OS is down, it returns `false` → the assertion **fails loudly**, it does not pass silently. And `openForPhase(...)` already asserts OS reachability up front (hard-fail). So a broken backend fails the test, not the opposite.

### 🟡 `ensureStoresForCurrentPhase()` not `@WrapInTransaction` → "config corruption if DB-backed + JVM crash"
The phase flag is an in-memory `Config` property set in test-init code (a test JVM), restored in `@After`. Not DB-backed, not a transactional concern.

### 🟡 `withPhaseOverride(...) returns Void` (deepseek, `ContentletDataGen.java`)
Stale — `ContentletDataGen` was **reverted to its baseline** in this PR; that code no longer exists.

### Minor (non-bug, optional follow-up)
`osHasDoc()` returns `false` on exception without a debug log. It doesn't cause false positives (the `assertTrue` fails regardless), but a `Logger.debug` on the swallowed exception would aid diagnosis — happy to add it if a reviewer wants.

Net: the PR is unchanged on this basis; happy to address any human-review feedback.

dotcms-integration/src/test/java/com/dotcms/util/MigrationPhaseStoreBootstrap.java
```diff
@@ -0,0 +1,307 @@
...
+        }
+
+        /** Refreshes the cluster, then polls until the doc is visible or the budget is exhausted. */
+        public boolean awaitDoc(final String docId) {
+            for (int i = 0; i < AWAIT_ATTEMPTS; i++) {
+                refresh();
+                if (osHasDoc(docId)) {
+                    return true;
+                }
+                Try.run(() -> Thread.sleep(AWAIT_SLEEP_MS));
```
ihoffmann-dot: Minor: Thread.sleep inside the polling loop is wrapped in Try.run(), which swallows InterruptedException. If the thread is interrupted, the loop runs all iterations before giving up. Suggestion:

```
try {
    Thread.sleep(AWAIT_SLEEP_MS);
} catch (InterruptedException e) {
    Thread.currentThread().interrupt();
    return false;
}
```

ihoffmann-dot (approved):

fabrizzio-dotCMS:
Thanks @ihoffmann-dot — addressed in d6f0801285.

`awaitDoc`'s inter-attempt sleep is now extracted into a `sleepBetweenAttempts()` helper that restores the interrupt flag and bails out instead of swallowing `InterruptedException`:

```java
private boolean sleepBetweenAttempts() {
    try {
        Thread.sleep(AWAIT_SLEEP_MS);
        return true;
    } catch (InterruptedException e) {
        Thread.currentThread().interrupt();
        return false;
    }
}
```

An interrupt now stops the poll loop immediately (`return false`) rather than running out all remaining attempts.

PR #36274, opened 2026-06-23
fix(push-publishing): remove fallback-rendered page artifacts on static un-publish (#35365)

### LINKED ISSUE #35365: Static publisher requires content to still be live, making local un-publish incompatible with in the same workflow
### Problem Statement

The native push/un-publish actionlets currently require opposite preconditions on the local content state depending on whether the target endpoint is static or dynamic. This makes it impossible to compose a single workflow that un-publishes the same content to both endpoint types.
Specifically:

Dynamic endpoints: PushNowActionlet with pushRemove=false produces a regular publish bundle. The remote un-publish only happens indirectly on the receiver inside ContentHandler when the payload indicates the sender content is no longer live (liveInode == null). This means the content must already be unpublished locally before the bundle is sent.
Static endpoints: StaticPublisher operates on the /live/... bundle filesystem, so it effectively requires the content to still be live when the bundle is built/executed.
The expected behaviour is that static publishing should also be allowed for unpublished content: when unpublished content is pushed to a static endpoint, the static publisher should remove the corresponding static content at the endpoint (rather than failing because /live/... no longer contains the asset).

https://drive.google.com/file/d/1tJACMryd2VXgoJJvGANofDw806LuC5qu/view?usp=sharing

### Steps to Reproduce

- Configure an authoring environment with at least one dynamic endpoint (e.g., a remote dotCMS delivery server) and one static endpoint (e.g., S3 ).

- Publish a piece of content so it is live on authoring and replicated to both endpoints.

- Create a workflow action containing:
- Un-publish Content
- Push/Remove Now → dynamic endpoint (pushRemove=false)
- Push/Remove Now → static endpoint (pushRemove=true)

- Execute the workflow against the live content.
- Observed: There is no ordering of these actionlets that yields a correct remote unpublish on both endpoints:
- If the local un-publish runs first, the static bundle cannot resolve the asset under /live/... and the static endpoint is not updated.
- If the static push runs first (while content is still live), the bundle sent to the dynamic endpoint carries liveInode != null and the receiver does not treat it as an un-publish.

### Acceptance Criteria

Static publishing should be a first-class participant in remote un-publish workflows:

- Static publishing should be allowed for unpublished content. The static publisher should not require the content to be live in /live/... in order to build/execute a bundle targeting a static endpoint.
- Pushing unpublished content to a static endpoint should remove the corresponding static artifact at the endpoint. The static publisher should interpret an unpublished-content payload as a deletion instruction and remove the rendered asset/page from the static target (e.g., delete the corresponding object from the S3 bucket).

- The preconditions for static and dynamic endpoints should be aligned, so that a single workflow can un-publish content across both endpoint types without requiring the author to orchestrate local publish/un-publish state transitions between actionlets

Expected: The workflow un-publishes the content from the dynamic endpoint, removes the rendered artifact from the static endpoint, and un-publishes locally — regardless of actionlet ordering.

### dotCMS Version

Current Release (dotEvergreen)

### Severity

### Links

https://helpdesk.dotcms.com/a/tickets/36287

dsolistorres:
## What

QA follow-up for #35365 (PR #36193). Static **Push Remove** of a page removed only the **default-language** artifact from the static endpoint (e.g. S3); other-language copies were left orphaned.

## Root cause — publish/un-publish language asymmetry

- **Publish:** `HTMLPageAsContentBundler` renders a file for **every configured bundle language** via `findByIdLanguageFallback(...)` (default-language fallback). So a page with a real version only in the default language is still published to S3 for *every* configured language (the others fall back to the default content). → *"Push Publish always creates a page on other languages."*
- **Un-publish (before this PR):** `StaticDependencyBundler` derived marker languages from `findContentletVersionInfos(identifier)` — **only languages with a real version**. The fallback-rendered copies have no real version, so they were never marked for removal. → *"Push Remove only removes the current language."*

This was never caught because every existing test page is built in a *fresh, non-default* language, so default-language fallback never triggers — no test created a page in the default language.

## Fix

`StaticDependencyBundler` now mirrors the page bundler: for **HTML pages**, it marks every configured language in which the page **resolves** (directly or via default-language fallback), using the same `findByIdLanguageFallback` call. Removal is therefore symmetric with publish, and fallback-rendered artifacts are removed.

**Scope:** page-specific. File assets and URL-mapped content are published per-language by *real version* (no fallback), so they keep the per-version-language behavior unchanged.

## Testing

- New integration case `getWorkingPageInDefaultLangConfiguredForFallbackLang` reproduces the QA scenario (page real-version only in default language, bundle configured for an extra language). It **fails before the fix** (`expected:<2> but was:<1>`) and **passes after**.
- Full `StaticPublisherIntegrationTest`: **76/76 green** (verified locally against PostgreSQL + OpenSearch). Unit tests green.

## Related

Still tracked separately: an enhancement to let static push-remove of *working (non-live)* content remove only specific language versions rather than all (pending product direction).

Refs: #35365

This PR fixes: #35365

### DIFF
--- a/dotCMS/src/enterprise/java/com/dotcms/enterprise/publishing/staticpublishing/StaticDependencyBundler.java
+++ b/dotCMS/src/enterprise/java/com/dotcms/enterprise/publishing/staticpublishing/StaticDependencyBundler.java
@@ -23,6 +23,7 @@ import com.dotmarketing.beans.Identifier;
 import com.dotmarketing.business.APILocator;
 import com.dotmarketing.business.IdentifierAPI;
 import com.dotmarketing.business.UserAPI;
+import com.dotmarketing.exception.DoesNotExistException;
 import com.dotmarketing.exception.DotDataException;
 import com.dotmarketing.exception.DotSecurityException;
 import com.dotmarketing.portlets.contentlet.business.ContentletAPI;
@@ -31,8 +32,10 @@ import com.dotmarketing.portlets.contentlet.model.Contentlet;
 import com.dotmarketing.portlets.contentlet.model.ContentletVersionInfo;
 import com.dotmarketing.portlets.folders.business.FolderAPI;
 import com.dotmarketing.portlets.folders.model.Folder;
+import com.dotmarketing.portlets.htmlpageasset.model.IHTMLPage;
 import com.dotmarketing.portlets.languagesmanager.business.LanguageAPI;
 import com.dotmarketing.portlets.languagesmanager.model.Language;
+import com.dotmarketing.util.Logger;
 import com.dotmarketing.util.UUIDUtil;
 import com.dotmarketing.util.UtilMethods;
 import com.google.common.collect.Lists;
@@ -180,14 +183,19 @@ public class StaticDependencyBundler implements IBundler {
                     }
 
                     if (staticUnpublish && contentPath != null && h != null) {
-                        // A marker is only needed for languages where the content has NO live
-                        // version: live content is rendered by the content bundlers and removed by
-                        // the publisher's delete branch, so it needs no marker. The publish queue
-                        // does not track the asset language (PublisherAPIImpl hardcodes
-                        // language_id=1), so derive the languages from the content's own versions
-                        // rather than the bundle languages. See issue #35365.
+                        // Un-publish must remove every artifact Push Publish created. The publish
+                        // queue does not track the asset language (PublisherAPIImpl hardcodes
+                        // language_id=1), so derive the languages from the content itself:
+                        //  - HTML pages are published for every configured language they resolve in
+                        //    (directly or via default-language fallback), so mirror that — otherwise
+                        //    fallback-rendered artifacts in other languages are orphaned. See #35365.
+                        //  - Other content (file assets, URL-mapped) is published only where it has a
+                        //    real version, so the per-version languages are the right set.
+                        final Collection<String> markerLanguages = contentlet.isHTMLPage()
+                                ? pageMarkerLanguages(id, languages)
+                                : nonLiveLanguages(id.getId());
                         StaticUnpublishMarker.writeContentMarkers(config, output, h.getHostname(),
-                                nonLiveLanguages(id.getId()), contentPath);
+                                markerLanguages, contentPath);
                     }
                 }
             }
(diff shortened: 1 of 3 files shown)


dsolistorres:
Thanks — addressed in `9400c42f69`.

### 🟠 High — `NumberFormatException` parsing `languageId` — **fixed**
`Long.parseLong(languageId)` is now guarded; a non-numeric bundle language id is skipped (with a warning) instead of aborting the bundle.

### 🟡 Missing null-check for `identifier` — **fixed**
Added an early `identifier == null || configLanguages == null` guard returning an empty set (in practice `id` is already non-null at the call site — guarded by `if (id == null) continue;` upstream — but the defensive guard is cheap).

### 🟡 `systemUser` not defined in method scope — **not an issue (false positive)**
`systemUser` is an instance field of `StaticDependencyBundler`, set in `generate()` (`systemUser = userAPI.getSystemUser();`) before `buildPathIncludes()` runs, which is the only caller of `pageMarkerLanguages`. It compiles and resolves correctly — no change needed.

### 🟡 `findByIdLanguageFallback(live=false)` and fully-deleted content — **already guarded**
The fully-deleted case (no version in the index) is handled upstream: the contentlet search returns empty and we `continue` (`if (cons.isEmpty()) { continue; }`) before reaching `pageMarkerLanguages`. So this method only runs when a working version exists, and `findByIdLanguageFallback(false)` resolves it for at least the content's own language — matching the publish-time fallback structure. No change needed.

### 🟡 `DoesNotExistException` logged at debug — **intentional**
This branch is an expected, common occurrence: it fires for every configured language the page does **not** resolve in (e.g. when the bundle has more languages than the page is published to). Logging it at info would be noisy on every un-publish; debug is the appropriate level for a non-error, expected path.

The guards are behavior-preserving for valid numeric language ids, so the previously-verified `StaticPublisherIntegrationTest` (76/76) result is unaffected; CI will re-run it.

dsolistorres:
Quick note on the latest re-run of this review: the 🟠 **High `NumberFormatException`** from the previous round is resolved (no longer reported), and the `dotCMS Backend Review` came back **✅ no issues found**. The three items now shown are new 🟡 Medium notes — addressed in `7fab09d430`:

1. **Catching `DoesNotExistException` not declared by `findByIdLanguageFallback`** — by design. It's an *unchecked* (runtime) exception the impl throws when the page resolves in neither the requested language nor the default; catching it is valid, and the `if (page != null)` branch covers the alternative (null) return, so the method is robust either way. No change.
2. **Debug log inside a lambda** — intentional, and the note itself concludes it's fine (the lambda avoids string-concat overhead when debug is off). This branch fires for every configured language a page isn't published to, so debug is the right level. No change.
3. **Test language might equal the default** — `new LanguageDataGen().nextPersisted()` always inserts a new language with a fresh id, so it can't equal the default; still, I added an explicit `assertNotEquals(defaultLanguage.getId(), fallbackLanguage.getId())` to document the repro's premise and guard the scenario.

dsolistorres:
Addressed both items from the latest review (commit `cfd34c15b7`):

1. **`Logger.warn` deferred message** — `pageMarkerLanguages` now uses the `Supplier<String>` (lambda) overload `Logger.warn(StaticDependencyBundler.class, () -> "Skipping non-numeric bundle language id: " + languageId)`, so the message is only built when warn logging is enabled — consistent with the `Logger.debug` call just below it.

2. **Unused `throws WebAssetException`** — removed from `getWorkingPageInDefaultLangConfiguredForFallbackLang` in `StaticPublisherIntegrationTestHelper`. The method only calls `build()` (which does not declare `WebAssetException`), so the declaration was dead.

Core build and `dotcms-integration` test-compile both green.

gortiz-dotcms (approved):

dsilvam (approved):

PR #36275, opened 2026-06-23
36267 publishing api include scheduled bundles in apiv1publishing

### LINKED ISSUE #36267: Publishing API: include scheduled bundles in /api/v1/publishing
### Description

The `/api/v1/publishing` endpoint currently misrepresents the publishing-jobs resource. A publishing job has a real lifecycle:

```
SCHEDULED → BUNDLE_REQUESTED → BUNDLING → SENDING → ... → SUCCESS / FAILED_*
```

The endpoint surfaces every state EXCEPT `SCHEDULED`. Any client (UI, automation, 3rd-party integration) querying "list my publishing jobs" gets an incomplete picture.

**What is a scheduled bundle?**

A bundle that has been pushed with a future `publishDate` but has not yet been picked up by the publisher cron job. It lives in `publishing_queue` (with `publish_date > now()`) and in `publishing_bundle` (with `publish_date` set), but does NOT yet have a row in `publish_audit` — which is the only table `/api/v1/publishing` reads from today.

**Reproduce**

1. Push any asset with publishDate = NOW + 5 minutes
2. Within that window, `GET /api/v1/publishing` → bundle is absent
3. After the cron job runs, bundle appears with `status=BUNDLE_REQUESTED`

The legacy JSP `view_publish_queue_list.jsp` showed these bundles by querying `publishing_queue` directly via `PublisherAPI.getQueueBundleIds`. That method has no REST callers and was not migrated to the v1 surface.

**Related**

- Epic: #34734
- Spike (audit + migration mapping): #36039 — note that the spike incorrectly claimed `/v1/publishing?status=…` replaces the legacy queue query; this ticket closes the gap.

### Acceptance Criteria

- [ ] A new value `SCHEDULED` is added to `com.dotcms.publisher.business.PublishAuditStatus.Status`
- [ ] `GET /api/v1/publishing` includes bundles where `publishing_queue.publish_date IS NOT NULL` AND `publish_audit` has no row for that `bundle_id`
- [ ] These bundles are returned with `status = SCHEDULED` in the existing `PublishingJobView` shape:
    - `bundleName`, `filterName`, `filterKey` from `publishing_bundle`
    - `assetCount` from `publishing_queue`
    - `createDate` = `bundle.publishDate` (the scheduled execution time)
    - `statusUpdated` = `null`
    - `numTries` = `0`
    - `environmentCount` from `publishing_bundle_environment`
- [ ] `?status=SCHEDULED` filters to only this tier
- [ ] Mixed enum filter still works: `?status=SCHEDULED,SUCCESS` returns both
- [ ] Pagination works correctly across the unioned result set (total count is accurate)
- [ ] Search filter (`?filter=...`) applies to scheduled rows the same way it applies to audit rows (bundle name + bundle id, case-insensitive)
- [ ] OpenAPI annotations updated: `@Parameter` description for `status` lists the new value
- [ ] OpenAPI regenerated and committed (`./mvnw compile -pl :dotcms-core -DskipTests`)
- [ ] Auth/permissions unchanged — same backend-user gating as the existing endpoint
- [ ] Integration test added covering: scheduled bundle visible, status filter, pagination total, transition (cron picks up → status changes to BUNDLE_REQUESTED)

### Priority
Medium

### Additional Context

**Files likely affected:**

- `com/dotcms/publisher/business/PublishAuditStatus.java` (add `SCHEDULED` enum value)
- `com/dotcms/publisher/business/PublishAuditAPIImpl.java` (extend the SQL in `getPublishAuditStatus` / `countPublishAuditStatus` around line 735+)
- `com/dotcms/rest/api/v1/publishing/PublishingResource.java` (update `@Parameter` description for `status`, line 199-214)
- `com/dotcms/rest/api/v1/publishing/PublishingJobsHelper.java` (`parseStatuses` validation around line 179, `getValidStatusNames` around line 213, `toPublishingJobView` may need to handle synthesized rows around line 152)
- `dotCMS/src/main/webapp/WEB-INF/openapi/openapi.yaml` (regenerated)

**Suggested implementation approach** (not prescribed):

Extend `PublishAuditAPIImpl.getPublishAuditStatus(...)` so its SQL builder (currently `buildCombinedFilterQuery` around line 735) UNIONs synthesized rows from `publishing_queue` joined with `publishing_bundle`, filtered by:

```sql
publishing_queue.bundle_id NOT IN (SELECT bundle_id FROM publishing_queue_audit)
```

The synthesized rows need to populate a `PublishAuditStatus` instance with `status = SCHEDULED` and the metadata above so the existing `PublishingJobsHelper.toPublishingJobView()` works without changes.

**Domain question for the implementer:**

Should `SCHEDULED` only mean "future publish_date AND no audit row," or should it ALSO catch the brief race window where `publish_date` has passed but the cron hasn't yet inserted the audit row? Recommend: the criterion is "queue row exists with no audit row" regardless of date — this avoids the race.

**Out of scope:**

- Drafts (`publishing_bundle.publish_date IS NULL`) — separate resource concept (bundle definition, not job). Track in a separate ticket if needed.
- Frontend changes — tracked separately; will land after this.

**Verification path:**

Reproduce with a real push-publish setup (e.g., the `docker/docker-compose-examples/push-publish/docker-compose.yml` two-node setup). A scheduled bundle should appear in `/api/v1/publishing` immediately after Push Publish, with `status=SCHEDULED`, then transition to `BUNDLE_REQUESTED` once the cron job picks it up.

hassandotcms:
###  Problem

  /api/v1/publishing only reads publishing_queue_audit. Bundles pushed with a future publish date have no audit row until the cron picks them up, so they were invisible to the API (the legacy JSP Queue tab showed them by reading publishing_queue directly).

###  Change

  Surface those bundles through the existing endpoints as a new SCHEDULED status — no new endpoint.

  - PublishAuditStatus.Status: add SCHEDULED (negative sentinel code, never persisted).
  - PublisherAPI: add countScheduledBundleIds / getScheduledBundleIds — queue rows with publish_date IS NOT NULL and no audit row (excludes drafts); asset count from queue rows,
  env count via correlated subquery.
  - GET /api/v1/publishing: unions the scheduled tier with the audit tier as a contiguous block; pagination total and slicing stay exact. ?status=SCHEDULED filters to that tier;
  mixed filters work.
  - GET /api/v1/publishing/{id}: returns a synthesized SCHEDULED detail instead of 404.
  - DELETE /api/v1/publishing/purge?status=SCHEDULED: rejected (audit-based purge can't match it; points to per-bundle DELETE).
  - Views: add nullable scheduledPublishDate (future run time) so createDate keeps its meaning.
  - Regenerated openapi.yaml.

###  Tests

  Integration tests in PublishingResourceIntegrationTest using the real push path: visibility/status, status filter (single + mixed), pagination total, detail-not-404, delete-cancels, transition to WAITING_FOR_PUBLISHING once an audit row appears, and a legacy-getQueueBundleIds-vs-API parity check.

This PR fixes: #36267


dotCMS/src/main/java/com/dotcms/rest/api/v1/publishing/PublishingJobsHelper.java
```diff
@@ -258,16 +293,41 @@ private Bundle getBundleSafely(final String bundleId) {
...
-            return bundle.getFilterKey();
+            Logger.debug(this, "Unable to resolve filter: " + filterKey, e);
+            return filterKey;
+        }
+    }
+
+    /**
+     * Coerces a DotConnect result value to a Date (JDBC returns java.sql.Timestamp), or null.
+     */
+    private static Date asDate(final Object value) {
```
jcastro-dotcms: There's a `com.dotmarketing.util.DateUtil` class in core. Please move this method to that class of check if the method already exists there.

dotCMS/src/main/java/com/dotcms/rest/api/v1/publishing/PublishingJobsHelper.java
```diff
@@ -258,16 +293,41 @@ private Bundle getBundleSafely(final String bundleId) {
...
+        return value instanceof Date ? (Date) value : null;
+    }
+
+    /**
+     * Coerces a DotConnect result value (Number or numeric String) to an int, defaulting to 0.
+     */
+    private static int asInt(final Object value) {
```
jcastro-dotcms: There's a `com.dotmarketing.util.NumberUtil` class in core. Please move this method to that class of check if the method already exists there.

jcastro-dotcms (changes requested):

dotCMS/src/main/java/com/dotcms/rest/api/v1/publishing/PublishingJobsHelper.java
hassandotcms: done

dotCMS/src/main/java/com/dotcms/rest/api/v1/publishing/PublishingJobsHelper.java
hassandotcms: done

jcastro-dotcms (approved):

PR #36282, opened 2026-06-23
feat(sitesearch): vendor-neutral Site Search — neutral aggregation + OpenSearch impl + phase-aware router (#35786)

### LINKED ISSUE #35786: [TASK] SiteSearchAPI/SiteSearchWebAPI — vendor-neutral aggregation abstraction
### Description

`SiteSearchAPI` (public interface) returns `org.elasticsearch.search.aggregations.Aggregation` and throws `org.elasticsearch.ElasticsearchException`. `SiteSearchWebAPI` (Velocity ViewTool) casts directly to concrete ES aggregation implementation classes (`InternalDateHistogram`, `StringTerms`, `StringTerms.Bucket`). This blocks Site Search from being served by OpenSearch in Phase 3.

**Priority:** Blocking Phase 3 — Site Search  
**Dependency:** `OSSiteSearchAPI` implementation (#34609)

### Scope

1. Define vendor-neutral DTO `IndexAggregation` (and sub-types for date histogram / terms) under `com.dotcms.content.index.domain`
2. Update **`SiteSearchAPI`** — replace `Aggregation` return type with `IndexAggregation`; replace `ElasticsearchException` throws with `DotSearchException`
3. Update **`SiteSearchWebAPI`** — remove casts to `InternalDateHistogram`, `StringTerms`, `Bucket`; use `IndexAggregation` API
4. Update **`ESSiteSearchAPI`** — adapt ES aggregation results to `IndexAggregation` DTO (adapter pattern)
5. **`OSSiteSearchAPI`** (from #34609) — implement using OS aggregation types, return same `IndexAggregation` DTO

### Checklist

- [ ] Define `IndexAggregation` DTO and sub-types in `com.dotcms.content.index.domain`
- [ ] Remove `Aggregation`, `ElasticsearchException` from `SiteSearchAPI` public signatures
- [ ] Remove `InternalDateHistogram`, `StringTerms`, `Bucket` casts from `SiteSearchWebAPI`
- [ ] Adapt `ESSiteSearchAPI` to produce `IndexAggregation` from ES results
- [ ] `OSSiteSearchAPI` returns `IndexAggregation` from OS results

### Acceptance Criteria

- Zero `org.elasticsearch.*` imports in `SiteSearchAPI` and `SiteSearchWebAPI`
- Velocity templates using `$SiteSearch` receive the same aggregation data structure
- Both ES and OS providers are swappable without changing `SiteSearchWebAPI`
- Existing site search integration tests pass

fabrizzio-dotCMS:
## What & why

Closes the Site Search portion of the ES → OpenSearch migration (#35786). Site Search is decoupled from Elasticsearch types and given a working OpenSearch backend plus a phase-aware router, so it dual-writes and reads correctly across all migration phases.

Two commits:

1. **Vendor-neutral aggregation abstraction** — removes `org.elasticsearch.*` from the `SiteSearchAPI` contract and `SiteSearchWebAPI`, reusing the existing `com.dotcms.content.index.domain.Aggregation` / `AggregationBucket` DTOs (from #36026) with histogram support, and introduces `DotSearchException`.
2. **OpenSearch implementation + router** — adds `OSSiteSearchAPI`, the `SiteSearchAPIImpl` phase router, and an integration test.

## Changes

| Area | Change |
|---|---|
| `SiteSearchAPI` / `SiteSearchWebAPI` | No vendor imports; `getAggregations`/`getFacets` return neutral `Aggregation`; `DotSearchException` added |
| `OSSiteSearchAPI` (new) | `@ApplicationScoped @Default` OpenSearch impl. Search/aggregations via the generic client → `ContentSearchResponse` (mirrors `OSSearchAPIImpl`); doc put/delete via `_doc` PUT/DELETE; get via typed `client.get(...)`. Default index resolved from `VersionedIndicesAPI` (not the deprecated `IndiciesAPI`) |
| `SiteSearchAPIImpl` (new, router) | `PhaseRouter<SiteSearchAPI>` mirroring `IndexAPIImpl`; the single fan-out point. Reads → read provider; doc/index writes → write fan-out; `listIndices`/`listClosedIndices` merge in dual-write; Quartz task methods route to a single provider (fan-out would double-schedule jobs) |
| `ESSiteSearchAPI` | Uses raw `ESIndexAPI` instead of the `IndexAPI` router so the SiteSearch router is the only fan-out point (avoids double dual-write of OS indices) |
| `APILocator` | `SITESEARCH_API` now returns `SiteSearchAPIImpl` |

## Design notes

- **Placement**: OS impl + router live next to `ESSiteSearchAPI` in the enterprise package (license-gated feature). The single `annotated` `beans.xml` covers the merged `target/classes`, so CDI still discovers the `@Default` bean.
- **OS index naming**: site-search indices use untagged logical names. `VersionedIndicesAPI` force-tags `.os` on store/load, so the default is `IndexTag.strip(...)`-ed on read. `deactivateIndex` calls `removeVersion(...)` when removing the slot would leave the version empty (`saveIndices` rejects empty).
- **Highlights**: the neutral `SearchHit` DTO carries no highlights, so OS `search()` returns empty highlight arrays (the ES path is best-effort too) — marked `TODO OS`.

## Testing

- `./mvnw compile -pl :dotcms-core` → BUILD SUCCESS (Java 25)
- `./mvnw test-compile -pl :dotcms-integration -am` → BUILD SUCCESS
- New `OSSiteSearchAPIIntegrationTest` (registered in `OpenSearchUpgradeSuite`) covers lifecycle, doc round-trip, aggregations, and default-index activation. Requires the `opensearch-upgrade` container:
  ```
  ./mvnw verify -pl :dotcms-integration -Dcoreit.test.skip=false -Dopensearch.upgrade.test=true -Dit.test=OSSiteSearchAPIIntegrationTest
  ```

This PR fixes: #35786


fabrizzio-dotCMS:
## Decision — neutral `Aggregation` vs ES `Aggregation`: equivalence scope & deliberate gaps

This records, as a reviewed decision, **how far** the vendor-neutral `com.dotcms.content.index.domain.Aggregation` / `AggregationBucket` are meant to be equivalent to the Elasticsearch types they replace on the VTL-facing surface (`$sitesearch.getAggregations(...)` / `getFacets(...)`), and which parts of the ES surface are **intentionally not** reproduced.

### What "equivalence" means here

The ES `org.elasticsearch.search.aggregations.Aggregation` interface is minimal — `getName()`, `getType()`, `getMetadata()`. Templates never reached buckets through that interface (it has no bucket accessors); Velocity duck-types against the **concrete runtime object** (`InternalTerms`, `InternalDateHistogram`, `TopHits`, `Terms.Bucket`…). So the contract we preserve is the **runtime accessor surface** a template could realistically have walked, not the 3-method interface.

### Covered (the common VTL path — what templates actually use)

| ES runtime accessor | Neutral equivalent |
|---|---|
| `agg.getName()` / `agg.getType()` | `getName()` / `getType()` |
| `(Terms/Histogram).getBuckets()` | `getBuckets()` |
| `bucket.getKeyAsString()` | `getKeyAsString()` |
| `bucket.getKeyAsNumber()` | `getKeyAsNumber()` (lenient parse; date-histogram key normalized to epoch-millis) |
| `bucket.getDocCount()` | `getDocCount()` |
| `bucket.getAggregations()` (nested) | `getAggregations()` |
| `topHits.getHits()…` (id / source / score) | `getHits()` → `SearchHits` / `SearchHit` |

This is verified end-to-end (`SiteSearchWebAPITest`, ES Phase 0), on the OpenSearch path (`OSSiteSearchAPIIntegrationTest`, `fromOS`), and at the factory level deterministically (`AggregationDomainTest`, incl. the date-histogram `ZonedDateTime`→epoch-millis conversion).

### Deliberate gaps (NOT reproduced — accepted)

These ES members are intentionally **not** exposed on the neutral types. They are analytics/lookup methods with no observed use in Site Search VTL, and reproducing them would re-introduce vendor coupling or carry no OpenSearch counterpart:

1. **`Aggregation.getMetadata()`** — the per-aggregation `meta` map. Not surfaced.
2. **`Terms.getBucketByKey(String)`** — random bucket lookup. Templates iterate buckets instead; a template doing `$agg.getBucketByKey("x")` would break.
3. **`Terms.getSumOfOtherDocCounts()` / `Terms.Bucket.getDocCountError()` / `Terms.getDocCountError()`** — approximate-count error/overflow accounting. Not surfaced.
4. **`bucket.getKey()` type narrowed `Object` → `String`** — for a date histogram, ES `getKey()` returned a `ZonedDateTime`; the neutral `getKey()` returns the epoch-millis **as a String**. A template calling date methods on the raw key (`$bucket.key.toInstant()`) would break. The supported idioms are `getKeyAsString()` / `getKeyAsNumber()`, which are preserved.

**Risk:** a customer/site VTL template relying on (2), (3), or the `Object` shape of (4) would fail after this change. We assess this as **low likelihood** — these are not part of the documented Site Search facet/aggregation idiom — and **accept it knowingly** rather than re-couple the neutral API to vendor types.

### Interaction with the H-8 rollback label

The deliberate gaps are a **forward-compatibility** consideration (a template adopting the neutral shape on N), distinct from the **rollback** (H-8) concern already flagged on this PR (a template adopting the *new* accessors `getBuckets()`/`getHits()`/`iterator()` and then being rolled back to N-1, whose `getAggregations()` returns the ES type). No template was co-migrated in this PR, so the H-8 risk window stays theoretical until one is. If/when a template is migrated to the neutral accessors, it must be tracked against the N+1 retirement of the old ES return type per H-8's two-phase guidance.

### If a gap proves needed

Add the specific accessor to the neutral type (e.g. a `metadata()` map, or a `getBucketByKey` helper on `Aggregation`) backed by both `from(...)` and `fromOS(...)` factories, with a unit test in `AggregationDomainTest` — rather than leaking the vendor type back onto the API.

fabrizzio-dotCMS:
### Update — `getMetadata()` equivalence gap is now CLOSED

Following the equivalence-gaps decision above, one of the four gaps has been implemented: **`getMetadata()`**.

**Why this one, and not the others.** It is the only gap that is **rollback-safe**: the ES `org.elasticsearch.search.aggregations.Aggregation` interface *already* exposes `getMetadata()`, so a template that adopts `$agg.metadata` resolves on **both** N (neutral `Aggregation`) **and** N-1 (ES `Aggregation`) after a rollback. Adding it therefore does **not** create a new H-8 vector — unlike `getBuckets()` / `getHits()`, which the ES interface lacks. It also completes equivalence with the exact interface the review anchored on: `getName` / `getType` / `getMetadata` (the third was the one missing).

**Implementation** (`Aggregation.java`):
- New 5th record component `Map<String, Object> getMetadata` (bean-named so `$agg.metadata` resolves via `getMetadata()`), with a null-safe empty-map default in the canonical constructor + a `Builder.metadata(...)`.
- ES factory: `.metadata(esAgg.getMetadata())` — already `Map<String, Object>`.
- OS factory: a `fromOSMeta(...)` helper unwraps each `JsonData.to(Object.class)` (falling back to the raw JSON string if it can't be mapped) from each variant's `meta()` (`AggregateBase.meta()`, inherited by the sterms/lterms/dterms/top_hits aggregates) — so the OpenSearch path is not left with a fresh divergence from ES.

**Tests:** `AggregationDomainTest` → 11/11 (added `esFactory_metadata_isPreserved` and `aggregation_metadata_defaultsToEmptyWhenUnset`). No existing test broke (unstubbed `getMetadata()` mocks return null → default empty map). `dotcms-core` install + `dotcms-integration` test-compile both BUILD SUCCESS (Java 25).

**Still deliberately deferred** (unchanged from the decision above): `getBucketByKey(String)`, `getSumOfOtherDocCounts()`, `getDocCountError()`, and the `getKey()` `Object`→`String` narrowing — no observed consumer, and adding accessors the ES interface lacks would widen the H-8 rollback surface for no benefit. They remain "add when a consumer is proven, backed by both factories + a unit test."

fabrizzio-dotCMS:
## Review feedback addressed — pushed in `ba8e9645cf`

Summary of the changes that resolve the open review findings on this PR, plus the aggregation test gaps.

### Findings resolved

| Sev | Finding | Resolution |
|---|---|---|
| 🔴 | `OSSiteSearchAPI.putToIndex` swallows write failures (silent data loss) | Now throws `DotSearchException` on non-2xx / caught exceptions. **It must propagate, not log-and-swallow:** `PhaseRouter` is already phase-aware — it re-throws the *primary* provider's failure and swallows the *shadow's* (WARN). Swallowing inside the impl defeated that, so a Phase-3 (OS-primary) write failure was lost. Now publishing observes it in Phase 3; OS shadow drift in phases 1/2 is logged WARN by the router. |
| 🔴 | `deleteFromIndex` swallows failures | Same fix; HTTP 404 stays benign (idempotent delete). |
| 🟠 | `setAlias` returns `false` on success | Returns `true` (`createAlias` is void + throws on failure → reaching the return = success). |
| 🟠 | Missing `indexName` validation → NPE / invalid chars reach OpenSearch | New `requireValidIndexName(idx)` guard (null/blank, non-lowercase, OS-forbidden chars) fails fast with `IllegalArgumentException` **before** the name is interpolated into the REST endpoint. |
| 🟠 | Mapping/settings read via `new File(url.getPath())` — NPEs if missing, fails inside a JAR | Read via `getResourceAsStream` (`readResource` helper, UTF-8, clear error if absent). JAR-safe. |
| 🟡 | Per-doc `Logger.info` in put/delete (log spam + URL at info) | Lowered to `Logger.debug(..., Supplier)` (no concatenation when off). Bulk delete log stays at info. |
| 🟡 | `setAlias` missing alias null/empty check | Already present (the `UtilMethods.isNotSet(indexName) || isNotSet(alias)` throw). |

**Deliberate scope note:** `ESSiteSearchAPI.putToIndex`/`deleteFromIndex` also swallow failures (legacy, pre-existing). Left unchanged — in Phase 3 OS is the sole writer so the OS fix covers the critical case, and making the legacy ES path throw would surface previously-hidden failures to existing Phase-0/1 deployments (riskier, out of scope here).

### Aggregation test coverage + equivalence

- **`AggregationDomainTest`** — added deterministic ES-factory unit tests (Mockito, no container), incl. the previously-untested `date_histogram` `ZonedDateTime`→epoch-millis key conversion. 11/11 pass.
- **`SiteSearchWebAPITest`** — added `date_histogram` cases (both `getAggregations` and the legacy `getFacets` wrapper).
- **`OSSiteSearchAPIIntegrationTest`** — strengthened the terms-agg assertion (bucket content/keys/counts, was only `containsKey`), added a nested `top_hits` case on the `fromOS` path, and index-name validation cases.
- **`getMetadata()` equivalence gap closed** — see the dedicated comment above; it's the one gap that's rollback-safe (the ES `Aggregation` interface exposes it too), so it completes interface equivalence without adding H-8 surface. `getBucketByKey`/`getSumOfOtherDocCounts`/`getDocCountError` remain deliberately deferred (no consumer; verified no shipped template uses `$sitesearch` aggregations).

### Verification

`./mvnw install -pl :dotcms-core` and `test-compile -pl :dotcms-integration` → BUILD SUCCESS (Java 25). `AggregationDomainTest` 11/11. The OpenSearch ITs require the `opensearch-upgrade` container to run.

fabrizzio-dotCMS:
### Note for reviewers — the latest `qwen` Bedrock review is re-reporting already-resolved findings

The most recent automated review (model `qwen.qwen3-next-80b-a3b`, against `86bf0fbb44`) lists several "New Issues 🔴/🟠" that are **already fixed**. It appears to be regurgitating the carried-forward *prior findings* block rather than re-reading the current code — the tell is that it cites **`OSSiteSearchAPI.java:1082` / `:1136` / `:1046`, but the file is only 910 lines long**, so those line numbers cannot exist.

Verified status of each, against the pushed code:

| Reported (stale) | Actual current state |
|---|---|
| `putToIndex` silently swallows write failures | **Fixed** — now `throw new DotSearchException(...)` on non-2xx / caught exceptions (`OSSiteSearchAPI.java` ~595/606); `PhaseRouter` applies the per-phase primary/shadow policy. |
| `deleteFromIndex` no index/idx validation, NPE risk | **Fixed** — `requireValidIndexName(idx)` guard before the REST call (~553/647); 404 stays benign by design. |
| `putToIndex` logs raw exception without context | **Fixed** — the thrown message includes `docId` + index. |
| `putToIndex` uses `ESMappingAPIImpl.toJsonString` on the OS path | **Intentional** — faithful ES-format doc JSON is the documented design for the OS site-search doc body, not a regression. |
| `SiteSearchAPIImpl:227` — `copyOf()` shallow-copy "concurrent writes corrupt state" | **False positive** — the fan-out is **sequential, not concurrent**; each provider gets an independent map copy; every `putToIndex` mutation is a top-level `map.put(...)` (entry replacement, not nested-object mutation), so a shallow copy is sufficient. This reasoning is already documented in the method's javadoc (`SiteSearchAPIImpl.java` ~240-245). |

A prior review pass (deepseek r1) had already marked these **Resolved**; the later qwen pass simply didn't re-diff. The substantive review feedback was addressed in `ba8e9645cf` (see the "Review feedback addressed" comment) and `86bf0fbb44`, and the rollback-safety check now labels the PR **Safe To Rollback**.

nollymar (approved):

PR #36284, opened 2026-06-23
feat(image): register libvips rotate filter and SHARED_COMPLETED dotGenerated mode  (#36283)

### LINKED ISSUE #36283: libvips: SHARED_COMPLETED dotGenerated mode + register rotate filter
## Problem Statement

The libvips image engine (`IMAGE_API_USE_LIBVIPS`) has two gaps when running in a clustered/pod deployment:

**1. No shared rendition reuse across instances.** `DOTGENERATED_DEFAULT_PATH` today supports only `LOCAL` (write renditions to instance-local `dotsecure`) or the default (write straight to the shared assets path). With `LOCAL`, every instance — and every restart — regenerates the same renditions. Writing directly to the shared path on the hot serving thread couples generation latency to (slow) shared storage and writes intermediate transforms there too.

**2. The `rotate` filter silently no-ops under libvips.** `VipsRotateImageFilter` exists but was never registered in `VipsImageFilterApiImpl.filterClasses`, so `?filter=rotate` resolves to nothing and the image is served unrotated (the legacy engine registers it correctly).

## Proposed Solution

**New `DOTGENERATED_DEFAULT_PATH=SHARED_COMPLETED` mode:**

- Before serving, check whether the final hashed rendition already exists in the **shared** `dotGenerated` store. If present, serve it directly — no regeneration.
- If absent, generate it. **All intermediate transforms stay instance-local under `dotsecure`** (fast scratch).
- Once the final rendition is ready locally, **return it immediately**, then a **background virtual thread** publishes the final file to the shared store under its hashed name. Staging goes through the shared temp dir (`getAssetTempPath()`, same mount as the store) and is atomically moved into place, so other instances — and restarts of this one — never regenerate it and never observe a partial file.

`LOCAL` and the legacy default behaviour are unchanged.

**Register `rotate` in the libvips engine** so `VipsRotateImageFilter` is actually used.

## Acceptance Criteria

- [ ] `DOTGENERATED_DEFAULT_PATH=SHARED_COMPLETED` generates renditions locally and serves finished renditions from the shared store when present.
- [ ] The serving request returns from the local copy; the shared publish happens on a background virtual thread.
- [ ] Shared publish is atomic (temp-on-shared-mount + move) and a no-op if already published.
- [ ] `LOCAL` and default modes behave exactly as before.
- [ ] `?filter=rotate` rotates the image under the libvips engine.
- [ ] Unit tests cover the path mapping and shared-store publish.

## Component / Area
Image API (libvips engine), `ImageFilterExporter`, `ConfigUtils`.

wezell:
## Proposed Changes

Closes #36283

### 1. Register `rotate` in the libvips engine
`VipsRotateImageFilter` existed but was missing from `VipsImageFilterApiImpl.filterClasses`, so `?filter=rotate` resolved to nothing and the image was served unrotated. Added `.put(ROTATE, VipsRotateImageFilter.class)`.

### 2 New `DOTGENERATED_DEFAULT_PATH=SHARED_COMPLETED` mode
For clustered/pod deployments. Renditions are generated on fast instance-local scratch and only the finished file is shared, so no other instance — or a restart of the same instance — regenerates it.

- **Serve from shared if present** — before generating, the final hashed rendition path is computed (`finalResultFile`, no pixel work) and checked in the shared store; a hit is served directly.
- **Generate locally** — all transforms (intermediate + final) write to `dotsecure`.
- **Publish in the background** — the local final is returned to the request immediately; a `Thread.ofVirtual()` then copies it to the shared store. Staging goes through `getAssetTempPath()` (shared `tmp_upload`, **same mount** as the store) and is atomically `Files.move`d into place, so readers never see a partial file. No-op if already published.

The legacy default behaviours are unchanged.

## Why
- `LOCAL` makes every instance/restart regenerate identical renditions.
- Writing straight to the shared assets path puts (slow) shared-storage latency on the serving thread and stores intermediates there too.
- `SHARED_COMPLETED` keeps generation local/fast and shares only the completed result.

## Testing
`ImageFilterExporterSharedStoreTest` (5 unit tests, container-free):
- `toSharedFile` rebases `localBase → sharedBase` preserving the `inode[0]/inode[1]/hashedName` layout; returns null outside the base; null-safe.
- `copyToShared` creates missing dirs, copies exact bytes, leaves no stray `.tmp` in the rendition dirs, and is idempotent on repeat publish.

```
Tests run: 5, Failures: 0, Errors: 0, Skipped: 0
```

Manual: set `DOT_DOTGENERATED_DEFAULT_PATH=SHARED_COMPLETED` + `DOT_IMAGE_API_USE_LIBVIPS=true`, request renditions, confirm finals appear under the shared `dotGenerated` and a second instance serves without regenerating; confirm `?filter=rotate&rotate_a=90` rotates.

## Notes
- Shared store resolves to the assets root (`getAbsoluteAssetsRootPath()/dotGenerated`) — the existing shared mount; no new volume wiring.
- The temp-on-shared-mount invariant is load-bearing for atomicity and is documented on `copyToShared`.

This PR fixes: #36283


freddyDOTCMS (approved):

dotCMS/src/main/java/com/dotmarketing/portlets/contentlet/business/exporter/ImageFilterExporter.java
```diff
@@ -123,22 +142,112 @@ private File runFilter(ImageFilter imageFilter, final File fileIn,final Map<Stri
...
+            final File parent = shared.getParentFile();
+            if (parent != null && !parent.exists()) {
+                parent.mkdirs();
+            }
+            if (!sharedTmpDir.exists()) {
+                sharedTmpDir.mkdirs();
+            }
+            final File tmp = new File(sharedTmpDir, shared.getName() + "_" + System.nanoTime() + ".tmp");
+            Files.copy(localFinal.toPath(), tmp.toPath(), StandardCopyOption.REPLACE_EXISTING);
+            Files.move(tmp.toPath(), shared.toPath(), StandardCopyOption.REPLACE_EXISTING);
```
oidacra: The JavaDoc here states the move is atomic so "concurrent readers never observe a partial file," but the call passes only `REPLACE_EXISTING`. `Files.move` is not guaranteed atomic without `StandardCopyOption.ATOMIC_MOVE` — on a POSIX same-mount setup it resolves to an atomic `rename(2)` in practice (so the documented deployment works), but the guarantee the JavaDoc makes isn't the one the code requests. Consider `Files.move(tmp, shared, StandardCopyOption.ATOMIC_MOVE)` and catch `AtomicMoveNotSupportedException` to fall back, so the atomicity invariant is explicit.

dotCMS/src/main/java/com/dotmarketing/portlets/contentlet/business/exporter/ImageFilterExporter.java
```diff
@@ -123,22 +142,112 @@ private File runFilter(ImageFilter imageFilter, final File fileIn,final Map<Stri
...
+    static void copyToShared(final File localFinal, final File shared, final File sharedTmpDir) {
+        try {
+            final File parent = shared.getParentFile();
+            if (parent != null && !parent.exists()) {
+                parent.mkdirs();
+            }
+            if (!sharedTmpDir.exists()) {
+                sharedTmpDir.mkdirs();
+            }
+            final File tmp = new File(sharedTmpDir, shared.getName() + "_" + System.nanoTime() + ".tmp");
```
oidacra: The staging temp name uses `shared.getName() + "_" + System.nanoTime()`. `nanoTime()` is per-JVM, so two cluster instances staging to the same shared `tmp_upload` mount can collide (low odds, but nonzero). Also, if `Files.move` throws, the staged `.tmp` is orphaned — the catch logs but never deletes it, so failures accumulate stray temp files on the shared store. Consider `Files.createTempFile(sharedTmpDir.toPath(), shared.getName()+"_", ".tmp")` (or add a UUID/node id) and `Files.deleteIfExists(tmp)` in a finally block.

dotCMS/src/test/java/com/dotmarketing/portlets/contentlet/business/exporter/ImageFilterExporterSharedStoreTest.java
```diff
@@ -0,0 +1,98 @@
...
+                .listFiles((d, name) -> name.contains(".tmp"));
+        assertTrue(strays == null || strays.length == 0, "stray .tmp file left in shared dir");
+    }
+
+    /**
+     * Publishing the same rendition twice (as two cluster instances would) leaves the shared file
+     * intact with the correct bytes — last write wins on identical content.
+     */
+    @Test
+    void copyToShared_idempotentOnRepeat(@TempDir final Path tmp) throws Exception {
```
oidacra: This "idempotent on repeat" test publishes the same payload twice, so it can't prove that `REPLACE_EXISTING` actually overwrites *different* stale content — which is the real concurrent-cluster / pre-restart case the design targets. Consider pre-creating `shared` with different bytes (or running with payload A then payload B from a second local file), then asserting the shared file holds the second payload.

dotCMS/src/main/java/com/dotmarketing/util/ConfigUtils.java
```diff
@@ -185,14 +185,37 @@ public static String getAssetTempPath() {
...
+     * dotsecure, see {@link #getDotGeneratedPath()}) and the finished file is published here so that
+     * other instances in the cluster — and the same instance after a restart — serve it directly
+     * instead of regenerating it.
+     */
+    public static String getDotGeneratedSharedPath() {
+        return ConfigUtils.getAbsoluteAssetsRootPath() + File.separator + "dotGenerated";
+    }
+
+    /** @return true when finished renditions should be published to the shared store after generation. */
+    public static boolean isDotGeneratedSharedCompleted() {
```
oidacra, line 206: `isDotGeneratedSharedCompleted()` (the feature gate) and the rewritten three-way `dotGeneratedPath` routing have no unit coverage, despite being the riskiest logic change in the PR. The gate reads `Config` live and is trivially testable; the `Lazy` routing is awkward to test because it memoizes. Consider a small test asserting case-insensitive `SHARED_COMPLETED`/`LOCAL` parsing and that an unrecognized value falls back to the assets-root path (the legacy behavior).

dotCMS/src/main/java/com/dotmarketing/portlets/contentlet/business/exporter/ImageFilterExporter.java
```diff
@@ -123,22 +142,112 @@ private File runFilter(ImageFilter imageFilter, final File fileIn,final Map<Stri
...
+     */
+    private void publishToShared(final File localFinal) {
+        final File shared = toSharedFile(localFinal);
+        if (shared == null || localFinal == null || !localFinal.exists()) {
+            return;
+        }
+        if (shared.exists() && shared.length() >= MIN_VALID_FILE_LENGTH) {
+            return;
+        }
+        Thread.ofVirtual().name("dotGenerated-share")
```
oidacra: The background publish spawns one virtual thread per uncached-rendition request, with no cap and no per-target de-duplication: concurrent requests for the same rendition all pass the `shared.exists()` guard and each copy the same bytes. This is benign for correctness once the move is made atomic, but a burst of cache misses fans out many simultaneous shared-mount copies with no backpressure (the existing semaphore guards pixel work, not publishing). Consider submitting publishes to a small bounded executor and/or a `ConcurrentHashMap.putIfAbsent` guard per target path to collapse duplicate in-flight publishes.

dotCMS/src/main/java/com/dotmarketing/portlets/contentlet/business/exporter/ImageFilterExporter.java
```diff
@@ -123,22 +142,112 @@ private File runFilter(ImageFilter imageFilter, final File fileIn,final Map<Stri
...
+     * loses atomicity. The caller passes {@link ConfigUtils#getAssetTempPath()} (the shared
+     * {@code tmp_upload} dir under the assets root), which sits on the same mount as the shared
+     * {@code dotGenerated} store. Package-visible and synchronous for tests; the virtual-thread
+     * hand-off lives in {@link #publishToShared(File)}.</p>
+     */
+    static void copyToShared(final File localFinal, final File shared, final File sharedTmpDir) {
+        try {
+            final File parent = shared.getParentFile();
+            if (parent != null && !parent.exists()) {
+                parent.mkdirs();
```
oidacra: `if (!parent.exists()) parent.mkdirs();` (and the same for the tmp dir) is a check-then-act that races under concurrent publishes, and the `mkdirs()` return value is ignored — a genuine failure surfaces later as an opaque copy/move IOException. Consider `Files.createDirectories(...)` for both, which is idempotent and concurrency-safe.

dotCMS/src/main/java/com/dotmarketing/portlets/contentlet/business/exporter/ImageFilterExporter.java
```diff
@@ -123,22 +142,112 @@ private File runFilter(ImageFilter imageFilter, final File fileIn,final Map<Stri
...
+            if (!sharedTmpDir.exists()) {
+                sharedTmpDir.mkdirs();
+            }
+            final File tmp = new File(sharedTmpDir, shared.getName() + "_" + System.nanoTime() + ".tmp");
+            Files.copy(localFinal.toPath(), tmp.toPath(), StandardCopyOption.REPLACE_EXISTING);
+            Files.move(tmp.toPath(), shared.toPath(), StandardCopyOption.REPLACE_EXISTING);
+            Logger.debug(ImageFilterExporter.class, "published rendition to shared store: " + shared);
+        } catch (Exception e) {
+            Logger.warnAndDebug(ImageFilterExporter.class,
+                    "failed to publish rendition to shared store " + shared + ": " + e.getMessage(), e);
```
oidacra, line 270: Swallowing the exception is correct for a fire-and-forget publish, and `warnAndDebug` does log at WARN. The gap is operational: a broken or read-only shared mount produces one WARN per request indefinitely, with no rate-limiting and no metric, so the real degradation (every instance regenerating, shared store never populated) is easy to lose in log noise. Consider keeping the swallow but emitting a metric/counter for publish failures, or rate-limiting the WARN (first occurrence + periodic summary), so a broken mount surfaces as a clear signal.

dotCMS/src/main/java/com/dotmarketing/portlets/contentlet/business/exporter/ImageFilterExporter.java
```diff
@@ -123,22 +142,112 @@ private File runFilter(ImageFilter imageFilter, final File fileIn,final Map<Stri
...
+     * Returns {@code null} when {@code localFinal} is not under {@code localBase}.
+     */
+    static File toSharedFile(final File localFinal, final String localBase, final String sharedBase) {
+        if (localFinal == null) {
+            return null;
+        }
+        try {
+            final String base = new File(localBase).getCanonicalPath();
+            final String full = localFinal.getCanonicalPath();
+            if (!full.startsWith(base)) {
```
oidacra: `full.startsWith(base)` on canonical paths is a raw string prefix match, so base `.../dotGenerated` would also match a path under a sibling like `.../dotGenerated2/...` and rebase it incorrectly. Low real-world risk given the fixed layout, but a latent bug. Consider comparing against `base + File.separator` (or using `Path.startsWith`), and add a negative test for the sibling-prefix case.

dotCMS/src/main/java/com/dotmarketing/portlets/contentlet/business/exporter/ImageFilterExporter.java
wezell: Fixed in bde9654e27: the move now attempts `ATOMIC_MOVE` and only falls back to a non-atomic `REPLACE_EXISTING` if the filesystem can't do it — so the 'no partial file' guarantee actually holds on the same-mount setup.

dotCMS/src/main/java/com/dotmarketing/portlets/contentlet/business/exporter/ImageFilterExporter.java
wezell: Fixed in bde9654e27: staging now uses `Files.createTempFile(tmpDir, name+"_", ".tmp")` (atomic unique name, no cross-JVM `nanoTime` collision) and the staged temp is deleted in a `finally` if the publish fails.

dotCMS/src/test/java/com/dotmarketing/portlets/contentlet/business/exporter/ImageFilterExporterSharedStoreTest.java
wezell: Fixed in bde9654e27: replaced with `copyToShared_overwritesStaleSharedContent` — pre-creates `shared` with different bytes/length and asserts they're replaced with the fresh payload, proving the REPLACE semantics for the real cross-instance/pre-restart case.

dotCMS/src/main/java/com/dotmarketing/util/ConfigUtils.java
wezell: Fixed in bde9654e27: added `isDotGeneratedSharedCompleted_readsConfigLive` covering the gate (sets the property and asserts it flips live). Left the memoized `Lazy` routing untested as you noted — it's awkward and lower-risk.

dotCMS/src/main/java/com/dotmarketing/portlets/contentlet/business/exporter/ImageFilterExporter.java
wezell: Addressed in bde9654e27: added an in-flight `ConcurrentHashMap`-backed set keyed by the shared path. Concurrent requests for the same rendition now spawn at most one publish thread; the rest no-op.

dotCMS/src/main/java/com/dotmarketing/portlets/contentlet/business/exporter/ImageFilterExporter.java
wezell: Fixed in bde9654e27: both dirs now use `Files.createDirectories(...)` (idempotent, race-free, throws loudly on real failure) instead of check-then-act `mkdirs()` with an ignored return.

dotCMS/src/main/java/com/dotmarketing/portlets/contentlet/business/exporter/ImageFilterExporter.java
wezell: Acknowledged — keeping the swallow (fire-and-forget). A broken/read-only mount would log one WARN per uncached request; with the in-flight dedupe (bde9654e27) that's now bounded per distinct rendition. A throttled WARN + a publish-failure metric is a reasonable operational follow-up, but out of scope for this PR — happy to file it.

dotCMS/src/main/java/com/dotmarketing/portlets/contentlet/business/exporter/ImageFilterExporter.java
wezell: Fixed in bde9654e27: `toSharedFile` now matches on a directory boundary (`base + File.separator`), so a sibling like `.../dotGenerated2/x` is no longer rebased against base `.../dotGenerated`.

oidacra:
QA: Rotation working now 
Image

PR #36317, opened 2026-06-25
feat(content-drive): add FILEASSET base-type → content-type resolution strategy (#36279)

### LINKED ISSUE #36279: [TASK] Content Drive: Add a FILEASSET base-type → content-type resolution strategy
### Description

Content Drive's upload flow (PR #36243, issue #35575) lets the user choose **Asset** vs **File** before uploading. Semantically these are **base types**, but the frontend currently sends the **content type** variable (`dotAsset` / `FileAsset`) because base-type → content-type resolution only works for `DOTASSET` on the backend.

Where it breaks for `FILEASSET`:
- `BaseTypeToContentTypeStrategyResolver` (`dotCMS/src/main/java/com/dotcms/contenttype/business/BaseTypeToContentTypeStrategyResolver.java`) registers a strategy **only** for `BaseContentType.DOTASSET` (`DotAssetBaseTypeToContentTypeStrategyImpl`, which uses `DotAssetAPI.tryMatch` to pick the best dotAsset content type by mime type). There is no `FILEASSET` strategy.
- `ContentHelper.tryToSetContentType` only sets the contentlet's content type when a strategy exists for the given base type; otherwise it is skipped.
- `MapToContentletPopulator.getContentTypeInode` resolves the type solely from `structureInode` / `structureName` / `contentType` — there is **no** fallback from `baseType`.
- Net result: firing the `NEW` workflow action with only `baseType: FILEASSET` (no `contentType`) fails with **"The content type or base type, is not set or is invalid."**

Latent bug to fix along the way:
- `BaseTypeToContentTypeStrategyResolver.subscribe(...)` ignores its `baseContentType` argument and always does `builder.put(BaseContentType.DOTASSET, strategy)`, so a `FILEASSET` strategy can't even be registered via `subscribe` today.

### Acceptance Criteria

- [ ] Register a `FILEASSET` base-type → content-type resolution strategy so that firing the `NEW` workflow action with `baseType: FILEASSET` (no explicit `contentType`) resolves to the appropriate File Asset content type (default/system `FileAsset`, or best match if multiple `FILEASSET` content types exist).
- [ ] Fix `BaseTypeToContentTypeStrategyResolver.subscribe(...)` so it registers the strategy under the passed `baseContentType` instead of always `DOTASSET`.
- [ ] Add coverage (unit/integration) for resolving `FILEASSET` (and a regression test for the `subscribe` fix).
- [ ] Once this lands, the Content Drive frontend (#35575) can send the **base type** for both Asset and File instead of the content type, matching what the upload dialog actually asks.

### Priority

Medium

### Additional Context

- Parent epic: #35436
- Follow-up from PR #36243 / issue #35575 (frontend ships sending content type as the working approach; this task enables the cleaner base-type approach).
- Reference for the working DOTASSET strategy: `DotAssetBaseTypeToContentTypeStrategyImpl` + `DotAssetAPIImpl.tryMatch`.

zJaaal:
## Proposed Changes

Resolves #36279.

Content Drive's upload dialog (PR #36243 / #35575) lets users choose **Asset** vs **File** — semantically these are **base types**. But the backend could only resolve a content type from a base type for `DOTASSET`, so firing the `NEW` workflow action with `baseType: FILEASSET` (no `contentType`) failed with *"The content type or base type, is not set or is invalid."* This adds the missing `FILEASSET` resolution strategy and switches the Content Drive frontend to send the base type for both options.

### Backend
- **`BaseTypeMimeTypeMatcher`** (new) — the dotAsset mime-matching algorithm extracted into a reusable matcher parameterized by base type + binary field var. `DotAssetAPIImpl.tryMatch` now delegates to it, so **DOTASSET behavior is unchanged**.
- **`DotAssetBaseTypeToContentTypeStrategyImpl`** — moved out of `BaseTypeToContentTypeStrategyResolver` into its own top-level class (per review feedback).
- **`FileAssetBaseTypeToContentTypeStrategyImpl`** (new) — mirrors dotAsset: matches the uploaded file's mime type against the `accept` (`ALLOWED_FILE_TYPES`) field variable of the `FILEASSET` content types, falling back to the system default `FileAsset`.
- **`BaseTypeToContentTypeStrategyResolver`** — registers both strategies and **fixes a latent bug** in `subscribe(...)`, which ignored its `baseContentType` argument and always registered under `DOTASSET` (it also threw `Multiple entries with same key` on any second registration).
- `MapToContentletPopulator.getContentTypeInode` is intentionally left untouched (resolution happens earlier in `ContentHelper.tryToSetContentType`).

### Frontend
- Reverted `uploadDotAsset` to its original 2-arg form (existing callers — browser selector, edit-content — unchanged).
- Added `newContentletByBaseType` (`DotWorkflowActionsFireService`) and `uploadFileByBaseType` (`DotUploadFileService`), which send `baseType` instead of `contentType`.
- Content Drive's upload selector now emits `DOTASSET` / `FILEASSET` and routes uploads through the new method.
- Root-level uploads now carry the **current site identifier** so the asset lands on the site being browsed instead of the backend default host.

https://github.com/user-attachments/assets/ce70909f-397c-44c3-a6c0-c09db78c1285

## Checklist
- [x] Tests — backend resolver unit test (subscribe regression + FILEASSET registration, written red-first), FileAsset strategy integration test, and frontend specs for the new methods, base-type selection, and current-site root upload.
- [x] Frontend: `data-access` (694) and `portlets-content-drive` (667) suites + lint pass.
- [x] Backend: unit test green; modules compile; integration test compiles (execution requires the enterprise-licensed integration environment / CI).

## Testing
1. Build core and run the integration test: `./mvnw verify -pl :dotcms-integration -Dcoreit.test.skip=false -Dit.test=FileAssetBaseTypeToContentTypeStrategyImplTest`
2. In Content Drive, upload a file and choose **File** → confirm a FileAsset is created (PDF → document type, image → image type, etc.); choosing **Asset** still creates a dotAsset.

This PR fixes: #36279


zJaaal:
Thanks for the automated review. I went through each finding against the actual code — the Critical/High items are false positives. Details below.

**🔴 Critical — `getMimeType` null/empty passed to matcher (`FileAssetBaseTypeToContentTypeStrategyImpl`)**
Not a bug. `FileAssetAPI.getMimeType(...)` returns `FileAsset.UNKNOWN_MIME_TYPE` (never null) when the type can't be determined. `MimeTypeUtils.match(...)` explicitly returns `false` when either side is `UNKNOWN_MIME_TYPE`, so the specific types (image/video/document) are correctly skipped and resolution falls through to the `*/*` tier → the default `FileAsset`. That's the intended fallback, not an incorrect match. This is also the exact same path the existing `DOTASSET` strategy already uses.

**🔴 Critical — `.get()` on optional without `isPresent()` → NPE (fallback)**
Hallucinated — there is no `.get()` call. The code is `Optional.ofNullable(APILocator.getContentTypeAPI(user).find(...))`. `find(...)` returns a `ContentType` or throws `NotFoundInDbException` (caught), so there's no NPE.

**🟠 High / 🟡 Medium (×2) — `DotAssetAPI.ALL_MIME_TYPE` not imported → `NoClassDefFoundError` (`BaseTypeMimeTypeMatcher`)**
Not a bug. `DotAssetAPI` lives in the same package (`com.dotcms.contenttype.business`), so no import is required, and `ALL_MIME_TYPE` / `PARTIAL_MIME_TYPE` are `public static final String` interface constants (inlined at compile time). The module compiles cleanly (`dotcms-core` build + `dotcms-integration` test-compile both pass).

**🟠 High — `MimeTypeUtils.match()` null `mimeType` → NPE**
Not a bug. `match()` is null-safe (`Try.of(() -> new MimeType(...)).getOrNull()` plus null guards), and the input is `UNKNOWN_MIME_TYPE`, not null.

**🟡 Medium — `Logger.debug` lacks stack trace**
Fair nit. It's a deliberate non-fatal swallow (the existing `DOTASSET` strategy logs nothing here), but I'm happy to include the exception object if preferred.

Net: no code changes needed for the Critical/High findings. Will address the logging nit if reviewers want it.

— Posted by Claude on behalf of @zJaaal

dotCMS/src/main/java/com/dotcms/contenttype/business/BaseTypeMimeTypeMatcher.java
```diff
@@ -0,0 +1,167 @@
...
+            throws DotDataException, DotSecurityException {
+
+        final List<ContentType> contentTypes = APILocator.getContentTypeAPI(user).findByType(baseContentType);
+
+        if (UtilMethods.isSet(contentTypes)) {
+
+            // Stores the content type indexed by mimetypes, on each index stores the subsets
+            // 0:Exact/on Site, 1:Exact/SYSTEM_HOST, 2:Partial Wildcard/on Site, 3:Partial Wildcard/SYSTEM_HOST,
+            // 4:Total Wildcard (or null)/on Site, 5:Total Wildcard (or null)/SYSTEM_HOST
+            final Map<String, ContentType>[] mimeTypeMappingArray = new Map[6];
```
fabrizzio-dotCMS: This code isn't very semantic. I know it is documented, but it's so hard to read.
Perhaps we could have a small static class instead of this array 
The fields on that class would make it a lot easier to read the precedence  of the resolved types

dotCMS/src/main/java/com/dotcms/contenttype/business/BaseTypeMimeTypeMatcher.java
```diff
@@ -0,0 +1,167 @@
...
+                        return Optional.ofNullable(entry.getValue());
+                    }
+                }
+            }
+        }
+
+        return Optional.empty();
+    }
+
+    private Map<String, ContentType> getMap(final Map<String, ContentType>[] mimeTypeMappingArray, final int index) {
```
fabrizzio-dotCMS: All these private methods are needed for the javadoc

dotCMS/src/main/java/com/dotcms/contenttype/business/FileAssetBaseTypeToContentTypeStrategyImpl.java
```diff
@@ -0,0 +1,97 @@
...
+            return Optional.ofNullable(APILocator.getContentTypeAPI(user)
+                    .find(FileAssetAPI.DEFAULT_FILE_ASSET_STRUCTURE_VELOCITY_VAR_NAME));
+        } catch (DotDataException | DotSecurityException e) {
+
+            Logger.debug(this, () -> "Could not resolve a FILEASSET content type: " + e.getMessage());
+            return Optional.empty();
+        }
+    }
+
+    private File getBinary (final List<File> binaryFiles, final Map<String, Object> contentletMap,
```
fabrizzio-dotCMS, line 92: in general try to add java doc to all your methods

dotCMS/src/main/java/com/dotcms/contenttype/business/BaseTypeMimeTypeMatcher.java
```diff
@@ -0,0 +1,167 @@
...
+ * <p>Matches are evaluated in priority order: exact mime type wins over a partial wildcard
+ * (e.g. {@code application/*}) which wins over a total wildcard ({@code *}{@code /}{@code *} or no
+ * "accept" restriction); within each tier, content types on the current site win over content
+ * types on the system host. A binary field with no "accept" field variable is treated as the total
+ * wildcard, so a content type that accepts everything resolves as the catch-all when nothing
+ * stricter matches.</p>
+ *
+ * @author dotCMS
+ */
+public class BaseTypeMimeTypeMatcher {
```
fabrizzio-dotCMS, line 42: This is a neat idea. love it

fabrizzio-dotCMS (commented):
it looks good however I think the inner logic of the matcher could be more semantic making it a lot easrier to read/understand

dotCMS/src/main/java/com/dotcms/contenttype/business/BaseTypeMimeTypeMatcher.java
zJaaal: Good call — done in c70c6d2b39. Replaced the `Map<String, ContentType>[6]` + magic indices (`? 5 : 4` etc.) with a `MimeTypeBuckets` inner class whose named fields make the precedence explicit:

```
exactCurrentSite → exactSystemHost → partialCurrentSite →
partialSystemHost → wildcardCurrentSite → wildcardSystemHost
```

A `bucketFor(accept, systemHost)` helper does the classification and `inPrecedenceOrder()` returns the buckets in resolution order. Behavior is unchanged (same precedence as the old indices), so the existing tests still hold.

— Claude, on behalf of @zJaaal

dotCMS/src/main/java/com/dotcms/contenttype/business/BaseTypeMimeTypeMatcher.java
zJaaal: Added javadoc to the matcher's private methods (`findContentType`, `mapToContentTypeMimeType`, `isSystemHostOrCurrentHost`) and to the new `MimeTypeBuckets` members in c70c6d2b39. The `getMap` helper that prompted this is gone with the array refactor.

— Claude, on behalf of @zJaaal

dotCMS/src/main/java/com/dotcms/contenttype/business/FileAssetBaseTypeToContentTypeStrategyImpl.java
zJaaal: Done in c70c6d2b39 — added javadoc to `apply(...)` and `getBinary(...)` in `FileAssetBaseTypeToContentTypeStrategyImpl`.

— Claude, on behalf of @zJaaal

dotCMS/src/main/java/com/dotcms/contenttype/business/BaseTypeMimeTypeMatcher.java
zJaaal: Thanks! 🙏

— Claude, on behalf of @zJaaal

fabrizzio-dotCMS (approved):

rjvelazco (approved):

dotcms-integration/src/test/java/com/dotcms/contenttype/business/FileAssetBaseTypeToContentTypeStrategyImplTest.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.Base64;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Integration test for {@link FileAssetBaseTypeToContentTypeStrategyImpl}, the strategy that
+ * resolves a {@link BaseContentType#FILEASSET} content type from an uploaded binary. Mirrors
+ * {@link DotAssetBaseTypeToContentTypeStrategyImplTest}.
+ */
+public class FileAssetBaseTypeToContentTypeStrategyImplTest extends IntegrationTestBase {
```
nollymar, line 52: This class must be included in a MainSuite

nollymar (changes requested):

dotcms-integration/src/test/java/com/dotcms/contenttype/business/FileAssetBaseTypeToContentTypeStrategyImplTest.java
zJaaal: Good catch — added `FileAssetBaseTypeToContentTypeStrategyImplTest` to `MainSuite2b` (import + class entry, right next to `DotAssetBaseTypeToContentTypeStrategyImplTest`) in 5230ab0442 so it runs in CI.

— Claude, on behalf of @zJaaal

nollymar (approved):

zJaaal:
Re: the latest automated review on `b558e20` — these are the same false positives already addressed [in this comment](https://github.com/dotCMS/core/pull/36317#issuecomment-4801579935); the bot re-fires per commit with shifted line numbers.

Quick confirmation against the current code:
- **`DotAssetAPI.ALL_MIME_TYPE` / `PARTIAL_MIME_TYPE` "undefined" → NoClassDefFoundError**: they're `public static final String` constants (`"*/*"`, `"/*"`) in `DotAssetAPI`, which is in the **same package** as `BaseTypeMimeTypeMatcher` (no import needed, inlined at compile time). The module compiles — `dotcms-core` build + unit test are green.
- **`.get()` on Optional → NoSuchElementException (FileAsset)**: the fallback is `Optional.ofNullable(APILocator.getContentTypeAPI(user).find(...))` — there is no `.get()` on it. The only `.get()` in the file is on a different optional guarded by `isPresent()`.
- **Fallback null/inactive**: `Optional.ofNullable` is null-safe; if empty, `ContentHelper.tryToSetContentType` throws a clean error. The default `FileAsset` is a system structure that always exists.

No code changes needed.

— Claude, on behalf of @zJaaal

PR #36353, opened 2026-06-29
fix(opensearch): repair bare orphan index on bootstrap without discarding populated indices (#36237)

### LINKED ISSUE #36237: OpenSearch: index bootstrap not idempotent — orphaned cluster index aborts startup

fabrizzio-dotCMS:
## Problem

Re-fix for **#36237** (QA failed the prior PRs #36238/#36240 on **TC-003**).

The idempotent-bootstrap reuse path introduced in #36238 re-asserted the custom mapping via `putMapping` against an orphaned cluster index (one that exists in the cluster but is missing from the dotCMS index store). On a **bare** orphan this failed:

```
INFO  Bootstrap: OS index already exists, reusing and re-asserting mapping: ...working_….os
ERROR MappingOperationsOS - putMapping failed for index ...working_….os — HTTP 400   (×8)
```

Root cause: the content mapping references the custom analyzer `my_analyzer`, defined in `os-content-settings.json`. Analyzers are **static** index settings that can only be applied at index creation — so a `putMapping`-only re-assert against a bare index is rejected (`analyzer [my_analyzer] not found`) and the index is left half-mapped, with the dotCMS dynamic templates missing.

## How the orphan arises

In the migration catch-up path the OS index name is **derived deterministically** by mirroring the ES name (`working_T0` → `cluster_X.working_T0.os`), not generated with a fresh timestamp. If a prior bootstrap created that physical index but crashed before committing its `VersionedIndices` store pointer, the next restart re-derives the **same** name, finds it already in the cluster, and the create fails with `resource_already_exists`.

## Fix

Decide the orphan's fate by **document count**, so a populated index is never discarded:

- **Empty orphan (0 docs)** → delete and recreate from scratch, restoring full **settings + base mapping + custom mapping**. An empty index has no data and no reindex progress, so recreating it costs nothing operationally — and it's the only case `putMapping`-reuse could not repair.
- **Populated orphan (>0 docs), or unknown count** → reuse in place, **untouched** (not deleted, not recreated, not remapped). A dotCMS-created index already carries the full mapping; deleting it would force a full reindex (hours, degraded/inconsistent search) — not justified to clean up an orphan. On any uncertainty (the count probe fails) we err toward reuse.

The delete only ever fires against a **demonstrably empty** orphan, and only in the bootstrap path for a store slot that is not registered — never against the active production index.

## Tests

- **Unit** (`ContentletIndexAPIImplBootstrapTest`, 8/8): empty→recreate, populated→reuse-untouched, doc-count-probe-fails→reuse, empty-orphan-delete-fails→still-creates, missing→create, create-fails→no-mapping, existence-probe-fails→create, OS-tag propagation.
- **Integration** (`ContentletIndexAPIImplMigrationIntegrationTest`): new regression IT creates a **bare** orphan against a real OpenSearch cluster, runs the bootstrap seam, and asserts the recreated index carries the dotCMS dynamic templates (`template_1`) **and** the `my_analyzer` setting. Verified green against OpenSearch:
  ```
  Tests run: 23, Failures: 0, Errors: 0, Skipped: 6  → BUILD SUCCESS
  ✅ bootstrap bare-orphan Phase 1 — recreated with full settings + mapping
  ```

Closes #36237

This PR fixes: #36237

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImpl.java
+++ b/dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImpl.java
@@ -671,7 +671,9 @@ public class ContentletIndexAPIImpl implements ContentletIndexAPI {
      *       {@code putMapping}-only re-assert against a bare orphan fails with {@code HTTP 400}
      *       (analyzer not found) and leaves the index half-mapped (issue #36237, QA TC-003). An
      *       empty index has no data and no reindex progress, so recreating it costs nothing
-     *       operationally and yields a clean index with full settings + base mapping.</li>
+     *       operationally and yields a clean index with full settings + base mapping. If the
+     *       delete cannot be confirmed and the index is still present, bootstrap fails loudly
+     *       rather than register a half-mapped index.</li>
      *   <li><b>Populated orphan (&gt; 0 docs), or count unknown</b> — reused in place, untouched.
      *       A populated orphan was created by dotCMS itself, so it already carries the full
      *       settings + base mapping + custom mapping; nothing needs to be (re)applied. The index is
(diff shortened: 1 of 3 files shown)


fabrizzio-dotCMS:
Addressed the 🟡 Medium finding (failed orphan deletion → create failure) in 47b6899.

The empty-orphan branch no longer falls through to a create after a non-acknowledged delete. If the delete isn't confirmed (so the index may still exist), the orphan is now **reused in place** instead of attempting a create that would throw `resource_already_exists` — which `bootstrapAndPointOS` doesn't catch (it only handles `IOException`) and would abort bootstrap, i.e. the exact original #36237 failure. The orphan is empty so nothing is lost, and a later clean restart recreates it properly once the cluster is healthy.

Unit test `test_emptyOrphanDeleteFails_reusedInPlace_notRecreated` updated to assert no create / no remap on a non-acknowledged delete. All 8 unit tests green.

fabrizzio-dotCMS:
Addressed the 🔴 Critical finding (failed orphan deletion leaves broken index state) in 8b8da22.

You're right — the previous revision's `return true` on an unacknowledged delete would register a bare, un-repairable orphan (no custom analyzer → mapping cannot be fixed in place), leaving search permanently broken.

New behavior when the empty-orphan delete is **not** acknowledged — re-probe existence:
- **index is gone** (delete took effect without an ack) → recreate cleanly;
- **index still present** → **throw** and fail bootstrap loudly with a clear message, rather than recreating (would throw `resource_already_exists`) or registering a half-mapped index. This is an abnormal cluster state (not the orphan-name collision the method resolves), so failing loud — consistent with #36240 — is the correct outcome.

Unit tests split into `test_emptyOrphanDeleteUnacked_butIndexGone_recreates` and `test_emptyOrphanDeleteFails_indexStillExists_failsLoud`. All 9 unit tests green.

jcastro-dotcms (approved):

PR #36359, opened 2026-06-29
fix(opensearch): malformed OS_ENDPOINTS falls back to ES instead of crashing startup (#35636)

### LINKED ISSUE #35636: [QA-G2] OpenSearch Migration — Invalid OS_ENDPOINTS Configuration (TC-006–TC-008)
## Parent Epic

#35476

## Glossary (quick)

- **Old engine (ES)** — Elasticsearch, what dotCMS uses today. Port `9200`, inspected with **Kibana** (`5601`).
- **New engine (OS)** — OpenSearch, the migration target. Port `9201`, inspected with **OpenSearch Dashboards** (`5602`).
- **Migration phase** — `0` = old engine only · `1` = write to *both*, search the old one · `2` = write to both, read OS.
- **Dual-write** — in Phase 1 every content change is written to *both* engines.
- **`.os` suffix (the "tag")** — marks a new-engine index. New-engine index names **end in `.os`** (e.g. `cluster_08abc3.working_20260406.os`); old-engine names do not. Visible everywhere: cluster, DB, REST. (NOT an `os::` prefix — that's an obsolete model.)

## Environment

```bash
docker compose -f docker/docker-compose-examples/os-migration/docker-compose.yml up -d
```

| Service | URL | Used for |
|---|---|---|
| Old engine (ES 7.10) | http://localhost:9200 | Direct REST queries |
| Kibana | http://localhost:5601 | Inspect/query the OLD index |
| New engine (OS 3.x) | http://localhost:9201 | Direct REST queries |
| OpenSearch Dashboards | http://localhost:5602 | Inspect/query the NEW index |
| dotCMS | http://localhost:8082 | Admin UI + REST API (`admin:admin`) |

Migration settings live in `dotmarketing-config.properties` (`FEATURE_FLAG_OPEN_SEARCH_PHASE`, `OS_ENDPOINTS=http://localhost:9201`, `OS_AUTH_TYPE=BASIC`, `OS_AUTH_BASIC_USER=admin`, `OS_AUTH_BASIC_PASSWORD=admin`, `OS_TLS_ENABLED=false`).

## Helpers (referenced by the steps below)

**H1 — Find your real index names.** Names contain a timestamp (differ per machine). **New-engine indexes end in `.os`**; old-engine names do not.
```bash
curl -s -u admin:admin http://localhost:8082/api/v1/esindex | jq .   # what dotCMS knows (keeps .os, drops cluster prefix)
curl -s -u admin:admin http://localhost:9200/_cat/indices?v          # OLD engine physical names (no .os)
curl -s -u admin:admin http://localhost:9201/_cat/indices?v          # NEW engine physical names (end in .os)
```
Or **Admin → System → Index**.

**H2 — Which phase did the system end up in?**
- **UI / index list:** Admin → System → Index (or H1). Only old-engine indexes (no `.os`) → Phase 0. New-engine indexes present (`.os`) → Phase 1.
- **Log:** find the migration-phase line (similar to `Migration Phase: ...`). If the shutdown fired you also see the error + "phase reset to 0" lines (see H5).

**H3 — Check the database** (SQL console; new-engine marker is the **`.os` suffix**, not an `os::` prefix):
```sql
SELECT * FROM indicies WHERE index_name LIKE '%.os';      -- NEW engine rows
SELECT * FROM indicies WHERE index_name NOT LIKE '%.os';  -- OLD engine rows
SELECT CASE WHEN index_name LIKE '%.os' THEN 'NEW' ELSE 'OLD' END AS engine, COUNT(*) FROM indicies GROUP BY 1;
```

**H5 — Exact log lines of an automatic migration shutdown** (Phase 1/2). Three lines, in order — **two `ERROR` + one `WARN`, NO `FATAL`**:
```
ERROR  OpenSearch configuration error — halting OS migration, dotCMS will fall back to ES-only: <reason>
ERROR  OpenSearch migration halted: invalid configuration detected at startup. Verify OS_ENDPOINTS, OS version, and FEATURE_FLAG_OPEN_SEARCH_PHASE, then restart dotCMS.
WARN   Migration phase reset to PHASE_0_MIGRATION_NOT_STARTED (was PHASE_1_DUAL_WRITE_ES_READS). This change is runtime-only — persist it in dotmarketing-config.properties to survive a restart.
```
`<reason>` is one of:
- unreachable / bad URL: `OpenSearch cluster is not reachable: <cause>. Check OS_ENDPOINTS configuration.`
- wrong version: `OpenSearch version mismatch: expected 3.x but connected cluster reports version <X>. Check OS_ENDPOINTS configuration.`
- same address: `ES and OS clients point to the same endpoint(s): <set>. … Set OS_ENDPOINTS to a separate OpenSearch instance.`

On a successful start you instead see `INFO OS version check passed: <version>`.

## Test Cases

> With a bad OpenSearch address, **dotCMS still starts normally and keeps serving from ES** — no crash, no hang. The migration is left off and a clear error is logged. Each case includes how to reproduce the bad config.

### TC-006 — A malformed `OS_ENDPOINTS` value does not crash dotCMS
- **Objective:** With a clearly invalid address, dotCMS still starts and serves from ES; the bad value is reported and the migration is left off.
- **How to induce:** set a value that is not a valid URL.
- **Risk:** High
- **Preconditions:** `FEATURE_FLAG_OPEN_SEARCH_PHASE=1`, `OS_ENDPOINTS=not-a-valid-url:xyz`.
- **Steps:**
  1. Start dotCMS; confirm the process stays up (no failed deployment).
  2. Check phase (**H2**) and DB (**H3**).
  3. Create/publish/search a content item.
- **Expected Result:**
  - **dotCMS starts and runs normally** — no uncaught exception, no failed deployment.
  - Migration left off (Phase 0); log shows H5 lines, reason `OpenSearch cluster is not reachable: <cause>. Check OS_ENDPOINTS configuration.`
  - Content operations work on the old engine. No raw stack trace.

### TC-007 — A non-existent OpenSearch host doesn't hang dotCMS forever
- **Objective:** If the host doesn't exist, dotCMS must eventually time out and finish starting (serving from ES) — not wait forever. Migration left off.
- **How to induce:** point `OS_ENDPOINTS` at a hostname that does not resolve.
- **Risk:** Medium
- **Preconditions:** `FEATURE_FLAG_OPEN_SEARCH_PHASE=1`, `OS_ENDPOINTS=http://nonexistent-host.invalid:9201`.
- **Steps:**
  1. Start dotCMS. It may pause while the connection times out — wait for startup to finish.
  2. Check the phase (**H2**).
- **Expected Result:**
  - **dotCMS finishes starting and is operational on the old engine — NOT stuck** waiting.
  - After timeout, migration left off; log shows H5 lines, reason `OpenSearch cluster is not reachable: <cause>. Check OS_ENDPOINTS configuration.`

- **Objective:** When the configured OpenSearch address answers but reports the wrong version (not 3.x), dotCMS still starts and serves from ES; migration left off.
- **How to induce:** start dotCMS with the OpenSearch connection string pointing at the **Elasticsearch** URL — `OS_ENDPOINTS=http://localhost:9200` (the ES port). ES replies version 7.10.x, which is not `3.x`.
- **Risk:** Medium
- **Preconditions:** Both engines up; `FEATURE_FLAG_OPEN_SEARCH_PHASE=1`, `OS_ENDPOINTS=http://localhost:9200`.
- **Steps:**
  1. Start dotCMS and read the startup log.
  2. Check phase (**H2**); list indexes (**H1**) and check DB (**H3**).
- **Expected Result:**
  - **dotCMS starts and runs normally on the old engine.**
  - Migration left off; log shows H5 lines, reason `OpenSearch version mismatch: expected 3.x but connected cluster reports version 7.10.x. Check OS_ENDPOINTS configuration.` then the phase-reset `WARN`.
  - Only old-engine indexes listed; no `.os` rows in DB.
  - (Because OS address == ES address, the version check fails first; the "same endpoint(s)" check is a secondary guard — see G12/TC-037.)

## Acceptance Criteria
- [ ] TC-006: malformed URL → dotCMS up & serving ES; H5 lines with "not reachable" reason; no crash/stack trace.
- [ ] TC-007: non-existent host → dotCMS finishes startup (not stuck); migration off; H5 "not reachable".
- [ ] TC-008: OS_ENDPOINTS=ES URL → dotCMS up & serving ES; H5 with "version mismatch: expected 3.x … 7.10.x"; 0 `.os` rows.

## Notes
- All Manual. Same underlying safety mechanism as Group 1, triggered through configuration.

fabrizzio-dotCMS:
## Problem

Re-fix for **#35636** (QA-G2, TC-006). QA result:

- **TC-007** (unreachable host) → ✅ falls back to ES.
- **TC-008** (wrong version / points at ES) → ✅ falls back to ES.
- **TC-006** (malformed URL, `DOT_OS_ENDPOINTS=invalid-url-format`) → ❌ **dotCMS crashes and is unrecoverable**, in every migration phase.

## Root cause

`OSDefaultClientProvider`'s no-arg constructor built the OpenSearch client **eagerly** (parsing the endpoint URLs). Weld's client-proxy subclass constructor invokes `super()`, so the build ran the moment the proxy was created — during `ContentletIndexAPIImpl`'s constructor (`InitServlet.init()`), **before** the phase-aware `IndexStartupValidator`.

```
java.lang.IllegalArgumentException: URI is not absolute
  at ConfigurableOpenSearchProvider.createHttpHosts(:204)
  at ConfigurableOpenSearchProvider.<init>(:54)
  at OSDefaultClientProvider.<init>(:37)
  at OSDefaultClientProvider$Proxy$_$$_WeldClientProxy.<init>(...)   ← Weld proxy ctor → super()
  at ContentletIndexAPIImpl.<init>(:178)
```

- Well-formed-but-unreachable URL (TC-007/008): eager build **succeeds**; the failure is deferred to the later `getClient().info()` call inside the validator → caught → fallback. ✅
- Malformed URL (TC-006): eager build **throws during construction** → escapes to `InitServlet.init()` (no try/catch) → startup aborts. ❌

Secondary: `createHttpHosts` only caught `MalformedURLException`, but a scheme-less value throws `IllegalArgumentException: URI is not absolute`, so the intended clear message was bypassed.

## Fix

- **`OSDefaultClientProvider`** builds its delegate **lazily** on the first `getClient()` (double-checked locking). The no-arg CDI constructor no longer builds; the explicit-config constructor stays eager (preserves the test contract). This routes a malformed `OS_ENDPOINTS` through the **existing** chokepoint that already passes TC-007/008 — `IndexStartupValidator.validateOSVersion()` (and `OSIndexAPIImpl.waitUtilIndexReady()`) catch `getClient()` failures, so **phases 1/2 `haltMigration()` → ES-only** and **phase 3 aborts** (acceptable: OS is primary and unreachable). No new fallback logic.
- **`ConfigurableOpenSearchProvider.createHttpHosts`** now also catches `IllegalArgumentException` and throws a clear `Invalid OS_ENDPOINTS URL: '<value>'` message.

No changes to `IndexStartupValidator`, `checkAndInitializeIndex`, `haltMigration`, or Phase 3 handling.

## Tests

`OSClientProviderIntegrationTest` (verified green against a real OpenSearch cluster — `Tests run: 10, Failures: 0, Errors: 0`):
- `test_createProvider_withInvalidEndpoint_throwsClearMessage` — malformed endpoint surfaces the clear "Invalid OS_ENDPOINTS URL" message.
- `test_defaultProvider_malformedEndpoints_isLazy_failsOnlyOnGetClient` — the default provider's no-arg construction does **not** crash on a malformed `OS_ENDPOINTS`; the failure appears only on `getClient()` (the regression guard for TC-006).

### Manual QA repro (TC-006)
`docker compose -f docker/docker-compose-examples/os-migration/docker-compose.yml up` with `FEATURE_FLAG_OPEN_SEARCH_PHASE=1`, `DOT_OS_ENDPOINTS=invalid-url-format` → dotCMS starts, serves from ES, logs the fallback (H5) lines, 0 `.os` rows.

Closes #35636

This PR fixes: #35636


jcastro-dotcms (approved):

fabrizzio-dotCMS:
### Re: dotBot finding — *"Repeated client rebuild attempts on non-transient errors"* (`OSDefaultClientProvider.java:60`, 🟡 Medium)

**Acknowledged — won't fix (intentional).**

The lazy `provider()` leaves the field `null` when `new ConfigurableOpenSearchProvider()` throws, so a subsequent `getClient()` retries the build. That retry behavior is deliberate and bounded:

- **Construction failure is cheap and permanent.** The build only throws on a malformed `OS_ENDPOINTS` (TC-006) — i.e. URL parsing, no connections opened. The config doesn't change at runtime, so there's nothing to "recover" and no resource churn (a well-formed-but-unreachable endpoint builds fine; that failure happens later at `getClient().info()`, not at construction).
- **Retries are bounded by the existing chokepoint.** `getClient()` failures are caught by `IndexStartupValidator.validateOSVersion()` / `OSIndexAPIImpl.waitUtilIndexReady()` — the very design this PR relies on. After `haltMigration()` in phases 1/2 the OS path is no longer exercised, so this is not a hot loop. In phase 3 startup aborts (acceptable — OS is primary and unreachable).
- **Memoizing the failure would add cost, not value.** Caching/rethrowing a `Throwable` to avoid a bounded, cheap retry trades real complexity for a non-issue.

Net: the retry is acceptable and intentional given how a malformed `OS_ENDPOINTS` is routed through the validator. No code change.

PR #36362, opened 2026-06-29
fix(maintenance): import DEFAULT variant so experiment starter restore works (#36337)

### LINKED ISSUE #36337: Starter export/import omits Experiments and Variants data
## Problem Statement

The maintenance **starter export** endpoints do not include **Experiments** or their **Variants**, and the import path does not restore them. As a result, a starter generated for backup or migration silently loses all A/B testing experiment data.

Affected endpoints:
- `GET /api/v1/maintenance/_downloadStarterWithAssets`
- `GET /api/v1/maintenance/_downloadStarter`

**Root cause:** `ExportStarterUtil` only exports (a) Hibernate-mapped entities plus a small hardcoded list (`getTableSet()`), and (b) a few hand-coded non-Hibernate "special" tables — Counter, Image, Portlet, Workflows, Rules (`getAdditionalDataAsJSON()`). The `experiment` table is persisted via `DotConnect` (not Hibernate) and is in neither list, so it is never written to the starter. The same applies to the `variant` table, which holds the experiment-variant rows (`dotexperiment-<id>-variant-<n>`) that the already-exported variant content (`contentlet_version_info` / `multi_tree`) references — without those rows, restored experiments resolve to orphaned variants.

**Impact / scope:**
- Affects starter-based backup and instance-to-instance migration.
- Particularly serious on **MSSQL** or any environment where the `pg_dump` binary is unavailable, because the JSON starter is then the *only* backup path. (The separate `pg_dump` path — `_pgDumpAvailable` / `_downloadDb` — already includes `experiment`/`variant` since they are not in its `--exclude-table-data` deny-list, so that path is **not** affected.)

## Steps to Reproduce

1. On a dotCMS instance, create an A/B testing **Experiment** on a page (this writes a row to the `experiment` table and `dotexperiment-…` rows to the `variant` table).
2. Call `GET /api/v1/maintenance/_downloadStarterWithAssets` (or `_downloadStarter`) as an admin.
3. Unzip the downloaded starter and inspect the JSON entries.

**Expected:** The ZIP contains the experiment data (e.g. `Experiment.json`) and the variant rows (e.g. `Variant.json`), and importing the starter into a clean instance restores the experiment and its variants.

**Actual:** No experiment or variant data is present in the ZIP. After importing such a starter, the experiment and its variants do not exist on the target instance.

## Acceptance Criteria

- [ ] `GET /api/v1/maintenance/_downloadStarter` and `_downloadStarterWithAssets` both include the `experiment` table data in the generated starter.
- [ ] Both endpoints include the `variant` table data (all rows, including archived and `DEFAULT`).
- [ ] Importing a starter restores experiments via `ImportStarterUtil`, preserving the experiment `id` and JSONB fields (`traffic_proportion`, `scheduling`, `goals`, `running_ids`).
- [ ] Importing restores variant rows, skipping the always-present `DEFAULT` variant and not failing on archived experiment variants.
- [ ] Import ordering is correct: variants are imported before contentlet version info / multi-tree; experiments are imported after contentlets (experiment `page_id` references a contentlet identifier).
- [ ] Round-trip (export → import into a clean instance) results in the experiment opening correctly in the UI with its variants intact.
- [ ] Works on both PostgreSQL and MSSQL (JSONB vs NVARCHAR handling via existing `DotConnect.addJSONParam`).
- [ ] The `pg_dump` path (`_downloadDb`) remains unchanged.

## dotCMS Version

Latest from `main` branch.

## Links

NA

erickgonzalez:
## Proposed Changes

Follow-up to #36339 (export/import of Experiments and Variants in the starter).

Manual round-trip testing surfaced a fatal error when importing a starter that contains experiments — startup aborts with:

```
Cannot construct instance of `com.dotcms.experiments.model.TrafficProportion`,
problem: Invalid Variants provided
```

### Root cause
On a fresh install the starter import runs in `MainServlet.init()` at **line 129** (`executeStartUpTasks()` → `Task00004LoadStarter` → `ImportStarterUtil`), while the `DEFAULT` variant is created later at **line 218** (`DotInitializationService` → `DefaultVariantInitializer`). `postgres.sql` does **not** seed the `DEFAULT` variant row, and the run-once upgrade tasks don't run on this path.

Every experiment's `trafficProportion` references the `DEFAULT` variant. Building the `Experiment` immutable during import triggers `AbstractTrafficProportion`'s `@Value.Check`, which validates that every referenced variant exists via `VariantAPI.get(...)`. Because the import branch was **explicitly skipping** the `DEFAULT` variant, it was absent at deserialize time → validation fails → startup is aborted.

### Fix
Stop skipping `DEFAULT` in the variant import branch — import it from the starter like any other variant. The existing `variantFactory.get(name).isPresent()` guard still prevents PK duplicates and makes it a no-op when the initializer already created the row. With `DEFAULT` (and the experiment variants) present before experiments deserialize, the model builds and `ExperimentsFactory.save()` works as designed.

This keeps the factory-based import (no raw SQL) and changes only `ImportStarterUtil`.

### Also included
- `fix(docker)`: bump `CUSTOM_STARTER_URL` references to the latest starter version (`20260629`) across docker-compose examples and docs, so deployments initialize from a starter that exercises this path.

## How to Test

1. On an instance with at least one A/B Experiment, download a starter (`_downloadStarter` / `_downloadStarterWithAssets`).
2. Point a fresh install at that starter (e.g. `CUSTOM_STARTER_URL`) and boot it.
3. Startup completes (no `Invalid Variants provided`), and the experiment + its variants are present and open correctly in the UI.

This PR fixes: #36337

This PR fixes: #36337

### DIFF
--- a/docker/dev-env/README.md
+++ b/docker/dev-env/README.md
@@ -72,7 +72,7 @@ docker run --rm \
 --pull always \
 -p 8443:8443 \
 -v $PWD/data:/data \
--e DOTCMS_STARTER_URL=https://repo.dotcms.com/artifactory/libs-release-local/com/dotcms/starter/20250613/starter-20250613.zip \
+-e DOTCMS_STARTER_URL=https://repo.dotcms.com/artifactory/libs-release-local/com/dotcms/starter/20260629/starter-20260629.zip \
 dotcms/dotcms-dev:nightly
 
 ```
--- a/docker/docker-compose-examples/cluster-mode/README.md
+++ b/docker/docker-compose-examples/cluster-mode/README.md
@@ -23,7 +23,7 @@ The license pack must contain at least two licenses (one for each node in the cl
 3. A custom starter can be set through this line (uncomment and change the starter url accordingly):
 
 ```
-#"CUSTOM_STARTER_URL": 'https://repo.dotcms.com/artifactory/libs-release-local/com/dotcms/starter/20250722/starter-20250722.zip'
+#"CUSTOM_STARTER_URL": 'https://repo.dotcms.com/artifactory/libs-release-local/com/dotcms/starter/20260629/starter-20260629.zip'
 ```
 
 #### Deploying nodes:
--- a/docker/docker-compose-examples/cluster-mode/docker-compose-node-1.yml
+++ b/docker/docker-compose-examples/cluster-mode/docker-compose-node-1.yml
@@ -44,7 +44,7 @@ services:
         DOT_ES_ENDPOINTS: 'https://opensearch:9200'
         DOT_INITIAL_ADMIN_PASSWORD: 'admin'
         DOT_DOTCMS_CLUSTER_ID:  'dotcms-production'
-        #CUSTOM_STARTER_URL: 'https://repo.dotcms.com/artifactory/libs-release-local/com/dotcms/starter/20260409/starter-20260409.zip'
+        #CUSTOM_STARTER_URL: 'https://repo.dotcms.com/artifactory/libs-release-local/com/dotcms/starter/20260629/starter-20260629.zip'
     depends_on:
       - opensearch
       - db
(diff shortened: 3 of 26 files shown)


erickgonzalez:
@claude create a description for this PR

jcastro-dotcms (approved):

dario-daza (approved):

Issue #36218, opened 2026-06-18
[QA-G12] OpenSearch Migration — Connection & Phase Behavior (TC-037–TC-040)


fabrizzio-dotCMS:
## Parent Epic

#35476

## Glossary (quick)

- **Old engine (ES)** — Elasticsearch, what dotCMS uses today. Port `9200`, inspected with **Kibana** (`5601`).
- **New engine (OS)** — OpenSearch, the migration target. Port `9201`, inspected with **OpenSearch Dashboards** (`5602`).
- **Migration phase** — `0` = old engine only · `1` = write to *both*, search the old one · `2` = write to both, read OS.
- **Dual-write** — in Phase 1 every content change is written to *both* engines.
- **`.os` suffix (the "tag")** — marks a new-engine index. New-engine index names **end in `.os`** (e.g. `cluster_08abc3.working_20260406.os`); old-engine names do not. Visible everywhere: cluster, DB, REST. (NOT an `os::` prefix — that's an obsolete model.)

## Environment

```bash
docker compose -f docker/docker-compose-examples/os-migration/docker-compose.yml up -d
```

| Service | URL | Used for |
|---|---|---|
| Old engine (ES 7.10) | http://localhost:9200 | Direct REST queries |
| Kibana | http://localhost:5601 | Inspect/query the OLD index |
| New engine (OS 3.x) | http://localhost:9201 | Direct REST queries |
| OpenSearch Dashboards | http://localhost:5602 | Inspect/query the NEW index |
| dotCMS | http://localhost:8082 | Admin UI + REST API (`admin:admin`) |

Migration settings live in `dotmarketing-config.properties` (`FEATURE_FLAG_OPEN_SEARCH_PHASE`, `OS_ENDPOINTS=http://localhost:9201`, `OS_AUTH_TYPE=BASIC`, `OS_AUTH_BASIC_USER=admin`, `OS_AUTH_BASIC_PASSWORD=admin`, `OS_TLS_ENABLED=false`).

## Helpers (referenced by the steps below)

**H2 — Which phase did the system end up in?**
- **UI / index list:** Admin → System → Index (or H1). Only old-engine indexes (no `.os`) → Phase 0. New-engine indexes present (`.os`) → Phase 1.
- **Log:** find the migration-phase line (similar to `Migration Phase: ...`). If the shutdown fired you also see the error + "phase reset to 0" lines (see H5).

[...]

Issue #35476, opened 2026-04-27
[QA EPIC] OpenSearch Migration — End-to-End QA Validation (Phases 0–3)


fabrizzio-dotCMS:
### Description

QA sub-epic for the ES → OpenSearch migration pipeline under [#34164](https://github.com/dotCMS/core/issues/34164). Covers end-to-end validation of the dual-write pipeline across Phases 0–3, plus acceptance criteria for PR #35632 (startup hardening), PR #35735 (TLS/HTTPS connectivity fix), and PR #35609 (vendor-neutral SearchAPI — Phase 2/3 reads now testable).

**Parent epic:** #34164 — Migrate Mapping Layer classes
**Fix PRs:**
- #35632 — Startup validation hardening, `haltMigration()`, phase-aware index init, thread-safe formatter
- #35735 — TLS/HTTPS connectivity fix: SSL context wired correctly, `OS_TLS_CERT_REQUIRED` property added
- #35609 — Vendor-neutral `SearchAPI` and phase-aware router: Phase 2/3 read path now migrated

**Test plans:**
- `docs/backend/OPENSEARCH_MIGRATION_TEST_PLAN.md` — base scenarios A–L, F (Phases 0–3, TC-001–TC-035)

### Environment Setup

```bash
# Start migration stack (ES 7.10 + Kibana + OpenSearch 3.x + Dashboards)
docker compose -f docker/docker-compose-examples/os-migration/docker-compose.yml up -d
```

| Service               | URL                          | Purpose                          |
|-----------------------|------------------------------|----------------------------------|
| Elasticsearch 7.10    | http://localhost:9200        | ES REST API                      |
| Kibana                | http://localhost:5601        | ES index inspection / queries    |
| OpenSearch 3.x        | http://localhost:9201        | OS REST API                      |
| OpenSearch Dashboards | http://localhost:5602        | OS index inspection / queries    |

```properties
# dotmarketing-config.properties — HTTP local setup
OS_ENDPOINTS=http://localhost:9201
OS_AUTH_BASIC_USER=admin
OS_AUTH_BASIC_PASSWORD=admin

# Phase flag: 0=ES only  1=dual-write+ES reads  2=dual-write+OS reads  3=OS only (search reads testable)
FEATURE_FLAG_OPEN_SEARCH_PHASE=0
```

```properties
# dotmarketing-config.properties — HTTPS / Kubernetes setup (PR #35735)
DOT_OS_ENDPOINTS=https://opensearch-host:9200
[...]

PR #36421, opened 2026-07-03
fix(opensearch): enforce ES/OS endpoint separation at the OS index-creation chokepoint (#36419)

### LINKED ISSUE #36419: OpenSearch: endpoint-separation guard runs after index bootstrap, so .os shadow indices are still created when ES==OS
## Problem Statement

During QA of the ES→OpenSearch migration (surfaced by #36218, test case TC-037; part of epic #35476), the endpoint-separation guard fires but too late to be effective.

When the old engine (ES) and new engine (OS) are configured with the **same** address (`DOT_ES_ENDPOINTS == DOT_OS_ENDPOINTS`) and `DOT_FEATURE_FLAG_OPEN_SEARCH_PHASE=1`, `IndexStartupValidator` correctly detects the overlap and halts the migration to Phase 0 — but it runs **~6 seconds after the index bootstrap has already created the `.os` shadow indices**. As a result the `.os` indices and their `indicies` rows persist despite the migration being refused, failing TC-037's "no `.os` rows" acceptance criterion.

**Root cause:** ordering. `IndexStartupValidator.validateIndexingConfig()` runs *after* `createContentIndexes()` creates the shadow index set.

**Impact:** the separation guard's purpose (leave nothing behind when ES==OS) is not achieved; the halt only stops *subsequent* dual-write, not the initial `.os` creation. Migration is behind a feature flag, so production impact is limited to migration testing/rollout.

## Steps to Reproduce

Rig: `docker/docker-compose-examples/single-node-os-migration/` (old engine = OpenSearch 1.3.20 `:9200`, new engine = OpenSearch 3.4.0 `:9201`).

1. Point both engines at the same cluster and set Phase 1:
   ```yaml
   DOT_FEATURE_FLAG_OPEN_SEARCH_PHASE: '1'
   DOT_ES_ENDPOINTS: 'https://opensearch3:9200'   # == DOT_OS_ENDPOINTS
   DOT_ES_TLS_TRUST_SELF_SIGNED: 'true'
   ```
2. Bring up fresh (`down -v` then `up -d`); wait for dotCMS ready.
3. Check the DB: `SELECT count(*) FROM indicies WHERE index_name LIKE '%.os';`
4. Check the cluster: `curl -sk .../9201/_cat/indices | grep '\.os'`

**Observed** (guard fires, but `.os` created first):
```
15:01:44–45  OSIndexAPIImpl createIndex ...working_150142.os / ...live_150142.os   (phase still 1)
15:01:51     validation FAILED — "ES and OS clients point to the same endpoint(s): [opensearch3:9200]..."
15:01:51     Migration phase reset to PHASE_0_MIGRATION_NOT_STARTED
```
Result: **2 `.os` rows** in `indicies` + 2 `.os` cluster indices (expected: 0).

## Acceptance Criteria

- [ ] When `DOT_ES_ENDPOINTS == DOT_OS_ENDPOINTS` at startup with Phase ≥ 1, no `.os` indices are created and no `.os` rows exist in `indicies`.
- [ ] The separation guard runs before any index bootstrap/dual-write (or `.os` artifacts created before the halt are cleaned up on halt).
- [ ] The existing halt behavior (reset to Phase 0, ES-only fallback, "same endpoint(s)" log) is preserved.

## dotCMS Version

`dotcms/dotcms:latest` (verified against release `26.06.30-01`), `single-node-os-migration` rig, migration Phase 1.

## Severity

## Links

NA — surfaced by QA of #36218 (epic #35476).

swicken:
### Proposed Changes

Fixes the endpoint-separation guard so it actually prevents OpenSearch `.os` shadow indices from being created when the ES and OS clients point at the **same** cluster.

**Root cause:** the separation check only ran in `InitServlet → checkAndInitializeIndex`, which executes *after* the empty-DB starter-load path (`Task00004LoadStarter → bootstrapAndPoint → bootstrapAndPointOS`) has already created the `.os` indices. With `ES==OS` the guard halted the migration but too late — the `.os` indices and their `indicies` rows already existed.

* **`ContentletIndexAPIImpl.bootstrapAndPointOS`** — add a config-only endpoint-separation gate at the single OS index-creation chokepoint, *before* the existing connection gate. On `ES==OS` overlap it logs, calls `haltMigration()` (ES-only fallback), and skips OS bootstrap — so no `.os` artifacts are created on **any** startup path (starter-load or InitServlet).
* **`IndexStartupValidator`** — extract the overlap logic into a shared `assertEndpointsSeparate(config)` and add a network-free `endpointsAreSeparate()` gate reused by both startup validation and the chokepoint (no duplicated comparison logic).
* **`IndexStartupValidatorTest`** — new unit tests for the separation gate (same-address → false, different → true, scheme-normalized overlap → false, assert throws with the overlap message).

Closes #36419

### Verification

Unit: `IndexStartupValidatorTest` 4/4 pass.

E2E (`single-node-os-migration` rig, branch image, `DOT_ES_ENDPOINTS == DOT_OS_ENDPOINTS == https://opensearch3:9200`, Phase 1):

| Check | Before | After |
|---|---|---|
| `.os` rows in `indicies` | 2 | **0** |
| `.os` cluster indices | 2 | **0** |
| `.os` createIndex attempts | created | **0 (prevented)** |
| Migration halted → Phase 0 | ✓ | ✓ |
| dotCMS serving on ES | ✓ | ✓ |

Log confirms the gate fires at the chokepoint before any OS creation:
```
ERROR IndexStartupValidator - OpenSearch endpoint-separation check failed: ES and OS clients point to the same endpoint(s): [opensearch3:9200]...
WARN  ContentletIndexAPIImpl - Skipping OpenSearch index bootstrap (...): OS endpoints overlap ES (same cluster). Migration halted (now ES-only)...
WARN  MigrationPhase - Migration phase reset to PHASE_0_MIGRATION_NOT_STARTED (was PHASE_1_DUAL_WRITE_ES_READS)...
```

### Checklist
- [x] Tests
- [ ] Translations
- [x] Security Implications Contemplated — config-only comparison, no new inputs; strengthens an existing safety guard (prevents dual-write to a shared cluster).

### Additional Info

Found during QA of #36218 (TC-037), part of the OpenSearch migration QA epic #35476. The existing halt behavior (reset to Phase 0, ES-only fallback, "same endpoint(s)" log) is preserved; this only closes the ordering window on the fresh-install/starter-load path.

### Known limitations / follow-ups (from review)

- **Runtime phase-flip (out of scope for #36419):** raising `FEATURE_FLAG_OPEN_SEARCH_PHASE` 0→1+ at runtime without a restart and then running a full reindex creates `.os` reindex slots via `initAndPointReindex → createContentIndex` fan-out, which does not pass through this bootstrap gate. Startup validation remains the primary guard.
- **Pre-existing (not introduced here):** the halt is runtime-only and subject to system-table shadowing (`MigrationPhase.reset()` javadoc). If the phase flag lives in the DB system table, the reset is a no-op from `current()`'s perspective and dual-writes could continue.

This PR fixes: #36419

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/content/index/opensearch/IndexStartupValidator.java
+++ b/dotCMS/src/main/java/com/dotcms/content/index/opensearch/IndexStartupValidator.java
@@ -87,6 +87,14 @@ public class IndexStartupValidator {
         final OSClientConfig config = resolveConfig();
         logConfigSummary(config);
         validateOSVersion();
+        if (IndexConfigHelper.isMigrationComplete()) {
+            // Phase 3: ES is decommissioned and ES_ENDPOINTS is no longer required, so the
+            // resolved ES side would just be the localhost:9200 default — comparing it against
+            // OS would false-positive on a legitimate config. Skip the separation check.
+            Logger.info(this, "Endpoint separation check skipped: PHASE_3_OPENSEARCH_ONLY"
+                    + " — ES is decommissioned and ES_ENDPOINTS is not required.");
+            return;
+        }
         assertEndpointsSeparate(config);
     }
 
@@ -242,10 +250,21 @@ public class IndexStartupValidator {
      * (empty-DB starter-load or populated-DB InitServlet), closing the window where the
      * starter-load path created {@code .os} indices before the late startup validation ran.
      *
+     * <p><strong>Phase-aware:</strong> in {@code PHASE_3_OPENSEARCH_ONLY} the check is skipped
+     * (returns {@code true}) — ES is decommissioned and {@code ES_ENDPOINTS} is no longer
+     * required, so the resolved ES side would just be the {@code localhost:9200} default and
+     * comparing it against OS would false-positive on a legitimate configuration.</p>
+     *
      * @return {@code true} when ES and OS point to distinct clusters (safe to create OS indices);
      *         {@code false} when they overlap (caller must skip the OS bootstrap and halt the migration)
      */
     public static boolean endpointsAreSeparate() {
+        if (IndexConfigHelper.isMigrationComplete()) {
+            Logger.info(IndexStartupValidator.class,
+                    "Endpoint separation check skipped: PHASE_3_OPENSEARCH_ONLY"
+                    + " — ES is decommissioned and ES_ENDPOINTS is not required.");
+            return true;
+        }
         final OSClientConfig config;
         try {
             config = ConfigurableOpenSearchProvider.configFromProperties();
--- a/dotCMS/src/main/java/com/dotcms/content/index/opensearch/IndexStartupValidator.java
+++ b/dotCMS/src/main/java/com/dotcms/content/index/opensearch/IndexStartupValidator.java
@@ -204,23 +204,13 @@ public class IndexStartupValidator {
     // -------------------------------------------------------------------------
 
     /**
-     * Resolves the effective ES and OS endpoint sets and asserts they do not share
-     * any {@code host:port} pair.
-     *
-     * <p>The OS side comes from the already-resolved {@link OSClientConfig#endpoints()} (the very
-     * endpoints the client is built with), and the ES side is read from config; both are passed
-     * through the same {@link #normalizeEndpoint} path so the comparison is consistent.</p>
-     *
-     * <p><strong>Best-effort:</strong> two configs that refer to the same host via different forms
-     * (e.g. {@code "127.0.0.1"} vs {@code "localhost"}) will not be detected as overlapping.</p>
-     *
-     * @throws DotRuntimeException if at least one endpoint is common to both clients
-     */
-    /**
      * Asserts that the ES and OS clients do not share any {@code host:port}. Shared by the
      * startup validator ({@link #validate()}) and the config-only OS index-creation gate
      * ({@link #endpointsAreSeparate()}) so both enforce separation with identical logic.
      * On success it logs the compared sets so the "passed" line is proof of what was compared.
+     *
+     * <p><strong>Best-effort:</strong> two configs that refer to the same host via different forms
+     * (e.g. {@code "127.0.0.1"} vs {@code "localhost"}) will not be detected as overlapping.</p>
      *
      * @throws DotRuntimeException if at least one endpoint is common to both clients
      */
--- a/dotCMS/src/main/java/com/dotcms/content/index/opensearch/IndexStartupValidator.java
+++ b/dotCMS/src/main/java/com/dotcms/content/index/opensearch/IndexStartupValidator.java
@@ -258,12 +258,23 @@ public class IndexStartupValidator {
      *         {@code false} when they overlap (caller must skip the OS bootstrap and halt the migration)
      */
     public static boolean endpointsAreSeparate() {
+        final OSClientConfig config;
         try {
-            assertEndpointsSeparate(ConfigurableOpenSearchProvider.configFromProperties());
-            return true;
+            config = ConfigurableOpenSearchProvider.configFromProperties();
         } catch (Exception e) {
+            // Config could not be resolved — this is NOT an ES/OS overlap. Log it distinctly so the
+            // operator is not sent chasing a same-endpoint misconfiguration that does not exist.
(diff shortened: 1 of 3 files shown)


dotCMS/src/main/java/com/dotcms/content/index/opensearch/IndexStartupValidator.java
```diff
@@ -232,8 +244,39 @@ private void validateEndpointSeparation(final OSClientConfig config) {
...
+        final OSClientConfig config;
+        try {
+            config = ConfigurableOpenSearchProvider.configFromProperties();
+        } catch (Exception e) {
+            // Config could not be resolved — this is NOT an ES/OS overlap. Log it distinctly so the
+            // operator is not sent chasing a same-endpoint misconfiguration that does not exist.
+            Logger.error(IndexStartupValidator.class,
+                    "Cannot resolve OpenSearch configuration for the endpoint-separation check: "
+                    + e.getMessage());
+            return false;
```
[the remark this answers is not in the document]
swicken: Thanks — I looked at this closely and I believe it is already addressed / a false positive:

1. **Cause is no longer conflated.** The two branches log distinct messages: config-resolution failure logs `"Cannot resolve OpenSearch configuration for the endpoint-separation check: ..."`, while a true overlap logs `"...point to the same endpoint(s)... Set OS_ENDPOINTS to a separate OpenSearch instance."` The caller WARN also defers to the preceding error rather than asserting an overlap. (This was the earlier review round's footgun; fixed in 4224e83d95.)

2. **Fail-closed on an unresolvable config is intended, not a bug.** There is no scenario where we should create OS indices from a config we cannot even resolve. Returning `false` → skip OS bootstrap and fall back to ES-only (Phase 1/2) or throw (Phase 3) is the safe outcome; letting the exception propagate would crash bootstrap, which contradicts the migration's explicit "OS problems fall back to ES-only" design.

3. **Consistent with the sibling guard.** `validateIndexingConfig()` deliberately does a broad `catch (Exception)` and funnels every failure — version mismatch, unreachable, and config-resolution — through one halt-and-fallback path. This gate mirrors that on purpose.

Config-failure and overlap do share the same *outcome* (skip + halt/throw), but that convergence is desirable (both mean "OS is not usable, do not dual-write"), and the *causes* are now logged distinctly. Leaving the code as-is.

dotCMS/src/test/java/com/dotcms/content/index/opensearch/IndexStartupValidatorTest.java
```diff
@@ -0,0 +1,108 @@
...
+
+import com.dotmarketing.exception.DotRuntimeException;
+import com.dotmarketing.util.Config;
+import org.junit.After;
+import org.junit.Test;
+
+/**
+ * Unit tests for {@link IndexStartupValidator} endpoint-separation logic.
+ *
+ * <p>Covers the config-only separation gate ({@link IndexStartupValidator#endpointsAreSeparate()})
```
fabrizzio-dotCMS, line 16: This validation must be phase-aware and consider that during Phase 3, ES_ENDPOINTS isn't  required

fabrizzio-dotCMS (changes requested):

dotCMS/src/test/java/com/dotcms/content/index/opensearch/IndexStartupValidatorTest.java
fabrizzio-dotCMS: Then I guess that if we're on phase 3, we shouldn't run this check at all

swicken: Done in 456799422f — the separation check no longer runs in Phase 3. `endpointsAreSeparate()` and `validate()` both skip it (with an info log) when the migration is complete, since ES_ENDPOINTS isn't required there and the resolved ES side would just be the `localhost:9200` default — a false positive against a legitimate OS config. The Phase-3 throw branch in `bootstrapAndPointOS` is gone with it (unreachable now), and there's a Phase-3 skip test.

fabrizzio-dotCMS (approved):

Issue #36219, opened 2026-06-18
[QA-G13] OpenSearch Migration — Divergent Index Names & Fan-out (TC-041–TC-042)


fabrizzio-dotCMS:
## Parent Epic

#35476

## Glossary (quick)

- **Old engine (ES)** — Elasticsearch, what dotCMS uses today. Port `9200`, inspected with **Kibana** (`5601`).
- **New engine (OS)** — OpenSearch, the migration target. Port `9201`, inspected with **OpenSearch Dashboards** (`5602`).
- **Migration phase** — `0` = old engine only · `1` = write to *both*, search the old one · `2` = write to both, read OS.
- **Dual-write** — in Phase 1 every content change is written to *both* engines.
- **`.os` suffix (the "tag")** — marks a new-engine index. New-engine index names **end in `.os`** (e.g. `cluster_08abc3.working_20260406.os`); old-engine names do not. Visible everywhere: cluster, DB, REST. (NOT an `os::` prefix — that's an obsolete model.)

## Environment

```bash
docker compose -f docker/docker-compose-examples/os-migration/docker-compose.yml up -d
```

| Service | URL | Used for |
|---|---|---|
| Old engine (ES 7.10) | http://localhost:9200 | Direct REST queries |
| Kibana | http://localhost:5601 | Inspect/query the OLD index |
| New engine (OS 3.x) | http://localhost:9201 | Direct REST queries |
| OpenSearch Dashboards | http://localhost:5602 | Inspect/query the NEW index |
| dotCMS | http://localhost:8082 | Admin UI + REST API (`admin:admin`) |

Migration settings live in `dotmarketing-config.properties` (`FEATURE_FLAG_OPEN_SEARCH_PHASE`, `OS_ENDPOINTS=http://localhost:9201`, `OS_AUTH_TYPE=BASIC`, `OS_AUTH_BASIC_USER=admin`, `OS_AUTH_BASIC_PASSWORD=admin`, `OS_TLS_ENABLED=false`).

## Test Cases

> The migration doc flags this area as **unresolved — test coverage needed**. These cases document what actually happens today so the team can decide the intended behavior; **expect to file findings as bugs rather than simple pass/fail.**

### TC-041 — Operating with an index name that exists in only one engine
- **Objective:** In a real catchup deployment ES and OS hold indexes with **different timestamps** (and the `.os` tag). Observe how dotCMS behaves when an operation is driven by a name that exists in one engine but not the other (the doc lists this as untested).
[...]

Issue #35640, opened 2026-05-08
[QA-G6] OpenSearch Migration — Index Delete: ES vs OS Behavior (TC-016–TC-020)


fabrizzio-dotCMS:
## Parent Epic

#35476

## Glossary (quick)

- **Old engine (ES)** — Elasticsearch, what dotCMS uses today. Port `9200`, inspected with **Kibana** (`5601`).
- **New engine (OS)** — OpenSearch, the migration target. Port `9201`, inspected with **OpenSearch Dashboards** (`5602`).
- **Migration phase** — `0` = old engine only · `1` = write to *both*, search the old one · `2` = write to both, read OS.
- **Dual-write** — in Phase 1 every content change is written to *both* engines.
- **`.os` suffix (the "tag")** — marks a new-engine index. New-engine index names **end in `.os`** (e.g. `cluster_08abc3.working_20260406.os`); old-engine names do not. Visible everywhere: cluster, DB, REST. (NOT an `os::` prefix — that's an obsolete model.)

## Environment

```bash
docker compose -f docker/docker-compose-examples/os-migration/docker-compose.yml up -d
```

| Service | URL | Used for |
|---|---|---|
| Old engine (ES 7.10) | http://localhost:9200 | Direct REST queries |
| Kibana | http://localhost:5601 | Inspect/query the OLD index |
| New engine (OS 3.x) | http://localhost:9201 | Direct REST queries |
| OpenSearch Dashboards | http://localhost:5602 | Inspect/query the NEW index |
| dotCMS | http://localhost:8082 | Admin UI + REST API (`admin:admin`) |

Migration settings live in `dotmarketing-config.properties` (`FEATURE_FLAG_OPEN_SEARCH_PHASE`, `OS_ENDPOINTS=http://localhost:9201`, `OS_AUTH_TYPE=BASIC`, `OS_AUTH_BASIC_USER=admin`, `OS_AUTH_BASIC_PASSWORD=admin`, `OS_TLS_ENABLED=false`).

## Helpers (referenced by the steps below)

**H1 — Find your real index names.** Names contain a timestamp (differ per machine). **New-engine indexes end in `.os`**; old-engine names do not.
```bash
curl -s -u admin:admin http://localhost:8082/api/v1/esindex | jq .   # what dotCMS knows (keeps .os, drops cluster prefix)
curl -s -u admin:admin http://localhost:9200/_cat/indices?v          # OLD engine physical names (no .os)
[...]

PR #36432, opened 2026-07-06
fix(os-index): surface primary index-delete failures instead of silent 200

### LINKED ISSUE #36430: OpenSearch: esindex DELETE reports success without checking the result — migration fan-out swallows primary delete failures that previously surfaced as errors
## Problem Statement

During QA of the ES→OpenSearch migration (surfaced by #36219, test case TC-041; part of epic #35476; independently verified 2026-07-06), `DELETE /api/v1/esindex/{indexName}` reports success without ever checking whether the delete succeeded — and the migration's fan-out delete made this worse by swallowing failures that previously surfaced as 5xx.

Two layers combine:

1. **REST layer (long-standing wart, since 2020 / #18584):** `ESIndexResource.deleteIndex` (~L393) calls `idxApi.delete(indexName)` and discards the returned boolean, then unconditionally sends the "Index:<name> deleted" admin message and returns 200.
2. **Migration regression (#35863):** pre-migration, `ContentletIndexAPIImpl.delete()` was `return indexAPI.delete(indexName)` and `ESIndexAPI.deleteMultiple` **throws** on failure — so real delete failures still propagated to the caller as exceptions/5xx despite the ignored boolean. The fan-out rewrite catches every provider's exception (including the **primary's**), converts it to `primaryResult=false`, and returns it. This is stricter than the codebase's own `PhaseRouter.writeBoolean` contract, which rethrows primary failures and only swallows shadow failures. Combined with (1), a failed delete is now fully silent: no exception, ignored boolean, success toast, HTTP 200.

Observed live (Phase 1): deleting a `.os`-only name — primary (ES) delete throws `index_not_found_exception`, `primaryResult=false`, caller gets HTTP 200 + "Index:live_<ts>.os deleted". In that everyday divergent-name case the harm is limited (the OS copy the user saw is actually gone), but the REST layer's success report is unverified in all cases.

Note: the worst-case scenario "new engine down + OS-only name → silent 200" does **not** occur — verified live it returns 404, because `OSIndexAPIImpl.listIndices` swallows the connection error and returns empty, so the name drops out of the pre-check's merged index list. The defect is the unchecked success report and the swallowed primary failure, not a both-engines-fail 200.

Related: #36423 (same fan-out's ERROR-noise on expected misses — different fix site: log policy vs. error propagation).

## Steps to Reproduce

Rig: `docker/docker-compose-examples/single-node-os-migration/` (old engine = OpenSearch 1.3.20 `:9200`, new engine = OpenSearch 3.4.0 `:9201`), `DOT_FEATURE_FLAG_OPEN_SEARCH_PHASE=1`.

1. Phase 0: demo content indexes into ES → `working/live_<tsA>`. Switch to Phase 1 (catchup creates `.os` pair), full reindex → active `<tsB>` pair in both engines.
2. Create a `.os`-only name: delete `cluster….live_<tsA>` directly on the **ES** cluster, leaving `cluster….live_<tsA>.os` in OS only.
3. `DELETE /api/v1/esindex/live_<tsA>.os` as admin.

**Observed:** HTTP 200 + admin toast "Index:live_<tsA>.os deleted", while the primary (ES) provider's delete failed:
```
15:06:39.477  ERROR business.ContentletIndexAPIImpl - Error while deleting index cluster_….live_<tsA>.os
org.elasticsearch.ElasticsearchException: [type=index_not_found_exception, reason=no such index [cluster_….live_<tsA>.os]]
	at …ESIndexAPI.deleteMultiple(ESIndexAPI.java:259)
	at …ContentletIndexAPIImpl.delete(ContentletIndexAPIImpl.java:1683)
	at …ESIndexResource.deleteIndex(ESIndexResource.java:393)
```
`ContentletIndexAPIImpl.delete()` returned `false`; the resource ignored it.

**Pre-migration behavior for comparison:** the same primary failure propagated as an exception out of `delete()` → error response, not a success toast.

## Acceptance Criteria

- [ ] `DELETE /api/v1/esindex/{indexName}` does not report success ("deleted" message / 2xx with success payload) when the delete did not succeed on the authoritative (primary) provider.
- [ ] `ContentletIndexAPIImpl.delete()` follows the same error contract as `PhaseRouter.writeBoolean`: shadow failures are fire-and-forget, primary failures are surfaced to the caller (rethrown or reflected in the result the caller actually checks).
- [ ] Successful deletes (including paired bare-name deletes where only the shadow side misses) still return 2xx.

## dotCMS Version

`dotcms/dotcms:latest` (migration code confirmed in release `26.06.30-01`); observed 2026-07-03, independently re-verified 2026-07-06 on the `single-node-os-migration` compose rig.

## Severity

Low - Minor issue or cosmetic

## Links

- QA ticket: #36219 (TC-041) · Epic: #35476 · Related: #36423 · Regression source: #35863 · Pre-existing REST wart: #18584
- NA (no Freshdesk ticket)

swicken:
### Proposed Changes

* `ContentletIndexAPIImpl.delete()` now **re-throws primary-provider failures** after all providers have been called, matching the `PhaseRouter.writeBoolean` contract (shadow failures remain fire-and-forget). Pre-migration, `delete()` was a straight `return indexAPI.delete(indexName)` and `ESIndexAPI.deleteMultiple` throws on failure — so real failures propagated to the caller. The dual-write fan-out introduced in #35863 caught every provider's exception, including the primary's, making failed deletes fully silent. The re-throw is folded into the **bidirectional broadcast `delete()` landed in #36399** (run all legs, clear the `indicies` DB pointers, then surface the primary failure).
* A primary `index_not_found` is treated as an **idempotent no-op** (`isIndexNotFound()` helper, vendor-neutral match): the delete goal — index absent — is met, so cleaning up a dangling DB pointer or a divergent name held only by the shadow engine does not surface as a failure.
* `ESIndexResource.deleteIndex` now **checks the delete result**: a thrown primary failure or an unacknowledged delete returns 500 with an ERROR admin message instead of unconditionally reporting "Index:<name> deleted" with HTTP 200. The active/building-index `DotStateException` stays a readable 400 (#35640, TC-018).

Surfaced by QA #36219 (TC-041, epic #35476): deleting a `.os`-only index name in Phase 1 failed on the primary (ES) provider yet returned 200 + a success toast. The REST boolean discard itself is long-standing (#18584), but the swallowed primary exception is a migration regression.

**Related: #36431** (log-noise side of the same fan-out; independent branch). Both PRs patch `delete()` on top of #36399's broadcast, so whichever merges second gets a conflict in `delete()` and the migration integration test — resolve by keeping **#36431's shadow exists-check + shadow-write log policy** and **this PR's `primaryEx` re-throw + `isIndexNotFound` no-op** together.

Note: the "both engines fail → still 200" worst case was tested and does NOT occur — with the OS engine down, the endpoint's index-list guard 404s the request before the fan-out.

### Checklist

* Unit tests: `ContentletIndexAPIImplDeletePropagationTest` (5 tests — primary rethrow with shadow still called, shadow swallow, unacknowledged-false, Phase 0 propagation).
* Integration test (in `OpenSearchUpgradeSuite` via `ContentletIndexAPIImplMigrationIntegrationTest`): delete of a name existing in neither engine propagates the primary failure. Suite run green against the `opensearch-upgrade` container.

Fixes #36430

This PR fixes: #36430


swicken:
Review finding addressed in 84e5f1578e: `delete()` now throws for the primary-failure case (this PR's own contract change), which bypassed the boolean false-branch — so the ERROR admin toast never fired and the response came from the generic exception mapper. `deleteIndex` now catches the throw, logs it, sends the same ERROR admin message, and returns 500; the false-branch continues to cover the unacknowledged-no-exception case.

fabrizzio-dotCMS:
👋 @swicken — same coordination note as on #36431: this PR patches `ContentletIndexAPIImpl.delete()` and `ESIndexResource.deleteIndex()`, which **#36399** (issue #35640) fully rewrites. Flagging to avoid parallel conflicting merges on the same method.

#36399's `delete()` is a superset (active-index guard, `indicies` DB-pointer cleanup, bidirectional cascade behind `FEATURE_FLAG_INDEX_DELETE_CASCADE`, resilient fan-out, name normalization, `errors`-array responses). Because of that, our proposal is **not** to merge the `delete()` changes from this PR — #36399 supersedes them.

On the substance of this PR (surface a failed primary delete instead of a silent 200): it's a valid point and #36399 has the same latent gap today (it swallows provider failures and the REST layer ignores the boolean). We'll reconcile the failure-surfacing **inside #36399** so it fits its resilient + always-clean-DB design (run all legs, clean DB, then surface a primary failure to the caller) rather than merging a second version of `delete()`.

Note this PR is stacked/related to #36431, and both are based on the pre-#36399 `delete()`, so they'll conflict once #36399 lands. Let's sync on sequencing (and the `.os` routing decision raised on #36431) before merging any of the three.

fabrizzio-dotCMS (approved):

Issue #36384, opened 2026-07-01
[Backend] API support for Content Drive field-based search/filter criteria


zJaaal:
## Description

**Placeholder — to be fleshed out after the field-scope discussion on the frontend issue (#36383).**

The frontend work in #36383 adds field-based search/filter criteria to Content Drive, reusing the `libs/edit-content` field components. This backend issue tracks the API support required to drive that UI.

**Goal:**
1. Expose, per content type, which fields are searchable — i.e. fields flagged **User Searchable** (`field.isSearchable()`) **and** **System Indexed** (`field.isIndexed()`) — equivalent to the legacy `StructureAjax.getSearchableStructureFields()`.
2. Support querying contentlets by those field values from Content Drive, equivalent to the legacy `ContentletAjax` search behavior (text contains, exact/terms match, date ranges, tag/relationship/category lookups, etc.).

**Backend limitations will determine the final field set.** The candidate field types and their expected query semantics are enumerated in the frontend issue (#36383). Not all candidates may be feasible/performant to query — that determination happens here and feeds back into #36383's scope.

## Acceptance Criteria

_To be refined after the scope discussion._

- [ ] Scope of supported field types agreed with frontend (#36383)
- [ ] Endpoint returns the searchable fields (`isSearchable && isIndexed`) for a given content type
- [ ] Contentlet query supports filtering by the agreed field types with correct query semantics (contains / exact / terms / range)
- [ ] Behavior parity documented against the legacy portlet where applicable

## Additional Context

Legacy references:
- `StructureAjax.getSearchableStructureFields()` — filters fields to `isSearchable() && isIndexed()`.
- `ContentletAjax` + `view_contentlets_js_inc.jsp` `renderSearchField()` — legacy per-type search widget mapping.

Companion frontend issue: #36383

zJaaal:
# QA: Not Needed

QA will be handled in this ticket https://github.com/dotCMS/core/issues/36383

PR #36452, opened 2026-07-07
feat(content-drive): field-based search/filter criteria reusing edit-content field components (#36383)

### LINKED ISSUE #36383: [TASK] Content Drive: field-based search/filter criteria reusing edit-content field components
## Description

## Candidate fields

|---|---|---|---|---|

### Out of scope

- [x] v1 field set confirmed

## Acceptance Criteria

## Additional Context

Reference — legacy behavior:

zJaaal:
## Proposed Changes

Brings the legacy "View Contentlets" field-based filtering forward into **Content Drive**. Adds a **"More"** button at the end of the filter toolbar that lists the selected content type's **User-Searchable + System-Indexed** fields; picking one adds a dynamic filter chip whose control matches the field type. Values are sent to `POST /api/v1/drive/search` under a new `userSearchable` object keyed by field variable, and round-trip through the URL like the other filters.

Fixes #36383.

### How it works
- **"More" button** — enabled only when **exactly one** content type is selected (tooltip explains the requirement otherwise). Lists eligible fields; the **title** field and **Host-Folder** are excluded (title is covered by the keyword search; host/folder is redundant with the folder nav).
- **Per-field controls** (chip + popover, reusing the existing filter pattern):
  | Field type | Control | `userSearchable` value |
  |---|---|---|
  | Text / Textarea / WYSIWYG | text input (debounced; numeric/decimal input mode by data type) | `string` |
  | Select / Radio | single-select listbox | `string` |
  | Multi-Select / Checkbox (multi) | multi-select listbox | `string[]` |
  | Checkbox (binary) | True/False listbox | `boolean` |
  | Date / Date-and-Time / Time | inline range picker | `{ from, to }` (ISO-8601) |
  | Tag | infinite-scroll multi-select (`DotTagsService`) | tag names `string[]` |
  | Category | infinite-scroll multi-select (`DotCategoriesService`, scoped to the field root) | category inode `string[]` |
  | Relationship | reused `dot-select-existing-content` picker dialog, **single-select** | related contentlet identifier `string` |
- Added chips render in add-order with enter/leave animations; a chip's **X** clears its value (keeps the chip, like the built-in filters); the set resets when the selected content type changes.
- The popover content is created only while open, so overlay-hosted controls measure real dimensions instead of rendering empty.
- Exposed the relationship picker (`DotSelectExistingContentComponent` + footer + helper utils) and `ExistingContentStore` from the `@dotcms/edit-content` public barrel for reuse.

### Tag / Category — shared infinite-scroll multi-select
Tag and Category lists are unbounded, so they're browsed through a new shared presentational component, **`dot-content-drive-lazy-multiselect`** (search + PrimeNG virtual-scroll lazy load, `signalState`-backed paging). This drops the previous fixed 1000-item fetch cap — pages load on demand as the user scrolls. "Select all" is disabled since it could only ever cover the loaded page.

### Relationship — single value, resolved on reload
Relationship is single-select (backend supports one related value). The persisted value is the **identifier** (version-stable; matches how the edit-content relationship field and the `tree` table store relationships). On a cold URL restore the chip only holds the identifier, so the contentlet is resolved once via the `/content/{id}` endpoint (which accepts an identifier) and cached — this fills the chip's title and lets the picker preselect the correct row by its **inode** (the picker keys selection by inode; passing the identifier would match every version of the same content). The shared picker is left untouched.

### Tests
`functions` helpers, store methods + `$request` reshaping + URL restore, the field-filter chip (control per type incl. Tag/Category → lazy multi-select, value paths, relationship reload resolve + title), the shared `dot-content-drive-lazy-multiselect` (init load, debounced search, infinite scroll, selection), the "More" menu (gating, eligibility, add), and the relationship footer. Full portlet suite: **27 suites / 736 tests** green; lint clean.

### ⚠️ Depends on backend contract (#36384 — [#36463](https://github.com/dotCMS/core/pull/36463) / [#36474](https://github.com/dotCMS/core/pull/36474))
Frontend value choices, to confirm against the backend:
- **Relationship** → contentlet **identifier** (single). Note: the backend currently rejects Relationship fields with a 400 (deferred to v1.1 / PR 3) — coordinate sequencing.
- **Category** → category **inode**. ✅ matches backend ("Category inodes").
- **Tag** → tag **labels**. ✅ matches backend ("Tag names").
- **Multi-Select / Checkbox** assumed **OR**-combined within a field. The backend currently **AND**s these tokens (and OR-s Tag/Category) — flagged on [#36474](https://github.com/dotCMS/core/pull/36474#issuecomment-4918348335) to align, since two identical-looking chips would otherwise behave oppositely.

These are isolated to the store's `$request` reshaping if adjustments are needed.

This PR fixes: #36383


core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-field-filter/dot-content-drive-field-filter.component.ts
```diff
@@ -0,0 +1,655 @@
...
+                );
+            });
+    }
+
+    /** Fetches and caches the related contentlet by identifier (the `/content/{id}` endpoint accepts one). */
+    #resolveRelationshipItem(identifier: string): void {
+        this.#contentletService
+            .getContentletByInode(identifier)
+            .pipe(take(1), takeUntilDestroyed(this.#destroyRef))
+            .subscribe((contentlet) => {
```
oidacra, line 506: `#resolveRelationshipItem` subscribes to `getContentletByInode()` with no error handler, and that service is a bare `HttpClient.get` with no internal `catchError`. On a 404 (related item deleted since it was referenced), a 500, or a network failure — likely during a cold URL restore — the chip silently never resolves its title and nothing is logged. Consider adding `catchError((e) => { console.error('Failed to resolve relationship contentlet', identifier, e); return of(null); })` before `subscribe`.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-lazy-multiselect/dot-content-drive-lazy-multiselect.component.ts
```diff
@@ -0,0 +1,212 @@
...
+            page: this.$state.currentPage(),
+            perPage: PER_PAGE,
+            filter: this.$state.filter()
+        })
+            .pipe(
+                take(1),
+                takeUntil(this.#cancel$),
+                // A failed page must not leave the list spinning forever; stop loading and stop
+                // paging so the empty state shows instead of a permanent spinner.
+                catchError(() => {
```
oidacra, line 197: This `catchError(() => EMPTY)` stops the spinner, but the resulting empty state is identical to a genuine "no results" and nothing is logged. Is it worth surfacing to the user that the load actually failed (backend error / endpoint down) rather than showing the normal "no results" copy — e.g. a distinct `error` flag with a "Something went wrong loading options" message plus a log for ops? As is, a Tag/Category endpoint 500 or a network failure is indistinguishable from an empty search.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/dot-content-drive.store.ts
```diff
@@ -51,47 +61,64 @@ const initialState: DotContentDriveState = {
...
+                            },
+                            language: filters()?.languageId,
+                            contentTypes: filters()?.contentType,
+                            baseTypes: filters()?.baseType?.map(
+                                (baseType) => MAP_NUMBERS_TO_BASE_TYPES[Number(baseType)]
+                            ),
+                            workflow: filters()?.workflow?.length
+                                ? parseWorkflowFilter(filters()?.workflow)
+                                : undefined,
+                            userSearchable: buildUserSearchablePayload(
```
oidacra: Was the cold URL restore path verified for field filters? Tracing it: `initContentDrive` sets `filters` (including `us.*`) and `userSearchableActive` but not `userSearchableFields` (stays `[]`), so the first `loadItems()` computes `buildUserSearchablePayload(filters, [])`, which skips every `us.*` entry (no matching field) and sends `userSearchable: undefined` — field metadata only arrives later from the field-filter-menu's async `getContentType`. That would make refreshing a URL like `…;us.body:hello` issue an initial `/drive/search` without the filter, briefly showing unfiltered results before a corrected second request. If so, consider populating `userSearchableFields` synchronously during restore, or deferring the first `loadItems()` until fields are loaded when `us.*` filters are present.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/dot-content-drive.store.ts
```diff
@@ -51,47 +61,64 @@ const initialState: DotContentDriveState = {
...
+                                filters(),
+                                userSearchableFields()
+                            ),
+                            contentCursor: page.contentCursor ?? 0,
+                            folderCursor: page.folderCursor ?? 0,
+                            maxResults: paginationSignal?.limit,
+                            sortBy: sort()?.field + ':' + sort()?.order,
+                            archived: false,
+                            showFolders:
```
oidacra, line 105: Is it intentional that the new `userSearchable` field filter is left out of the `showFolders` gate? It currently gates on `!baseType && !contentType && !languageId && !workflow`, and every prior filter added to this store got a matching clause here (e.g. workflow in #35906). As is, applying only a field-based filter (no other filter active) still lists folders in the results, which differs from how the other filters behave. If that's not intended, consider adding a `!<userSearchable-active>` term.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-field-filter-menu/dot-content-drive-field-filter-menu.component.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+    const createComponent = createComponentFactory({
+        component: DotContentDriveFieldFilterMenuComponent,
+        providers: [
+            mockProvider(DotContentDriveStore, {
+                getFilterValue: jest.fn().mockReturnValue(undefined),
+                userSearchableFields: signal<DotCMSContentTypeField[]>([]),
+                userSearchableActive: signal<string[]>([]),
+                addUserSearchableField: jest.fn(),
+                setUserSearchableFields: jest.fn(),
+                clearUserSearchableFilters: jest.fn()
```
oidacra, line 63: Is the "clear active field filters when the content type changes" behavior covered anywhere? It's described as a core responsibility of this component and `clearUserSearchableFilters` is mocked here, but no test seems to drive a content-type transition to assert it fires (and doesn't fire on first load). Might be worth a test transitioning `contentType` (`['blog']` → `['news']`, and → `undefined`).

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/dot-content-drive-toolbar.component.spec.ts
```diff
@@ -59,7 +59,12 @@ describe('DotContentDriveToolbarComponent', () => {
                 clearFilters: jest.fn(),
                 filters: filtersSignal,
                 setDialog: jest.fn(),
-                selectedItems: selectedItemsSignal
+                selectedItems: selectedItemsSignal,
+                userSearchableFields: signal([]),
+                userSearchableActive: signal<string[]>([]),
+                setUserSearchableFields: jest.fn(),
```
oidacra, line 70: The store mock was extended for the new field-filter API, but I don't see tests covering the new `$activeFieldFilters` computed (resolving active variables against loaded fields and filtering out unresolved ones — the URL-restore case) or the new field-filter/field-filter-menu components in the template. Since this is the toolbar entry point for the feature, would it be worth adding a computed test (including the unresolved-variable branch) plus a render assertion via `byTestId`?

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-field-filter-menu/dot-content-drive-field-filter-menu.component.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+            }),
+            mockProvider(DotHttpErrorManagerService),
```
oidacra, line 65: `DotHttpErrorManagerService` is mocked here but I don't see a test that makes `getContentType` error, so the `catchError` → `handle(error)` → `of([])` branch (the component's only error handling) looks uncovered. Would it be worth a test where `getContentType` returns `throwError(...)`, asserting `httpErrorManager.handle` is called and `setUserSearchableFields([])` runs?

core-web/libs/portlets/dot-content-drive/portlet/src/lib/utils/functions.spec.ts
```diff
@@ -18,11 +19,28 @@ import {
...
+    parseUserSearchableValue,
+    serializeUserSearchableValue,
+    buildUserSearchablePayload,
+    getUserSearchableActive
 } from './functions';
 
 import { DotContentDriveFilters } from '../shared/models';
 
+/** Minimal content-type field fixture for the user-searchable helpers. */
+const field = (overrides: Partial<DotCMSContentTypeField> = {}): DotCMSContentTypeField =>
```
oidacra: This hand-rolled `field()` helper casts an object literal `as DotCMSContentTypeField`, and its base `dataType: 'TEXT'` isn't adjusted when `fieldType` is overridden to Tag/Checkbox/Date, so the `fieldType`/`dataType` combos can be incoherent. `@dotcms/utils-testing` already exports type-correct factories (`createFakeTagField`, `createFakeCheckboxField`, `createFakeDateField`, …) — would using those be preferable here (and in `dot-content-drive.store.spec.ts`, where the same helper is duplicated)?

core-web/libs/portlets/dot-content-drive/portlet/src/lib/utils/functions.ts
```diff
@@ -266,3 +289,193 @@ export function getFolderNodesByPath(
...
+export function serializeUserSearchableValue(
+    value: DotContentDriveUserSearchableValue | null | undefined,
+    fieldType: string
+): string {
+    if (value == null) {
+        return '';
+    }
+
+    if (isDateFieldFilterType(fieldType)) {
+        const range = value as DotContentDriveDateRange;
```
oidacra: `value` here is a `string | string[] | DotContentDriveDateRange | null | undefined` union and `fieldType` is an unrelated parameter, so `value as DotContentDriveDateRange` isn't validated — a mismatched `fieldType`/`value` pair would silently yield `range.from === undefined` rather than surfacing the mismatch. Would a small runtime guard (`typeof value === 'object' && !Array.isArray(value)`) or an `isDateRange` type guard be worth adding before the cast?

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-field-filter/dot-content-drive-field-filter.component.ts
zJaaal: Good catch — fixed in ec96ee32bc. `#resolveRelationshipItem` now has `catchError(() => of(null))`, so a 404/500/network failure keeps the chip's neutral count fallback instead of silently breaking. I kept it best-effort (no toast) since it's a background label resolve, not a user action.

<sub>— drafted by Claude Code on my behalf</sub>

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-lazy-multiselect/dot-content-drive-lazy-multiselect.component.ts
zJaaal: Addressed in ec96ee32bc. The lazy multiselect now sets a distinct `error` flag on load failure and renders a separate "Something went wrong loading options" message (`lazy-multiselect-error`) instead of the normal empty copy. I stopped short of injecting DotHttpErrorManagerService here to keep the component presentational — happy to add an ops log if you'd prefer.

<sub>— drafted by Claude Code on my behalf</sub>

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/dot-content-drive.store.ts
zJaaal: Confirmed and fixed in ec96ee32bc. Added a `userSearchableFieldsLoaded` flag; `loadItems` now holds the search while a restored `us.*` filter has no field metadata yet (read `untracked` so adding an empty chip doesn't re-fire a search). Once the menu loads the fields, `$request` changes and the search runs with the value shaped in. Covered by a new store test.

<sub>— drafted by Claude Code on my behalf</sub>

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/dot-content-drive.store.ts
zJaaal: Fixed in ec96ee32bc — `showFolders` now also gates on `!userSearchable`. You're right it was usually masked by `contentType` being present, but this makes it explicit and consistent. New $request test asserts it.

<sub>— drafted by Claude Code on my behalf</sub>

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-field-filter-menu/dot-content-drive-field-filter-menu.component.spec.ts
zJaaal: Added in ec96ee32bc: a test transitions `contentType` ['blog']→['news'] and asserts `clearUserSearchableFilters` fires on the change but not on first load.

<sub>— drafted by Claude Code on my behalf</sub>

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/dot-content-drive-toolbar.component.spec.ts
zJaaal: Added in ec96ee32bc: a toolbar test asserts `$activeFieldFilters` resolves active variables against loaded fields and drops the unresolved one (the URL-restore case), plus a `byTestId` render check that the resolved chip appears and the unresolved one doesn't.

<sub>— drafted by Claude Code on my behalf</sub>

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-field-filter-menu/dot-content-drive-field-filter-menu.component.spec.ts
zJaaal: Added in ec96ee32bc: a test makes `getContentType` return `throwError(...)` and asserts `httpErrorManager.handle` is called and `setUserSearchableFields([])` runs.

<sub>— drafted by Claude Code on my behalf</sub>

core-web/libs/portlets/dot-content-drive/portlet/src/lib/utils/functions.spec.ts
zJaaal: Done in ec96ee32bc — replaced the hand-rolled `field()` helper in both functions.spec and dot-content-drive.store.spec with the type-correct `createFake*Field` factories (`createFakeTextField`, `createFakeTagField`, `createFakeCheckboxField`, `createFakeSelectField`, `createFakeDateField`).

<sub>— drafted by Claude Code on my behalf</sub>

core-web/libs/portlets/dot-content-drive/portlet/src/lib/utils/functions.ts
zJaaal: Fixed in ec96ee32bc — added an `isDateRange` type guard (`typeof === 'object' && !Array.isArray`) before using the value, so a mismatched fieldType/value yields '' instead of a misleading partial range. New unit test covers a non-range value on a Date field.

<sub>— drafted by Claude Code on my behalf</sub>

oidacra (approved):

PR #36461, opened 2026-07-07
fix(edit-content): locale-aware compare and in-place version switching in History Tab

### LINKED ISSUE #35756: Compare feature shows default locale version instead of selected locale on the current side of the diff
## Problem Statement

In the new Edit Contentlet mode, when a user views a contentlet version in a **non-default locale** and uses the History Tab compare feature to compare an older version, the diff is incorrect: the **previous** side correctly shows the selected locale, but the **current** side shows the **default locale** version instead of the currently viewed locale.

This means the user is comparing two different locales against each other, which produces meaningless diffs and could mislead editors into thinking content is different when it isn't (or vice versa).

**Browser & OS:** N/A — behavior in the new Edit Contentlet mode (Angular frontend).

## Steps to Reproduce

1. Create a content type and activate the new Edit Contentlet mode.
2. Create a multilingual contentlet with a few saved versions.
3. Switch to a locale that is **not the default** (e.g., Spanish when English is the default).
4. Open the History Tab → Versions accordion and click **Compare** on an older version.
5. Observe: the **Previous** panel shows the selected (non-default) locale version correctly, but the **Current** panel shows the **default locale** version instead of the selected locale version.

**Expected:** Both the Previous and Current panels show versions in the **currently selected locale**.

https://github.com/user-attachments/assets/bce9758c-5283-4141-be33-7b6f0bf91577

## Acceptance Criteria

- [ ] When viewing a non-default locale version and clicking Compare, the "current" side of the diff shows the latest version in the **selected locale** — not the default locale
- [ ] The "previous" side shows the selected older version in the same selected locale
- [ ] Both sides of the compare diff always reflect the same locale context the user is currently viewing
- [ ] Original reproduction scenario (non-default locale → compare old version → current side shows default locale) no longer reproduces
- [ ] Compare in the default locale continues to work correctly (no regression)

## dotCMS Version

Latest from main branch

## Severity

## Links

Related: #35559 — History Tab fixes (compare order, return button)

### LINKED ISSUE #36375: Compare view: clicking another version exits to full-viewport instead of updating the comparison
### Problem Statement

In the new Edit Contentlet **compare** view, the list of versions is shown on the right. Clicking another previous version from that list **opens it full-viewport** (leaving compare mode) instead of updating the comparison.

The version-click behavior should depend on whether the user is currently in compare view:
- **In compare view** → clicking a version should **stay in compare** and update the comparison.
- **Not in compare view** → clicking a version should load it **full-viewport** (current behavior — keep it).

**Impact:** Users comparing versions are unexpectedly kicked out of compare mode when they try to compare against a different version, breaking the compare workflow.

**Browser & OS:** All (frontend behavior).

### Steps to Reproduce

1. Open a contentlet in the new Edit Contentlet with multiple versions.
2. Open the **History** and start a **compare** between two versions.
3. With the version list on the right, **click another previous version**.
4. **Observe:** the clicked version opens **full-viewport**, exiting compare view.
5. **Expected:** the comparison updates in place — the clicked version becomes the compared (older) version against the current version, staying in compare view.

### Acceptance Criteria

**In compare view**
- [x] While in compare view, clicking another previous version **keeps the compare layout** (does not navigate full-viewport)
- [x] The clicked version becomes the **older/compared** version; the **current-version** side stays fixed
- [x] The comparison content/diff refreshes to reflect the newly selected version
- [x] The active/selected state in the version list reflects the version now being compared

**Not in compare view**
- [x] When **not** in compare view, clicking a previous version loads it **full-viewport** (existing behavior is unchanged)

**Mode detection**
- [x] The behavior is driven by reliably detecting whether the user is currently in compare view at click time

### Suggested Tests

Frontend behavior/state branching (no backend change — versions load through existing endpoints), so coverage is **Unit-first** with a single E2E smoke. Integration/Postman are not applicable.

- [ ] When in **compare** mode, clicking a version does **not** trigger full-viewport navigation
- [ ] When in compare mode, clicking a version sets it as the **older/compared** version and leaves the current-version side unchanged
- [ ] The comparison diff/content refreshes to the newly selected version
- [ ] The version list's **active/selected** state updates to the compared version
- [ ] When **not** in compare mode, clicking a version triggers the **full-viewport** load (existing behavior)
- [ ] The compare-mode detection flag correctly drives the branch (covered for both true/false)

- [ ] Start a compare → click a different version in the list → the view **stays in compare** and the diff updates (does not go full-viewport). *(Single regression-critical flow; the both-modes branching is already unit-covered, so no broader E2E.)*

**Integration / Postman**
- [ ] No tests — version data loads through the **existing** content/version endpoints; there is no API contract change, only client-side routing/state logic.

### dotCMS Version

Latest from main branch (new Edit Contentlet)

### Severity

### Links

NA

oidacra:
## Summary

Fixes two issues in the History Tab of the new Edit Contentlet mode, plus a set of UX adjustments to the same flows that surfaced while verifying them.

### 1. Compare used the default locale on the "current" side (#35756)

Comparing versions while viewing a non-default locale produced a cross-locale diff: the previous side showed the selected locale, but the current side showed the default locale's working version.

Root cause: the `compareData` computed in the history feature store hardcoded `language: 'en-us'` as the version-map key passed to `DotContentCompareStore`. Since the compare store resolves the "current" (working) version from the version list of that language key, it always came back in the default locale.

The fix derives the key from the store's `currentLocale`, matching the backend `Language.getIsoCode()` key format used by `GET /api/v1/content/versions?groupByLang=1`: lowercase `languageCode-countryCode`, or bare `languageCode` for language-only locales. `DEFAULT_LOCALE_ISO_KEY` (`'en-us'`) remains only as the fallback when no locale is loaded yet.

Closes #35756

### 2. Clicking another version exited compare view (#36375)

While in compare view, clicking another version in the History list opened it full-viewport (leaving compare) instead of updating the comparison.

The `VIEW` action in `handleHistoryAction` now branches on the current view: in compare view, clicking a previous version updates the compared (older) side in place via `loadCompareVersionContent` (which also keeps the list's active state in sync through `historicalVersionInode`), and clicking the working version exits compare. Outside compare view the full-viewport load is unchanged.

Closes #36375

### UX adjustments to the same flows

- **No more skeleton collapse on reloads**: the history list stays visible while reloading; skeletons only show on the initial load and are now sized to match the real version cards.
- **Scoped reloads**: versions refetch only when `identifier` or `languageId` changes (viewing another version no longer refetches the list); push publish history refetches only when `identifier` changes and reads its own status, so it never flashes while versions load.
- **Inline loading feedback**: the store tracks `loadingVersionInode` while a clicked version is being fetched (view or compare), and the matching timeline card shows a spinner, cleared on success and error.
- **Compare exits when its context dies**: leaving the History sidebar tab exits compare view, and a locale/identifier change discards stale compare/historical state (a stale `originalContentlet` could otherwise restore the previous locale's content when exiting compare or historical view).

CleanShot 2026-07-07 at 16 32 39@2x

CleanShot 2026-07-07 at 16 32 25@2x

## Acceptance Criteria

#35756 — locale-aware compare

- [x] When viewing a non-default locale version and clicking Compare, the "current" side of the diff shows the latest version in the selected locale — not the default locale
- [x] The "previous" side shows the selected older version in the same selected locale
- [x] Both sides of the compare diff always reflect the same locale context the user is currently viewing
- [x] Original reproduction scenario (non-default locale, compare old version, current side shows default locale) no longer reproduces
- [x] Compare in the default locale continues to work correctly (no regression)

#36375 — version click while comparing

- [x] While in compare view, clicking another previous version keeps the compare layout (does not navigate full-viewport)
- [x] The clicked version becomes the older/compared version; the current-version side stays fixed
- [x] The comparison content/diff refreshes to reflect the newly selected version
- [x] The active/selected state in the version list reflects the version now being compared
- [x] When not in compare view, clicking a previous version loads it full-viewport (existing behavior unchanged)
- [x] The behavior is driven by the current view state (`uiState.view === 'compare'`) at click time

## Test Plan

New coverage in `history.feature.spec.ts`:

- `compareData` locale derivation: non-default locale (`es-es`), language-only locale (`it`), mixed-case `isoCode` normalization, null-locale fallback, default-locale regression
- `VIEW action while in compare view`: stays in compare and updates the compared side, active-state sync, exit via working version, full-viewport branch outside compare
- `loadingVersionInode`: set while fetching (view and compare), cleared on success and on error
- Scoped reload effect: no refetch on version-inode change, versions-only refetch on language change, compare/historical state discarded on locale change, compare exits when leaving the History sidebar tab

New coverage in component specs:

- Sidebar history: skeleton only on initial load (list stays during reloads), push publish loading driven by its own status, `loadingVersionInode` maps the spinner to the matching item only
- Timeline item: spinner shown/hidden by `isLoadingVersion`

Verification run: `pnpm nx lint edit-content`, `pnpm nx test edit-content` (104 suites, 1973 tests passing), `pnpm nx affected:lint` and `pnpm nx affected:test` against `origin/main` — all green.

## Changed Files

- `core-web/libs/edit-content/src/lib/store/features/history/history.feature.ts` — locale-aware compare key; compare-aware `VIEW` branching; `loadingVersionInode` lifecycle; scoped reload effect with compare-context cleanup; exit compare on tab leave
- `core-web/libs/edit-content/src/lib/store/edit-content.store.ts` — `loadingVersionInode` state field
- `core-web/libs/edit-content/src/lib/components/dot-edit-content-sidebar/dot-edit-content-sidebar.component.html` — wire `pushPublishStatus` and `loadingVersionInode` inputs
- `core-web/libs/edit-content/src/lib/components/dot-edit-content-sidebar/components/dot-edit-content-sidebar-history/dot-edit-content-sidebar-history.component.{ts,html}` — independent push publish status; initial-load-only skeletons; pass per-item loading flag
- `core-web/libs/edit-content/src/lib/components/dot-edit-content-sidebar/components/dot-edit-content-sidebar-history/components/dot-history-timeline-item/dot-history-timeline-item.component.{ts,html}` — inline loading spinner
- Matching `.spec.ts` files for all of the above

This PR fixes: #35756


nicobytes (commented):
Automated review pass — traced the two bug fixes and the reload-scoping logic directly against the PR branch's code (not just the diff) to verify correctness.

## Summary

Both root-cause fixes check out:

- **Locale-aware `compareData`** (`history.feature.ts`): the `language` key derivation (`isoCode` → `languageCode-countryCode` → `languageCode`, all lowercased) matches `DotLanguage`'s actual shape (`languageCode: string`, `countryCode?: string`, `isoCode?: string`) and the stated backend `Language.getIsoCode()` format. Traced all 5 spec cases (non-default locale, language-only locale, mixed-case isoCode, null-locale fallback, default-locale regression) against the implementation — all correct.
- **In-place version switching while comparing**: the new `isCompareView` branch in `handleHistoryAction`'s `VIEW` case correctly routes to `loadCompareVersionContent` (the same method `COMPARE` uses to enter compare mode), so there's no async window where both `loadVersionContent` and `loadCompareVersionContent` could race on the shared `loadingVersionInode` field — the view-state branching makes the two mutually exclusive by construction.
- **Scoped reload effect**: the `versionsKey`/`loadedPushPublishIdentifier` closures correctly gate reloads to identifier+language changes (versions) and identifier-only changes (push publish), and the `page === 1` short-circuit in the existing `loadVersions` accumulation logic (`isNewContent || page === 1`) means dropping the old `clearVersions()`/`clearPushPublishHistory()` calls doesn't reintroduce stale-accumulation on a fresh load — full replacement still happens on every page-1 response.
- **`HISTORY_SIDEBAR_TAB_INDEX = 1`** matches the actual tab order in `dot-edit-content-sidebar.component.html` (`[value]="1"` is the History tab) — confirmed against the current template rather than assumed.
- Both `loadVersionContent` and `loadCompareVersionContent` clear `loadingVersionInode` symmetrically on both the success and error paths.

## Minor notes (non-blocking)

- The tab-list padding tweak in `dot-edit-content-sidebar.component.html` (`[&_.p-tab]:flex-1` → `px-6`/`py-0!`) isn't mentioned in the PR description — looks like incidental cleanup rather than something worth calling out, but flagging for awareness in case it wasn't intentional.
- Test coverage for the compare-load error path only asserts `loadingVersionInode` clears; it doesn't also assert `compareContentlet`/`uiState.view` are left untouched (which the code does correctly — verified by reading `loadCompareVersionContent`'s error handler). Not blocking, just a coverage gap if a future refactor accidentally starts clearing that state on error.

Nice, well-scoped fix with test coverage that matches the implementation closely. Approve.

oidacra:
Re: the tab-list padding tweak — it is intentional. The icon tabs previously stretched to a third of the sidebar width each via flex-1, so the active-tab underline was much wider than the icon; they now hug their content. It rode along in this PR since it's the same sidebar the History work touches.

oidacra:
Re: the compare-load error-path coverage — done in 4a819606a9. Added a test asserting a failed comparison update clears loadingVersionInode while compareContentlet and the compare view stay untouched, so a failed fetch can't silently disturb the active compare session.

nicobytes (approved):

PR #36464, opened 2026-07-07
fix(edit-content): style comment Send/Clear buttons and add empty-message validation

### LINKED ISSUE #36372: Comment section: primary Send / tertiary Clear buttons with empty-message validation (Edit Contentlet)
### Description

Update the buttons and validation in the **Comment section** of the new Edit Contentlet.

- The **Send** button is styled as a **primary** button.
- The **Clear** button is styled as a **tertiary** button.
- Both buttons are **enabled at all times** (never disabled).
- If the user clicks **Send** with an **empty** message, show inline validation:
  - A **red border** on the comment input.
  - An **error message** below the input: **"A message is required."**

### Acceptance Criteria

**Button styling**
- [x] The **Send** button uses the **primary** button style
- [x] The **Clear** button uses the **tertiary** button style
- [x] Both **Send** and **Clear** are **enabled at all times**, regardless of input state

**Validation — empty message**
- [x] Clicking **Send** with an empty (or whitespace-only) comment does **not** submit
- [x] On that attempt, the comment input shows a **red border**
- [x] An error message **"A message is required."** is shown below the input
- [x] Once the user enters a non-empty message, the red border and error message are **cleared**
- [x] Clicking **Send** with a valid message submits the comment normally

**Clear button**
- [x] Clicking **Clear** empties the comment input
- [x] Clicking **Clear** also dismisses any active validation error/red border

### Suggested Tests

Frontend styling + client-side validation only (no backend/endpoint change), so coverage is **Unit-first** with a single E2E smoke. Integration/Postman are not applicable.

- [ ] The **Send** button renders with the **primary** style; **Clear** renders with the **tertiary** style
- [ ] Both buttons are **enabled** regardless of input state (empty and non-empty)
- [ ] Clicking **Send** with an empty / whitespace-only message does **not** emit a submit
- [ ] On empty Send, the input shows the **error/red-border** state and the message **"A message is required."**
- [ ] Entering a non-empty message **clears** the error/red-border state
- [ ] Clicking **Send** with a valid message **emits the submit** with the message
- [ ] Clicking **Clear** empties the input and dismisses any active error state

- [ ] In the Comments tab, click **Send** with no text → red border + "A message is required." appears → type a comment → error clears → Send posts the comment. *(Single flow; the styling/enabled-state and per-branch validation are already unit-covered, so no broader E2E.)*

**Integration / Postman**
- [ ] No tests — this is button styling and client-side validation; the comment-submit endpoint and its contract are unchanged. *(If the empty-message rule is also enforced server-side as part of this work, add a Postman check that a blank comment is rejected.)*

### Priority

Medium

### Additional Context

Part of the new Edit Contentlet work (`dotCMS: New Edit Contentlet`), Comments section. Buttons stay enabled so the validation is driven by the Send action rather than a disabled state. Error copy is **"A message is required."** (final wording).

jcastro-dotcms:
## Summary

Closes #36372.

Updates the Comments section of the new Edit Contentlet sidebar to match the requested button styling and adds client-side validation for empty comments.

- **Send** is now a primary button; **Clear** is a tertiary (text/ghost) button — matching the pattern already used for de-emphasized secondary actions elsewhere in the app.
- Both buttons stay **enabled at all times** regardless of textarea content. They're still disabled while a save request is in flight, to prevent double-submits — that's request state, not input state, so it isn't affected by the "enabled at all times" requirement.
- Clicking **Send** with an empty or whitespace-only comment no longer submits: the textarea gets a red border and **"A message is required."** renders below it. This clears as soon as the user types a non-empty value or clicks **Clear**.
- Added a new `commentRequired` validation error key (instead of reusing Angular's built-in `required`) so this field shows our specific copy instead of the shared, generic "This field is required" string used elsewhere in the app.
- Fixed a latent ordering bug in `onSubmit()`: `markAsDirty()`/`markAsTouched()` must run *before* `setErrors()`, since the shared `dot-field-validation-message` component recomputes its visibility synchronously off the `setErrors()` emission. With the old order, the message could silently never render even though the field's error state was otherwise correct — this applied to the original `required` error too, it just wasn't previously covered by any test that checked the rendered text.
- Moved `dot-field-validation-message` out of absolute positioning into normal document flow so it no longer visually overlaps the character counter above it.

## Test plan

- [x] Updated/added unit tests in `dot-edit-content-sidebar-activities.component.spec.ts` (38/38 passing): button enabled/disabled states, required-error message content, error clearing on input/Clear.
- [x] Verified manually end-to-end in the running app (local Docker backend + `nx serve dotcms-ui`): button styling, empty-submit validation, error-clears-on-type, Clear dismisses the error, valid comment submits successfully.
- [x] `eslint` clean on all touched files.

This PR fixes: #36372


nicobytes:
The `commentRequired` custom error key works, but semantically it duplicates Angular's built-in `required` — both mean "this field is required," the only difference is the copy. We're using the *custom validator* path just to change the message for what is really a standard required error. A couple of alternatives that avoid introducing a competing error key:

**Option A — use the `message` input that `dot-field-validation-message` already exposes** (smallest change, nothing shared touched):

```ts
this.commentControl.setErrors({ required: true });
```
```html
<dot-field-validation-message
    [field]="form.get('comment')"
    [message]="'edit.content.sidebar.activities.required' | dm" />
```

In `getErrors()` the component returns `this.defaultMessage ? this.defaultMessage : ...`, so passing `message` overrides the generic copy while keeping Angular's semantic `required`. No new error key, no conceptual duplication.

Caveat: `message` applies to *any* error on the control. This field only validates `required` today, so it's fine — but if it later gained e.g. `maxlength`, it would show the same text for both.

**Option B — extend the shared component** to accept a per-key message override for built-in validators (a `requiredErrorMessage` input, mirroring the existing `patternErrorMessage`). This is the proper fix at the root, but it's a wider change since it touches all consumers of `dot-field-validation-message`.

My preference is Option A here: it gets the same field-specific copy through an API the component already offers, without adding an error key that overlaps with `required`.

jcastro-dotcms:
Good catch, thanks @nicobytes. Went with **Option B** (extend the shared component) rather than Option A as originally described — the literal snippet in Option A binds `message` statically to always show our copy, but `getErrors()` returns `defaultMessage` unconditionally whenever *any* error is present on the control (`return this.defaultMessage ? this.defaultMessage : ...`). Since this field's max-length validator returns `{ invalid: true }` (not a string), once a comment exceeds 500 chars while non-empty, `dirty && !valid` is true and the same "A message is required." text would render — which is wrong for that case. A conditional `message` binding avoids that, but introduces a timing issue: `errorMsg` is recomputed synchronously off the `field` input's `statusChanges` subscription inside `setErrors()`, which fires *before* Angular's own change-detection cycle re-evaluates the conditional `message` expression — so the override value is effectively one tick stale.

Implemented in `dd0f0fa`:
- Reverted to Angular's built-in `required` key (`this.commentControl.setErrors({ required: true })`) — no more semantic duplication.
- Added a `requiredErrorMessage` input to `DotFieldValidationMessageComponent`, mirroring the existing `patternErrorMessage` pattern — it only overrides the copy for a `required` error, leaving every other validator's message (and every other consumer of the component) untouched. It's a static value bound once, so it doesn't hit the timing issue above.
- Added test coverage in `dot-field-validation-message.component.spec.ts` for the new input, including a regression test confirming it doesn't leak into a `pattern` error on the same field.

All 45 tests across the two specs pass. Let me know if you'd rather see this scoped differently.

adrianjm-dotCMS (approved):

jcastro-dotcms:
Fixed the textarea shift (commit \`c091148\`).

Root cause: \`activities-footer\` is \`sticky bottom-0\`. \`dot-field-validation-message\` conditionally mounts/unmounts its \`<small>\` internally (\`@if\`), so the row's height changed whenever the error appeared or cleared — and because the footer is bottom-anchored, that height change pushed the textarea upward instead of just growing/shrinking downward.

Rather than changing the shared component's internal \`@if\` to always render (which would add persistent reserved space to all ~25 other usages of \`dot-field-validation-message\` across the app — login, content types, personas, push-publish, etc. — none of which asked for that), I scoped the fix to this component: wrapped the validation message in a `div.h-3.5` slot (`data-testid="activities-error-slot"`) that's always present with a fixed height, whether or not the error is showing. The footer's total height is now constant across all states, so there's no more movement.

Added a regression test asserting the slot exists and keeps its height class through the no-error → error → cleared cycle.

PR #36463, opened 2026-07-07
feat(content-drive): userSearchable field-filter contract, parse/validate + routing switch (#36384)

ihoffmann-dot:
### Proposed Changes

First slice (PR 1 of 3) of the Content Drive field-based search backend — issue #36384. Ships the plumbing only: request contract, parse/validation and routing switch. No DB/ES clause construction yet (that lands in PR 2).

* **Request contract** — `AbstractDriveRequestForm` gains `userSearchable` (`Map<String, Object>`, `@Value.Default` empty), a map keyed by field variable name, additive to the existing global `filters.text`.
* **Typed criteria model** — new `FieldSearchCriteria` (in `com.dotcms.browser` to avoid a `browser → rest.drive` dependency): resolved `Field`, `FilterKind` (SCALAR/RANGE/MULTI/BOOLEAN), `RoutingBucket` (DB/INDEX) and values.
* **Parse + validate** — new `ContentDriveFieldFilterResolver`: requires exactly one content type, rejects unknown keys and fields that are not `searchable && indexed`, rejects out-of-scope types (Relationship = v1.1; Binary, JSON, Key/Value, Block Editor, Constant, Hidden, Custom, dividers, File/Image, Host/Folder), infers the operator from the value shape and validates it against the field type. Failures raise `BadRequestException` (HTTP 400) naming the offending field.
* **Query model** — `BrowserQuery` carries `List<FieldSearchCriteria> fieldCriteria` (+ builder/getter/copy-ctor/toString). Legacy `BrowserQueryForm` (`/v1/browser`) untouched.
* **Routing switch** — `BrowserAPIImpl.isUseElasticSearchForTextFiltering` renamed to `isUseElasticSearchForFiltering`; now routes to the index when there is a text/fileName filter **or** any INDEX-routed field criterion. Tag (DB-routed) does not flip the switch. Default heuristic `HYBRID_SINGLE_CHUNKED_QUERY_ES` unchanged.
* **Wiring** — `ContentDriveHelper` parses `userSearchable`, sets the criteria on the builder and enables ES filtering when an index-routed criterion is present.

Follows the ADR-0018 database-first routing contract. Tag criteria are validated and carried but remain no-op in this PR (resolved DB-side in PR 2). The legacy Site Browser path is behaviorally unchanged.

### Checklist
- [ ] Tests (land with PR 2 — routing/validation integration coverage)
- [ ] Translations
[...]

PR #36474, opened 2026-07-08
feat(content-drive): field-filter resolver — Tag (DB) + index clauses via search strategies (#36384)

### LINKED ISSUE #36384: [Backend] API support for Content Drive field-based search/filter criteria

ihoffmann-dot:
### Proposed Changes

Second slice (PR 2 of 3) of the Content Drive field-based search backend — issue #36384. Builds on #36463 (contract + plumbing) and implements the actual resolution of `userSearchable` field filters into DB predicates and index clauses, per the ADR-0018 routing contract.

> **Stacked on [#36463](https://github.com/dotCMS/core/pull/36463).** Base branch is `36384-content-drive-field-search-contract`; review that one first. Once #36463 merges, this can be retargeted to `main`.

* **Index-routed clauses (reuse strategies)** — `BrowserAPIImpl.buildBaseESQuery` now appends the per-field Lucene clauses for index-routed criteria (Text/Textarea/WYSIWYG, Select/Radio, Multi-Select/Checkbox, Boolean, Date/Time/Date-Time, Category). The field-value → Lucene translation is delegated to the shared strategies in `com.dotcms.rest.api.v1.content.search` (`FieldHandlerRegistry` + `FieldContext`) so the syntax matches the modern content search — no hand-rolled Lucene. Clauses are injected into the existing hybrid per-chunk narrowing, so they AND with the text term and the DB candidate set.
* **DB-routed predicate: Tag** — `selectQuery` now appends a `tag`/`tag_inode` sub-select for Tag criteria, resolved DB-side regardless of ES filtering to preserve read-your-writes. Tag names within a field combine with OR; criteria across fields combine with AND. Category stays index-only (no SQL tree resolution).
* **Value formatting** — a small formatter maps each `FieldSearchCriteria` to the value string its strategy expects (scalar, comma-joined multi, `true`/`false`, `from TO to` range with `*` for open bounds).
* **Model** — `FieldSearchCriteria` now carries its `ContentType` (needed to build the `ct.field` Lucene name and provide strategy context).
* **Routing preserved** — index-routed criteria never touch SQL; Tag never touches the index; Category never hits SQL; no structural predicate is re-routed. Default heuristic stays `HYBRID_SINGLE_CHUNKED_QUERY_ES`.

### Checklist
- [x] Tests — new `ContentDriveFieldFilterTest` (integration): text→index, tag→DB (read-your-writes), tag OR within field, date range→index, index+DB AND composition, and the three 400 validations. **8/8 passing** locally against the integration environment (Postgres + OpenSearch).
- [ ] Translations
- [x] Security Implications Contemplated — Category resolution reuses `CategoryFieldStrategy`'s per-user permission filtering; all field values are validated against the content type schema (PR 1); tag/field values are bound as SQL parameters (no concatenation of user input into SQL).

### Additional Info

* `/v1/drive/search` is `@Hidden` → no `openapi.yaml` change.
* **Known semantic to confirm with FE:** Multi-Select/Checkbox "in list" reuses the shared `TextFieldStrategy`, which ANDs the tokens (matches the current content-search behavior) rather than OR-ing them. Category and Tag are OR-in-list. If FE expects OR for multi-select, we can special-case it — flagged rather than silently diverging from content-search.
* **Tag matching** is by name only (case-insensitive, exact), mirroring the existing text-filter tag join. Not scoped to `tag_inode.field_var_name` to avoid false negatives from legacy rows; a content type with two tag fields could cross-match — acceptable v1 limitation, easy to tighten later.
* Relationship remains v1.1 (rejected at parse time) — PR 3.

### Screenshots

N/A — backend only.

This PR fixes: #36384


zJaaal:
Re: the *"Known semantic to confirm with FE"* note on Multi-Select/Checkbox —

On the FE these render as a multi-pick listbox: the user ticks several values expecting **"match any of these" (OR)**. Two concerns with the current AND behavior (from reusing `TextFieldStrategy`):

1. **Inconsistency within the same feature** — Tag and Category are already OR-in-list, so a Multi-Select/Checkbox row that ANDs its tokens would behave the opposite way from a Tag/Category row that looks identical in the UI.
2. **AND rarely matches intent** for a filter — selecting `A` and `B` to mean "items that have both" is a narrow case; "items that have either" is the common expectation.

So the FE expectation is **OR** for Multi-Select/Checkbox. Could we special-case those two to OR the tokens rather than inheriting the text strategy's AND? If there's a reason to keep AND, happy to talk it through — just flagging before this merges since the behavior would diverge from what the UI implies.

<sub>Drafted by Claude Code on my behalf.</sub>

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/drive/ContentDriveFieldFilterTest.java
```diff
@@ -0,0 +1,317 @@
...
+ *   index.</li>
+ *   <li><b>Date-Time range → index</b>: {@code from/to} range.</li>
+ *   <li><b>AND composition</b>: an index-routed and a DB-routed criterion combine.</li>
+ *   <li><b>Validation</b>: unknown key, non-searchable key and more than one content type all
+ *   yield a {@link BadRequestException} (HTTP 400).</li>
+ * </ul>
+ */
+@ApplicationScoped
+@RunWith(DataProviderWeldRunner.class)
+public class ContentDriveFieldFilterTest extends IntegrationTestBase {
```
fabrizzio-dotCMS, line 71: 🔴 This IT isn't registered in any suite, so it won't run in CI. Failsafe in `dotcms-integration/pom.xml` only includes `MainSuite*/Junit5Suite*/QuickSuite/OpenSearchUpgradeSuite`; standalone classes aren't discovered. The sibling drive tests are wired into `MainSuite3a` (lines 70-71) — this one isn't, so the 8 "passing locally" tests never execute in the pipeline.

Fix: add `ContentDriveFieldFilterTest.class` to `MainSuite3a`. And since the Text/Date cases depend on the index, consider also adding it to `OpenSearchUpgradeSuite` so it's validated under the ES→OS migration matrix.

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
```diff
@@ -1163,13 +1168,97 @@ String buildBaseESQuery(final BrowserQuery browserQuery) {
...
+        }
+        final StringBuilder clauses = new StringBuilder();
+        for (final FieldSearchCriteria criteria : browserQuery.getFieldCriteria()) {
+            if (criteria.getBucket() != FieldSearchCriteria.RoutingBucket.INDEX) {
+                continue;
+            }
+            final Field field = criteria.getField();
+            final ContentType contentType = criteria.getContentType();
+            final String luceneFieldName = contentType.variable() + "." + field.variable();
+            final Function<FieldContext, String> handler = FieldHandlerRegistry.getHandler(field.type());
```
fabrizzio-dotCMS, line 1243: `getHandler` returns `getOrDefault(type, ctx -> BLANK)`, so an INDEX-routed field type without a registered handler silently produces no clause → the criterion returns everything unfiltered instead of failing. All 11 current index-routed types (PR 1's `isIndexRouted`) are registered, so this is correct today, but the coupling is silent. Consider a fail-fast/WARN when `bucket == INDEX` and the resolved handler yields BLANK for a non-empty value, so a future type added to `isIndexRouted` without a handler doesn't degrade to "returns everything".

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
```diff
@@ -1163,13 +1168,97 @@ String buildBaseESQuery(final BrowserQuery browserQuery) {
...
+            case BOOLEAN:
+                return String.valueOf(criteria.getBooleanValue());
+            case RANGE:
+                final String from = UtilMethods.isSet(criteria.getRangeFrom())
+                        ? criteria.getRangeFrom() : "*";
+                final String to = UtilMethods.isSet(criteria.getRangeTo())
+                        ? criteria.getRangeTo() : "*";
+                return from + " TO " + to;
+            case MULTI:
+                return String.join(",", criteria.getValues());
```
fabrizzio-dotCMS, line 1293: MULTI joins values by comma into `TextFieldStrategy`, which ANDs the tokens (the semantic you flagged). Recommend confirming the OR-vs-AND expectation with FE before they consume it. Related and not called out in the description: Boolean Checkbox also routes through the TEXT handler group with `"true"`/`"false"` — worth a quick confirm that it matches an indexed boolean checkbox correctly.

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/drive/ContentDriveFieldFilterTest.java
```diff
@@ -0,0 +1,317 @@
...
+                .setProperty(TEXT_VAR, "angular")
+                .setProperty(DATE_VAR, date(2026))
+                .folder(testFolder)
+                .nextPersisted();
+
+        Logger.info(ContentDriveFieldFilterTest.class,
+                "Field-filter test data ready under " + testAssetPath);
+    }
+
+    private static Date date(final int year) {
```
fabrizzio-dotCMS, line 221: `new Date(year - 1900, 0, 1)` is deprecated and JVM-local-timezone; the date-range boundary assertions could get flaky on a non-UTC CI agent. Prefer an explicit UTC `Instant`/ISO-8601 string.

fabrizzio-dotCMS (commented):
Reviewed the resolver slice (stacked on #36463). Clean implementation — reusing the shared `FieldStrategy` handlers instead of hand-rolling Lucene is the right call, and the INDEX↔SQL routing is respected. One blocking item (test not wired into CI, inline) plus coverage gaps below.

**Coverage gaps worth closing:**
- The new IT only exercises Text (index), Tag (DB), Date range and the 3 validations. Missing: Select/Radio SCALAR equals, **Multi-Select/Checkbox MULTI** (exactly the AND-vs-OR semantic you flagged — deserves a test pinning the current behavior), Boolean checkbox, and **Category** (the permission-filtered path the security note relies on). Also: open-ended range (`*` bound), out-of-scope type → 400, kind-mismatch → 400, Relationship → 400.
- **No test goes through JSON deserialization.** The IT builds `DriveRequestForm` via the builder, so `userSearchable` (`Map<String,Object>`) never passes through Jackson. `inferKind`'s `Boolean`/`Map`/`Iterable` branches are validated with Java types, not with what Jackson actually produces from a JSON body. A **Postman test** on `POST /v1/drive/search` would cover the real HTTP/JSON contract and the 400s — the existing `ContentDriveResource.postman_collection.json` currently has 0 references to `userSearchable`. Recommend adding one.
- A pure unit test for `ContentDriveFieldFilterResolver` (see the note on #36463) still applies.

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/drive/ContentDriveFieldFilterTest.java
ihoffmann-dot: Done! Added to MainSuite3a

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
ihoffmann-dot: Done! WARN when an INDEX field yields no clause

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
ihoffmann-dot: Done! Multi-Select/Checkbox now OR their values

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/drive/ContentDriveFieldFilterTest.java
ihoffmann-dot: Done!

ihoffmann-dot:
> Reviewed the resolver slice (stacked on #36463). Clean implementation — reusing the shared `FieldStrategy` handlers instead of hand-rolling Lucene is the right call, and the INDEX↔SQL routing is respected. One blocking item (test not wired into CI, inline) plus coverage gaps below.
> 
> **Coverage gaps worth closing:**
> 
> * The new IT only exercises Text (index), Tag (DB), Date range and the 3 validations. Missing: Select/Radio SCALAR equals, **Multi-Select/Checkbox MULTI** (exactly the AND-vs-OR semantic you flagged — deserves a test pinning the current behavior), Boolean checkbox, and **Category** (the permission-filtered path the security note relies on). Also: open-ended range (`*` bound), out-of-scope type → 400, kind-mismatch → 400, Relationship → 400.
> * **No test goes through JSON deserialization.** The IT builds `DriveRequestForm` via the builder, so `userSearchable` (`Map<String,Object>`) never passes through Jackson. `inferKind`'s `Boolean`/`Map`/`Iterable` branches are validated with Java types, not with what Jackson actually produces from a JSON body. A **Postman test** on `POST /v1/drive/search` would cover the real HTTP/JSON contract and the 400s — the existing `ContentDriveResource.postman_collection.json` currently has 0 references to `userSearchable`. Recommend adding one.
> * A pure unit test for `ContentDriveFieldFilterResolver` (see the note on #36463) still applies.

Done! ✅ Added ITs for Multi-Select (OR), Boolean, open-ended range, and kind-mismatch→400, plus unit tests for out-of-scope→400, Relationship→400, and a Jackson round-trip (so inferKind's Boolean/Map/Iterable branches are validated against real deserialized JSON, not hand-built types). Select-equals and Category behavioral ITs + the Postman test are the only remaining nice-to-haves.

ihoffmann-dot:
> Re: the _"Known semantic to confirm with FE"_ note on Multi-Select/Checkbox —
> 
> On the FE these render as a multi-pick listbox: the user ticks several values expecting **"match any of these" (OR)**. Two concerns with the current AND behavior (from reusing `TextFieldStrategy`):
> 
> 1. **Inconsistency within the same feature** — Tag and Category are already OR-in-list, so a Multi-Select/Checkbox row that ANDs its tokens would behave the opposite way from a Tag/Category row that looks identical in the UI.
> 2. **AND rarely matches intent** for a filter — selecting `A` and `B` to mean "items that have both" is a narrow case; "items that have either" is the common expectation.
> 
> So the FE expectation is **OR** for Multi-Select/Checkbox. Could we special-case those two to OR the tokens rather than inheriting the text strategy's AND? If there's a reason to keep AND, happy to talk it through — just flagging before this merges since the behavior would diverge from what the UI implies.
> 
> Drafted by Claude Code on my behalf.

Done! Multi-Select/Checkbox now use OR ✅

dario-daza (approved):

PR #36491, opened 2026-07-09
feat(edit-content): related-content navigation via agnostic EditContentHost (#36349)

### LINKED ISSUE #36349: Navigate between related content via breadcrumb banner (Edit Contentlet)
### Description

In the new Edit Contentlet there is currently **no way to navigate between related content**. In the old edit mode, clicking a related content opens that content's editor (see attached video). This issue restores that capability in the new mode using a **breadcrumb navigation banner**.

**Behavior:**
- When a content has relationships and the user clicks a related content (e.g. from a relationship field), the new mode opens **that** content's editor.
- A **banner under the title**, labeled **"Related content"**, shows the **navigation trail** as breadcrumbs.
- The trail **grows as the user navigates**: open Content A → click related B → trail shows `A › B` → click related C from B → `A › B › C`.
- **Each breadcrumb is clickable** and redirects to that specific content's editor (clicking an earlier crumb returns to it).
- Navigating to another content while the current one has **unsaved changes** triggers the existing **unsaved-changes confirmation** before leaving (consistent with #36347).

> 🎥 See the attached video (old-mode behavior reference) and 🖼️ the attached screenshot (target look in the new mode). *Note: attach these to the issue — they are the visual source of truth.*

### Acceptance Criteria

**Breadcrumb banner**
- [ ] A banner labeled **"Related content"** appears **under the title** when the user has navigated into related content
- [ ] The banner renders the navigation path as **breadcrumbs**
- [ ] The breadcrumb trail **grows** each time the user opens a related content (e.g. `A › B › C`)
- [ ] The current content is reflected as the **last/active** crumb in the trail

**Navigation**
- [ ] Clicking a related content (e.g. from a relationship field) opens **that content's** editor in the new mode
- [ ] Each breadcrumb is **clickable** and redirects to that specific content's editor
- [ ] Clicking an **earlier** breadcrumb returns to that content and trims the trail accordingly

**Unsaved changes**
- [ ] Navigating to a related content (via related-content click or a breadcrumb) while the current content has **unsaved changes** shows the existing unsaved-changes confirmation
- [ ] Choosing **discard** proceeds with navigation; choosing **keep editing** cancels navigation and stays on the current content
- [ ] With **no** unsaved changes, navigation proceeds immediately with no confirmation

**Parity & visuals**
- [ ] The navigation behavior matches the old edit mode (per the attached video)
- [ ] The banner matches the attached screenshot design

### Suggested Tests

Frontend navigation/state change, so coverage is **Unit-first** with a single focused E2E. Navigation reuses existing content-load endpoints, so Integration/Postman are not warranted. Unit tests are prioritized over E2E as they are cheaper and cover the trail/guard logic directly.

- [ ] The banner renders the **"Related content"** label only once a navigation trail exists
- [ ] Navigating into a related content **appends** a crumb to the trail (`A` → `A › B` → `A › B › C`)
- [ ] The **current** content is the last/active crumb
- [ ] Clicking a related content invokes navigation to **that** content's id
- [ ] Clicking an **earlier** breadcrumb navigates back to that content and **trims** the trail to that point
- [ ] Navigating with a **dirty** form triggers the unsaved-changes confirmation (both related-content click and breadcrumb paths)
- [ ] Confirming **discard** proceeds with navigation; **keep editing** cancels and the trail/current content are unchanged
- [ ] Navigating with a **clean** form proceeds immediately (no confirmation)

- [ ] Open a content with a relationship → click a related content → trail shows `A › B` → click the `A` crumb → returns to A and the trail trims. *(Single navigation round-trip; the per-state matrix and the dirty-navigate guard are already unit-covered, so no broader E2E is warranted.)*

**Integration / Postman**
- [ ] No tests — navigation loads each content through the **existing** content endpoints; there is no new API contract. *(If a dedicated endpoint to resolve the relationship trail is introduced during implementation, add Postman coverage for it then.)*

### Priority

Medium

### Additional Context

Part of the new Edit Contentlet work (`dotCMS: New Edit Contentlet`). This brings the related-content navigation of the legacy editor to the new Angular mode. The unsaved-changes guard should be the same one used by the dialog container (#36347). Related to the relationship field work in #36155.

adrianjm-dotCMS:
## Proposed Changes

Restores **navigating between related content** in the new Edit Contentlet and, to enable it cleanly, makes the editor **presentation-agnostic**.

Fixes #36349

### Related-content navigation (the feature)
- Clicking a related content's title opens it; a **"Relating content"** breadcrumb banner tracks the trail (grows/trims, click an earlier crumb to go back).
- The existing unsaved-changes prompt still fires when navigating with a dirty form.
- Works **full-screen** (URL-driven trail via the `rc` query param) **and inside a dialog** (in-place reload, in-memory trail).

### Agnostic editor via `EditContentHost` (the mechanism)
The editor and its store no longer branch on `isDialogMode` — that conditional is **removed entirely**. Everything presentation-specific goes through an injected `EditContentHost` port:

| Concern | `RouterEditContentHost` (full-screen) | `DialogEditContentHost` (dialog) |
|---|---|---|
| Identity | route params | dialog config |
| Navigation (related/crumb/locale/post-save/restore) | router | in-place reload |
| Title / breadcrumb | document title + shell trail | no-op |
| Save result | URL is source of truth | reported to opener via `saved$` |

The shell provides the router host; the dialog component provides the dialog host. Adding a new presentation (e.g. side panel) is then just a new adapter + a thin host wrapper — no editor changes.

## How to test
**Full-screen** (`/content/:id`): open a content with a relationship field → click a related content's title → it opens with the "Relating content" banner → click a crumb to go back. Verify locale switch, version restore, and post-save navigation still work; and that unsaved changes prompt.

**Dialog** (UVE or "create new related content"): edit/create content in the dialog → click a related content in a relationship field → the editor reloads **in place** within the dialog with the breadcrumb → go back via a crumb → save reports back to the opener.

## Checklist
- [x] Tests: `pnpm nx test edit-content` — 107 suites green
- [x] Lint + typecheck clean
- [ ] Manual QA (full-screen + dialog)

This PR fixes: #36349


core-web/libs/edit-content/src/lib/services/host/overlay-edit-content-host.ts
```diff
@@ -0,0 +1,132 @@
...
+
+        this.#relatedNav.registerTitle(contentlet.inode, contentlet.title);
+
+        const trail = this.#trailInodes();
+        if (trail.length) {
+            this.#trailInodes.set([...trail.slice(0, -1), contentlet.inode]);
+        }
+    }
+
+    goToRestoredVersion(): void {
```
oidacra: `OverlayEditContentHost.goToRestoredVersion()` is a no-op here, and `restoreVersion`'s success handler (in `history.feature.ts`) only calls `host.goToRestoredVersion(...)` without patching `contentlet`. In full-screen the router host re-navigates and re-initializes the store, but in the dialog nothing seems to re-trigger a reload — when a version is restored while editing inside the dialog, does the editor still reflect the restored version, or could it keep showing the pre-restore state?

core-web/libs/ui/src/lib/components/dot-collapse-breadcrumb/dot-collapse-breadcrumb.component.html
```diff
@@ -13,7 +13,7 @@
                     data-testid="btn-collapse" />
             </div>
             <div class="mx-2">
-                <ChevronRightIcon />
+                <i class="pi pi-angle-right text-gray-700"></i>
```
oidacra: `dot-collapse-breadcrumb` is a shared component (also used by the category field), and this swaps `<ChevronRightIcon />` — added by the Angular 21 / PrimeNG 21 migration in PR #34531 — back to `<i class="pi pi-angle-right">`. What's the reason for reverting the icon here? Since the change reaches other consumers of this shared breadcrumb, it'd help to know whether it's intentional.

core-web/libs/edit-content/src/lib/store/edit-content.store.spec.ts
```diff
@@ -33,19 +33,28 @@ import {
...
 
 describe('DotEditContentStore', () => {
     let spectator: SpectatorService<InstanceType<typeof DotEditContentStore>>;
     let store: InstanceType<typeof DotEditContentStore>;
-    let mockActivatedRoute: Partial<ActivatedRoute>;
+
+    // The host resolves which content to open. Tests set its return value to
+    // drive `initialize()` (route params in prod are the host's concern, not the
+    // store's).
+    const mockHost = {
```
oidacra, line 45: This `mockHost` implements ~7 of the ~13 `EditContentHost` members, and the same partial shape is hand-rolled across five spec files (store, content/history/locales/workflow features). As the port grows, wouldn't these mocks silently drift out of sync? Would a shared `createFakeEditContentHost()` helper (typed against the full interface) be worth extracting so all specs reuse one source of truth?

core-web/libs/edit-content/src/lib/store/features/workflow/workflow.feature.ts
```diff
@@ -307,17 +307,17 @@ export function withWorkflow() {
...
-                                        ) {
-                                            router.navigate(['/content', contentlet.inode], {
-                                                replaceUrl: true,
-                                                queryParamsHandling: 'preserve'
-                                            });
-                                        }
+                                        // Navigate to the saved content. The host decides how:
+                                        // the full-screen host reconciles the related-content
+                                        // trail and navigates when the inode changed; the dialog
+                                        // host stays in place (no-op).
```
oidacra: This comment says the dialog host "stays in place (no-op)", but `OverlayEditContentHost.goToSavedContent` isn't a no-op — it registers the new title and repoints the trail's last crumb to the new inode. Should this be reworded (e.g. "repoints its in-memory trail without navigating") to avoid implying nothing happens?

core-web/libs/edit-content/src/lib/components/dot-edit-content-layout/dot-edit-content.layout.component.html
```diff
@@ -55,6 +56,22 @@
...
         }
 
+        <!-- "Relating content" navigation banner: shown once the user has drilled
+             into related content (trail has an origin + at least one step). Each
+             crumb navigates back (router in full-screen, in-place in a dialog), so
+             the unsaved-changes check still prompts on dirty forms. Rendered below
+             the beta message. -->
+        @if (relatingTrail.length >= 2) {
+            <div
+                class="flex min-h-11 items-center gap-3 bg-blue-50 px-8 py-2"
```
oidacra: The "Relating content" banner is a hand-styled `bg-blue-50 / text-blue-600` div, while the sibling banners in this topBar (beta / lock / workflow / invalid) all use `<p-message severity>` with theme tokens. Would using `<p-message severity="info">` here keep it consistent with the other banners and theme-aware?

core-web/libs/edit-content/src/lib/services/host/router-edit-content-host.ts
```diff
@@ -0,0 +1,145 @@
...
+        const rc = this.#relatedNav.buildTrailForSavedInode(contentlet.inode);
+
+        this.#router.navigate(['/content', contentlet.inode], {
+            replaceUrl: true,
+            queryParams: { rc },
+            queryParamsHandling: 'merge'
+        });
+    }
+
+    goToRestoredVersion(inode: string, previousInode: string | undefined): void {
```
oidacra, line 134: Both `reloadContent()` (locale switch) and `goToRestoredVersion()` navigate with `queryParamsHandling: 'preserve'`, so the `rc` trail is kept verbatim — unlike `goToSavedContent()`, which repoints the last crumb via `buildTrailForSavedInode()`. When either happens while a trail is active (length ≥ 2), does the last/"current" crumb end up still labeling the pre-switch/pre-restore inode instead of the content now being edited (AC-B4)? If so, would repointing the trail's last crumb here — mirroring `goToSavedContent()` — keep the breadcrumb accurate?

core-web/libs/edit-content/src/lib/services/host/overlay-edit-content-host.ts
adrianjm-dotCMS: Good catch. Fixed: `OverlayEditContentHost.goToRestoredVersion()` now reloads the editor in place via the in-place navigation stream (`this.#navigation$.next({ inode })`), mirroring `reloadContent()`, so a version restored while editing inside the dialog is reflected instead of keeping the pre-restore state. Added a spec covering it.

core-web/libs/ui/src/lib/components/dot-collapse-breadcrumb/dot-collapse-breadcrumb.component.html
adrianjm-dotCMS: Good call flagging it. Rather than keep `pi pi-angle-right` (or the old `<ChevronRightIcon />`), I moved both breadcrumb separators — the collapse chevron and the `p-breadcrumb` separator template — to Material Symbols (`chevron_right`), which is the platform's standard icon set going forward, so the shared component now aligns with it. Note that `<ChevronRightIcon />` as an element no longer renders under PrimeNG 21.2 (icons became attribute-selector components, `[data-p-icon="chevron-right"]`), which is why the migration's element form had to be swapped out in the first place.

core-web/libs/edit-content/src/lib/store/edit-content.store.spec.ts
adrianjm-dotCMS: Agreed, good idea. The partial mocks work today, but you're right they'll drift as the port grows (it just gained `identityChanges$` and `setNavigationGuard`). Rather than expand this PR's scope, I'll extract a typed `createFakeEditContentHost(overrides?: Partial<EditContentHost>): EditContentHost` and migrate the five specs to it in a small follow-up, so a growing port forces every mock to stay complete.

core-web/libs/edit-content/src/lib/store/features/workflow/workflow.feature.ts
adrianjm-dotCMS: Done. Reworded: the comment now says the dialog host does not navigate but repoints its in-memory trail's last crumb to the new inode (and registers the new title), instead of calling it a no-op.

core-web/libs/edit-content/src/lib/components/dot-edit-content-layout/dot-edit-content.layout.component.html
adrianjm-dotCMS: Done. Converted the "Relating content" banner to `<p-message severity="info">` with a container template holding the label + breadcrumb, so it now matches the sibling top-bar banners (beta/lock/workflow/invalid), inherits the shared top-bar banner sizing, and is theme-aware via the info tokens instead of the hand-coded `bg-blue-50`/`text-blue-600`.

core-web/libs/edit-content/src/lib/services/host/router-edit-content-host.ts
adrianjm-dotCMS: Good catch — you're right, that broke AC-B4 with an active trail. Fixed: `reloadContent()` (locale switch) and `goToRestoredVersion()` now repoint the trail's current crumb to the new inode via `buildTrailForSavedInode()` + `merge` (extracted a shared `#navigateRepointingCurrentCrumb()` that `goToSavedContent()` also uses), instead of preserving `rc` verbatim. So when the edited inode changes in place while a trail is active, the current crumb follows the content actually being edited; with no active trail `buildTrailForSavedInode()` returns null and just clears `rc`. Added specs for both.

nicobytes (commented):
## Review (Angular + simplification + risk)

Nice work on the `EditContentHost` abstraction — I verified the design against the actual code (not just the diff) and it holds up well. Leaving inline comments on the specific sections below; overall I didn't find any blocking bugs.

core-web/apps/dotcms-ui/src/app/app.routes.ts
```diff
@@ -129,9 +129,14 @@ const PORTLETS_ANGULAR: Route[] = [
     {
         canActivate: [editContentGuard],
         path: 'content',
-        data: {
-            reuseRoute: false
-        },
+        // No `reuseRoute: false` here: the editor REUSES its component subtree across
```
nicobytes, line 132: ✅ **Verified.** I confirmed this against `DotCustomReuseStrategyService.shouldReuseRoute` (`future.data.reuseRoute !== false`, default `true`): removing the flag from the ancestor `/content` route (not the `:id` child) is exactly what's needed, since a non-reusable ancestor tears down the whole subtree regardless of the child's own setting. The reasoning in this comment is accurate.

core-web/libs/edit-content/src/lib/services/host/router-edit-content-host.ts
```diff
@@ -0,0 +1,188 @@
...
+
+    addBreadcrumb(crumb: { label: string; url: string }): void {
+        this.#globalStore.addNewBreadcrumb({
+            label: crumb.label,
+            target: '_self',
+            url: crumb.url
+        });
+    }
+
+    goToSavedContent(
```
nicobytes, line 119: ✅ **Verified — navigation guard coverage.** I traced every navigating method on this class (`reloadContent`, `goToRestoredVersion`, `goToRelatedContent`, `goToCrumb`) and all of them go through `#navigationGuard`, except `goToSavedContent` (below). Confirmed that's intentional: its only call site (`workflow.feature.ts`) runs right after a successful save, when the form is already pristine — skipping the guard there is safe.

core-web/libs/edit-content/src/lib/store/dot-related-content-navigation.store.ts
```diff
@@ -0,0 +1,217 @@
...
+            trailInodes,
+            /** The trail as crumbs (inode + cached title) for the breadcrumb. */
+            trail: computed<DotRelatedContentCrumb[]>(() =>
+                toRelatedContentCrumbs(trailInodes(), store.titleCache())
+            )
+        };
+    }),
+    withMethods((store) => {
+        /** Caches a title so the breadcrumb can label a crumb by its inode. */
+        const registerTitle = (inode: string | undefined, title: string | undefined): void => {
```
nicobytes, line 157: 🔎 **Minor.** The `titleCache` persisted to sessionStorage has no cap/eviction — it grows unbounded over a long session browsing many related contents. Low risk (it's per-tab and the values are short strings), but might be worth a TODO to bound it.

adrianjm-dotCMS: Good catch added a TODO on registerTitle. Documented rather than capped: it lives in sessionStorage (per-tab, cleared on tab close) and each entry is a short inode → title string, so hitting the quota would need an unrealistic number of distinct contents in one session. The TODO marks where to bound it (keep the last N on write) if a heavy-navigation use case ever shows up.

core-web/apps/dotcms-ui/src/app/app.routes.ts
adrianjm-dotCMS: Thanks for verifying

core-web/libs/edit-content/src/lib/services/host/router-edit-content-host.ts
adrianjm-dotCMS: Thanks for verifying

nicobytes (approved):

PR #36503, opened 2026-07-09
fix(osgi): restart on live-export state, not the shared upload folder (#36434)

### LINKED ISSUE #36434: [DEFECT] OSGi plugin fails to resolve on a cluster (osgi.wiring.package … version>=0.0.0) because the system export is written to osgi-extra.conf but the framework restart is skipped
**Freshdesk:** https://dotcms.freshdesk.com/a/tickets/37894

Code references below are pinned to `main` @ `5efb47a8d9fa26cab0911f25eb5fbe8525026d5a`.

## Problem statement

On a **clustered** dotCMS (2+ nodes, Evergreen/SaaS or self-hosted with a shared `/data/shared`), uploading an OSGi plugin that imports a dotCMS package **not yet in the exported-packages list** can leave the plugin permanently unresolved on one or more nodes, with:

```
org.osgi.framework.BundleException: Unable to resolve <bundle>: missing requirement
[<bundle>] osgi.wiring.package; (&(osgi.wiring.package=<pkg>)(version>=0.0.0))
```

…even though the Exported Packages list (`osgi-extra.conf`) **does** contain `<pkg>;version=0`. Customers "fix" it by adding `<pkg>;version="0.0.0"` and re-deploying, which makes it look like a version-matching problem. **It is not.**

### This is NOT a version-string problem

In OSGi, `version=0` parses to `0.0.0`, and an import with no floor compiles to `(version>=0.0.0)`, which a bare `version=0` export satisfies. A bundle importing `pkg;version=0` resolves to ACTIVE when `pkg` is exported as bare `version=0`, and fails **only** when `pkg` is absent from the running framework at resolve time. The `;version="0.0.0"` workaround works only because editing exports / re-deploying triggers a fresh framework restart — the value is incidental. (Regression tests demonstrating this will accompany the fix PR.)

## Root cause

The real cause is a **restart being skipped (or applied too late) on a cluster**, so a package that *is* in `osgi-extra.conf` is not live in the running framework when the bundle is resolved. Two variants, both reproduced:

### Mode A — restart skipped entirely (persistent failure) — primary

[`OSGIUtil.restartOsgi()`](https://github.com/dotCMS/core/blob/5efb47a8d9fa26cab0911f25eb5fbe8525026d5a/dotCMS/src/main/java/org/apache/felix/framework/OSGIUtil.java#L554-L568) only restarts if the **upload** folder still contains jars (`if (UtilMethods.isSet(pathnames))`).

On a cluster, both nodes run the upload-folder poller ([`startWatchingUploadFolder`](https://github.com/dotCMS/core/blob/5efb47a8d9fa26cab0911f25eb5fbe8525026d5a/dotCMS/src/main/java/org/apache/felix/framework/OSGIUtil.java#L410-L429) → [`checkUploadFolder`](https://github.com/dotCMS/core/blob/5efb47a8d9fa26cab0911f25eb5fbe8525026d5a/dotCMS/src/main/java/org/apache/felix/framework/OSGIUtil.java#L453)) against the **shared** upload folder. The race:

1. Node B's poller processes the jar → [`processOsgiPackages`](https://github.com/dotCMS/core/blob/5efb47a8d9fa26cab0911f25eb5fbe8525026d5a/dotCMS/src/main/java/org/apache/felix/framework/OSGIUtil.java#L521) sees new packages → writes `osgi-extra.conf` and schedules a **debounced** `restartOsgi`. Jar stays in `upload`.
2. Node A's poller runs the reload ~fractions of a second later. The export is **already written**, so it hits the [`containsAll == true` branch and moves the jar from the shared `upload` → `load`](https://github.com/dotCMS/core/blob/5efb47a8d9fa26cab0911f25eb5fbe8525026d5a/dotCMS/src/main/java/org/apache/felix/framework/OSGIUtil.java#L529-L531) — no restart.
3. Node B's debounced `restartOsgi` fires, finds `upload` **empty** (`isSet(pathnames)` false), and **silently does nothing**.

Net result: export written to the shared file, jar deployed to `load`, **no node ever restarts** → the package is never published to the running framework → the bundle fails to resolve on **every** node, persistently, until an unrelated restart. fileinstall then auto-starts the jar from the shared `load` folder against the stale framework and logs the `(version>=0.0.0)` error.

### Mode B — remote restart applied too late (transient failure) — secondary

Even when a restart does happen, the origin node publishes a cluster restart request over pub/sub ([`restartOsgiClusterWide`](https://github.com/dotCMS/core/blob/5efb47a8d9fa26cab0911f25eb5fbe8525026d5a/dotCMS/src/main/java/org/apache/felix/framework/OSGIUtil.java#L574-L586); handled at [`OsgiRestartTopic`](https://github.com/dotCMS/core/blob/5efb47a8d9fa26cab0911f25eb5fbe8525026d5a/dotCMS/src/main/java/org/apache/felix/framework/OsgiRestartTopic.java#L83-L85)). A remote node's fileinstall can auto-start the jar from the **shared `load` folder before that node processes the restart**, so it resolves against its still-stale `system.packages.extra` → same error, until the node restarts (then it recovers).

Contributing factor for both: [`org.osgi.framework.storage.clean=onFirstInit`](https://github.com/dotCMS/core/blob/5efb47a8d9fa26cab0911f25eb5fbe8525026d5a/dotCMS/src/main/java/org/apache/felix/framework/OSGIUtil.java#L194) — the felix bundle cache is not cleaned on a restart, only on JVM first init.

## Steps to reproduce

Reproduced on a 2-node cluster sharing `/data/shared` (shared felix `upload`/`load`/`undeployed` + `osgi-extra.conf`). A docker-compose harness + minimal plugin will accompany the fix PR.

1. Bring up a 2-node cluster sharing `/data/shared`.
2. Upload a plugin (with a `Bundle-Activator`, so OSGi resolves it on start) that imports a real dotCMS package **not currently in `osgi-extra.conf`**, e.g. `Import-Package: com.dotcms.ai.model;version=0`.
3. **Mode A:** with normal timing, observe that the export is written to `osgi-extra.conf` and the jar lands in `load`, but no `Restarting OSGI Cluster Wide` occurs → the bundle fails to resolve on both nodes and stays failed.

### Observed evidence

The node that wrote the export, then the skipped restart:
```
17:01:48  There are a new changes into the exported packages
17:01:48  OSGI Extra Packages Saved
17:01:51  Debouncing : restartOsgi after 3000 MILLISECONDS
17:01:51  Restarting OSGI            <-- entered, but upload folder already emptied by the other node → guard fails, no restart
```
`osgi-extra.conf` contains `com.dotcms.ai.model;version=0` the whole time, yet:
```
BundleException: Unable to resolve com.dotcms.osgi.repro-37894d:
  osgi.wiring.package; (&(osgi.wiring.package=com.dotcms.ai.model)(version>=0.0.0))
```
No `Restarting OSGI Locally` after the upload on either node → package never applied.

## Expected vs actual

- **Expected:** once a package is added to `osgi-extra.conf`, every cluster node restarts (or otherwise re-publishes system exports) before the plugin is resolved, so it resolves consistently across the cluster.
- **Actual:** the restart is skipped (Mode A) or races the plugin start on a remote node (Mode B), leaving the plugin unresolved despite the export being present.

## Affected versions

- Current Evergreen [26.06.30-01](https://dev.dotcms.com/docs/changelogs?v=26.06.30-01). 
- Requires a cluster (2+ nodes) sharing `/data/shared`. Not reproducible on a single node.

## Suggested fixes (for discussion)

1. **Don't gate `restartOsgi()` on the upload folder.** Track whether exports changed (or whether jars were moved to `load`) and restart on that, so a concurrent `containsAll` move on another node can't cancel the restart.
2. **Make the export-change → restart atomic per cluster**, coordinated by the existing `OSGI_RESTART_LOCK_KEY` cluster lock, so exactly one node owns "write exports + move jars + trigger cluster-wide restart" for a given upload.
3. **Before auto-starting a bundle from the `load` folder**, verify the required exports are live in the current framework; if not, defer until after the pending restart (addresses Mode B).
4. Consider cleaning the felix bundle cache on restart when exports changed (`storage.clean`), or otherwise forcing re-resolution.

## Diagnostics (proposed alongside the fix)

- Make `felix.log.level` `Config`-driven (e.g. `FELIX_LOG_LEVEL`, default `3`) so `4` can be set to capture the Felix resolver's candidate evaluation during OSGi restarts — directly useful for diagnosing this class of issue in the field.

## External Links

https://dotcms.freshdesk.com/a/tickets/37894

dsolistorres:
## Problem (#36434 — Stage 1 / Mode A)

On a **clustered** dotCMS (2+ nodes sharing `/data/shared`), uploading an OSGi plugin that imports a dotCMS package **not yet in the exported-packages list** could leave the plugin **permanently unresolved** on every node:

```
org.osgi.framework.BundleException: Unable to resolve <bundle>: missing requirement
  osgi.wiring.package; (&(osgi.wiring.package=<pkg>)(version>=0.0.0))
```

…even though `osgi-extra.conf` already contained `<pkg>`. This is **not** a version-string problem (`version=0` ≡ `0.0.0`); it's a **restart that was skipped on the cluster**.

### Root cause (Mode A)
Two flaws combined, both in the upload pipeline of `OSGIUtil`:
1. **The "already exported?" decision read the shared `osgi-extra.conf` file** rather than this node's running framework. When a peer node had just written the new package to that shared file, this node read it back, concluded "no restart needed," and only moved the jar.
2. **The restart trigger was gated on "is the `upload` folder still non-empty?"** The real restart ran debounced, *after* the cluster lock was released, so a peer's jar-move could empty `upload` and silently cancel it.

Net: the export was written to the shared file and the jar deployed to `load`, but **no node ever restarted** → the package was never published to any running framework → persistent resolve failure.

## Fix (Stage 1)
- **Decide against live state, not the file.** Each node snapshots the packages its framework actually started with (`liveExportedPackages`) and compares the plugin's packages against that, so a peer writing the shared file can't trick a node into skipping its restart.
- **Make the decision atomic under the cluster lock.** Writing exports + moving jars to `load` + capturing `needsRestart` all happen under `osgi_restart_lock`; the boolean can't be raced away.
- **Submit the cluster-wide restart off-thread** (single submitter) — the upload endpoint still returns immediately (matching prior behavior) and the restart starts ahead of fileinstall's next poll, so the freshly-moved jar isn't started against a still-stale framework on the origin node.
- **Remove the fragile upload-folder guard** from the debounced restart path (also fixes the manual `PUT /extra-packages` edit, which must restart even with no jar in `upload`).
- **Read exports after the boot reload** in `initializeFramework`, fixing a latent boot-with-preloaded-jars gap.

### Also included
- **`felix.log.level` is now `Config`-driven** (default `3`) so `4` can capture the Felix resolver's candidate evaluation in the field — per the issue's Diagnostics ask. dotCMS runs **two separate Felix frameworks**, so there are **two independent config keys**, one per framework (the system framework namespaces all its felix config under `system.`, matching the existing `system.felix.base.dir` pattern):

  | Framework | Class / line | Config key | Env var | Default |
  |---|---|---|---|---|
  | User / plugin | `OSGIUtil.java:201` | `felix.log.level` | `DOT_FELIX_LOG_LEVEL` | `3` |
  | System (saml, tika) | `OSGISystem.java:96` | `system.felix.log.level` | `DOT_SYSTEM_FELIX_LOG_LEVEL` | `3` |

  Levels: `1`=ERROR, `2`=WARNING, `3`=INFO, `4`=DEBUG. Set the relevant key to `4` to enable resolver DEBUG on that framework only.
- **Tier-A regression tests** in `OSGIUtilTest` proving `version=0` ≡ `0.0.0` and that the error only appears when the package is absent from the running framework.

## Verification
- `./mvnw compile -pl :dotcms-core` — BUILD SUCCESS.
- `OSGIUtilTest` — 8 run, 0 failures/errors/skipped.
- **2-node Docker cluster** sharing `/data/shared`, package absent from exports, plugin dropped into the shared `upload` folder (both pollers racing — the Mode A trigger): the origin node took the new `Restarting OSGI Cluster Wide` path, the peer restarted via pub/sub, and `com.dotcms.osgi.repro-37894d` reached **`ACTIVE` on both nodes** with **zero** `wiring.package` errors.

## Scope / follow-up
This is **Stage 1**: it eliminates the **persistent** failure (Mode A). A **transient, self-healing** window remains: because the jar is moved into the *shared*, fileinstall-watched `load` folder before the async restart republishes the new exports, a still-stale framework can briefly auto-start the bundle and log a `wiring.package=…(version>=0.0.0)` error that clears on the pending restart. This window can occur on **both the origin node and remote nodes** — it is intentionally deferred to **Stage 2** (node-local `load` + verify-exports-before-start / version-gated apply), tracked in #36504. It is documented in a code comment at the jar-move site so it isn't mistaken for an oversight. In practice the origin node's restart is submitted immediately and typically outruns fileinstall's poll (the 2-node test below saw zero errors), but the window is not formally closed until Stage 2.

Refs: #36434

This PR fixes: #36434

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/rest/api/v1/osgi/OSGIResource.java
+++ b/dotCMS/src/main/java/com/dotcms/rest/api/v1/osgi/OSGIResource.java
@@ -2,6 +2,7 @@ package com.dotcms.rest.api.v1.osgi;
 
 import com.dotcms.rest.ResponseEntityBooleanView;
 import com.dotcms.rest.ResponseEntityListView;
+import com.dotcms.rest.ResponseEntitySetStringView;
 import com.dotcms.rest.ResponseEntityStringView;
 import com.dotcms.rest.ResponseEntityView;
 import com.dotcms.rest.WebResource;
@@ -662,7 +663,7 @@ public class OSGIResource {
                                     + "OSGI_BUNDLES_LOADED / OSGI_BUNDLES_UPLOAD_FAILED system events for "
                                     + "the outcome.",
                             content = @Content(mediaType = "application/json",
-                                    schema = @Schema(implementation = ResponseEntityStringView.class))),
+                                    schema = @Schema(implementation = ResponseEntitySetStringView.class))),
                     @ApiResponse(responseCode = "403", description = "Can not access the upload folder or invalid OSGI Upload request"),
             })
     public final Response uploadBundles(@Context final HttpServletRequest request,
--- a/dotCMS/src/main/java/com/dotcms/rest/api/v1/osgi/OSGIResource.java
+++ b/dotCMS/src/main/java/com/dotcms/rest/api/v1/osgi/OSGIResource.java
@@ -718,7 +718,7 @@ public class OSGIResource {
         // refresh strategy is running by schedule job
         OSGIUtil.getInstance().checkUploadFolder();
 
-        return Response.ok(new ResponseEntityView<>(
+        return Response.ok(new ResponseEntitySetStringView(
                 files.stream().map(File::getName).collect(Collectors.toSet())))
                 .build();
     }
--- a/dotCMS/src/main/java/com/dotcms/rest/api/v1/osgi/OSGIResource.java
+++ b/dotCMS/src/main/java/com/dotcms/rest/api/v1/osgi/OSGIResource.java
@@ -624,7 +624,14 @@ public class OSGIResource {
     }
     
     /**
-     * This endpoint receives multiples jar files in order to upload to the osgi.
+     * This endpoint receives multiple jar files in order to upload them to the OSGi framework.
+     * <p>
+     * The response returns as soon as the jars are copied and processing is scheduled. If the
+     * uploaded plugins require new exported packages, the OSGi framework is restarted cluster-wide
+     * <strong>asynchronously</strong> after this method returns, so a {@code 200} confirms the upload
+     * was accepted, not that the plugin is live. Callers should react to the {@code OSGI_FRAMEWORK_RESTART}
+     * and {@code OSGI_BUNDLES_LOADED} system events for completion, and {@code OSGI_BUNDLES_UPLOAD_FAILED}
+     * for asynchronous failures.
      *
      * @param request
      * @param response
@@ -638,9 +645,22 @@ public class OSGIResource {
     @Produces({MediaType.APPLICATION_JSON, "application/javascript"})
     @Consumes(MediaType.MULTIPART_FORM_DATA)
     @Operation(summary = "Upload bundles to the OSGI framework",
+            description = "Uploads one or more plugin JARs to the OSGi upload folder and schedules "
+                    + "their processing. A 200 confirms the upload was accepted, NOT that the plugin "
+                    + "is live: if new packages must be exported the OSGi framework is restarted "
+                    + "cluster-wide asynchronously, after this response returns. Restart completion is "
+                    + "signalled by the OSGI_FRAMEWORK_RESTART system event and successful deployment by "
+                    + "the OSGI_BUNDLES_LOADED event; an asynchronous failure is reported via the "
+                    + "OSGI_BUNDLES_UPLOAD_FAILED event and an error notification (check the server logs "
+                    + "for the full stack trace). Clients should react to these events rather than "
+                    + "assume the plugin is active once the 200 is received.",
             responses = {
                     @ApiResponse(
                             responseCode = "200",
+                            description = "Bundle upload accepted and scheduled. Any required OSGi "
+                                    + "restart happens asynchronously; watch the OSGI_FRAMEWORK_RESTART / "
+                                    + "OSGI_BUNDLES_LOADED / OSGI_BUNDLES_UPLOAD_FAILED system events for "
+                                    + "the outcome.",
                             content = @Content(mediaType = "application/json",
                                     schema = @Schema(implementation = ResponseEntityStringView.class))),
                     @ApiResponse(responseCode = "403", description = "Can not access the upload folder or invalid OSGI Upload request"),
--- a/dotCMS/src/main/java/org/apache/felix/framework/OSGISystem.java
+++ b/dotCMS/src/main/java/org/apache/felix/framework/OSGISystem.java
@@ -36,6 +36,7 @@ public class OSGISystem {
     private static final String AUTO_DEPLOY_DIR_PROPERTY = AutoProcessor.AUTO_DEPLOY_DIR_PROPERTY;
     private static final String FELIX_FILEINSTALL_DIR = "felix.fileinstall.dir";
     private static final String FELIX_UNDEPLOYED_DIR = "felix.undeployed.dir";
+    private static final String FELIX_LOG_LEVEL = "felix.log.level";
     private static final String UTF_8 = "utf-8";
     private static final String PROPERTY_OSGI_PACKAGES_EXTRA = "org.osgi.framework.system.packages.extra";
     private final Framework felixFramework;
@@ -91,7 +92,8 @@ public class OSGISystem {
         felixProps.put("felix.fileinstall.log.level", "3");
         felixProps.put("org.osgi.framework.startlevel.beginning", "2");
         felixProps.put("org.osgi.framework.storage.clean", "onFirstInit");
-        felixProps.put("felix.log.level", "3");
+        // felix.log.level: 1=ERROR, 2=WARNING, 3=INFO, 4=DEBUG.
+        felixProps.put(FELIX_LOG_LEVEL, Config.getStringProperty("system." + FELIX_LOG_LEVEL, "3"));
         felixProps.put("felix.fileinstall.disableNio2", "true");
         felixProps.put("gosh.args", "--noi");
 
--- a/dotCMS/src/main/java/org/apache/felix/framework/OSGIUtil.java
+++ b/dotCMS/src/main/java/org/apache/felix/framework/OSGIUtil.java
@@ -13,6 +13,7 @@ import com.dotcms.dotpubsub.DotPubSubProvider;
 import com.dotcms.dotpubsub.DotPubSubProviderLocator;
 import com.dotcms.rest.ErrorEntity;
 import com.dotcms.util.ReturnableDelegate;
+import com.dotcms.util.VoidDelegate;
 import com.dotmarketing.business.APILocator;
 import com.dotmarketing.exception.DotRuntimeException;
 import com.dotmarketing.osgi.HostActivator;
@@ -220,14 +221,39 @@ public class OSGIUtil {
         if(UtilMethods.isEmpty(extraPackages)) {
             return ;
         }
-        
-        if("RESET".equals(extraPackages)) {
-            new File(OSGIUtil.getInstance().getOsgiExtraConfigPath()).delete();
-            debouncer.debounce("restartOsgi", this::restartOsgi, delay, TimeUnit.MILLISECONDS);
-            return;
+
+        final boolean reset = "RESET".equals(extraPackages);
+
+        // Validate before locking: the dry-run is a pure validation with no side effects, so it stays
+        // outside the cluster lock (keeps the lock hold short and preserves the OsgiException the REST
+        // layer maps to a 400).
+        if (!reset && Config.getBooleanProperty("OSGI_TEST_DRY_RUN", true) && testDryRun) {
+            this.testDryRun(extraPackages);
         }
 
-        this.persistOsgiExtras(extraPackages, testDryRun);
+        // Serialize osgi-extra.conf mutations cluster-wide under the SAME lock the upload pipeline uses
+        // (checkUploadFolder), so a manual edit / RESET on one node cannot clobber an in-progress upload
+        // merge on another node (lost update). Keeping the shared file consistent with live state is
+        // exactly what this fix is about — this closes the gap for the other writer of the file (#36434).
+        final ClusterLockManager<String> lockManager =
+                DotConcurrentFactory.getInstance().getClusterLockManager(OSGI_RESTART_LOCK_KEY);
+        try {
+            lockManager.tryClusterLock((VoidDelegate) () -> {
+                if (reset) {
+                    new File(OSGIUtil.getInstance().getOsgiExtraConfigPath()).delete();
+                } else {
+                    // Dry-run already performed above; do not repeat it inside the lock.
+                    this.persistOsgiExtras(extraPackages, false);
+                }
+            });
+        } catch (final Throwable t) {
+            // Preserve the original unchecked exception type (e.g. DotRuntimeException) so the REST
+            // layer maps it to the same response as before.
+            if (t instanceof RuntimeException) {
+                throw (RuntimeException) t;
+            }
+            throw new DotRuntimeException(getExceptionMessage(t), t);
+        }
 
         //restart OSGI after delay (manual extra-packages edit path keeps the debounce)
         debouncer.debounce("restartOsgi", this::restartOsgi, delay, TimeUnit.MILLISECONDS);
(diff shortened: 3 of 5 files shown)


dsolistorres:
### ✅ Addressed review observation — log-level config keys documented

Thanks for catching the mismatch between the PR description and the code. The two `felix.log.level` reads use **different config keys on purpose**, because dotCMS runs **two separate Felix frameworks**, each with its own config namespace:

| Framework | Code | Config key | Env var |
|---|---|---|---|
| User / plugin | `OSGIUtil.java:201` | `felix.log.level` | `DOT_FELIX_LOG_LEVEL` |
| System (saml, tika) | `OSGISystem.java:96` | `system.felix.log.level` | `DOT_SYSTEM_FELIX_LOG_LEVEL` |

The system framework already namespaces all of its felix configuration under `system.` (e.g. `system.felix.base.dir`), so `system.felix.log.level` keeps that convention and lets each framework's resolver verbosity be tuned independently. Levels: `1`=ERROR, `2`=WARNING, `3`=INFO, `4`=DEBUG (default `3`).

No code change needed — the **PR description has been updated** with the table above so the two keys are explicit for reviewers and QA.

dsolistorres:
### Addressed backend-review feedback (commit `5939b80`)

**🟠 High — jar moved to shared `load` before the async restart (pre-restart fileinstall window).**
Confirmed real, and it is the **same class of race as "Mode B"** deferred to #36504 — with the valid nuance the reviewer raised: it can occur on the **origin node too**, not only remote nodes. A complete fix requires the Stage 2 machinery (node-local `load` + verify-exports-before-start / version-gated apply); partial origin-only patches are fragile (the jar-move sequencing breaks or delays deploy, and they don't help remote nodes anyway). Per the reviewer's first suggestion, I've:
- Added an explicit **code comment at the jar-move site** documenting the window (origin + remote) as a transient, self-healing, Stage-2-tracked condition, so it isn't mistaken for an oversight.
- Updated the **PR description** and **broadened #36504** to state the origin-node case explicitly.

This is transient/self-healing (not the persistent Mode A this PR fixes), and the 2-node cluster test saw zero errors because the origin restart outruns fileinstall's poll — but it's honestly not *formally* closed until Stage 2.

**🟡 Medium — addressed now:**
- **M3** (publish/async-restart failure invisible to admin): the async catch now calls `pushBundleUploadError(...)` → system event + error toast, covering pub/sub publish failures that `restartOsgiOnlyLocal` doesn't handle itself.
- **M5** (cast to disambiguate the `tryClusterLock` overload): replaced with a typed `ReturnableDelegate<Boolean>` local.
- **M6** (tests didn't exercise the new mechanism): added `test_zzz_04` — seeds `osgi-extra.conf` with the package (simulating a peer's write) while this node's `liveExportedPackages` lacks it, and asserts `processOsgiPackages` still returns `needsRestart=true` (Mode A), plus a control asserting `false` once it's live. Full class green locally: **9 run, 0 failures**.

**🟡 Medium — intentionally deferred (with rationale):**
- **M1** (startup `fireReload` mutates `osgi-extra.conf` without the cluster lock): legit, but boot-time cluster locking has startup-ordering risk; better handled deliberately (candidate for Stage 2 / its own change) than bolted on here.
- **M2** (event pushed before restart vs after): current order is deliberate so the UI shows "restarting" promptly; the failure path now surfaces an error toast (M3), so an optimistic event is corrected by a subsequent error. Left as-is.
- **M4** (no debounce/coalescing on the async restart): not a correctness issue — the single-thread `OSGIRestart` submitter serializes restarts; at most a redundant restart under rapid uploads.

dsolistorres:
### Addressed backend-review observations (commit `8959815`)

Both items were documentation/observability gaps around the new fire-and-forget restart path — fixed:

**🟡 Item 1 — upload endpoint docs didn't reflect async restart.**
`OSGIResource.uploadBundles` now documents (in both `@Operation`/`@ApiResponse` and the method Javadoc) that a `200` confirms the upload was **accepted/scheduled, not that the plugin is live**: any required OSGi restart runs cluster-wide asynchronously after the response, with completion signalled by `OSGI_FRAMEWORK_RESTART`, successful deployment by `OSGI_BUNDLES_LOADED`, and async failure by `OSGI_BUNDLES_UPLOAD_FAILED`. Clients should react to those system events rather than assume the plugin is active on the `200`. `openapi.yaml` was regenerated from the annotations and is included in the commit.

**🟡 Item 2 — async restart-failure message lacked jar names.**
`checkUploadFolder` now captures the uploaded jar names (before the reload moves them out of the upload folder) and includes them in the async restart-failure message: `Cluster-wide OSGI restart failed after bundle upload [<jars>]: …`, so an admin who sees the standalone error toast can correlate it back to the triggering upload — matching the other `pushBundleUploadError` call sites in the file.

`./mvnw compile -pl :dotcms-core` — BUILD SUCCESS (openapi.yaml regen included).

dsolistorres:
### Addressed @Schema observation (commit `79fe5e0`)

`uploadBundles` returns `ResponseEntityView<Set<String>>` (the set of uploaded jar names), but the 200 `@ApiResponse` declared `@Schema(implementation = ResponseEntityStringView.class)`. Corrected it to `ResponseEntitySetStringView.class` (`com.dotcms.rest.ResponseEntitySetStringView extends ResponseEntityView<Set<String>>`) so the OpenAPI contract matches the actual return type, and added the import.

Regenerated `openapi.yaml` from the annotations — the OSGi upload response now `$ref`s `#/components/schemas/ResponseEntitySetStringView` (component definition added). Verified the committed `openapi.yaml` matches build output (no drift), so the CI OpenAPI check will pass.

`./mvnw compile -pl :dotcms-core` — BUILD SUCCESS.

dsolistorres:
### Re: OSGI_FRAMEWORK_RESTART event ordering (commit `943db07`)

Reviewed this one closely — the ordering here is **intentional**, and I've added a code comment documenting why so it isn't re-flagged.

`OSGI_FRAMEWORK_RESTART` is a *"restart is happening"* signal (the `dot-plugins` store reacts by showing a `restarting` spinner), **not** a success signal. In this path the restart runs **off-thread (fire-and-forget)** and blocks for several seconds, so pushing the event **before** `restartOsgiClusterWide()` is what shows the spinner *while* the restart is in progress. Pushing it *after* (as the synchronous `restartOsgi()` / `_restart` endpoint paths do) would only flip the UI to "restarting" once the work is already finished — backwards for this path.

The "optimistic success" concern is covered: a restart **failure** is not reported as success — the catch block pushes `OSGI_BUNDLES_UPLOAD_FAILED` (via `pushBundleUploadError`), which the store handles by resetting status to `loaded` and showing an error toast. So a failed restart self-corrects the UI.

No behavior change — just a clarifying comment on the intentional difference from the synchronous paths.

dsolistorres:
### Addressed the latest review round (commit `2ddd862`)

**🟠 High #1 — `uploadBundles` return type drift.** Correct catch — my earlier commit fixed the `@Schema` but left the return building a raw `ResponseEntityView<>`. Now constructs `ResponseEntitySetStringView`, so the object matches the declared schema. (No `openapi.yaml` change — the annotation was already correct.)

**🟠 High #2 — manual `osgi-extra.conf` writer not mutually exclusive with the upload writer.** Confirmed true: the upload path mutates the file under `OSGI_RESTART_LOCK_KEY`, but `writeOsgiExtras` (manual `PUT /extra-packages` + `RESET`) had no lock, so a manual edit on one node could clobber an in-progress upload merge on another (lost update). This is a Stage-1 consistency gap (same principle this PR establishes), **not** Stage 2 — so fixed here: `writeOsgiExtras` now serializes its file mutation under the same cluster lock. The dry-run validation runs *before* the lock (pure validation, keeps lock-hold short and preserves the `OsgiException` → 400 REST behavior); no reentrancy with the upload path (that path calls `persistOsgiExtras` directly, never `writeOsgiExtras`).

**🟡 Medium #3 — GLOBAL-visibility broadcast leaked jar names + exception text.** The new async restart-failure path now logs the full detail (jar names + exception) to the **server log only** and broadcasts a **generic** message ("check the dotCMS server logs for details") to avoid exposing it to every logged-in backend user. (The pre-existing `pushBundleUploadError` sites and a proper permission-scoped `Visibility` are broader and left out of this PR.)

`./mvnw compile -pl :dotcms-core` — BUILD SUCCESS; `openapi.yaml` unchanged.

_Note: I couldn't run the docker-orchestrated `OSGIUtilTest` locally this round (Docker daemon down in my environment) — but these three changes don't touch the paths the existing 9 tests cover, and CI runs `OSGIUtilTest` (MainSuite2a/2b) + the Osgi postman collection._

dsolistorres:
@claude the changes you requested have been addressed in commit `2ddd862` (on top of the earlier review rounds). Summary of what was fixed since your last review of `943db07`:

- **High** — `uploadBundles` now constructs `ResponseEntitySetStringView` so the returned object matches the declared `200` `@Schema` (previously still `ResponseEntityView<>`).
- **High** — `writeOsgiExtras` (manual `PUT /extra-packages` + `RESET`) now serializes its `osgi-extra.conf` mutation under the same `OSGI_RESTART_LOCK_KEY` cluster lock the upload pipeline uses, so the two writers are mutually exclusive cluster-wide (dry-run validation stays outside the lock to preserve the `OsgiException` → 400 behavior).
- **Medium** — the async cluster-restart-failure broadcast now sends a generic message and keeps jar names + exception text in the server log only (no longer leaked to every logged-in backend user via `GLOBAL` visibility).

`OSGIUtilTest` passes locally (9/9). Please re-review the current head — thanks!

dsolistorres:
### Non-blocking follow-up tracked in #36537

The backend review's note on the High #2 lock fix — that `ClusterLockManager.tryLock` (ShedLock) **silently skips** when the lock is already held, so a manual `PUT /extra-packages` edit that loses the lock to an in-progress upload merge still returns an optimistic `200 "OSGI Extra Packages Saved"` — is tracked as a **non-blocking follow-up** in #36537.

It's split out (not fixed here) because returning a `409` requires a paired frontend change: `DotHttpErrorManagerService` has no `CONFLICT`/409 handler, so a raw 409 would be swallowed silently in the UI (no toast, dialog stays open). #36537 covers both the backend `409`-on-skipped-write and the UI handling.

This does not affect #36503's correctness — the lock fix itself is sound (the shared file never diverges); #36537 only replaces the optimistic `200` on the contention path with an honest `409` + matching UI feedback.

dsilvam (approved):

PR #36510, opened 2026-07-10
fix(permissions): enforce contentlet-level WRITE check on Copy and Edit (#34215)

gortiz-dotcms:
## Problem

When a user only has **View** permission on a contentlet instance, they could still use the **Copy and Edit** button in the Page Edit screen to create a copy of that contentlet. Additionally, a user with **Read-only** access on a page could call the `_deepcopy` endpoint and receive a fully editable duplicate of that page — same class of bug, different object.

Both code paths called `contentletAPI.copyContentlet()` directly, which only enforces READ, not WRITE.

## Root Cause

1. **Copy and Edit (contentlet):** `PageResourceHelper.copyContent()` fetched the source contentlet and immediately invoked `contentletAPI.copyContentlet()` without verifying the user had WRITE permission on the specific contentlet instance. Permission was only evaluated at the content type level.

2. **Deep copy (page):** `PageResourceHelper.copyPage()` called `contentletAPI.copyContentlet(page, ...)` without a WRITE check on the page itself. A READ-only user could create a page copy before the child-contentlet WRITE checks caused a rollback — wrong failure mode, wrong place.

## Fix

Added explicit `permissionAPI.doesUserHavePermission(..., PERMISSION_WRITE, ...)` checks in both code paths before any copy is attempted. Both throw `DotSecurityException` (→ 403) if the user lacks WRITE on the instance.

## Changes

- `dotCMS/src/main/java/com/dotcms/rest/api/v1/page/PageResourceHelper.java`
  - `copyContent()` — added instance-level WRITE check before copying a contentlet (Copy and Edit)
  - `copyPage()` — added WRITE check on the page before deep-copying it

## Test Plan

- [ ] Log in as a user with View-only on a contentlet but Publish on the content type → "Copy and Edit" should return 403
- [ ] Log in as a user with WRITE on the contentlet → "Copy and Edit" should succeed as before
- [ ] Log in as a user with Read-only on a page → `PUT /page/{pageId}/_deepcopy` should return 403
- [ ] Log in as a user with WRITE on the page → deep copy should succeed as before
- [ ] Run: `./mvnw verify -pl :dotcms-integration -Dcoreit.test.skip=false -Dit.test=PageResourceTest`

Fixes: #34215

This PR fixes: #34215

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/rest/api/v1/page/PageResourceHelper.java
+++ b/dotCMS/src/main/java/com/dotcms/rest/api/v1/page/PageResourceHelper.java
@@ -329,6 +329,10 @@ public class PageResourceHelper implements Serializable {
 
         if (page instanceof HTMLPageAsset) {
 
+            if (!permissionAPI.doesUserHavePermission(HTMLPageAsset.class.cast(page), PermissionAPI.PERMISSION_WRITE, user, pageMode.respectAnonPerms)) {
+                throw new DotSecurityException(String.format("User '%s' does not have WRITE permission on Page '%s'",
+                        user.getUserId(), page.getIdentifier()));
+            }
             final Contentlet newPage = this.contentletAPI.copyContentlet(
                     HTMLPageAsset.class.cast(page), user, pageMode.respectAnonPerms);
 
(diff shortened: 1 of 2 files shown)


dsilvam:
@gortiz-dotcms is it possible to include a test here?

dsilvam (approved):

PR #36512, opened 2026-07-10
fix(viewtools): keep aggregations in $json.generate() reflection JSON (#36435)

### LINKED ISSUE #36435: $json.generate() reflection-based navigation of aggregation results silently breaks after #36026 records migration
## Problem Statement

#36026 fixed the ES→OpenSearch aggregation return-type regression so that **direct** Velocity object navigation of aggregations works again: `$results.aggregations.<name>.buckets`, `getKeyAsNumber()`, `getDocCount()`, nested `getAggregations()`/`top_hits` — all restored and covered by `ContentSearchToolTest` (PR #36027).

A related but distinct access pattern is still silently broken: templates that call `$json.generate($rawResults.response)` (`JSONTool.generate(Object)` → `new JSONObject(o)`, org.json's reflection/bean-introspection constructor) and then navigate the resulting `JSONObject`, e.g. `$results.aggregations.get("asMap").folders.buckets`.

### Root cause (verified against actual bytecode/behavior, not inferred)

**Before the migration:** `ESSearchResults.getResponse()` returned the raw `org.elasticsearch.action.search.SearchResponse` — a real bean. `SearchResponse.getAggregations()` returns `org.elasticsearch.search.aggregations.Aggregations`, which itself has a bean-style `getAsMap()` (confirmed via `javap` against the vendored ES 7.10.2 jar). So `new JSONObject(searchResponse)` reflected `getAggregations()` → nested-reflected `getAsMap()` → produced `{"aggregations": {"asMap": {"folders": {...}}}}`. That's the only reason `.aggregations.get("asMap").folders.buckets` ever worked.

**After the migration:** `ContentSearchResults.getResponse()` now returns `ContentSearchResponse` — a Java record with bare-named components (`hits`, `scrollId`, `tookMillis`, `aggregationTree`) and a derived `aggregations()` method, **none prefixed with `get`**. I confirmed with a compiled test against the actual `org.json:json:20231013` jar that `new JSONObject(x)` on a record shaped this way (no `get`-prefixed accessors) produces a **completely empty `{}`** — not a map missing a sub-key, but nothing at all. So `$results.aggregations` itself resolves to nothing; the `.get("asMap")` hop never even gets a real map to look into.

Notably, `Aggregation` and `AggregationBucket` (the nested per-bucket types) *were* designed with this in mind — their record components are deliberately named `getName`, `getType`, `getBuckets`, `getHits`, `getKey`, `getDocCount`, etc., which I confirmed round-trips correctly through `JSONObject`'s bean reflection. `ContentSearchResponse` itself just didn't get the same treatment, and it's the object the customer's template happens to fetch via `.response` before JSON-ifying it.

This produces the same "no error thrown, just missing data" failure mode #36026 was originally raised for — just one hop earlier in the chain, and via a path #36026/#36027's tests didn't exercise (they navigate the Java object directly; this bug only shows up when a template round-trips the response through `$json.generate()` first).

## Steps to Reproduce

```velocity
#set($esQuery = '{"query":{"query_string":{"query":"+baseType:5 +conhost:$hostIdentifier"}},"aggs":{"folders":{"terms":{"field":"parentpath_dotraw","size":2000}}},"size":0}')
#set($rawResults = $estool.search($render.eval($esQuery)))
#set($results = $json.generate($rawResults.response))
#foreach($folder in $results.aggregations.get("asMap").folders.buckets)
    <url><loc>https://${host}${folder.key}</loc></url>
#end
```
`$results` (i.e. `$json.generate($rawResults.response)`) serializes to `{}`. `$results.aggregations` is therefore always absent, so the `#foreach` silently iterates zero times — no exception, no log line.

## Real-World Impact

Found via Freshdesk ticket 38225 — a customer's `sitemap.vtl` across multiple sites rendered only the primary/base URL; all ES-aggregation-driven sub-page loops (folders, pages, urlmaps) came up empty. Customer is on current dotEvergreen (already has the #36026 fix), so the direct-navigation fix alone does not help this pattern. Blast radius on other customers using the same `json.generate()`-then-navigate idiom is unknown — worth a docs/KB sweep.

## Workaround (confirmed working)

Skip `$json.generate()` entirely and navigate the Velocity object directly — this is the path #36026 actually restored and tested:
```velocity
#set($rawResults = $estool.search($render.eval($esQuery)))
#foreach($folder in $rawResults.aggregations.get("folders").buckets)
    <url><loc>https://${host}${folder.key}</loc></url>
#end
```

## Acceptance Criteria

- [ ] Decide: give `ContentSearchResponse` the same `get`-prefixed record-component treatment as `Aggregation`/`AggregationBucket` so `$json.generate()` round-trips correctly again, OR treat this as a documented breaking change and proactively identify/migrate affected customer templates.
- [ ] If (a), add a regression test that drives the customer's exact `$json.generate($rawResults.response)` → `.aggregations...` pattern end-to-end (not just direct object navigation, which #36026 already covers).
- [ ] Note possible impact on other customers using this pattern beyond ticket 38225.

## dotCMS Version

[26.06.30-01](https://dev.dotcms.com/docs/changelogs?v=26.06.30-01)

## Links
Follow-up to #36026 (originating migration); PR #36027 (fix whose scope didn't cover this path). Freshdesk ticket: https://dotcms.freshdesk.com/a/tickets/38225
https://helpdesk.dotcms.com/a/tickets/38486

## Severity
Defect, not urgent — confirmed template-level workaround exists.

fabrizzio-dotCMS:
## Summary

**The problem:** When a VTL template does `$json.generate($rawResults.response)` to turn search results into JSON and then reads the aggregations, **the JSON came out without the aggregations**. The `#foreach` printed nothing — no error, no log. A customer hit this because their `sitemap.vtl` rendered only the base URL (Freshdesk 38225).

**The cause:** During the ES→OpenSearch migration, `getAggregations()` was annotated `@JsonIgnore` (so Jackson wouldn't duplicate the tree in the API JSON). But the serializer behind `$json.generate()` **also honours that annotation** → it skipped the aggregations → they vanished from the generated JSON.

**The fix:** Move that suppression from a method-level `@JsonIgnore` to a class-level `@JsonIgnoreProperties`. Jackson still honours it → the API JSON is unchanged. The `$json.generate()` serializer does **not** honour it → the aggregations show up again. Extended to `tookInMillis` and `suggest`, which shared the same latent bug.

**The breaking change:** The legacy `.get("asMap")` hop no longer works (it was an Elasticsearch artifact). The correct navigation is now direct: `aggregations.<name>.buckets`.

**Verification:** unit tests 14/14 ✅, end-to-end integration 10/10 ✅ (real Velocity engine + Elasticsearch).

---

## Problem

Templates that round-trip a raw search response through `$json.generate($rawResults.response)` and then navigate the resulting `JSONObject` (e.g. `$results.aggregations.<name>.buckets`) silently got **no aggregations** after the ES→OpenSearch migration. A customer's `sitemap.vtl` rendered only the base URL — every aggregation-driven loop (folders, pages, urlmaps) iterated zero times, with no exception and no log line (Freshdesk 38225).

This is a distinct path from #36026/#36027, which restored **direct** Velocity object navigation (`$rawResults.aggregations...`). This bug shows up only when the template serializes the response through `$json.generate()` **first**.

## Root cause

`JSONTool.generate(Object)` is literally `new com.dotmarketing.util.json.JSONObject(bean)` — a reflection/bean serializer that:
1. only reads public zero-arg `getX()`/`isX()` accessors, and
2. **honours Jackson's `@JsonIgnore`** (`JSONObject.isIgnorable()`).

`ContentSearchResponse.getAggregations()` carried a method-level `@JsonIgnore` (added so the neutral Jackson wire is single-sourced from `aggregationTree` and not double-emitted). Because the vendored `JSONObject` also honours `@JsonIgnore`, it skipped the getter too — so the aggregations vanished from the generated JSON and the customer's `#foreach` iterated nothing.

An audit of the object graph found **two more Velocity back-compat aliases with the exact same latent regression**: `getTookInMillis()` and `getSuggest()` also carried a method-level `@JsonIgnore`. Pre-migration the raw ES `SearchResponse` exposed both via getters, so `$json.generate()` templates reading timing/suggestions regressed the same way (lower blast radius — no ticket yet).

## Fix

Move the Jackson suppression from **method-level** `@JsonIgnore` on the getters to a **class-level** `@JsonIgnoreProperties({"aggregations", "tookInMillis", "suggest"})`:

| Serializer | Reads method-level `@JsonIgnore`? | Reads class-level `@JsonIgnoreProperties`? |
|---|---|---|
| Jackson (API wire) | yes | **yes** → still suppresses all three; wire shape unchanged (tree=`aggregationTree`, timing=`tookMillis`, suggest omitted) |
| vendored `JSONObject` (`$json.generate()`) | yes | **no** → the `getX()` aliases are visible again → data restored |

Net: `$json.generate()` templates work again, and the neutral Jackson JSON gains **no** duplicate `aggregations` / `tookInMillis` / `suggest` key.

Rest of the graph audited clean: `SearchHits`, `SearchHit`, `Aggregation`, `AggregationBucket`, `TotalHits` and `ContentSearchResults` carry no `@JsonIgnore` getters, so no other attribute is affected.

## Test plan

**Automated**
- `AggregationDomainTest` (unit, no search engine): drives the exact `new JSONObject(response)` reflection path and asserts `aggregations` (down to each bucket's `key`/`docCount`), `tookInMillis` and `suggest` all survive; the Jackson-shape test asserts none of the three leak a duplicate wire key (`tookMillis`/`aggregationTree` stay; `tookInMillis`/`suggest`/`aggregations` absent). 14/14 green. Verified failing on the pre-fix source (1 failure) and passing with the fix.
- `ContentSearchToolTest` (integration): drives the customer's `$json.generate($rawResults.response)`-then-navigate idiom through the real dotCMS Velocity engine — the path #36026/#36027 did not cover.

**Manual (QA)**

_TC1 — aggregations (the reported bug):_
1. On a site with published content, create a VTL page/widget:
   ```velocity
   #set($esQuery = '{"aggs":{"folders":{"terms":{"field":"contentType","size":20}}},"size":0,"query":{"bool":{"filter":[{"term":{"live":true}}]}}}')
   #set($rawResults = $estool.search($esQuery))
   #set($results = $json.generate($rawResults.response))
   RAW JSON: $results
   #foreach($group in $results.aggregations.folders.buckets)
     key: $!{group.key} — docCount: $!{group.docCount}
   #end
   ```
2. **Expected:** `RAW JSON` contains an `"aggregations"` block, and the loop prints one `key: … docCount: …` line per bucket.
3. **Before the fix:** `RAW JSON` has no `aggregations` block and the loop prints nothing.

_TC2 — tookInMillis & suggest (the two siblings fixed in the same PR):_
1. On any site, create a VTL page/widget:
   ```velocity
   #set($esQuery = '{"size":1,"query":{"match_all":{}}}')
   #set($rawResults = $estool.search($esQuery))
   #set($results = $json.generate($rawResults.response))
   RAW JSON: $results
   tookInMillis: $!{results.tookInMillis}
   suggest key present: $!{results.suggest}
   ```
2. **Expected:** `RAW JSON` contains both a `"tookInMillis"` number and a `"suggest"` key (an object, `{}` when the query has no suggester); `tookInMillis:` renders a number and `suggest key present:` renders (not blank).
3. **Before the fix:** `RAW JSON` has neither `tookInMillis` nor `suggest`; both output lines render blank.
4. _(Optional, to see `suggest` populated)_ add a suggester to the query, e.g. `"suggest":{"my-suggestion":{"text":"lorem","term":{"field":"title"}}}`, and confirm the `suggest` block carries the suggester's entries.

## QA Note

- **Breaking-detail for affected templates:** the legacy `.get("asMap")` hop (e.g. `$results.aggregations.get("asMap").folders.buckets`) only ever worked because the pre-migration object was an ES `Aggregations` (which had `getAsMap()`). The neutral tree is flatter, so the correct navigation is now `aggregations.<name>.buckets` **directly** — drop the `asMap` hop. Worth a docs/KB note for customers using the `json.generate()`-then-navigate idiom.
- No REST/wire contract changes: the `/api/es/raw` and `/api/es/search` neutral JSON shape is unchanged (verified by the Jackson no-duplicate assertions).

Closes #36435.

### Rollback safety (H-8 — reviewed)

The rollback-safety bot flags this as a VTL viewtool contract change (H-8, HIGH): the PR **restores** the `$json.generate($rawResults.response).aggregations` (and `.tookInMillis` / `.suggest`) contract, so rolling back to N-1 would silently re-break any template that relies on it. This is inherent to the forward-fix and is expected:

- Templates already using this idiom (e.g. Freshdesk 38225) are **already broken on the current release**; rollback leaves them no worse off.
- The only new exposure is templates *authored/updated to adopt the idiom during this cycle*, which would silently regress on rollback (zero-iteration `#foreach`, no error/log).
- **No REST/wire contract change** — the neutral `/api/es/raw` & `/api/es/search` JSON is unchanged (pinned by the Jackson no-duplicate assertions).

**Release-notes action for ops:** rolling back this release silently reintroduces the `$json.generate(...).aggregations` regression (Freshdesk 38225) for any template adopting the restored idiom this cycle. Weigh that in any rollback decision.

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/content/index/domain/ContentSearchResponse.java
+++ b/dotCMS/src/main/java/com/dotcms/content/index/domain/ContentSearchResponse.java
@@ -39,22 +39,25 @@ import javax.annotation.Nullable;
  *                        sub-aggregations and {@code top_hits} preserved); Velocity resolves
  *                        {@code $results.aggregations.<name>} through this map
  */
-// {@code aggregations} is suppressed for Jackson at the CLASS level (not with a method-level
-// {@code @JsonIgnore} on {@link #getAggregations()}) on purpose: the neutral Jackson JSON must not
-// carry {@code aggregations} — the tree is already serialized as {@code aggregationTree}, and a
-// method-level {@code @JsonIgnore} would ALSO hide it from the reflection-based
+// The Velocity-only aliases {@code aggregations}, {@code tookInMillis} and {@code suggest} are
+// suppressed for Jackson at the CLASS level (not with method-level {@code @JsonIgnore} on the
+// getters) on purpose. Each is redundant on the neutral Jackson wire (the tree is serialized as
+// {@code aggregationTree}, timing as {@code tookMillis}, and suggestions are wire-omitted), but a
+// method-level {@code @JsonIgnore} would ALSO hide them from the reflection-based
 // {@code com.dotmarketing.util.json.JSONObject} bean constructor that {@code JSONTool.generate(Object)}
-// uses (it honours {@code @JsonIgnore}). That reflection path is what {@code $json.generate($response)}
-// templates walk, so hiding {@code aggregations} there is exactly the #36435 regression. A class-level
-// {@code @JsonIgnoreProperties} is read by Jackson but NOT by the vendored JSONObject, so it keeps the
-// wire shape unchanged while leaving {@code getAggregations()} visible to Velocity's json tool.
-@com.fasterxml.jackson.annotation.JsonIgnoreProperties({"aggregations"})
+// uses — it honours {@code @JsonIgnore}. That reflection path is exactly what
+// {@code $json.generate($response)} templates walk, so a method-level ignore silently drops the data
+// there (issue #36435 — originally only {@code aggregations} was reported, {@code tookInMillis} and
+// {@code suggest} shared the same latent regression). A class-level {@code @JsonIgnoreProperties} is
+// read by Jackson but NOT by the vendored JSONObject, so it keeps the wire shape unchanged while
+// leaving the {@code getX()} aliases visible to Velocity's json tool.
+@com.fasterxml.jackson.annotation.JsonIgnoreProperties({"aggregations", "tookInMillis", "suggest"})
 public record ContentSearchResponse(
         SearchHits hits,
         @Nullable String scrollId,
         long tookMillis,
         Map<String, Aggregation> aggregationTree,
-        @com.fasterxml.jackson.annotation.JsonIgnore Map<String, Object> suggest) {
+        Map<String, Object> suggest) {
 
     /**
      * Canonical constructor. {@code aggregationTree} and {@code suggest} default to an empty map when
@@ -82,8 +85,10 @@ public record ContentSearchResponse(
     // that access WITHOUT changing the JSON wire shape:
     //  - getHits()/getScrollId() return the same values as the record components, so Jackson merges
     //    them into the existing "hits"/"scrollId" fields (no new key, no duplicate).
-    //  - getTookInMillis()/getAggregations() carry @JsonIgnore so they remain Velocity-only and
-    //    never add a field to the neutral JSON.
+    //  - getTookInMillis()/getAggregations()/getSuggest() are Velocity-only and must NOT add a field
+    //    to the neutral JSON; that suppression lives in the class-level @JsonIgnoreProperties above
+    //    (Jackson-only) rather than a method-level @JsonIgnore, so they stay visible to the
+    //    reflection-based $json.generate() path (issue #36435).
 
     /** Velocity/back-compat alias for {@link #hits()}; serializes as the same {@code hits} field. */
     public SearchHits getHits() {
@@ -97,9 +102,11 @@ public record ContentSearchResponse(
 
     /**
      * Velocity/back-compat alias mirroring Elasticsearch {@code SearchResponse.getTookInMillis()}.
-     * {@code @JsonIgnore} keeps the neutral JSON unchanged (timing is serialized as {@code tookMillis}).
+     * Deliberately NOT annotated {@code @JsonIgnore}: it must stay visible to the reflection-based
+     * {@code $json.generate($response)} path (issue #36435). The neutral Jackson JSON is kept
+     * unchanged (timing is serialized only as {@code tookMillis}) by the class-level
+     * {@code @JsonIgnoreProperties("tookInMillis")}.
      */
-    @com.fasterxml.jackson.annotation.JsonIgnore
     public long getTookInMillis() {
         return tookMillis;
     }
@@ -122,11 +129,12 @@ public record ContentSearchResponse(
 
     /**
      * Velocity/back-compat alias for {@link #suggest()}, exposing search suggestions as
-     * {@code $r.suggest}. Kept {@code @JsonIgnore} (like the {@code suggest} component itself) so the
-     * neutral JSON shape of {@code /api/es/raw} is unchanged; the ES-wire {@code suggest} block is
-     * emitted only by the {@code /api/es/search} legacy adapter.
+     * {@code $r.suggest}. Deliberately NOT annotated {@code @JsonIgnore}: it must stay visible to the
+     * reflection-based {@code $json.generate($response)} path (issue #36435). The neutral JSON shape of
+     * {@code /api/es/raw} is unchanged (suggestions are wire-omitted) via the class-level
+     * {@code @JsonIgnoreProperties("suggest")}; the ES-wire {@code suggest} block is emitted only by the
+     * {@code /api/es/search} legacy adapter.
      */
-    @com.fasterxml.jackson.annotation.JsonIgnore
     public Map<String, Object> getSuggest() {
         return suggest;
     }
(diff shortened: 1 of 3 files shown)


fabrizzio-dotCMS:
Thanks — the H-8 flag is understood and addressed as a documentation/ops concern, not a code change, since (as the bot notes) restoring the `$json.generate()` contract is inherent to the forward-fix and largely unavoidable.

Actions taken:
- Added a **Rollback safety (H-8)** section to the PR description: rolling back this release silently reintroduces the `$json.generate($rawResults.response).aggregations` regression (Freshdesk 38225) for any template that adopts the restored idiom this cycle → flagged for release notes.
- Clarified scope: templates already using the idiom are **already broken on the current release**, so rollback leaves them no worse off; the only new exposure is templates authored to adopt it during this cycle.
- Confirmed **no REST/wire contract change** — the neutral `/api/es/raw` & `/api/es/search` JSON is unchanged, pinned by the Jackson no-duplicate assertions in `AggregationDomainTest`.

No safer code alternative exists without abandoning the fix itself, so proceeding with the documented rollback caveat.

wezell (approved):
Good test.

Issue #36501, opened 2026-07-09
bug(index): product fixes & findings from phased (ES→OS) integration runs


fabrizzio-dotCMS:
## Purpose

Single tracking issue for **genuine product fixes and findings** surfaced while running the integration battery under a non-zero ES→OS migration phase (see #36320 for the phase-testing tooling and PR #36500). Test-only decoupling of `ContentletIndexAPIImplTest` stays in #36320; product-code fixes land here.

---

## Finding #1 — `deactivateIndex` fails / leaves a stale OS store row when the last index is deactivated

**Where:** `dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImpl.java` — `deactivateIndex(String)`

When the index being deactivated holds the **last populated slot** of the OS `VersionedIndices`, `osBuilder.build()` produces an empty record and `VersionedIndicesAPI.saveIndices(...)` rejects it by contract (`IndicesFactoryImpl:120` — *"At least one index must be specified when saving versioned indices for version: 3.X"*). Two affected sites, two severities:

| Site | Phase | Current behavior | Severity |
|---|---|---|---|
| `:3148` | 3 (OS-only) | `saveIndices(build())` with **no try/catch** → exception **propagates** → `deactivateIndex` fails hard | 🔴 High |
| `:3186` | 1 & 2 (best-effort OS mirror) | caught → `WARN "Could not mirror index deactivation to OS store"` + **stale OS store row left dangling** | 🟠 Medium |

The stale-row case is the exact failure mode `clearOsStorePointer` (`:1966-1978`) was written to prevent for issue #35640: on the next restart, `initOSCatchup` would treat the stale row as authoritative and recreate the deleted index empty.

**Root cause:** `deactivateIndex` never got the empty-record guard that `clearOsStorePointer` already has.

**Fix (parity with `clearOsStorePointer`):** at both save sites, if `build().hasAnyIndex()` → `saveIndices(...)`, else → `versionedIndicesAPI.removeVersion(VersionedIndices.OPENSEARCH_3X)`. This gives the OS versioned store the same "empty = OK" semantics the legacy ES path already has (`legacyIndiciesAPI.point(...)` accepts an all-null `IndiciesInfo`).

**Reproduction:**
```bash
just test-integration-phase 1 ContentletIndexAPIImplTest
# WARN "Could not mirror index deactivation to OS store for index: live_..." during activateDeactivateIndex
# (under phase 3 the same path throws instead of warning)
[...]

Issue #36320, opened 2026-06-25
test: decouple ContentletIndexAPIImplTest from ElasticSearch internals (phase-aware, vendor-neutral) - Make it possible to run MainSuite Tests in all 3 phases


fabrizzio-dotCMS:
## Description

`ContentletIndexAPIImpl` is the vendor-neutral router of the ES→OpenSearch migration, but its integration test `ContentletIndexAPIImplTest` (`dotcms-integration/src/test/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImplTest.java`, 1261 lines, 14 `@Test`) is still coupled to ElasticSearch internals.

Running the suite under a non-zero migration phase (verified at phase 3 via the new `opensearch-phase-suite` profile) makes **6 of 14** tests fail — most of them purely because of ES-specific assertions/helpers, **not** product defects.

```bash
./mvnw verify -pl :dotcms-integration -Dcoreit.test.skip=false \
  -Dopensearch.phase=3 \
  -Dit.test=ContentletIndexAPIImplTest
```

**In scope (test decoupling — fix here):**
1. Replace the ES-direct search helper `indexSearch()` (~L1088-1110: `RestHighLevelClientProvider` + `SearchRequest`/`QueryBuilders`/`SearchSourceBuilder`/`SearchResponse`/`SearchHits`) with the vendor-neutral read path (`SearchAPI` / `ContentletAPI.search` / `ContentSearchResponse`) so the read-back hits whichever store the active phase routes to.
2. Replace `@Test(expected = ElasticsearchStatusException.class)` (~L1257, `testGetIndexDocumentCountWithInvalidIndexNameFails`) with the neutral exception the router throws (`DotRuntimeException` wrapping the OS index-not-found). Observed under phase 3: *expected `ElasticsearchStatusException` but was `DotRuntimeException`*.
3. Make the `.os` naming assertion in `activateDeactivateIndex` phase-aware (observed `ComparisonFailure: expected:<live_…[.os]> but was:<live_…[]>`).
4. Re-evaluate `testGetIndexDocumentCountSuccess` and `createContentIndexAndDelete` (`AssertionError`) against the store the active phase uses.
5. Remove the 8 `org.elasticsearch.*` imports (L80-87) once neutralized.

**Out of scope (genuine product gaps — file as separate bug issues):**
- `optimize` → `index_not_found_exception` looking up the index **without** the `.os` suffix (optimize-path routing/naming gap under phase 3).
[...]

PR #36558, opened 2026-07-13
fix(os-index): resolve getIndexDocumentCount physical name via the read provider (.os tag) (#36501)

fabrizzio-dotCMS:
## Proposed Changes
`ContentletIndexAPIImpl.getIndexDocumentCount(String)` resolved the physical name with `getNameWithClusterIDPrefix` (cluster prefix but **no** `.os` tag). Under ES→OS migration phase 2/3 the read provider is OpenSearch, whose indices are `.os`-tagged, so the lookup hit a non-existent index and threw. Now resolves via the read provider's own `toPhysicalName()` (ES: prefix only; OS: prefix + `.os`), mirroring the sibling `getIndexDocumentCount(String, IndexTag)` overload. Contributes to #36501.

## Verification
Phase 2, isolated env: `ContentletIndexAPIImplTest#testGetIndexDocumentCountSuccess` **1/0/0** (previously failing).

Refs #36501, #36320.

This PR fixes: #36501

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImpl.java
+++ b/dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImpl.java
@@ -3264,8 +3264,11 @@ public class ContentletIndexAPIImpl implements ContentletIndexAPI {
      */
     @Override
     public long getIndexDocumentCount(final String indexName) {
-        return router.readProvider().getIndexDocumentCount(
-                indexAPI.getNameWithClusterIDPrefix(indexName));
+        // Resolve the physical name through the read provider's own toPhysicalName so the OS read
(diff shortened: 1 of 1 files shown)


ihoffmann-dot (approved):

ihoffmann-dot:
PR [#36558](https://github.com/dotCMS/core/pull/36558) is a clean one-line backend fix. It reroutes getIndexDocumentCount through readProvider.toPhysicalName(indexName) instead of indexAPI.getNameWithClusterIDPrefix(indexName), so the OpenSearch read path receives the .os-tagged physical name (the old path added only the cluster prefix, causing the index to be reported as not found).

Verified correct: toPhysicalName is idempotent, the OS override appends .os, and the same pattern is already used at ContentletIndexAPIImpl.java:3319. No 🔴 or 🟡 issues; only stylistic 🟢 notes, discarded.

PR #36559, opened 2026-07-13
fix(os-mapping): getFieldMappingAsMap must return the neutral (ES-shaped) map (#36501)

fabrizzio-dotCMS:
## Proposed Changes
`MappingOperationsOS.getFieldMappingAsMap` serialized the whole OpenSearch `FieldMapping`, producing the wrapper shape `{"full_name":"...","mapping":{"<field>":{...}}}`. The vendor-neutral contract (established by `MappingOperationsES` returning `sourceAsMap()`, keyed by the leaf field name) was violated, so callers doing `map.get(fieldVariable)` got null and NPE'd under migration phase 2. Now returns the inner `mapping` sub-map so the OS result matches the ES shape. Contributes to #36501.

## Verification
Phase 2, isolated env: `FieldAPITest` **97/0/0** (incl. `test_SaveNewIndexedField_ShouldAddESMapping` and `test_SaveNewRelationshipField_ShouldAddESMapping`, previously NPE).

Refs #36501, #36320.

This PR fixes: #36501

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/content/index/opensearch/MappingOperationsOS.java
+++ b/dotCMS/src/main/java/com/dotcms/content/index/opensearch/MappingOperationsOS.java
@@ -118,6 +118,7 @@ public class MappingOperationsOS implements IndexMappingRestOperations {
     }
 
     @Override
+    @SuppressWarnings("unchecked")
     public Map<String, Object> getFieldMappingAsMap(final String index,
             final String fieldName) throws IOException {
         final String prefixed = physicalName(index);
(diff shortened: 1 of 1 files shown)


ihoffmann-dot (approved):

ihoffmann-dot:
The fix is correct, minimal, and well-scoped. It aligns the OpenSearch path with the established ES vendor-neutral contract (sourceAsMap(), keyed by leaf field name), the cast is guarded (instanceof Map), and the behavior is exercised by FieldAPITest on the OS route.

PR #36562, opened 2026-07-13
test(hardening): fix order-dependent tests surfaced by phased (ES→OS) runs (#36501)

fabrizzio-dotCMS:
## Proposed Changes
Two test-quality issues exposed by running the integration suite under a non-zero ES→OS migration phase (phase 2, reads from OpenSearch). Neither is a product bug. Contributes to #36501.

- **`RelationshipFactoryImplTest.testGetDBRelatedChildrenMultilingualContent`** — the DB-focused test's setup relied on index read-after-write consistency: a legacy-relationship checkin with no explicit relationships auto-preserves related content by *reading the index*, and under OS reads (phase 2) that read isn't read-after-write consistent mid-test, dropping one related version (3 vs 4). Added `TestDataUtils.waitForEmptyQueue()` after the relationship is established so the language-variant check-ins see a consistent index.
- **`ESSiteSearchAPITest.test_deleteOldSiteSearchIndices`** — used fixed index names with no cleanup, so a class retry (`rerunFailingTestsCount`) collided on a leftover index (`createSiteSearchIndex` is fail-hard on already-exists). Now uses unique, strictly increasing timestamps per run (relative order preserved).

## Verification
Both verified under phase 2 in an isolated worktree: `RelationshipFactoryImplTest` 1/0/0, `ESSiteSearchAPITest` 1/0/0.

## Not included
`PublisherTest.autoUnpublishContent` (initially bucketed here as flaky) is **excluded**: re-firing `IdentifierDateJob` with the correct content type on every poll still leaves the identifier's `SysExpireDate` null under phase 2 (deterministic across 4 reruns), so it is deeper than a race — a **candidate product gap in the OS auto-expire path**, being re-investigated separately under #36501.

Refs #36501.

This PR fixes: #36501

### DIFF
--- a/dotcms-integration/src/test/java/com/dotcms/content/elasticsearch/business/ESSiteSearchAPITest.java
+++ b/dotcms-integration/src/test/java/com/dotcms/content/elasticsearch/business/ESSiteSearchAPITest.java
@@ -236,29 +236,36 @@ public class ESSiteSearchAPITest {
      */
     @Test
     public void test_deleteOldSiteSearchIndices() throws IOException, DotDataException {
-        final String timestamp = String.valueOf(new Date().getTime());
-        siteSearchAPI.createSiteSearchIndex(ES_SITE_SEARCH_NAME + "_20230101000000", "Index1_deleteTest", 1);
-        siteSearchAPI.createSiteSearchIndex(ES_SITE_SEARCH_NAME + "_20230201000000", "", 1);
-        siteSearchAPI.createSiteSearchIndex(ES_SITE_SEARCH_NAME + "_20230301000000", "", 1);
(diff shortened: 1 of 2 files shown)


ihoffmann-dot (approved):

ihoffmann-dot:
Both changes are test-only hardening, well-reasoned and correct:

ESSiteSearchAPITest.java: Replaces fixed literal timestamps with dynamic strictly-increasing ones to avoid collisions on class retries where createSiteSearchIndex is fail-hard on already-exists.

RelationshipFactoryImplTest.java: Adds TestDataUtils.waitForEmptyQueue() to force indexing to settle before the language-variant check-ins, fixing a read-after-write inconsistency under the OpenSearch read path.

No 🔴 critical or 🟡 intermediate findings. No production code touched, assertions remain faithful refactors of the originals, and both changes are correctly commented.

PR #36589, opened 2026-07-15
fix(content-drive): field-based search follow-up fixes (#36384)

### LINKED ISSUE #36384: [Backend] API support for Content Drive field-based search/filter criteria

zJaaal:
## Proposed Changes

Follow-up fixes for the Content Drive **field-based search/filter** feature (issue #36384), on top of the merged frontend feature (#36452) and backend contract (#36463). Groups frontend + backend corrections found while testing the feature end-to-end.

### Frontend — date/time filter controls

The Date, Time, and Date-and-Time field filters were rebuilt to work around PrimeNG's `p-datePicker` range-mode limitations and to give a usable range UX:

- **Time field** — PrimeNG has no `timeOnly` range mode, so a single control can't express a time span. Now renders **two independent time-only pickers** (from / to). Either bound can be left empty (open-ended range); the "To" picker is gated until a "From" is chosen. The range is resolved against `_dotraw` as time-of-day (date-agnostic).
- **Date-and-Time field** — range + `showTime` crashes PrimeNG on hour change (`value.toDateString is not a function` — it calls `toDateString()` on the range array). Replaced with a **date-range calendar (no `showTime`) + two time-only pickers**; the picked date and time combine into each bound's instant.
- **Inverted-range validation** — an inline error is shown when From is after To (no silent swap); the invalid range is not applied. Uses seconds-since-midnight for time-only fields and full timestamp comparison for date-and-time.
- **Calendar stability** — capped the panel width so paging years no longer bounces the popover horizontally; centered the calendar.
- **Debounced search** — field-filter value changes are debounced so a settled change triggers a single search instead of one per keystroke/click.

### Frontend — lazy multiselect & relationship (Tag / Category / Relationship)

- **Duplicate rows** — Tag/Category options could render twice (on scroll and after a search) because the virtual-scroll listbox had no `dataKey`. Added `dataKey="value"` so PrimeNG tracks options uniquely.
- **Category labels on reload** — a stored Category filter showed the raw inode instead of its name after a cold reload. The chip now resolves each stored inode to its category name via `GET /api/v1/categories/{idOrKey}` (resolves by inode) and caches the result.
- **Relationship title on reload** — relationships persist by **identifier** (version-stable), so the chip resolves the stored identifier to a contentlet (`getContentletByInode`, which accepts identifier or inode) to show its title and preselect it in the picker. Resolution is guarded with `catchError` so a failed lookup degrades gracefully.

### Frontend — robustness (review follow-ups)

- Defer the first search until `userSearchable` field metadata has loaded on a cold URL restore, so restored field filters aren't dropped from the initial request. The search-effect reads `userSearchableFieldsLoaded` tracked so the deferral reliably releases once metadata arrives — even when the built request is structurally identical (e.g. a restored `us.*` key for an ineligible/removed field), which previously left the portlet stuck in `LOADING`.
- Gate `showFolders` on the presence of `userSearchable` filters.
- Type-guard the date-range cast in `serializeUserSearchableValue`.
- Distinct lazy-load error state in the multiselect.
- **Always-on in-development banner** — removed the dismiss (X) control and its `localStorage` persistence; the banner now always renders (message + learn-more link).

### Backend

- **Normalize ISO date-range bounds for ES** — the resolver parses incoming `{ from, to }` bounds flexibly (`Instant` / `OffsetDateTime` / `LocalDateTime` / `LocalDate`, with a `DateUtil` fallback) and formats them to the ES query pattern `yyyy-MM-dd'T'HH:mm:ss`; open/blank bounds become `*`, and unparseable bounds are escaped (with a normalization log) rather than breaking the query.
- **Accept single-value (scalar) filters on list fields** — Multi-Select / Checkbox / Tag / Category / Relationship now accept a single scalar value in addition to an array, so filtering by one option validates and resolves correctly.

### Content Drive — "New" button & folder navigation

Separate Content Drive fixes bundled on this branch (the related sidebar right-click context-menu work is tracked in #36595).

- **New content is created in the current folder** — the "New" menu carries the browsed folder to both editors:
  - New editor (`content/new/:type`): a `folderPath` (`hostname/path`) query param pre-fills the Host/Folder field.
  - Legacy editor (`c/content/new/:type`): `folder=<inode>` is appended to the resolved action URL by `DotCreateContentletResolver`.
  - The folder `inode` (from `/api/v1/folder/byPath`) is threaded through `DotFolder` and the sidebar tree node data.
- **Cold-load selection regression** — reloading with a `path` in the URL cleared the sidebar selection back to root instead of opening the deep-linked folder. `setPathEffect` no longer clobbers the URL-restored path while the sidebar tree is still resolving; it waits for `sidebarLoading` to settle.

### Testing

- Frontend: `pnpm nx test portlets-content-drive` — green (772 tests). Added/updated specs for the Time & Date-and-Time controls (two-picker render, change handlers driven through the UI, inverted-range inline error + no-patch guard, the Date-and-Time full-instant invalidity branch, valid range), debounce coalescing, plain-Date range serialization, single/multi/lazy write paths, the relationship/category resolution error fallbacks, lazy-multiselect duplicate fix + Category label resolution, relationship identifier resolution/preselect, the cold-load field-metadata deferral (including the deadlock-release regression for an ineligible `us.*` field), `showFolders` gating, the date-range serialization guard, and the always-on banner (renders, no dismiss control). New-button folder params (navigation service), legacy action-URL append (`DotCreateContentletResolver`), folder-context resolution (content-type selector dialog), inode threading (`createTreeNode`), and the `setPathEffect` guard (shell) also covered.
- Backend: `ContentDriveFieldFilterResolverTest` and `ContentDriveFieldFilterTest` extended for date-normalization and scalar-on-list-field paths.

This PR fixes: #36384


dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
```diff
@@ -1297,6 +1305,58 @@ private String fieldCriteriaLuceneValue(final FieldSearchCriteria criteria) {
...
+
+    /**
+     * Best-effort parse of a date bound across the shapes a client may send: ISO-8601 instant
+     * (with offset/{@code Z}), ISO offset date-time, ISO local date-time, ISO date-only, and finally
+     * the dotCMS content date formats. Returns {@code null} when none match.
+     */
+    private Date parseFlexibleDate(final String value) {
+        Date date = Try.of(() -> Date.from(Instant.parse(value))).getOrNull();
+        if (null == date) {
+            date = Try.of(() -> Date.from(OffsetDateTime.parse(value).toInstant())).getOrNull();
```
dario-daza, line 1414: Shall we add a log to know which path was used build the date?

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
```diff
@@ -1297,6 +1305,58 @@ private String fieldCriteriaLuceneValue(final FieldSearchCriteria criteria) {
...
+     * @param raw The raw bound value.
+     * @return The normalized bound, or the original value if it isn't a recognizable date.
+     */
+    private String normalizeDateBound(final String raw) {
+        final String value = raw.trim();
+        if (value.isEmpty() || "*".equals(value)) {
+            return "*";
+        }
+        final Date parsed = parseFlexibleDate(value);
+        return null != parsed ? new SimpleDateFormat(ES_QUERY_DATE_PATTERN).format(parsed) : value;
```
dario-daza: If the parsed value is null, we return the raw string unchanged. Should we validate and escape special characters in case a client attempts to inject something that isn’t a date? or how could I handle this scenario?

dario-daza (changes requested):

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
ihoffmann-dot: Good catch! Fixed ✅ When the bound can't be parsed as a date, we no longer pass the raw string through, it's escaped with ESUtils.escape, so a crafted value can't alter the query structure, an escaped non-date just matches nothing

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
ihoffmann-dot: Done!

dario-daza (approved):

Issue #36541, opened 2026-07-13
BinaryExporterServlet returns 401 (redirect-to-login) instead of 403 for authenticated front-end users lacking READ on a protected file asset


dsolistorres:
## Problem Statement

When an authenticated **front-end** user (logged in, but not a backend/admin user) requests a protected **file asset** (e.g. a PDF) that they do **not** have READ permission on, dotCMS returns **HTTP 401 Unauthorized** and sets `REDIRECT_AFTER_LOGIN` in the session, instead of a clean **HTTP 403 Forbidden**.

**Page** assets already behave correctly — an authenticated user without permission gets a clean 403 via `VelocityServlet`. Only **file/binary** assets diverge.

**Impact:** On a plain site the user simply receives the wrong status code (401 + redirect intent instead of 403). On a **SAML-protected** site this is more severe: the 401 causes the front-end SAML interceptor to re-trigger IdP authentication on every request, producing an **infinite redirect loop** — `ERR_TOO_MANY_REDIRECTS` in Chrome, or a constantly-refreshing white page in other browsers — leaving the protected file effectively inaccessible for the affected user.

This is the **front-end file-asset counterpart** to the backend-only issue #32365 (SAML users without a backend role). This issue is scoped **only** to the `BinaryExporterServlet` status-code behavior; any separate problem in the SAML interceptor logic will be filed independently if found.

**Root cause (code):**

In `dotCMS/src/main/java/com/dotmarketing/servlets/BinaryExporterServlet.java`, the `catch (DotSecurityException)` block (~lines 675–686) branches only on `isLoggedToBackend()`:

```java
} catch (DotSecurityException e) {
    if (req.getSession() != null) {
        if (WebAPILocator.getUserWebAPI().isLoggedToBackend(req)) {
            req.getSession().removeAttribute(WebKeys.REDIRECT_AFTER_LOGIN);
            resp.sendError(HttpServletResponse.SC_FORBIDDEN);      // 403 — backend users
        } else {
            final String requestUri = (String) req.getAttribute("javax.servlet.forward.request_uri");
            req.getSession().setAttribute(WebKeys.REDIRECT_AFTER_LOGIN, requestUri);
            resp.sendError(HttpServletResponse.SC_UNAUTHORIZED);   // 401 — ANY non-backend user
        }
    }
}
```

[...]

PR #36590, opened 2026-07-15
fix(permissions): access-denied 403 page + auth/authz split + SAML loop-guard (#32365)

### LINKED ISSUE #32365: Add Default "Access Denied" Page for SAML Users Without Backend Role
### Problem Statement

When a user logs into dotCMS via SAML SSO successfully but does not have the dotcms_back_end_user role (i.e., lacks permission to access the admin UI), they are currently redirected to a blank dotCMS backend page that continuously reloads. This creates confusion and appears broken to the end user.

### Steps to Reproduce

1. Set up SAML SSO in dotCMS.

2. Attempt to log in via SAML with a user that exists in the SSO provider but does not have the dotcms_back_end_user role.

3. User is redirected to the backend admin page.

4. The page remains blank and continuously refreshes.

### Acceptance Criteria

If a SAML-authenticated user lands on the backend admin page but lacks the required backend role, dotCMS should display a friendly "Access Denied" message (or similar), explaining that they do not have permissions to access the admin area. 

### dotCMS Version

current

### Proposed Objective

Customer Support

### Proposed Priority

Priority 3 - Average

### External Links... Slack Conversations, Support Tickets, Figma Designs, etc.

https://dotcms.freshdesk.com/a/tickets/31277

### Assumptions & Initiation Needs

_No response_

### Quality Assurance Notes & Workarounds

_No response_

### Sub-Tasks & Estimates

_No response_

dsolistorres:
## Problem

On SAML-protected sites, a **SAML user without a back-end role** (front-end-only) who lands on a back-end URL (e.g. `/dotAdmin/`) — or any authenticated front-end user requesting a resource they lack READ on — was bounced back to the IdP repeatedly. Because the IdP re-authenticates the already-signed-in user but that never grants the missing access, the result was an **infinite redirect loop** (`ERR_TOO_MANY_REDIRECTS`) with no default "Access Denied" landing. This is the counterpart to the file-asset status-code fix in #36541.

## What this PR does

Cleanly separates **authentication** (401 → start login) from **authorization** (403 → Access Denied, never re-authenticate), and gives those users a real destination:

- **Centralized 401-vs-403 decision** — new `SecurityUtils.sendPermissionDenied(user, uri, req, resp)`: authenticated (non-anonymous) → **403** and clears `REDIRECT_AFTER_LOGIN`; anonymous/null → **401** + sets it. Adopted by `BinaryExporterServlet`, `SpeedyAssetServlet`, and `CMSUrlUtil` (which previously keyed only on `user == null`).
- **Default 403 "Access Denied" page** in `custom-error-page.jsp` with a provider-agnostic **sign-out** link: SAML SLO (`/api/v1/dotsaml/logout/{host}`) when the host has SAML enabled (Entra ID, Okta, …), else `/dotCMS/logout`. Honors the existing per-site `/cms403Page` vanity override.
- **Deterministic SAML loop-guard** in `SamlWebInterceptor`: if auto-login resolved a SAML user yet the request is still not logged in for its destination (`isAutoLogin() && isNotLogged()`), return a clean **403** instead of redirecting to the IdP again. This trips on the first futile cycle and is immune to the per-callback session churn caused by SameSite cookies.

Anonymous (not-logged-in) users still get the legitimate 401 → login flow; back-end/admin users still get 403.

## Fixes / Refs

- Fixes #32365
- Refs #36541

## Testing

- Unit: `SecurityUtilsTest` (401-vs-403 resolver) — 19/19 green.
- Manual (native login **and** real Entra ID): authenticated-but-unauthorized front-end user → clean **403** on pages and file assets (no `REDIRECT_AFTER_LOGIN`); anonymous → 401 → login; SAML front-end user hitting a back-end URL now gets a **403 Access Denied with a sign-out link** instead of an infinite loop.

## Follow-up (tracked separately)

- Preserve the original front-end URL through the SAML round-trip (default RelayState → original request, with open-redirect validation) so users return to the exact page they requested.
- Fix `DotSamlResource.logoutGet/logoutPost` missing `return` after redirect (spurious `DoesNotExistException`).

This PR fixes: #32365

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/auth/providers/saml/v1/DotSamlResource.java
+++ b/dotCMS/src/main/java/com/dotcms/auth/providers/saml/v1/DotSamlResource.java
@@ -430,6 +430,9 @@ public class DotSamlResource implements Serializable {
 							SamlName.DOT_SAML_LOGOUT_SERVICE_ENDPOINT_URL,
 							()-> "/dotAdmin/#/public/logout");
 					RedirectUtil.sendRedirectHTML(httpServletResponse, logoutPath);
+					// The redirect commits the response; return so we don't fall through to the
+					// unconditional DoesNotExistException throw below (which would write a second response).
+					return;
 
 
 				}
@@ -485,6 +488,9 @@ public class DotSamlResource implements Serializable {
 					
 					
 					RedirectUtil.sendRedirectHTML(httpServletResponse, logoutPath);
+					// The redirect commits the response; return so we don't fall through to the
+					// unconditional DoesNotExistException throw below (which would write a second response).
+					return;
 
 				}
 			} finally {
--- a/dotCMS/src/main/java/com/dotcms/filters/interceptor/saml/SamlWebInterceptor.java
+++ b/dotCMS/src/main/java/com/dotcms/filters/interceptor/saml/SamlWebInterceptor.java
@@ -162,6 +162,24 @@ public class SamlWebInterceptor implements WebInterceptor {
                             // if the auto login couldn't log in the user, then send it to the IdP login page (if it is not already logged in).
                             if (null == session || !autoLoginResult.isAutoLogin() || this.samlWebUtils.isNotLogged(request)) {
 
+                                // Loop-guard: if auto-login actually resolved a SAML user (isAutoLogin)
+                                // yet the request is STILL not logged in for this destination, re-authenticating
+                                // will resolve the very same user and produce the same result — that is
+                                // an infinite IdP redirect loop (e.g. a front-end-only SAML user sent
+                                // to a back-end URL like /dotAdmin). Break it
+                                // with a clean 403 instead of bouncing to the IdP again.
+                                if (autoLoginResult.isAutoLogin() && this.samlWebUtils.isNotLogged(request)) {
+
+                                    Logger.warn(this, ()-> "SAML auth redirect loop detected for URI '"
+                                            + request.getRequestURI() + "': the user is authenticated but not "
+                                            + "authorized for this destination. Returning 403 instead of "
+                                            + "redirecting to the IdP again.");
+                                    SecurityLogger.logInfo(this.getClass(), "SAML auth redirect loop broken for URI: "
+                                            + request.getRequestURI());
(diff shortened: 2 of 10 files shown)


dsolistorres:
Thanks for the review — addressed all four items in `8e2a308ebf`:

**🟠 High — loop-guard skipped clearing `REDIRECT_AFTER_LOGIN`** (`SamlWebInterceptor`)
The authenticated-403 guard now clears `WebKeys.REDIRECT_AFTER_LOGIN` before returning the 403, so a stale redirect can't survive the session — including for `/api/*` destinations where `custom-error-page.jsp` returns before its 403 branch would clear it. I cleared it **inline** rather than routing through `SecurityUtils.sendPermissionDenied` on purpose: the resolved `User` isn't in scope at that point, and this path must **never** fall to the helper's 401 branch (which would re-set `REDIRECT_AFTER_LOGIN` and re-enter the very loop we're breaking). Added a test assertion that the attribute is cleared.

**🟡 Medium — `logoutGet`/`logoutPost` fall-through** (`DotSamlResource`)
Good call that the new sign-out link makes this user-facing on every click — pulled the one-line fix into this PR instead of deferring it: `return;` after `RedirectUtil.sendRedirectHTML(...)` in both methods, so they no longer fall through to the unconditional `DoesNotExistException` throw over the committed response. (Removed from the follow-up ticket #36591.)

**🟡 Medium — `SpeedyAssetServlet` session allocation** (`SpeedyAssetServlet`)
Reverted this file entirely to the original sessionless `sendError(SC_UNAUTHORIZED)` on the auth-time `SecurityException` — no new `HttpSession` / `REDIRECT_AFTER_LOGIN` write on this high-traffic static-asset endpoint. Asset authorization is still enforced downstream by `BinaryExporterServlet` via the `/contentAsset` forward, so nothing is lost.

**🟡 Medium — FQN-inline call in `CMSUrlUtil`**
Added `import com.dotcms.util.SecurityUtils;` and call it unqualified at both sites, matching `BinaryExporterServlet`.

The two `SpeedyAssetServlet` notes from the earlier review are moot now that the file is reverted. Unit tests green (`SamlWebInterceptorLoopGuardTest` 2/2 incl. the new clear-redirect assertion, `SecurityUtilsTest` 19/19).

PR #36606, opened 2026-07-16
feat(cicd): 36605 changelog auto-publish to dev site 

sfreudenthaler:
## What

End-to-end implementation of **automated release changelog publishing to dev.dotcms.com**, spec-first (speckit flow: spec → plan → tasks → TDD implementation → adversarial verification), plus two CI-hygiene cleanups. All pieces are independently mergeable — no dependency on #36416 or any other open PR.

### 1. Spec (`specs/36605-changelog-site-publish/`)
Spec (13 FRs incl. human-edit protection, no-auto-backfill, older-version patch semantics; LTS/CLI descoped v1) plus `data-model.md` (live-verified `Dotcmsbuilds` field contract). The working artifacts (plan/tasks/research/checklists — incl. Constitution Check and ADR-gate consultation, 3 relevant ADRs, none conflicting; 44/44 TDD tasks complete) were used during development and deliberately not committed.

### 2. Implementation
- **`.github/actions/core-cicd/changelog-publisher/`** — new Python CLI (mirrors the evergreen-tracks package pattern): searches `Dotcmsbuilds` by version (`minor_dotraw`), upserts create/update-in-place, fires the System Workflow Publish action with `disabledWYSIWYG:["releaseNotes"]` (markdown-preservation guard). Dry-run by default; `--apply` to fire; `--force` (manual only) to override human-edit protection, which skips any entry last modified by a non-service-account user (immutable `modUser` id comparison, display-name fallback). Exit contract: 0 = success/protective-skip (`::changelog-skip::` marker), 2 = validation (out-of-scope version rejected before any network call), 1 = runtime failure. 27 pytest (responses-mocked) green.
- **Single-inference design**: there is no separate site-notes generation. The existing `gather-release-data/prompt-template.md` becomes the one shared editorial template (site style: prose intro, no emoji, site section set, `[[#N](issues url)]` links), the GitHub release body is the single shared artifact, and the publisher injects the site's per-version heading anchors mechanically at publish time. One inference per release; the site path has zero Claude dependency at publish time; a hand-edited GitHub release re-syncs the site on re-run. Note: this restyles the GitHub releases page to the site's editorial format (deliberate).
[...]

PR #36611, opened 2026-07-16
fix(ci): link-issue check must not fail on transient GitHub API errors

### LINKED ISSUE #36610: link-issue check fails on transient GitHub API errors — unguarded curl|jq in comment-maintenance steps
## Symptom

The required `Add Issue to PR / link-issue` check fails intermittently with:

```
jq: parse error: Invalid numeric literal at line 1, column 10
##[error]Process completed with exit code 5.
```

even when the PR is correctly linked (the run resolves `ISSUE_NUMBER` fine and the linked issue has a `Team :` label). Observed on PR #36606 (3 consecutive failures 2026-07-16 ~22:23–22:29 UTC, after 12 straight passes with identical linkage) and PR #36608 (failure 19:56, pass on retry 20:15). During the same window, direct `gh api` calls from a workstation intermittently returned HTML error bodies (`invalid character '<' looking for beginning of value`) with rate limits nowhere near exhausted — i.e. a transient GitHub API brownout.

## Root cause

`.github/workflows/issue_comp_link-issue-to-pr.yml` pipes raw `curl -s` output into `jq` in three steps with no HTTP-status or JSON-validity guard:

1. "Remove failure comment if resolved" — `existing_comments=$(curl -s …/comments)` → `echo … | jq -r`
2. "Remove issue-validation failure comment if resolved" — same pattern
3. "Get existing issue comments" — `curl -s …/comments | jq -c . > /tmp/issue_comments.json` (this is the step that failed on #36606: its env group ends with `ISSUE_NUMBER`, matching the crash location in the logs)

When the API returns a transient non-JSON error body (5xx HTML, secondary-rate-limit text), `jq` exits 5 and the `bash -e` step fails the whole required check. `curl -s` without `-f` also happily forwards HTTP error bodies as if they were data.

## Fix

Make the three fetches brownout-tolerant: `curl -sf --retry 3 --retry-all-errors`, and fall back to an empty JSON array (`[]`) if the fetch or parse still fails — the affected steps are best-effort comment maintenance, so the correct degraded behavior is "skip the cleanup / treat as no existing comments", never "fail the merge-gating check".

## Impact

Org-wide: any PR whose link-issue run coincides with a GitHub API hiccup gets a red required check and a confusing signal ("issue linking required" when linkage is fine). Retry usually passes, but authors have no way to know that.

sfreudenthaler:
## What

Makes the merge-gating `link-issue` check resilient to transient GitHub API errors. Three comment-maintenance steps in `issue_comp_link-issue-to-pr.yml` piped raw `curl -s` output straight into `jq`; during an API brownout the 5xx/HTML error body made `jq` exit 5 and fail the required check even on correctly linked PRs (observed on #36606 — 3 consecutive failures — and #36608 on 2026-07-16).

## How

Same minimal guard at all three sites:
- `curl -sf --retry 3 --retry-all-errors` (fail on HTTP errors, retry through brownouts)
- fall back to an empty JSON array if the fetch or a `jq -e` validity check still fails

These steps are best-effort comment maintenance, so the correct degraded behavior is "skip cleanup / post a fresh comment", never "fail the check". Verified: YAML parses; guard pattern exercised against a live 502 endpoint and against non-JSON garbage — both degrade cleanly.

Closes: #36610

https://claude.ai/code/session_012ogYXVeZsBVUHkwzdStYPJ

This PR fixes: #36610

### DIFF
--- a/.github/workflows/issue_comp_link-issue-to-pr.yml
+++ b/.github/workflows/issue_comp_link-issue-to-pr.yml
@@ -256,12 +256,15 @@ jobs:
           pr_url="$PR_URL"
           pr_number=$(echo "$pr_url" | grep -o '[0-9]*$')
 
-          # Check for existing failure comment
-          existing_comments=$(curl -s \
+          # Check for existing failure comment. Best-effort cleanup: a transient API
+          # error (5xx/HTML body) must degrade to "no comment found", never fail the
+          # merge-gating check by feeding non-JSON to jq (#36610).
+          existing_comments=$(curl -sf --retry 3 --retry-all-errors \
             -H "Accept: application/vnd.github+json" \
             -H "Authorization: Bearer ${{ env.GH_TOKEN }}" \
             -H "X-GitHub-Api-Version: 2022-11-28" \
-            "https://api.github.com/repos/${{ github.repository }}/issues/$pr_number/comments")
+            "https://api.github.com/repos/${{ github.repository }}/issues/$pr_number/comments") || existing_comments="[]"
+          echo "$existing_comments" | jq -e . >/dev/null 2>&1 || existing_comments="[]"
 
           # Find failure comment by looking for the distinctive header
           failure_comment_id=$(echo "$existing_comments" | jq -r '.[] | select(.user.login == "github-actions[bot]" and (.body | contains("❌ Issue Linking Required"))) | .id' | head -1)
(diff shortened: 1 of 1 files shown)


dcolina (approved):
Looks good the fallback added to avoid the error when get API errors. 👍🏽

PR #36618, opened 2026-07-17
36597 uve style editor styling non default persona content fails with content not found

### LINKED ISSUE #36597: UVE Style Editor: styling non-default Persona content fails with HTTP 400 CONTENT_NOT_FOUND
## Problem Statement

In UVE (Universal Visual Editor), using the **Style Editor** to style content that is personalized for a **non-default Persona** fails with an HTTP 400 Bad Request / `CONTENT_NOT_FOUND` error.

The backend error shows `personalization=dot:default` even though the content being styled is personalized for a *different*, non-default persona — indicating the persona tag isn't reaching the save request. This blocks styling of any personalized content variant; users can only style content under the default persona.

**Impact:** Any team using Personas + the UVE Style Editor together cannot style personalized content variants, a core combination of two dotCMS features.

## Steps to Reproduce

1. Open a page in UVE that has a container with content personalized for a non-default Persona (e.g., "Returning Visitor").
2. Switch the active persona in UVE to that non-default Persona.
3. Open the Style Editor for the personalized content in that container.
4. Change a style property and save.

**Expected:** The style change saves successfully and applies to the content under the selected (non-default) persona's personalization key.

**Actual:** Save fails with:

```
com.dotcms.rest.exception.BadRequestException: HTTP 400 Bad Request
ResponseEntityView{errors=[ErrorEntity{errorCode='CONTENT_NOT_FOUND', message='Contentlet: c12464812d5aff9a5f28e22f9605cac7 not found for page=92bfbfff74adad3a8ceb8c7b37917291, container=969a9dd1b64f20a5e4f2ad0743cdb84c, uuid=1, personalization=dot:default, variant=DEFAULT.', fieldName='null'}], entity=, messages=[], i18nMessagesMap={}, pagination=null}
```

Note the mismatch: even with a non-default persona active, the backend resolves `personalization=dot:default` — the persona tag isn't making it into the save call.

### Technical note (initial research)

`PageResourceHelper.updateContentletStyles()` (`dotCMS/src/main/java/com/dotcms/rest/api/v1/page/PageResourceHelper.java:786-845`) correctly derives the personalization key from `containerEntry.getPersonaTag()`, falling back to `dot:default` only when that field is unset in the request body (see `PageContainerForm.java:48-71` for the expected `personaTag` JSON field).

This suggests the UVE frontend's style-save call is not including the active persona's tag in the request payload when saving styles for personalized content. The exact frontend call site was not pinpointed during initial triage and should be confirmed as part of the fix — the backend contract itself appears correct.

## Acceptance Criteria

- [ ] Selecting a non-default persona in UVE and saving a style change for content personalized under that persona no longer throws HTTP 400 / `CONTENT_NOT_FOUND`
- [ ] Style changes saved for non-default-persona content are persisted under that persona's correct personalization key (verified via API/DB — not `dot:default`)
- [ ] Reloading the page under the same non-default persona shows the newly applied style
- [ ] Styling content under the DEFAULT persona continues to work with no regression
- [ ] Switching personas mid-session (default → non-default → default) and styling content under each does not cross-contaminate styles between personalization keys
- [ ] If the persona tag cannot be resolved (e.g., persona deleted), the UI surfaces a clear error message instead of a generic 400

## dotCMS Version

Latest from main branch

## Severity

## Links

NA

dario-daza:
### Proposed Changes
* Fixed a bug in the UVE Style Editor where styling content that was personalized for a specific Persona (i.e. not the "Default Visitor") failed with an error and the change wasn't saved.
* The editor now correctly keeps track of which Persona you're viewing while you make style changes, so the update is saved against that Persona's version of the content instead of always falling back to the default one.

### Additional Info
Previously, saving a style change while viewing a page under a non-default Persona threw an error ("Contentlet not found...") because the save request always looked for the default Persona's version of the content, even when a different Persona was selected in the editor. This fix makes sure the currently selected Persona is included when the style change is saved, so it lands on the correct personalized content.

This PR fixes: #36597

This PR fixes: #36597

### DIFF
--- a/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-palette/components/dot-uve-style-editor-form/dot-uve-style-editor-form.component.spec.ts
+++ b/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-palette/components/dot-uve-style-editor-form/dot-uve-style-editor-form.component.spec.ts
@@ -27,6 +27,7 @@ import {
 import { UVE_STATUS } from '../../../../../shared/enums';
 import { ActionPayload, SelectedContentlet } from '../../../../../shared/models';
 import { UVEStore } from '../../../../../store/dot-uve.store';
+import { PageData } from '../../../../../store/features/editor/models';
 import { PageType } from '../../../../../store/models';
 
 // Workaround: the `schema` input alias causes a compilation error when used directly.
@@ -43,6 +44,7 @@ type MockUveStore = {
     editorSelected: ReturnType<typeof signal<SelectedContentlet | null>>;
     pageAsset: ReturnType<typeof computed<DotCMSPageAsset | null>>;
     pageType: ReturnType<typeof signal<PageType>>;
+    $pageData: ReturnType<typeof signal<PageData>>;
     saveStyleEditor: jest.Mock;
     rollbackPageAssetResponse: jest.Mock;
     addCurrentPageToHistory: jest.Mock;
@@ -189,6 +191,13 @@ describe('DotUveStyleEditorFormComponent', () => {
                 return pageAsset ? { ...pageAsset, clientResponse: pageAsset } : null;
             }),
             pageType: signal(PageType.HEADLESS),
+            $pageData: signal<PageData>({
+                containers: [],
+                personalization: 'dot:default',
+                id: 'test-page',
+                languageId: 1,
+                personaTag: undefined
+            }),
             saveStyleEditor: jest.fn().mockReturnValue(of({})),
             rollbackPageAssetResponse: jest.fn().mockReturnValue(true),
             addCurrentPageToHistory: jest.fn(),
--- a/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-palette/components/dot-uve-style-editor-form/dot-uve-style-editor-form.component.ts
+++ b/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-palette/components/dot-uve-style-editor-form/dot-uve-style-editor-form.component.ts
@@ -479,7 +479,8 @@ export class DotUveStyleEditorFormComponent {
                 contentletIdentifier: activeContentlet.contentlet.identifier,
                 styleProperties: filteredFormValues,
                 pageId: activeContentlet.pageId,
-                containerUUID: activeContentlet.container.uuid
+                containerUUID: activeContentlet.container.uuid,
+                personaTag: this.#uveStore.$pageData().personaTag
             })
             .pipe(
                 map((): SaveResult => ({ ok: true, isTraditionalPage })),
--- a/core-web/libs/portlets/edit-ema/portlet/src/lib/services/dot-page-api/dot-page-api.service.spec.ts
+++ b/core-web/libs/portlets/edit-ema/portlet/src/lib/services/dot-page-api/dot-page-api.service.spec.ts
@@ -239,5 +239,36 @@ describe('DotPageApiService', () => {
                 }
             ]);
         });
+
+        it('should include personaTag in the payload when styling non-default-persona content', () => {
+            const payload = {
+                pageId: 'test-page-123',
+                containerIdentifier: 'container-id-456',
+                containerUUID: 'container-uuid-789',
+                contentletIdentifier: 'contentlet-id-abc',
+                styleProperties: {
+                    'font-size': '16px'
+                },
+                personaTag: 'persona.returning_visitor'
+            };
+
+            spectator.service.saveStyleProperties(payload).subscribe();
+
+            const { request } = spectator.expectOne(
+                '/api/v1/page/test-page-123/styles',
+                HttpMethod.PUT
+            );
+
+            expect(request.body).toEqual([
+                {
+                    identifier: 'container-id-456',
+                    uuid: 'container-uuid-789',
+                    personaTag: 'persona.returning_visitor',
+                    'contentlet-id-abc': {
+                        'font-size': '16px'
+                    }
+                }
+            ]);
+        });
     });
 });
--- a/core-web/libs/portlets/edit-ema/portlet/src/lib/services/dot-page-api/dot-page-api.service.ts
+++ b/core-web/libs/portlets/edit-ema/portlet/src/lib/services/dot-page-api/dot-page-api.service.ts
@@ -107,6 +107,7 @@ export class DotPageApiService {
      * @param {Record<string, unknown>} payload.styleProperties - Style properties to apply.
      * @param {string} payload.pageId - The page ID where styles are being saved.
      * @param {string} payload.containerUUID - UUID of the container.
+     * @param {string} [payload.personaTag] - Persona key tag to personalize this style update for.
      * @returns {Observable<unknown>} Observable that completes when properties are saved.
      * @memberof DotPageApiService
      */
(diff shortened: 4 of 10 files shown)


zJaaal:
### 🟠 Possible default-visitor regression — worth verifying before merge

Nice fix for the non-default persona path. One case I'd double-check: the **Default Visitor**.

The backend converts the tag here ([`PageResourceHelper.java:207-209`](https://github.com/dotCMS/core/blob/main/dotCMS/src/main/java/com/dotcms/rest/api/v1/page/PageResourceHelper.java#L207)):

```java
final String personalization = UtilMethods.isSet(containerEntry.getPersonaTag()) ?
        Persona.DOT_PERSONA_PREFIX_SCHEME + StringPool.COLON + containerEntry.getPersonaTag() : // "dot:persona:" + tag
        MultiTree.DOT_PERSONALIZATION_DEFAULT;                                                   // "dot:default"
```

For the Default Visitor, `$pageData().personaTag` is **not** empty — it's `"dot:persona"`:
- `DEFAULT_PERSONA.keyTag = 'dot:persona'` (frontend `consts.ts`)
- `getDefaultPersona()` sets `keyTag = DOT_PERSONA_PREFIX_SCHEME = "dot:persona"` (`PersonaAPIImpl.java:82`)

So `UtilMethods.isSet("dot:persona")` is `true`, and the personalization resolves to **`dot:persona:dot:persona`** instead of `dot:default`. Before this PR the default case used the hardcoded `null` → `dot:default`, so this changes the target for the most common (default) case and could reintroduce the same "not found" symptom — or write styles to a personalization the read path never looks at.

It likely wasn't caught because:
- the integration test only covers a *real* non-default persona, and
- the frontend unit test mocks `personaTag: undefined` for `$pageData`, so it never exercises the real `"dot:persona"` value.

**Suggestions (pick one):**
- **Frontend (simplest):** only send the tag for a real persona — gate on the already-computed `isDefaultPersona` so the default sends `undefined`.
- **Backend (most robust):** treat the bare prefix as default: `UtilMethods.isSet(tag) && !Persona.DOT_PERSONA_PREFIX_SCHEME.equals(tag)`. This also protects the existing `saveContent` content-add path, which sends the same `"dot:persona"` for the default visitor.

Either way, adding a **default-visitor** test asserting the row lands on `dot:default` would lock this down.

> Side note: the drag-and-drop content flow (`utils/index.ts` → `pageContainer.personaTag`) already sends `"dot:persona"` for the default through this same backend method. If that flow works today, it's worth checking whether default content is already being stored under `dot:persona:dot:persona` — that tells you which of the two fixes is the correct one.

_Reviewed and drafted by Claude (Claude Code), on behalf of @zJaaal._

dario-daza:
> ### 🟠 Possible default-visitor regression — worth verifying before merge
> Nice fix for the non-default persona path. One case I'd double-check: the **Default Visitor**.
> 
> The backend converts the tag here ([`PageResourceHelper.java:207-209`](https://github.com/dotCMS/core/blob/main/dotCMS/src/main/java/com/dotcms/rest/api/v1/page/PageResourceHelper.java#L207)):
> 
> ```java
> final String personalization = UtilMethods.isSet(containerEntry.getPersonaTag()) ?
>         Persona.DOT_PERSONA_PREFIX_SCHEME + StringPool.COLON + containerEntry.getPersonaTag() : // "dot:persona:" + tag
>         MultiTree.DOT_PERSONALIZATION_DEFAULT;                                                   // "dot:default"
> ```
> 
> For the Default Visitor, `$pageData().personaTag` is **not** empty — it's `"dot:persona"`:
> 
> * `DEFAULT_PERSONA.keyTag = 'dot:persona'` (frontend `consts.ts`)
> * `getDefaultPersona()` sets `keyTag = DOT_PERSONA_PREFIX_SCHEME = "dot:persona"` (`PersonaAPIImpl.java:82`)
> 
> So `UtilMethods.isSet("dot:persona")` is `true`, and the personalization resolves to **`dot:persona:dot:persona`** instead of `dot:default`. Before this PR the default case used the hardcoded `null` → `dot:default`, so this changes the target for the most common (default) case and could reintroduce the same "not found" symptom — or write styles to a personalization the read path never looks at.
> 
> It likely wasn't caught because:
> 
> * the integration test only covers a _real_ non-default persona, and
> * the frontend unit test mocks `personaTag: undefined` for `$pageData`, so it never exercises the real `"dot:persona"` value.
> 
> **Suggestions (pick one):**
> 
> * **Frontend (simplest):** only send the tag for a real persona — gate on the already-computed `isDefaultPersona` so the default sends `undefined`.
> * **Backend (most robust):** treat the bare prefix as default: `UtilMethods.isSet(tag) && !Persona.DOT_PERSONA_PREFIX_SCHEME.equals(tag)`. This also protects the existing `saveContent` content-add path, which sends the same `"dot:persona"` for the default visitor.
> 
> Either way, adding a **default-visitor** test asserting the row lands on `dot:default` would lock this down.
> 
> > Side note: the drag-and-drop content flow (`utils/index.ts` → `pageContainer.personaTag`) already sends `"dot:persona"` for the default through this same backend method. If that flow works today, it's worth checking whether default content is already being stored under `dot:persona:dot:persona` — that tells you which of the two fixes is the correct one.
> 
> _Reviewed and drafted by Claude (Claude Code), on behalf of @zJaaal._

Good catch, and worth double-checking carefully given how easy it'd be to reintroduce the exact bug we just fixed. I traced through the actual render path to see whether $pageData().personaTag really can be "dot:persona" for the Default Visitor:

- `VisitorAPIImpl.java` — when com.dotmarketing.persona.id is sent (including the literal modes.persona.no.persona sentinel the frontend uses for "Default Visitor"), it tries personaAPI.find(...), which does `findContentletByIdentifier("modes.persona.no.persona", ...)`. That's not a real contentlet ID, so it throws — caught explicitly (catch(`DotContentletStateException e`), comment: "meant to catch the 'Can't find contentlet' error") — and falls through to visitor.setPersona(null).
- `ViewAsPageStatus.persona is set from visitor.getPersona()` → null in this case.
- `ViewAsPageStatusSerializer.java` — when persona is null, the "persona" key is never added to the response JSON at all — not null, structurally absent. Same for GraphQL (PageAPIGraphQLTypesProvider pulls persona off the same ViewAsPageStatus via PropertyDataFetcher).
- `PersonaAPIImpl.getDefaultPersona()` — the method that produces keyTag = "dot:persona" — is only called from getPersonas() (the dropdown-list builder), never from the visitor/render path that populates viewAs.

So viewAs.persona is genuinely absent for both "no persona param sent" and "explicitly selected Default Visitor," which means $pageData().personaTag is undefined in practice, not "dot:persona".

That said, the underlying worry is legitimate as a robustness concern, personaTag is client-supplied on a public **REST** endpoint, and if this assumption were ever wrong (a future change, a different client, a bug in persona resolution) the effect would be a silently-wrong personalization key rather than a loud error. So we added the guard anyway rather than relying on "we checked, it can't happen today": `PageResourceHelper.resolvePersonalizationFromTag()` now treats a bare "dot:persona" (no tag segment) as default, same as a missing tag. It's a one-line change and it also covers the pre-existing addContent()/drag-and-drop save path, which goes through the same logic and reads the same $pageData().personaTag.

Also added test_updateStyles_with_defaultVisitor to lock down the contract — asserts both null and the bare "dot:persona" prefix resolve to dot:default, so this can't silently regress.

zJaaal (approved):

ihoffmann-dot (approved):
Well-scoped, well-tested (2 new integration tests + FE unit tests), OpenAPI regenerated. 
Low risk. Behavior change is additive (new optional field), correctly defaulted, labeled AI: Safe To Rollback. No regression risk. No blocking issues.

PR #36633, opened 2026-07-20
refactor(deployment): update SDK publishing workflow to align with dotCMS release strategy

### LINKED ISSUE #36587: Align SDK publish pipeline with dotCMS date-lockstep versioning
### Description

[ADR-0019](https://github.com/dotCMS/platform-adrs/blob/main/decisions/0019-sdk-cms-date-lockstep-versioning.md) decides that the
dotCMS SDKs (JS/TS packages under `@dotcms/*`) must stop versioning independently and instead adopt
**date-lockstep versioning**: the SDK version is always identical to the dotCMS release version it
ships with (e.g. dotCMS release `2026.6.24` → `@dotcms/*@2026.6.24`).

Today the two are fully decoupled:

- **dotCMS core release** (`cicd_6-release.yml`) is a manual `workflow_dispatch` where a human types
  the release version by hand, creates the release branch/tag, and creates a GitHub Release.
- **SDK publish** (`cicd_manual-release-sdks.yml` → `deploy-javascript-sdk` action) is a *separate*
  manual `workflow_dispatch` with its own independent SemVer (patch/minor/major), plus an always-on
  `@next` prerelease tag published on every run (`BASE_VERSION-next.RUN_NUMBER`).
- The main release pipeline (`cicd_comp_deployment-phase.yml`) already has the wiring to call the SDK
  publish action, but it's explicitly disabled: `publish-npm-sdk-libs: false`.

This task implements the ADR: fold SDK publishing into the dotCMS release pipeline and retire the
standalone SDK release workflow.

## Precedent: this is exactly how dotCLI already works

dotCLI solved this identical problem already, and we should copy its mechanism rather than invent a
new one:

- `cicd_release-cli.yml` is triggered by `on: release: types: [created]` — i.e. it fires automatically
  when `cicd_6-release.yml` creates the GitHub Release object. No separate manual "publish CLI" step
  exists.
- It reads the version directly off the release tag:
  ```bash
  TAG_NAME=${{ github.event.release.tag_name }}
  RELEASE_VERSION=${TAG_NAME:1}   # strips leading "v", e.g. v2026.6.24 -> 2026.6.24
  ```
- That `RELEASE_VERSION` is written into `.mvn/maven.config` and into the published npm
  `package.json` for `@dotcms/dotcli`.
- Cutting a dotCMS release **stays manual** (a human still decides when to cut one). What becomes
  automatic is *only* the publish step that rides along on the release event.

## Proposed mechanism for the SDKs

1. **Do not automate cutting a dotCMS release.** `cicd_6-release.yml` remains a manual,
   human-triggered `workflow_dispatch`. This is intentional — release cutting has real blast radius
   (Docker `latest` tag promotion, Evergreen track pushes, changelog generation) and must stay
   human-gated.
2. **Automate the SDK publish step**, mirroring dotCLI:
   - Trigger the SDK publish off `release: created` (or hook it directly into
     `cicd_comp_release-phase.yml` alongside the existing dotCLI job), not a standalone
     `workflow_dispatch`.
   - Read the version from `github.event.release.tag_name` (same source dotCLI already uses) instead
     of from `core-web/libs/sdk/VERSION`.
   - Publish every `@dotcms/*` package at that exact version to the `latest` npm dist-tag.
3. **Drop the `next` dist-tag entirely.** No prerelease/canary channel is being kept. Every SDK
   package version published corresponds 1:1 to a real dotCMS release — no exceptions.
4. **Retire `cicd_manual-release-sdks.yml`** and the independent patch/minor/major SemVer flow once
   the new mechanism is confirmed to publish every SDK package correctly. Per the ADR's explicit
   caution, don't delete outright — disable it / make it manual-only first, and confirm there are no
   leftover responsibilities (pre-release builds, dry-run validation, internal-only artifacts) that
   aren't covered by the new flow before removing it for good.
5. **Resulting developer flow:**
   - Developer changes SDK code, tests locally, opens a PR, merges to `main`.
   - **Nothing publishes to npm at merge time.** The change simply sits in `main`.
   - Someone (dev or release engineering) manually cuts a dotCMS release, same as today, whenever
     they choose (multiple releases per week are fine — there is no fixed release cadence; only
     Evergreen's ~2-week *upgrade* cadence is fixed, which is unrelated to how often we can cut a
     release).
   - That release automatically publishes every `@dotcms/*` package at the release's version,
     including SDK packages with **no code changes** (identity-preserving re-publish, per the ADR).

## Things to Consider / Open Questions

- **LTS backport policy is still undefined** (ADR explicitly flags this as needing further
  investigation). Decide: does an LTS release publish SDKs under a `lts` dist-tag, under `latest`, or
  not at all? dotCLI's current behavior (skip LTS tags) may not be the right model for SDKs if
  customers on LTS still need SDK updates — confirm with SaaS & Platform / Support before deciding.
- **Urgent hotfix escalation path**: per the ADR, an urgent SDK fix does not get its own out-of-band
  publish — it still requires cutting a release. Confirm with Customer Support that the existing
  dotCLI-style escalation ("Support force-upgrades the affected customer directly") is understood to
  apply the same way to SDK hotfixes.
- **Runtime compatibility check** (ADR section "Runtime check"): out of scope for this specific
  pipeline ticket, but should be tracked as a related/follow-up task — implement the fail-open,
  warn-first version check in `@dotcms/client` reading `/api/v1/appconfiguration`.
- **Communication**: customers currently on independent SemVer versions of the SDKs need advance
  notice of this scheme change (a `2026.6.24` version looks nothing like a `1.4.2` version). Coordinate
  release notes / doc updates (ADR Implementation Note 9) so this isn't a silent breaking change to
  the versioning contract itself.
- **`examples/*` and docs** currently reference `latest`/caret ranges; confirm scope — is updating them
  part of this ticket or a separate follow-up?
- **Rollback**: if a release with a bad SDK publish needs to be pulled, what's the process? (e.g. `npm
  deprecate`, or does a subsequent release simply supersede it since customers should pin exact
  versions anyway?)
- **Non-version-publishing subset**: confirm whether anything in `cicd_manual-release-sdks.yml` today
  (dry-run validation, internal artifacts, etc.) has no equivalent in the release pipeline and must be
  migrated or preserved separately, per the ADR's explicit "confirm before deleting" instruction.

## Out of Scope

- The runtime SDK↔CMS compatibility check (`__SDK_VERSION__` injection + warn-first console check) —
  tracked separately per ADR Implementation Note 5.
- Any change to when/how often a dotCMS release is manually cut — that trigger and process are
  unchanged by this task.
- LTS backport policy definition (needs its own decision, referenced above).

## References

- ADR-0019: Date-Lockstep Versioning for the dotCMS SDKs —
  `platform-adrs/decisions/0019-sdk-cms-date-lockstep-versioning.md`
- Current SDK publish action: `.github/actions/core-cicd/deployment/deploy-javascript-sdk/action.yml`
- Current standalone SDK release workflow: `.github/workflows/cicd_manual-release-sdks.yml`
- Current dotCMS release workflow: `.github/workflows/cicd_6-release.yml`
- dotCLI release workflow (mechanism to mirror): `.github/workflows/cicd_release-cli.yml`
- Deployment phase wiring (currently disabled for SDKs): `.github/workflows/cicd_comp_deployment-phase.yml`
- Issue: [SDK ↔ CMS version alignment #35474](https://github.com/dotCMS/core/issues/35474)

### Acceptance Criteria

- [ ] Cutting a dotCMS release via `cicd_6-release.yml` results in every `@dotcms/*` SDK package being
      published to npm under the `latest` dist-tag at the exact release version (date-based, e.g.
      `2026.6.24`), with no manual SDK-specific step required.
- [ ] The published version is sourced from the GitHub Release tag (`github.event.release.tag_name`),
      not from `core-web/libs/sdk/VERSION` or any independently maintained version file.
- [ ] SDK publishing happens automatically as part of / triggered by the release pipeline (mirroring
      `cicd_release-cli.yml`'s `release: created` trigger) — there is no separate button or workflow a
      human must run to get the SDK onto npm.
- [ ] A dotCMS release with zero SDK code changes still re-publishes all SDK packages at the new
      version (preserves the one-dotCMS-version ⇄ one-SDK-version invariant).
- [ ] The `@next` dist-tag and all `-next.<run>` prerelease publishing logic are removed. No canary /
      prerelease channel is published for the SDKs.
- [ ] `package.json` versions (and inter-package `peerDependencies` referencing other `@dotcms/*`
      packages) are updated to the release version at publish time, consistent with current logic in
      `deploy-javascript-sdk/action.yml`.
- [ ] `cicd_manual-release-sdks.yml` (and the independent patch/minor/major SemVer flow) is disabled or
      made manual-only, pending confirmation that no leftover responsibilities exist outside the new
      release-pipeline flow.
- [ ] `core-web/libs/sdk/VERSION` file usage is removed or repurposed (confirm nothing else in the
      build depends on it before deleting).
- [ ] `examples/*` package.json dependencies pin an exact date version rather than `latest`/`^` ranges
      (per ADR Implementation Note 6) — or at minimum, this is tracked as a explicit follow-up if out
      of scope for this ticket.
- [ ] LTS releases (tag containing `_lts_v`) are considered: confirm whether SDK publish should be
      excluded for LTS tags (dotCLI already excludes `LTS` tags in `cicd_release-cli.yml`) or should
      publish under a distinct dist-tag per the ADR's LTS backport policy note.
- [ ] CI validates that the release-triggered publish job actually ran and published successfully
      (i.e. a release isn't silently considered "done" if the SDK publish step failed).

### Priority

None

### Additional Context

_No response_

KevinDavilaDotCMS:
This PR introduces the changes to align the SDK release with the dotCMS release process

Already have the same code in core-workflow-test and the tests passed

https://github.com/dotCMS/core-workflow-test/pull/514
https://github.com/dotCMS/core-workflow-test/pull/516

This PR fixes: #36587

This PR fixes: #36587


KevinDavilaDotCMS:
Went through all 3 findings from the Claude code review against the actual code at this commit — unlike the earlier deepseek-bot round on the fork PR, these were all real and confirmed by direct inspection. Fixed all three.

## Fixed

- **Missing `checkout` in `notify-precheck-failure` (🟠 High)** — confirmed real, and on us: we added `actions/checkout` to the `publish` job when fork testing caught this exact issue there, but never propagated the same fix to `notify-precheck-failure` (added earlier, in the previous review round, to make a pre-check failure non-silent). It went untested because our fork tests never exercised a `precheck` failure path (only success, LTS-skip, and dry-run). Added the missing checkout step — same reasoning as the `publish` job's comment: a local (`./`) action can't be resolved on an empty runner workspace, so the Slack notification was a silent no-op.
- **`IS_LTS` never computed for `workflow_dispatch` (🟡 Medium)** — confirmed real. The regex validating manual-dispatch versions explicitly allows the `_lts_v##` shape, but `IS_LTS` was only ever derived inside the `release`-event branch, defaulting to `false` for manual runs. A manual dispatch with an LTS-shaped version would pass validation and then actually publish, contradicting the documented skip-LTS policy. Fixed by deriving `IS_LTS` from whichever value carries the marker for that event type (`tag_name` for `release`, `RELEASE_VERSION` for `workflow_dispatch`) instead of only the former.
- **Publish loop not retriable on partial failure (🟡 Medium)** — confirmed real, and a legitimate operational trap under the new model: the version is fixed to the dotCMS release tag and can never be bumped (unlike the old `@next` scheme, which side-stepped this with a fresh run-number suffix each retry). If package N of 10 failed mid-loop, packages 1..N-1 were already published immutably, and a re-run would hit npm's hard rejection on republishing an existing version — permanently stalling that release's SDK publish with no clean recovery. Made the loop idempotent: each package now checks `npm view @dotcms/$sdk@$RELEASE_VERSION` first and treats an already-published version as success, so a re-run (or a run resumed after `cancel-in-progress` killed a prior one) can finish whatever's left.

Same changes mirrored into the `core-workflow-test` fork branches, ready to be re-verified there before this goes back for another look.

mbiuki (approved):

.github/actions/core-cicd/deployment/deploy-javascript-sdk/action.yml
```diff
@@ -176,63 +139,55 @@ runs:
...
           echo "📦 Processing @dotcms/${sdk}..."
           cd "$sdk"
 
-          ORIGINAL_VERSION=$(jq -r '.version' package.json)
-
-          if [ "$PUBLISH_LATEST" = "true" ]; then
-            echo "  🚀 Publishing latest: @dotcms/${sdk}@${LATEST_VERSION}"
+          # Idempotency: the version is fixed to the dotCMS release tag and can never be
+          # bumped, so if a prior (failed/cancelled) run already published this exact
+          # version, npm would reject a re-publish with a hard 403 — permanently stalling
```
sfreudenthaler, line 169: TIL

.github/workflows/cicd_manual-release-sdks.yml
sfreudenthaler, line 1: Clear and meaningful error, but why keep it at all?  why not just remove all together?

.github/workflows/cicd_release-sdk.yml
```diff
@@ -0,0 +1,203 @@
...
+# for @dotcms/dotcli: publishing is triggered automatically by the GitHub Release created
+# during cicd_6-release.yml, and the version is read straight off the release tag — there
+# is no independent SDK version, no separate manual "publish SDK" step, and no `@next`
+# prerelease channel.
+#
+# Key features:
+# - Automatic trigger on release creation; workflow_dispatch kept for manual re-publish/testing
+# - Version is derived from the GitHub Release tag (mirrors cicd_release-cli.yml exactly)
+# - Skips LTS releases (LTS SDK publish policy is a separate, not-yet-decided follow-up)
+# - Publishes every @dotcms/* SDK package at the exact release version, to the `latest` tag only
```
sfreudenthaler, line 16: this is fine even for Evergreen Tracks.  Because the same immutable builds eventually make it to all three tracks.  We don't need to release for each track, just latest will do and they'll trickle down automatically

sfreudenthaler (approved):
A few considerations and open questions, but no blocking commentary.

.github/workflows/cicd_release-sdk.yml
```diff
@@ -0,0 +1,203 @@
...
+        required: false
+        default: 'main'
+      dry-run:
+        description: 'Dry run (skip the actual npm publish)'
+        default: 'false'
+        required: false
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref || github.run_id }}
+  cancel-in-progress: true
```
dcolina: `cancel-in-progress: true` on a publish workflow can leave packages half-published if a re-run cancels the in-flight one. `cicd_6-release.yml` already uses `cancel-in-progress: false` for exactly this reason, I think it worth aligning here too.

KevinDavilaDotCMS: Good catch! 
Added!

.github/workflows/cicd_release-sdk.yml
KevinDavilaDotCMS: Thanks for confirming!
 that matches the design intent to publishing only latest per release (no per-track republish) is correct because all three tracks eventually receive the same immutable build as it trickles down

.github/workflows/cicd_manual-release-sdks.yml
KevinDavilaDotCMS: It's part of the ADR: https://github.com/dotCMS/platform-adrs/blob/main/decisions/0019-sdk-cms-date-lockstep-versioning.md#:~:text=Confirm%20before%20deleting,off%2Dversion%20publish.

We'll test it in a live release first, then we can delete it

PR #36635, opened 2026-07-20
fix(publish-date-workflow-resolution): fix publish date

### LINKED ISSUE #36634: Prevent scheduled publishDate from bypassing unresolved workflows
## Problem

When a content item has a future `publishDate` (also referred to as `timeOpen`), the `PublishDateUpdater` job can publish it locally as soon as that date is reached, even when the item is still in an unresolved workflow step.

For example, an editor can schedule content for next Monday while its approval workflow is still pending. Under the current behavior, the scheduled publishing job can make the content live without an actual workflow publish action, unintentionally bypassing the approval process.

## Goal

Provide an opt-in safeguard for scheduled local publishing so that a matured `publishDate` cannot publish content before the applicable workflow step has been resolved, while preserving the historical behavior by default.

## Proposed Approach

- Introduce the feature flag `PUBLISH_JOB_QUEUE_RESPECT_WORKFLOW_RESOLUTION`, disabled by default.
- When enabled, require the current workflow step to have `resolved=true` before `PublishDateUpdater` performs local scheduled publication.
- Introduce `PUBLISH_JOB_QUEUE_RESPECT_WORKFLOW_RESOLUTION_SCHEMES` to optionally scope the safeguard to comma-separated workflow scheme IDs or variable names.
  - Empty value: apply the safeguard to all workflows.
  - Configured value: apply it only to the listed workflows.
- In global mode, hold content that has no current workflow step.
- In scoped mode, hold content without a current step only when its content type is assigned to a configured workflow scheme.
- Fail closed for an affected content item if its workflow state cannot be read, while continuing to process the remaining queue.
- Ensure failed or missing content items still count toward batch transaction boundaries so transactions are always closed.
- Do not change immediate workflow publishing, push publishing, or scheduled expiration/unpublishing.

## Acceptance Criteria

- When `PUBLISH_JOB_QUEUE_RESPECT_WORKFLOW_RESOLUTION=false`, scheduled publishing behavior is unchanged.
- When the flag is enabled and the workflow scheme list is empty, the safeguard applies globally.
- When the flag is enabled and workflow schemes are configured, the safeguard applies only to those schemes.
- A configured workflow at a step with `resolved=true` may be published by `PublishDateUpdater`.
- A configured workflow at a step with `resolved=false` is not published by `PublishDateUpdater`.
- In global mode, content without a current workflow step is not published.
- In scoped mode, content without a current step is not published when its content type is assigned to a configured workflow.
- Workflows outside the configured list keep the historical scheduled publishing behavior.
- Workflow lookup failures hold the affected content item and do not leave the batch transaction open.
- Immediate publishing is unaffected.
- Scheduled expiration/unpublishing is unaffected.
- Push publishing is unaffected.
- Automated tests cover enabled and disabled flags, resolved and unresolved steps, scoped and global behavior, missing steps, workflow lookup failures, transaction closure, and unpublish behavior.

## Related PR

- #36594

syedATdot:
Fixes #36634

Imported from external fork contribution by @riccardoruocco (`riccardoruocco/core`, branch `fix/publish-date-workflow-resolution`) so CI can run on it..

This PR fixes: #36634


dotCMS/src/enterprise/java/com/dotcms/enterprise/publishing/PublishDateUpdater.java
```diff
@@ -590,6 +647,78 @@ public String getOperationName() {
...
+     * @param contentlet content selected by the scheduled publishing job
+     * @param workflowAPI workflow API used to inspect the current step
+     * @return {@code true} when scheduled publication is allowed
+     * @throws DotDataException when the current workflow state cannot be read
+     * @throws DotSecurityException when the current workflow scheme cannot be read
+     */
+    @VisibleForTesting
+    static boolean workflowAllowsScheduledPublish(final Contentlet contentlet,
+            final WorkflowAPI workflowAPI) throws DotDataException, DotSecurityException {
+        if (!Config.getBooleanProperty(RESPECT_WORKFLOW_RESOLUTION_PROPERTY, false)) {
```
fabrizzio-dotCMS: No need to re-read this value over and over again; do it once at the beginning, at the constructor level

fabrizzio-dotCMS:
**Suggestion: read the config once per run, not per contentlet**

`workflowAllowsScheduledPublish` currently calls `Config.getBooleanProperty(...)` and re-parses `configuredWorkflowSchemes()` (`split` + `stream` + `collect`) for **every** contentlet. Since `Publisher` is created once per job run, snapshot these in its constructor:

```java
// Publisher constructor
this.respectWorkflowResolution = Config.getBooleanProperty(RESPECT_WORKFLOW_RESOLUTION_PROPERTY, false);
this.configuredSchemes = configuredWorkflowSchemes(); // parsed once
```

`workflowAllowsScheduledPublish` then takes the resolved values instead of reading `Config` itself.

**Alternative:** if you prefer all run-scoped config in one place, resolve them at the top of `updatePublishExpireDates(...)` (the job entry point) and pass them down.

**Please don't use a `static final Lazy<>`** here: it caches for the whole JVM lifetime, so changing `PUBLISH_JOB_QUEUE_RESPECT_WORKFLOW_RESOLUTION` would need a **Tomcat restart**. Reading once per run keeps the values stable during the run and still picks up changes on the next fire.

fabrizzio-dotCMS:
**Test gap: mid-batch commit triggered by a failure**

`batchClosesTransactionWhenProcessorFails` uses `transactionBatchSize=10` with a single inode, so it only exercises the **final** commit path. The core of the transaction refactor is that a failed/missing record now counts toward the commit boundary mid-batch. Suggest adding a case with several inodes and `transactionBatchSize=2` where an early item throws, to lock in that behavior.

fabrizzio-dotCMS (changes requested):
Overall the approach is solid and the opt-in flag keeps historical behavior safe by default. Requesting changes so we address a scaling issue plus a couple of smaller items before merge.

### Must address

1. **Held content is re-scanned every run (scaling).** A gated contentlet stays `live:false`, so it keeps matching `PUBLISH_LUCENE_QUERY` and is re-fetched (`find`), re-checked (`findCurrentStep`), and re-logged on every run — the job runs every minute, so the cost grows linearly with the backlog of unresolved content and never clears on its own. Push the workflow predicate into the publish query so held items are excluded at the index level (`+wfstep:(<resolvedStepIds>)`, resolved at query-build time since `resolved` is a step-definition property, not an indexed per-content field). Keep `workflowAllowsScheduledPublish(...)` as the authoritative final gate for index-staleness and the scoped/no-step cases. Details in a separate comment below.

2. **Log level for held items.** `workflowAllowsScheduledPublish(...)` logs at `WARN` per held item per run. A held item is expected steady state, not an anomaly — this becomes per-item-per-minute log spam. Lower to `DEBUG` (matches the existing "already auto-published and manually unpublished" skip).

### Minor

3. **Read the config once per run, not per contentlet.** `Config.getBooleanProperty(...)` / `configuredWorkflowSchemes()` are evaluated inside the per-contentlet `process(...)` path. Resolve them once when the `Publisher` is built and reuse. (See inline comment.)

4. **Test gap: mid-batch commit triggered by a failure.** The transaction-boundary change now commits on *visited* count including failed/missing contentlets — please add a test that exercises a commit boundary reached via a failure, not only via successful processing. (See inline comment.)

fabrizzio-dotCMS:
**On the "records pile up and come back every run" concern — agreed, and I think the right fix is to push the workflow predicate into the search query.**

The candidate set is fetched here via an Elasticsearch query:

```java
// PublishDateUpdater.java:449
final List<String> allInodes = contentletAPI.searchIndex(luceneQuery, 0, 0, null,
        systemUser, false).stream().map(ContentletSearch::getInode).collect(Collectors.toList());
```

Since a held contentlet stays `live:false`, it keeps matching `PUBLISH_LUCENE_QUERY` and is re-fetched, re-checked (`find` + `findCurrentStep`), and re-logged (WARN) on **every** run — the job runs every minute, so this cost scales linearly with the backlog of unresolved content and never clears on its own.

**To be precise:** filtering in the query does *not* stop the content from accumulating (that's inherent to the feature — we intentionally hold it until the workflow resolves). What it fixes is that the accumulation stops being expensive to skip repeatedly. Per-run cost goes from `O(backlog)` to `O(actually-publishable)`:
- held items are excluded at the index level → no `find`, no `findCurrentStep`, no WARN for them each minute;
- `searchIndex(..., 0, 0, ...)` no longer materializes the full matching set in memory every run.

**Feasibility:** the workflow fields are indexed — `wfstep` / `wfscheme` (`ESMappingConstants`), and there's precedent for querying them (`WorkflowAPIImpl` builds `+wfscheme:( … ) +working:true` and `+(wfstep:… )`). The one caveat is that `resolved` is **not** a per-content indexed field — it's a property of the step *definition* (`WorkflowStep.resolved`). So we can't add `+resolved:true`; instead we'd resolve the set of resolved step IDs at query-build time and add an inclusion clause, e.g. `+wfstep:(<resolvedStepId1> <resolvedStepId2> …)`.

This should go into the **publish** query only — built around `getPublishLuceneQuery(...)` / the `PUBLISH_LUCENE_QUERY` template (`PublishDateUpdater.java:80`, `:323`) — and must **not** touch the unpublish/expire path (`:331`), per the issue's constraint that expiration is unaffected.

**Suggestion — keep both, they solve different problems:**
1. Add the `+wfstep:(resolvedStepIds)` pre-filter to the publish query → fixes the scale/every-minute concern.
2. Keep `workflowAllowsScheduledPublish(...)` as the authoritative final gate. The index reflects the last reindex of the content, so there's a small staleness window after a workflow transition; the Java check reads live DB state and also covers the scoped/no-current-step semantics that a pure inclusion filter can't express cleanly (in scoped mode, step-less content outside the configured schemes should still publish).

Net: the query filter is the scaling optimization; the Java gate stays as the correctness safety net.

fabrizzio-dotCMS (approved):

fabrizzio-dotCMS (changes requested):
## Follow-up: `resolved` step is not the same as `approved` — clarify the contract before this feature is switched on

Approving my earlier pass was premature — I think there is a semantic gap worth resolving before this ships, even though it is behind an opt-in flag.

### The issue

The whole scheduled-publish gate hinges on `WorkflowStep.isResolved()`:

- `getPublishLuceneQuery(...)` appends `+(wfstep:<resolved step ids>)` in global mode.
- `workflowAllowsScheduledPublish(...)` returns `currentStep.map(WorkflowStep::isResolved).orElse(false)`.

But `resolved` on a workflow step only means **"this step is a terminal/resolution state"** — it does **not** mean the content was *approved*. Any terminal step marked `resolved=true` qualifies, including **Archived** or **Rejected** steps. So a contentlet sitting in an archived/rejected-but-resolved step, with a publish date in range, would match the query and be published — which is almost certainly not what "ready to be published" is meant to express.

### Evidence is in this PR's own test

`PublishDateUpdaterTest#publishQueryRestrictsGlobalModeToResolvedSteps` asserts:

```java
assertTrue(query.contains(" +(wfstep:published-step wfstep:archived-step )"));
```

i.e. `archived-step` is explicitly treated as eligible for scheduled publish.

### Why it still matters even though it is opt-in

`Config.getBooleanProperty("PUBLISH_JOB_QUEUE_RESPECT_WORKFLOW_RESOLUTION", false)` defaults to `false`, so there is no out-of-the-box regression — good. But the day an operator enables it expecting "only publish approved content", archived/rejected-but-resolved content will go live. The ambiguity is latent behind the flag rather than removed.

### Requested changes

1. Decide and document the intended criterion: is it truly "any resolved step", or specifically the **approval** terminal step(s)? If the latter, `isResolved()` is insufficient — we need a way to distinguish the approval resolution from archive/reject resolutions (e.g. per-step configuration, not just per-scheme).
2. Add an integration test that indexes real content and runs the query against the index (not just asserting the query string), including a contentlet in a **resolved Archived/Rejected step** with a past publish date, and assert the intended behavior. That test turns the ambiguity above into an explicit, agreed-upon decision. It should live in `dotcms-integration` (class `*IT`) and be registered in the appropriate `@SuiteClasses` suite.

Happy to pair on either. Everything else in the PR looks solid (immutable `WorkflowResolutionConfig` record, per-run config caching, manual chunked transactions, fail-closed fallback, Lucene-id sanitization).

fabrizzio-dotCMS (approved):
## Updating my earlier review: approving with non-blocking comments

Reconsidering my previous *request changes*: since this feature is **opt-in and defaults to `false`** (`PUBLISH_JOB_QUEUE_RESPECT_WORKFLOW_RESOLUTION`), there is no out-of-the-box behavior change, and the implementation itself is solid — immutable `WorkflowResolutionConfig` record, config read once per run, manual chunked transactions (correct for a large batch), fail-closed fallback, and Lucene-id sanitization. The `transactionVisited`→`transactionProcessed` change also fixes a latent uncommitted-transaction bug. So I'm switching to **approve**, and downgrading the items below to non-blocking follow-ups rather than merge blockers.

### Non-blocking: `resolved` step ≠ `approved` (please document + follow up)

The gate uses `WorkflowStep.isResolved()`, but `resolved` only means "terminal/resolution step", not "approved". Terminal **Archived**/**Rejected** steps marked `resolved=true` also qualify — this PR's own test asserts `+(wfstep:published-step wfstep:archived-step )`, i.e. archived content is treated as publishable. When an operator enables the flag expecting "only approved content", that is a surprise.

Because it is opt-in, this doesn't need to block the merge, but I'd ask for two cheap things:
1. Make the criterion explicit in a code comment and in the property description in `dotmarketing-config.properties` (i.e. "resolved step, which includes terminal archive/reject steps").
2. Open a follow-up issue to evaluate a more precise criterion (see below).

### Suggested follow-up: prefer `WorkflowAction.hasPublishActionlet()` over `isResolved()`

Since workflows are fully configurable, there is no fixed "approved" step — but the workflow already self-describes the publish point through the **publish actionlet**. The rest of the core determines publish capability this way, not via `resolved`:
- `ESContentletAPIImpl` (~L5600), `ImportUtil` (~L4506), `DotWebdavHelper` (~L944), `WorkflowResource` (~L848), `SystemActionApiFireCommandFactory` (~L247) all use `WorkflowAction.hasPublishActionlet()`.

A criterion like *"the contentlet's current step exposes an action with `hasPublishActionlet()`"* (via `WorkflowAPI.findActions(step, systemUser)` / `findAvailableActions(contentlet, user)`) would naturally exclude archived/rejected steps and let admins define the publish point simply by where they place the publish action — no new convention needed. The existing `..._SCHEMES` property still scopes which workflows participate. Same Lucene `+(wfstep:...)` technique, just computing the step-id list from "steps with a publish action" instead of "resolved steps".

### Also worth adding (follow-up ok): an integration test

The current tests assert the query *string* and the mocked gate. An IT under `dotcms-integration` (class `*IT`, registered in a `@SuiteClasses` suite) that indexes real content across resolved/unresolved/archived steps and asserts what the query actually returns from the index would lock down the contract — including the archived/rejected-resolved case above.

None of the above blocks the merge given the opt-in default. Nice work overall.

PR #36639, opened 2026-07-20
fix(content-drive): load folder tree via /folder/search to remove 40-folder cap (#36638)

### LINKED ISSUE #36638: [BUG] Content Drive: Folder tree caps each level at 40 folders (folders beyond 40 are inaccessible)
## Problem Statement

The Content Drive folder tree (left sidebar) only ever renders the **first 40 subfolders** of any folder level. Folders beyond the 40th are silently dropped — there is no "load more", no pagination control, and no indication that folders are missing — so they are effectively **inaccessible** from the tree.

**Root cause:** the frontend `DotFolderService.getFolders()` posts to the endpoint **without a `limit` parameter**, so the backend applies its default of `40`.

- FE — `core-web/libs/data-access/src/lib/dot-folder/dot-folder.service.ts` → `POST /api/v1/folder/byPath` with body `{ path }` only, no `limit`.
- BE — `dotCMS/.../rest/api/v1/folder/FolderResource.java#findSubFoldersByPath` → `@DefaultValue("40") @QueryParam("limit")`.

Because the tree lazy-loads one level at a time (on node expand), **each level** is independently truncated to 40.

**Impact:** any site or folder that has more than 40 direct subfolders in a single level. Front-end, browser-agnostic.

## Steps to Reproduce

1. Use a site (or folder) that has **more than 40 direct subfolders** in one level.
2. Open **Content Drive**.
3. Expand that folder in the left sidebar tree.
4. Observe that only 40 subfolders render.

**Expected:** every subfolder in that level is reachable in the tree.
**Actual:** the level is capped at 40; the remaining subfolders never appear and cannot be navigated to.

## Technical Context

- `POST /api/v1/folder/byPath` is **`@Deprecated(forRemoval = true)`** (since Jun 2026, PR #36308 / issue #36151). It was replaced by **`GET /api/v1/folder/search`** because `byPath` runs a **per-folder permission check in a loop** (O(N)), which caused ~2s responses on large sites.
- `GET /api/v1/folder/search` already exists and returns paginated results with a real total count and **batch permission filtering** (one SQL round-trip, O(1)). The FE service already has a `searchFolders()` method for it — it just isn't wired into the Content Drive tree.

## Recommended Fix

Migrate the Content Drive folder tree's data source from the deprecated `POST /byPath` to **`GET /api/v1/folder/search`** (non-recursive, per level), adapting the response in the FE data/util layer so the **tree component and its behavior stay identical**. This:

- removes the 40-folder cap (search supports pagination and returns a total),
- avoids the ~2s large-site regression (batch permission filter vs. per-folder loop),
- retires the deprecation debt before `byPath` is removed.

Core tree behavior (lazy per-level expand, selection, highlighting, URL restore) is preserved — only the data plumbing changes, plus the "Load more" affordance below.

### "Load more" pagination

To keep the DOM bounded on very large levels (a level with thousands of siblings would otherwise render thousands of nodes at once and freeze the browser), expanding a folder loads its children **one page at a time** (`FOLDER_TREE_PAGE_SIZE = 50`):

- When a level has more children than a page, a non-selectable **"Load more (N)"** node is appended at the bottom. Clicking it fetches and appends the next page, then refreshes the remaining count or removes the node when the level is exhausted.
- **Cold load / deep-link restore is not paginated:** each ancestor level of the URL's target path is fetched in full so the selected folder always resolves and the tree expands to it. (Consequence: a level reached via URL restore renders in full with no "Load more", whereas the same level reached by clicking paginates. Making cold-load paginate too — "page until target" — is a possible follow-up.)

**Rejected alternatives:**
- `byPath` + `limit=-1` — one-line, but keeps the deprecated endpoint and re-introduces the exact per-folder permission cost `byPath` was deprecated for.
- Virtual scroll — bounds the DOM but per-node virtual scroll in a lazy nested tree is high-complexity for a rare case; "Load more" achieves the same DOM bound far more simply.

**Files (FE):**
- `.../portlet/src/lib/utils/functions.ts` (`getFolderHierarchyByPath`, `getFolderNodesByPath`, `folderSearchViewToDotFolder`, `buildLoadMoreNode`)
- `.../portlet/src/lib/utils/tree-folder.utils.ts` (`buildTreeFolderNodes`, `createTreeNode`)
- `.../portlet/src/lib/store/features/sidebar/withSidebar.ts`
- `.../portlet/src/lib/components/dot-content-drive-sidebar/*` (expand + "Load more" handling)
- `.../ui/src/lib/dot-tree-folder/*` (`loadMore` output + "Load more" node rendering)
- `.../ui/src/lib/shared/models.ts` + `constants.ts` (`load-more` node type, `LOAD_MORE_*`)
- `libs/dotcms-models/.../dot-folder.model.ts` (`hasChildren`)
- `Language.properties` (`content-drive.tree.load-more`)

No backend change required — `GET /api/v1/folder/search` already exists.

## Acceptance Criteria

- [ ] A folder level with more than 40 subfolders is fully reachable — the old 40-cap is gone.
- [ ] Expanding a folder loads its children a page at a time (50); when more remain, a **"Load more (N)"** node appears at the bottom. Clicking it appends the next page and refreshes/removes the node when the level is exhausted.
- [ ] The "Load more" node is not selectable and does not navigate; clicking it never changes the selected folder.
- [ ] Cold load / deep-link to a folder that lives beyond the first page still resolves — the tree expands to and selects it.
- [ ] Core tree behavior is preserved: lazy per-level expand, node selection/highlighting, initial hierarchy load to the current path, and URL restore.
- [ ] Folder tree data is sourced from `GET /api/v1/folder/search` (non-recursive per level); `POST /api/v1/folder/byPath` is no longer used by Content Drive.
- [ ] Folders the user lacks read permission for are not shown, and listing a large level performs well (batch permission filtering; no per-folder N+1).
- [ ] Childless folders render without an expand chevron (using the endpoint's `hasChildren`).
- [ ] An empty folder (0 children) renders correctly as a leaf; root / site-level listing works.
- [ ] The existing error path (folder load failure) is preserved and handled gracefully.
- [ ] Other `getFolders` callers (site/folder field, browsing service) are unaffected — either left untouched or verified.
- [ ] Sidebar/tree/util unit tests cover the new data source, the >40 case, `hasChildren`→chevron, and the "Load more" expand/append/exhaust flow.

## dotCMS Version

Latest from `main` branch.

## Severity

## Links

NA

zJaaal:
## Proposed Changes

Fixes the Content Drive sidebar folder tree silently capping every level at **40 folders**.

**Root cause:** the tree fetched each level through the deprecated `POST /api/v1/folder/byPath` with no `limit`, so the backend's default of `40` truncated the level — any folder beyond the 40th was unreachable.

**Fix:** migrate the tree's data source to the current `GET /api/v1/folder/search` (non-recursive, one call per level), requesting a single large page so every folder in the level is returned.

### Why `search`
`byPath` is `@Deprecated(forRemoval = true)` (PR #36308) — it was replaced by `search` because its per-folder permission loop caused ~2s responses on large sites. `search` batches permission filtering (one SQL round-trip) and returns a real total, so uncapping is safe.

### Behavior preserved
The visible tree UX is unchanged. Lazy per-level expand, node selection, routing/URL path sync, and folder-create / upload / editor forwarding all read the same node data (`id`, `inode`, `hostname`, full `path`) — the adapter reconstructs each field identically to what `byPath` returned.

### Intentional improvements
- **No expand chevron for empty folders** — `search` returns `hasChildren`, so childless folders no longer show a toggle that loads nothing on click.
- **Alphabetical order** — levels are now ordered by name (ASC); `byPath` returned unspecified DB order.
- **No silent truncation** — a level exceeding the page size logs a warning (see below).

### Pagination decision
The tree renders a whole level at once (no "load more"), so we request a single large page (`FOLDER_TREE_SEARCH_PAGE_SIZE = 10000`, matching the backend safety cap). Folder levels essentially never exceed this, and the backend loads-all-then-slices regardless of page size, so infinite scroll would add per-node pagination complexity with no backend-cost benefit. **Deferred as a follow-up** if a huge-level site ever appears; a warning is logged if a level is truncated so it is not silent.

### Files
- `functions.ts` — `FolderSearchView → DotFolder` adapter; `getFolderHierarchyByPath` / `getFolderNodesByPath` call `searchFolders` with `siteId` + folder path; truncation warning
- `tree-folder.utils.ts` — drop the `byPath` parent-first placeholder slice; derive `leaf` from `hasChildren`
- `withSidebar.ts` — pass `siteId` / `hostname` / folder path
- `dot-folder.model.ts` — optional `hasChildren`
- `constants.ts` — `FOLDER_TREE_SEARCH_PAGE_SIZE`

No backend change — `GET /api/v1/folder/search` already exists. The old `getFolders`/`byPath` method is untouched and still used by the site/folder field and browsing service.

### Coordination note
This removes Content Drive's last dependency on `POST /api/v1/folder/byPath`, unblocking its eventual removal.

## Checklist
- [x] Tests (unit: `functions.spec.ts`, `tree-folder.utils.spec.ts`, `withSidebar.spec.ts` — 774 pass)
- [ ] Translations — N/A
- [x] Security Implications Contemplated — `search` permission-filters all results server-side; tree remains scoped to the current site

## Testing
- `pnpm nx test portlets-content-drive` — 27 suites, 774 tests pass
- Added a >40-folder-level regression case, `hasChildren`→chevron cases, and a truncation-warning case
- Verified saving, editor/upload forwarding, and routing consume identical node data

This PR fixes: #36638

This PR fixes: #36638


zJaaal:
Addressed the 🟡 parent-path trailing-slash finding in `957658c863`.

`folderSearchViewToDotFolder` now normalizes the parent path before composing the folder's own path (mirroring the guard in `dot-browsing.service.ts`):

```ts
const parentPath = view.path.endsWith('/') ? view.path : `${view.path}/`;
// path: `${parentPath}${view.name}/`
```

Note: a live stress test on a running instance (10k+ folders in one level) confirmed the backend currently *does* return trailing-slashed parent paths (e.g. `/zz-stress-test-36638/`), so this was not a live defect — but the guard keeps the two `/folder/search` adapters consistent and protects `isOnTargetPath` if that ever changes. Added a unit test covering `/main` + `sub` → `/main/sub/`.

— Posted by Claude Code on behalf of @zJaaal

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/features/sidebar/withSidebar.ts
```diff
@@ -58,9 +58,13 @@ export function withSidebar() {
...
                 const urlFolderPath = store.path() || '';
-                const fullPath = `${currentSite.hostname}${urlFolderPath}`;
 
-                getFolderHierarchyByPath(fullPath, dotFolderService)
+                getFolderHierarchyByPath(
+                    urlFolderPath,
+                    currentSite.identifier,
+                    currentSite.hostname,
+                    dotFolderService
+                )
```
KevinDavilaDotCMS: This use too many params, maybe use an object?
Probably that object type already exists

KevinDavilaDotCMS (approved):

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/features/sidebar/withSidebar.ts
zJaaal: Good call — collapsed the `siteId` + `hostName` params into the existing `DotSite` object (which is what `currentSite` already is), so both `getFolderHierarchyByPath` and `getFolderNodesByPath` now take `(folderPath, site, dotFolderService[, page])`. Done in `c2d9da242a`.

— Posted by Claude Code on behalf of @zJaaal

PR #36642, opened 2026-07-20
feat(analytics): expose visitor postal/zip code in Geolocation ($visitor.geo.postal) (#36232)

### LINKED ISSUE #36232: Expose visitor postal/zip code in Visitor Geolocation ($visitor.geo.postal)
## Summary
Expose the visitor's **postal / zip code** in the Visitor Geolocation feature so it is available via `$visitor.geo.postal`, the visitor JSON, and GraphQL.

## Context
Reported via Freshdesk ticket: https://dotcms.freshdesk.com/a/tickets/37915

A customer needs the visitor's zip code for geolocation-based site behavior and asked whether purchasing a MaxMind database update would expose it. It would not — this is not a data gap. The bundled MaxMind GeoLite2-City database already contains postal code (`CityResponse.getPostal().getCode()`), but dotCMS never reads it. This has come up before (Jan 2026, City-vs-Country DB discussion).

## Current behavior
The `Geolocation` domain object exposes city, country, countryCode, continent, continentCode, subdivision, subdivisionCode, latitude, longitude, timezone, company, and ipAddress — but **no postal/zip field**. As a result `$visitor.geo`, the serialized visitor JSON, and the GraphQL geo type all omit postal code.

Root cause (code references):
- `dotCMS/src/main/java/com/dotcms/visitor/domain/Geolocation.java` — no `postal` field on the model or builder.
- `dotCMS/src/main/java/com/dotcms/util/GeoIp2CityDbUtil.java` (`getGeolocation`, ~lines 405-416) — builds the object from the MaxMind `CityResponse` but never calls `cityResponse.getPostal()`.
- `dotCMS/src/main/java/com/dotcms/graphql/business/PageAPIGraphQLTypesProvider.java` (~lines 313-340) — explicit geo field registry; postal is not registered.

## Expected behavior
`$visitor.geo.postal` (and the JSON / GraphQL equivalents) return the visitor's postal/zip code when MaxMind provides it.

## Proposed change (read surfaces only — low risk)
1. `Geolocation.java` — add a `postal` field, `getPostal()` getter, and `withPostal()` builder method; include in the constructor, builder copy-constructor, `equals()`, and `hashCode()` (mirror the existing `subdivision` handling). Velocity and JSON serialization pick the getter up via reflection.
2. `GeoIp2CityDbUtil.java` — add `.withPostal(cityResponse.getPostal().getCode())` (null-guarded) to `getGeolocation(InetAddress)`, plus a `getPostalCode(String ip)` accessor mirroring `getCityName()`.
3. `PageAPIGraphQLTypesProvider.java` — register `geoFields.put("postal", new TypeFetcher(GraphQLString, new PropertyDataFetcher<Geolocation>("postal")))`.
4. `GeoCharacter.java` — add `.put("g.postalCode", geo.getPostalCode(ipAddress))` to the `$visitor.g` map.
5. Tests — `GeoIp2CityDbUtilTest`: the serialize round-trip passes once the field is in the builder + equals/hashCode; add a postal assertion for the test IP `108.49.249.66` (note: the bundled GeoLite2 db is ~4 years old; if it returns no postal for that IP, assert non-null instead of an exact value).

## Out of scope
Postal-based personalization **rules** (rules-engine conditionlet + Angular UI) — the conditionlet is distance-only today and would require new comparison types. Separate, larger effort.

## Workaround (available today, no core change)
Customers can write a custom OSGi viewtool plugin that bundles the MaxMind GeoIP2 jar, opens the `.mmdb`, and exposes `$myGeo.postal` via `reader.city(ip).getPostal().getCode()`. (`com.maxmind` is not OSGi-exported, so the plugin must bundle its own copy.)

## Affected versions
Current Release / dotEvergreen.

dsilvam:
### Proposed Changes
Exposes the visitor's postal/zip code across the Visitor Geolocation surfaces. The bundled MaxMind GeoLite2-City database already contains postal code (`CityResponse.getPostal().getCode()`) — dotCMS just never read it. This is read-surface only (additive, backward-compatible).

* `Geolocation.java` — add `postal` field, `getPostal()`, `withPostal(...)` builder, plus constructor/copy-constructor/`equals`/`hashCode` (mirrors `subdivision`). Picked up by Velocity (`$visitor.geo.postal`) and JSON via reflection.
* `GeoIp2CityDbUtil.java` — populate postal in `getGeolocation(InetAddress)` (null-guarded) and add a `getPostalCode(String ip)` accessor mirroring `getCityName(ip)`.
* `GeoCharacter.java` — add `g.postalCode` to the `$visitor.g` map, **null-safe** (empty string fallback) so a missing postal can't collapse the whole geo map to `"ukn"` via `ImmutableMap`'s null-value NPE.
* `PageAPIGraphQLTypesProvider.java` — register `postal` on the GraphQL geo type.
* `openapi.yaml` — regenerated; `postal` added to the `Geolocation` schema.

### Checklist
- [x] Tests — `GeoIp2CityDbUtilTest` extended: exact-value assertions on `getPostal()`, the `getPostalCode(ip)` accessor, and the JSON serialize round-trip (exercises `equals`/`hashCode`). `Tests run: 4, Failures: 0, Errors: 0`.
- [ ] Translations — n/a
- [x] Security Implications Contemplated — no new input; reads an existing local DB. Postal is already personal-ish geo data on par with city/subdivision already exposed.

### Additional Info
Postal-based personalization **rules** (rules-engine conditionlet + Angular UI) are intentionally out of scope — the conditionlet is distance-only today and would need new comparison types (separate, larger effort).

`openapi.yaml` was regenerated via `./mvnw compile -pl :dotcms-core`; the committed diff matches the generator output exactly, so the CI openapi drift-check stays green.

Refs: #36232

This PR fixes: #36232

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/visitor/filter/characteristics/GeoCharacter.java
+++ b/dotCMS/src/main/java/com/dotcms/visitor/filter/characteristics/GeoCharacter.java
@@ -21,11 +21,13 @@ public class GeoCharacter extends AbstractCharacter {
             try {
                 String ipAddress = visitor.getIpAddress().getHostAddress();
                 GeoIp2CityDbUtil geo = GeoIp2CityDbUtil.getInstance();
+                final String postalCode = geo.getPostalCode(ipAddress);
                 m = new ImmutableMap.Builder<String, String>().put("g.latLong", geo.getLocationAsString(ipAddress))
                     .put("g.countryCode", geo.getCountryIsoCode(ipAddress))
                     .put("g.cityName", geo.getCityName(ipAddress))
                     .put("g.continent", geo.getContinent(ipAddress))
                     .put("g.company", geo.getCompany())
+                    .put("g.postalCode", postalCode != null ? postalCode : "")
                     .build();
             } catch (Exception e) {
                 m = ImmutableMap.of("g.ip", "ukn");
--- a/dotCMS/src/main/java/com/dotcms/graphql/business/PageAPIGraphQLTypesProvider.java
+++ b/dotCMS/src/main/java/com/dotcms/graphql/business/PageAPIGraphQLTypesProvider.java
@@ -333,6 +333,8 @@ public enum PageAPIGraphQLTypesProvider implements GraphQLTypesProvider {
                 new PropertyDataFetcher<Geolocation>("subdivision")));
         geoFields.put("subdivisionCode", new TypeFetcher(GraphQLString,
                 new PropertyDataFetcher<Geolocation>("subdivisionCode")));
+        geoFields.put("postal", new TypeFetcher(GraphQLString,
+                new PropertyDataFetcher<Geolocation>("postal")));
         geoFields.put("ipAddress", new TypeFetcher(GraphQLString,
                 new PropertyDataFetcher<Geolocation>("ipAddress")));
 
--- a/dotCMS/src/main/java/com/dotcms/util/GeoIp2CityDbUtil.java
+++ b/dotCMS/src/main/java/com/dotcms/util/GeoIp2CityDbUtil.java
@@ -331,6 +331,29 @@ public class GeoIp2CityDbUtil {
 	}
 
 	/**
+	 * Returns the postal/zip code the specified IP address belongs to, when the
+	 * MaxMind database provides it.
+	 *
+	 * @param ipAddress
+	 *            - The IP address to get information from.
+	 * @return The postal/zip code, or {@code null} if the database has none for
+	 *         this IP.
+	 * @throws UnknownHostException
+	 *             If the IP address of a host could not be determined.
+	 * @throws IOException
+	 *             If the connection to the GeoIP2 service could not be
+	 *             established, or the result object could not be created.
+	 * @throws GeoIp2Exception
+	 *             If the IP address is not present in the service database.
+	 */
+	public String getPostalCode(String ipAddress) throws UnknownHostException,
+			IOException, GeoIp2Exception {
+		InetAddress inetAddress = InetAddress.getByName(ipAddress);
+		CityResponse cityResponse = getDatabaseReader().city(inetAddress);
+		return cityResponse.getPostal() != null ? cityResponse.getPostal().getCode() : null;
+	}
+
+	/**
 	 * Returns the {@link TimeZone} associated with location, as specified by
 	 * the <a href="http://www.iana.org/time-zones">IANA Time Zone Database</a>.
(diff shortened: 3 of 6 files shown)


dsolistorres (approved):

dotCMS/src/main/java/com/dotcms/visitor/filter/characteristics/GeoCharacter.java
```diff
@@ -21,11 +21,13 @@ public GeoCharacter(AbstractCharacter incomingCharacter) {
             try {
                 String ipAddress = visitor.getIpAddress().getHostAddress();
                 GeoIp2CityDbUtil geo = GeoIp2CityDbUtil.getInstance();
+                final String postalCode = geo.getPostalCode(ipAddress);
                 m = new ImmutableMap.Builder<String, String>().put("g.latLong", geo.getLocationAsString(ipAddress))
                     .put("g.countryCode", geo.getCountryIsoCode(ipAddress))
                     .put("g.cityName", geo.getCityName(ipAddress))
                     .put("g.continent", geo.getContinent(ipAddress))
                     .put("g.company", geo.getCompany())
+                    .put("g.postalCode", postalCode != null ? postalCode : "")
```
dsolistorres, line 30: It seems here we are returning an empty string when postalCode is null, but in GraphQL we are returning null instead of empty string, is that ok?

dsilvam: Good catch — the divergence is intentional.

`GeoCharacter` builds a Guava `ImmutableMap<String, String>`, which throws an NPE on a null *value*. If `getPostalCode(...)` returned null and we put it straight in, the whole `try` block would blow up and fall into the `catch`, collapsing the **entire** geo map to `ImmutableMap.of("g.ip", "ukn")` — wiping `g.countryCode`, `g.cityName`, `g.continent`, etc. for that visitor. So the empty-string fallback here is a guard against that regression, not a style choice.

Everywhere the value flows through the real model/contract — `Geolocation.getPostal()`, the visitor JSON, `$visitor.geo.postal`, and the GraphQL geo type — we keep it as `null`, which is the honest representation of "MaxMind has no postal for this IP" and matches how the sibling fields already behave (e.g. the serialize test asserts `"company":null`). Emitting `""` there would misrepresent missing data as a present-but-empty value.

Net: the legacy `$visitor.g` string-map can't hold null so it uses `""`; the model/GraphQL layer can and does preserve `null`. Happy to add a short code comment noting the Guava constraint if that helps future readers.

Issue #17401, opened 2019-09-30
Searching by `urltitle` not giving results


dsilvam:
When using the Content Search Screen to filter content by `urltitle` we are not getting any results back. This is after [changing default ES text analyzer from whitespace to standard](https://github.com/dotCMS/core/issues/16141).

Steps to reproduce:
1. Go To Content Search screen
2. Filter by content type 'News'
3. Paste `the-gas-price-rollercoaster` into the Search field

No results are back.

This works in 5.1.6

dsilvam:
PR: https://github.com/dotCMS/core/pull/17402

dsilvam:
Note to QA:
In addition to test the `urlTitle` in both "Search" and "URL Title" fields, please test partial searches like: 

- [ ]  `Ga`

- [ ] `Gas`

- [ ] `Gas Pr`

- [ ] `Gas Price`

dsilvam:
PR 2: https://github.com/dotCMS/core/pull/17429

erickgonzalez:
Working as it should, when searching with special characters at the `Search Field` return records.

But we have an issue with any other field, that is not escaping special characters, reported here:  https://github.com/dotCMS/core/issues/17434

bryanboza:
Fixed, tested on master and now we are returning results

PR #36652, opened 2026-07-21
feat(evergreen-tracks): manual-only standard/trailing promotion, drop daily cron

### LINKED ISSUE #36520: Rebuild track promotion engine to honor 2-week maintenance-window cadence
# Rebuild track promotion engine to honor 2-week maintenance-window cadence

_Parent epic: dotCMS/core#35693 · Workstream A (Launch mechanics) · Target: Jul 22_

## Background
This is a **follow-on to the existing promotion engine** built in dotCMS/core#36160, which lives in the repo at `core/cicd/evergreen-tracks/` (planner/executor/markers with taint + hold + forward-only). That engine currently advances `standard` and `trailing` on a **daily** age check (age ≥ ~14 / ~28 days). The launch decision is that Standard and Trailing recalculate **every two weeks on the existing maintenance-window schedule**, not continuously. The engine's planner and cron need to be reworked so a track only moves on a scheduled maintenance window, giving customers a predictable "your next possible update is on window date Y" answer.

## Scope
- Change the promotion trigger from a daily cron to the existing maintenance-window schedule (every two weeks); `latest` may continue to advance on each release cut.
- Rework the planner so `standard`/`trailing` compute their target as of the maintenance window, not the calendar day — a release only becomes eligible for a track on a window boundary.
- Preserve existing forward-only, taint, and hold semantics from dotCMS/core#36160.
- Emit/log the next scheduled window date so downstream comms (doc, picker) can surface it.
- Update `cicd/evergreen-tracks/` planner tests and fixtures for window-based eligibility.
- Validate end-to-end against the `dotcms/dotcms-test` sandbox.

## Out of scope
- The registry/executor mechanics themselves (built in dotCMS/core#36160).
- Customer-selectable cadence (explicitly deferred by the epic).
- Any control-plane/API work (Workstream B).

## Acceptance criteria
- [ ] `standard`/`trailing` only re-point on a maintenance-window boundary, verified in the sandbox across at least two simulated windows.
- [ ] Planner is deterministic for a given window date and set of tags; unit tests cover the boundary cases.
- [ ] Forward-only / taint / hold behavior from dotCMS/core#36160 still holds under the new schedule.
- [ ] The next scheduled window date is exposed in the promotion run output/logs.
- [ ] README in `cicd/evergreen-tracks/` updated to describe the 2-week cadence.

## Depends on
- Builds directly on dotCMS/core#36160 (promotion engine). This is the first launch task.

## Blocks
- dotCMS/core#36521 (windows described in docs), dotCMS/private-issues#653 (enablement), dotCMS/infrastructure-as-code#7987 (canary move), dotCMS/dotusage#71 (telemetry).

_Overlap note: closely related to existing issue dotCMS/core#36160 — this is a follow-on cadence change, not a duplicate._

Part of the Evergreen Release Tracks epic. Parent: dotCMS/core#35693.

---

## Concrete scope (from Steve, 2026-07-11)

1. **Cron schedule:** adjust the promotion schedule so tag re-points cannot cause updates to deploy outside the maintenance window (no unexpected off-window deployments).
2. **Floating-tag drift detection:** build a tool — like the one in the OVH cloud-control cluster — that detects a SHA change under a floating tag and ensures the cluster actually picks it up.
3. **Release-job isolation + regression test:** when the release job triggers the floating-tag move, it must move `latest` only — never `standard`/`trailing` (moving those could trigger a deployment/pod roll). Cover this with a test so it can't regress later.

sfreudenthaler:
## What

`standard`/`trailing` track tags now move **only on a manual `evergreen-tracks-promote` dispatch** — there is no cron. A human running the action at a maintenance-window tag update is the cadence gate, so tracks can never re-point off-window. `latest` is unaffected: it still moves per-release, unattended, via `cicd_6-release.yml` (`--tracks latest`).

This reflects the launch decision to hold off on any scheduled promotion until the customer-experience team green-lights automatic (e.g. daily) movement.

## Manual dispatch is a two-stage, human-approved flow

- **`plan`** job — always runs a dry-run and prints the intended tag moves.
- **`apply`** job — waits on the `evergreen-tracks-apply` environment's **required-reviewer gate**. Nothing moves until a human reviews the plan output and approves the deployment.

> - [x]  **One-time repo setup:** Settings → Environments → `evergreen-tracks-apply` → Required reviewers. Until reviewers are configured, the apply job runs unattended — the gate is the protection rule, not the YAML.

## Why the 2-week planner rework in the issue was NOT built

The original issue scoped a "compute the target as of a 2-week maintenance window" planner rework. Manual invocation makes that unnecessary — the operator *is* the scheduler — so the planner's age-threshold logic is unchanged. Simplest thing that satisfies the launch decision.

## Changes

- **`cicd_evergreen-tracks-promote.yml`** — drop the daily `schedule:` cron; split into `plan` (dry-run) → `apply` (gated on `evergreen-tracks-apply` environment). Removed the redundant `apply` input.
- **Regression test** — pins `--tracks latest` to move **only** `latest`, never standard/trailing (concrete-scope item 3: a release cut must not roll pods on those tracks).
- **`cli.py` / `cicd_6-release.yml`** — refreshed stale "daily cron ages standard/trailing" comments.
- **README** — documents the new cadence and the approval gate.

## Out of scope (tracked separately)

- Floating-tag drift detection (concrete-scope item 2) → its own follow-on issue #36653 .

## Test

`uv run pytest` → 65 passed (includes the new isolation test).

Closes: #36520

### DIFF
--- a/.github/workflows/cicd_6-release.yml
+++ b/.github/workflows/cicd_6-release.yml
@@ -200,7 +200,9 @@ jobs:
     permissions:
       contents: read
     # Share the registry-mutation lock with the promote/admin workflows so this
-    # on-demand promote can never race the daily cron or an admin run.
+    # on-demand promote can never race a manual promote apply or an admin run.
+    # The promote workflow holds this lock only in its post-approval apply job,
+    # so a promotion pending human approval never queues this job.
     concurrency:
       group: evergreen-tracks-registry
       cancel-in-progress: false
--- a/.github/workflows/cicd_6-release.yml
+++ b/.github/workflows/cicd_6-release.yml
@@ -172,9 +172,10 @@ jobs:
   # Promote latest - move the floating `latest` Docker tag to this GA release.
   #
   # evergreen-tracks (cicd/evergreen-tracks) is the single controller of the
-  # latest/standard/trailing tags. Its daily cron ages standard/trailing, but
-  # `latest` must move the instant a GA ships — so the release pipeline invokes
-  # the same engine on-demand, scoped to the latest track only (--tracks latest).
+  # latest/standard/trailing tags. standard/trailing move only on a manual
+  # evergreen-tracks-promote dispatch, but `latest` must move the instant a GA
+  # ships — so the release pipeline invokes the same engine on-demand, scoped to
+  # the latest track only (--tracks latest).
   # This replaces the old deploy-docker `latest: true` path (now `latest: false`).
   #
   # Runs for every GA release from main on dotcms/core by default — moving latest
--- a/.github/workflows/cicd_evergreen-tracks-promote.yml
+++ b/.github/workflows/cicd_evergreen-tracks-promote.yml
@@ -66,9 +66,13 @@ jobs:
           TRAILING_DAYS: ${{ github.event.inputs.trailing_days }}
         run: |
           echo "repo=$REPO (dry-run) — review this plan, then approve the apply job"
+          # --tracks standard,trailing: this workflow never moves `latest` — the
+          # release pipeline owns it and moves it unattended (possibly mid-approval).
+          # Scoping here keeps the drift check below immune to `latest` churn.
           # 2>/dev/null: keep only the engine's stdout plan, drop uv's stderr chatter.
           PLAN=$(uv run evergreen-tracks promote \
             --repo "$REPO" \
+            --tracks standard,trailing \
             --standard-days "$STANDARD_DAYS" \
             --trailing-days "$TRAILING_DAYS" 2>/dev/null)
           echo "$PLAN"
@@ -101,10 +105,11 @@ jobs:
   # and leaving tags half-moved.
   #
   # `apply` re-derives the plan from live registry state at run time and FAILS
-  # if it differs from the approved plan (a GA landed, a hold/taint changed,
-  # etc.), so it can never move tags nobody reviewed. Standard/trailing age
-  # slowly, so drift is rare — when it happens, re-dispatch to review the new
-  # plan.
+  # if it differs from the approved plan (e.g. a hold/taint marker changed, or a
+  # release aged past a track threshold during a long approval), so it can never
+  # move tags nobody reviewed. The plan is scoped to standard,trailing, so an
+  # unattended `latest` move by the release pipeline mid-approval does NOT trip
+  # it. Drift is rare — when it happens, re-dispatch to review the new plan.
   apply:
     needs: [ plan, gate ]
     runs-on: ubuntu-${{ vars.UBUNTU_RUNNER_VERSION || '24.04' }}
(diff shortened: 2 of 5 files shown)


sfreudenthaler:
### Addressed the review findings (commit b8fde8e4e2)

**🟠 High — approval wait held the shared registry lock, could stall unattended `latest`.** Confirmed the root cause: a job waiting on an environment required-reviewer gate acquires its concurrency slot *before* the gate evaluates ([community discussion #17401](https://github.com/orgs/community/discussions/17401)), so moving the lock to job-level on `apply` alone would **not** have helped — the waiting job would still hold `evergreen-tracks-registry`.

Fix: split the approval into its own lock-free `gate` job. `apply` now `needs: [plan, gate]` and carries the job-level `evergreen-tracks-registry` lock, so it only enters concurrency evaluation *after* approval — the lock covers just the fast mutation, never the human wait. A GA release's `promote-latest` can now move `latest` immediately even while a standard/trailing approval sits pending. Workflow-level concurrency removed.

**🟡 Medium — approved plan ≠ what runs.** Documented in the workflow + README that `apply` re-derives its plan from live registry state, so the approval authorizes the operation (repo + thresholds), not the exact digest set the plan printed. Given standard/trailing age slowly and the operator controls timing, a documented re-derive is the right cost — no plan-pinning machinery added.

**Verified** end-to-end against `dotcms/dotcms-test` ([run 29798854833](https://github.com/dotCMS/core/actions/runs/29798854833)): `plan` → `gate` waits for approval → approved → `apply` moved the tags. 66 unit tests green.

sfreudenthaler:
### 🟡 Medium now enforced, not advisory (commit 2328ea4c67)

Per follow-up: `apply` no longer trusts that the approved plan still holds — it **re-derives the plan from live registry state and fails if it differs** from what was approved (a GA landed, a hold/taint changed, a digest was re-tagged). On drift it prints both plans and exits non-zero; the operator re-dispatches to review the new plan. So a stale approval can never move tags nobody reviewed.

Implementation (kept minimal):
- `cli.py` logs to stdout instead of stderr, so the workflow captures the plan text cleanly (`2>/dev/null` drops uv's chatter).
- `plan` job exposes its dry-run plan as a job output; `apply` re-derives, compares (sorted — held-track lines come from a set), and fails on mismatch. Also fails on an empty plan so a broken capture can't silently pass.

**Verified:** comparison logic passes on identical/reordered/no-move plans and fails on drift/empty (local); and an end-to-end dispatch ([run 29799379094](https://github.com/dotCMS/core/actions/runs/29799379094)) logged `plan unchanged since approval; applying` before moving the tags. 66 unit tests green.

sfreudenthaler:
### 🟡 Scoped manual promote to `standard,trailing` (commit 3174160d2b)

Good catch — the drift check was diffing the full plan including `latest`, but `latest` moves unattended via the release pipeline during the approval wait (the exact thing the separate `gate` job enables). A GA shipping mid-approval would have changed the re-derived plan's `latest ->` line and failed `apply`, forcing a needless re-dispatch even though standard/trailing were untouched — undoing the concurrency decoupling at the plan-diff layer.

Fix: all three promote invocations (plan capture, apply re-derive, apply) now pass `--tracks standard,trailing`. This workflow never moves `latest` anyway (the release pipeline owns it), so scoping is also just *more correct* and keeps the "one engine, two triggers" model intact. Verified the scoped dry-run emits no `latest` line. 66 tests green.

erickgonzalez (approved):

PR #36669, opened 2026-07-21
feat(contenttype): add optional 'system' query param to exclude system types (#36072)

### LINKED ISSUE #36072: [TASK] Content type list endpoint — support excluding system types + global sort across base types

ihoffmann-dot:
## Issue

Part 1 of #36072 (Content Drive "New" button content-type picker).

`GET /api/v1/contenttype` returns non-creatable **system** content types (e.g. `dotFavoritePage`, Host) that share a base type with real content, so the `type` base-type filter can't exclude them. Frontends strip them client-side (`!ct.system`, `-1` count adjustment), which breaks pagination counts.

## Change

Adds an optional `system` query param to `GET /api/v1/contenttype`:

- **`system` = "include system types", Boolean, default `true`.** Absent or `true` ⇒ current behavior (system types included). `system=false` ⇒ excludes them (`and structure.system = false`). No system-only mode.
- Threaded as a **dedicated predicate**, independent of the `search`/condition string, so it **ANDs with the `filter`** (name search). Previously the string was name-search XOR condition, so they couldn't combine.
- Applied in **both** the multi-type UNION path (`searchMultipleTypes`) and the single-type path (`dbSearch`), and in the **count path** (`countForSites` → `dbSearch`) so `totalResults` stays exact — no client-side `-1`.
- New API/factory method overloads are additive; existing signatures delegate with `includeSystemTypes = true` (backward compatible).
- `openapi.yaml` regenerated (endpoint is not `@Hidden`) — diff is only the new `system` param.

## Testing

`ContentTypesPaginatorTest` (registered in `MainSuite1a`) — **16 tests, 0 failures**:
- Multi-type: `system` omitted ⇒ system types present + exact total (backward compat); `system=false` ⇒ only regular types, exact total, no system type returned.
- Single-type: `filter` + `system=false` combine ⇒ only the regular type, exact total.

## Rollback safety

Additive optional API param + query changes. No schema/migration, no ES mapping. N-1 ignores the new param. **Rollback-safe.**

This PR fixes: #36072

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/contenttype/business/ContentTypeAPI.java
+++ b/dotCMS/src/main/java/com/dotcms/contenttype/business/ContentTypeAPI.java
@@ -509,7 +509,11 @@ public interface ContentTypeAPI {
    *                            passed by param. e.g:
    *                            offset = 0 -> start from the first record
    *                            offset = 10 -> start from the #11 record
-   * @param requestedContentTypes The Content Types that are explicitly requested to be included.
+   * @param requestedContentTypes The "ensure" set: Velocity Variable Names (or keys) of Content
+   *                            Types that must always be included in the result, regardless of the
+   *                            {@code condition} (name search) and any Site filter. This is not a
+   *                            Site filter -- it guarantees inclusion. Pass {@code null} or an empty
+   *                            list for no forced inclusions.
    * @return List of Content Types Objects
    * @throws DotDataException Error occurred when performing the action.
    */
@@ -527,7 +531,11 @@ public interface ContentTypeAPI {
    * @param limit                Amount of results
    * @param offset               Start position of the resulting list
    * @param hostId               hostId where the content type lives, pass null to bring from all sites.
-   * @param requestedContentTypes The Content Types that are explicitly requested to be included.
+   * @param requestedContentTypes The "ensure" set: Velocity Variable Names (or keys) of Content
+   *                            Types that must always be included in the result, regardless of the
+   *                            {@code condition} (name search) and any Site filter. This is not a
+   *                            Site filter -- it guarantees inclusion. Pass {@code null} or an empty
+   *                            list for no forced inclusions.
    * @param includeSystemTypes   When {@code false}, excludes system Content Types via a dedicated
    *                             {@code and system = false} predicate. When {@code true}, behavior
    *                             is unchanged.
@@ -581,8 +589,10 @@ public interface ContentTypeAPI {
    *                  by param. e.g:
    *                  offset = 0 -> start from the first record
    *                  offset = 10 -> start from the #11 record
-   * @param includeContentTypeIds
-   *                  The Content Types that are explicitly required to be included.
+   * @param includeContentTypeIds The "ensure" set: Velocity Variable Names (or keys) of Content
+   *                  Types that must always be included in the result, regardless of the
+   *                  {@code condition} (name search) and the Site filter. This is not a Site filter
+   *                  -- it guarantees inclusion. Pass {@code null} or an empty list for none.
    *
    * @return The list of {@link ContentType} objects matching the specified search criteria.
    *
@@ -603,7 +613,11 @@ public interface ContentTypeAPI {
    * @param orderBy              The order-by clause, which is internally sanitized by this Factory.
    * @param limit                The maximum number of returned items in the result set.
    * @param offset               Start position of the result list.
-   * @param includeContentTypeIds The Content Types that are explicitly required to be included.
+   * @param includeContentTypeIds The "ensure" set: Velocity Variable Names (or keys) of Content
+   *                             Types that must always be included in the result, regardless of the
+   *                             {@code condition} (name search) and the Site filter. This is not a
+   *                             Site filter -- it guarantees inclusion. Pass {@code null} or empty
+   *                             for none.
    * @param includeSystemTypes   When {@code false}, excludes system Content Types via a dedicated
    *                             {@code and system = false} predicate. When {@code true}, behavior
    *                             is unchanged.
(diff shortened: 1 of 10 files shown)


dotCMS/src/main/java/com/dotcms/contenttype/business/ContentTypeAPI.java
```diff
@@ -550,6 +593,30 @@ List<ContentType> search(final List<String> sites, final String condition,
...
+   * Returns a list of Content Types living in the specified list of Sites, optionally excluding
+   * system Content Types.
+   *
+   * @param sites                The list of one or more Sites to search for Content Types.
+   * @param condition            Allows you to add more conditions to the query via SQL code.
+   * @param base                 The {@link BaseContentType} to search for.
+   * @param orderBy              The order-by clause, which is internally sanitized by this Factory.
+   * @param limit                The maximum number of returned items in the result set.
+   * @param offset               Start position of the result list.
+   * @param includeContentTypeIds The Content Types that are explicitly required to be included.
```
dario-daza: The description is a bit confusing, especially regarding the `includeContentTypeIds` parameter. Does this method retrieve a list of content types by site IDs, or does it return only the content types explicitly required to be included?

dotCMS/src/main/java/com/dotcms/rest/api/v1/contenttype/ContentTypeResource.java
```diff
@@ -1715,7 +1715,14 @@ public final Response getContentTypes(@Context final HttpServletRequest httpRequ
             @QueryParam(ContentTypesPaginator.ENSURE) @Parameter(schema = @Schema(type = "string"),
                     description = "Guarantee Content Types to be included in the response: " +
                             "Comma-separated content type keys (e.g. `activity, blog, product`)."
-            ) String ensureContentTypesParam) throws DotDataException {
+            ) String ensureContentTypesParam,
```
dario-daza, line 1718: Shall we include a postman test to this? to assert the paths when we send a system=true/false

ihoffmann-dot: Done!

dotCMS/src/main/java/com/dotcms/contenttype/business/ContentTypeAPI.java
ihoffmann-dot: Done!

dario-daza (approved):

PR #36670, opened 2026-07-21
feat(content-drive): add JSON/Story Block/Custom/Binary/Key-Value field filters (#36383)

### LINKED ISSUE #36383: [TASK] Content Drive: field-based search/filter criteria reusing edit-content field components

zJaaal:
## Proposed Changes

https://github.com/user-attachments/assets/cfb3ec97-49c5-4d46-9f2f-7050f2eea8f9

Extends Content Drive field-based search (#36383) to the remaining **text-backed** field types the legacy "View Contentlets" content search only ever offered via its plain-textbox fallback: **JSON, Story Block, Custom, Binary, and Key/Value**. They are offered as *contains* filters.

The key enabler: the backend Lucene-query handlers for all five **already exist** and are **already wired into Content Drive's query builder** (`BrowserAPIImpl` → shared `FieldHandlerRegistry`). So the backend change is just widening the resolver allowlist — no new query logic and no handler-registry change.

### Backend
- **`ContentDriveFieldFilterResolver`** — route `JSONField` / `StoryBlockField` / `CustomField` / `BinaryField` / `KeyValueField` to `RoutingBucket.INDEX` (`isIndexRouted`) and accept them as a single `SCALAR` contains term (`supportsKind`). They're INDEX-routed exactly like Text/Select/Date — the DB-first contract (ADR-0018) is unchanged.
- Out of scope (still rejected with 400): **Hidden, Constant, File/Image, Host/Folder, dividers.** File/Image index only the linked asset identifier (not a filename/content) so a contains match isn't usable; Constant is identical on every contentlet (no-op); Hidden/dividers per product decision.
- Unit test: positive cases for the five new types; the out-of-scope rejection test repointed to `HiddenField`.

### Frontend
- **`constants`** — `FIELD_FILTER_TEXT_FALLBACK_TYPES` (`JSON-Field`, `Story-Block`, `Custom-Field`, `Binary`) + `FIELD_FILTER_KEY_VALUE_TYPE`, added to `USER_SEARCHABLE_FIELD_TYPES` (this alone lists them in the "More" menu).
- **Text-fallback types** reuse the existing text control (the `$control` default) → a plain contains filter. No component change needed.
- **Key/Value** — a **single input** with a `key:value` shorthand and a hint (*"Use key:value to match an exact pair"*). The literal input is stored (so the chip and URL round-trip verbatim); `parseUserSearchableValue` translates it when building the `userSearchable` payload:
  - `key:value` → joined `key_value` term → **exact pair** match against the indexed `.key_value` subfield.
  - a bare term → passed through → **loose** contains (matches a key *or* a value).
  - a raw colon is never sent (that path is metadata-only and wouldn't match a regular Key/Value field).
- i18n: `content-drive.field-filter.key-value.placeholder` / `.hint`.

### How Key/Value matching was verified

Confirmed end-to-end against a running instance (create a Key/Value content type → publish content → query `/api/content/query`):
- `.key_value` is stored as `(key + "_" + value).toLowerCase()`, **analyzed** (whitespace-tokenized) and **case-insensitive**.
- The joined `key_value` term is the only **precise** pair match; a bare term is **position-agnostic** (matches a key or a value) — documented so it isn't mistaken for a bug.
- The legacy per-key query format (`<field>.<key>:value`) no longer matches the current index, and the `key:value` colon syntax is metadata-only — hence the FE sends the joined/underscore form.

### Also in this PR — folder pagination fix

The table paginator's `$totalItems` only checked `page.hasMoreContent`, so a folder holding **only sub-folders** (`hasMoreContent: false`, `hasMoreFolders: true`) computed the exact total from the current page and PrimeNG offered no next page — the list was stuck at one page (e.g. 60 of 120 folders). Now it offers another page when **either** content or folders have more. Regression test added; the two "exact total" tests updated to set `hasMoreFolders: false` for the true "no more" case.

### Also in this PR — special-character escaping in filter terms

A field-filter term containing a Lucene query-syntax character was wrapped in `*term*` **unescaped**, so ES failed to parse the query and the **whole search errored** (not just missed) — e.g. a Binary filter `quarterly-report`, a time `12:30`, or a URL `https://…`. The per-field strategies now escape the term via a **vendor-neutral `LuceneQueryUtils.escape`** before adding their own `*` wildcards:

- `TextFieldStrategy` (Text/Textarea/WYSIWYG/Select/Radio/Multi-Select/Checkbox/JSON/Story Block/Custom) — escapes each token in the contains case; the explicit quoted-phrase (`"…"`) case is untouched.
- `BinaryFieldStrategy` — escapes the file-name term.
- `KeyValueFieldStrategy` — escapes the plain (colon-less) term Content Drive sends; the metadata `key:value` shorthand path is left as-is.

Details:
- **Not just hyphens** — the whole Lucene set is escaped (`+ - ! ( ) : ^ [ ] " { } ~ * ? | & / \`), and it also **neutralizes Lucene injection** (`x") OR title:*` becomes a literal). `+`, `,`, `|` and whitespace stay token delimiters.
- **Still a wildcard contains** — the `*…*` the strategy adds are unchanged; only the term *inside* is made literal.
- **URLs are now searchable** on the text-content fields — a `:`/`/`-bearing term matches the stored value via the `_dotraw` keyword sub-field instead of erroring.
- **Scope:** only the per-field searchable strategies. The global keyword/title search (`GlobalSearchAttributeStrategy`) is untouched, so raw `+`/`-` keyword behavior there is unchanged. Clean terms are a no-op through `escape`, so **no existing `LuceneQueryBuilderTest` assertion changes**.
- **Vendor-neutral escape (review by @fabrizzio-dotCMS):** the escape lives in `LuceneQueryUtils.escape` — a pure loop over the stable `query_string` special-char set — instead of `org.apache.lucene…QueryParser`, which is only on the classpath transitively via the ES client and would vanish with the ES→OS migration. The Lucene import is dropped from all three strategies (and the unused Lucene imports from `LuceneQueryUtils`). A unit test asserts **byte-for-byte parity** with the original `QueryParser.escape`, so the refactor changes no behavior.

### Also in this PR — Content Drive JSON-body escaping (the escaping was silently broken)

The escaping above fixed the *Lucene* parse error, but exposed a second bug: Content Drive's hybrid search builds its ES request by interpolating the Lucene query into a hand-written **JSON template** (`ES_QUERY_TEMPLATE`). The escaped term contains a backslash (`angular\-cms`), and `\-` is an **invalid JSON escape** — so ES rejected the request body with `json_parse_exception` and the search **silently returned nothing** for *any* special-char term. Confirmed live against a running instance via `/api/v1/drive/search`. Fixed by JSON-escaping the query before interpolation (`jsonEscape`, escaping `\` and `"`). The modern `/api/v1/content/search` path was never affected — it passes the query to `QueryBuilders.queryStringQuery(...)`, which does no JSON interpolation. No-op for any query without a `\`/`"`, so every previously-working browse/search is byte-identical.

### Also in this PR — Binary filenames match by exact/hyphenated name

`BinaryFieldStrategy` queried only the **analyzed** file-name field, which the analyzer tokenizes on `-`/`/`/`.` (`doc-dev-blue-cold.pdf` → `[doc, dev, blue, cold, pdf]`) — so a hyphenated term like `doc-dev-blue-cold` never matched, only single tokens (`doc`). It now queries the analyzed field **OR** its `_dotraw` keyword (which stores the whole file name), exactly like `TextFieldStrategy`. The `_dotraw` keyword is already indexed for every field — this is a **query-side** change only, no reindex. It's additive (OR), so it can only match *more*, never fewer. Shared with `/api/v1/content/search`, which gains the same (improved) exact-filename behavior. Verified on a running instance that `file_dotraw` holds the full filename and matches; IT asserts the hyphenated `quarterly-report` now matches.

### Rollback safety

**Category: M-3 (REST/Headless API contract change) — 🟡 MEDIUM. Additive, recoverable.**

This PR widens `ContentDriveFieldFilterResolver`'s allowlist so JSON / Story Block / Custom / Binary / Key-Value are accepted as `SCALAR` INDEX-routed filters (previously rejected with 400), and the frontend is updated in the same PR to offer them — the M-3 signal.

- **No 🔴/🟠 risk:** no DB migration, no ES mapping change (the Lucene handlers already existed — only the resolver allowlist changed), no new content-type field type (these are pre-existing types, newly *permitted* as filters).
- **Not a public API:** `/api/v1/drive/search` is an internal endpoint powering the **in-development** Content Drive admin UI (still behind the "in development" banner), not a public/documented headless API. There are no third-party consumers — the "contract" is between first-party FE and BE that ship and roll back together, so the practical rollback surface is only in-house admin usage of a pre-GA feature.
- **Rollback impact:** if the backend is rolled back to N-1 while N's frontend is still deployed/cached — or a user opens a bookmarked/saved Content Drive URL with a chip for one of these new field types — N-1's resolver rejects that filter with a 400. Old filter types and all other search paths are unaffected.
- **Recoverable:** the store catches the failed search into an error state (the portlet doesn't crash); the user recovers by removing that chip or clearing filters. Impact is scoped to a single saved filter on a single content type.
- **Ops note:** a rollback of these field-filter types isn't fully clean *if* they were already in use — flag as "rollback degrades new Content Drive filter types" rather than assuming zero impact.

## Testing
- `pnpm nx test portlets-content-drive` — green (**801 tests**); new coverage for the menu eligibility of the new types, the text-fallback controls, the Key/Value input + hint + verbatim storage, and the `parseUserSearchableValue` Key/Value translation (`key:value`→`key_value`, bare term, partial, empty).
- `ContentDriveFieldFilterResolverTest` — green (new-type routing + repointed rejection).
- `FieldStrategyEscapingTest` — fast unit test (runnable locally, JDK 25, **10/10 green**) asserting the escaped query strings across Text, Binary and Key/Value strategies: clean (no-op), hyphen, colon, the full operator set (`/`, `(`, `)`), a neutralized injection (`x")OR`), the preserved quoted phrase, and the Binary analyzed-OR-`_dotraw` clause.
- `LuceneQueryUtilsTest` — asserts `LuceneQueryUtils.escape` output and **byte-for-byte parity** with the original `QueryParser.escape` across a sample set (the guard that the vendor-decoupling changed no behavior).
- `BrowserAPIImplTest` — unit-tests `jsonEscape` (the Content Drive JSON-body fix) with and without a hyphen: a clean query is unchanged, a Lucene-escaped hyphen term has its backslash doubled for valid JSON, and a quoted phrase is escaped.
- **Integration** (`ContentDriveFieldFilterTest`, dotcms-integration): end-to-end `driveSearch` coverage that each new type actually matches — JSON/Custom (contains via index), Key/Value (joined `color_red` = exact pair; bare `color` = loose), Binary (contains on file **name**), Story Block (contains on stored text), plus `testHyphenatedTermIsEscapedAndMatches` (a hyphenated Text term matches via the escaped query instead of erroring) and a hyphenated-filename assertion on Binary (`quarterly-report` matches via `_dotraw`). Compiles under JDK 25; runs under the integration harness (PostgreSQL + ES).

### Follow-up (not in this PR)
`KeyValueFieldStrategy` mishandles a raw `key:value` colon on a regular Key/Value field (emits a malformed query); the FE works around it by sending the joined form. Worth fixing in the shared strategy separately — flagged to its owner.

This PR fixes: #36383


core-web/libs/portlets/dot-content-drive/portlet/src/lib/utils/functions.ts
```diff
@@ -468,9 +469,45 @@ export function parseUserSearchableValue(
...
+ * @param {string} raw - The literal value the user typed (also what's kept in the URL/chip).
+ * @return {*}  {(string | undefined)}
+ */
+function toKeyValueTerm(raw: string): string | undefined {
+    const trimmed = raw.trim();
+    if (!trimmed) {
+        return undefined;
+    }
+
+    const separator = trimmed.indexOf(':');
```
ihoffmann-dot, line 509: Using that separator with a bare value that contains a colon (URLs like https://…, times like 12:30, ratios like 16:9) will be misread as key:value and joined with _, it silently matches nothing.

Could we document the greedy shorthand explicitly? Maybe a short note on the method.

ihoffmann-dot (approved):

core-web/libs/portlets/dot-content-drive/portlet/src/lib/utils/functions.ts
zJaaal: Good catch, thanks. Documented the greedy shorthand on `toKeyValueTerm` in 0ca43f3f31.

One nuance worth noting: the split is on the **first** colon only, so a *keyed* colon-bearing value is preserved — `link:https://x` → `link_https://x`, `start:12:30` → `start_12:30` (added a test for that). The genuinely ambiguous case is only a **bare** colon value with no key (e.g. just `12:30`), which is inherently indistinguishable from `key:value`; the doc now tells users to prefix such values with their key. Left the behavior as-is since it's a deliberate shorthand, but happy to add an escape hatch if you'd prefer.

_(reply authored by Claude on @zJaaal's behalf)_

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-field-filter/dot-content-drive-field-filter.component.ts
```diff
@@ -235,6 +242,18 @@ export class DotContentDriveFieldFilterComponent {
         return type === 'integer' ? 'numeric' : type === 'decimal' ? 'decimal' : 'text';
     });
     protected readonly $textPlaceholderKey = computed(() => {
+        // Field types whose contains-search targets something more specific than "the value" get a
+        // clarifying placeholder (Binary matches the file name, not its contents).
+        const byFieldType: Record<string, string> = {
+            [FIELD_FILTER_BINARY_TYPE]: 'content-drive.field-filter.binary.placeholder',
+            [FIELD_FILTER_JSON_TYPE]: 'content-drive.field-filter.json.placeholder',
```
oidacra, line 249: Would it be worth adding placeholder assertions for the JSON-Field and Story-Block branches? Right now `$textPlaceholderKey` maps three types (Binary, JSON, Story-Block) but only Binary and the generic Custom fallback are asserted in the spec — the JSON/Story-Block entries are exercised only by the control-rendering table (which checks the control testid, not the placeholder), so a typo or dropped map entry in either wouldn't be caught by any test. Two `it`s mirroring the existing Binary placeholder test would close the gap.

oidacra (approved):

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-field-filter/dot-content-drive-field-filter.component.ts
zJaaal: Addressed!

zJaaal: Good call — added in 3368ad019a. Two `it`s mirroring the Binary placeholder test now assert the JSON (`content-drive.field-filter.json.placeholder`) and Story-Block (`content-drive.field-filter.story-block.placeholder`) branches, so a dropped/typo'd map entry in either is caught.

_(reply authored by Claude on @zJaaal's behalf)_

dario-daza (approved):

fabrizzio-dotCMS:
### Backend / ES→OS migration: confine the Lucene `QueryParser` dependency to a neutral util

The three strategies (`BinaryFieldStrategy`, `KeyValueFieldStrategy`, `TextFieldStrategy`) now import `org.apache.lucene.queryparser.classic.QueryParser` directly. That artifact (`lucene-queryparser:8.7.0`) is **only** on the classpath transitively via the Elasticsearch client (`elasticsearch-rest-high-level-client:7.10.2 → elasticsearch:7.10.2 → lucene-queryparser`) — it is **not** declared in `bom/application/pom.xml`, and the modern `opensearch-java` client we use for OS 3.x does **not** bring it. So once the ES client is dropped as part of the ES→OS migration, `QueryParser` disappears from the classpath and these REST-layer strategies fail with `NoClassDefFoundError`.

The behavior itself is fine — `QueryParser.escape()` is a pure static string transform whose `query_string` output is valid Lucene syntax on both ES and OS — so this isn't a blocker, but it deepens a vendor coupling across three more classes in the REST layer.

**Suggested fix — Option B: reimplement the escape in a neutral dotCMS util (best aligned with the project's "no vendor types past the adapter" rule).** `QueryParser.escape()` is literally a ~15-line loop over a fixed, stable special-character set (`+ - && || ! ( ) { } [ ] ^ " ~ * ? : \ /`). We already have `com.dotmarketing.util.LuceneQueryUtils` — add an `escape(String)` there and drop the vendor import entirely. Zero dependency risk, and it confines the concept to one place. The character set is a stable contract of `query_string` syntax that ES and OS honor identically.

Alternative (Option A) if you'd rather not reimplement: declare `org.apache.lucene:lucene-queryparser` explicitly in the BOM so it survives ES-client removal. Option B is preferred since it removes the Lucene artifact from the equation altogether.

zJaaal:
@fabrizzio-dotCMS good catch — implemented **Option B**. The escape now lives in a vendor-neutral `LuceneQueryUtils.escape(String)` (a pure loop over the fixed `query_string` special-char set — `\ + - ! ( ) : ^ [ ] " { } ~ * ? | & /`), and the `org.apache.lucene…QueryParser` import is dropped from all three strategies (`Text`/`Binary`/`KeyValue`). I also removed the 5 **unused** Lucene imports that were already sitting in `LuceneQueryUtils`, so that class is now genuinely Lucene-free too.

To guarantee the reimplementation changed nothing, `LuceneQueryUtilsTest` asserts **byte-for-byte parity** with `org.apache.lucene…QueryParser.escape` across a sample set (it's still on the test classpath via the ES client, so we can compare directly). Green.

Two related fixes rode along in the same commit while verifying this on a live instance:
- **JSON-body escaping** — the strategy escaping fixed the *Lucene* parse error but exposed that Content Drive interpolates the query into a hand-built JSON template; the `\` from the escape was an invalid JSON escape → `json_parse_exception` → the search silently returned nothing for any special-char term. Fixed by JSON-escaping before interpolation.
- **Binary `_dotraw`** — `BinaryFieldStrategy` now matches the analyzed file name **OR** its `_dotraw` keyword, so hyphenated/exact filenames match (previously only single analyzer tokens did).

_(Posted by Claude Code on @zJaaal's behalf.)_

fabrizzio-dotCMS (approved):

PR #36678, opened 2026-07-21
feat(sdk): Implement SDK compatibility handshake server advertises version + minimum supported SDK, SDKs warn on mismatch

### LINKED ISSUE #36609: SDK compatibility handshake: server advertises version + minimum supported SDK, SDKs warn on mismatch
### Description

**Problem**

- With release Tracks, a customer's environments (dev / staging / prod) run different dotCMS versions at different times.
- Client SDKs are published on npm in lockstep with the dotCMS version (dotCMS `26.07.13` → npm `26.7.13`).
- Most dotCMS releases do not change the SDK contract. Some do.
- Customers have no reason to update the SDK when nothing requires it, so deployed SDKs drift behind the server over time.
- Nothing tells a customer that the SDK deployed in their app is no longer supported by the dotCMS version an environment was just updated to.
- Source: customer feedback on Tracks (per-environment update notifications, SDK versioning best practices).

**Solution**

A version handshake: the server advertises, the SDK compares and warns.

**Server**

- Introduce a release-level constant `MIN_SDK_VERSION`: the oldest SDK version supported by this server release. Bumped only when a release breaks SDK compatibility; most releases never touch it.
- Every REST API response includes two static headers emitted from constants (no request parsing, no per-request logic):

```
X-DotCMS-Version:  26.7.13
X-DotCMS-Min-SDK:  26.5.1
```

**SDK** (`@dotcms/client`; `react`/`angular` packages inherit through it)

- Inject the package version at build time so the SDK knows its own version at runtime.
- Read `X-DotCMS-Min-SDK` from responses it already makes. If own version < min, log one console error per session:

```
[dotCMS SDK] SDK 26.5.1 is not supported by dotCMS 26.7.13. Upgrade required: <docs link>
```

- If own version > `X-DotCMS-Version`, log a console warning: the SDK is newer than the server and may call APIs the server does not have. Catches code built against a newer environment (dev) running against an older one (prod).
- Version comparison must treat versions as numeric calver segments (`26.10.1` > `26.7.13`), not string compare.

Tracks updates environments staggered (dev before prod), so the warning surfaces in the customer's dev console weeks before production updates. The customer's development workflow becomes the early-warning system.

### Acceptance Criteria

- [ ] Server: `MIN_SDK_VERSION` exists as release metadata with a documented bump procedure (release checklist: "does this release break SDK compatibility? If yes, bump `MIN_SDK_VERSION` to this release's version")
- [ ] Server: all REST API responses include `X-DotCMS-Version` and `X-DotCMS-Min-SDK` headers
- [ ] Server: both headers are listed in `Access-Control-Expose-Headers` so browser-side SDK calls can read them cross-origin
- [ ] SDK: own version available at runtime via build-time injection, matching the published npm version
- [ ] SDK: when own version < `X-DotCMS-Min-SDK`, logs a single console error per session including server version, minimum required version, and upgrade link
- [ ] SDK: when own version > `X-DotCMS-Version`, logs a single console warning per session that the SDK is newer than the server
- [ ] SDK: numeric calver comparison; no warning and no behavior change when versions are compatible or when the headers are absent (older servers)
- [ ] Docs: compatibility behavior and the two headers documented on the SDK documentation page

### Priority

High

KevinDavilaDotCMS:
## Videos

Without missmatch:

https://github.com/user-attachments/assets/14921ec1-3eea-4b12-a5b9-04cfe26f2109

With missmatch:

https://github.com/user-attachments/assets/a71f6943-9b04-4bc7-87e5-24045d5ec6a9

## Summary

Implements the version handshake described in #36609: the server advertises its own
version and the oldest SDK version it still supports via two response headers;
`@dotcms/client` reads them off responses it already makes and warns in the console when
the installed SDK is outside the compatible range. Built directly on top of the
date-lockstep versioning work in #36587/#36633 (ADR-0019) — this feature is only possible
because the SDK's own version now reliably matches a real dotCMS release version.

## Problem this solves

With Evergreen Tracks, a customer's environments (dev/staging/prod) run different dotCMS
versions at different times, and SDKs are published in lockstep with the dotCMS version.
Most releases don't change the SDK contract, so customers have no reason to update the
SDK when nothing requires it — deployed SDKs drift behind the server over time, and
nothing today tells a customer their SDK is no longer supported by the dotCMS version an
environment was just updated to.

## What changed — Backend (Java)

- **`MinSdkVersion.java`** (new, `com.dotcms.rest.config`) — a single manually-maintained
  constant: the oldest `@dotcms/*` SDK version this dotCMS instance still supports.
  Starts at baseline `"0.0.0"` (no breaking change has been introduced under this
  mechanism yet, so every previously published SDK version is still considered
  compatible). See "Decisions" below for the bump procedure.
- **`SdkVersionWebInterceptor.java`** (new, `com.dotcms.filters.interceptor.meta`) — adds
  two static headers to every response:
  ```
  X-DotCMS-Version:  <ReleaseInfo.getVersion()>
  X-DotCMS-Min-SDK:  <MinSdkVersion.VALUE>
  ```
  Registered in `InterceptorFilter.addInterceptors()`, alongside the existing
  `ResponseMetaDataWebInterceptor` (the `x-dot-server` header) — same mechanism, same
  file, same pattern.
- **`Access-Control-Expose-Headers`** — no change needed. `dotmarketing-config.properties`
  already defaults `api.cors.default.Access-Control-Expose-Headers` to `*`, so both new
  headers are already exposed cross-origin for browser-side SDK calls.

### Why a `WebInterceptor` and not a JAX-RS `ContainerResponseFilter`

The first implementation used a JAX-RS `ContainerResponseFilter` (`SdkVersionHeaderFilter`,
now removed). It worked for JAX-RS resources (confirmed via `curl` against
`/api/v1/appconfiguration`) but **not** for GraphQL (`/api/v1/graphql`), because
`DotGraphQLHttpServlet` is a plain `HttpServlet`, not a JAX-RS resource — it never enters
the Jersey filter chain a `ContainerResponseFilter` hooks into. Since `@dotcms/client`'s
page-loading flow goes through GraphQL by default, the JAX-RS-only filter silently missed
the majority of real SDK traffic (confirmed by manual end-to-end testing against a local
instance — see "Testing" below).

`InterceptorFilter` is the first filter in the pipeline and maps every request regardless
of how it's ultimately handled (JAX-RS or plain servlet) — the exact reason
`ResponseMetaDataWebInterceptor` already uses this mechanism for `x-dot-server` instead of
a JAX-RS filter. Switched to the same approach for full coverage.

## What changed — SDK (`@dotcms/client` only)

Only `libs/sdk/client` needed changes. `react` and `angular` never make their own raw
`fetch` calls — they route everything through `@dotcms/client`'s `FetchHttpClient`, so they
inherit the check automatically, exactly as the issue describes.

- **`sdk-compatibility.ts`** (new) — pure, dependency-free comparison logic:
  - Numeric calver comparison (`26.10.1 > 26.7.13`), not string comparison.
  - `checkSdkCompatibility(headers, ownVersion)`: reads `X-DotCMS-Version` /
    `X-DotCMS-Min-SDK` off a `Headers` object; logs `console.error` once per session if
    `ownVersion < minSdkVersion`; logs `console.warn` once per session if
    `ownVersion > serverVersion`.
  - Fails open by design: missing headers (older server), unparsable versions, or any
    unexpected exception → silently does nothing, never throws, never changes
    request/response behavior.
  - `sdk-compatibility.spec.ts` — 11 unit tests, 100% line coverage.
- **`rollup.config.cjs`** — a small, dependency-free virtual-module Rollup plugin
  (`sdkVersionPlugin`) injects the package's own version (from its own `package.json`,
  already set to the exact dotCMS release version by the release pipeline before this
  build runs) as a build-time constant, importable as `virtual:sdk-version`. No new npm
  dependency added — verified `@nx/rollup`'s `withNx()` concatenates a `plugins` array
  passed in rather than replacing its own, so this doesn't interfere with Nx's generated
  plugins.
- **`fetch-http-client.ts`** — one fire-and-forget call added right after
  `await fetch(url, options)`: `checkSdkCompatibility(response.headers, SDK_VERSION)`.
- **`virtual-modules.d.ts`** (new) — ambient `declare module 'virtual:sdk-version'` so
  TypeScript accepts the import.
- **`__mocks__/virtual-sdk-version.ts`** + **`jest.config.ts`** `moduleNameMapper` — Jest
  never runs the Rollup build, so the virtual module is mapped to a real stub file for
  unit tests.

### 1. Why `MIN_SDK_VERSION` (a floor) instead of a simple `ownVersion !== serverVersion` check

Considered and rejected a simpler "warn on any mismatch" approach. The issue explicitly
calls this out as the reason a floor is needed, not equality:

> Most dotCMS releases do not change the SDK contract. Some do.

With Tracks, a customer's SDK almost never exactly equals the current server version at
any given moment — most releases don't require an SDK update at all. Warning on any
mismatch would fire constantly for nearly every customer regardless of real
incompatibility, training people to ignore it (alert fatigue) and defeating the point of
the signal. `MIN_SDK_VERSION` only moves when a release actually breaks the contract, so
the warning stays meaningful.

### 2. `MIN_SDK_VERSION` bump procedure — manual, approximate, decided as "Option A"

`MIN_SDK_VERSION` is manually maintained, not computed — per the issue's own AC
("documented bump procedure"). The exact mechanics weren't specified in the issue, so we
picked between:

- **Option A (chosen):** when a PR breaks SDK compatibility, the developer sets
  `MinSdkVersion.VALUE` to whichever `@dotcms/client` version is already published as
  `latest` on npm at PR time. Doesn't need to be exact — it only needs to correctly gate
  SDKs older than what's needed.
- **Option B (rejected):** try to set it to the exact future dotCMS release version this
  PR will ship in. Rejected because that version is a date assigned only when a release is
  actually cut (per ADR-0019's date-lockstep scheme) — it isn't known yet at PR-review
  time, so this isn't practically achievable.

This is documented directly in `MinSdkVersion.java`'s Javadoc for future contributors.

### 3. "Once per session" is what the issue specifies, not an addition

Both the description and both relevant AC bullets say "per session" explicitly:
`logs a single console error per session` / `logs a single console warning per session`.
Implemented via two module-level boolean flags. Note the practical implication: "per
session" server-side means per Node.js process lifetime (a long-running production server
might log this once across days/weeks), not per page reload — which matches the issue's
own stated intent of this being a **developer-console early-warning system** (Tracks
updates dev before prod, so the signal reaches a developer's local console before a
production rollout), not a continuous ops-monitoring signal.

### 4. Baseline `MIN_SDK_VERSION` value: `"0.0.0"`

No breaking change has been introduced under this mechanism yet, so the correct, honest
starting value is "no SDK is too old" — `0.0.0` is lower than every real published
version, so nothing is flagged as incompatible until the first real bump happens.

## Testing performed

- **Unit tests:** `sdk-compatibility.spec.ts` — 11/11 passing, 100% line coverage.
  Full `sdk-client` suite (298 tests) still green after the change, including the
  pre-existing `fetch-http-client.spec.ts`.
- **Lint:** clean (`nx lint sdk-client`).
- **Real Rollup build:** ran `nx build sdk-client` and inspected the actual compiled
  `dist/libs/sdk/client/index.cjs.js` — confirmed `SDK_VERSION` is correctly inlined
  (matching `package.json`'s version exactly), no leftover unresolved
  `virtual:sdk-version` import, and the compatibility-check internals are not leaked into
  the public `.d.ts` API surface.
- **Backend compile:** `./mvnw compile -pl :dotcms-core -DskipTests` → `BUILD SUCCESS`,
  zero warnings attributable to the new/changed files.
- **Manual end-to-end test**, real local dotCMS instance (`just build && just dev-run`) +
  a real Next.js example app, packing `@dotcms/client` with `npm pack` and installing the
  `.tgz`:
  - Confirmed via `curl` that both headers appear on JAX-RS responses
    (`/api/v1/appconfiguration`).
  - **Found and fixed a real gap** this way: `curl` against `/api/v1/graphql` (the actual
    endpoint the Next.js example's page load hits) was missing both headers with the
    original JAX-RS-filter implementation — this is what led to discovering the
    GraphQL-is-a-plain-servlet issue and switching to `SdkVersionWebInterceptor` (see
    "Why a WebInterceptor" above). Re-verified via `curl` that the GraphQL endpoint
    returns both headers correctly after the fix.
  - Verified the outdated-SDK console error fires correctly and matches the expected
    message format, using a real packed/installed SDK build with a temporarily lowered
    `MIN_SDK_VERSION` floor to force the mismatch.
  - Confirmed the once-per-session throttle behaves as designed (logged once across two
    consecutive requests in the same server process).

## Out of scope / follow-ups

- SDK documentation page update describing the two headers and compatibility behavior
  (issue AC item — not part of this PR, should be tracked as a fast-follow doc task).
- `examples/*` pinning to exact versions instead of ranges — separate, pre-existing
  follow-up from ADR-0019 (Implementation Note 6), unrelated to this change.
- LTS-specific `MIN_SDK_VERSION` policy — not addressed here; LTS SDK publishing itself
  is still an open follow-up from #36587.

## Acceptance criteria

- [x] Server: `MIN_SDK_VERSION` exists as release metadata with a documented bump
      procedure (documented in `MinSdkVersion.java`'s Javadoc)
- [x] Server: all REST/GraphQL responses include `X-DotCMS-Version` and `X-DotCMS-Min-SDK`
      headers (verified via `curl` against both a JAX-RS resource and the GraphQL servlet)
- [x] Server: both headers are exposed via `Access-Control-Expose-Headers` (already
      defaulted to `*`, verified in `dotmarketing-config.properties`)
- [x] SDK: own version available at runtime via build-time injection, matching the
      published npm version (verified in the actual compiled bundle)
- [x] SDK: console error once per session when `ownVersion < X-DotCMS-Min-SDK`
- [x] SDK: console warning once per session when `ownVersion > X-DotCMS-Version`
- [x] SDK: numeric calver comparison; fails open (no warning, no behavior change) when
      headers are absent or unparsable
- [ ] Docs: compatibility behavior and the two headers documented on the SDK
      documentation page — **not done in this PR**, tracked as a follow-up

This PR fixes: #36609

This PR fixes: #36609


KevinDavilaDotCMS:
Went through all 3 findings + the 2 notes against the actual code/diff. One was real and is fixed; the other two are valid observations but don't need a code change.

## Fixed

- **`core-web/libs/sdk/react/package.json` — out-of-scope packaging change**: confirmed via `git diff` this PR moved `@dotcms/client`/`@dotcms/uve` from `dependencies` to `peerDependencies`, unrelated to this feature and never mentioned in the PR description. It slipped in during manual testing, unintentionally. Reverted — `react/package.json` is now identical to `main` again.

## Reviewed, no code change needed

- **Version-scheme mismatch risk (semver vs. date-lockstep)**: the underlying concern is real — once `MinSdkVersion.VALUE` moves off `0.0.0` to a date-lockstep value, any SDK still on an old semver version will correctly compare as "older" and get flagged. On reflection this isn't a false positive, though — a customer still on old semver *after* dotCMS has moved on and broken compatibility genuinely does need to upgrade. The one real edge case (the transition release itself) is already covered by the documented bump procedure in `MinSdkVersion.java`'s Javadoc: bump to whatever's *currently* published as `latest` on npm at PR time, not a future date version — so the first bump during the semver→date-lockstep transition would itself still be a semver value, keeping the comparison consistent. Worth being deliberate about that procedure when the first real bump happens, but not a code issue today.

- **Version disclosure via `X-DotCMS-Version` (fingerprinting)**: checked — dotCMS already exposes the exact running version publicly and unauthenticated today, via `LoginFormResource` (`/api/v1/loginform`, a pre-login endpoint that returns `ReleaseInfo.getVersion()` directly, since it has to be callable before a session exists). This isn't a new attack surface introduced by this PR; it's a second, more convenient way to read information that's already public dotCMS-wide. If suppressing version disclosure entirely is something worth doing, that's a broader, pre-existing conversation unrelated to this specific change.

## Notes, agreed as-is

- `rollup.config.cjs`'s virtual-module plugin only wired into the `build` target, not `build:js` (the esbuild editor target) — correct, and fine for now since that target doesn't reach `FetchHttpClient` today.
- No backend test asserting the interceptor writes the headers — agreed, low value given how trivial the body is; covered by the manual `curl` verification described in the PR description instead.

dario-daza (approved):

dotCMS/src/main/java/com/dotmarketing/filters/InterceptorFilter.java
```diff
@@ -53,6 +54,7 @@ private void addInterceptors(final FilterConfig config) {
         delegate.add(new EMAWebInterceptor());
         delegate.add(new GraphqlCacheWebInterceptor());
         delegate.add(new ResponseMetaDataWebInterceptor());
+        delegate.add(new SdkVersionWebInterceptor());
```
oidacra: `SdkVersionWebInterceptor` is registered after `GraphqlCacheWebInterceptor`, which returns `Result.SKIP_NO_CHAIN` on a cache hit (after writing the cached body itself), breaking the delegate chain before this interceptor runs. That means `X-DotCMS-Version` / `X-DotCMS-Min-SDK` are absent on cached GraphQL responses — and GraphQL caching is on by default with a ~15s TTL. Since the class Javadoc names GraphQL as the main reason for using a WebInterceptor, should this be registered before `GraphqlCacheWebInterceptor`, or should `writeResponse()` set the two headers on the cache-hit path too?

core-web/libs/sdk/client/src/lib/utils/sdk-compatibility.ts
```diff
@@ -0,0 +1,105 @@
...
+        const minSdkVersion = headers.get(DOTCMS_MIN_SDK_HEADER);
+
+        if (!serverVersion || !minSdkVersion) {
+            return;
+        }
+
+        if (!hasWarnedOutdatedSdk && (compareVersions(ownVersion, minSdkVersion) ?? 0) < 0) {
+            hasWarnedOutdatedSdk = true;
+            console.error(
+                `[dotCMS SDK] SDK ${ownVersion} is not supported by dotCMS ${serverVersion} ` +
```
oidacra, line 90: AC5 asks the console error to include an upgrade link (the issue example shows `Upgrade required: <docs link>`), but the message ends with `Install @dotcms/client@<min> or newer.` with no URL. Worth appending a docs/release-notes link so the error points somewhere actionable?

core-web/libs/sdk/client/src/lib/utils/sdk-compatibility.ts
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        const serverVersion = headers.get(DOTCMS_VERSION_HEADER);
+        const minSdkVersion = headers.get(DOTCMS_MIN_SDK_HEADER);
+
+        if (!serverVersion || !minSdkVersion) {
+            return;
+        }
+
+        if (!hasWarnedOutdatedSdk && (compareVersions(ownVersion, minSdkVersion) ?? 0) < 0) {
```
oidacra, line 87: `(compareVersions(...) ?? 0)` treats an unparsable comparison (`null`) the same as "equal", so the check is silently skipped. That hits both ends: an LTS-shaped own version like `26.7.13_lts_v1` never triggers the outdated-SDK error, and an LTS / `-SNAPSHOT` / `UNVERSIONED` server version (all values `ReleaseInfo.getVersion()` can return) never triggers the newer-SDK warning. Would it be worth handling `null` explicitly (distinct from a real `0`), rather than coalescing to 0?

dotCMS/src/main/java/com/dotcms/filters/interceptor/meta/SdkVersionWebInterceptor.java
```diff
@@ -0,0 +1,56 @@
...
+ * {@code ContainerResponseFilter}: a JAX-RS filter only covers requests routed through
+ * Jersey, but SDK traffic also hits plain-servlet endpoints outside JAX-RS entirely — e.g.
+ * GraphQL ({@code /api/v1/graphql}, served by {@code DotGraphQLHttpServlet}, a raw
+ * {@code HttpServlet}, not a JAX-RS resource). {@code InterceptorFilter} is the first filter
+ * in the pipeline and maps every request, so this covers both uniformly — the same reason
+ * {@link ResponseMetaDataWebInterceptor} (the {@code x-dot-server} header) uses this same
+ * mechanism instead of a JAX-RS filter.
+ *
+ * <p>Unrelated to CORS request handling — {@code Access-Control-Expose-Headers} already
+ * defaults to {@code *} (see {@code dotmarketing-config.properties}), so these two headers
```
oidacra: The Javadoc says exposure is free because `Access-Control-Expose-Headers` defaults to `*`. That holds for `api.cors.default`, but there is a resource-scoped `api.cors.graphql.Access-Control-Expose-Headers=Content-Type,Cache-Control` override (in the integration config) that would exclude these two headers for the `graphql` resource specifically. Since GraphQL is the primary SDK path, worth confirming the production `graphql` CORS config exposes `X-DotCMS-Version` / `X-DotCMS-Min-SDK`?

core-web/libs/sdk/client/src/lib/client/adapters/fetch-http-client.ts
```diff
@@ -68,6 +73,12 @@ export class FetchHttpClient extends BaseHttpClient {
             // Use native fetch API - no additional configuration needed
             const response = await fetch(url, options);
 
+            // Fire-and-forget: reads X-DotCMS-Version / X-DotCMS-Min-SDK off the
+            // response and logs a console warning on mismatch. Fails open (no headers,
+            // e.g. an older server) and never throws, so this can't affect the actual
+            // request/response handling below.
+            checkSdkCompatibility(response.headers, SDK_VERSION);
```
oidacra, line 80: This is the production integration point, but `fetch-http-client.spec.ts` was not updated and does not reference it — nothing verifies the check is called once per successful response with the response headers and `SDK_VERSION`. Worth a spec that mocks `checkSdkCompatibility` and asserts the wiring?

dotCMS/src/main/java/com/dotcms/filters/interceptor/meta/SdkVersionWebInterceptor.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public static final String DOTCMS_VERSION_HEADER = "X-DotCMS-Version";
+    public static final String DOTCMS_MIN_SDK_HEADER = "X-DotCMS-Min-SDK";
+
+    @Override
+    public Result intercept(final HttpServletRequest request, final HttpServletResponse response) {
+        try {
+            response.setHeader(DOTCMS_VERSION_HEADER, ReleaseInfo.getVersion());
+            response.setHeader(DOTCMS_MIN_SDK_HEADER, MinSdkVersion.VALUE);
+        } catch (Exception e) {
```
oidacra, line 53: This `catch (Exception e)` swallows the exception with no log line. Fail-open (never break the request) is the right intent, but per the backend logging standard, would a `Logger.debug(this, ...)` here help make a systemic failure of the header advertisement discoverable instead of the feature silently going dark?

KevinDavilaDotCMS:
Went through all 6 inline findings from this latest round of review against the actual code on this branch. Applied 4, left 1 as-is (already correct/intentional), and left 1 documentation-only. No behavior change beyond what's described below.

## Applied

**1. Interceptor registration order (`InterceptorFilter.java`) — real bug, now fixed**
Confirmed: `SdkVersionWebInterceptor` was registered *after* `GraphqlCacheWebInterceptor`. On a GraphQL cache hit, `GraphqlCacheWebInterceptor.intercept()` writes the response and returns `Result.SKIP_NO_CHAIN`, breaking the delegate chain before this interceptor ever runs — so `X-DotCMS-Version`/`X-DotCMS-Min-SDK` were silently absent on cached GraphQL responses (~15s TTL, on by default), which is exactly the primary path this WebInterceptor approach was chosen to cover. Moved `SdkVersionWebInterceptor` to register before `GraphqlCacheWebInterceptor`; verified `writeResponse()` on the cache-hit path doesn't reset headers, so headers set earlier in the chain survive onto the cached response too.

**2. Missing upgrade link in the outdated-SDK console error (`sdk-compatibility.ts`)**
AC5 in the source issue explicitly calls for the console error to include an upgrade link (`Upgrade required: <docs link>`). The shipped message ended at `Install @dotcms/client@<min> or newer.` with no URL. Added a docs link to the message, following the existing `https://www.dotcms.com/docs/latest/<slug>` convention used elsewhere in this SDK — the exact page (`sdk-version-compatibility`) still needs to be authored per this issue's own "Docs" acceptance criterion.

**3. Swallowed exception with no log line (`SdkVersionWebInterceptor.java`)**
The `catch (Exception e)` around the two `setHeader` calls had no logging at all. Fail-open (never break the request) is correct, but a systemic failure of the whole feature would have gone completely invisible. Added `Logger.debug(this, ...)` in the catch block — doesn't change the fail-open behavior, just makes it discoverable.

**4. Javadoc overclaim on CORS exposure (`SdkVersionWebInterceptor.java`)**
Confirmed `api.cors.graphql.Access-Control-Expose-Headers=Content-Type,Cache-Control` is a real, supported per-resource override (used today in the integration-test config, resolved by `CorsFilter`'s per-resource-then-default fallback) that isn't present in the production default config — so out of the box this is fine, but any environment that sets that override for the `graphql` resource specifically would silently stop exposing these two headers on the primary SDK path. Softened the Javadoc to note this instead of asserting exposure unconditionally.

## Reviewed, no code change

**`compareVersions(...) ?? 0` treating an unparsable comparison as "equal"**
Checked the actual semantics: since both call sites only ever test the result with `< 0` or `> 0` (never equality), `(null ?? 0) < 0` and `(null ?? 0) > 0` are both `false` — identical to an explicit `result !== null && result < 0` check. There's no behavioral difference between the current code and the suggested change; a `null` (unparsable/LTS-shaped version) already correctly results in "don't warn" either way. This is also the same fail-open-on-unparsable behavior the previous review round already looked at and confirmed as intentional. Leaving as-is.

## Test coverage added (not a "finding" fix per se, but related)

Added a spec to `fetch-http-client.spec.ts` asserting `checkSdkCompatibility` is actually called with the response headers and the SDK version on the production wiring path (`fetch-http-client.ts`) — this had zero coverage before. Ran both this suite and `sdk-compatibility.spec.ts` locally; all tests pass, including with the updated error message from fix #2.

Also re-ran `mvn compile -pl :dotcms-core` locally to confirm the `InterceptorFilter`/`SdkVersionWebInterceptor` changes build clean.

oidacra (approved):

PR #36722, opened 2026-07-24
chore(sdk): Bring back 'next' tag publish on sdks

### LINKED ISSUE #36701: Reintroduce next npm dist-tag for internal dev/QA testing on merge to main
### Description

Follow-up to #36587/PR #36633 (date-lockstep SDK versioning), which removed the standalone `@next` prerelease channel entirely — SDK packages are now only published to the `latest` npm dist-tag, and only when a real dotCMS release is cut (`cicd_release-sdk.yml`, triggered by `release: created`). Nothing publishes to npm on a plain merge to `main` anymore.

This created a gap for internal QA: the team runs a post-merge QA phase against "real" environments, not local code (`npm pack` + `.tgz` install isn't usable there). Without a `next` channel, there's no installable, non-production SDK version to point those environments at between merges and the next real release cut.

This issue reintroduces a `next` npm dist-tag, published automatically on merge to `main` when SDK files change — but explicitly as an **internal dev/QA-only channel**, not a customer-facing one, and without reopening the "second door for an off-version SDK publish" concern ADR-0019 originally removed `next` to close. The `latest`/date-lockstep mechanism (real releases, `MinSdkVersion`, the compatibility handshake) is not touched at all by this change.

**Naming scheme**

The `next` version is computed at publish time — no local `VERSION` file is reintroduced:

1. Query the currently-published `latest` version live from npm (e.g. `npm view @dotcms/client version`).
2. Append `-next.<github.run_number>` (unique, monotonically increasing per workflow run).

Example: if `latest` is currently `26.3.17-1`, a trunk run publishes `26.3.17-1-next.87`. Verified with Node's `semver` package that this format is valid strict and loose semver — safe for `npm publish`.

Because this string always carries an explicit `-next.` prerelease suffix, it can never be mistaken for a real date-lockstep release version — and it's never something we tell customers to install.

**Proposed changes**

1. **`deploy-javascript-sdk/action.yml`** — add a new optional input `npm-tag` (default `'latest'`); use it in place of the currently hardcoded `--tag latest` in the publish step. Everything else in the action (build, per-package version rewrite, idempotency check) is reused unchanged for both tags.
2. **`cicd_3-trunk.yml`** — re-add SDK-changed detection and a `next`-publish step:
   - Reuse the **existing** `sdk_libs` change-detection filter (defined in `.github/filters.yaml`, computed by `cicd_comp_initialize-phase.yml`) — this filter was never deleted in #36587/#36633, only the wiring in trunk that consumed it for publishing was removed.
   - Add a step that computes the `next` version as described above and calls `deploy-javascript-sdk` with `npm-tag: next`, gated on the `sdk_libs` filter matching and `github.event_name != 'workflow_dispatch'` (mirroring the exact condition the pre-#36587 mechanism used).

**Explicitly not touched**

`cicd_comp_deployment-phase.yml`, `cicd_release-sdk.yml`, `cicd_6-release.yml`, `cicd_comp_release-prepare-phase.yml`, `MinSdkVersion.java`, `SdkVersionWebInterceptor.java`, and all SDK-compatibility-check TypeScript code (`sdk-compatibility.ts` and friends) — the real-release, date-lockstep, and version-handshake mechanisms are unaffected.

**Why this doesn't reopen the ADR-0019 "second door" concern**

- `next` versions are never date-lockstep-shaped, so they can never satisfy or interfere with the one-dotCMS-version⇄one-SDK-version invariant, which only concerns `latest`.
- The SDK's own compatibility check (`sdk-compatibility.ts`'s `parseVersionSegments`/`compareVersions`) already fails open (returns `null`, no warning logged) on a version string like `26.3.17-1-next.87`, since the `next`/run-number segments aren't plain integers — verified against the existing, unmodified parsing logic. No code change needed there for this safety property to hold.
- `next` is purely an internal dev/QA convenience; it must never be documented or recommended to customers as an installable channel.

### Acceptance Criteria

- [ ] `deploy-javascript-sdk/action.yml` has a new optional `npm-tag` input (default `latest`), used in the publish step instead of the hardcoded value
- [ ] `cicd_3-trunk.yml` computes a `next` version from the currently-published `latest` npm version + `-next.<run_number>` suffix, with no local `VERSION` file reintroduced
- [ ] `cicd_3-trunk.yml` only publishes to `next` when SDK files changed (reusing the existing `sdk_libs` filter) and the run is not a manual `workflow_dispatch`
- [ ] End-to-end verified: a real push to `main` with SDK changes results in a new `@dotcms/client@<version>-next.<n>` published under the `next` dist-tag, installable via `npm install @dotcms/client@next`
- [ ] Confirmed no changes were needed to `cicd_release-sdk.yml`/`cicd_6-release.yml`/`MinSdkVersion`/`SdkVersionWebInterceptor` — the real-release/date-lockstep/compatibility-handshake mechanism remains fully untouched
- [ ] Confirmed a locally-installed `next` SDK build does not trigger any false-positive/false-negative in the SDK compatibility check (fails open due to the unparsable version string)

### Priority

Medium

### Additional Context

**Testing plan**: mirror how #36587/PR #36633 was validated — implement and test first in the `core-workflow-test` fork (dry-run trigger, then a real trunk-style push with SDK changes) before applying the same changes to `dotcms/core`.

**Out of scope**:
- Any change to how `latest` is published or versioned (date-lockstep stays exactly as-is).
- Any change to `MinSdkVersion` or the SDK compatibility handshake (#36609/#36698).
- Publishing a `next` channel for `@dotcms/dotcli` — scope is limited to the SDK libs (`core-web/libs/sdk/*`) only.

**References**:
- #36587 / PR #36633 — removed the original `@next` channel as part of date-lockstep versioning (ADR-0019)
- #36609 / PR #36678 — SDK compatibility handshake (unaffected by this change)
- #36698 — `MIN_SDK_VERSION` release-pipeline automation (separate, unaffected concern)

KevinDavilaDotCMS:
- Updated the  action to include an optional  input for specifying the npm dist-tag (e.g., latest, next) during SDK package publishing.
- Modified the action's description to reflect the new functionality.
- Introduced a new job in the CI workflow to publish SDK packages under the  tag for internal development and QA use, ensuring that it only runs when SDK files change.
- Added Slack notifications for both successful and failed SDK  publishes, providing visibility into the publishing process.

This code is already tested in core-workflow-test: https://github.com/dotCMS/core-workflow-test/actions/runs/30069848966

This PR fixes: #36701

This PR fixes: #36701

### DIFF
--- a/dotcms-integration/src/test/java/com/dotmarketing/portlets/personas/business/PersonaAPITest.java
+++ b/dotcms-integration/src/test/java/com/dotmarketing/portlets/personas/business/PersonaAPITest.java
@@ -61,8 +61,12 @@ public class PersonaAPITest {
     persona3 = new PersonaDataGen().hostFolder(host.getIdentifier()).nextPersisted();
     persona4 = new PersonaDataGen().hostFolder(host.getIdentifier()).nextPersisted();
     
-    allPersonasOnHost  =  personaAPI.getPersonasIncludingDefaultPersona(host, "", false, 100, 0 , null, APILocator.systemUser(), false);
-    assertTrue("total allPersonas should be 5, got:" + allPersonasOnHost._2, allPersonasOnHost._2 == 5);
+    // wait for ES to catch up on the deleteAllPersonas() removals + the 4 personas just created above,
+    // otherwise a not-yet-refreshed stale/leftover persona from another test can inflate this count
+    Awaitility.await().atMost(15, TimeUnit.SECONDS).pollInterval(1, TimeUnit.SECONDS).until(() -> {
+      allPersonasOnHost = personaAPI.getPersonasIncludingDefaultPersona(host, "", false, 100, 0, null, APILocator.systemUser(), false);
+      return allPersonasOnHost._2 == 5;
+    });
   }
 
   private static void deleteAllPersonas() throws Exception{
--- a/dotcms-integration/src/test/java/com/dotmarketing/portlets/personas/business/PersonaAPITest.java
+++ b/dotcms-integration/src/test/java/com/dotmarketing/portlets/personas/business/PersonaAPITest.java
@@ -61,12 +61,8 @@ public class PersonaAPITest {
     persona3 = new PersonaDataGen().hostFolder(host.getIdentifier()).nextPersisted();
     persona4 = new PersonaDataGen().hostFolder(host.getIdentifier()).nextPersisted();
     
-    // wait for ES to catch up on the deleteAllPersonas() removals + the 4 personas just created above,
-    // otherwise a not-yet-refreshed stale/leftover persona from another test can inflate this count
-    Awaitility.await().atMost(15, TimeUnit.SECONDS).pollInterval(1, TimeUnit.SECONDS).until(() -> {
-      allPersonasOnHost = personaAPI.getPersonasIncludingDefaultPersona(host, "", false, 100, 0, null, APILocator.systemUser(), false);
-      return allPersonasOnHost._2 == 5;
-    });
+    allPersonasOnHost  =  personaAPI.getPersonasIncludingDefaultPersona(host, "", false, 100, 0 , null, APILocator.systemUser(), false);
+    assertTrue("total allPersonas should be 5, got:" + allPersonasOnHost._2, allPersonasOnHost._2 == 5);
   }
 
   private static void deleteAllPersonas() throws Exception{
--- a/.github/actions/core-cicd/deployment/deploy-javascript-sdk/action.yml
+++ b/.github/actions/core-cicd/deployment/deploy-javascript-sdk/action.yml
@@ -2,16 +2,22 @@
 #
 # VERSIONING (ADR-0019 — date-lockstep versioning):
 # - The SDK version is always the dotCMS release version it ships with. There is no
-#   independent SDK versioning and no `@next` prerelease channel.
+#   independent SDK versioning.
 # - Single source of truth: the `version` input, supplied by the caller (the dotCMS
 #   release pipeline — see cicd_release-sdk.yml, which derives it from the GitHub
 #   Release tag, mirroring how cicd_release-cli.yml versions @dotcms/dotcli).
-# - Every invocation publishes every SDK package at that exact version to the `latest`
-#   npm dist-tag — including releases where the SDK has no code changes, to preserve
-#   the one-dotCMS-version <-> one-SDK-version invariant.
+# - Every real-release invocation publishes every SDK package at that exact version to
+#   the `latest` npm dist-tag — including releases where the SDK has no code changes, to
+#   preserve the one-dotCMS-version <-> one-SDK-version invariant.
+#
+# `next` TAG (internal dev/QA use only):
+# - cicd_3-trunk.yml also calls this action on merges to main that touch SDK files,
+#   passing npm-tag: 'next' and a version computed from the current `latest` plus a
+#   `-next.<run_number>` suffix. This is never a date-lockstep-shaped version and must
+#   never be recommended to customers — see issue #36701.
 #
 name: 'SDK Publish NPM Packages - Stable'
-description: 'Publish the dotCMS SDK libs on NPM registry at the given version.'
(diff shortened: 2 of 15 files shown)


KevinDavilaDotCMS:
Applied two fixes from the Claude Code review, plus a note on why a third finding is being accepted as-is rather than code-fixed.

**Fix 1 — `continue-on-error: true` on the `next` publish step**

`cicd_comp_finalize-phase.yml` doesn't only look at its own `needs:` — its `prepare-report-data` job independently queries `GET /repos/{repo}/actions/runs/{run_id}/jobs` and scans **every** job's `conclusion` in the run, excluding only names matching `^(Finalize|Final Status|.*[Ff]inalize.*)`. `"Publish SDK (next)"` doesn't match that exclusion, so without this fix a transient npm/registry hiccup on this internal dev/QA publish would mark the entire trunk workflow red. Added `continue-on-error: true` to the publish step so it can't sink the run.

One subtlety this surfaced: simply adding `continue-on-error: true` would have silently broken the existing "Slack Notification (SDK next failure)" step. `continue-on-error` masks a step's *conclusion* to `success`, and the job-level `failure()`/`success()` context functions key off `conclusion`, not `outcome` — so `if: failure()` would never fire again once the step above it has `continue-on-error: true`. Fixed by checking the step's own outcome directly instead: `if: steps.deploy-javascript-sdk.outcome == 'failure'`. `outcome` reflects the real pre-continue-on-error result, so the failure Slack message still fires correctly.

**Fix 2 — fallback text for the failure Slack message**

If "Compute next version" itself fails (e.g. `npm view` fails to resolve the current `latest`), `steps.next-version.outputs.version` is empty, and the failure message would render `` publish version `` with nothing after it. Added a fallback: `${{ steps.next-version.outputs.version || 'unknown — the version-compute step itself failed' }}`.

**Finding #3 — semver precedence / caret-range risk for `next` versions — accepted as a known limitation, not code-fixed here**

`next` versions look like `<latest>-next.<run_number>` (e.g. `26.7.14-1-next.42`), which per semver precedence sorts *below* `26.7.14-1` itself. This is expected and fine for the intended use — devs/QA install it explicitly via `@dotcms/client@next`, never via a bare version range. The real risk is only if something resolves it through a caret range (`^26.7.14-1`), which would never match a prerelease tag like this per semver rules anyway (so in practice this mostly self-protects) — but this is the same class of caret-pinning fragility already tracked as follow-up debt for `examples/*` in the base date-lockstep work (#36587). Not scoping a fix into this PR since nothing in this repo currently consumes `next` versions via a range, but flagging it explicitly so it isn't re-discovered as a surprise later.

dotcms-integration/src/test/java/com/dotmarketing/portlets/personas/business/PersonaAPITest.java
```diff
@@ -61,8 +61,12 @@ public static void initData() throws Exception {
     persona3 = new PersonaDataGen().hostFolder(host.getIdentifier()).nextPersisted();
     persona4 = new PersonaDataGen().hostFolder(host.getIdentifier()).nextPersisted();
     
-    allPersonasOnHost  =  personaAPI.getPersonasIncludingDefaultPersona(host, "", false, 100, 0 , null, APILocator.systemUser(), false);
-    assertTrue("total allPersonas should be 5, got:" + allPersonasOnHost._2, allPersonasOnHost._2 == 5);
+    // wait for ES to catch up on the deleteAllPersonas() removals + the 4 personas just created above,
+    // otherwise a not-yet-refreshed stale/leftover persona from another test can inflate this count
```
nollymar: why was this await interval included in this PR?

nollymar (changes requested):
This PR should not require an IT fix and an await might not be the right approach

dotcms-integration/src/test/java/com/dotmarketing/portlets/personas/business/PersonaAPITest.java
dario-daza: I have included to remove a flaky test, I'll removed

nollymar (approved):

PR #36728, opened 2026-07-25
feat(workflows): Add Claude AI SDK breaking change check workflow

### LINKED ISSUE #36698: Automate MIN_SDK_VERSION bump via release pipeline + AI-based SDK-breaking-change detection
### Description

Follow-up to #36609 (SDK compatibility handshake), which shipped `MinSdkVersion.java` as a manually-edited Java constant. The bump procedure documented there ("Option A": developer sets `MinSdkVersion.VALUE` to whichever `@dotcms/client` version is currently published as `latest` on npm at PR time) works correctly for backend-only breaking changes, but has a real gap when a breaking change *also* requires an SDK-side code fix: the "latest published version at PR time" necessarily predates that fix, so setting the floor to it would incorrectly mark an unfixed SDK version as "supported."

This issue proposes replacing the manual, PR-time Java-constant-edit procedure with a release-pipeline-driven mechanism, so:
- Developers never write a version number by hand or guess a not-yet-existent future date.
- The value only ever gets set at the moment it is actually known (when a release is cut).
- Forgetting to bump it is a loud pipeline failure, not a silent gap.

**Proposed solution**

**1. New release-pipeline input (no separate date field — reuses `release_version`)**

Add a new `workflow_dispatch` boolean input to `cicd_6-release.yml`:

- **`bump_min_sdk_version`** (boolean, default `false`) — "Does this release include a change that breaks SDK compatibility?"

**Important design correction (found while scoping this issue):** the bump must land on **`main`**, on its own — it must *not* be folded into `cicd_comp_release-prepare-phase.yml`'s existing automated commit (the one that also touches `.mvn/maven.config` and `LICENSE`). That commit only ever lives on the disposable `release-${release_version}` branch, which is never merged back into `main` (verified: no step anywhere in the release pipeline merges a release branch back to `main`). That's fine for `LICENSE` and `maven.config`, because `update-license-date.sh` *recomputes* the Change Date fresh from the current date on every run — it never depends on prior state, so it doesn't matter that it never reaches `main`. `MinSdkVersion.VALUE` is the opposite: it's a ratchet that must *persist* across releases. If its bump only ever lived on the release branch, the very next release-prepare run would start again from `main`'s stale value, silently undoing the bump.

This also means **timing matters**, not just placement: the mechanism to update `main` must only fire *after* the release has actually succeeded (`build` + `deployment` both green), not eagerly in `release-prepare`. If `release-prepare` bumped `main` immediately and the release then failed downstream, `main` would advertise a stricter `MIN_SDK_VERSION` for a dotCMS version that was never actually shipped — breaking currently-valid customer SDK installs for no reason. Concretely: this needs to be a new step/job gated on `needs: [release-prepare, build, deployment]` with `if: success()` (likely in `cicd_6-release.yml`'s `release` job, or a new job right after it).

**Confirmed (not just suspected): `main` cannot be pushed to directly — this must go through a PR.** Checked the precedent this repo already used for an almost identical problem: the now-retired `cicd_manual-release-sdks.yml` had to persist a version bump (`core-web/libs/sdk/VERSION`) onto `main` after a release, and its solution was a dedicated step, `Open post-release PR to bump VERSION on main` — it committed to its own branch, force-pushed that branch, then ran `gh pr create --base main` (reusing an existing PR via `gh pr list --head ... --json url` if a previous run had already opened one still pending). Verified via the actual resulting PR (#36563, "chore(sdk): bump SDK version to 1.7.1 after 1.7.0 release"): opened by `github-actions[bot]`, but **merged by a human**, not auto-merged — confirmed no auto-merge mechanism exists anywhere in this repo's workflows.

This means our new mechanism should follow the exact same shape:
- Create a branch, commit the `MinSdkVersion.java` bump there, push it, and `gh pr create --base main` (with the same dedupe-against-an-existing-open-PR check as the retired workflow, in case a previous release run already opened one that's still unmerged).
- **Because a human has to merge this PR, an open-but-unmerged bump PR is itself the same kind of silent gap this issue is trying to eliminate** — if nobody notices it, the ratchet never actually advances even though the release believes it handled it. So this step must also post a Slack notification (reusing the existing `notify-slack` action already used elsewhere in these pipelines) explicitly asking someone to review and merge the PR, not just open it quietly.
- Idempotency still holds the same way as before: nothing lands on `main` until a human merges the PR, so a failed/retried release before that point leaves `main` untouched — there's nothing to roll back. The new step should still no-op (skip opening a duplicate PR) if `MinSdkVersion.VALUE` on `main` already matches this release's version, or if a PR bumping to this same version is already open.

No separate free-text version field is needed regardless: `MIN_SDK_VERSION`, when it needs to move, is always equal to the release doing the bumping — never a separately-typed, potentially-mistyped, or future-guessed value.

**2. AI-based label detection (safety net, mirrors the existing rollback-safety mechanism)**

Add a new workflow, `ai_claude-sdk-breaking-change.yml`, structurally mirroring the existing `ai_claude-rollback-safety.yml`:
- Same org-membership security gate.
- Same "human override" escape hatch: a `Human: SDK Breaking Change` / `Human: Not SDK Breaking` label on the PR skips the AI evaluation.
- Same stale-label-clearing preflight (re-evaluates fresh on every push).
- Calls the same shared reusable workflow (`dotCMS/ai-workflows/.github/workflows/claude-orchestrator.yml@v3`) with a new prompt.
- The prompt evaluates the PR diff against a **new reference document**, `docs/core/SDK_BREAKING_CHANGE_CATEGORIES.md` (to be authored as part of this issue, modeled on the existing `docs/core/ROLLBACK_UNSAFE_CATEGORIES.md`), listing concrete categories of SDK-breaking changes (e.g. removed/renamed GraphQL field the SDK depends on, changed `postMessage` editor protocol shape, changed REST response shape the SDK's TS types model, etc.).
- Applies label `AI: SDK Breaking Change` when a match is found (with an explanatory PR comment, same format as the rollback-safety check), or no label / a "clean" label otherwise.

This mechanism is entirely decoupled from the release pipeline timing change above — it runs per-PR, independent of when/whether a release is ever cut.

Important: treat the AI's label as a strong-assist signal for human PR review, not an infallible autopilot decision — semantic "does this break the SDK contract" judgment is inherently less mechanical than the rollback-safety categories (DB migrations, ES mappings, etc.), so false negatives are a real risk the human reviewer should stay alert to.

**3. Release-time validation (closes the "forgot to check" gap)**

Add a validation step to the release pipeline (likely in `cicd_comp_release-prepare-phase.yml` or a dedicated step in `cicd_6-release.yml`) that:
- Queries merged PRs between the previous release tag and the current release commit (the same kind of tag-range query the existing `.github/scripts/release-qa-status` script already does for the QA-coverage Slack warning).
- Checks whether any of those PRs carry the `AI: SDK Breaking Change` (or human-confirmed) label.
- If such a PR exists and `bump_min_sdk_version` was left `false` for this run, **fail the release explicitly** with a clear error message (mirroring the existing strict, no-silent-fallthrough validation pattern already used for the `skip_latest` input in `cicd_6-release.yml`'s `verify-branch` job).

Unlike the actual `MinSdkVersion.java` bump, this validation only reads PR/label state and doesn't mutate anything — it can safely stay early in the pipeline (e.g. in `release-prepare` or `verify-branch`) to fail fast, before spending a full build+deployment cycle on a misconfigured release.

**Out of scope**
- Any change to the runtime SDK↔CMS compatibility handshake itself (headers, comparison logic, `SdkVersionWebInterceptor`) — that shipped in #36609/PR #36678 and does not change here; this issue is purely about *how* `MinSdkVersion.VALUE` gets updated, not the mechanism that reads/compares it.
- LTS-specific `MIN_SDK_VERSION` policy.

**References**
- #36609 — SDK compatibility handshake (the feature this automates the maintenance of)
- #36587 / PR #36633 — date-lockstep SDK versioning pipeline (ADR-0019), the precedent for automating release-time version handling
- `.github/workflows/ai_claude-rollback-safety.yml` — the mechanism this issue's AI-detection step is modeled on
- `docs/core/ROLLBACK_UNSAFE_CATEGORIES.md` — the reference-doc pattern to follow for the new SDK-breaking-change categories doc
- `.github/scripts/release-qa-status/` — existing precedent for querying merged PRs/labels across a release tag range
- Retired `cicd_manual-release-sdks.yml`'s "Open post-release PR to bump VERSION on main" step (and its resulting PR #36563) — the precedent confirming `main` requires a PR (no direct push, no auto-merge) for this exact kind of post-release persistent-state bump

### Acceptance Criteria

- [ ] `cicd_6-release.yml` has a new `bump_min_sdk_version` boolean input (default `false`)
- [ ] When `true`, a step gated on the release having fully succeeded (`build` + `deployment` both green) opens a PR against `main` with the `MinSdkVersion.java` `VALUE` bump — not folded into `release-prepare`'s release-branch commit, and not a direct push
- [ ] That step posts a Slack notification asking a human to review and merge the bump PR (reusing the existing `notify-slack` action)
- [ ] The step is idempotent: it reuses an already-open bump PR from a previous run instead of opening a duplicate, and no-ops entirely if `main`'s `MinSdkVersion.VALUE` already matches this release's version
- [ ] If the release fails before reaching this step, `main` is left completely untouched with respect to `MinSdkVersion.VALUE` (nothing to roll back)
- [ ] New workflow `ai_claude-sdk-breaking-change.yml` exists, mirroring `ai_claude-rollback-safety.yml`'s security gate, human-override labels, and stale-label-clearing preflight
- [ ] New reference doc `docs/core/SDK_BREAKING_CHANGE_CATEGORIES.md` defines concrete, evaluable categories of SDK-breaking changes
- [ ] AI workflow applies `AI: SDK Breaking Change` label (with explanatory comment) when a match is found against the reference doc
- [ ] Release pipeline fails explicitly (not silently), early (before build/deployment), if a merged PR since the last release tag carries the breaking-change label but `bump_min_sdk_version` was left `false`
- [ ] `MinSdkVersion.java`'s Javadoc (from #36609) is updated to describe this pipeline-driven procedure (bumped via an automated PR against `main`, opened after a successful release, merged by a human) instead of the superseded manual "Option A" (developer edits the constant directly in their PR)

### Priority

High

### Additional Context

Purely a CI/CD process-automation improvement for maintaining `MIN_SDK_VERSION` going forward — does not touch the already-shipped, already-tested runtime handshake feature from #36609.

KevinDavilaDotCMS:
# Automate `MinSdkVersion.VALUE` maintenance via the release pipeline

Closes #36698

## Problem

Follow-up to #36609 / #36678 (the SDK compatibility handshake). That work shipped
`MinSdkVersion.java` — a constant read by `SdkVersionWebInterceptor` and sent to every
`@dotcms/*` SDK consumer via the `X-DotCMS-Min-SDK` response header — but left its maintenance
fully manual: a developer had to remember to edit the constant by hand in their PR whenever a
change broke SDK compatibility ("Option A" from the original ticket: set it to whatever
`@dotcms/client` version was currently `latest` on npm at PR time).

That's fragile in two ways:
- Nothing stops a developer from simply forgetting.
- The "npm latest at PR time" heuristic has a known gap when the breaking change also needs an
  SDK-side fix — the version that's `latest` at PR time necessarily predates that fix.

This PR replaces the manual edit with a release-pipeline-driven mechanism with a safety net at
both ends: PR-time detection and release-time enforcement.

## What this ships

1. **`SDK Breaking Change` label + AI detector** (`ai_claude-sdk-breaking-change.yml`) — runs on
   every PR, evaluates the diff against a new reference doc, and labels PRs that break SDK
   compatibility.
2. **`docs/core/SDK_BREAKING_CHANGE_CATEGORIES.md`** — the reference doc the AI (and human
   reviewers) use to judge what counts as SDK-breaking, grounded in the actual `@dotcms/client`
   surface (GraphQL page/content API, REST responses, the UVE `postMessage` protocol, the
   compatibility headers themselves).
3. **A new `bump_min_sdk_version` release input** on `cicd_6-release.yml` — "does this release
   include an SDK-breaking change?"
4. **A release-time validation gate** — fails the release outright if a merged PR since the last
   release carries the label but the operator left the checkbox unchecked.
5. **A post-release bump job** — once the release fully succeeds, opens a PR (never a direct
   push) bumping `MinSdkVersion.VALUE` to the release version and pings Slack for human review.
6. Updated Javadoc on `MinSdkVersion.java` describing this procedure.

## Design decisions worth knowing about for review

**Why the bump happens *after* the release succeeds, not during `release-prepare`.**
The obvious-looking approach — folding the bump into `release-prepare`'s existing automated
commit (the same one that already updates `LICENSE` and `.mvn/maven.config`) — is wrong for this
value specifically. That commit only ever lives on the disposable `release-${version}` branch,
which is never merged back to `main`. That's *fine* for `LICENSE`, because
`update-license-date.sh` recomputes the date fresh from "today" on every run — it never depends
on prior state. `MinSdkVersion.VALUE` is the opposite: it's a ratchet that must persist forward
across releases. If it only ever lived on the release branch, the next release-prepare run would
silently start again from `main`'s stale value.

That in turn means **timing matters**: the bump can only safely happen once we *know* the release
shipped (build + deployment both green). If we bumped `main` eagerly in `release-prepare` and the
release then failed downstream, `main` would advertise a stricter compatibility floor for a
dotCMS version that customers never actually received — breaking currently-valid SDK installs for
nothing.

**Why it's a PR, never a direct push to `main`.**
Confirmed via the exact historical precedent for this same problem: the now-retired
`cicd_manual-release-sdks.yml` had to persist a version bump (`core-web/libs/sdk/VERSION`) onto
`main` after a release, and its solution was a dedicated "Open post-release PR to bump VERSION on
main" step — never a direct push. The PR it produced (#36563) was opened by `github-actions[bot]`
but merged by a human. There's no auto-merge anywhere in this repo's workflows to lean on instead.

**Why one label, not four.**
An earlier draft of this mechanism mirrored `ai_claude-rollback-safety.yml`'s label scheme
exactly: `AI: X` / `AI: Not X` / `Human: X` / `Human: Not X`. For rollback-safety that split earns
its keep — a "clear stale AI labels on every push" preflight needs to not clobber a human's
deliberate override. For this mechanism we simplified to a single label, **`SDK Breaking
Change`**:
- The AI only ever *adds* it when it detects a break (with an explanatory PR comment). It never
  removes it.
- Removing the label — because a human disagrees with the AI's verdict — is a plain manual
  action, available to anyone with write access. No separate `Human: ...` label needed.
- A human can just as easily *add* the label manually — covering both a false negative from the
  AI, and PRs from external contributors, where the AI never runs at all (the org-membership
  security gate blocks it).
- Net effect: no "stale label" preflight step needed, no risk of the AI clobbering a human
  decision on the next push, one label to reason about.

### 1. PR time
Every PR triggers `ai_claude-sdk-breaking-change.yml`:
- `security-check` — gates on dotCMS org membership (same mechanism as rollback-safety).
- `claude-sdk-breaking-change-check` — reads `docs/core/SDK_BREAKING_CHANGE_CATEGORIES.md`, diffs
  the PR, and either:
  - posts a comment explaining which category matched + adds `SDK Breaking Change`, or
  - does nothing (no label, no comment) if it judges the change non-breaking.

A human reviewer can add or remove the label manually at any time regardless of the AI's verdict.

### 2. Release time — the checkbox is forgotten
Say a PR merged with the `SDK Breaking Change` label (correctly, from the AI or a human), two
more unrelated PRs merge after it, and then someone cuts a release *without* checking
`bump_min_sdk_version`. The very first job, `verify-branch`, runs a new step
(`Validate bump_min_sdk_version against merged PR labels`) that:
- resolves the previous standard release tag,
- pulls every PR merged since that tag (via squash-merge commit messages, same convention the
  existing `release-qa-status` tool relies on),
- checks each one's labels,
- and if any carries `SDK Breaking Change` while `bump_min_sdk_version` is `false`, **fails the
  release immediately**, before `release-prepare` even creates a branch or tag:

  > `::error::This release (v26.10.01-1) includes PR(s) labeled 'SDK Breaking Change' (#36800)
  > since v26.7.14-1, but 'bump_min_sdk_version' was left false. Re-run with
  > bump_min_sdk_version=true, or remove the label from those PRs first if none of them actually
  > break SDK compatibility.`

The operator sees the error, realizes what happened, and re-runs with the checkbox checked.

If the checkbox is checked but *no* labeled PR is found in range, the release still proceeds — it
just prints a non-fatal warning, since a human explicitly opted in and may know something the
labels don't capture:

> `::warning::bump_min_sdk_version=true but no merged PR since v26.7.14-1 carries an SDK-breaking
> label. Proceeding, since a human explicitly opted in — but double-check this is intentional.`

LTS releases (`_lts_v##`) skip this validation entirely — `@dotcms/*` SDKs track the `@latest` npm
cadence, not LTS patch cadence.

### 3. Release time — the release fails partway through
Say the checkbox *was* checked correctly, but the release fails during `build` or `deployment`
for unrelated reasons. The new `bump-min-sdk-version` job is gated with a strict `success()` on
`needs: [release-prepare, build, deployment]` — not the `always() && !failure() && !cancelled()`
idiom used by sibling jobs. If any of those three didn't fully succeed, this job **doesn't run at
all**:
- `MinSdkVersion.VALUE` is not touched.
- No PR is opened.
- No Slack message fires (the job never starts).

Nothing needs to be rolled back, because nothing was ever written. Once the operator fixes the
issue and re-runs the same release successfully, the bump job runs then instead.

### 4. Release time — the release succeeds
Once `release-prepare`, `build`, and `deployment` are all green, `bump-min-sdk-version` runs:
- Checks out fresh `main`.
- **Idempotency check #1**: if `MinSdkVersion.VALUE` already equals this release's version
  (already-bumped, or a retry after a previous bump PR already merged), it no-ops — no duplicate
  commit, no duplicate PR.
- Otherwise, creates branch `sdk/bump-min-sdk-version-<version>` off `main`, edits the constant,
  and **verifies the edit actually took effect** (guards against the `VALUE = "..."` literal
  silently failing to match if the file is ever reformatted — in which case it fails loudly
  instead of quietly opening an empty PR).
- Commits, force-pushes the branch.
- **Idempotency check #2**: reuses an already-open PR for the same branch/version instead of
  opening a duplicate on a retried run.
- Opens the PR (title: `chore(sdk): bump MinSdkVersion.VALUE to <version>`, body explains why),
  and posts to Slack (`log-sdk-libs` channel):

  > :large_orange_circle: **Attention dotters:** dotCMS `26.10.01-1` was released with an
  > SDK-breaking change.
  >
  > A PR bumping `MinSdkVersion.VALUE` to `26.10.01-1` is open on `main`.
  > **Please review and merge the post-release PR ASAP:** [View PR]
  > [View workflow run]

  If the job itself fails partway (e.g. `gh pr create` errors), a separate failure message goes
  to the same channel instead. The PR does **not** auto-merge — a human always reviews and merges
  it manually, same as the retired SDK-VERSION-bump precedent.

## Test scenarios walked through during design (not yet live-fire tested — see Known limitations)

| # | Scenario | Expected outcome | Confirmed by |
|---|----------|-------------------|--------------|
| 1 | AI detects a breaking change; human agrees and leaves the label | Label stays; release-time gate later catches it if the checkbox is forgotten | Design walkthrough + exact error message traced above |
| 2 | AI flags a PR as breaking; human disagrees | Human removes `SDK Breaking Change` manually; no separate label needed, no re-add risk since the AI never removes/re-adds on its own | Confirmed the AI check only ever *adds*, per the workflow's own prompt instructions |
| 3 | AI misses a real breaking change (false negative), or PR is from an external contributor (AI never runs — blocked by the org-membership gate) | A human adds `SDK Breaking Change` manually; the release-time gate treats it identically to an AI-applied label | Confirmed via the gate's label check (`grep -qxF 'SDK Breaking Change'`) — no distinction by origin |
| 4 | Dev forgets `bump_min_sdk_version` on a release that includes a labeled PR | Release fails immediately in `verify-branch`, before any release branch/tag is created | Exact `::error::` message traced above |
| 5 | Dev checks `bump_min_sdk_version` but release fails in `build`/`deployment` | `bump-min-sdk-version` job doesn't run at all — nothing written, nothing to roll back | Confirmed via the job's strict `success()` gate semantics (cross-checked against this repo's own `promote-latest` job, which relies on the same implicit behavior) |
| 6 | Release succeeds with the checkbox checked | PR opens against `main`, Slack notification fires to `log-sdk-libs` | Exact PR title/body and Slack message traced above |
| 7 | Same release re-run after a transient failure, or bump PR already merged | No duplicate PR, no duplicate commit (idempotency checks #1 and #2) | Traced through the exact `grep`/`gh pr list` dedupe logic |
| 8 | `VALUE = "..."` literal gets reformatted some day | Job fails loudly with a clear error instead of silently opening an empty, misleading PR | Locally simulated a reformatted file against the `sed` + verify logic — confirmed the mismatch is caught |
| 9 | Interaction with the existing `next` npm dist-tag publish job (`cicd_3-trunk.yml`) | No interference — `MinSdkVersion.java` falls under the `backend` change-detection filter, not `sdk_libs`, so merging the bump PR never triggers `publish-sdk-next` | Verified directly against `.github/filters.yaml` |

## Known limitations / accepted risk (not fixed here, flagged for awareness)

- **Two releases in flight at once**, both flagged as SDK-breaking, would open two separate bump
  PRs (different branch names, keyed by version) — no collision, but a human merging both should
  merge the later version last.
- **A stale, unmerged bump PR from a previous release** isn't caught by this run's own dedupe
  (different branch name). The validation step does emit a non-fatal `::warning::` if it finds an
  already-open `sdk/bump-min-sdk-version-*` PR against `main`, nudging the operator to merge it
  first.
- **Squash-merge dependency**: both this mechanism's PR-label lookup and the existing
  `release-qa-status` tool assume squash-merge commit messages (`... (#12345)`) to map commits back
  to PR numbers. If the repo's merge policy ever changes, both silently stop finding PR numbers —
  fails safe (never blocks a release), but a real breaking PR could then slip through undetected.
  Pre-existing repo-wide assumption, not new to this PR.
- **This has not yet been live-fire tested against a real GitHub Actions run in this repo** (only
  reasoned through carefully and unit-tested locally where possible — e.g. the `sed`/`grep`
  verification logic). A parallel attempt to validate this end-to-end in the disposable
  `core-workflow-test` fork hit an org-level branch-protection wall (a required status check that
  neither `--admin` merge nor a direct push could bypass with the available token permissions), so
  the AI-labeling behavior, the release-time gate, and the bump job's real GitHub Actions execution
  should be watched closely on this repo's first real usage.

## Files changed

- `.github/workflows/cicd_6-release.yml` — new `bump_min_sdk_version` input, new validation step
  in `verify-branch`, new `bump-min-sdk-version` job.
- `.github/workflows/ai_claude-sdk-breaking-change.yml` (new).
- `docs/core/SDK_BREAKING_CHANGE_CATEGORIES.md` (new).
- `dotCMS/src/main/java/com/dotcms/rest/config/MinSdkVersion.java` — Javadoc updated to describe
  the automated procedure.

## One-time setup already done

The `SDK Breaking Change` label has been created on this repo (color `#b60205`).

This PR fixes: #36698

### DIFF
--- a/.github/workflows/ai_claude-sdk-breaking-change.yml
+++ b/.github/workflows/ai_claude-sdk-breaking-change.yml
@@ -1,0 +1,119 @@
+name: Claude AI SDK Breaking Change Check
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+
+jobs:
+  # Security gate: Check if user is dotCMS organization member
+  #
+  # REQUIREMENTS FOR CLAUDE ACCESS:
+  # 1. Must be a member of the dotCMS organization
+  # 2. Membership must be set to PUBLIC visibility
+  #
+  # TROUBLESHOOTING: If blocked, visit https://github.com/orgs/dotCMS/people
+  # and ensure your membership is public (click "Make public" if needed)
+  security-check:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read           # Allow repository checkout
+      # Note: Organization membership checking uses fine-grained token
+      # so no additional GITHUB_TOKEN permissions needed for that API
+    outputs:
+      authorized: ${{ steps.membership-check.outputs.is_member }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Check organization membership
+        id: membership-check
+        uses: ./.github/actions/security/org-membership-check
+        with:
+          username: ${{ github.event.pull_request.user.login || github.actor }}
+
+      - name: Log security decision
+        run: |
+          if [ "${{ steps.membership-check.outputs.is_member }}" = "true" ]; then
+            echo "✅ Access granted: User is a dotCMS organization member"
+          else
+            echo "❌ Access denied: User failed dotCMS organization membership check"
+            echo ""
+            echo "📋 TROUBLESHOOTING: If you are a dotCMS team member:"
+            echo "   1. Visit https://github.com/orgs/dotCMS/people"
+            echo "   2. Ensure your membership is set to 'Public'"
+            echo "   3. If you're not listed, contact an organization owner"
+            echo ""
+            echo "::warning::Unauthorized user attempted to trigger Claude workflow: ${{ github.event.pull_request.user.login || github.actor }}"
+          fi
+
+  # SDK breaking-change analysis — runs on every PR push. Single label design: the AI only
+  # ever ADDS "SDK Breaking Change" when it detects a break; it never removes it. Removing
+  # the label (because a human disagrees with the AI's call, or because a later push fixed
+  # the issue) is always a manual human action — no separate "Human: ..." label needed, and
+  # no preflight step to clear/protect anything: a human's removal simply stands until the
+  # AI sees reason to re-add it on some future push.
+  claude-sdk-breaking-change-check:
+    needs: security-check
+    # Cancel in-progress check when a new push arrives — always analyze latest state
+    concurrency:
+      group: claude-sdk-breaking-${{ github.event.pull_request.number }}
+      cancel-in-progress: true
+    if: needs.security-check.outputs.authorized == 'true'
+    permissions:
+      contents: write
+      id-token: write
+      pull-requests: write
+      issues: write
+    uses: dotCMS/ai-workflows/.github/workflows/claude-orchestrator.yml@v3
+    with:
+      model_id: ${{ vars.BEDROCK_MODEL_ID }}
+      bedrock_role_arn: ${{ vars.BEDROCK_ROLE_ARN }}
+      trigger_mode: automatic
+      prompt: |
+        You are a dotCMS SDK-compatibility analyst. Determine whether the changes in this PR
+        break compatibility for `@dotcms/*` SDK consumers (`@dotcms/client`, `@dotcms/react`,
+        `@dotcms/angular`, `@dotcms/uve`) — i.e. whether an SDK version built against the
+        server contract *before* this change would misbehave against the server *after* this
+        change.
+
+        STEP 1 — Read the SDK breaking-change categories reference:
+          cat docs/core/SDK_BREAKING_CHANGE_CATEGORIES.md
+
+        STEP 2 — Get the full PR diff:
+          git diff ${{ github.event.pull_request.base.sha }}...${{ github.event.pull_request.head.sha }}
+
+        STEP 3 — Analyze the diff against EVERY category in the reference document.
+          Focus on: GraphQL schema surface reachable via the page API's graphql.page /
+          graphql.content extension (new required fields/args, removed/renamed types or
+          fields), REST response shape changes to /api/v1/page/*, /api/v1/content, and
+          /api/v1/nav, changes to the UVE/editor postMessage protocol (message names or
+          payload shapes in DotCMSUVEAction / __DOTCMS_UVE_EVENT__), and changes to the
+          SdkVersionWebInterceptor / X-DotCMS-Version / X-DotCMS-Min-SDK headers or the
+          compareVersions() comparison contract in sdk-compatibility.ts. Ignore pure admin-UI
+          (dotcms-ui) changes, test-only changes, or documentation changes unless they touch
+          one of the above surfaces.
+
+        STEP 4 — If the changes break SDK compatibility, post this comment on the PR
+        using: gh pr comment ${{ github.event.pull_request.number }} --body "..."
+
+          Format:
+          SDK Breaking Change Detected!!!
+          - Category: <category ID and name, e.g. "G-1 — Removing or Renaming a Reachable GraphQL Type/Field">
+          - Why it breaks compatibility: <specific explanation tied to the actual code changed>
+          - Code that makes it breaking: <file path(s) and the specific lines or block>
+          - Safer alternative (if possible): <the safer alternative from the reference, adapted to this change>
+
+          If multiple categories match, repeat the block for each one.
+
+          Then add the label (only if it isn't already on the PR):
+          gh pr edit ${{ github.event.pull_request.number }} --add-label "SDK Breaking Change"
+
+        If the changes do NOT break SDK compatibility, do nothing — no comment, no label, and
+        do NOT remove an existing "SDK Breaking Change" label even if you disagree with it.
+        Removing that label is a human-only decision.
+
+        Be specific: quote actual file names and code lines, not generic descriptions.
+      claude_args: '--allowedTools "Bash(git diff*),Bash(git log*),Bash(cat docs/core/SDK_BREAKING_CHANGE_CATEGORIES.md),Bash(gh pr comment*),Bash(gh pr edit*)"'
+      timeout_minutes: 15
+      runner: ubuntu-latest
+      enable_mention_detection: false
--- a/.github/workflows/cicd_6-release.yml
+++ b/.github/workflows/cicd_6-release.yml
@@ -333,6 +333,15 @@ jobs:
 
       - name: Bump MinSdkVersion.VALUE and open PR
         id: bump
+        # Notify-but-don't-fail: cicd_comp_finalize-phase.yml scans every job in the run via
+        # the GitHub API (not just finalize's own `needs`), so without this a transient
+        # git/gh error here — after the release has already fully shipped — would mark the
+        # whole release run red over a best-effort housekeeping PR. The failure Slack step
+        # below still fires (it checks the step's own outcome, not the job-level
+        # failure()/success() context functions, which continue-on-error would otherwise
+        # mask) — same pattern already applied to the SDK next-tag publish job in
+        # cicd_3-trunk.yml.
+        continue-on-error: true
         env:
           GH_TOKEN: ${{ secrets.CI_MACHINE_TOKEN || github.token }}
           CI_MACHINE_USER: ${{ secrets.CI_MACHINE_USER || 'github-actions[bot]' }}
@@ -411,7 +420,11 @@ jobs:
           slack-bot-token: ${{ secrets.SLACK_BOT_TOKEN }}
 
       - name: Slack Notification (bump PR failure)
-        if: failure()
+        # Checks the bump step's own outcome directly rather than the job-level
+        # failure()/success() context functions — those reflect conclusion (which
+        # continue-on-error masks to "success"), not outcome, so failure() would never
+        # fire here once that step has continue-on-error: true.
+        if: steps.bump.outcome == 'failure'
         continue-on-error: true
         uses: ./.github/actions/core-cicd/notification/notify-slack
         with:
--- a/.github/workflows/cicd_6-release.yml
+++ b/.github/workflows/cicd_6-release.yml
@@ -59,6 +59,11 @@ on:
         type: boolean
         default: true
         required: false
+      bump_min_sdk_version:
+        description: 'Does this release include a change that breaks SDK compatibility? If true, opens a post-release PR bumping MinSdkVersion.VALUE and notifies Slack for human review. Ignored for LTS releases.'
+        type: boolean
+        default: false
+        required: false
       java-version:
         description: 'Override Java version (SDKMAN format, e.g., 25.0.1-open). Compiler release auto-defaults to Java major version.'
         type: string
@@ -99,6 +104,79 @@ jobs:
             echo "::error::skip_latest must be blank (update the latest tag) or exactly 'DO NOT UPDATE LATEST' (skip it). Got: '${SKIP_LATEST}'. Re-run with a valid value."
             exit 1
           fi
+
+      # Closes the "forgot to check" gap for MinSdkVersion (see #36698): if a PR merged since
+      # the last standard release carries an SDK-breaking-change label but this run left
+      # bump_min_sdk_version false, fail loudly here rather than silently under-reporting the
+      # compatibility floor after the release ships. Read-only (no mutation), so it's safe to
+      # run before release-prepare even exists — deliberately plain `gh`/bash, not the
+      # release-qa-status TS tool, since we only need a label check, not QA-verdict aggregation.
+      - name: Validate bump_min_sdk_version against merged PR labels
+        env:
+          GH_TOKEN: ${{ secrets.CI_MACHINE_TOKEN || github.token }}
+          REPO: ${{ github.repository }}
+          RELEASE_VERSION: ${{ github.event.inputs.release_version }}
+          BUMP_MIN_SDK_VERSION: ${{ github.event.inputs.bump_min_sdk_version }}
+        run: |
+          set -euo pipefail
+
+          # Only standard (non-LTS) releases carry the -## suffix; LTS patch releases don't
+          # participate in the @latest SDK contract, so skip entirely.
+          if [[ ! "${RELEASE_VERSION}" =~ ^[0-9]{2}\.[0-9]{2}\.[0-9]{2}-[0-9]{1,2}$ ]]; then
+            echo "Non-standard release version '${RELEASE_VERSION}' (LTS or custom) — skipping SDK breaking-change validation."
+            exit 0
+          fi
+
+          THIS_TAG="v${RELEASE_VERSION}"
+
+          # Resolve the previous standard release tag the same way release-qa-status does
+          # (same STANDARD_RELEASE_PATTERN, kept in sync on purpose).
+          PREV_TAG=$(gh api "repos/${REPO}/releases?per_page=100" --paginate \
+            --jq '[.[] | select(.tag_name | test("^v[0-9]{2}\\.[0-9]{2}\\.[0-9]{2}-[0-9]{1,2}$"))] | .[].tag_name' \
+            | grep -v -F -- "${THIS_TAG}" | sort -V | tail -1 || true)
+
+          if [ -z "${PREV_TAG}" ]; then
+            echo "::warning::No previous standard release tag found — skipping SDK breaking-change validation (nothing to compare against)."
+            exit 0
+          fi
+
+          echo "Comparing ${PREV_TAG}...main for merged PRs since the last release."
+
+          # release-prepare hasn't created this release's tag yet at this point in the
+          # pipeline — main is the correct proxy for "what's about to ship" (release_commit
+          # defaults to main's tip the same way).
+          PR_NUMBERS=$(gh api "repos/${REPO}/compare/${PREV_TAG}...main" --paginate \
+            --jq '.commits[].commit.message' \
+            | grep -oE '\(#[0-9]+\)$' | grep -oE '[0-9]+' | sort -un || true)
+
+          BREAKING_PRS=()
+          for pr in ${PR_NUMBERS}; do
+            LABELS=$(gh pr view "${pr}" --repo "${REPO}" --json labels --jq '.labels[].name' 2>/dev/null || true)
+            if echo "${LABELS}" | grep -qxF 'SDK Breaking Change'; then
+              BREAKING_PRS+=("#${pr}")
+            fi
+          done
+
+          if [ "${#BREAKING_PRS[@]}" -gt 0 ] && [ "${BUMP_MIN_SDK_VERSION}" != "true" ]; then
+            echo "::error::This release (${THIS_TAG}) includes PR(s) labeled 'SDK Breaking Change' (${BREAKING_PRS[*]}) since ${PREV_TAG}, but 'bump_min_sdk_version' was left false. Re-run with bump_min_sdk_version=true, or remove the label from those PRs first if none of them actually break SDK compatibility."
+            exit 1
+          fi
+
+          if [ "${#BREAKING_PRS[@]}" -eq 0 ] && [ "${BUMP_MIN_SDK_VERSION}" = "true" ]; then
+            echo "::warning::bump_min_sdk_version=true but no merged PR since ${PREV_TAG} carries an SDK-breaking label. Proceeding, since a human explicitly opted in — but double-check this is intentional."
+          fi
+
+          # Non-fatal nudge: a stale, unmerged bump PR from a previous release won't be caught
+          # by this run's own dedupe (different branch name, keyed on this release's version).
+          # `gh pr list --head` only matches an exact branch name, not a prefix, so this has
+          # to go through the REST API directly and filter on head.ref with startswith().
+          STALE_PR=$(gh api "repos/${REPO}/pulls?state=open&base=main" --paginate \
+            --jq '.[] | select(.head.ref | startswith("sdk/bump-min-sdk-version-")) | .html_url' | head -1 || true)
+          if [ -n "${STALE_PR}" ]; then
+            echo "::warning::An unmerged MinSdkVersion bump PR from a previous release is still open: ${STALE_PR}. Consider merging it before this release ships another one."
+          fi
+
+          echo "SDK breaking-change validation passed (${#BREAKING_PRS[@]} breaking PR(s) found, bump_min_sdk_version=${BUMP_MIN_SDK_VERSION})."
 
   # Initialize - standard initialization phase (always first)
   initialize:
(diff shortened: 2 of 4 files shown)


KevinDavilaDotCMS:
Went through all 4 findings from the Claude Code review against the actual code — verified two empirically, one against direct precedent in this repo, and applied a real fix for the last one.

## Applied

**Finding 4 — bump-job failure would redden the whole release run — real, fixed**

Confirmed: `cicd_comp_finalize-phase.yml`'s `prepare-report-data` job scans **every** job in the run via the GitHub API (not just its own `needs:`), excluding only names matching `^(Finalize|Final Status|.*[Ff]inalize.*)`. `"Bump MinSdkVersion.java"` doesn't match that exclusion, so a transient `git`/`gh` error in that step — happening *after* the release has already fully shipped (build + deployment green) — would have marked the entire release run red over a best-effort housekeeping PR. This is the same class of issue already fixed for the SDK `next`-tag publish job in #36722.

Fix: added `continue-on-error: true` to the "Bump MinSdkVersion.VALUE and open PR" step, and changed the failure Slack notification's condition from `if: failure()` to `if: steps.bump.outcome == 'failure'` — `continue-on-error` masks the step's *conclusion* to `success` (which `failure()` checks), but `outcome` still reflects the real pre-mask result, so the failure notification still fires correctly.

## Reviewed, findings don't hold

**Finding 1 (partially) — 250-commit cap on the compare endpoint**

Tested this directly rather than take it on faith: ran the exact command our validation step uses (`gh api repos/dotCMS/core/compare/<sha>...main --paginate`) against a real range of 824 commits in this repo. It returned all 824 — `--paginate` does correctly page past the default 250-commit page size on this endpoint. The claim that "`--paginate` does not lift it" doesn't hold up empirically here.

The other half of that finding — that the PR-number extraction only matches squash-merge commit subjects (`(#N)`) and would miss merge-commit-merged PRs — is accurate, but it's an existing, already-documented limitation shared with the `release-qa-status` tool this repo already relies on for release QA reporting. Not new to this PR, and already called out as an accepted risk in the PR description.

**Finding 2 — shallow checkout could break `git diff base...head` in the AI prompt**

Checked this against direct precedent: `ai_claude-rollback-safety.yml` uses the byte-for-byte identical `git diff ${{ base.sha }}...${{ head.sha }}` construct, via the same `claude-orchestrator.yml@v3` reusable workflow. We've seen that exact mechanism successfully execute this diff and produce specific, code-referencing review comments on real PRs in this repo multiple times already. If checkout depth were insufficient, that already-shipped mechanism would already be failing. No changes made.

**Finding 3 — label doesn't exist**

Was accurate when the review ran, but already resolved — the `SDK Breaking Change` label was created on this repo before this PR was opened. Confirmed still present.

## Notes, agreed as-is

- The "unreachable nothing-to-commit branch" observation is technically correct (the preceding idempotency guard + the post-sed verification together guarantee a diff exists by that point) but harmless dead code — not changing it.
- Thanks for the callout on the doc quality and the H-1 self-referential note — that one was intentional.

dcolina:
@KevinDavilaDotCMS — two findings specific to the new logic (not caught in the prior review rounds). #1 can take down a legitimate release, so I'd fix it before this ships.

### 1. `verify-branch` can falsely block a legitimate release from a `(#N)` in a commit *body* — `cicd_6-release.yml:148-163`

```bash
PR_NUMBERS=$(gh api ".../compare/${PREV_TAG}...main" --paginate \
  --jq '.commits[].commit.message' \
  | grep -oE '\(#[0-9]+\)$' | grep -oE '[0-9]+' | sort -un || true)
```

`--jq '.commits[].commit.message'` emits the **full** message (subject + body), and `grep` runs per line. Squash-merge bodies routinely carry lines ending in `(#N)` — rolled-up bullets, revert/cherry-pick footers. Every one of those `N`s gets treated as a PR in this release's range.

**Failure case:** a commit body has a line ending in `(#450)`; `#450` is an **old, unrelated** PR that still carries the `SDK Breaking Change` label. `BREAKING_PRS` becomes non-empty → with `bump_min_sdk_version=false` → `exit 1`, **release blocked**. Worse, the error message then tells the operator to re-run with `bump_min_sdk_version=true`, triggering an **unwarranted floor bump**. The earlier review only flagged the false-*pass* direction (merge commits missing `(#N)`); this is the false-*block* direction.

**Fix** — match only the subject line:
```bash
--jq '.commits[].commit.message | split("\n")[0]'
```

### 2. The bump isn't monotonic — a re-run/backfill of an older release lowers the floor — `cicd_6-release.yml:~360-369`

The job only guards `CURRENT == RELEASE_VERSION`; it never checks `RELEASE_VERSION >= CURRENT`. If `main` is already at `VALUE = "25.10.20-1"` and someone backfills/re-runs an older `25.10.10-5` with `bump=true`, the `sed` rewrites `VALUE` to the **older** version, the NEW_VALUE check passes, and a PR is opened that **loosens** the compatibility contract — easy to merge without noticing.

`MinSdkVersion.VALUE` is a floor; it should only ever move forward. Suggest comparing with `sort -V` and no-op'ing (or failing loud) when `RELEASE_VERSION < CURRENT`.

dcolina (approved):
I'll approve it but it would be fine to fix the potential issues I've reported.

KevinDavilaDotCMS:
@dcolina Both real, both fixed — verified with concrete evidence from this repo's actual history before applying anything.

## 1. `(#N)` matches from the commit body, not just the subject

Confirmed this is not just theoretically possible — it's already happening in this repo. Scanned the last 3000 commits on `main`: **20 of them** have a body line (not the subject) ending in `(#N)`, e.g.:

```
commit c53feafbe | subject: fix(content-drive): field-based search follow-up fixes...
  body line: feature (issue #36384), on top of the merged frontend feature (#36452)
```

Your fix is exactly right. Applied:

```diff
- --jq '.commits[].commit.message' \
+ --jq '.commits[].commit.message | split("\n")[0]' \
```

Re-verified against the real compare range used in testing (825 commits) — this now extracts exactly one PR number per commit, from the subject only, with zero stray matches.

## 2. Non-monotonic bump (backfill could lower the floor)

Also confirmed — the code only guarded the exact-match no-op case, never checked direction. Added a monotonicity guard right after the existing idempotency check:

```bash
HIGHEST=$(printf '%s\n%s\n' "${CURRENT}" "${RELEASE_VERSION}" | sort -V | tail -1)
if [ "${HIGHEST}" != "${RELEASE_VERSION}" ]; then
  echo "::error::Refusing to bump MinSdkVersion.VALUE backward: main is already at '${CURRENT}', which is newer than this release's version '${RELEASE_VERSION}'. ..."
  exit 1
fi
```

Verified `sort -V` handles our date-lockstep version shape correctly (`26.7.14-1` vs `26.10.20-1` vs `26.7.14-2`, etc.) before relying on it. Since this lives inside the step that already has `continue-on-error: true` (from the earlier review round's fix), tripping this guard surfaces via the failure Slack notification without reddening the whole release run — the release already shipped by the time this step runs, so a refused bump is a "needs human attention" signal, not a release failure.

Both fixes pushed. Thanks for catching these — #2 especially, since a silently-loosened floor is exactly the kind of failure mode this whole mechanism exists to prevent.

KevinDavilaDotCMS:
Following up on the two remaining findings from the second Claude Code review round (PR review comment above dcolina's) — reviewed both, no code changes for either.

**`PREV_TAG` resolution assumes chronological order (`cicd_6-release.yml` — the `sort -V | tail -1` over standard release tags).** Confirmed the theoretical gap: if a release were ever cut out of chronological order (an older-dated backfill release after a higher-versioned one already shipped), the compare range would be wrong. No evidence this repo's release process ever cuts backfills out of order — every release we've observed moves strictly forward in time. Accepting as a known, low-probability limitation; not changing the resolution logic (which would otherwise need to switch to ordering by actual release creation date, like `release-qa-status`'s `findPreviousTag` does, rather than by version-string sort).

**`git push --force` runs before checking for an already-open PR (the bump job).** Confirmed: a retried run would force-reset the bot's branch from `origin/main` before checking if a PR already exists, which would clobber any commits a human manually pushed onto that branch in the meantime. This is intentionally the same convention already established by the retired `cicd_manual-release-sdks.yml`'s equivalent mechanism (its own comment: "these are bot-managed branches so force push is safe") — not a new risk introduced here, and consistent with how this kind of bot-owned branch is expected to be handled: review/merge the PR, don't push directly to the bot's branch. Leaving as-is.

Issue #33280, opened 2025-09-16
[EPIC] Select columns to show on table


MariaFernandez778:
### EPIC Title

[EPIC] Select columns to show on table

## Vision
User is able to select the columns to show in their content drive table. 

### Problem or opportunity
Currently the content drive displays a number of predefined and popular columns but the user may need other custom columns to display there instead. 

### Solution
Show in the context menu for the table the ability to pick which columns to show on the content drive table. 

### Out-of-Scope
Creating or defining new columns is out of scope. 

### Market Validation & Competitive Positioning
Few of the popular file sharing systems provide this functionality but more advanced and customizable CMS's and DAM.s do. We actually already provide this functionality in content search I think, or in some other table. 

### Target Personas

- [x] Developer teams
- [x] Content teams
- [ ] DevOps teams
- [ ] System administrators (dotCMS)

zJaaal:
Will revisit later, but we will work on the Show in List here https://github.com/dotCMS/core/issues/36736

PR #36737, opened 2026-07-27
feat(content-drive): dynamic Show In List table columns (#36736)

### LINKED ISSUE #36736: [TASK] Content Drive: dynamic table columns from content type "Show In List" fields
### Description

Part of epic #33280 (*Select columns to show on table*).

Make the Content Drive results table support **extra, dynamic columns** on top of its fixed set (name, status, locale, type, edited-by, last-edited, actions).

For the **Content Drive** context: when a single content type is selected in the filters, the extra columns are that content type's **"Show In List"** fields (`field.listed === true`), inserted right after the **Type** column. They follow the selected content type automatically — no manual picking. With multiple/no content type selected, only the fixed columns show.

> Note: this delivers the epic's goal via a different solution than the epic's original "context-menu picker" note — columns are derived from the content type's *Show In List* fields, reusing the field data we already fetch for **User Searchable**.

### Approach

**1. Splice a dynamic columns block into the table** — `libs/portlets/dot-content-drive/ui/src/lib/dot-folder-list-view/`
- **Only the Show In List columns are dynamic** — keep every fixed column exactly as it is today (header entries + the hand-written `<tbody>` cells with their custom rendering). No full table generalization; this keeps things simple.
- Add an `extraColumns` input and render it as a block inserted **immediately after the Type (`contentType`) column**, in both the header and the body:
  - Header: emit the extra column headers right after the Type `<th>`.
  - Body: right after the Type `<td>`, `@for` over the extra columns rendering the field value generically (e.g. `item[column.field]` with a default renderer).
  - Skeleton: same insertion so the loading row matches.
- Update the hardcoded `colspan` / `SKELETON_SPAN` to include the extra-column count. Widths: the extra columns take their own space after Type (fixed-layout percentages may need a light rebalance).

> Per the table's original design it's already set up to accommodate this — the change is a localized insertion after the Type column, not a rewrite.

**2. Derive Content Drive extra columns from "Show In List" fields**
- Reuse the field fetch built for **User Searchable**: single-selected-content-type detection (`$activeContentType`) + `DotContentTypeService.getContentType(variable)` — the response already returns every field, so no extra request is needed.
- Filter with `field.listed` (instead of the User Searchable `searchable && indexed` predicate) to get the Show In List fields.
- Expose them on the store (e.g. a `showInListFields` signal alongside `userSearchableFields`) and pass them to the table as `extraColumns`.

### Acceptance Criteria

- [ ] `dot-folder-list-view` accepts an `extraColumns` input and renders those columns after the **Type** column (header, body, skeleton all in sync).
- [ ] Fixed columns are unchanged — they keep their hardcoded custom rendering (thumbnail, status, locale, relative date, actions).
- [ ] `colspan` / skeleton span account for the extra columns (no hardcoded mismatch).
- [ ] In Content Drive, selecting a single content type adds that type's **Show In List** (`listed`) fields as extra columns.
- [ ] Selecting multiple content types or none shows only the fixed columns.
- [ ] The column set updates reactively when the selected content type changes (and clears with the filter).

### Technical Notes

- **Table:** `libs/portlets/dot-content-drive/ui/src/lib/dot-folder-list-view/` — columns array `HEADER_COLUMNS` and `DotFolderListViewColumn` in `.../ui/src/lib/shared/{constants,models}.ts`. Header/skeleton already loop the array; the `<tbody>` cells and `colspan` are the parts to touch.
- **Show In List flag:** `DotCMSContentTypeField.listed` — `libs/dotcms-models/src/lib/dot-content-types.model.ts`.
- **Field fetch to reuse:** `dot-content-drive-field-filter-menu.component.ts` (`$activeContentType`, `getContentType().fields`, per-type cache) + store signals in `dot-content-drive.store.ts` (`userSearchableFields`, `setUserSearchableFields`, `clearUserSearchableFilters`). Add a parallel Show-In-List path off the same fetch.

### Out of Scope

- Manual column selection / column-picker UI (the epic's original framing).
- Creating or defining new columns or fields.
- Persisting column preferences per user.
- Extra columns for contexts other than Content Drive (the table stays agnostic; only the Content Drive derivation is in scope here).

### Additional Context

Epic: #33280

zJaaal:
## Proposed Changes

Implements #36736 (under epic #33280): the Content Drive results table shows extra columns after the **Type** column, derived from the selected single content type's **"Show In List"** (`field.listed`) fields. Frontend only.

- **Agnostic table** (`dot-folder-list-view`): new `extraColumns` input, spliced in right after the Type column in the header, body, and skeleton, with a dynamic `colspan`. Fixed columns are untouched. Columns are **de-duplicated by field key** (so a `title` field can't double the name column). The table **owns sizing**: an explicit width wins, otherwise text/number columns size to the **average value length** of the current rows (in `ch`, header-floored, clamped 8–32) and other types get a fixed per-type width; the widths sum into the table's `min-width` for **horizontal scroll** instead of squeezing. Column display types are an `as const` (`DOT_FOLDER_LIST_VIEW_COLUMN_TYPE`) — no magic strings.
- **Type-aware cells**: `date` → `MM/dd/yyyy`, `datetime` → `MM/dd/yyyy h:mm a`, `time` → `h:mm a` (Angular `DatePipe`, **absolute** on purpose — not the relative `modDate` pipe); `boolean` → Material `check_circle` / `cancel`, and **blank when the value is absent** (a missing boolean doesn't read as `false`); `image` → a **thumbnail of the field's own asset** (see below); folders render blank.
- **Image / Binary / File columns**: render a `dot-content-thumbnail` pointed at the field via `[options]="{ fieldVariable }"`, so a secondary/gallery image shows **its own** picture rather than the contentlet's title image (non-image assets fall back to a file-type icon). Same size/styling as the row's leading title thumbnail (fixed `4.5rem × 3.5rem` box, no click). Fixed column width — never measured from content.
- **Sortable columns**: a Show-In-List column is sortable when its field is `indexed` (the `(sort)` → `store.setSort` → `sortBy` pipeline was already wired). The sortable header keeps its label + sort icon on one line.
- **Reuses the User Searchable fetch**: one `getContentType(variable)` response now feeds both the searchable-field chips and the new `showInListFields` store signal (filtered by `field.listed`). Cleared when the single-content-type selection changes/clears.
- **Shell**: maps each field's `fieldType`/`dataType` to the generic column type using the canonical `DotCMSFieldTypes` enum — Image/Binary/File → `image`, Date-and-Time/Time resolved from `fieldType` (they share `dataType` DATE), the rest from `dataType`.

## Review follow-ups (in this PR)

Addressing @nicobytes' review:
- Column typing **decoupled** from the field-filter constants → canonical `DotCMSFieldTypes` enum (no behavior change).
- Date/time columns **documented** as intentionally absolute (vs. the relative `modDate` pipe).
- **Sortable** Show-In-List columns (was hardcoded `sortable:false`).
- **Image support** — Image/Binary/File fields showed raw text; they now render the field's own thumbnail.
- Header sort-icon wrap fixed (`whitespace-nowrap`); stale/duplicated JSDoc removed.

**Deferred (not here):** convergence of the relationship field's tables onto the Content Drive table (as a constrained subset) — tracked in **#36749** (near-term: add single/multi-select to the Content Drive table); a dev-warning for `$safeExtraColumns` when it drops a field-key collision.

## Testing

- `content-drive-ui`: 186 passing · `portlets-content-drive`: 810 passing. Lint clean.
- New unit tests: table (render/order after Type, dedupe by key, boolean true/false/absent, date vs datetime, `ch` vs fixed width, clamp, no-extras, **image thumbnail cell**); store (`setShowInListFields` set/clear); field-filter menu (populate from `listed`, clear on deselect); shell (field→type mapping, **Image/Binary/File → image**, **sortable-when-indexed**).

## Notes

- The extra-column mechanism is agnostic; only the Content Drive derivation is wired here.
- `item[field]` renders the raw value for text/number; rich rendering of complex field types (relationship/block) is out of scope.

Closes #36736

This PR fixes: #36736

### DIFF
--- a/core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-field-filter-menu/dot-content-drive-field-filter-menu.component.spec.ts
+++ b/core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-field-filter-menu/dot-content-drive-field-filter-menu.component.spec.ts
@@ -62,6 +62,7 @@ describe('DotContentDriveFieldFilterMenuComponent', () => {
                 userSearchableActive: signal<string[]>([]),
                 addUserSearchableField: jest.fn(),
                 setUserSearchableFields: jest.fn(),
+                setShowInListFields: jest.fn(),
                 clearUserSearchableFilters: jest.fn()
             }),
             mockProvider(DotHttpErrorManagerService),
@@ -180,5 +181,40 @@ describe('DotContentDriveFieldFilterMenuComponent', () => {
 
         expect(httpErrorManager.handle).toHaveBeenCalled();
         expect(store.setUserSearchableFields).toHaveBeenCalledWith([]);
+    });
+
+    it('should populate the Show In List fields from the content type listed fields', () => {
+        contentTypeService.getContentType.mockReturnValue(
+            of({
+                id: 'blog',
+                fields: [
+                    field({ variable: 'title', name: 'Title' }), // not listed
+                    field({ variable: 'summary', name: 'Summary', listed: true }),
+                    field({ variable: 'author', name: 'Author', listed: true })
+                ]
+            } as DotCMSContentType)
+        );
+        store.getFilterValue.mockReturnValue(['blog']);
+
+        spectator.detectChanges();
+
+        expect(store.setShowInListFields).toHaveBeenCalledWith([
+            expect.objectContaining({ variable: 'summary' }),
+            expect.objectContaining({ variable: 'author' })
+        ]);
+    });
+
+    it('should clear the Show In List fields when the single content-type selection is removed', () => {
+        const contentType = signal<string[] | undefined>(['blog']);
+        store.getFilterValue.mockImplementation((key: string) =>
+            key === 'contentType' ? contentType() : undefined
+        );
+
+        spectator.detectChanges();
+
+        contentType.set(undefined);
+        spectator.detectChanges();
+
+        expect(store.setShowInListFields).toHaveBeenLastCalledWith([]);
     });
 });
--- a/core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-field-filter-menu/dot-content-drive-field-filter-menu.component.ts
+++ b/core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-field-filter-menu/dot-content-drive-field-filter-menu.component.ts
@@ -66,7 +66,7 @@ export class DotContentDriveFieldFilterMenuComponent {
 
     protected readonly $loading = signal(false);
 
-    /** Monotonic cache of eligible fields per content-type variable, to avoid refetching. */
+    /** Monotonic cache of the raw fields per content-type variable, to avoid refetching. */
     readonly #fieldsCache = signal<Record<string, DotCMSContentTypeField[]>>({});
 
     /** The previously-active single content-type variable, used to detect real changes. */
@@ -109,13 +109,13 @@ export class DotContentDriveFieldFilterMenuComponent {
                     return this.#contentTypeService.getContentType(contentTypeVar).pipe(
                         take(1),
                         map((contentType: DotCMSContentType) => {
-                            const eligible = this.#eligibleFields(contentType?.fields ?? []);
+                            const raw = contentType?.fields ?? [];
                             this.#fieldsCache.update((cache) => ({
                                 ...cache,
-                                [contentTypeVar]: eligible
+                                [contentTypeVar]: raw
                             }));
 
-                            return eligible;
+                            return raw;
                         }),
                         catchError((error) => {
                             this.#httpErrorManager.handle(error);
@@ -127,7 +127,10 @@ export class DotContentDriveFieldFilterMenuComponent {
                 takeUntilDestroyed(this.#destroyRef)
             )
             .subscribe((fields) => {
-                this.#store.setUserSearchableFields(fields);
+                // One fetch feeds two consumers: the field-filter chips (searchable + indexed
+                // fields) and the results table's extra columns ("Show In List" fields).
+                this.#store.setUserSearchableFields(this.#eligibleFields(fields));
+                this.#store.setShowInListFields(fields.filter((field) => field.listed));
                 this.$loading.set(false);
             });
 
@@ -158,6 +161,7 @@ export class DotContentDriveFieldFilterMenuComponent {
 
         if (active === null) {
             this.#store.setUserSearchableFields([]);
+            this.#store.setShowInListFields([]);
 
             return;
         }
--- a/core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.html
+++ b/core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.html
@@ -37,6 +37,7 @@
         [items]="$items()"
         [loading]="$loading()"
         [totalItems]="$totalItems()"
+        [extraColumns]="$extraColumns()"
         [offset]="$offset()"
         (paginate)="onPaginate($event)"
         (sort)="onSort($event)"
--- a/core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.spec.ts
+++ b/core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.spec.ts
@@ -2029,5 +2029,20 @@ describe('DotContentDriveShellComponent', () => {
                 expect.objectContaining({ field: 'noidx', sortable: false })
             ]);
         });
+
+        it('should map Image/Binary/File fields to the image (thumbnail) column type', () => {
+            showInListFieldsSignal.set([
+                { variable: 'photo', name: 'Photo', dataType: 'SYSTEM', fieldType: 'Image' },
+                { variable: 'doc', name: 'Doc', dataType: 'SYSTEM', fieldType: 'Binary' },
+                { variable: 'attach', name: 'Attach', dataType: 'SYSTEM', fieldType: 'File' }
+            ] as DotCMSContentTypeField[]);
+            spectator.detectChanges();
+
+            expect(spectator.component.$extraColumns()).toEqual([
+                expect.objectContaining({ field: 'photo', type: 'image' }),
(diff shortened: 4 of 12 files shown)


nicobytes:
## Review: this duplicates the relationship field's dynamic columns

Not blocking — the feature works. But "content type fields → table columns" is now implemented twice, independently, and the two will drift.

### The overlap

| | Relationship field (`libs/edit-content/.../dot-edit-content-relationship-field`) | This PR (content-drive) |
|---|---|---|
| Column source | `showFields` field variable | `field.listed` (Show In List) |
| Column model | `TableColumn {nameField, header, type}` — `models/relationship.models.ts:79` | `DotFolderListViewColumn {field, header, width?, order, type?}` — `content-drive/ui/shared/models.ts` |
| Type vocabulary | `text \| title \| language \| status \| image` | `text \| number \| boolean \| date \| datetime \| time` |
| Field → type | `getTypeField()`, by `clazz` — `utils/index.ts:118` | `#columnTypeForField()`, by `fieldType` + `dataType` — `dot-content-drive-shell.component.ts` |
| Cell rendering | `@switch column.type` — `dot-relationship-field.component.html:28` (header) + `:126` (body) | `@switch column.type` — `dot-folder-list-view.component.html:374-407` |

(The relationship field also carries two column models of its own: `TableColumn` and `Column {field, header}` in `models/column.model.ts`.)

### Concrete unification — follow-up PR

Target is `libs/ui`: `content-drive/ui` already imports `@dotcms/ui` (`dot-tree-folder.component.ts:18`), and going the other way would trip `import/no-cycle`.

**1. `DotContentTableColumn` in `@dotcms/ui`**, splitting the two axes this PR conflates in a single `type`:

```ts
type: 'text' | 'number' | 'boolean' | 'date' | 'datetime' | 'time';   // value formatting  (this PR)
renderer?: 'title' | 'status' | 'language' | 'thumbnail';             // contentlet cells   (relationship)
```

The relationship field's `title`/`status`/`language`/`image` aren't data types, they're components — that's why the two vocabularies can't merge as-is. Split them and both tables can consume the same model: the relationship field gains date/number/boolean formatting (today everything non-special falls through to `'text'`, so a date renders as a raw timestamp), content-drive can gain thumbnail/status later.

**2. Move the field → column-type mapping into a shared pure util.** `(field: DotCMSContentTypeField) => DotContentTableColumnType`, carrying the union of both rule sets: image/binary by `clazz` (relationship) + date sub-type by `fieldType` + `dataType` fallback (this PR).

Worth fixing while doing it: this PR pulls `FIELD_FILTER_DATE_TIME_TYPE` / `FIELD_FILTER_TIME_ONLY_TYPE` from the **field-filter** constants to resolve *columns*. It works, but it couples two unrelated concerns — those values should come from a field-type enum.

**3. `<dot-content-cell [contentlet]="item" [column]="column" />` in `libs/ui`.** The `@switch` over column type is written twice today. Each table keeps its own chrome (checkbox, actions, frozen columns, drag & drop) and only drops the shared cell in for dynamic columns.

I would **not** merge the two tables themselves — different selection, pagination, actions, drag & drop. Only the column model, the mapping and the cell.

### Notes on this PR itself

- **Two date conventions in the same row.** The extra columns use `DatePipe` with hardcoded `'MM/dd/yyyy'` / `'MM/dd/yyyy h:mm a'`, while `modDate` right next to them uses `| dotRelativeDate` (`dot-folder-list-view.component.html:135`). The literal formats also ignore locale; `dotRelativeDate` already takes a format argument and defaults to `'MM/dd/yyyy'`.
- **Duplicated JSDoc** on `#columnTypeForField`: two consecutive `/** */` blocks, and the first is stale — it still describes the width logic that moved into the table.
- **Mixed width units.** `HEADER_COLUMNS` uses percentages, extra columns use `ch`/`rem`, and they combine into `min-width: calc(100% + 32ch + 12rem)` under `table-layout: fixed`. The percentages resolve against the already-grown table, so the fixed columns shrink as extra columns are added. Worth eyeballing with 5-6 extra columns.
- **Width from the average value length of the current page** means a text column resizes when you paginate. Documented as deliberate, but it does jump.
- **`$safeExtraColumns` drops colliding columns silently.** A content type with a `title` field marked Show In List produces a configured column that simply never appears, with no feedback.
- **`sortable: false` on every extra column** — the backend supports sorting by field. Good follow-up.

### Order

1. Land this as-is.
2. Refactor PR: column model + mapping util + `dot-content-cell` into `@dotcms/ui`, migrate content-drive and the relationship field onto them.

zJaaal:
Thanks @nicobytes — agree with the sequencing (land this, unify in a follow-up). We're **not** merging the two tables or the column model in this PR: as you note, the two `type` vocabularies are really two axes (data-formatting `text|number|boolean|date…` vs. component-renderers `title|status|language|image`), so collapsing them now would risk regressions in **both** the relationship field and content-drive. That's the `DotContentTableColumn` (type + renderer) + shared mapping util + `<dot-content-cell>` refactor — worth its own PR in `@dotcms/ui`, migrating both consumers together.

Took the safe, no-regression bits now (pushed `8fb557eaab`):

- ✅ **Decoupled column typing from the field-filter constants** — `#columnTypeForField` now reads the canonical `DotCMSFieldTypes.DATE_AND_TIME` / `.TIME` enum instead of `FIELD_FILTER_*`. Same string values, zero behavior change — just removes the coupling to an unrelated feature you flagged.
- ✅ **Documented the date convention as intentional** — the extra date/time columns render **absolute** values on purpose; `dotRelativeDate` (used for `modDate`) shows relative text ("3 days ago") that's right for an audit timestamp but would misrepresent an arbitrary content date field. Added a comment so it doesn't read as an oversight. (Left the format absolute rather than switching to the relative pipe.)
- ✅ **Removed the stale/duplicated JSDoc** on `#columnTypeForField`.

Acknowledged for the follow-up (not touched here): the shared model/util/cell, `sortable` on extra columns (backend supports it), and `$safeExtraColumns` dropping a colliding column silently. The mixed width-unit / paginate-resize notes are the deliberate "width from current page" behavior.

_(Posted by Claude Code on @zJaaal's behalf.)_

zJaaal:
Update — the flagged in-PR items are addressed (latest: `728ccf9983`):

- ✅ **Field-type coupling** — column typing now uses the canonical `DotCMSFieldTypes` enum instead of the field-filter constants.
- ✅ **Date convention** — documented that date/time columns render absolute values on purpose (vs. `modDate`'s relative pipe).
- ✅ **Stale/duplicated JSDoc** on `#columnTypeForField` removed.
- ✅ **Sortable columns** — Show-In-List columns are now sortable when the field is `indexed` (the sort pipeline was already wired; only the hardcoded `sortable:false` blocked it). +test.
- ✅ **Image support** — Image/Binary/File fields rendered raw text before; they now render a **thumbnail of the field's own asset** (`dot-content-thumbnail` + `fieldVariable`), matching the row's title-thumbnail size/styling (no click). Fixed column width, header icon stays inline. +tests.

**Deferred (not in this PR):**
- `$safeExtraColumns` silently dropping a column whose field collides with a fixed column — small dev-warning follow-up.

_(Posted by Claude Code on @zJaaal's behalf.)_

rjvelazco (approved):

Issue #36750, opened 2026-07-27
Backfill dev.dotcms.com changelog entries for releases since 26.07.13-01 (pre-automation gap)


sfreudenthaler:
## Summary

The dev.dotcms.com site changelog (the `Dotcmsbuilds` content type on the corpsites headless backend) was hand-maintained by Jamie Mauro through version **26.07.13-01** (last manual entry, released 2026-07-13). Changelog automation merged 2026-07-21 (#36606), but its trigger wiring was missing until #36743 (see #36742), so GA releases published in the gap between the last manual entry and the trigger fix have **no site changelog entry**.

This issue tracks the one-time operator backfill of those missing entries.

## Confirmed missing versions

Verified read-only against the corpsites headless backend (`+contentType:Dotcmsbuilds +Dotcmsbuilds.minor_dotraw:<version>` returns 0 hits with the admin token):

| Version | GA release date | Site entry exists? |
|---|---|---|
| 26.07.06-3 | 2026-07-15 | ❌ missing |
| 26.07.17-01 | 2026-07-17 | ❌ missing |
| 26.07.20-01 | 2026-07-20 | ❌ missing |
| 26.07.21-01 | 2026-07-21 | ❌ missing |
| 26.07.27-01 | 2026-07-27 | ❌ missing |

Control checks (confirming the query and the boundary are correct): **26.07.13-01** (last manual entry) is present; **26.07.06-01** and **26.07.06-02** are present. Because the 26.07.06 line otherwise has entries, **26.07.06-3** — an older-line patch published 2026-07-15, i.e. after the manual process ended — is a genuine gap, not an intentionally-skipped patch. CLI (`dotcms-cli-*`) and LTS (`*_lts_*`) releases are out of scope.

## Why the gap exists

- The manual authoring process ended at 26.07.13-01.
- Automation merged in #36606 (2026-07-21) but the trigger that fires it on release was not wired up until #36743 (tracked by #36742).
- Releases published in between therefore never triggered an automated publish and were never hand-entered.

## Backfill method

Per spec **FR-012**, backfill is an **explicit, per-version operator action** — not an automatic catch-up. For each missing version, run one deliberate invocation of the changelog publisher:

```
changelog-publisher publish \
  --version <v> \
  --notes-file <release body> \
  --docker-image dotcms/dotcms:<v>_<sha> \
  --released-date <actual release date> \
  --apply
```

Notes:
[...]

PR #36759, opened 2026-07-27
fix(changelog-publisher): send required eolDate field (create path 400s without it)

### LINKED ISSUE #36758: changelog-publisher omits required eolDate field — fire 400s on create
Discovered during the #36750 backfill (the first live fires of the publisher): the `Dotcmsbuilds` type requires `eolDate` ("Planned End of Life Date"), which the tool's payload omitted — every create fails with a 400 validation error, so the maiden release run would have failed the same way. The field wasn't flagged as required by the original field inspection.

Fix: payload now sends `eolDate`; CLI gains `--eol-date`, defaulting to released date + 1 year (the convention every current-track entry on the site follows — verified against live entries). Backfill of 26.07.06-3 / 26.07.17-01 / 26.07.21-01 / 26.07.27-01 succeeded with the patch.

Also for the ops runbook, discovered in the same session: the service account additionally needs (a) an explicit entry in the System Workflow Publish action's *Who can use* list, and (b) type-row View+Edit+Publish on `dotCMS Builds` (create-path checks structure-level write), beyond the children-row grants.

sfreudenthaler:
First live fires of the changelog publisher (the #36750 backfill) hit a 400: `Dotcmsbuilds` requires `eolDate`, which the payload omitted — the maiden release run would have failed identically. Payload now includes it; CLI gains `--eol-date` defaulting to released date + 1 year (the convention every current-track entry follows, verified live). 31 tests green; the four-version backfill ran successfully on this exact code.

Closes: #36758

https://claude.ai/code/session_012ogYXVeZsBVUHkwzdStYPJ

This PR fixes: #36758

### DIFF
--- a/.github/actions/core-cicd/changelog-publisher/src/changelog_publisher/cli.py
+++ b/.github/actions/core-cicd/changelog-publisher/src/changelog_publisher/cli.py
@@ -49,6 +49,11 @@ def cmd_publish(args: argparse.Namespace) -> int:
         return 2
     # Default per site convention: EOL = released date + 1 year (current track).
     if args.eol_date:
+        try:
+            dt.date.fromisoformat(args.eol_date)
+        except ValueError:
+            log.error("invalid --eol-date %r: expected yyyy-MM-dd", args.eol_date)
+            return 2
         eol_date = args.eol_date
     else:
         try:
--- a/.github/actions/core-cicd/changelog-publisher/src/changelog_publisher/cli.py
+++ b/.github/actions/core-cicd/changelog-publisher/src/changelog_publisher/cli.py
@@ -40,6 +40,15 @@ def cmd_publish(args: argparse.Namespace) -> int:
     with open(args.notes_file, encoding="utf-8") as fh:
         release_notes = fh.read()
     released_date = args.released_date or dt.date.today().isoformat()
+    # Default per site convention: EOL = released date + 1 year (current track).
+    if args.eol_date:
+        eol_date = args.eol_date
+    else:
+        r = dt.date.fromisoformat(released_date)
+        try:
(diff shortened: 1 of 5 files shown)


sfreudenthaler:
Review finding addressed in `c6c2ebe8c8` (+ test-name tidy): malformed `--released-date` is now validated up front — clean one-line error, exit 2 (usage/validation bucket per the exit contract), zero network calls, no traceback. New test `test_malformed_released_date_exits_two_with_clean_error` covers it; suite at 32 green. (For the record: the release workflow always passes `$(date -u +%F)`, so this path was manual-invocation-only — but the clean-error contract should hold everywhere.)

sfreudenthaler:
Both second-pass findings addressed in `ddafbef8aa` (33 tests green):

1. **Stacked decorator** — fixed, and it was actually worse than flagged: my earlier insertion had *stolen* `test_missing_backend_url`'s decorator (leaving it undecorated) while double-decorating the new test. Both now have exactly one.
2. **`--eol-date` validation symmetry** — a user-supplied `--eol-date` now gets the same up-front `fromisoformat` check: clean one-line error, rc 2, zero network calls, with a matching test.

sfreudenthaler:
Third-pass finding addressed (35 tests green): the default eolDate derivation — the branch production actually runs — now has value-pinning tests: `test_default_eol_date_is_released_plus_one_year` asserts the fired payload carries releasedDate + 1 year exactly, and `test_default_eol_date_handles_feb_29` covers the 2028-02-29 → 2029-02-28 fallback. Good catch — a wrong/empty default would previously have passed the suite.

dcolina (approved):
Go for it.

PR #36778, opened 2026-07-28
fix(content-drive): date/time filter timezone coherence + typeable Time input

### LINKED ISSUE #36383: [TASK] Content Drive: field-based search/filter criteria reusing edit-content field components

zJaaal:
## What & why

A set of **frontend-only** Content Drive fixes. The core of the PR is timezone coherence for the **Date / Date-and-Time / Time** field filters (the filter now sends what the user sees instead of a browser-shifted instant, and the table renders the same server-time value). While in the area, it also folds in a typeable Time input, clearer range-validation copy, and a small folder-dialog preview regression fix.

### 1. Timezone coherence — filter + table (the main fix)
- **Filter serialization** now sends the **wall-clock** the user picked, not a UTC instant. `toISOString()` shifted by the browser offset (a UTC-3 user's `10:00` went out as `13:00Z`); the backend then parsed that `Z` as an instant and reformatted it in the **server** zone, so the bound no longer matched the pick. Serialization now emits a no-offset `yyyy-MM-ddTHH:mm:ss` string that round-trips as identity against `BrowserAPIImpl#parseFlexibleDate` / `normalizeDateBound`.
- **Table display** (`dot-folder-list-view`) formats date/datetime/time content columns in **UTC** (`| date: … : 'UTC'`). dotCMS stores these fields against the server zone and the editor shows them there; formatting in the browser zone shifted a `09:00` value to `12:00` for a UTC+3 viewer, so the table disagreed with both the editor and the filter. Now editor == index == table == filter.

### 2. `toLocalIsoString` — reuse `date-fns` + guard Invalid Date *(from Claude review)*
- Replaced the hand-rolled `getTimezoneOffset()` math with `date-fns` `format(date, "yyyy-MM-dd'T'HH:mm:ss")` (already a repo dependency), which produces the byte-identical wall-clock string.
- Guards `!date || Number.isNaN(date.getTime())` → returns `''`. The new typeable Time input (`[keepInvalid]="true"`) can emit an `Invalid Date` mid-typing, and both the old impl and `date-fns` `format` throw `RangeError` on one; now a partial time simply clears that bound.

### 3. UTC list-column formatting test *(from Claude review)*
- Added a spec that runs under a **non-UTC** zone (`America/New_York`) with a **near-midnight-UTC** instant, so the day *and* hour differ from UTC, and asserts the exact rendered text for date/datetime/time columns. Verified sensitive: dropping `:'UTC'` makes it fail — unlike the prior `contains ':'` check, which was trivially green in Jest's UTC zone.

### 4. Typeable Time filter input
- The Time-only filter's From/To pickers are now typeable (input + clock overlay, `[keepInvalid]="true"`) instead of click-only spinners; no "Today" button. Date-and-Time keeps its original inline calendar.

### 5. Clearer range-validation copy
- `content-drive.field-filter.time.invalid-range` → `'From' must be earlier than 'To'` (matches existing codebase phrasing).

### 6. Folder create form previews the currently opened folder
- Opening **New folder** inside a subfolder showed the **site root** in the read-only Path preview instead of the current folder. `#getAssetPath` dropped `store.path()` on the empty-name branch — a regression from #34081, where an empty-name guard added to strip a dangling `//` over-corrected and also discarded the path. The branch now keeps `store.path()` (empty-slug `//` still avoided).
- **Loop-safe:** this is a pure `computed` (`$finalPath`) reading `store.path()` — no form-control write, so it can't reintroduce the `title→name` effect loop that #34081 removed. Does not affect where folders are actually created (creation always used `store.path()` via the non-empty-name branch).

## Testing

https://github.com/user-attachments/assets/a0572bc8-88ad-40b8-8ff1-c26cbb6a6060

- `pnpm nx test portlets-content-drive` — **835 pass** (incl. new folder-preview, invalid-date, and typeable-time cases).
- `pnpm nx test content-drive-ui` — **113 pass** (incl. the new UTC list-column test).
- Lint + format clean.

## Notes / follow-ups (out of scope, FE-only)
- UTC table formatting hardcodes the server zone as UTC (correct for the default dotCMS server). Sourcing the zone from `DotSystemConfigService` would generalize a non-UTC server — tracked as a follow-up.
- The folder dialog derives its parent from `store.path()` only, not `store.selectedNode()`. After a **global search** (which resets `path` to root but leaves a node selected) a new folder would preview/create at root. Left as a separate item pending a decision on whether to add a `selectedNode` fallback (as the content-type-selector dialog does).

Fixes #36383

This PR fixes: #36383


rjvelazco (approved):

dario-daza (approved):

zJaaal:
Addressed the Medium finding on `toLocalIsoString` in 054ba237cf.

Two things in one change:
- **Invalid Date guard** — confirmed your read: `[keepInvalid]="true"` on the typeable Time input can emit an `Invalid Date` mid-typing, and `$timeRangeInvalid()` doesn't catch it (`NaN > x` is `false`), so `#applyRange()` reached the serializer. `toLocalIsoString` now returns `''` for `!date || Number.isNaN(date.getTime())`, so a partial time just clears that bound instead of throwing.
- **De-duplicated the impl** — while here, replaced the hand-rolled `getTimezoneOffset()` math with `date-fns` `format(date, "yyyy-MM-dd'T'HH:mm:ss")` (already a repo dependency, used in analytics / `dot-format-date`). Byte-identical output, and it made the guard necessary anyway since `date-fns` `format` also throws `RangeError` on an invalid instant. Added a unit test for the invalid-date path.

On the non-blocking UTC note: the non-UTC-server generalization (sourcing the zone from `DotSystemConfigService`) is captured as a follow-up — leaving it out of this PR to keep it FE-only and scoped.

— posted by Claude on @zJaaal's behalf

zJaaal:
Addressed the new Medium (UTC-formatting coverage) in 365dfbdfc0.

You were right that the old assertion (`datetime contains ':'`, `date doesn't`) was trivially green under Jest's UTC zone. New test:
- Runs under a **non-UTC zone** (`America/New_York`, UTC-5 in Jan) and seeds a **near-midnight-UTC** instant (`2026-01-15 02:30 UTC` → `2026-01-14 21:30 EST`) so both the calendar day **and** the hour differ from UTC.
- Asserts the exact rendered text for date / datetime / time columns (`01/15/2026`, `01/15/2026 2:30 AM`, `2:30 AM`).

Verified it's actually sensitive: temporarily dropping `:'UTC'` makes it render `01/14/2026 11:30 PM` and fail — so a future regression removing the arg is caught.

Non-UTC-server generalization (zone from `DotSystemConfigService`) still tracked as a follow-up, out of this FE-only PR.

— posted by Claude on @zJaaal's behalf

PR #36781, opened 2026-07-28
feat(dev-tools): persist last query + unify theming across dev-tool portlets (#36692)

### LINKED ISSUE #36692: Dev tools consistency: remove dark mode from velocity-playground and persist last query across query-tool, es-search, and velocity-playground
### Description

Two consistency gaps across the migrated dev-tool portlets — `dot-query-tool`, `dot-es-search`, and `dot-velocity-playground` — that should be addressed together so the three portlets behave uniformly.

**1. Remove dark mode from `dot-velocity-playground`**

Only `dot-velocity-playground` ships a dark theme toggle; the other two dev tools do not. This creates an inconsistent UX across otherwise-sibling portlets. Remove the dark mode from `dot-velocity-playground` so all three follow the platform's default theming.

**2. Persist the last query per portlet — homogeneous across the three**

Users lose their in-progress work when they leave and return to any of these portlets. Dev-tool queries (Lucene, ES JSON, VTL) are complex and time-consuming to recompose, so the last query should be restored when the user comes back to the portlet.

Deliver a **single, shared implementation** used by all three portlets, backed by `localStorage`. `dot-velocity-playground` already persists splitter ratio and history via `localStorage` (`readJson` / `writeJson` utilities in `dot-velocity-playground.utils.ts`) — extract that pattern into a shared location and reuse it across the three portlets.

**Suggested approach**

- Promote the persistence utilities (or wrap them as a `withPersistedQuery()` SignalStore feature) into `libs/data-access/` or `libs/ui/` so all three portlets consume the same helper.
- Key naming convention: `dotcms.devtools.{portlet}.lastQuery` (e.g. `dotcms.devtools.query-tool.lastQuery`).
- Debounce writes (~300 ms) while the user is typing.
- Load and restore the last query on portlet init.
- Provide an explicit "Clear" action per portlet that removes the stored value.
- Per-portlet payloads:
  - `dot-query-tool`: Lucene query string (+ minimal filters if any).
  - `dot-es-search`: ES query JSON.
  - `dot-velocity-playground`: VTL body — reconcile with existing history/splitter storage, do not duplicate.

### Acceptance Criteria

- [ ] Dark mode / dark theme toggle is removed from `dot-velocity-playground` (UI, config, and any related code paths).
- [ ] All three portlets (`dot-query-tool`, `dot-es-search`, `dot-velocity-playground`) render with the same default theming; no portlet-specific theme switch remains.
- [ ] A shared persistence helper exists in a shared lib (`libs/data-access/` or `libs/ui/`) — not duplicated per portlet.
- [ ] Each of the three portlets restores the last query on load from `localStorage`.
- [ ] Each portlet saves the current query to `localStorage` on change (debounced ~300 ms).
- [ ] Each portlet exposes a "Clear" action that removes the stored value.
- [ ] Storage keys follow the convention `dotcms.devtools.{portlet}.lastQuery`.
- [ ] `dot-velocity-playground` existing `localStorage` usage (history, splitter ratio) is reconciled with the new helper — no duplicated storage for the same data.
- [ ] Unit tests cover: load-on-init, save-on-change (debounced), clear action, and corrupt / missing storage value handling.
- [ ] Manual QA: reload the browser after typing a query in each portlet and confirm the query is restored.

### Priority

Medium — quality-of-life fix for power users of the dev tools; not blocking.

### Additional Context

Existing persistence pattern to draw from: `libs/portlets/dot-velocity-playground/src/lib/dot-velocity-playground.utils.ts` (`readJson`, `writeJson`, `HISTORY_STORAGE_KEY`, `SPLITTER_STORAGE_KEY`).

AP2300:
## Summary

Closes #36692. Two consistency gaps across the migrated dev-tool portlets (`dot-query-tool`, `dot-es-search`, `dot-velocity-playground`), addressed together.

- **Persist last query per portlet** — new shared `withPersistedQuery()` SignalStore feature in `libs/data-access/dot-localstorage/` (backed by pure `readJson`/`writeJson`/`removeKey` helpers, SSR-safe, error-swallowing). Storage key convention: `dotcms.devtools.{portletKey}.lastQuery`. Hydrates on init, debounces writes at 300 ms, exposes a `clearPersistedQuery()` action. Wired into all three stores.
- **Clear-query button** in each editor's panel header (`pi pi-eraser`, tooltip via `| dm`, disabled-when-empty, no confirmation — conscious action per issue discussion). Sits inline next to Help so it doesn't fight PrimeNG's `pTemplate="icons"` slot on portlets that also have a Wrap-code checkbox.
- **Removed dark Monaco theme** from `dot-velocity-playground` — kept the VTL syntax tokenizer, dropped the hardcoded `dot-velocity-dark` theme. Editor now inherits `DOT_MONACO_BASE_OPTIONS`' `theme: 'vs'`, the same light theme used by the other two dev tools.
- **Reconciled velocity persistence** — the implicit "on init, prime `code` from `history[0]`" was a duplicate persistence path for the query. Removed; the new persisted-query key is now the single source of truth. History remains for the multi-entry selector.
- **i18n**: added `queryTool.action.clearQuery`, `esSearch.action.clearQuery`, `velocityPlayground.action.clearQuery` to `Language.properties`.

## Acceptance criteria coverage

- [x] Dark theme removed from `dot-velocity-playground` (theme definition + `defineTheme`/`setTheme` calls + `theme:` bindings)
- [x] All three portlets render with the same default Monaco theme
- [x] Shared persistence helper in `libs/data-access/` (not duplicated)
- [x] Restores last query on load
- [x] Saves on change (debounced 300 ms)
- [x] Per-portlet Clear action
- [x] Storage keys follow `dotcms.devtools.{portlet}.lastQuery`
- [x] Velocity's existing localStorage (history, splitter, wrap) reconciled — no duplicate storage of the query
- [x] Unit tests: load-on-init, debounced-write, clear, corrupt/missing storage

## Test plan

- [x] Type a query in each portlet, reload the page, confirm the query is restored
- [x] Click Clear query, confirm the editor empties and reload leaves it empty (localStorage entry gone)
- [x] Rapid typing writes only the final value to localStorage after ~300 ms (Application → Local Storage in devtools)
- [x] Velocity editor renders with the shared light theme, VTL syntax still highlighted (directives, variables, strings)
- [x] Wrap-code checkbox on es-search / velocity headers renders on the right without wrapping
- [x] Manually corrupt the stored value (`{not-json`) and confirm the portlet loads with an empty editor rather than erroring

This PR fixes: #36692


core-web/libs/data-access/src/lib/dot-localstorage/dot-persisted-query.utils.ts
```diff
@@ -0,0 +1,54 @@
...
+/** Build the full localStorage key for a given portlet's last query. */
+export const buildPersistedQueryKey = (portletKey: string): string =>
+    `${PERSISTED_QUERY_KEY_PREFIX}.${portletKey}.lastQuery`;
+
+/**
+ * Read a JSON value from localStorage with a typed fallback. Returns the
+ * fallback in non-browser environments, when the key is missing, or when the
+ * payload can't be parsed.
+ */
+export const readJson = <T>(key: string, fallback: T): T => {
```
hmoreras, line 24: These three helpers (`readJson` / `writeJson` / `removeKey`) already exist in `libs/portlets/dot-velocity-playground/src/lib/dot-velocity-playground.utils.ts` (lines 28-57) with identical implementations — same SSR guard, same try/catch, same JSON parse/stringify.

Since one AC of this PR is *"Shared persistence helper in `libs/data-access/` (not duplicated)"*, it would be worth finishing the consolidation: keep the copies here as the single source, drop them from the velocity utils, and have the velocity store import `readJson` / `writeJson` / `removeKey` from `@dotcms/data-access` for its HISTORY / SPLITTER / WRAP keys. Otherwise the same three functions live in two places inside the very portlet the AC targets.

core-web/libs/data-access/src/lib/dot-localstorage/with-persisted-query.feature.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    return signalStoreFeature(
+        { state: type<Record<Field, string>>() },
+        withMethods((store) => ({
+            /**
+             * Reset the persisted query to an empty string and delete the
+             * localStorage entry. The debounced-write effect will observe the
+             * empty value but the removeKey call has already cleared storage.
+             */
+            clearPersistedQuery(): void {
```
hmoreras, line 96: Small race worth double-checking: the state change here (`patchField(store, '')`) re-fires the tracked `effect` below, which schedules a `setTimeout(() => writeJson(storageKey, ''), 300)`. 300 ms later, `writeJson` writes `""` (JSON-encoded empty string) back to storage — the entry `removeKey` just deleted gets re-created.

User-visible impact is small (on reload, `stored.length > 0` is false → initial state kept), but it contradicts the docstring's *"removeKey call has already cleared storage"* and the unit test's `getItem(persistedKey) === null` assertion. The test passes only because it doesn't `advanceTimersByTime(300)` after `clearPersistedQuery()`.

Two-line fix inside the effect works — skip the write when `value === ''` (and `removeKey` there too for safety) — plus a regression test that advances timers past the debounce window.

core-web/libs/data-access/src/lib/dot-localstorage/with-persisted-query.feature.ts
```diff
@@ -0,0 +1,118 @@
...
+
+                effect((onCleanup) => {
+                    const value = readField();
+                    if (!hydrated) {
+                        hydrated = true;
+
+                        return;
+                    }
+
+                    const timer = setTimeout(() => writeJson(storageKey, value), debounceMs);
```
hmoreras: try to avoid use setTimeout here

hmoreras (changes requested):

core-web/libs/data-access/src/lib/dot-localstorage/dot-persisted-query.utils.ts
AP2300: all utils now imported from data-access

core-web/libs/data-access/src/lib/dot-localstorage/with-persisted-query.feature.ts
AP2300: Fixed in b099a6fe5f: empty values in the debounced tail route through removeKey instead of writeJson, so clearing no longer resurrects the entry. Added a regression test that advances timers past the debounce after clearPersistedQuery() and asserts storage stays null.

core-web/libs/data-access/src/lib/dot-localstorage/with-persisted-query.feature.ts
AP2300: timeout has been replaced to listen to the rxMethod

hmoreras (approved):

Issue #36244, opened 2026-06-19
Unify OpenSearch startup connection gate: phase-aware retry/fallback across both startup paths


fabrizzio-dotCMS:
## Description

dotCMS has **inconsistent OpenSearch (OS) connection gating at startup**, and the OS readiness gate does not respect the migration phase. Three concrete gaps were confirmed against the current code (see file:line below). Together they cause an opaque, late-failing crash when OS is unreachable/misconfigured during startup, instead of a fast, actionable failure with phase-correct behavior (fallback during migration, abort once OS is primary).

This is the follow-up to a finding made on 2026-06-12 while smoke-testing Phase 3 against the `os-migration` limited-user stack: the real failure was a TLS scheme mismatch (`http://` against an OS 3.x port that only accepts `https`), surfacing as `ConnectionClosedException` deep inside `createContentIndex` ~30s into startup (`SystemExitManager - Startup failure`), with no actionable message.

### Current behavior (confirmed)

| Gate | Where | Today |
|---|---|---|
| ES readiness | `ESIndexAPI.waitUtilIndexReady()` (`ESIndexAPI.java:1150`) | Retries `ES_CONNECTION_ATTEMPTS` (default 24, 5s sleep); on exhaustion → `SystemExitManager.immediateExit(1)`. Always aborts. |
| OS readiness | `OSIndexAPIImpl.waitUtilIndexReady()` (`OSIndexAPIImpl.java:512`) | Mirrors ES retry (`OS_CONNECTION_ATTEMPTS` / `ES_CONNECTION_ATTEMPTS`, configurable sleep); on exhaustion → `immediateExit(1)`. **Always aborts — not phase-aware.** |
| Semantic validator | `IndexStartupValidator.validateIndexingConfig()` via `ContentletIndexAPIImpl.checkAndInitializeIndex()` (`ContentletIndexAPIImpl.java:524-541`) | Single probe, **no retry**. Phase-aware: Phase 3 → `DotRuntimeException`; Phase 0–2 → `haltMigration()` (fallback). |

### Startup ordering problem (empty-DB path)

By servlet `load-on-startup` order:

1. `MainServlet` (=1) calls `APILocator.getESIndexAPI().waitUtilIndexReady()` (`MainServlet.java:113`) — **waits on ES only, hardcoded**, even in Phase 3 where OS is primary — then runs the startup tasks.
[...]

PR #36783, opened 2026-07-28
fix(opensearch): degrade OS index-bootstrap failures to ES-only in dual-write phases (#36222)

### LINKED ISSUE #36222: [QA-G16] OpenSearch Migration — Limited (Non-Admin) OpenSearch User (TC-053–TC-058)
## Parent Epic

#35476

## Goal

Prove dotCMS can run the OpenSearch migration against an OS 3.x cluster reached through a
**restricted, non-admin user** (the managed-cloud setup validated in spike #35922) — not just an
open, security-disabled cluster. Covers provisioning the user, pointing dotCMS at it, and confirming
index operations work through the limited role.

## Glossary (quick)

- **ES (old engine):** Elasticsearch — `http://localhost:9200`, plain HTTP, no auth.
- **OS (new engine):** OpenSearch 3.x — `https://localhost:9201`, **TLS + basic auth**.
- **Migration phase:** 0 = ES only, 1 = dual-write (search ES), 2 = dual-write (search OS).
- **`.os` tag:** new-engine index names end in `.os` (e.g. `cluster_dotcms….working_….os`).

## Setup — limited-user stack + the provisioning script

**Stack:** `docker/docker-compose-examples/os-migration/docker-compose.limited-user.yml` (OS 3.x with
the security plugin ON, provisioned with the non-admin `dotcms-es-user`). The
`docker/docker-compose-examples/single-node-os-migration/` variant runs OS 1.x + OS 3.x, provisioned
the same way.

```bash
docker compose -f docker/docker-compose-examples/os-migration/docker-compose.limited-user.yml up -d
```

**Users on OS 3.x (port 9201, HTTPS):**

| User | Password | Role |
|---|---|---|
| `admin` | `Dev!Search3-Kx9mP-2026` | cluster admin |
| `dotcms-es-user` | `Dev!dotcms-EsUser-2026` | `dotcms-role` (limited) |

**Provisioning script `opensearch.py`** — runs once via the `opensearch-provision` service; creates
per customer an internal user `<customer>-es-user`, action groups, a role `<customer>-role`, and the
role mapping. Run it manually:

```bash
# Re-run the bundled provisioner against the running stack (idempotent):
docker compose -f docker/docker-compose-examples/os-migration/docker-compose.limited-user.yml run --rm opensearch-provision

# Or standalone against any reachable cluster:
./opensearch.py --admin-user admin --admin-pass 'Dev!Search3-Kx9mP-2026' \
  --password 'Dev!dotcms-EsUser-2026' --customer dotcms --host localhost --port 9201
```

The `dotcms-role` grants:
- **Index** (pattern `cluster_<customer>*`): `indices_all`, `indices_monitor`
- **Cluster:** `cluster:monitor/health`, `cluster:monitor/state`, `cluster:monitor/nodes/stats`,
  `indices:data/write/bulk`, `indices:data/read/scroll`, `indices:data/read/scroll/clear`, and
  **`cluster:monitor/main`** (GET `/` version probe — the spike #35922 fix; without it dotCMS misreads
  OS as unreachable and silently falls back to ES)
- **All-indices:** `indices:monitor/stats`, `indices:monitor/settings/get`, `indices:admin/aliases/get`

⚠️ **Cluster-id gotcha:** the role only grants index permissions on names matching `cluster_<customer>*`.
dotCMS names indexes `cluster_<DOT_DOTCMS_CLUSTER_ID>.*`, so **`DOT_DOTCMS_CLUSTER_ID` must start with
the customer name** (e.g. `dotcms-os-migration`) or the limited user gets 403 on its own indexes.

⚠️ **The cluster id can only be set on the very first boot.** dotCMS stores it in the database
(`dot_cluster`) the first time it starts, and from then on the stored value wins: the environment
variable, the system property and the config file are only consulted when that row does not exist yet
(it is also blocked in the system-table API). So **changing `DOT_DOTCMS_CLUSTER_ID` and restarting an
instance that already has data has no effect at all** — the index names stay as they were. To get a
different cluster id you need a **fresh database volume**. This matters for TC-056, which is written
around a different way of producing the same rejection.

**dotCMS config:**
```properties
DOT_ES_ENDPOINTS=http://<host>:9200
DOT_OS_ENDPOINTS=https://<host>:9201
DOT_OS_AUTH_TYPE=BASIC
DOT_OS_AUTH_BASIC_USER=dotcms-es-user
DOT_OS_AUTH_BASIC_PASSWORD=Dev!dotcms-EsUser-2026
DOT_OS_TLS_TRUST_SELF_SIGNED=true
DOT_FEATURE_FLAG_OPEN_SEARCH_PHASE=1
DOT_DOTCMS_CLUSTER_ID=dotcms-os-migration   # MUST start with the customer name
```

## Test Cases

**TC-053 — Provision the limited user and verify the role/permissions** _(Risk: Medium)_
1. Launch the limited-user stack; let `opensearch-provision` finish.
2. List users: `curl -sk https://localhost:9201/_plugins/_security/api/internalusers?pretty -u admin:'Dev!Search3-Kx9mP-2026'` → `dotcms-es-user` present.
3. Inspect role: `curl -sk https://localhost:9201/_plugins/_security/api/roles/dotcms-role?pretty -u admin:'Dev!Search3-Kx9mP-2026'`.
4. `curl -sk https://localhost:9201/ -u dotcms-es-user:'Dev!dotcms-EsUser-2026'` → HTTP 200.
- **Expected:** `dotcms-es-user` + `dotcms-role` exist with the listed permissions (incl. `cluster:monitor/main`); `GET /` as the limited user returns **200**.

**TC-054 — dotCMS starts against OS as the limited user; migration NOT disabled** _(Risk: High)_
1. Stack up; dotCMS configured as in Setup (Phase 1, cluster id starts with `dotcms`).
2. Start dotCMS; watch the log; confirm the phase (Admin → System → Index, or log).
- **Expected:** dotCMS starts and runs normally; migration **NOT** auto-disabled (no "OpenSearch cluster is not reachable" / phase-reset lines); log shows `OS version check passed: 3.x`; both ES and `.os` indexes listed → Phase 1.

**TC-055 — Dual-write works through the limited role (Phase 1)** _(Risk: High)_
1. (TC-054 passed.) Create and publish a content item; note its identifier.
2. As the limited user, confirm it in OS (index ends in `.os`): `curl -sk "https://localhost:9201/<cluster_dotcms….working_….os>/_doc/<id>" -u dotcms-es-user:'Dev!dotcms-EsUser-2026'`.
- **Expected:** document present in the OS working index (200/found); no 403 in the dotCMS log for the OS write.

**TC-056 — OpenSearch refuses the index names dotCMS asks for** _(Risk: Medium)_

_Why this case exists:_ the limited user may only work on indexes whose name starts with the customer
name it was created for (here `cluster_dotcms…`). dotCMS builds its index names from the cluster id,
so when the two do not line up, OpenSearch refuses every index dotCMS tries to **create** with a
**403**. In Phase 1 Elasticsearch is still the engine serving the site, so this must stay an
OpenSearch-only problem: dotCMS has to keep running and the site has to keep working.

> **The permission to take away is the one to CREATE indexes** — not the one to write documents.
> Removing the write permission produces a different, already-correct behaviour (rejected document
> writes, logged and ignored) and does **not** exercise this case. Both previous QA rounds ended
> inconclusive on exactly this point.

_Automated coverage:_ the product contract below is asserted by
`OsBootstrapForbiddenIndexTest` (`dotcms-integration`, registered in `OpenSearchUpgradeSuite`, runs in
CI): Phase 1 bootstrap absorbs the 403 and halts the migration, the OpenSearch index store is left
without pointers to indexes that were never created, Phase 1 reindex still registers its Elasticsearch
slots, and Phase 3 propagates the failure instead of degrading. **The manual case below only covers
what a machine cannot see: that a human using dotCMS notices nothing broken.**

**Setup — remove the create-index permission (OpenSearch Dashboards UI, no cluster shell needed)**

1. In dotCMS, go to Admin → System → Index and read the prefix of any listed index: it looks like
   `cluster_<clusterid>.working_<timestamp>`. Note that `cluster_<clusterid>` — that is the pattern
   OpenSearch must stop allowing.
2. In OpenSearch Dashboards → Security → Roles → the role mapped to the dotCMS user
   (`dotcms-role`), open **Index permissions** and change the index pattern from
   `cluster_dotcms*` to something that cannot match, e.g. `cluster_nomatch*`. Save.
   - This removes create *and* write in one move, which is what the case needs. Leave the
     second block (pattern `*` with the monitor/stats actions) untouched.
3. **Arming check — do this before anything else:** in Dashboards → Index Management, list the
   indexes ending in `.os` and write down what is there. This list is the reference for step 6.

**Steps**

4. In dotCMS, go to Admin → System → Index and press **Reindex**. Wait for it to finish.
5. Create and publish a content item, then search for it in the back end.
6. In Dashboards → Index Management, list the `.os` indexes again and compare with step 3.
7. In the dotCMS log (unfiltered, INFO level and above), search for these three strings.

**Expected**

- **Step 4:** the Reindex finishes with **no error in the UI**, and the Index screen keeps listing the
  Elasticsearch indexes.
- **Step 5:** publishing works and the item is found by search — everything is served by
  Elasticsearch.
- **Step 6:** **no new `.os` index appeared.** The list is the same as in step 3.
- **Step 7:** all three strings are present, in this order:

  | # | Log string | What it proves |
  |---|---|---|
  | 1 | `Error creating index: cluster_….reindex_working_….os` | OpenSearch refused the creation — the case was actually exercised |
  | 2 | `OpenSearch index creation failed — working=… likelyCause=AUTH_FORBIDDEN` | dotCMS recognised the cause and says how to fix it |
  | 3 | `Migration phase reset to PHASE_0_MIGRATION_NOT_STARTED` | the migration switched itself off; dotCMS is on Elasticsearch only |

**Recovery / re-enable**

8. Put the index pattern back (`cluster_dotcms*`), set the phase back to 1, restart dotCMS and
   reindex once more → the `.os` indexes are created again and both engines are listed
   (back to TC-054 / TC-055 behaviour).

**Verdict rules — please use these literally**

- **PASS** — steps 4, 5, 6 behave as above **and** all three log strings from step 7 are present.
- **FAIL** — string 1 **is** present (the creation was refused) but string 2 or 3 is missing, or the
  Reindex errored, or publishing/searching broke, or a new `.os` index was registered anyway. This is
  a product problem: report it.
- **INCONCLUSIVE — not a failure** — string 1 is **absent**. That means OpenSearch never refused a
  creation, so the condition under test never happened (usually: the role still allows creating
  indexes, or no new index needed to be created). Do not mark the case failed; re-check the role in
  step 2 and repeat, or hand the log over and we will read it with you.

_Alternative setup (needs a clean install, only if the role cannot be edited):_ start the stack with a
**new database volume** and `DOT_DOTCMS_CLUSTER_ID=acme-test` (does not start with `dotcms`) in Phase
1, leaving `dotcms-role` as provisioned. Same expected results, read from the start-up log instead of
from the Reindex button. Do **not** try this by editing the variable on an existing instance — the
cluster id is stored in the database on first boot and the stored value always wins, so nothing gets
refused (see the note in **Setup**). This is how round 2 ended up testing a different thing.

**TC-057 — Read/search works through the limited user** _(Risk: Medium)_
1. Content dual-written (TC-055). As the limited user: `curl -sk "https://localhost:9201/<…os>/_search?q=*:*" -u dotcms-es-user:'Dev!dotcms-EsUser-2026'`.
2. (Phase 2) run a dotCMS content search and confirm results.
- **Expected:** search returns results (no 403); scroll-based reads succeed.

**TC-058 — (Regression) Without `cluster:monitor/main`, dotCMS misreads OS as unreachable** _(Risk: Medium)_
1. Remove `cluster:monitor/main` from `dotcms-role` (security API, or re-provision without it).
2. Confirm `GET /` as the limited user now returns **403**.
3. Restart dotCMS in Phase 1 against this user; watch the log.
- **Expected:** version probe fails → migration auto-disabled (the "OpenSearch cluster is not reachable" + phase-reset lines), dotCMS falls back to ES. Restoring `cluster:monitor/main` (re-run the provisioner) → TC-054 passes again.

## Acceptance Criteria

- [ ] TC-053: `dotcms-es-user`/`dotcms-role` provisioned with expected perms (incl. `cluster:monitor/main`); `GET /` as limited user = 200.
- [ ] TC-054: dotCMS starts as the limited user; migration stays ON (Phase 1); no halt lines.
- [ ] TC-055: published content lands in the OS index via the limited role; no 403.
- [ ] TC-056: index names OpenSearch refuses keep dotCMS running on Elasticsearch (start-up, reindex, publish and search all work, no error in the UI), the log names the refused indexes with 403 and how to fix it, the migration turns itself off (phase 0) and no `.os` index is registered; restoring the permission brings Phase 1 back. In Phase 3 the same misconfiguration must fail visibly instead.
- [ ] TC-057: search/read through the limited user succeeds.
- [ ] TC-058: removing `cluster:monitor/main` reproduces the unreachable→halt failure; restoring it recovers.

## Notes

- Validates the managed-cloud limited-user setup from **spike #35922**; the `cluster:monitor/main` grant is the key fix that keeps the OS version probe (`client.info()`) working for a non-admin user.
- **TC-056 was rewritten on 2026-07-28.** The previous wording expected the OpenSearch write to be ignored and the site to carry on; what it actually described was the behaviour of content writes, not of index creation. The rejected index creation used to take the whole start-up down (see the QA-Failed report below), which is fixed in **PR #36783**: an index rejected by OpenSearch now switches the migration off and leaves dotCMS running on Elasticsearch, with a log line that names the cluster-id fix. Re-test TC-056 against a build that includes that PR.
- **TC-056 was updated again on 2026-08-05**, after the second QA run. Two things were wrong in the previous wording, both about *how to test it* rather than about the product: (1) it told the tester to change `DOT_DOTCMS_CLUSTER_ID`, which cannot be done on an instance that already has data — the value is stored in the database on first boot and the stored one always wins, so the mismatch never happened; the case now induces the same rejection by narrowing the OpenSearch role instead. (2) it did not say **where** the failure is reported: in Phase 1 the reindex is an Elasticsearch operation that succeeds, so no error is shown in the UI on purpose, and the evidence lives in the log (the `AUTH_FORBIDDEN` / "falling back to ES-only" line plus the phase dropping to 0). A Phase 3 caveat was added as well: there the same misconfiguration must fail visibly.
- ES stays plain HTTP / no auth; only OS 3.x is HTTPS + authenticated in this stack.

fabrizzio-dotCMS:
### What QA hit

With a limited OS user (role scoped to `cluster_dotcms*`) and `DOT_DOTCMS_CLUSTER_ID` **outside** that pattern, TC-056 expected the OS side to fail harmlessly while ES kept working. Instead index creation failed outright, no index was created at all, and the reindex endpoint returned **HTTP 400**:

```
INFO util.SecurityLogger - class com.dotcms.rest.exception.mapper.DotBadRequestExceptionMapper : Failed to create index:
```

### Root cause

OpenSearch rejects the create with `HTTP 403` (`security_exception`) because the index name carries a cluster-id prefix the role does not cover. `OSIndexAPIImpl.createIndex` surfaces that as an **unchecked** `DotStateException`, but `ContentletIndexAPIImpl.bootstrapAndPointOS` only caught `IOException`, so the rejection escaped:

- **startup** — `checkAndInitializeIndex()` rethrows as `DotRuntimeException` while the migration is on, aborting index initialisation for *both* providers;
- **runtime** — `POST /api/esindex/reindex` → `refreshAllContent()` → `fullReindexStart()` → `initIndex()` wraps it in `DotReindexStateException` (a `DotStateException`), which the JAX-RS mapper turns into the 400 QA saw.

The existing OS connection gate (#36244) cannot catch this: it probes `GET /`, which the limited user **is** allowed to call (`cluster:monitor/main`), so the gate passes and the 403 only appears on the create request.

### Fix

`bootstrapAndPointOS` now catches `Exception` and applies the same phase-aware policy the connection gate already uses, in `handleOsBootstrapFailure`:

| Phase | Behaviour |
|---|---|
| 1 / 2 (OS is a shadow store) | Log an actionable ERROR, `haltMigration()` → ES-only, skip `pointOS` so no store pointer is left for an index that was never created. dotCMS starts and reindexes normally on ES. |
| 3 (OS is primary) | Propagate as `DotDataException` — there is no ES to fall back to and the phase is never auto-rolled back. |

When the failure classifies as `AUTH_FORBIDDEN` the log names the actual remediation — the OS role's index pattern must cover the dotCMS index names, i.e. `DOT_DOTCMS_CLUSTER_ID` must start with the customer name the role was provisioned for.

`OSIndexAPIImpl.classifyConnectionError` / `ConnectionFailureKind` are widened to `public` so the bootstrap path reuses the existing classification instead of duplicating it.

### Note on the test case

TC-056's expected result ("dotCMS logs the OS write failure, fire-and-forget") describes the *document write* path, which already degrades correctly (`PhaseRouter` + the `createContentIndex` fan-out — that is why TC-055/TC-057 passed). For *index bootstrap* the correct Phase-1 outcome is the one implemented here: OS bootstrap fails → migration halted → dotCMS keeps serving from ES with an actionable log. Worth updating the wording of TC-056 when re-testing.

### Tests

- `ContentletIndexAPIImplOsBootstrapFailureTest` (unit, 4 cases) — the per-phase policy: absorbed in phases 1/2 with the phase reset to 0, propagated in phase 3 with the phase untouched, and degradation driven by the phase rather than the failure kind. **Passing locally.**
- `OsBootstrapForbiddenIndexTest` (integration, registered in `OpenSearchUpgradeSuite`) — real DB-backed index stores, real ES provider and `IndexAPI` router against a stub OS provider that reports the cluster reachable and then rejects the create with a 403: Phase 1 must not throw, must halt the migration and must leave no OS store pointer; Phase 3 must throw `DotDataException` and keep the phase.

```bash
./mvnw verify -pl :dotcms-integration -Dcoreit.test.skip=false \
    -Dopensearch.upgrade.test=true -Dit.test=OsBootstrapForbiddenIndexTest
```

### Manual verification (reproduces TC-056)

1. Bring up `docker/docker-compose-examples/os-migration/docker-compose.limited-user.yml`.
2. Configure dotCMS in Phase 1 with `DOT_DOTCMS_CLUSTER_ID=acme-test` (outside `cluster_dotcms*`).
3. Start dotCMS → it comes up on ES, logs the 403 with the cluster-id remediation and resets the phase to 0; reindex from the UI works instead of returning 400.
4. Set `DOT_DOTCMS_CLUSTER_ID=dotcms-os-migration`, re-enable Phase 1, restart → OS indices are created and dual-write resumes (TC-054/TC-055).

This PR fixes: #36222


fabrizzio-dotCMS:
## Audit: the rest of the OpenSearch admin operations under a permission-scoped role

Reviewed every OS write/admin operation reachable in a dual-write phase, asking the same question that #36222 exposed: *can a 403 from an index-scoped OS role propagate and break something that Elasticsearch was serving fine?*

**Two more sites had the same defect, both fixed in c47de3a** (`optimize`, `flushCaches`). Everything else was already protected.

### Findings

| Operation / path | OS provider on failure | Phase 1 outcome | Verdict |
|---|---|---|---|
| `optimize` (tag-dispatched, `IndexAPIImpl`) | throws `RuntimeException` | **propagated** → 500 on `POST /api/esindex/optimize` and the CMS-maintenance AJAX | **was broken → fixed** |
| `flushCaches` (tag-dispatched, `IndexAPIImpl`) | throws `RuntimeException` | **propagated** → 500 on `DELETE /api/esindex/cache` | **was broken → fixed** |
| Index bootstrap (`bootstrapAndPointOS`) | throws `DotStateException` | propagated → start-up abort / 400 on reindex | fixed in cd1ea03 |
| `create/clear/delete/deleteMultiple/updateReplicas/closeIndex/openIndex/createAlias/deleteInactiveLiveWorkingIndices` | throw | absorbed — all go through `PhaseRouter` (OS is the shadow) | OK |
| Content writes (`putToIndex`, bulk processor) | throw | absorbed — ES is hardcoded as authoritative in both dual phases; OS listener is a shadow listener | OK |
| Base + custom mapping (`putMapping` via `addCustomMapping`) | throws | absorbed — `ESMappingUtilHelper` catches per field/relationship; the base mapping runs inside `createContentIndex`, covered by cd1ea03 | OK |
| Post-switchover optimize (`optimizeNewActiveIndicesAsync`) | throws | absorbed — per-provider try/catch, async | OK |
| Activate / deactivate OS mirror | DB only | absorbed — try/catch around the mirror | OK |
| Listing / stats / health (`listIndices`, `getIndices`, `getClosedIndexes`, `getIndicesStats`, `getClusterHealth`, `getClusterStats`) | swallow internally, return empty | absorbed | OK |
| Site search (`SiteSearchAPIImpl`) | throws | absorbed — `PhaseRouter`; the `listIndices`/`listClosedIndices` merge builds on the swallowing primitives | OK |
| Orphan probes (`indexExists`, `getIndexDocumentCount`) | throw | absorbed — `Try` with a safe default | OK |
| Start-up validation (`IndexStartupValidator`) | — | only `client.info()` + config comparison, no index-level call | OK (this is precisely the blind spot) |

### One design point worth a decision (not changed here)

`PhaseRouter` treats the **read provider** as primary, so in **Phase 2** OS is primary: a 403 from OS then propagates on every router write op, while an ES failure is the one being swallowed. Content writes are exempt (`putToIndex` pins ES as authoritative in both dual phases), so this only affects index-admin actions in Phase 2. That is the router's documented contract, and changing it has a wide blast radius — flagging it rather than touching it. Happy to open it as a separate discussion if we want Phase 2 to keep degrading to ES for admin operations too.

Note that with a correctly configured cluster id the limited role grants `indices_all` inside its pattern, so none of these 403s occur — all of them require the TC-056 misconfiguration (or a role narrowed further than the one the provisioner creates).

fabrizzio-dotCMS:
## Review response

Thanks — four of these were real defects and are now fixed; two I'm answering with reasoning rather than code, and one I'm treating as out of scope for this PR. Verdict per finding, in the reviewer's numbering.

### 1. `haltMigration()` may be a no-op when the phase comes from the system table — **valid, addressed**

Correct, and worth stating that this hole is not new: the shipped connection gate (`bootstrapAndPointOS` → `waitUtilIndexReady` → `handleConnectionExhausted`) already calls `haltMigration()` and skips `pointOS` on the same assumption. What this PR added was a second trigger for it.

I did not make the handler propagate in that case, though: aborting index initialisation is exactly the failure mode this PR exists to remove, and with no OS pointer written the dual-write path already skips OS cleanly (`loadProviderIndices` returns null → "skipping OS write"), so Phase 1 keeps working. What was genuinely wrong was staying silent. The handler now verifies the reset stuck and, when it did not, logs a FATAL that names the cause (the phase is set through the DB-backed config, where a runtime reset does not apply), what state the instance is in, and what to do before OpenSearch reads are enabled.

### 2. First-exception halt with no retry and no resync — **valid concern, no code change (reasoning below)**

Two parts. On **retry**: the connection gate immediately upstream of this code already retries OS reachability `OS_CONNECTION_ATTEMPTS` times (default 24) and only then halts, so by the time we reach the create call connectivity was just verified; adding a second retry loop around the create would mostly duplicate it. The alternative — propagating transient failures — is the #36222 regression itself.

On **resync**: an OS index that is bootstrapped after a halt does not get backfilled, correct, but that is the migration's existing documented behaviour, not something this PR introduces — `hasEmptyIndices()` deliberately triggers a reindex off ES emptiness only in phases 0-2, with the rationale in the code comment ("in phases 1/2 the OS index is expectedly empty during catchup … the OS catchup happens via normal dual-write as content changes"). Changing that is a migration-wide decision about when a full reindex is forced, and I don't want to smuggle it into a bug fix. Worth its own discussion — I'd rather do it deliberately than as a side effect here.

### 3. `initAndPointReindex` registers OS reindex pointers for indices that were never created — **valid, fixed**

This was the best catch: the reindex path pointed the OS store at slots the fan-out had silently failed to create, so reindex writes would let OpenSearch auto-create a dynamically-mapped twin that `fullReindexSwitchover` later promotes. Creation is now targeted per provider so each store is only pointed at what was actually created: ES failures propagate (primary), an OS failure in a dual-write phase goes through the same `handleOsBootstrapFailure` policy and leaves the OS reindex slots unset, Phase 3 propagates. A soft `false` return is handled too — it aborts in Phase 3 and logs + skips the pointer in dual-write.

### 4. `classifyConnectionError` misclassifies via bare `403`/`401` substrings — **valid, fixed**

Confirmed: the message handed to the classifier embeds the physical index name, and a `_yyyyMMddHHmmss` timestamp routinely contains those digit runs (12:04:03 → `…120403…`), so a settings-parse or orphan-delete failure was reported as `AUTH_FORBIDDEN` and the new branch then told the operator to change `DOT_DOTCMS_CLUSTER_ID` for something unrelated. Status codes are now matched with `(?<![0-9])(401|403)(?![0-9])`, which still recognises `status: 403` and `HTTP/1.1 403` but not digits inside a timestamp. Two tests added, including the false-positive case.

### 5. Phase 2 absorbs failures of the provider that serves reads — **valid, fixed**

Right, and my gate was wrong: `isMigrationComplete()` treated Phase 2 like Phase 1. It is now `isReadEnabled()`, so `optimize`/`flushCaches` absorb an OS failure only while OS is the shadow store and propagate from Phase 2 on. That also makes the tag-dispatched pair consistent with `PhaseRouter`, which absorbs shadow failures only. Tests updated: the Phase-2 cases now assert propagation.

### 6. `indexSuffixOS()` still advertises a timestamp after an absorbed failure — **valid, fixed**

`bootstrapAndPointOS` now returns whether it actually registered the OS indices, and `initIndex` / `initOSCatchup` report `indexSuffixOS` accordingly (all three catchup cases). Latent as noted, but it was a lie in the return value, so worth closing while the code is open.

### 7. Partial success leaves an orphan the log does not mention — **valid, addressed in the message**

Self-healing via orphan reuse, as noted, so no logic change; the ERROR now says explicitly that an index created before the failure stays in the cluster, is reused or repaired by the next bootstrap, and is never registered in the store — which is what an operator cleaning up a permission-scoped cluster needs to know.

dotcms-integration/src/test/java/com/dotcms/OpenSearchUpgradeSuite.java
```diff
@@ -54,6 +55,7 @@
         ContentletIndexAPIImplMidReindexRollbackIT.class,
         MigrationPhaseStoreBootstrapIT.class,
         DeactivateIndexEmptyStoreIT.class,
+        OsBootstrapForbiddenIndexIT.class,
```
nollymar: Fab, the IT should be sufixed by Test not IT. It might be worth adding this rule to the Claude's context

jcastro-dotcms (approved):

dario-daza (approved):

PR #36796, opened 2026-07-29
Update CI CD to ensure SDK publishing occurs

### LINKED ISSUE #36701: Reintroduce next npm dist-tag for internal dev/QA testing on merge to main

KevinDavilaDotCMS:
### Proposed Changes
* change 1
* change 2

### Checklist
- [ ] Tests
- [ ] Translations

### Additional Info

### Screenshots

This PR fixes: #36701

This PR fixes: #36701

### DIFF
--- a/.github/workflows/cicd_3-trunk.yml
+++ b/.github/workflows/cicd_3-trunk.yml
@@ -137,12 +137,17 @@ jobs:
   publish-sdk-next:
     name: Publish SDK (next)
     needs: [ initialize, test ]
-    # Explicit success check on `test` (not the always()/!failure() idiom used by
-    # sibling jobs) — this job publishes to a real, public npm registry, so it should
-    # require `test` to have actually run and passed, not merely "didn't fail/wasn't
-    # cancelled" (which also admits a skipped test job).
+    # Same always()/!failure()/!cancelled() idiom as sibling jobs (build-cli, deployment).
+    # A prior version of this gate required needs.test.result == 'success', but on an
+    # ordinary push cicd_3-trunk.yml never wires any per-suite flag (jvm_unit_test, cli,
+    # frontend, postman, karate) into the reusable test phase — only run-all-tests, which
+    # is only ever true on a manual workflow_dispatch. So the test matrix is always empty
+    # (0 configurations) on a regular push, `test` always resolves to "skipped", and that
+    # stricter gate meant this job silently never ran on ordinary merges — defeating its
+    # whole purpose (any SDK-touching merge must publish a `next` version). This still
+    # blocks if `test` genuinely ran and failed (e.g. a manual run-all-tests dispatch).
     if: >-
-      needs.test.result == 'success' &&
+      always() && !failure() && !cancelled() &&
       fromJSON(needs.initialize.outputs.filters).sdk_libs == 'true' &&
       github.event_name != 'workflow_dispatch'
     runs-on: ubuntu-${{ vars.UBUNTU_RUNNER_VERSION || '24.04' }}


KevinDavilaDotCMS:
Went through both findings from the Claude Code review — both are 🟡 Medium, but worth settling since they question the core rationale of this fix.

## 1. Was `needs.test.result` really `'skipped'`, or would it actually resolve to `'success'`?

Good question to ask, and it's already answered by real data, not just theory: this fix was written specifically because of a real production run — [run 30474411639](https://github.com/dotCMS/core/actions/runs/30474411639) — where a push that genuinely touched `core-web/libs/sdk/client/...` (so `sdk_libs` was `true`, and the event was a plain `push`, not `workflow_dispatch`) still showed:

```
Publish SDK (next) | completed | skipped
```

With `sdk_libs == 'true'` and `github.event_name != 'workflow_dispatch'` both satisfied, the *only* way the old gate (`needs.test.result == 'success'`) could still skip the job is if `test`'s result genuinely wasn't `'success'`. So the diagnosis holds up against a real run, not just against reading the YAML. Confirmed root cause: `cicd_3-trunk.yml` never wires any per-suite flag (`jvm_unit_test`, `cli`, `frontend`, `postman`, `karate`) into the reusable test phase for a normal push — only `run-all-tests`, which is only ever `true` on a manual `workflow_dispatch`. So the test matrix is always empty on an ordinary push, and the caller-side `test` job resolves to `skipped`, not `success`.

## 2. Would this still publish `next` if `build` genuinely failed (not just got skipped for artifact reuse)?

Checked this against how `failure()` actually behaves here: it evaluates the *whole transitive dependency chain*, not just a job's direct `needs`. `publish-sdk-next` needs `[initialize, test]`, and `test` itself needs `build` — so if `build` truly fails, that failure propagates through `test` and is still visible to `publish-sdk-next`'s `!failure()` check, which correctly skips it too. This is the exact same idiom already relied on by the sibling `deployment` job, whose direct `needs` also don't list `build`, and which is already trusted to correctly refuse to deploy when `build` fails. No gap here.

Both findings check out — no changes needed. Thanks for flagging them; happy to add the run-30474411639 reference as a comment in the YAML if that'd help future readers, but didn't want to touch the diff without checking first.

erickgonzalez (approved):

rjvelazco (approved):

PR #36821, opened 2026-07-30
feat(content-drive): add Action Center for bulk workflow actions

### LINKED ISSUE #36817: Workflow Center: bulk workflow actions with one action per execution
### Description

Ship the **Action Center**: a dialog for running bulk workflow actions over a multi-item selection in Content Drive, restricted to **one action per execution**.

It opens from the toolbar once more than one contentlet is selected, replacing the flat workflow-action buttons with a single "Action Center" button. The dialog has two sections:

**Quick Actions** — system actions fired over the whole eligible selection in one request via `POST /api/v1/workflow/actions/default/fire/{systemAction}`, which accepts a contentlet list. Counts are derived client-side from row state (`live`, `archived`) that the grid already has; an action that applies to nothing in the selection is omitted rather than shown as `(0)`.

**Workflow Actions** — one collapsible panel per workflow scheme from `POST /api/v1/workflow/contentlet/actions/bulk`, with real per-action eligibility counts (the backend aggregates on the `wfstep` field, so counts are per-action, not per-selection). Fired via `PUT /api/v1/workflow/contentlet/actions/bulk/fire`.

**On the single-action restriction.** The design mock shows checkboxes, implying several bulk actions triggered together. This ticket deliberately ships single-select. A contentlet occupies exactly one workflow step at a time — `workflow_task` carries `UNIQUE (webasset, language_id)` and its `status` column holds a single step id — so firing one action moves the contentlet to `nextStep` and invalidates the other actions' preconditions and counts. Whether multiple actions can be offered at all is the subject of the linked spike.

### Acceptance Criteria

- [ ] An "Action Center" button appears in the Content Drive toolbar when more than one contentlet is selected, replacing the flat workflow-action buttons
- [ ] Folders may be selected but are excluded from every request, and the dialog states how many were ignored
- [ ] Quick Actions lists Publish, Unpublish, Archive, Unarchive and Delete, each with the number of selected contentlets it applies to
- [ ] A quick action that applies to zero selected contentlets is not listed at all
- [ ] Executing a quick action fires a single request for the whole eligible selection and reports the outcome
- [ ] Workflow Actions shows one collapsible panel per scheme, with the scheme's contentlet count and each action's own eligibility count
- [ ] Only one workflow action can be selected at a time, and Execute is enabled only for the scheme that owns the current selection
- [ ] Collapsing a scheme clears its pending selection, so Execute cannot stay armed for a hidden panel
- [ ] Actions that need extra input (push publish, move target path, assign/comment) are visibly disabled with an explanation
- [ ] An action whose count is an upper bound (`conditionPresent` — the backend does not evaluate the action's Velocity condition when aggregating) is marked as approximate
- [ ] After a successful execution the grid refreshes, the selection clears and the dialog closes, because counts go stale as soon as an action runs
- [ ] `skippedCount` / `skipReason` is surfaced when contentlets are skipped for not being on a scheme that owns the action
- [ ] Empty state when no scheme exposes actions; inline error when the lookup fails
- [ ] Behaves correctly on Community, where System Workflow is the only scheme that can exist
- [ ] Dialog has a custom header (title + selected count) and footer, and the body is the only scrollable region
- [ ] Unit tests cover the eligibility/count mapping and the dialog's execute paths

**Explicitly out of scope** (tracked separately):

- Multiple actions selected and run together — see the linked spike
- Per-item drill-down with execution results — separate ticket
- Lock / Unlock and Add to Bundle — neither has a bulk REST endpoint (see Additional Context)
- SSE progress streaming via `POST /api/v1/workflow/contentlet/actions/_bulkfire`

### Priority

High

### Additional Context

**Endpoints used** — all already exist, so this is frontend-only work:

| Purpose | Endpoint |
|---|---|
| Discover actions + counts | `POST /api/v1/workflow/contentlet/actions/bulk` |
| Fire a workflow action | `PUT /api/v1/workflow/contentlet/actions/bulk/fire` |
| Fire a system action over many items | `POST /api/v1/workflow/actions/default/fire/{systemAction}` |

Note that `contentletIds` on these endpoints holds contentlet **inodes**, not identifiers, despite the property name — the backend documents this explicitly.

**Why Lock/Unlock and Add to Bundle are excluded.** `LOCK`/`UNLOCK` are not `SystemAction` values and only single-item `PUT /api/v1/content/_lock|_unlock/{inode}` exist; the legacy JSP drives bulk unlock through a Struts command (`full_unlock_list` → `EditContentletAction._batchUnlock`) that loops server-side. Add to Bundle is only reachable through the legacy `RemotePublishAjaxAction` servlet. Both would need either a client-side N-call loop or new endpoints — worth their own ticket rather than baking workarounds into this one.

**Why System Workflow stays visible.** Hiding it would remove bulk Copy entirely (Copy has no system-action mapping, so the workflow list is its only route) and would leave the section permanently empty on Community. The system-action mapping is also per content type, so hiding one scheme would not reliably de-duplicate Publish anyway.

**Reference implementation.** The same feature exists in the legacy content-search portlet at `dotCMS/src/main/webapp/html/portlet/ext/contentlet/view_bulk_actions_inc.jsp`. It hits the same endpoints, is also one-action-per-click, and treats a Velocity condition as a reason to prompt before firing rather than firing blind.

**Draft branch:** `draft-content-drive-action-center` — a working implementation with unit tests and lint passing.

**Note on the parent epic:** #33999 describes surfacing workflow actions as secondary buttons in the command bar with an overflow menu. The Action Center dialog is a different shape from that description, so the epic body may need updating to match the current direction.

rjvelazco:
Draft implementation of the Content Drive **Action Center** — a dialog for acting on a multi-item selection.

Opening as a draft: this is a working v1 to demo and discuss with the team, with a few deliberate scope limits documented below.

Fixes #36817

### Proposed Changes

* **`getBulkActions()` on `DotWorkflowsActionsService`** — wraps `POST /api/v1/workflow/contentlet/actions/bulk`, which returns the workflow actions available for a set of contentlets, grouped by scheme and step, each with the number of selected contentlets it applies to. New `DotBulkActionView` models in `dotcms-models`.
* **`DotContentDriveActionCenterComponent`** — the dialog, with two sections:
  * **Quick Actions** — Publish / Unpublish / Archive / Unarchive / Delete, each fired over the whole eligible selection in one request via `POST /api/v1/workflow/actions/default/fire/{systemAction}`. Counts are derived client-side from row state (`live`, `archived`), and an action that applies to nothing is omitted rather than shown as `(0)`.
  * **Workflow Actions** — one collapsible panel per workflow scheme, with real per-action eligibility counts from the backend's Elasticsearch aggregation on `wfstep`. Single-expand; one action selected at a time.
* **Toolbar** — once more than one *contentlet* is selected, the flat action buttons are replaced by a single "Action Center" button. Folders are excluded from the count and from every payload, since the bulk endpoints take contentlet inodes and folders carry no workflow step.
* **Shell** — `DIALOG_TYPE.ACTION_CENTER` is routed out of the shared dialog's content switch and mounted as a sibling, because this dialog owns a custom header and footer (see Additional Info). The store stays the single source of open/close state.
* **i18n** — 19 `content-drive.action-center.*` keys.

### Checklist

- [x] Tests — 37 new (`action-center.spec.ts` for the pure mapping/eligibility logic, plus the component spec). Full portlet suite green: 872/872, lint clean.
- [x] Translations — English keys added to `Language.properties`; other locales not translated.
- [x] Security Implications Contemplated — no new endpoints and no new permissions surface; this consumes existing authenticated REST endpoints. Worth knowing for review: the bulk-actions endpoint filters by **action-level role permission only** and deliberately lets actions with special workflow roles through, because there is no per-contentlet permissionable at aggregation time (`WorkflowActionUtils#filterBulkActions`). So a listed action can still fail per item at fire time, which the backend reports in `fails[]` and the dialog surfaces. Enforcement stays server-side; the counts here are advisory, never an authorization decision.

### Additional Info

**Scope limits, all deliberate:**

* **One action per execute.** No endpoint fires several different actions in one call, and this isn't just a missing endpoint — `workflow_task` has a `UNIQUE (webasset, language_id)` constraint, so a contentlet+language sits on exactly one step, owned by exactly one scheme. Firing action A moves it to a new step, which invalidates the other actions' preconditions. The legacy JSP dialog has the same constraint. A useful consequence: the scheme counts **partition** the selection rather than overlapping.
* **Actions needing extra input are disabled** with a tooltip — push-publish settings, a move target path, or an assign/comment prompt. Wiring those means reusing `DotWorkflowEventHandlerService`; out of scope here.
* **Conditional counts are shown as `≤ N`.** When `conditionPresent` is true the backend does not evaluate the action's Velocity condition while aggregating, so the count is an upper bound.
* **Lock / Unlock and Add to Bundle are omitted** — neither has a bulk REST endpoint. Legacy drives unlock through a Struts command (`full_unlock_list`) that loops server-side, and add-to-bundle through the legacy `RemotePublishAjaxAction` servlet.
* **Fires synchronously** via `bulkFire`. The legacy dialog uses the SSE endpoint (`_bulkfire`) for live progress counters, which is the better path for large batches but needs an SSE shim since native `EventSource` cannot POST a body.
* **System Workflow stays visible.** Hiding it would remove bulk **Copy** outright — Copy has no system-action mapping, so it is only reachable through the workflow list — and would empty the section on Community, where System Workflow is the only scheme that can exist.

**Known follow-ups (backend asks):**

* `_bulklock` / `_bulkunlock` REST endpoints (the `_batchUnlock` logic already exists to port).
* An endpoint returning **eligible inodes per action** — the only thing that unblocks the design's per-item drill-down, which needs to know *which* items an action applies to. The API returns aggregate counts only.
* A REST replacement for the legacy add-to-bundle AJAX action.

**Deviations from the design prototype**, both following from the above: workflow steps use radio semantics rather than checkboxes (multi-select would need sequential fires with re-validation between each), and there are no per-step icon chips (the API returns dotCMS icon names that don't map to Material Symbols).

**Styling note:** the prototype's raw `slate-*` / `#1D1B4B` / `#1D4ED8` values are mapped onto the theme tokens `tailwindcss-primeui` exposes (`surface-*`, `primary-*`), so the dialog follows the active PrimeNG theme instead of pinning one palette. If that navy is meant to be a real brand token, that's a theme-level change rather than something hardcoded here.

### Video

https://github.com/user-attachments/assets/db8dec61-0ba4-4f2c-910c-0c63301098a5

This PR fixes: #36817


rjvelazco:
@claude core review this PR and propose some unit test

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dialogs/dot-content-drive-action-center/dot-content-drive-action-center.component.html
zJaaal, line 1: We already have a dialog for content drive, we can reuse it and just put the content of this one there

core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.html
```diff
@@ -67,6 +67,12 @@
     </ng-template>
 </p-toast>
 
+<!-- Owns its own p-dialog (custom header/footer, body-only scroll), so it renders as a sibling of
+     the shared dialog rather than inside its content switch. -->
+@if ($actionCenterVisible()) {
```
zJaaal: Lets reuse the dialog below, that way we dont have double logic

zJaaal (changes requested):

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dialogs/dot-content-drive-action-center/dot-content-drive-action-center.component.html
rjvelazco: Done!

core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.html
rjvelazco: Done!

zJaaal (approved):

PR #36840, opened 2026-07-31
feat(site-search): allow overriding the index analyzer via SITE_SEARCH_ANALYZER

### LINKED ISSUE #36839: Allow overriding the Site Search index analyzer via config to support CJK/RTL languages
## Description

Site Search hardcodes an English-tuned analyzer chain into its bundled index resources (`es/os-sitesearch-settings.json` + `es/os-sitesearch-mapping.json`): `standard` tokenizer + `lowercase` + `asciifolding` + a hardcoded **English stemmer** on the `content` field, with `title`/`description`/`author` falling back to the plain `standard` analyzer. There is no supported way to change tokenization short of repackaging the webapp.

This makes Site Search a poor fit for non-Latin languages:

- **CJK** text degrades to single-character tokens — a query for 東京 matches any document containing 東 *or* 京 anywhere, so precision and ranking are near-meaningless.
- **RTL** (Arabic/Hebrew) gets no normalization or stemming — `كتاب` does not match `الكتاب` — and the English stemmer/asciifolding actively work against non-Latin text.

Elasticsearch/OpenSearch already ship ~35 built-in language analyzers (`cjk`, `arabic`, `thai`, ...) that are referenced by name with no plugin or settings declaration, and plugins add dictionary-grade ones (`kuromoji`, `nori`, `smartcn`). The gap is purely that Site Search offers no hook to select one.

**Proposal:** a single `SITE_SEARCH_ANALYZER` Config property (env: `DOT_SITE_SEARCH_ANALYZER`) naming any analyzer known to the cluster. When set, it is applied to the text fields (`content`, `title`, `description`, `author`) of the site-search mapping at index creation, for both the Elasticsearch and OpenSearch providers. Unset keeps the bundled defaults byte-for-byte. Deliberately **not** a settings/mapping file override — one env var covers every named analyzer without customers authoring engine JSON.

## Acceptance Criteria

- [ ] Setting `DOT_SITE_SEARCH_ANALYZER` (e.g. `cjk`, `arabic`) applies that analyzer to the `content`, `title`, `description`, and `author` fields of newly created Site Search indexes on both the ES and OS providers
- [ ] The `content.ngram` subfield's `search_analyzer` follows the override so query-time tokenization stays consistent
- [ ] With the property unset, index creation behavior and the bundled settings/mapping are unchanged
- [ ] An analyzer name unknown to the cluster fails index creation loudly with the engine's error, rather than silently building an index with default analyzers
- [ ] The override takes effect at index creation (existing indexes require a rebuild); documented in the code
- [ ] Unit tests cover the default (untouched) path, the override applied to text fields only, and missing-resource failure

## Priority

Medium

## Additional Context

- Requested by a partner needing best-practice guidance for RTL and CJK site search.
- Built-in analyzers (`cjk`, `arabic`, `thai`, ...) require nothing on the cluster; plugin analyzers (`kuromoji`, `nori`, `smartcn`) additionally require the plugin installed on every node of the ES/OS cluster.
- Known limitation (accepted): one analyzer per index — mixed-language sites pick a single tokenization. Per-language subfields/indexes are out of scope until a customer needs them.

swicken:
### Proposed Changes
* Add a `SITE_SEARCH_ANALYZER` Config property (env: `DOT_SITE_SEARCH_ANALYZER`) naming any analyzer known to the search cluster — built-in language analyzers (`cjk`, `arabic`, `thai`, ...) or plugin-provided ones (`kuromoji`, `nori`, `smartcn`). When set, it is applied at index creation to the text fields of the site-search mapping (`content`, `content_raw`, `title`, `description`, `author`) on both the Elasticsearch and OpenSearch providers, and the `content.ngram` subfield's `search_analyzer` follows it so query-time tokenization stays consistent. Unset keeps the bundled defaults byte-for-byte.
* Declare `content_raw` explicitly in both bundled mappings. The publishers add it to every `text/*` document but the mapping never declared it, so it was dynamically mapped with the `standard` analyzer — and since site-search queries use `query_string` with `default_field:"*"`, it silently reintroduced CJK unigram false positives even with the override applied. Declared type matches the previous dynamic mapping, so the unset path is unchanged.
* Centralize settings/mapping resource loading in `SiteSearchIndexResources` (also fixes the ES path reading resources via `new File(url.getPath())`, which breaks when packaged inside a JAR), with a loud `DotSearchException` on missing resources or a failed override instead of silently building an index with default analyzers.

### Why
Site Search hardcodes an English-tuned analyzer chain (standard tokenizer + lowercase + asciifolding + English stemmer). CJK text degrades to single-character tokens (a query for 東京 matches any document containing 東 or 京 anywhere) and RTL languages get no normalization or stemming (كتاب does not match الكتاب). Resolves #36839.

### Checklist
- [x] Tests
- [ ] Translations (n/a — no user-facing strings)
- [x] Security Implications Contemplated (analyzer name is operator-controlled config, inserted via the JSON API — no injection path; value is trimmed and failures are loud)

### Additional Info
**Verified end-to-end against a live stack**, not just unit tests: built the branch image, booted dotCMS + OpenSearch with `DOT_SITE_SEARCH_ANALYZER=cjk`, published Japanese HTML file assets, and ran a real run-now Site Search job:
* The dotCMS-created index mapping carries `cjk` on all text fields, including `content_raw`.
* Query 東京 (Tokyo) via the exact `query_string`/`default_field:"*"` shape Site Search uses returns **only** the document about Tokyo Tower; a control document containing 東 and 京 in unrelated words no longer matches (it did on the default mapping, at an identical score).
* Control queries 建物 and 静か each return exactly their one matching document.
* A misconfigured analyzer name fails index creation with the engine's `analyzer [x] has not been configured in mappings` error.

**Scope notes:**
* One analyzer per index (a mixed-language site picks one tokenization) — per-language subfields/indices are deliberately out of scope until a customer needs them.
* Plugin analyzers (`kuromoji`/`nori`/`smartcn`/`icu`) additionally require the plugin installed on every node of the ES/OS cluster; built-ins need nothing.
* `content_raw` (raw un-stripped HTML, no readers anywhere) roughly 2–5×'s the indexed text per document; making it `index: false` is a good follow-up but changes default-path search behavior, so it is not part of this PR.

This PR fixes: #36839


jcastro-dotcms (dismissed):

swicken:
Re the two non-blocking notes:

1. **Key order on the set path** — agreed and intended; the byte-for-byte guarantee is only claimed (and only matters) for the unset path. Mapping property order is not significant to either engine.

2. **`content.ngram` index/query divergence** — verified: nothing in the codebase queries `content.ngram` (it is reachable only via the raw-JSON query path; grep shows no reader). The divergence is also deliberate defense, not an accident: with `search_analyzer` following the override, a CJK query produces bigrams that cannot match the standard-tokenized edge-ngram index, so the subfield is inert under `query_string(default_field:"*")`. Leaving its `search_analyzer` at `standard_content` would decompose CJK queries into unigrams that DO match single-character edge-grams — reintroducing the same false-positive side door this PR closes for `content_raw`, through a third field. The E2E result (東京 returning only the genuinely matching doc across all fields) covers this path.

dario-daza (approved):

Issue #36658, opened 2026-07-21
Block Editor: lossless Markdown round-trip for all stored content (dotcms-* fences for rich nodes, custom-node fallback, table attrs, marks) on the fire/save path


hassandotcms:
## Description

Since #36002 (PR #36253), the workflow fire endpoints convert a Block Editor field supplied as Markdown into Tiptap/ProseMirror JSON server-side. But Markdown has no syntax for the rich nodes the editor persists (`dotContent` embedded contentlets, `dotVideo`, `youtube`, `gridBlock`, `aiContent`, or customer-supplied custom blocks), so today:

- A Markdown write to a document containing rich blocks is **ignored** (existing document kept, warning logged). The save returns 200 and the caller has no signal that nothing was applied.
- The overwrite guard's whitelist is missing `youtube` and `image`, so a single YouTube embed wrongly blocks all Markdown writes to that document.
- Constructs the guard does not inspect are lost silently: merged table cells (`colspan`/`rowspan`/`colwidth`), text alignment, underline/sub/sup, link `target`/`rel`.
- The JSON-to-Markdown direction flattens rich nodes to plain text (`dotContent` emits its title), so a `JSON -> MD -> JSON` round trip destroys them.

This blocks non-interactive clients (AI agents via MCP, headless imports) from updating any content that contains an embed, which is exactly the content Markdown-driven agents are expected to maintain.

### What we want

Give every rich node a Markdown form so the round trip is lossless, using **fenced code blocks with `dotcms-*` info strings**. This is standard CommonMark, preserved verbatim by every parser and by LLMs, and parseable with the fence handling `TiptapMarkdown` already has:

    ```dotcms-content
    {"identifier": "2d5d1c4c-...", "languageId": 1, "title": "Q3 Report"}
    ```

- **Thin payloads**: `dotContent`/`dotImage`/`dotVideo` need only `identifier` (plus `languageId`); the server already rehydrates `attrs.data` from the identifier on every read (`StoryBlockAPI.allowedTypes`). `title` is emitted for readability and ignored on parse. `gridBlock` carries its node subtree verbatim (children preserved, not Markdown-editable). Unknown/custom node types (remote `customBlocks` extensions) are carried by a generic `dotcms-node` fence with the raw node JSON.
[...]

PR #36841, opened 2026-07-31
36659 feat(block-editor): dotcms-* rich-node vocabulary on HTML input (parity with the Markdown leg)

### LINKED ISSUE #36659: TiptapHtml: accept the same rich-node vocabulary on HTML input (parity with the Markdown leg)
## Description

#36470 (PR #36473) made the save path convert Block Editor HTML input to Tiptap JSON server-side via `TiptapHtml`, deliberately clamped to the same primitive subset as the Markdown leg. Once the Markdown leg gains rich-node support (#36658), HTML input becomes the lossier of the two: a client sending HTML cannot express an embedded contentlet, video, YouTube embed, or grid, and content that contains them still rejects HTML writes.

### What we want

Extend `TiptapHtml` so HTML input can express the same rich-node vocabulary, keeping its sanitization contract intact. Proposed markup: explicit, namespaced elements mirroring the Markdown fences (exact form to be settled in the PR):

    <dotcms-content identifier="2d5d1c4c-..." language-id="1"></dotcms-content>
    <dotcms-video src="https://.../clip.mp4" mime-type="video/mp4"></dotcms-video>

- Same thin-payload rule as the Markdown leg: asset-bound nodes carry only identifier/src plus essentials; the server rehydrates `attrs.data` on read.
- Same trust model: attribute allow-list per element, hard-coded node type per element, size caps. Unknown or invalid `dotcms-*` elements degrade gracefully (dropped or reduced to their text content) and never throw.
- The existing sanitization contract must not widen otherwise: the `DROP_SUBTREE` set, URL scheme allow-list, and attribute-breakout rejection stay exactly as shipped; the new elements are additions to the allow-list, not a relaxation.
- The overwrite guard treats HTML-expressible documents the same as Markdown-expressible ones; response messages on ignored/degraded writes match the Markdown leg.

### Out of scope

- A JSON-to-HTML server serializer (there is none today; `toHtml()` remains the render-time Velocity path).
- Marks/alignment beyond what the Markdown leg supports.

### Related work

- #36470 / PR #36473: TiptapHtml converter (jsoup walker, sanitization contract)
- #36658: Markdown rich-node vocabulary (this issue mirrors its frozen vocabulary and depends on it)

## Acceptance Criteria

- [ ] Firing with HTML containing each `dotcms-*` element stores a valid Tiptap document; embedded contentlets read back fully hydrated with no human re-save.
- [ ] HTML and Markdown inputs produce equivalent Tiptap JSON for the same rich content (parity test suite shared with the Markdown leg).
- [ ] Sanitization regression suite passes unchanged: script/style/iframe subtrees dropped, `javascript:`/`data:` URLs rejected, no event-handler attributes survive, including inside the new `dotcms-*` elements.
- [ ] Invalid or unknown `dotcms-*` elements degrade gracefully; conversion never throws.
- [ ] A document containing supported rich nodes accepts HTML writes; ignored/degraded writes carry a response message (no new rejections, no status-code changes).
- [ ] Server-converted JSON opens and re-saves unchanged in the Block Editor (golden test).

hassandotcms:
### What

  Closes #36659. HTML input now expresses the same rich nodes the Markdown leg gained in #36658 (dotContent, dotImage, dotVideo, youtube, aiContent, gridBlock, custom blocks), as namespaced elements — HTML is no longer the lossier save-path format.

      <dotcms-content identifier="2d5d1c4c-…" language-id="1"></dotcms-content>
      <dotcms-grid>{"type":"gridBlock","attrs":{"columns":[6,6]},"content":[…]}</dotcms-grid>
      

  - **Markup form:** the seven fence labels as element names. Scalars ride as attributes, hyphenated because HTML lowercases attribute names (`language-id` → `languageId`); the verbatim labels and `dotcms-ai` take the fence-identical JSON as element body (attributes can't carry a subtree, and JSON in an attribute corrupts on the first quote). The editor's own HTML was rejected as a model: `dotContent` has no `parseHTML`, and `dotVideo`/`youtube` parse from `<video>`/`<iframe>` — both in `DROP_SUBTREE`, which stays.
  
  - **Shared validator:** both carriers normalize into the same `TiptapMarkdown.DotcmsFences` (now package-visible), so parity holds by construction — same schema, node type, 64KB/depth caps, `doc` rejection.
  
  - **Degrade, never throw:** invalid body labels keep their payload as a visible `codeBlock` like an invalid fence; scalar labels drop; unknown `dotcms-*` stays transparent.
  
  - Sanitization does not widen: no blanket attribute copy, every URL through `sanitizeUrl`.

  ### Behavior change

  Two narrow routing fixes in `MapToContentletPopulator`: a value opening with `<dotcms-*>` used to store as literal text (`HTML_START` misses hyphenated tags) and now routes to HTML via a `<dotcms-` carve-out; and routing is decided on what *follows* leading `dotcms:attrs` comments, so comment-then-HTML reaches the HTML converter instead of losing the decoration. Stored JSON still passes through untouched; same status codes, no new rejections.

  ### Verification

  204 unit (51 new, incl. HTML≡Markdown parity suite + hostile attrs on the new elements; existing sanitization suite unchanged), 20 integration (5 new). Live pass: the HTML-created doc opens in the Block Editor with every rich node rendering, takes a real edit, and re-saves intact; the Markdown-fence sibling renders identically. OpenAPI + MCP guidance updated.
  
  

This PR fixes: #36659


fabrizzio-dotCMS (commented):
## Code review

The design holds up well: one grammar validated by a single `DotcmsFences`, so parity is enforced by construction rather than by duplicated logic. The routing carve-out is tightly scoped (only the seven known labels; `<dotcms-block-editor>` keeps its current route), and the test suite is genuinely thorough — the HTML≡Markdown parity block and the hostile-input cases are the right shape for this change.

Three things worth addressing, none of them merge blockers.

---

### 🟡 Important

#### 1. A `dotcms:attrs` decoration leaks onto the next block when a scalar element fails validation

`dotCMS/src/main/java/com/dotcms/tiptap/TiptapHtml.java:310-326`

In the `DOTCMS_ATTR_LABELS` branch, `st.takePendingAttrs()` is only called on the `node != null` path. On the `else` (validation failed) path the pending decoration stays live:

```html
<!-- dotcms:attrs {"textAlign":"center"} -->
<dotcms-content language-id="1"></dotcms-content>   <!-- no identifier -> dropped -->
<p>Unrelated paragraph</p>                           <!-- <- receives textAlign:center -->
```

This breaks the invariant the PR itself states ("a pending decoration is consumed but never applied to a rich node"). The `DOTCMS_BODY_LABELS` branch does consume it on failure via `decorate(code, st)` (line 342), and the Markdown leg consumes it on the degraded fence. Only this path skips it. `decoration_before_rich_element_is_consumed_not_applied` uses a **valid** element, so it takes the `node != null` branch and does not cover this.

Fix: call `st.takePendingAttrs();` in the `else` branch too.

#### 2. `<dotcms-node>` / `<dotcms-grid>` bypass the sanitization contract the class promises

`dotCMS/src/main/java/com/dotcms/tiptap/TiptapHtml.java:328-355` + `TiptapMarkdown.DotcmsFences#parseVerbatim`

The `TiptapHtml` class javadoc and the PR description both state *"no blanket attribute copy, every URL through `sanitizeUrl`"*. That does not hold for the body labels: `isPlausibleNodeTree` validates **shape** (textual `type`, no `doc`, depth ≤ 64) but never inspects `attrs`, and the payload is stored with `deepCopy()` as-is. So this passes:

```html
<dotcms-node>{"type":"dotImage","attrs":{"src":"/dA/x.jpg","href":"javascript:alert(1)"}}</dotcms-node>
```

while `javascript_href_on_image_is_omitted_but_safe_src_kept` shows that blocking exactly this was the intent. And the sink is live: `storyblock/dotImage.vtl` emits `<a href="$!item.attrs.href">` and `alt="$!item.attrs.alt"` unescaped, and `render.vtl` emits `style="text-align: $element.attrs.textAlign;"` unescaped (only the `table` branch uses `$esc.html`) — a stored-XSS path on the front-end render.

To be fair about scope: **this PR does not introduce the exposure at the product level.** The same endpoint has accepted the ```` ```dotcms-node ```` fence since #36658, and a plain `` already reached the renderer unescaped. What this PR does add is (a) extending the bypass to the HTML carrier, which previously could never emit an unsanitized URL, and (b) asserting in the javadoc and the PR description that sanitization does not widen — which is no longer accurate.

Suggestion: run URL-bearing fields of verbatim payloads through `sanitizeUrl` (or reject `javascript:`/`data:` schemes inside `isPlausibleNodeTree`), and soften the javadoc so it does not promise a guarantee the body labels do not provide. Hardening the VTL output encoding deserves its own issue — it is pre-existing and broader than this change.

---

### 🔵 Quality

**3. Decoration silently lost when the decorated node is discarded** — `TiptapHtml.java:495-514`, `:517-527`, `:531-545`. `decorate()` runs *before* it is known whether the node survives, so an empty `<h2></h2>`, an empty `<blockquote></blockquote>`, or a `<ul></ul>` with no items consumes the pending decoration and is then dropped — the decoration evaporates. Same root cause as #1: the pending-decoration lifecycle is spread across six call sites instead of being centralized in one place that knows whether the node was emitted.

**4. `dotcms-ai` is missing from the label list** — `WorkflowResource.java` (`BLOCK_EDITOR_FIELD_NOTE`, ~line 264). The new HTML section names it as a body label, but it has no bullet in the "supported labels and payload fields" list in either the Markdown or the HTML section, so neither a human nor an AI agent can discover that its payload is `{"content": "<string>"}` (per `DotcmsFences#parseAi`). Since this PR rewrites that block across seven endpoints, it is the natural place to close the gap.

**5. `ContentResource` → `WorkflowResource` coupling** — `ContentResource.java:214` references `com.dotcms.rest.api.v1.workflow.WorkflowResource.BLOCK_EDITOR_FIELD_NOTE` as an inline fully-qualified name. One REST resource importing a constant from another is a mild smell; a shared documentation-constants holder would be cleaner and would drop the embedded FQN.

---

### Checked, no findings

- **Routing in `MapToContentletPopulator`**: `HTML_START` (`^<(!|/?[a-zA-Z][a-zA-Z0-9]*[\s/>])`) genuinely cannot match hyphenated tags, so the carve-out is necessary and correctly scoped. The unterminated-comment fallback to the Markdown route is correct and tested. No value that routes one way today changes route, apart from the documented behavior change (leading decoration comment + HTML now reaches the HTML converter instead of storing as literal text).
- **Size/depth caps**: body labels go through the `parse(String, String)` overload, so the 64KB cap and depth-64 guard apply exactly as they do to fences. The new `parse(String, JsonNode)` overload skips the byte cap, but only the scalar labels use it and their payloads are flat objects of scalars.
- **ReDoS**: `DOTCMS_ELEMENT_START` and `DOTCMS_ATTRS_COMMENT_DATA` are simple alternation / single greedy quantifier — linear.
- **Degrade-never-throw**: every new path wraps in `catch (Exception)` returning `null`.
- **OpenAPI**: the seven Java usages map 1:1 to the seven changed description blocks in the yaml; no `operationId` churn, no `@Schema` changes.
- **Test registration**: `StoryBlockMarkdownPopulatorTest` is pre-existing and already registered in `MainSuite1b`; `TiptapHtmlRichNodeTest` is a plain unit test and correctly needs no suite.
- **`<p><dotcms-content/></p>` degrading to text** is deliberate and tested (`inline_position_degrades_to_text`). Only flagging it because an LLM emitting HTML wraps things in `<p>` quite naturally and the loss is silent — it does not surface in `messages`. Worth confirming whether that should warn.

fabrizzio-dotCMS (approved):
Approving. The findings in my review comment are non-blocking; the pending-decoration leak in the scalar-label failure path (`TiptapHtml.java:310-326`) is a one-line fix worth folding in before merge.

PR #36842, opened 2026-07-31
ci(autodoc): route Claude execution through ai-workflows to fix Bedrock OIDC

### LINKED ISSUE #36949: ci(autodoc): fix Bedrock OIDC and harden autodoc workflow
## Summary

The autodoc workflow (`issue_autodoc.yml`) was failing at `AssumeRoleWithWebIdentity` in production with a `Not authorized to perform sts:AssumeRoleWithWebIdentity` error that never appeared on `core-workflow-test`. Root cause: the IAM trust policy on `GitHubActions-BedrockCodeReview` gates on the OIDC `job_workflow_ref` claim, which only resolves correctly when the job is invoked via a reusable workflow — not when it runs inline.

## Changes

- **Restructure workflow into a multi-job pipeline**: `preflight → prep → claude → finalize`. The `claude` job calls `dotCMS/ai-workflows/.github/workflows/autodoc-executor.yml` as a reusable workflow, ensuring the `job_workflow_ref` OIDC claim resolves to `dotCMS/ai-workflows/...@refs/tags/*` and satisfies the Bedrock role trust condition.
- **modUser guardrail**: skip applying a draft when a human (not the CI Autodoc service account) has an outstanding working draft on the content, to avoid overwriting human edits.
- **Pin action SHAs** to address Semgrep supply-chain findings.
- **Replace `secrets: inherit` with explicit secret map** for least-privilege.
- **Move `AUTODOC_DOTCMS_SITE_FOLDER` and `AUTODOC_DOTCMS_BASE_URL` from Secrets to Variables** — neither is a credential; treating them as secrets obscures config unnecessarily.

## Related

- `dotCMS/ai-workflows#65` (merged) — introduces `autodoc-executor.yml` reusable workflow
- Bedrock IAM role: `GitHubActions-BedrockCodeReview`

jdcmsd:
relates to: https://github.com/dotCMS/ai-workflows/pull/65

## Summary

- Restructures the `autodoc` job into `prep → claude → finalize` so Claude/Bedrock runs inside `dotCMS/ai-workflows/autodoc-executor.yml`
- This sets `job_workflow_ref` to the `ai-workflows` ref, satisfying the trust condition on `GitHubActions-BedrockCodeReview` that was blocking OIDC assumption with `Not authorized to perform sts:AssumeRoleWithWebIdentity`
- `prep` and `finalize` jobs run directly in this workflow and do not touch AWS
- Report path standardised to `/tmp/autodoc-report.md`; passed between jobs via Actions artifact

## Root cause

The IAM role's trust policy requires `job_workflow_ref` to match `dotCMS/ai-workflows/.github/workflows/*@refs/tags/*`. When the Claude step ran directly inside `issue_autodoc.yml`, `job_workflow_ref` was `dotCMS/core/.github/workflows/issue_autodoc.yml@...` — a hard deny. Routing through the `ai-workflows` reusable workflow fixes the claim.

## Dependency

**Merge `dotCMS/ai-workflows#65` and tag `v3.5.0` before merging this PR.** The `claude` job references `autodoc-executor.yml@v3.5.0`.

## Test plan

- [x] Confirm `dotCMS/ai-workflows#65` is merged and tagged `v3.5.0`
- [ ] Merge this PR
- [ ] Close a test Epic and confirm: OIDC succeeds, Claude runs, artifact uploaded, comment posted on issue, dotCMS draft saved

This PR fixes: #36949


.github/workflows/issue_autodoc.yml
```diff
@@ -0,0 +1,480 @@
...
+          fi
+
+          # Parse action and urlTitle from the machine-readable meta block and export
+          # for the downstream apply steps.
+          ACTION=$(awk '/BEGIN_DOC_META/{p=1;next} /END_DOC_META/{p=0} p && /^action:/{print $2; exit}' "$REPORT")
+          URL_TITLE=$(awk '/BEGIN_DOC_META/{p=1;next} /END_DOC_META/{p=0} p && /^urlTitle:/{sub(/^urlTitle:[[:space:]]*/,"");print;exit}' "$REPORT")
+          echo "ACTION=${ACTION:-none}" >> "$GITHUB_ENV"
+          echo "URL_TITLE=$URL_TITLE"   >> "$GITHUB_ENV"
+
+      - name: Apply doc update to dotCMS
```
sfreudenthaler, line 406: Addressed in the latest commit — added a modUser guardrail before the EDIT fire. Logic: fetch the working inode first (GET /api/v1/content/{id} always returns working), then proceed only if live==true (no pending draft) or modUser == the CI Autodoc service account (user-3244a6be-d28b-4d80-a3fc-9f66a1d832ae). If a human has an outstanding draft, the step skips with a clear log message pointing to dotCMS admin for manual review.

.github/workflows/issue_autodoc.yml
```diff
@@ -0,0 +1,480 @@
...
+          draft = report.split(marker, 1)[1].lstrip('\n') if marker in report else ''
+          payload = {'contentlet': {
+              'identifier': os.environ['IDENTIFIER'],
+              'languageId': 1, 'documentation': draft,
+              'disabledWYSIWYG': ['documentation']}}
+          json.dump(payload, open('/tmp/payload.json', 'w'))
+          PYEOF
+
+          curl -s --fail-with-body -X PUT \
+            "$AUTODOC_DOTCMS_BASE_URL/api/v1/workflow/actions/default/fire/EDIT?identifier=$IDENTIFIER" \
```
sfreudenthaler, line 472: No publish gate needed here — this workflow never fires PUBLISH. EDIT saves a working draft; NEW creates a new contentlet in draft state. Nothing goes live without a human opening the dotCMS admin UI and explicitly publishing. This is the same draft-first default as apply_doc.py (the --publish flag is required to go live locally).

sfreudenthaler (commented):
These look good overall — nice work routing through ai-workflows to fix the OIDC trust issue, and the finalize step follows the same draft-then-apply pattern we landed for the changelog site publisher. Two things before this is ready to merge:

1. No human-edit guardrail on the update path (see inline comment below). My changelog publisher checks `modUser` on the existing content and skips (with a visible skip marker) if the last edit wasn't made by the service account, specifically so the automation never clobbers a manual edit made in the dotCMS backend. This workflow's "Apply doc update to dotCMS" step has no equivalent check — it will overwrite the `documentation` field unconditionally, and there's no log line or alert anywhere that would tell us it happened. Given Jamie (or anyone) could be hand-editing these pages between Epic closures, I'd want that guardrail before this runs against real content.

2. Publish-gate question (also inline below): my changelog automation auto-publishes with no approval gate, but that's safe *because* the content is deterministic — the GitHub release body IS the site content, mechanically transformed. Here Claude is generating original documentation prose, so there's real hallucination risk. Before this merges I want to confirm what `AUTODOC_DOTCMS_BASE_URL` points at and what the default `EDIT`/`NEW` workflow actions actually do on the `DotcmsDocumentation` content type — does firing them here land the page in Draft/Review, or does it go straight to Published with no human in the loop? If it's the latter, we need either a review gate before publish, or to restrict this to a staging/review instance until there's human sign-off on the draft.

.github/workflows/issue_autodoc.yml
jdcmsd: I like the idea, though I think a little extra nuance is called for. If it'll only edit docs last edited by a service user, then it's undermining its own core function, since human edits are the sole source of publish actions by design; using the same logic here would run into a bootstrapping problem. So, here's what I decided: 

- If the automation finds an existing document in a published/live state, then it shouldn't matter who last edited it, because all it's doing is saving a new draft without destroying any human edits;
- if it finds a draft document that the service account last edited, then same;
- if it finds that an existing document is specifically in a working/draft state last saved by someone other than the service account, THEN it skips with a warning and requests a manual review.

.github/workflows/issue_autodoc.yml
jdcmsd: Yeah, EDIT leaves it as a draft/review; actions/default/fire/PUBLISH would be required to get the content version live. Publishing has been inveighed against through the entire design process, and also is expressly disallowed to the service account, which operates on a strict "draft only" role that I made for it.

image

.github/workflows/issue_autodoc.yml
```diff
@@ -0,0 +1,501 @@
...
+      EPIC_NUMBER: ${{ github.event.issue.number }}
+      REPO: ${{ github.repository }}
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      AUTODOC_DOTCMS_API_TOKEN_DRAFTING: ${{ secrets.AUTODOC_DOTCMS_API_TOKEN_DRAFTING }}
+      AUTODOC_DOTCMS_BASE_URL: ${{ secrets.AUTODOC_DOTCMS_BASE_URL }}
+      AUTODOC_DOTCMS_SITE_FOLDER: ${{ vars.AUTODOC_DOTCMS_SITE_FOLDER }}
+
+    steps:
+      - name: Download report artifact
+        uses: actions/download-artifact@v4
```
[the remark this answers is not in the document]
jdcmsd: Added SHA!

.github/workflows/issue_autodoc.yml
```diff
@@ -0,0 +1,501 @@
...
+            echo 'content<<__AUTODOC_EOF__'
+            cat /tmp/eval_context.md
+            echo '__AUTODOC_EOF__'
+          } >> "$GITHUB_OUTPUT"
+
+  claude:
+    name: Run Claude (Bedrock)
+    needs: prep
+    uses: dotCMS/ai-workflows/.github/workflows/autodoc-executor.yml@v3.5.0
+    secrets: inherit
```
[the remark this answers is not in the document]
jdcmsd: Replaced with explicit map of the 3 secrets autodoc-executor.yml actually declares

jdcmsd:
Flurry of activity above is just a rebase to resolve a conflict. I think this is the same branch as merged the autodoc automation in originally; if we'd cut a new branch, it would have avoided the need for this step.

jdcmsd:
Last push to attempts to fix the issues [raised above](https://github.com/dotCMS/core/pull/36842#issuecomment-5219421330).

sfreudenthaler:
> * Medium: `.github/workflows/issue_autodoc.yml:324-326` — **verify the secrets map is complete.**

Should be good here.  We dropped down to two secrets and moved the other one over to input in dotcms/ai-workflows#85

sfreudenthaler (approved):
LGTM. Fixed the set -e dead-code issue in the content-check curl guard (`.github/workflows/issue_autodoc.yml:437-443`) — moved the failure check onto the assignment itself with `if ! CONTENT_CHECK=$(...)` so the SKIP message actually fires on a failed curl instead of the step just going red before the guard runs.

PR #36843, opened 2026-07-31
Migrate Analytics Dashboard to domain-driven query API (Angular changes)

### LINKED ISSUE #36628: Implement domain-driven query REST API for Analytics Dashboard
## Description

Implement the domain-driven query REST API redesign for the Analytics Dashboard, following the design produced in the prior spike (#35736 — closed).

**Background:** The current analytics query endpoints encode pre-computed metrics as URL path segments (`/total-events`, `/top-content`, `/pageviews-by-device-browser`, `/unique-visitors`, `/engagement`, etc.) across the Analytics Events, Analytics Conversions, and Analytics Sessions resource groups. This bakes the metric, grouping, and aggregation into the URL, so every new dashboard chart requires a new endpoint and customer questions outside the pre-defined set (e.g. "clicks by country") can't be answered.

**Goal:** Replace these endpoints with a resource-oriented query API (resource + dimensions + metrics + filters) so new questions can be answered without adding new endpoints. **The new API must return the exact same underlying information the current endpoints return — only reshaped into a better, more consistent JSON response.** No new metrics or data are in scope; this is a contract/response redesign, not a data/feature change.

**Design source of truth:** [Analytics Query REST API Redesign — Google Doc](https://docs.google.com/document/d/1BeW8zIuUgUI5gMLkKzeY_LVjeI5VTfaoiv7tRspNwzo/edit?tab=t.0). Implement per the endpoint(s), query contract, and response schema defined there.

**Scope:**
- Backend: new query endpoint(s) implementing the resource + dimensions + metrics + filters model
- Frontend: update the Analytics Dashboard to call the new endpoint(s) and render from the new response shape
- Out of scope: ingestion (POST) endpoints — these are working correctly and must not be modified

**Cutover:** Hard cutover. The existing per-metric query endpoints are removed and replaced — no versioning or backward-compatibility layer, per the spike's conclusion.

## Acceptance Criteria

- [x] New domain-driven query REST API endpoint(s) implemented per the design in the linked Google Doc (resource + dimensions + metrics + filters model)
- [x] New API returns the exact same underlying data/information as the current endpoints (total-events, top-content, pageviews-by-device-browser, unique-visitors, engagement), only reshaped into the improved JSON response format defined in the design doc
- [x] Old per-metric query endpoints are removed (hard cutover) — no versioning or backward-compatibility layer added
- [x] Ingestion (POST) endpoints for events/conversions/sessions are unaffected and remain unchanged
- [x] Analytics Dashboard frontend updated to call the new endpoint(s) and consume the new response shape, producing the same charts/metrics as before (total events, top content, pageviews by device/browser, unique visitors, engagement)
- [x] Integration tests added/updated covering the new query endpoint(s) for each dashboard metric currently supported
- [x] `openapi.yaml` regenerated (via `@Operation`/`@Parameter` annotations, not hand-edited) to reflect the new endpoint(s) and the removal of the deprecated ones

## Priority

High

## Additional Context

- Related spike (research/design, closed): #35736 — "Spike: Redesign Analytics Query REST API"
- Design detail (contract, dimensions, metrics, filters, response schema, example queries): [Google Doc](https://docs.google.com/document/d/1BeW8zIuUgUI5gMLkKzeY_LVjeI5VTfaoiv7tRspNwzo/edit?tab=t.0)
- **Breaking change:** removing the old query endpoints is a public REST API contract change. Confirm with the Falcon team whether release notes / upgrade documentation are needed for any external consumers of these query endpoints (dashboard is first-party, but the endpoints are public REST API).

jcastro-dotcms:
### Proposed Changes
* Migrate all 7 `DotAnalyticsService` methods off the old per-metric endpoints (`/event/*`, `/session/*`, `/conversion/content/attribution`) onto the new domain-driven query API (`/analytics/events`, `/analytics/sessions`, `/analytics/content`), mapping each response back to the existing DTO shapes so no store/component code needed to change downstream
* Remove `getConversionsOverview()` and the "Conversions Overview" table widget (component, store slice, types, i18n keys) from the Conversions tab entirely — the backend never built a domain-driven replacement for `/analytics/conversion`, since Product decided to drop that resource
* Fix the dotCMS Java proxy allowlist (`EventAnalyticsProxyHelper`) to permit the new `events`/`sessions`/`content` path prefixes — without this, calls to the new endpoints 400 before reaching the upstream service
* Change the dashboard's default landing tab from Pageview to Engagement (product decision)

### Additional Changes (Not Part of the Original Ticket Requirements)
dotCMS/core#36628 is scoped to the REST API migration itself. The two items below were separate asks/fixes that came up during implementation and review of this PR — flagging them explicitly since they go beyond that ticket's requirements:
* Add a full-report loading overlay shown while a tab's data is loading, matching the look of `dot-edit-content-layout`'s reload overlay
* Fix `analyticsHealthGuard` from `canMatch` to `canActivate` — `canMatch` was re-firing the `/api/v1/analytics/health` check (and blocking navigation on its response) on every single tab click; `canActivate` only fires when entering/re-entering the Analytics Dashboard portlet, not on internal tab switches

### Checklist
- [x] Tests
- [ ] Translations

### Additional Info
Related backend work (separate repo): dotCMS/dot-ca-event-manager#92, following the design/migration guide in that repo's `docs/analytics-frontend-migration.md`.

Migration was done gradually, one endpoint at a time: for each method, the exact outgoing request from the Angular dev build was captured and replayed against the old endpoint in Postman to diff the values before moving to the next.

**Security note:** the `EventAnalyticsProxyHelper` allowlist change only adds 3 new literal path-prefix strings (`events`, `sessions`, `content`) to an existing exact-match/prefix allowlist — it does not loosen the matching logic itself. Covered by existing + new unit tests in `EventAnalyticsProxyHelperTest`.

**Translations:** no new user-facing strings were added; only unused i18n keys tied to the removed Conversions Overview widget were deleted from the default `Language.properties` (no other locale files had these keys). The new loading overlay's `aria-label` reuses the existing generic `Loading` key.

**Review follow-ups incorporated:** dead-code cleanup (unused constant/type, dead store slice), a shared `#queryEventsMetric` helper to de-duplicate the aggregate/series query pattern, trimmed migration-era JSDoc down to enduring facts, guard redirect switched to returning a `UrlTree` instead of `router.navigate()` + `false`, a11y attributes on the loading overlay (`role="status"`, `aria-live="polite"`, `aria-hidden` on the decorative spinner), and moving the overlay's styling from custom SCSS to Tailwind utility classes per `STYLING_STANDARDS.md`.

### Screenshots
No visual changes to chart/table rendering. User-visible changes are: default tab is now Engagement instead of Pageview, the Conversions tab no longer shows the "Conversions Overview" table, and a loading spinner overlay now appears briefly while a tab's data loads.

This PR fixes: #36628


core-web/libs/portlets/dot-analytics/portlet/src/lib/guards/analytics-health.guard.ts
```diff
@@ -8,8 +8,16 @@ import { DotAnalyticsService } from '@dotcms/portlets/dot-analytics/data-access'
...
+ * That fired a fresh `/api/v1/analytics/health` request (and blocked navigation on its response)
+ * on every single tab click. `canActivate` only runs when the guarded route is newly activated —
+ * once when entering `/dashboard` (or `/search`) from outside it, not on every child-only
+ * navigation within it.
  */
-export const analyticsHealthGuard: CanMatchFn = (_route, _segments) => {
+export const analyticsHealthGuard: CanActivateFn = (_route, _state) => {
     const analyticsService = inject(DotAnalyticsService);
     const router = inject(Router);
     const activatedRoute = inject(ActivatedRoute);
```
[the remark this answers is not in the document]
jcastro-dotcms: Good catch — you're right. `isEnterprise` is resolved on the parent `analytics` route (via `DotEnterpriseLicenseResolver` in `app.routes.ts`) and Angular merges resolved `data` down the ancestor chain unconditionally, so it's already present on this guard's own `route` snapshot param. Injecting `ActivatedRoute` instead risked resolving to whatever route was active *before* this navigation, since the target route's own instance isn't created until after guards pass — exactly the case that matters most now that this guard runs on `canActivate` (cross-portlet entry) rather than `canMatch`. Fixed in 4ce129ab29: reading from the snapshot param, dropped the now-unused `ActivatedRoute` injection, and updated the guard's tests accordingly.

jcastro-dotcms:
Thanks for the review — addressed both new findings:

1. **`$isReportLoading` stuck forever with no site** — confirmed, this was a real regression. `isUnsettled` now only treats `LOADING` as unsettled (not `INIT`), so the no-site branches in `loadAllPageviewData`/`loadConversionsData`/`loadEngagementData` — which never fire a request and leave slices at `INIT` — no longer keep the overlay up indefinitely. A real load still transitions to `LOADING` synchronously the instant it starts, so the overlay still covers the actual in-flight wait. Fixed in 9240b6cee6.
2. **`isEnterprise`/`ActivatedRoute` guard issue** — already fixed in 4ce129ab29 (in response to Copilot's comment on the same line), confirming your independent recheck.
3. **No test coverage for `$isReportLoading`** — added `dot-analytics-dashboard.store.spec.ts` (this store had no spec file at all before) in 9240b6cee6, covering the no-site regression case directly plus the normal loading→settled transition.

All three commits pushed to this branch.

jcastro-dotcms:
@Claude regarding your feedback:

All three findings from your review were verified and addressed:

1. **`$isReportLoading` can leave the overlay spinning forever when no site is selected (Medium).** Confirmed as a real regression. `isUnsettled` treated `ComponentStatus.INIT` as "loading", and when `currentSiteId` is empty, `loadAllPageviewData()`/`loadConversionsData()`/`loadEngagementData()` all return without firing any request, leaving their slices at `INIT` with nothing ever going to resolve that status. Fix: restricted `isUnsettled` to `ComponentStatus.LOADING` only. A real load still transitions to `LOADING` synchronously the instant it starts (before any network round trip), so the overlay still covers the actual in-flight wait — it just no longer hangs when nothing was asked to load. Fixed in commit `9240b6cee6`.

2. **`isEnterprise` read from an injected `ActivatedRoute` instead of the guard's own snapshot param (Medium).** Confirmed — same root cause you and Copilot both independently identified: `inject(ActivatedRoute)` inside a `CanActivateFn` can resolve against the route active *before* the navigation, since the target route's own `ActivatedRoute` instance isn't created until after guards pass. Fix: read `isEnterprise` from the guard's `route` snapshot parameter instead (Angular merges resolved `data` down the ancestor chain unconditionally, so it's already there — `isEnterprise` is resolved on the parent `analytics` route via `DotEnterpriseLicenseResolver`), and dropped the now-unused `ActivatedRoute` injection. This was fixed in commit `4ce129ab29` (in response to Copilot's identical comment on the same line, before your review posted).

3. **No test coverage for `$isReportLoading` (Medium).** Confirmed — this store had no spec file at all before this PR. Added `dot-analytics-dashboard.store.spec.ts` with two tests: one reproducing the no-site regression from finding #1 directly (asserts `$isReportLoading()` stays `false` when `currentSiteId` is empty), and one covering the normal loading → settled transition (asserts it flips `true` while a request is in flight and `false` once it resolves). Fixed in commit `9240b6cee6`.

The non-blocking notes (Conversions Overview removal, allowlist additivity, the `getSessionEngagement`/`toSessionEngagementData` day-field observations) were also reviewed and confirmed accurate — no changes needed there.

All changes verified: full test suites pass, lint clean, `nx format:check` clean, and `tsc --noEmit` shows no type errors in any touched file.

nicobytes (commented):
## Review: code simplification + Angular

Net positive migration (−1017 / +710). Domain-driven query API is encapsulated in `DotAnalyticsService`, DTO contracts stay stable, Conversions Overview removal is clean, and the `canActivate` + `$isReportLoading` fixes address real UX bugs.

**Verdict:** approve-with-follow-ups (non-blocking). Inline notes cover dead-code cleanup, service duplication/comment debt, and Angular guard/a11y polish.

### Already solid
- Shared `#buildDomainQueryParams` + `toSessionEngagementData`
- Guard switched from `canMatch` → `canActivate` (correct Router lifecycle)
- Overlay driven by store computed `$isReportLoading` (right layer)
- Regression coverage for no-site overlay stuck state

### Nit: PR description
Body still says `getConversionsOverview()` was left wired to the old endpoint — in the diff it was removed entirely. Worth updating for future readers.

core-web/libs/portlets/dot-analytics/data-access/src/lib/constants/dot-analytics.constants.ts
```diff
@@ -9,12 +9,17 @@ export const ANALYTICS_CONVERSION_CONTENT_ATTRIBUTION_URL =
     '/api/v1/analytics/conversion/content/attribution' as const;
```
nicobytes: **Simplification — dead constant**

`ANALYTICS_CONVERSION_CONTENT_ATTRIBUTION_URL` is no longer imported anywhere after the migration to `ANALYTICS_CONTENT_URL`. Safe to delete this export (and the JSDoc above it) in a small follow-up so we don't leave a stale proxy path in the public constants surface.

core-web/libs/portlets/dot-analytics/data-access/src/lib/store/features/with-conversions.feature.ts
```diff
@@ -81,8 +78,7 @@ const initialConversionsState: ConversionsState = {
     conversionRate: createInitialRequestState(),
```
nicobytes: **Simplification — dead store slice**

`conversionRate: RequestState<number>` is still initialized/reset here, but nothing loads it and the Conversions report computes the rate locally from `convertingVisitors`. Leftover from before the overview removal — drop the field from `ConversionsState`, `initialConversionsState`, and this reset block.

core-web/libs/portlets/dot-analytics/data-access/src/lib/types/analytics-api.types.ts
```diff
@@ -106,56 +106,6 @@ export interface ContentAttributionData {
...
- * `pagination` is required on this endpoint; shape differs from content attribution (see {@link ContentAttributionApiEntity}).
- */
-export interface ConversionsOverviewApiEntity {
-    data: ConversionOverviewData[];
-    pagination: AnalyticsConversionPagination;
-    params?: Record<string, string>;
-}
-
 /** Response wrapper from analytics event endpoints (entity.data field) */
 export interface AnalyticsEventResponse<T> {
```
nicobytes: **Simplification — unused legacy type**

`AnalyticsEventResponse<T>` modeled the old `entity.data` envelope. The service now consumes `AnalyticsQueryResponse` (`rows`). Nothing imports this anymore — delete it (and update any stale JSDoc in this file that still references `/event/total-events`, `/event/top-content`, etc.).

core-web/libs/portlets/dot-analytics/data-access/src/lib/services/dot-analytics.service.ts
```diff
@@ -102,208 +111,282 @@ export class DotAnalyticsService {
...
     getUniqueVisitors(params: GetUniqueVisitorsWithoutGranularity): Observable<UniqueVisitorsData>;
-    /** Time series: each row uses `day` as bucket label (see {@link UniqueVisitorsByDayData}). */
     getUniqueVisitors(
         params: GetUniqueVisitorsWithGranularity
     ): Observable<UniqueVisitorsByDayData[]>;
     getUniqueVisitors(
         params: GetUniqueVisitorsParams
     ): Observable<UniqueVisitorsData | UniqueVisitorsByDayData[]> {
-        const httpParams = this.#buildUniqueVisitorsParams(params);
+        const httpParams = this.#buildDomainQueryParams({
```
nicobytes: **Simplification — duplicated aggregate/series flow**

`getTotalEvents` and `getUniqueVisitors` are the same shape: build params → GET `/analytics/events` → map `rows[0]` vs map series with `dimensions=granularity`. Consider a private helper (e.g. `#queryEventsMetric`) so the next metric doesn't copy-paste this again. Same idea applies lightly to `getPageviewsByDeviceBrowser` (device vs browser only renames the dimension field).

core-web/libs/portlets/dot-analytics/data-access/src/lib/services/dot-analytics.service.ts
```diff
@@ -102,208 +111,282 @@ export class DotAnalyticsService {
...
+     * `eventType` as a dimension, that flips the mode to attribution. `eventType` here is a plain
+     * filter param, confirmed against `ContentAnalyticsService.fetchTopContent()` to have no effect
+     * on mode resolution. `orderBy`/`orderDir` are sent explicitly even though the old caller never
+     * did, to avoid relying on a server-side default for the same "descending" behavior the old
+     * endpoint returned implicitly. No field rename needed — `identifier`/`title`/`totalEvents`
+     * already match {@link TopContentData} exactly.
      *
      * @param params - Date range plus optional `siteId` and `eventType` query params
      */
     getTopContent(params: GetRangeSiteEventParams): Observable<TopContentData[]> {
```
nicobytes, line 172: **Simplification — JSDoc length**

The migration "why" notes on methods like `getTopContent` / `getContentAttribution` / `getSessionEngagementGroupBy` are useful in the PR, but as permanent service docs they're heavy (mode resolution, Postman confirmation, field renames). Prefer 1–2 lines of enduring intent (e.g. "restrict metrics to these 4 or grouped queries 400") and keep the archaeology in the PR / migration guide.

core-web/libs/portlets/dot-analytics/portlet/src/lib/guards/analytics-health.guard.ts
```diff
@@ -8,19 +8,33 @@ import { DotAnalyticsService } from '@dotcms/portlets/dot-analytics/data-access'
...
+            // Read from the guard's own snapshot param, not an injected ActivatedRoute — at
+            // guard-execution time the target route's ActivatedRoute instance doesn't exist yet
+            // (it's only created on activation, after guards pass), so `inject(ActivatedRoute)`
+            // can resolve to whatever route was active *before* this navigation instead of the
+            // one being guarded. `isEnterprise` is resolved on the parent `analytics` route (see
+            // app.routes.ts) and Angular merges resolved `data` down the whole ancestor chain
+            // unconditionally, so it's already present here correctly.
+            const isEnterprise = route.data?.['isEnterprise'] ?? true;
 
             router.navigate(['/analytics/error'], {
```
nicobytes: **Angular — prefer `UrlTree` over navigate + `false`**

`canActivate` change is correct ✅. While here: Angular's preferred pattern is to *return* the redirect instead of side-effecting:

```ts
return router.createUrlTree(['/analytics/error'], {
  queryParams: { status: healthStatus, isEnterprise }
});
```

Avoids a separate `navigate` call, plays nicer with cancellation, and makes the guard test assert the tree directly. Pre-existing smell, but this file already moved — good moment to clean it up.

core-web/libs/portlets/dot-analytics/portlet/src/lib/dot-analytics-dashboard/dot-analytics-dashboard.component.html
```diff
@@ -46,7 +46,18 @@
...
     </p-tabs>
-    <div class="p-8">
+    <div class="analytics-dashboard-body p-8">
+        <!-- Full-report loading overlay: mirrors dot-edit-content-layout's reload overlay
+             (same spinner + semi-transparent backdrop) so the ~2s customer-environment load
+             time for each tab reads consistently with the rest of the app, instead of just
+             the per-widget skeletons underneath. -->
+        @if (store.$isReportLoading()) {
+            <div
+                class="analytics-dashboard-body__loading-overlay flex items-center justify-center"
```
nicobytes: **Angular / a11y — loading overlay**

Visual parity with edit-content's reload overlay is good. For WCAG AA, consider announcing the busy state:

- `role="status"` + `aria-live="polite"` (or `aria-busy` on `.analytics-dashboard-body`)
- `aria-hidden="true"` on the decorative spinner icon
- optional `aria-label` via `| dm`

Non-blocking — matches the reference overlay — but worth a small polish since this is new UI.

core-web/libs/portlets/dot-analytics/portlet/src/lib/dot-analytics-dashboard/dot-analytics-dashboard.component.scss
```diff
@@ -34,3 +34,17 @@
...
 }
+
+// Same look as dot-edit-content-layout's reload overlay (semi-transparent white backdrop +
+// centered spinner), adapted from that component's CSS-grid positioning to a relatively
+// positioned block here since this layout isn't grid-based.
+.analytics-dashboard-body {
+    position: relative;
+}
+
+.analytics-dashboard-body__loading-overlay {
```
nicobytes: **Angular / styling — prefer Tailwind utilities**

Per project standards (`STYLING_STANDARDS.md` / Tailwind-first), this overlay does not need custom SCSS. Move these to the template on the overlay `div`:

```html
class="analytics-dashboard-body__loading-overlay absolute inset-0 z-10 flex items-center justify-center bg-white/60"
```

Then drop this rule (and ideally the parent `position: relative` if you can put `relative` on `.analytics-dashboard-body` via Tailwind in the HTML too). Keep SCSS only when PrimeNG/`::ng-deep` forces it.

core-web/libs/portlets/dot-analytics/data-access/src/lib/store/dot-analytics-dashboard.store.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+
+import { DotAnalyticsDashboardStore } from './dot-analytics-dashboard.store';
+
+import { DotAnalyticsService } from '../services/dot-analytics.service';
+
+describe('DotAnalyticsDashboardStore', () => {
+    let spectator: SpectatorService<InstanceType<typeof DotAnalyticsDashboardStore>>;
+    let store: InstanceType<typeof DotAnalyticsDashboardStore>;
+    let analyticsService: SpyObject<DotAnalyticsService>;
+    let globalStore: SpyObject<GlobalStore>;
```
nicobytes: **TypeScript — `GlobalStore` is a value, not a type**

```
'GlobalStore' refers to a value, but is being used as a type here. Did you mean 'typeof GlobalStore'?
```

`GlobalStore` is a Signal Store (runtime value), same pattern as `DotAnalyticsDashboardStore` above. Fix:

```ts
let globalStore: SpyObject<InstanceType<typeof GlobalStore>>;
```

(or mirror whatever typing other portlet store specs use for `GlobalStore`).

core-web/libs/portlets/dot-analytics/data-access/src/lib/constants/dot-analytics.constants.ts
jcastro-dotcms: Confirmed dead — removed `ANALYTICS_CONVERSION_CONTENT_ATTRIBUTION_URL` and its JSDoc in 4adc8a96b0.

core-web/libs/portlets/dot-analytics/data-access/src/lib/store/features/with-conversions.feature.ts
jcastro-dotcms: Confirmed dead — removed `conversionRate` from `ConversionsState`, `initialConversionsState`, and the no-site reset block in 4adc8a96b0.

core-web/libs/portlets/dot-analytics/data-access/src/lib/types/analytics-api.types.ts
jcastro-dotcms: Confirmed unused — deleted `AnalyticsEventResponse<T>` and fixed the stale JSDoc across this file that still referenced `/event/total-events`, `/event/top-content`, `/event/pageviews-by-device-browser`, `/session/engagement`, and `/conversion/content/attribution` (now describing the current `DotAnalyticsService` methods instead of the retired proxy paths). Done in 4adc8a96b0.

core-web/libs/portlets/dot-analytics/data-access/src/lib/services/dot-analytics.service.ts
jcastro-dotcms: Extracted a private `#queryEventsMetric<K>(metric, params)` helper — `getTotalEvents`/`getUniqueVisitors` are now one-line delegations. Structural typing worked out cleanly so no cast was needed at the call sites (K infers as the literal metric name, e.g. `Record<'totalEvents', number>` is structurally `TotalEventsData`).

Held off on `getPageviewsByDeviceBrowser` per your "lightly" framing — the device/browser duplication there is really just one output key name differing (2 lines), and unifying it would need a mapped-type generic (`{ [P in K]: string } & { total: number }`) that seemed like more type complexity than the duplication warranted. Happy to do it too if you'd rather have it unified for consistency. Done in 4adc8a96b0.

core-web/libs/portlets/dot-analytics/data-access/src/lib/services/dot-analytics.service.ts
jcastro-dotcms: Trimmed `getTopContent`/`getContentAttribution`/`getSessionEngagementGroupBy` (and lightly `getTotalEvents`/`getUniqueVisitors`, now mostly documented on the shared `#queryEventsMetric` helper instead) down to 1-2 lines of enduring intent — dropped the mode-resolution deep-dive, the Postman-confirmation note, and the "old caller never did this" framing, kept the hard constraints (e.g. the 400-risk on `getSessionEngagementGroupBy`'s metric restriction, the `eventType`-as-dimension mode flip on `getTopContent`). Done in 4adc8a96b0.

core-web/libs/portlets/dot-analytics/portlet/src/lib/guards/analytics-health.guard.ts
jcastro-dotcms: Done — the guard now returns `router.createUrlTree(['/analytics/error'], {...})` instead of calling `router.navigate()` then returning `false`. Updated the spec to mock `createUrlTree` (echoing its args back) and assert the guard's return value directly instead of asserting a `navigate` side-effect. Done in 4adc8a96b0.

core-web/libs/portlets/dot-analytics/portlet/src/lib/dot-analytics-dashboard/dot-analytics-dashboard.component.html
jcastro-dotcms: Added `role="status"`, `aria-live="polite"`, and `aria-hidden="true"` on the decorative spinner icon. Also added an `aria-label` (reusing the existing generic `Loading` i18n key rather than adding a new one) since without it the live region has no accessible name to announce — the icon is the only content and it's now hidden from AT. Done in 4adc8a96b0.

core-web/libs/portlets/dot-analytics/portlet/src/lib/dot-analytics-dashboard/dot-analytics-dashboard.component.scss
jcastro-dotcms: Moved to Tailwind utilities on the template (`relative` on `.analytics-dashboard-body`, `absolute inset-0 z-10 flex items-center justify-center bg-white/60` on the overlay), dropped both SCSS rules entirely — kept the BEM class names for the `data-testid`/selector hooks per your snippet. Done in 4adc8a96b0.

core-web/libs/portlets/dot-analytics/data-access/src/lib/store/dot-analytics-dashboard.store.spec.ts
jcastro-dotcms: Good catch, real compile error (`ts-jest` transpile-only mode doesn't catch it, only a real `tsc` run does — same lesson from a type error caught in a build earlier in this PR). Fixed: `SpyObject<InstanceType<typeof GlobalStore>>`. Verified with `tsc --noEmit` directly this time, not just the test run. Done in 4adc8a96b0.

jcastro-dotcms:
Thanks for the thorough review, @nicobytes! All 9 inline suggestions addressed and pushed in `4adc8a96b0` (replied individually on each thread) — dead code removed (unused constant, unused type, dead `conversionRate` slice, stale JSDoc referencing retired endpoints), `#queryEventsMetric` helper extracted, migration-archaeology JSDoc trimmed, guard switched to returning a `UrlTree`, a11y attributes added to the overlay, overlay styling moved from SCSS to Tailwind utilities, and a real `GlobalStore` type error in the new store spec fixed.

Also fixed the PR description nit — it still said `getConversionsOverview()` was left wired to the old endpoint; updated it to reflect that it (and the whole Conversions Overview widget) was removed entirely.

nicobytes (approved):

Issue #36360, opened 2026-06-29
[QA-G17] OpenSearch Migration — Site Search Functional Validation (Phases 0–3, $sitesearch viewtool)


fabrizzio-dotCMS:
### Description

Manual QA validation of **Site Search** across the Elasticsearch → OpenSearch migration phases, verifying
the merged work from #35786 (vendor-neutral aggregation abstraction + full OpenSearch `SiteSearchAPI`
implementation + phase-aware `SiteSearchAPIImpl` router). Goal: confirm Site Search behaves identically
whether served by Elasticsearch or OpenSearch in Phases 0–3, and that the Velocity viewtool **`$sitesearch`**
(`SiteSearchWebAPI`) renders search results, aggregations, and the legacy facet wrappers **without** the old
Elasticsearch-type cast failure.

**Parent epic:** #35476 · **Based on:** #35786 · **Mode:** QA post-merge (manual hand-checks)

Before #35786, `SiteSearchWebAPI` cast aggregation results directly to Elasticsearch implementation classes
(`InternalDateHistogram`, `StringTerms`, `StringTerms.Bucket`), which blocked Site Search from being served
by OpenSearch. #35786 made the two public Site Search APIs vendor-neutral, added the OpenSearch provider, and
routed all calls through a phase-aware router. This group verifies that work from a tester's perspective.

---

## Environment Setup

```bash
# Start migration stack (ES 7.10 + Kibana + OpenSearch 3.x + Dashboards)
docker compose -f docker/docker-compose-examples/os-migration/docker-compose.yml up -d
```

| Service               | URL                          | Purpose                          |
|-----------------------|------------------------------|----------------------------------|
| dotCMS admin          | http://localhost:8080/dotAdmin | Admin UI                       |
| Elasticsearch 7.10    | http://localhost:9200        | ES REST API                      |
| Kibana                | http://localhost:5601        | ES index inspection / queries    |
| OpenSearch 3.x        | http://localhost:9201        | OS REST API                      |
| OpenSearch Dashboards | http://localhost:5602        | OS index inspection / queries    |

```properties
# dotmarketing-config.properties (or DOT_-prefixed env vars)
OS_ENDPOINTS=http://localhost:9201
OS_AUTH_BASIC_USER=admin
OS_AUTH_BASIC_PASSWORD=admin
ENABLE_SCRIPTING=true          # REQUIRED — see Setup H2; without it .vtl files return raw source

# Migration phase flag. Restart dotCMS after every change.
[...]

PR #36849, opened 2026-07-31
feat(migration): internal migration-readiness endpoint (content + site search) (#36360)

fabrizzio-dotCMS:
## Problem

Support/QA had no single place to answer *"is it safe to change the OpenSearch migration phase right now, and if not, what do I fix?"* before promoting. QA Round 2 on #36360 showed the failure mode: promoting a phase while an index's ES and OpenSearch copies were out of sync led to silent empty/partial results, and the downgrade case (Phase-3-only content invisible after a rollback) was undocumented.

This adds an internal, read-only **migration-readiness endpoint** that condenses that status with actionable recommendations, and retires the role-gated `.os` reveal in the index portlets in favor of it.

## What this adds

**`GET /api/v1/index/migration/readiness`** — internal, read-only, advisory. Never mutates anything; the fix is always the operator re-running the crawl/reindex, which self-heals through the existing write-path gates (#36797 read half, #36825 write half).

- **Not public.** `@Hidden` (absent from the OpenAPI / API-playground schema) and gated to CMS administrators or members of the migration support role (`OS_MIGRATION_INDEX_VISIBILITY_ROLE_KEY`, default `os_migration_qa`); anyone else gets a 403.
- **Covers both mirrored families** — versioned content indices (`working`/`live`) **and** Site Search indices. `content` is a keyed object (`WORKING` / `LIVE` — a fixed pair); `siteSearch` is a list (an open set). Each entry is a per-index ES↔OS diff: `{indexName, es:{exists,docCount,physicalName}, os:{exists,docCount,physicalName}, verdict, recommendation}`, verdict `IN_SYNC` / `MISSING_COUNTERPART` / `COUNT_DRIFT`. `physicalName` is the full name as stored on each server (cluster-prefixed; `.os`-tagged on OpenSearch), and the report carries the top-level `clusterId`.
- **Overall verdict:** current phase + read/write engines + a `dualWrite` flag; `safeToAdvance` (toward OpenSearch-only) and `safeToRollback` (downgrade) with `outOfSyncCount`, a human `summary`, and per-index `blockers`. Field order: `clusterId, phase, content, siteSearch, verdict`.
- **Bare model response** — returns the readiness object directly (no `ResponseEntityView` envelope), serialized by Jackson from typed records.
- **Stateless, exact counts.** Every field derived at request time. The Site Search half uses `SiteSearchAPI.documentCount`; the content half reads each engine leaf's `getIndicesStats()` (index `_stats` `primaries.docs.count`) — never a search total (capped at 10,000, which would hide drift on large indices). Both reconcilers query the two engine leaves directly (not the phase-aware router), so both sides show in every phase.
- **`safeToRollback` needs no history:** a downgrade routes reads back to Elasticsearch, so it is unsafe when any index's ES copy is behind its OpenSearch counterpart — derivable from the same snapshot, nothing persisted.

**Teardown — index portlet visibility is now phase-only (reverts I-4).** `MigrationIndexVisibility` no longer reveals `.os` by role: `.os` indices are hidden in Phases 0/1/2 and shown in Phase 3, for everyone. The role key is retained only to gate this endpoint, which is now the single source of truth for migration/QA.

## Layers

- `com.dotcms.content.index.migration`: `MirrorStatus` (shared per-index diff), `SiteSearchMirrorReconciler`, `ContentIndexMirrorReconciler`, `MigrationReadinessService` (composition + verdict), `MigrationReadiness` (report DTO). Pure, no mutation.
- `com.dotcms.rest.api.v1.index.MigrationReadinessResource`: the JAX-RS resource + role gate.

## Tests — 31 unit green

- `MigrationReadinessServiceTest` (16): advance/rollback verdicts across phases 0–3, missing-counterpart and >10k count-drift blocking advance, content-keyed-by-slot / siteSearch-as-list, clusterId, `driftPercent`, the phase-aware `outOfSyncCount` (Phase 0 expected-missing excluded, other mismatches still counted), and rollback fail-safe on unmeasurable counts.
- `ContentIndexMirrorReconcilerTest` (5): in-sync, missing OS counterpart, count drift, null/absent slots, physical names.
- `MigrationReadinessResourceTest` (5): role gate — admin/role-member allowed, non-admin-without-role denied, null user denied, access-lookup failure fails closed.
- `MigrationIndexVisibilityTest` (5): rewritten to the phase-only contract.

**Remaining test (flagging):** a container integration test that drives HTTP 403/200 through the JAX-RS stack is not included yet — the gate decision and both reconcilers are unit-covered. Happy to add it if you'd like it before merge.

## Notes

- **Stacked on #36825** (`issue-36360-sitesearch-mirror-reconciliation`) because it consumes `SiteSearchAPI.documentCount` from that PR. Based against that branch so the diff shows only the readiness changes; will retarget to `main` once #36825 merges.
- Docs: new "Migration-readiness endpoint" subsection in `docs/backend/OPENSEARCH_MIGRATION.md`.

This PR fixes: #36360

### DIFF
--- a/docs/backend/OPENSEARCH_MIGRATION.md
+++ b/docs/backend/OPENSEARCH_MIGRATION.md
@@ -374,8 +374,10 @@ through the write-path gate above.
   for **both** mirrored families — the versioned content indices (`working`/`live`) and the Site
   Search indices. `content` is a keyed object by slot (`WORKING` / `LIVE` — a fixed pair);
   `siteSearch` is a list (an open set). Each entry carries `{indexName, es:{exists,docCount,physicalName},
-  os:{exists,docCount,physicalName}, verdict, recommendation}` — `physicalName` is the full name as
-  stored on each server (cluster-prefixed; `.os`-tagged on OpenSearch) — with verdict `IN_SYNC` /
+  os:{exists,docCount,physicalName}, driftPercent, verdict, recommendation}` — `physicalName` is the
+  full name as stored on each server (cluster-prefixed; `.os`-tagged on OpenSearch), and
+  `driftPercent` is the signed % the OpenSearch (mirror) count deviates from the Elasticsearch
+  (original) — negative = behind, positive = ahead, `null` when a count is unknown — with verdict `IN_SYNC` /
   `MISSING_COUNTERPART` / `COUNT_DRIFT`. The top level also carries the `clusterId` embedded in every
   physical name. The response is the model itself (no `ResponseEntityView` envelope).
 - **Stateless, from live counts.** Every field is derived at request time. Counts are **exact** — the
--- a/docs/backend/OPENSEARCH_MIGRATION.md
+++ b/docs/backend/OPENSEARCH_MIGRATION.md
@@ -353,10 +353,40 @@ closed when a crawl actually runs in a dual-write phase (1 or 2). In the window 
   twin that was never rebuilt, and never crawling, is a hard gap.
 
 **Operational rule (pairs with the code):** before promoting the phase — especially into Phase 3 —
-ensure every Site Search index has been crawled at least once so its OS twin exists and is in sync. A
-targeted verify/repair job (detect twins that are missing **or** whose counts diverge → rebuild full)
-that closes the no-crawl window without depending on a crawl is deferred as a follow-up under the same
-issue.
+ensure every Site Search index has been crawled at least once so its OS counterpart exists and is in
+sync. The migration-readiness endpoint below is what tells the operator *which* indices still need
+that crawl, before they change the phase.
+
+#### Migration-readiness endpoint (pre-phase-change advisory)
+
+`GET /api/v1/index/migration/readiness` is an internal, read-only report a support technician runs
+**before changing the migration phase** to see whether it is safe and, if not, what to do. It never
+mutates anything — the fix is always the operator re-running the crawl / reindex, which self-heals
+through the write-path gate above.
+
+- **Not public.** The resource is `@Hidden` (absent from the OpenAPI / API-playground schema) and
+  gated to CMS administrators or members of the migration support role
+  (`OS_MIGRATION_INDEX_VISIBILITY_ROLE_KEY`, default `os_migration_qa`); anyone else gets a 403.
+- **What it reports.** The current phase with its read/write engines and an `evaluable` flag; an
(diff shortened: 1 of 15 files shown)


fabrizzio-dotCMS:
Thanks — all three findings were real. Addressed in e6ae14ce1b.

**1. `canEvaluate()` dead code / phantom `evaluable` field — removed.** Confirmed no callers anywhere in `main` or `test`, and there is no `evaluable` field on `MigrationReadiness`/`PhaseInfo`. The phase interpretation the method encoded already lives in `MigrationReadinessService` (the per-phase advance/rollback branches), so the method was a duplicate API surface with no wiring. Dropped it and fixed the stale PR-body sentence, which now describes the `dualWrite` flag that actually ships.

**2. Phase 0 `outOfSyncCount` contradicting the summary — now phase-aware.** `outOfSyncCount` is computed through a new `needsAttentionIn(phase, status)` instead of the raw `needsAttention()`: in Phase 0 an index whose ES copy exists but whose OpenSearch counterpart does not is the *expected* state (counterparts are built during dual-write), so it no longer counts. The exclusion is deliberately narrow — an OpenSearch copy with no ES source, or a drift between two copies that both exist, is still unexpected before the migration starts and stays counted. To avoid re-creating the same contradiction in that rarer case, the Phase 0 "nothing to reconcile yet" summary now appends a note naming the leftover indices when the residual count is non-zero.

**3. `safeToRollback` false green on an unmeasurable count — fixed.** The check moved into `blocksRollback(status)`, which treats `docCount() < 0` on *either* engine as unsafe before any numeric comparison, so `es=100 / os=-1` no longer resolves through `100 < -1` to a green. Your assumption about the sentinel was right for the Site Search half; for the content half I traced it: those counts come from `getIndicesStats()` and are existence-gated (absent key → `exists=false`, count `0`), so `-1` does not currently arise there — the guard is uniform anyway so the two halves cannot diverge later. The ES-side `-1` was already fail-safe by accident (`-1 < any real count`) but is now explicit, which also covers `es=-1 / os=-1`.

Tests: 31 unit green (`MigrationReadinessServiceTest` 12 → 16), covering Phase 0 expected-missing excluded, Phase 0 unexpected mismatch still counted, and unknown OS / unknown ES counts blocking rollback. `docs/backend/OPENSEARCH_MIGRATION.md` updated for both contract changes.

PR #36852, opened 2026-07-31
feat(rendering): minify rendered HTML behind FEATURE_FLAG_MINIFY_HTML

### LINKED ISSUE #36851: Native, configurable HTML minification in the core rendering engine
## Description

dotCMS currently serves rendered HTML exactly as it comes out of the Velocity engine, which means the indentation, blank lines, and line breaks authors use to keep VTL templates, containers, and widgets readable are all shipped to the browser. The result is production HTML bloated with bytes that serve no functional purpose.

This is a customer-driven request from **ThousandEyes**, escalated through Product Feedback. They have asked for a **native, configurable HTML-minification option in the core rendering engine**, so they do not have to build and maintain a custom OSGi plugin.

### Customer rationale

- **Page weight / transfer size** — excess whitespace adds bytes to every HTML response. At thousands of page views per day this affects bandwidth and slightly increases TTFB / download time, especially on mobile and slower connections.
- **Performance scoring tools** — Lighthouse / PageSpeed Insights and GTmetrix flag unminified HTML as an improvement opportunity. It is a recurring audit item they would rather resolve than repeatedly justify as "acceptable."
- **Consistency with other assets** — they already minify CSS and JS. HTML being the one unoptimized layer is inconsistent with their front-end performance standards and internal benchmarks/SLAs.
- **Core Web Vitals and AI crawlers** — reduced page weight has a measurable effect on page speed at their scale, and cleaner, denser HTML is easier for LLM-based tools and AI crawlers to parse accurately.
- **Best practice / competitive standard** — most other CMS platforms support this out of the box or via simple configuration.

### Why the existing workarounds were rejected

The customer was offered two workarounds and formally declined both:

- **Template hygiene (removing indentation from VTL)** — indentation is how their developers visually parse nested `#if` / `#foreach` blocks. Flattening templates to suppress whitespace trades source readability for something that should be handled at the output layer. Not a tradeoff they will accept.
- **OSGi / WebInterceptor plugin using htmlcompressor** — buffering and rewriting every page response in production puts a new point of failure directly in the render/response path. If the library or their configuration mishandles an edge case (malformed markup, inline JS/JSON, a new content type), it could corrupt or break live pages sitewide, with that risk sitting entirely outside dotCMS core and on them to catch and roll back.

Their position is that the correct place for this is the core rendering engine, behind a toggle.

## Acceptance Criteria

- [ ] Rendered HTML is minified — insignificant whitespace, line breaks, and indentation removed — before it is written to the response
- [ ] Behavior is controlled by a feature flag (`FEATURE_FLAG_MINIFY_HTML`), **off by default**, so existing installs are unaffected until opted in
- [ ] Minification covers **all page render paths**, not just one: LIVE mode (streaming), PREVIEW / EDIT / ADMIN / NAVIGATE_EDIT modes, and the `getPageHtml` callers (`PageResource` REST and `PageRenderDataFetcher` GraphQL)
- [ ] Whitespace-sensitive content is preserved byte-for-byte: `<pre>`, `<textarea>`, `<script>`, `<style>`
- [ ] Inline layout whitespace is preserved so words are never joined — `<span>a</span> <span>b</span>` keeps its space, while `<div>a</div>\n<div>b</div>` does not
- [ ] JavaScript relying on newlines for automatic semicolon insertion (ASI) continues to work
- [ ] HTML comments are stripped, but downlevel conditional comments (`<!--[if IE]>`) are retained as markup
- [ ] Markup is **not** otherwise rewritten — no `<html>`/`<body>` injection into fragments, no DOCTYPE case changes, no auto-closing of tags (important for partials, URL-mapped fragments, and non-HTML templates)
- [ ] Minification failures degrade gracefully — the original markup is served rather than erroring the page
- [ ] For LIVE mode, minification happens **before the page cache write**, so the cost is paid once per cache fill rather than on every request
- [ ] Unit test coverage for the whitespace-significance and preserved-region edge cases
- [ ] Documented: flag name, default, what it does and does not minify, and the need to flush the page cache for existing cached pages to pick up the change

### Scope

This request covers **HTML whitespace minification only**. Minifying inline JS and CSS is explicitly out of scope — it is substantially riskier and should be a separate discussion if asked for later.

### Implementation notes

Investigation on the current codebase surfaced a few things worth recording for whoever picks this up:

**There is no single chokepoint for rendered HTML.** `VelocityLiveMode.serve()` streams directly to `response.getOutputStream()` and writes into the static page cache — it never returns through `getPageHtml`. A servlet-level filter or a hook in `VelocityServlet` would therefore miss the highest-traffic path entirely. The two seams that together cover everything are:

- `VelocityLiveMode.writePage()` — LIVE mode, and the point at which content enters the page cache
- `VelocityModeHandler.eval()` — all other modes plus the REST/GraphQL `getPageHtml` callers (this already does post-processing for CSP, so it is an established pattern)

**Library evaluation.** Two candidates were assessed against the criteria above:

- **jsoup** (already a dependency at 1.21.1) is a parser, not a minifier. With `prettyPrint(false)` it preserves whitespace verbatim (no minification); with `prettyPrint(true)` it re-indents. It also normalizes markup — injecting `<html><head></head><body>` into every fragment and lowercasing `<!DOCTYPE html>` — which would break fragment and URL-mapped output.
- **htmlcompressor** — the original `com.googlecode.htmlcompressor` is abandoned (last release 2011); the maintained fork is `com.github.hazendaz:htmlcompressor:2.0.2` (2024). It is safe and handles the preserved regions correctly, but it deliberately collapses inter-tag whitespace to a *single space* rather than removing it, so output still carries a space between every tag. Notably, this is the same library the customer rejected running as a plugin.

Neither delivers full whitespace removal without additional custom logic on top, so a small dependency-free minifier is a reasonable path — the risky part is whitespace significance (inline vs. block elements), which is what the test coverage above is guarding.

**Rollout note.** Enabling the flag does not retroactively minify already-cached pages; they update as cache entries refill. A page cache flush makes it immediate.

### References

- Asana (Product Feedback): https://app.asana.com/1/10691720482678/project/1213421055126718/task/1217069221210622?focus=true
- Helpdesk ticket: https://helpdesk.dotcms.com/a/tickets/38291
- Customer has offered to act as a **reference customer** to help prioritization.

fmontes:
Fixes #36851

Adds an opt-in HTML minifier so rendered pages are served without the indentation, blank lines, and line breaks that VTL templates, containers, and widgets carry for readability.

Off by default — enable with `FEATURE_FLAG_MINIFY_HTML=true`.

### Proposed Changes

* **`HtmlMinifier`** (new) — dependency-free minifier that strips insignificant whitespace and HTML comments. Deliberately conservative: it does not minify JS/CSS, rewrite attributes, or strip optional end tags.
* **`VelocityLiveMode.writePage()`** — minifies LIVE mode output **before the page cache write**, so the cost is paid once per cache fill rather than on every request.
* **`VelocityModeHandler.eval()`** — covers preview/edit/admin/navigate modes plus the `getPageHtml` callers (`PageResource` REST and `PageRenderDataFetcher` GraphQL). This method already post-processes for CSP, so minification follows the established pattern.
* **`FeatureFlagName`** — adds the `FEATURE_FLAG_MINIFY_HTML` constant.
* **`HtmlMinifierTest`** (new) — 19 test methods / 57 assertions covering the whitespace-significance, attribute-value, literal-angle-bracket, Unicode-whitespace, comment-boundary and preserved-region edge cases.

### Why two seams instead of one filter

There is no single chokepoint for rendered HTML. `VelocityLiveMode.serve()` streams directly to `response.getOutputStream()` and writes into the static page cache — it never returns through `getPageHtml`. A servlet filter or a hook in `VelocityServlet` would therefore have missed the highest-traffic path entirely. The two seams above are the minimum that covers every render path.

### Safety

The risky part of HTML minification is whitespace that *looks* removable but is actually rendered. This implementation:

* Copies `<pre>`, `<textarea>`, `<script>`, and `<style>` content **byte-for-byte** — protects rendered output and JavaScript automatic semicolon insertion.
* Copies **quoted attribute values byte-for-byte**, so `<input value="a    b">` keeps its spacing. Tags are parsed as a unit with quote tracking, which also means a `>` inside a quoted value does not end the tag early. Whitespace *between* attributes is still collapsed to one space.
* Treats `<` and `>` as literal text when they are not part of a tag, so `<p>Home > About</p>` and `<p>3 < 4</p>` are left alone.
* **Collapses** whitespace between inline elements rather than removing it, so `<span>a</span> <span>b</span>` keeps its space and words are never joined. Whitespace bordering block elements is removed.
* Collapses only the five characters HTML treats as collapsible whitespace (space, tab, LF, CR, FF). Unicode spaces that browsers **render** — the ideographic space `U+3000` common in CJK copy, the thin space `U+2009` — are left untouched. `Character.isWhitespace` is deliberately not used, since it matches those too.
* Strips HTML comments but **retains** downlevel conditional comments (`<!--[if IE]>`).
* Does **not** otherwise rewrite markup — no `<html>`/`<body>` injection into fragments, no DOCTYPE case changes, no auto-closing of tags, and no dropping of the space before a self-closing `/` (removing it would append the slash to an unquoted attribute value). This matters for partials, URL-mapped fragments, and non-HTML templates.
* Degrades gracefully — any failure logs a warning and serves the **original** markup, so a bug here cannot take a page down.
* Is idempotent, which matters because LIVE mode can minify on write and again through `eval()`.

### Checklist
- [x] Tests
- [ ] Translations — n/a, no user-facing strings
- [x] Security Implications Contemplated — see notes below

**Security notes:** minification only removes whitespace and comments; it does not decode, re-encode, or re-escape content, so it cannot introduce XSS by unescaping. Comment stripping removes HTML comments from delivered pages, which slightly *reduces* incidental information disclosure. Ordering with CSP is preserved — in `eval()`, `ContentSecurityPolicyUtil.apply()` still runs first, so nonce injection is unaffected.

### Additional Info

**Library evaluation.** Two candidates were assessed before writing custom code:

* **jsoup** (already a dependency at 1.21.1) is a parser, not a minifier. It also **injects `<tbody>`** into tables written without one, which changes the DOM and breaks `table > tr` selectors in CSS and JS that previously worked, and it keeps comments, so the information-disclosure benefit would be lost. It is the right tool as a *test oracle* here, not as the implementation. `prettyPrint(false)` preserves whitespace verbatim (no minification at all); `prettyPrint(true)` *re-indents*. It also normalizes markup — injecting `<html><head></head><body>` into every fragment and lowercasing `<!DOCTYPE html>` — which would break fragment and URL-mapped output.
* **htmlcompressor** — the original `com.googlecode.htmlcompressor` is abandoned (last release 2011). The maintained fork `com.github.hazendaz:htmlcompressor:2.0.2` is safe and handles preserved regions correctly, but deliberately collapses inter-tag whitespace to a *single space* rather than removing it, so output still carries a space between every tag. It is also the same library the customer explicitly rejected running as a plugin (see #36851).

Neither delivers full whitespace removal without custom logic layered on top, so a small owned minifier — guarded by tests — was the path chosen. No new dependency, no BOM change.

**Scope.** HTML whitespace only. Inline JS/CSS minification is intentionally out of scope; it is substantially riskier and should be a separate discussion.

**Rollout.** Enabling the flag does not retroactively minify already-cached pages — they update as cache entries refill. Flush the page cache to make it immediate.

**Testing note.** Run the unit tests with the build cache off, or Maven can report `BUILD SUCCESS` without executing anything:

```
./mvnw test -pl :dotcms-core -Dmaven.build.cache.enabled=false -Dtest='HtmlMinifier*Test'
Tests run: 29, Failures: 0, Errors: 0, Skipped: 0
```

Always check the "Tests run" count rather than the build status.

### Known limitations

Three cases minification cannot detect. All are reachable only with the flag on, which is off by default and per instance, so the mitigation in every case is to turn the flag off for that instance.

**1. `white-space: pre` applied via CSS.** Minification is driven by tag names, so whitespace made significant purely through CSS is not recognised:

Minification is driven by tag names, so whitespace made significant purely through CSS is not recognised:

```
in     : <div style="white-space:pre">a\n  b</div>
actual : <div style="white-space:pre">a b</div>
```

`<pre>` and `<textarea>` are preserved byte-for-byte, but a styled `div` used as a code block is not, and that is a realistic pattern -- close to the shape of the original request in #36851. This is not fixable with a tag heuristic, since the information lives in CSS the minifier does not parse.

An opt-out is under discussion on the PR; `data-no-minify` is the preferred form, since the failure is per element and an operator needs to fix one page rather than disable the feature globally.

**2. Elements hidden by attribute or by CSS rather than by tag name.** `INVISIBLE_TAGS` covers `script`, `style`, `template`, `noscript` and `dialog`, so whitespace beside them survives. It is keyed on the tag name, so it cannot see:

```
a <div hidden></div> b                      ->  a<div hidden></div>b                      "ab"
a <div style="display:none">x</div> b       ->  a<div style="display:none">x</div>b       "ab"
```

The `hidden` attribute is fixable (`appendTag` already parses attributes with quote tracking) but has not been done; `display:none` is the same CSS problem as (1). Both are far less common than the `script`/`svg` cases that were fixed, since they need a hidden **block** element sitting between two text runs.

**3. Fragments that begin with text are not minified.** A VTL page can render JSON, XML or CSV through these seams, where collapsing whitespace changes data rather than formatting: a run of spaces inside a JSON string is part of the value, a newline in CSV separates records, and whitespace in an XML text node is significant.

`minifyBestEffort` therefore requires the payload to **begin** with an element (after any leading whitespace, doctype or comment) and that element to be one HTML defines. Beginning with an element keeps out JSON that carries an HTML fragment in a string value; requiring a known name keeps out XML written without a declaration, such as `<rss>` or `<urlset>`.

The price is that a fragment beginning with *text* rather than a tag is no longer minified:

```
Hello   <b>world</b>    ->  unchanged
```

Compression lost rather than content changed, which is the direction to err in, but worth knowing if you serve text-leading fragments.

Any content-shape test is guesswork; the real signal is the response media type. That is unusable as things stand, because `VelocityLiveMode` (and the preview and edit handlers) call `setContentType(CHARSET)` with a charset rather than a media type, so a `text/html` check would disable the feature outright. A side effect worth its own issue: every rendered page currently ships `Content-Type: UTF-8`, and browsers are sniffing their way to the correct answer.

### Screenshots

n/a — no UI changes.

This PR fixes: #36851


jcastro-dotcms (approved):

dotCMS/src/main/java/com/dotcms/rendering/util/HtmlMinifier.java
zJaaal, line 1: We need to check this against commented code, we had an issue before that we were inserting the UVE scripts on wrong body tags, because our query to find the body tag was not taking into account commented out tags

zJaaal:
Both review findings are addressed in `345b9788`, plus a third case found while verifying them. Each has a test that fails against the previous implementation (`Tests run: 19, Failures: 3` before, `OK (19 tests)` after).

### 1. Literal `>` read as a tag end (@Copilot)

Confirmed: `<p>Home > About</p>` came out as `<p>Home >About</p>`, because `out.lastIndexOf("<")` found the surrounding `<p>` and judged the space against a block element.

Rather than confirm that the last `>` closes the last tag, the scan now tracks the tag it emitted last as it goes, so the output buffer is never re-parsed:

```java
// Name of the tag emitted last, or null when text was emitted last.
String lastTag = null;
...
private static boolean isSignificantBefore(final String lastTag) {
    return null == lastTag || isInlineTag(lastTag);
}
```

A literal `>` in text leaves `lastTag` null, so it reads as text and the space survives. A new `isMarkupStart()` gives the same treatment to a bare `<`, so `<p>3 < 4</p>` is also left alone. Side benefit: this removes an O(n) backward scan per whitespace run.

### 2. Attribute-value whitespace collapsed (`claude[bot]`)

Confirmed: `<input value="a    b">` became `<input value="a b">`. Tags are now copied as a unit by a new `appendTag()` that tracks quoting, so attribute values survive byte-for-byte and a `>` inside a quoted value no longer ends the tag early. Whitespace *between* attributes is still collapsed to one space.

### 3. Unicode whitespace that HTML renders (found while verifying the above)

`Character.isWhitespace` matches characters HTML renders rather than collapses. Verified against the JDK that `U+3000` (ideographic space, ordinary in CJK copy), `U+2009`, `U+200A` and `U+2028/9` all return `true`, so they were being replaced by an ASCII space or dropped: `<p>a　b</p>` became `<p>a b</p>`. Replaced with an `isHtmlWhitespace()` that matches only the five characters HTML treats as collapsible.

### Also

The non-blocking notes and a latent crash: `findPreserveTagEnd` no longer wraps a loop that always returned on its first iteration, the unreachable `<![endif]` branch nested under `startsWith("<!--")` is gone, and two `Set.of(...).contains(null)` paths that degenerate markup such as `</>` would have hit are guarded.

Left deliberately alone: the space before a self-closing `/`. Dropping it would append the slash to an unquoted attribute value.

### Test note

The `${net.bytebuddy:byte-buddy-agent:jar}` surefire failure described above reproduces on a clean `main` checkout too, and `-DargLine` does not override it since the pom sets it. It blocks the whole `dotcms-core` unit suite locally, not just this class. I ran JUnit directly against the module classpath instead. Worth its own issue, but unrelated to this PR. CI is what should confirm the suite.

_Investigated and written by Claude, posting under @zJaaal._

dario-daza (approved):

fabrizzio-dotCMS (commented):
Reviewed by running the algorithm rather than reading it: I extracted `minify()` into a standalone class (it is pure) and drove it through cases the suite does not cover, then re-ran this PR's own integrity oracle (jsoup 1.21.1) over the failures to see which ones it catches.

**Two content-changing bugs, one test-oracle gap, one limitation to document.** Details inline. The one I would fix first is the invisible-element whitespace case, because it is the same bug class already fixed for comments in round 1 and it silently joins words on any page with a `<script>` between inline elements.

### On the library evaluation

The jsoup verdict holds, and there is one more reason to add. With `prettyPrint(true).indentAmount(0)` jsoup also **injects `<tbody>`** into tables written without one, which changes the DOM and breaks `table > tr` selectors in CSS and JS — on top of the `<html><head><body>` injection and the lowercased DOCTYPE already noted. It also keeps comments, so the information-disclosure benefit would be lost. jsoup is the right tool as a test oracle here, not as an implementation.

### Feature-flag cost: measured, and it is a non-issue

I built a faithful replica of the `Config.getBooleanProperty` hot path as it runs in production — `systemTableConfigSource` active, `SYSTEM_GROUP` served by `cache.default.chain` (Caffeine only):

| | cost |
|---|---|
| `isEnabled()`, one read | 179 ns |
| the two reads per render in `writePage()` | ~0.36 us |
| `minify()` on the 53 KB `demo-home.html` | 256 us |
| the two reads as a share of minifying alone | 0.14 % |

So nobody should spend time optimising the flag read, and a `static final` cache would be the wrong trade — it breaks runtime flipping, which is the point of a flag. Two things are worth knowing anyway:

- **This cannot be measured in a unit test.** `systemTableConfigSource` is null until `initSystemTableConfigSource()` runs at boot, so in `HtmlMinifierTest` the expensive half of the path never executes.
- **Each read is two guaranteed cache misses.** `getSystemTableValue` probes the system table for both `DOT_FEATURE_FLAG_MINIFY_HTML` and `FEATURE_FLAG_MINIFY_HTML`, neither of which exists unless someone set it via DB. The `DOT_SYSTEM_CACHE_LOADED` flag keeps that off the database, so it stays in memory — but if an operator adds the Redis provider to `cache.default.chain`, those guaranteed misses become network round trips per read. That is the one scenario where it stops being free, and it is the argument for reading the flag once instead of twice in `writePage()` (`isEnabled()` and then again inside `minifyIfEnabled()`). Reading once also removes the window where a flag flip between the two reads buffers the whole page only to serve it unminified.

### The buffer in `writePage()` is worth one line of change

`StringWriter` is backed by a synchronized `StringBuffer`, and Velocity emits a page as many small writes. Measured on the same 53 KB page, chunked into 1359 writes:

```
initial capacity default   StringWriter 39.5 us    StringBuilderWriter 28.5 us   (1.4x)
initial capacity 32KB      StringWriter 69.9 us    StringBuilderWriter 22.4 us   (3.1x)
```

`org.apache.commons.io.output.StringBuilderWriter` is unsynchronized and already on the classpath — this same file imports `TeeOutputStream` from commons-io — so this is a one-line swap plus an initial capacity, with no new dependency. ~17 us per render, small in absolute terms, but free.

### CI note

Per the note about the local byte-buddy failure: `./mvnw test` can report BUILD SUCCESS without running anything, because the Maven build cache skips surefire. `./mvnw test -pl :dotcms-core -Dmaven.build.cache.enabled=false -Dtest='HtmlMinifier*Test'` runs them for real — check the "Tests run" count in the output.

dotCMS/src/main/java/com/dotcms/rendering/util/HtmlMinifier.java
```diff
@@ -0,0 +1,399 @@
...
+                }
+                index = commentEnd;
+                continue;
+            }
+
+            final String preserveTag = matchPreserveTag(html, index);
+            if (null != preserveTag) {
+                // Copy the element, including its content and closing tag, verbatim.
+                final int end = findPreserveTagEnd(html, index, preserveTag);
+                if (pendingSpace && INLINE_TAGS.contains(preserveTag)) {
```
fabrizzio-dotCMS: **Invisible elements are not transparent to whitespace collapsing — this changes rendered content.**

This is the same bug class you already fixed for comments in round 1, but it was not applied to `script` / `style` / `template` / `noscript`. They are `display:none`, so they generate no box, and the whitespace on either side of them is rendered — exactly like whitespace around a comment.

I ran the algorithm from this PR (extracted `minify()`, which is pure) against cases the suite does not cover:

```
in       : hola <script>var x=1</script> mundo
actual   : hola<script>var x=1</script>mundo      -> a browser paints "holamundo"

in       : <b>a</b> <style>.x{}</style> <b>b</b>
actual   : <b>a</b><style>.x{}</style><b>b</b>    -> "ab"

in       : <b>a</b> <template><i>t</i></template> <b>b</b>
actual   : <b>a</b><template><i>t</i></template><b>b</b>
```

The fix is the pattern you already built for comments — `isSignificantAfter` should skip the invisible region the way it skips a comment, and this branch should leave `pendingSpace` and `lastTag` untouched when the preserved tag is invisible:

```java
private static final Set<String> INVISIBLE_TAGS = Set.of("script", "style", "template", "noscript");
...
if (INVISIBLE_TAGS.contains(preserveTag)) {
    out.append(html, index, end);
    index = end;
    continue;   // pendingSpace and lastTag survive the region
}
```

I checked that idempotence still holds with that change: the space is emitted on the far side of the region, and the second pass reaches the same decision.

dotCMS/src/main/java/com/dotcms/rendering/util/HtmlMinifier.java
```diff
@@ -0,0 +1,399 @@
...
+     * Tags whose text content must be preserved byte-for-byte.
+     */
+    private static final Set<String> PRESERVE_TAGS = Set.of("pre", "textarea", "script", "style");
+
+    /**
+     * Elements that participate in inline layout, where whitespace between tags is rendered and is
+     * therefore significant. Whitespace touching any element <i>not</i> in this set can be removed
+     * outright; whitespace between two inline elements is collapsed to a single space instead.
+     */
+    private static final Set<String> INLINE_TAGS = Set.of(
```
fabrizzio-dotCMS, line 51: **`INLINE_TAGS` is missing inline elements that appear in most modern templates.**

Missing: `svg`, `iframe`, `canvas`, `video`, `audio`, `embed`, `math`, `acronym`, `rt`, `rp`, `rtc`. Since `img` and `input` are both here, the omission looks accidental rather than deliberate.

`svg` is the one that will bite: it is how every icon in a modern template is rendered.

```
in       : Ver <svg width="8"></svg>
actual   : Ver<svg width="8"></svg>          -> the icon is glued to the word

in       : <a href="#">x</a> <svg></svg>
actual   : <a href="#">x</a><svg></svg>

in       : <b>a</b> <iframe src="x"></iframe> <b>b</b>
actual   : <b>a</b><iframe src="x"></iframe><b>b</b>
```

`iframe` and `video`/`canvas` are the same class: replaced inline elements, so the whitespace before them is painted.

dotCMS/src/test/java/com/dotcms/rendering/util/HtmlMinifierIntegrityTest.java
```diff
@@ -0,0 +1,359 @@
...
+            tags.add(element.tagName());
+        }
+        return tags;
+    }
+
+    /**
+     * @return the document's rendered text, flattened and whitespace-normalised, so that legitimate
+     * collapsing is ignored but joined or lost words are not
+     */
+    private static String visibleText(final Document document) {
```
fabrizzio-dotCMS, line 537: **The integrity oracle has a blind spot, and it is what let the two content bugs above through.**

The class doc sells this as an invariant — "anything that changes what a browser would render is a failure". It is not, and the gap is structural rather than a missing fixture: jsoup has no CSS model, so `visibleText()` cannot see a spacing change next to an element that contributes **no text of its own**.

I ran this oracle exactly as written (jsoup 1.21.1, same four comparisons plus idempotence) over the 12 corruption cases I found. **10 pass clean**, including all four `script`/`style`/`template` cases and four of the `svg`/`canvas`/`video` ones. Only `iframe` and `acronym` are caught, and only incidentally.

The `option` workaround right below this line is the same limitation surfacing: jsoup is not layout.

The real corpus does not cover the gap either — `<svg>`, `<iframe>`, `<canvas>`, `<video>` and `<audio>` appear **0 times** across both demo pages, so `test_minify_preserves_integrity_of_real_pages` could not have caught this.

Suggested, in order of value:
1. An adjacency assertion that does not depend on the element having text — compare the leaf sequence in document order, flagging whether each leaf had adjacent whitespace. Without this, the next tag added to `INLINE_TAGS` gets the same free pass.
2. Fixtures for every invisible tag and every replaced inline element.
3. One corpus page with inline SVG icons.

The invariant idea is right, and it is the strongest part of this PR — it just needs an oracle that can actually observe the property it claims.

dotCMS/src/main/java/com/dotcms/rendering/util/HtmlMinifier.java
```diff
@@ -0,0 +1,399 @@
...
+ * first two, and JavaScript automatic semicolon insertion for {@code <script>}).
+ *
+ * @see FeatureFlagName#FEATURE_FLAG_MINIFY_HTML
+ */
+public class HtmlMinifier {
+
+    /**
+     * Tags whose text content must be preserved byte-for-byte.
+     */
+    private static final Set<String> PRESERVE_TAGS = Set.of("pre", "textarea", "script", "style");
```
fabrizzio-dotCMS, line 30: **Worth calling out as a documented limitation: `white-space: pre` applied via CSS rather than via `<pre>`.**

```
in       : <div style="white-space:pre">a\n  b</div>
actual   : <div style="white-space:pre">a b</div>
```

Not fixable with a tag heuristic — the information lives in CSS — so I am not asking for code here. But this is the ticket a customer will open: rendering a code block inside a styled `div` instead of a `<pre>` is common, and it is precisely the shape of the original request in #36851.

Two things would help: mention it explicitly in the PR body and the release note alongside the existing safety notes, and give operators an escape hatch — a configurable additional preserve-tag list, or honouring a `data-no-minify` attribute — so a broken page can be fixed without turning the whole feature off.

fabrizzio-dotCMS (commented):
Left a couple of comments, please read

dotCMS/src/main/java/com/dotcms/rendering/util/HtmlMinifier.java
zJaaal: Confirmed and fixed in `25a94ac9e1`. Reproduced all four of your cases before touching anything.

You were right that it is the comment bug class again. Rather than special-case the preserve branch, I unified the two ideas, since `isInlineTag` was never really asking about inline layout:

```java
private static final Set<String> INVISIBLE_TAGS = Set.of("script", "style", "template", "noscript");

/** Every element beside which whitespace is painted ... */
private static final Set<String> WHITESPACE_SIGNIFICANT_TAGS =
        Stream.concat(INLINE_TAGS.stream(), INVISIBLE_TAGS.stream())
                .collect(Collectors.toUnmodifiableSet());
```

The three decision points now consult that union and `isInlineTag` became `keepsAdjacentWhitespace`. One benefit over the `continue` version: it covers `template` and `noscript` without needing to locate their closing tags, since they are not preserved regions. The existing `pendingSpace && keepsAdjacentWhitespace(preserveTag)` line in the preserve branch already did the emitting, which is why `textarea` worked.

Corpus reduction is unchanged, 28.1% and 46.3%, so preserving whitespace around `<style>` blocks cost 1 to 2 bytes on the demo pages.

_Investigated and fixed by Claude, posting under @zJaaal._

dotCMS/src/main/java/com/dotcms/rendering/util/HtmlMinifier.java
zJaaal: All eleven added in `25a94ac9e1`: `svg`, `iframe`, `canvas`, `video`, `audio`, `embed`, `math`, `acronym`, `rt`, `rp`, `rtc`. Every case you listed reproduced.

Accidental rather than deliberate, yes. The set was assembled from an inline-elements list without cross-checking against replaced elements, which is exactly how `img` and `object` ended up present while `iframe` and `canvas` did not.

Agreed that `svg` is the one that would have been reported. Neither demo page in the corpus contains one, so nothing here would have caught it — see the reply on the oracle thread for what now does.

_Investigated and fixed by Claude, posting under @zJaaal._

dotCMS/src/test/java/com/dotcms/rendering/util/HtmlMinifierIntegrityTest.java
zJaaal: This was the most valuable comment on the PR, and I have reproduced your measurement. Running my own oracle over these cases: **9 of 10 pass clean**, only `iframe`, and incidentally. So the claim in the class doc was false and it now states the limitation instead of denying it.

I did not take your suggestion #1 verbatim, for a reason worth flagging. A blanket "no adjacency flag may change" assertion fails on the *legitimate* removals around block elements, and deciding which are legitimate is precisely what the minifier does, so the oracle would be checking the implementation against itself.

What landed instead keeps the adjacency idea but breaks the circularity with an independent specification list:

```java
/**
 * ... held here as an independent specification of the HTML rendering model rather than read
 * from HtmlMinifier ... the two must agree, and a tag present in one but missing from the other
 * fails this test instead of quietly changing what pages render. Reading the implementations
 * own set would make the test agree with any bug.
 */
private static final String[] WHITESPACE_SIGNIFICANT_NEIGHBOURS = { ... };
private static final String[] BLOCK_NEIGHBOURS = { ... };
```

`test_minify_keeps_whitespace_beside_every_significant_neighbour` minifies `alpha <tag></tag> omega`, removes the element, and asserts a separator survives. It never asks the element to render text, which is what makes it see what `visibleText()` cannot. `test_minify_removes_whitespace_beside_block_neighbours` is the negative half, so the significant set cannot be widened until nothing is minified — which was the obvious way to make the first test pass dishonestly.

It answers your "the next tag added gets the same free pass" point: coverage is data driven, so a tag added to the spec list is tested without writing an assertion, and a tag added to the implementation alone fails.

On your suggestion #3, a corpus page with inline SVG icons: agreed, and not done. Both corpus pages come from the demo starter and neither has one, so it needs a purpose-built page rather than another capture. Worth a follow-up rather than holding this PR.

One note on the red-then-green, since it is the interesting part: on the old code the adjacency test named exactly the 15 missing tags and passed every tag already handled, which is the check that it is measuring the right thing rather than failing for its own reasons.

_Investigated and implemented by Claude, posting under @zJaaal._

dotCMS/src/main/java/com/dotcms/rendering/util/HtmlMinifier.java
zJaaal: Reproduced, and agreed it is not fixable with a tag heuristic — the information is in CSS and this minifier does not parse CSS.

```
in     : <div style="white-space:pre">a\n  b</div>
actual : <div style="white-space:pre">a b</div>
```

Taking the documentation half now: it goes in the PR body next to the other safety notes, and it belongs in the release note for the same reason you give, that rendering a code block in a styled `div` is common and is close to the shape of the original request in #36851.

On the escape hatch, I agree with the need and would rather you or @zJaaal pick the form, since it is a new operator-facing config surface rather than a bug fix:

* a configurable additional preserve-tag list is the smaller change, but it cannot express "this one div", which is the actual failure
* honouring `data-no-minify` addresses the real case and is roughly as small, since the preserve-region machinery already exists — the cost is that it is dotCMS-specific markup a template author has to know about
* a third option: skip minification for a page entirely when the markup contains an opt-out marker, which is coarser but easiest to reason about during an incident

My preference is `data-no-minify`, because the failure is per-element and an operator hitting this needs to fix one page without turning the feature off globally. Happy to implement whichever you prefer, in this PR or a follow-up.

_Investigated by Claude, posting under @zJaaal._

zJaaal:
@fabrizzio-dotCMS thanks — running the algorithm rather than reading it found two real content bugs and a structural hole in the tests. All four inline threads are answered; this covers the review body, which raised four more items.

Both content bugs are fixed in `25a94ac9e1`, verified red-then-green, corpus reduction unchanged at 28.1% and 46.3%.

### Flag cost: taking the action, not the optimisation

Your measurement settles it — 179 ns against 256 us of minification, 0.14%. So no caching, and I agree a `static final` would be the wrong trade since it breaks the runtime flip the flag exists for.

But your second point is the actionable one, and it is now done: `writePage()` reads the flag **once** instead of twice.

```java
// minifyBestEffort, not minifyIfEnabled, because the flag was already read above and reading
// it twice per render buys nothing.
```

`minifyBestEffort` keeps the best-effort `Try` for callers that have already checked. Two reasons it is worth the line, both yours: the two guaranteed system-table misses become network round trips per read if an operator adds a remote provider to `cache.default.chain`, and reading once closes the window where a flag flip between the reads buffers a whole page only to serve it unminified.

Also noting your point that this path cannot be measured in a unit test, because `systemTableConfigSource` is null until `initSystemTableConfigSource()` runs at boot — worth knowing before anyone tries to add a benchmark for it.

### StringWriter: taken

```java
final StringBuilderWriter merged = new StringBuilderWriter(MERGE_BUFFER_INITIAL_CAPACITY);
```

Confirmed commons-io is already imported in this file (`TeeOutputStream`, line 56), so no new dependency. Sized the buffer at 32 KB, which is where your numbers show the gap widening to 3.1x, since the synchronised `StringBuffer` takes the lock on each of Velocity's hundreds of small writes.

### The jsoup `<tbody>` injection

Added to the library evaluation in the PR body. It is the strongest item on that list, because unlike the `<html>`/`<body>` injection it breaks `table > tr` selectors in CSS and JS that were previously working, and it would do so silently. Your framing is the right one: jsoup is the correct tool as a test oracle here, not as the implementation.

### The CI note, which was affecting my own verification

This one mattered more than it looks. My PR body claimed the local surefire run failed on a byte-buddy agent property. With your invocation it runs fine:

```
./mvnw test -pl :dotcms-core -Dmaven.build.cache.enabled=false -Dtest='HtmlMinifier*Test'
Tests run: 29, Failures: 0, Errors: 0, Skipped: 0
BUILD SUCCESS
```

I had been working around it by running JUnit directly against the module classpath, so the tests were real but the note in the PR body was wrong. Corrected there.

### Not done

* **A corpus page with inline SVG icons** (your suggestion #3). Both corpus pages are demo-starter captures and neither contains an `<svg>`, so this needs a purpose-built page. Follow-up.
* **The `white-space: pre` escape hatch.** Documented; I have asked on that thread which form you want, since it is a new operator-facing config surface rather than a bug fix. My preference is `data-no-minify`, because the failure is per-element and an operator needs to fix one page without disabling the feature globally.

_Investigated and implemented by Claude, posting under @zJaaal._

zJaaal:
@fabrizzio-dotCMS ready for another look when you have time. Every thread is answered individually; this is the short version of what changed.

**Both content bugs fixed** in `25a94ac9e1`, reproduced first, then red-then-green.

* Invisible elements and the 11 missing inline tags are now one idea rather than two: `INVISIBLE_TAGS` declared for its own reason, `WHITESPACE_SIGNIFICANT_TAGS` the union that the decision points consult, and `isInlineTag` renamed to `keepsAdjacentWhitespace`, which is what it was always asking. That covers `template` and `noscript` without having to locate their closing tags.
* `writePage()` reads the flag once, `StringWriter` → `StringBuilderWriter` with a sized buffer.
* Corpus reduction unchanged at 28.1% / 46.3%.

**The oracle gap, in `7df4ab7b2e`.** Your third comment was the one that mattered, and chasing it properly turned up more than the original two bugs.

* The adjacency test is data driven from a specification list held in the test, not read from the implementation, plus a block-tag negative control. 63 tags × 8 layouts = 504 checks.
* Writing those layouts exposed a bug in my own test: against the old code only 105 of the expected 120 combinations failed. The "beside inline markup" shape could never fail, because the intervening `</b><b>` left the separator non-empty whatever the minifier did. Fixed, all 8 shapes now have teeth.
* Your suggestion #3 is done, and it needed a second step I had not expected. `icons-and-media.html` is the first corpus entry built rather than captured — 45 inline SVG icons plus iframe/canvas/video/audio, none of which exist in either demo page. But adding it alone changed nothing: I checked, and the oracle is exactly as blind on the new page as on the captured ones. So `test_minify_keeps_word_to_element_separations_in_real_pages` counts word-to-element separations in the input and requires them in the output. On the old code it reports `41 of 41 <svg> elements lost the space separating them from the preceding word`.

30 tests, verified with `-Dmaven.build.cache.enabled=false` per your CI note — which also corrected the stale claim in the PR body about the local surefire failure.

**Two decisions are yours, not code I want to guess at:**

1. **The `white-space: pre` escape hatch.** Documented in the PR body. On its thread I laid out three forms (configurable preserve-tag list, `data-no-minify`, whole-page opt-out) and argued for `data-no-minify`, since the failure is per-element and an operator needs to fix one page without disabling the feature globally. Your call on the form, and on whether it belongs here or in a follow-up.
2. **A corpus page is not a substitute for a browser.** The oracle still cannot see CSS, so the limitation is documented rather than solved. If you think that is not good enough for merge, the honest answer is a real headless-browser check, which is a bigger piece of work than this PR.

**Fair warning on CI**: still in flight as I write this, and `Postman Tests - Default` has never completed a full run — the previous attempt aborted partway on a bug in my own assertion, so the LIVE cache requests, EDIT_MODE and the flag-off restore are still unexercised. Worth waiting for that job before signing off.

_Written by Claude, posting under @zJaaal._

dotCMS/src/main/java/com/dotcms/rendering/util/HtmlMinifier.java
```diff
@@ -0,0 +1,448 @@
...
+                        && isSignificantAfter(html, lookahead);
+                index = lookahead;
+                continue;
+            }
+
+            if (html.startsWith("<!--", index)) {
+                // Drop comments entirely, but keep conditional comments, which are markup.
+                final int end = html.indexOf("-->", index);
+                final int commentEnd = end < 0 ? length : end + 3;
+                if (html.startsWith("<!--[if", index)) {
```
fabrizzio-dotCMS, line 268: **The conditional-comment branch discards the pending space without emitting it — this changes rendered content.**

This is the only branch that sets `pendingSpace = false` without the `out.append(' ')` that the preserve branch (line 180) and the markup branch (line 193) both do first. Its own comment says whitespace around it is kept, but only the whitespace *after* it survives:

```
in     : a <!--[if IE]>x<![endif]-->b
actual : a<!--[if IE]>x<![endif]-->b     -> a browser paints "ab", the source paints "a b"

in     : <b>a</b> <!--[if IE]><b>x</b><![endif]--><b>b</b>
actual : <b>a</b><!--[if IE]><b>x</b><![endif]--><b>b</b>     -> "ab"

in     : <b>a</b> <!--[if !IE]>--><b>x</b><!--<![endif]-->
actual : <b>a</b><!--[if !IE]>--><b>x</b>     -> "ax"
```

Not an IE-only concern: a conditional comment is just a comment to every modern browser, so it paints nothing and the runs either side become adjacent — exactly the case `isSignificantAfter` already handles for a removed comment. The space only survives today when there happens to be whitespace on the far side too, which is why `a <!--[if IE]>x<![endif]--> b` looks fine and hides the bug.

The fix is to emit it the way the other two branches do:

```java
if (html.startsWith("<!--[if", index)) {
    if (pendingSpace) {
        out.append(' ');
    }
    out.append(html, index, commentEnd);
    pendingSpace = false;
    lastTag = null;
}
```

`pendingSpace` being true already means `isSignificantBefore` passed and `isSignificantAfter` returned `true` for the conditional comment, so emitting is the correct reading of the state, not a new judgement. I checked that idempotence holds and that `a <!--[if IE]>x<![endif]--> b` gaining a space on each side is harmless, since the browser collapses across the comment.

**Why no test caught it.** Every conditional-comment fixture in both suites uses the comment in isolation — `HtmlMinifierTest:135` and `:352`, `HtmlMinifierIntegrityTest:112` and `:138` — never with content beside it. The new adjacency test cannot reach it either: it is driven by tag names, and a comment is not a tag, so comment-like constructs never get the `alpha %s omega` treatment. That is the same shape of gap as my oracle comment last round, one level up. The difference from the `svg` case is that this one *is* visible to `visibleText()`, because there is real text on both sides — `"a b"` against `"ab"`. A fixture is enough here; no new oracle needed.

Third case above, separately: in the downlevel-revealed form the closing `<!--<![endif]-->` is stripped while the opening `<!--[if !IE]>-->` is kept, since only `<!--[if` matches. Harmless in a modern browser, but the asymmetry is worth a line of comment if it is deliberate.

dotCMS/src/main/java/com/dotcms/rendering/util/HtmlMinifier.java
```diff
@@ -0,0 +1,448 @@
...
+
+    /**
+     * Elements that generate no box at all ({@code display: none}). They are <b>transparent</b> to
+     * whitespace collapsing in exactly the way a removed comment is: the element paints nothing, so
+     * the whitespace on either side of it ends up separating the surrounding content and is
+     * rendered. Dropping it joins words, which is why
+     * {@code hola <script>x</script> mundo} must not become {@code hola<script>x</script>mundo}.
+     */
+    private static final Set<String> INVISIBLE_TAGS =
+            Set.of("script", "style", "template", "noscript");
```
fabrizzio-dotCMS: **Two more invisible elements, and one of them is detectable in markup.**

`INVISIBLE_TAGS` is keyed on the tag name, which leaves the same corruption reachable through `<dialog>` without `open` and through the `hidden` attribute — both `display: none` in the UA stylesheet:

```
in     : a <dialog>x</dialog> b
actual : a<dialog>x</dialog>b     -> "ab"

in     : a <div hidden></div> b
actual : a<div hidden></div>b     -> "ab"

in     : a <div style="display:none">x</div> b
actual : a<div style="display:none">x</div>b     -> "ab"
```

`dialog` is a tag name, so it fits the set exactly as it stands — a one-word addition. `hidden` needs an attribute check, and `appendTag` already parses attributes with quote tracking, so the machinery is there if you want it. The `style="display:none"` variant is not fixable with a tag heuristic and belongs beside the `white-space: pre` note in the PR body, since it is the same "the information lives in CSS" limitation.

I would not block on any of this. The frequency is well below `svg` or `script`, and a `hidden` block element sitting between two text runs is not a common template shape. But it is the last corner of the class, and I would rather it be a decision than an unnoticed gap.

dotCMS/src/test/java/com/dotcms/rendering/util/HtmlMinifierIntegrityTest.java
```diff
@@ -0,0 +1,553 @@
...
+            "noscript", "object", "output", "picture", "progress", "q", "rp", "rt", "rtc", "ruby",
+            "s", "samp", "script", "select", "slot", "small", "span", "strike", "strong", "style",
+            "sub", "sup", "svg", "template", "textarea", "time", "tt", "u", "var", "video", "wbr"};
+
+    /**
+     * Block-level tags, as the negative control. Whitespace beside these is <i>not</i> rendered, so
+     * it must be removed. Without this, widening the significant set until everything is preserved
+     * would pass the test above while minifying nothing.
+     */
+    private static final String[] BLOCK_NEIGHBOURS = {
```
fabrizzio-dotCMS, line 219: **The negative control has holes, so a widened significant set could still pass both tests.**

The control does its job for the tags it lists, and pairing it with the significant list is the right structure. But `dd`, `dt`, `caption`, `figcaption`, `h3`–`h6`, `details`, `summary`, `legend` and `pre` are in neither list, so adding any of them to `WHITESPACE_SIGNIFICANT_TAGS` would pass this test and the one above it.

That matters because the stated point of holding the specification in the test is that drift fails it. Right now drift only fails for the 28 tags named here. I checked all of the above against `separatorAround(tag, ELEMENT_BARE, DOCUMENT_BARE)` as written — including `pre`, where the preserve branch copies `<pre></pre>` verbatim so `between.contains(element)` still holds — and every one of them already produces an empty separator. So completing the list is a one-line change that passes today and buys the guarantee for the tags most likely to be reached for next.

fabrizzio-dotCMS (commented):
Round 2 verified. All four inline findings are fixed, and I checked the same way as last time — by running the algorithm rather than reading it. I re-extracted `minify()` from `7df4ab7b2e` and drove it through the 21 cases from my last review plus new variants: uppercase `<SCRIPT>`/`<SVG>`, two adjacent script regions, and the `<embed>` void element.

| finding | status |
|---|---|
| invisible elements | fixed in `25a94ac9e1` — all four cases round-trip and stay idempotent, and the uppercase and back-to-back-region variants are correct too |
| `INLINE_TAGS` | all eleven present, every case I listed now keeps its space |
| oracle blind spot | `7df4ab7b2e` — the 63-tag specification list matches the implementation's union exactly, block negative control, 8 layouts, the class doc now states the limitation instead of denying it, and `icons-and-media.html` closes the corpus gap |
| `white-space: pre` | documented in the PR body next to the other safety notes |

Two things I want to name as better than what I asked for. Unifying `INVISIBLE_TAGS` and `INLINE_TAGS` into `WHITESPACE_SIGNIFICANT_TAGS` beats the `continue` I sketched, because it covers `template` and `noscript` without having to locate their closing tags. And you were right not to take suggestion #1 verbatim: a blanket "no adjacency flag may change" assertion would have the test validating the implementation against itself, which is the failure mode the specification list avoids. `separatorAround` stripping the element and the remaining tags before measuring is the detail that makes it real — without that line the "beside inline markup" shape would pass whatever the minifier did.

On the escape hatch, my preference stands: `data-no-minify`, because the failure is per element and an operator hitting it needs to fix one page rather than turn the feature off globally. Either here or as a follow-up, your call.

**One new bug of the same class, in the conditional-comment branch, plus two smaller notes — inline.** The conditional-comment one is not a regression from this round, it predates both fix rounds, but it is the last instance of the pattern those rounds were chasing and it changes what a modern browser renders. The four threads above are answered as far as I am concerned, feel free to resolve them.

fabrizzio-dotCMS (commented):
Round 3 verified. All three findings are fixed, and I re-extracted `minify()`/`minifyBestEffort()` from `90cd646971` and ran them rather than reading the diff.

| finding | status |
|---|---|
| conditional comment dropped the pending space | fixed — all three shapes keep their space and stay idempotent, including the downlevel-revealed one |
| `dialog` / `hidden` | `dialog` in the set; `hidden` and `style="display:none"` documented in the class doc, which is the call I said I would not block on |
| negative-control holes | all twelve tags landed (`dd`, `dt`, `caption`, `figcaption`, `h3`–`h6`, `details`, `summary`, `legend`, `pre`) |

I re-ran the full round-1 and round-2 case set against this commit as a regression check: invisible elements, the eleven inline tags, literal `>`/`<`, comment transparency and block collapsing all still behave, all idempotent.

Two things worth naming. `<dialog open>` is a block element and now keeps a space it does not need — that is the correct direction of error, since an extra byte is never a corruption, and it is not worth special-casing. And the fixture comment explaining *why* the conditional-comment gap existed is the part that stops it coming back; the five new fixtures cover exactly the shapes that were missing.

**I verified your content-type justification independently, and it holds.** `VelocityModeHandler:39` defines `CHARSET` as `Config.getStringProperty("CHARSET", "UTF-8")`, and `VelocityLiveMode:117`, `VelocityPreviewMode:61` and `VelocityEditMode:54` all pass that to `setContentType`. Nothing else on that path sets one — the other `setContentType("text/html")` calls in the tree are error, redirect and login interceptors. So gating on `text/html` would indeed disable the feature outright, and you were right not to reach for it here. It also means every rendered page is shipping `Content-Type: UTF-8`, which is not a media type. That is pre-existing and squarely outside this PR, but it is worth its own look — right now browsers are sniffing their way to the correct answer.

**The new scrutiny is on `looksLikeHtml`**, since it is new behaviour that now decides whether the feature runs at all, and it was not something I asked for. Net it is clearly better than minifying unconditionally, and it errs in the safe direction. But it has two escapes, and only one of them is documented — details inline, plus a measured nit.

One cosmetic thing: `dialog` went into `WHITESPACE_SIGNIFICANT_NEIGHBOURS` between `ruby` and `s`, and the reflow left `"style"` alone on its own line. The list is alphabetical everywhere else.

dotCMS/src/main/java/com/dotcms/rendering/util/HtmlMinifier.java
```diff
@@ -0,0 +1,494 @@
...
+     * @param content the rendered payload, never {@code null} or empty
+     * @return {@code true} when the payload carries markup and is safe to minify
+     */
+    private static boolean looksLikeHtml(final String content) {
+
+        if (content.stripLeading().regionMatches(true, 0, "<?xml", 0, 5)) {
+            return false;
+        }
+
+        for (int index = 0; index < content.length(); index++) {
```
fabrizzio-dotCMS: **The undocumented escape: XML without a declaration is still minified, and the class doc right above says why that is wrong.**

The declaration is the only XML signal here, and it is optional — a document without a prolog is well-formed XML 1.0. So this check catches the polite case and misses the rest:

```
in     : <?xml version="1.0"?>\n<rss>  <item>a  b</item></rss>
actual : unchanged                                              -- correct

in     : <rss>\n  <title>a  b</title>\n</rss>
actual : <rss><title>a b</title></rss>                          -- text node collapsed
```

By the reasoning in your own javadoc ("whitespace in XML text nodes is significant") the second one is the same corruption class as the CSV case, just quieter. Realistic shapes reaching these seams: an Atom or RSS fragment, a sitemap written without the prolog, or a standalone SVG served from a VTL page.

**The documented residual is the more likely of the two to actually be hit.** A VTL page rendering JSON with content bodies in it is the classic pre-GraphQL dotCMS pattern, and one HTML fragment in one string value switches minification back on for the whole payload:

```
in     : {\n  "body": "<p>a   b</p>",\n  "note": "two  spaces"\n}
actual : { "body": "<p>a b</p>", "note": "two spaces" }
```

`"two  spaces"` lost a space it owned. Comment stripping reaches inside string values the same way.

**A stronger check, using only what the class already holds.** Require the payload to *start* with markup — after leading whitespace, and allowing a doctype or a comment first — and require the first element to be a known HTML tag. That closes all three:

* JSON and CSV start with `{`, `[` or text, so they never enter
* `<rss>`, `<feed>`, `<urlset>` are not in `WHITESPACE_SIGNIFICANT_TAGS` or `PRESERVE_TAGS`, so declaration-less XML is skipped on the tag name instead of the prolog
* an HTML page starts with `<!DOCTYPE` or `<html>`, and a tag-leading fragment still qualifies

The cost is explicit and worth stating in the javadoc: a fragment that *starts* with text (`Hello <b>world</b>`) would no longer be minified. That is compression lost, not content changed, and the PR body already flags fragment output as a supported shape — so it is your call whether that trade is worth it, or whether the first-element check alone (which fixes the XML case but not the JSON one) is the right stopping point.

Either way, both escapes belong in the limitations note, not just the JSON one. And I would say the same thing here as on `white-space: pre`: any content-shape heuristic is guesswork, and the real fix is the media-type signal — which lands back on `setContentType(CHARSET)`.

dotCMS/src/main/java/com/dotcms/rendering/util/HtmlMinifier.java
```diff
@@ -0,0 +1,494 @@
...
+     * string value. Gating on the response content type would be the stronger check, but
+     * {@code VelocityLiveMode} currently calls {@code setContentType(CHARSET)} with a charset rather
+     * than a media type, so that signal is not usable as it stands.
+     *
+     * @param content the rendered payload, never {@code null} or empty
+     * @return {@code true} when the payload carries markup and is safe to minify
+     */
+    private static boolean looksLikeHtml(final String content) {
+
+        if (content.stripLeading().regionMatches(true, 0, "<?xml", 0, 5)) {
```
fabrizzio-dotCMS: **`stripLeading()` copies the whole payload to test five characters, and Velocity output usually has the leading whitespace that triggers it.**

`String.stripLeading()` returns `this` only when there is nothing to strip; otherwise it allocates a copy of the entire remainder. Measured on a 53 KB payload with a single leading newline:

```
stripLeading() on 53 KB with a leading \n    11.5 us per call
minify() on the 53 KB demo-home.html        256 us      (your number, unchanged)
```

So 4.5% of the cost of minifying, spent on a 5-character comparison, and paid on every render whose first byte is whitespace — which is the normal shape of Velocity output, since a directive on line 1 emits nothing but the newline after it survives.

An index does the same job without the copy:

```java
int start = 0;
while (start < content.length() && isHtmlWhitespace(content.charAt(start))) {
    start++;
}
if (content.regionMatches(true, start, "<?xml", 0, 5)) {
    return false;
}
```

`isHtmlWhitespace` rather than `Character.isWhitespace` also keeps this consistent with the rest of the class, which deliberately avoids the latter.

Small in absolute terms, same as the `StringWriter` note last round — just free.

fabrizzio-dotCMS (approved):
Approving. Round 3 is verified, and I went looking for new bugs rather than only re-reading my own — nothing real came back.

### What I ran this round, against `90cd646971`

**Adversarial fixtures on malformed and hostile markup**, none of which corrupt anything and all of which are idempotent:

* `<script>var s = "</script>";</script>` — the region ends at the first `</script`, which is exactly what an HTML parser does; JS string context does not protect it in a browser either, so matching that is correct
* preserved regions in uppercase (`<PRE>`, `<SCRIPT>`), with nested inline children, and unterminated
* unterminated tag, unterminated comment, unterminated `<script>` — each degrades the way a browser does, and the unterminated comment swallowing to EOF is right
* unquoted attribute values, single-quoted JSON in a `data-` attribute, self-closing ``, `<div   class="a"   >`
* `a < b`, `a <! b`, `a </> b`, `<!DOCTYPE html>`
* U+3000 ideographic space untouched, and a surrogate pair (`😀`) survives the `charAt` scanning intact

**The real corpus, end to end.** I reproduce your numbers exactly, and all three pages are idempotent:

```
demo-home.html         54275 -> 39007   28.1% reduction   idempotent
demo-members.html      15141 ->  8124   46.3% reduction   idempotent
icons-and-media.html   21183 -> 15167   28.4% reduction   idempotent
```

**The comment-stripping concern I wanted to rule out.** Since `eval()` covers edit and preview, I checked whether anything in the render path uses HTML comments as functional markers that a client reads — the failure mode behind the commented-out `<body>` issue raised earlier in this PR. Every `<!--` I can find in the rendering path and the shipped VTL is decorative (`</div><!--/row-->` and similar). Nothing functional, so stripping them is safe here.

### Why the two open threads do not block

Both are real, and I am leaving them on record rather than withdrawing them, but neither is merge-blocking:

* **XML without a declaration** — reachable, but these seams only ever carry a rendered dotCMS page, and a feed or sitemap written as a VTL page essentially always carries the prolog. When it is hit, the damage is whitespace collapsing in text nodes, which is cosmetic in the realistic shapes. Nothing like the CSV case, where the row separator became a space — and that one is fixed.
* **JSON carrying an HTML fragment** — the more likely of the two, but bounded harder than I first framed it: JSON forbids literal newlines and tabs inside strings, so the only reachable corruption is a run of *spaces* inside a string value collapsing to one. Structural whitespace collapsing is harmless. Worth the documentation line; not worth holding the PR.

Both belong in the limitations note next to `white-space: pre`, since right now only the JSON one is written down. And `stripLeading()` is a measured nit, not a bug — your call whether to take it here.

One footnote from the fuzzing, for the record rather than for action: `isMarkupStart` accepts `<?`, so `a <?php echo 1; ?> b` becomes `a<?php echo 1; ?>b`. A processing instruction is a bogus comment to a browser, paints nothing, and is therefore the invisible-element class one more time. The frequency of a PI in dotCMS-rendered HTML is zero, so I would not touch it — but it is the same shape, if the pattern ever needs a general answer instead of a set of tag names.

Good work across three rounds. The shape that made the difference is the specification list held in the test: it is what turned my third comment last round into coverage rather than a fixture.

dotCMS/src/main/java/com/dotcms/rendering/util/HtmlMinifier.java
zJaaal: Both escapes confirmed and fixed in `451074b04b`. Reproduced first:

```
<rss>\n  <title>a  b</title>\n</rss>              ->  <rss><title>a b</title></rss>
{"body": "<p>a   b</p>", "note": "two  spaces"}   ->  "two spaces"
```

You were right that the documented residual is the likelier of the two. A VTL page rendering JSON with content bodies is common enough that I should not have shipped it as a footnote.

Adopted your rule, with one change after checking it. The suggestion was to test the first element against `WHITESPACE_SIGNIFICANT_TAGS` or `PRESERVE_TAGS`, but `html`, `body`, `div` and `p` are in none of those sets, so that rule rejects an ordinary page and a `<div>`-leading fragment as readily as it rejects `<rss>`. It would have disabled the feature almost everywhere.

So there is now an explicit `HTML_TAGS` allow-list. Same effect for the cases you named, and better on the open-world point: an XML vocabulary nobody listed is skipped by default rather than minified until somebody reports it.

The trade is taken and tested rather than only described — `test_minifyBestEffort_still_minifies_html` pins `Hello <b>world</b>` as *not* minified, alongside the page, fragment, leading-whitespace and leading-comment cases that must keep working. Both escapes are in the limitations note now, not just the JSON one.

Agreed on the ending: this is guesswork standing in for a missing signal, and the signal is the media type. Your independent confirmation that nothing else on the path sets one is the useful part — it also means every rendered page is shipping `Content-Type: UTF-8` today. That is worth its own issue, and I have noted it in the PR body rather than reaching for it here.

_Investigated and fixed by Claude, posting under @zJaaal._

dotCMS/src/main/java/com/dotcms/rendering/util/HtmlMinifier.java
zJaaal: Taken in `451074b04b`, near enough verbatim:

```java
int cursor = skipHtmlWhitespace(content, 0);

if (content.regionMatches(true, cursor, "<?xml", 0, 5)) {
    return false;
}
```

`isHtmlWhitespace` rather than `Character.isWhitespace`, as you say — the class avoids the latter deliberately, since it treats U+3000 and U+2009 as collapsible when a browser paints them, which was one of the earlier bugs.

Your point about *when* it is paid is the part I had not thought about. I would have filed this as a micro-optimisation on a cold path; it is actually the common path, because Velocity output usually begins with the newline a first-line directive leaves behind. Same shape as the `StringWriter` note last round.

_Fixed by Claude, posting under @zJaaal._

PR #36860, opened 2026-08-02
Notify devs via Slack when their change hits the latest track

### LINKED ISSUE #36859: Notify devs when their change lands on the latest evergreen track
## Summary

When the release pipeline moves the floating `latest` Docker tag (the `promote-latest` job in `cicd_6-release.yml`), notify the GitHub users whose commits are newly included via Slack, telling them their change is live on `latest` and when the evergreen-tracks reconciler will roll it out to `latest`-track customer environments.

## Why

Developers currently have no visibility into when their merged change actually reaches a running customer environment. The evergreen-tracks reconciler already runs a predictable weekday cron for the `latest` track (`kubernetes/cluster/*/evergreen-tracks-reconciler/cronjob-latest.yaml`, 10:00 AM ET onward per cluster) — we can compute the next scheduled rollout and tell the author directly.

## Approach

- New reusable workflow `cicd_comp_notify-latest-promotion.yml`, called from `cicd_6-release.yml` right after `promote-latest` succeeds.
- Diffs commits between the current release tag and the previous one (via GitHub compare + commits/pulls API) to find PR authors.
- Resolves GitHub logins to Slack users via the existing `utility_slack-channel-resolver.yml`.
- Looks up which environments are currently subscribed to the `latest` track via the dotUsage `GET /environments` API (tenant/customer names), with a generic fallback if that lookup fails.
- Computes the next weekday 10:00 AM America/New_York rollout window and includes it in the Slack message.

## Out of scope

- Per-cluster exact times (message references the earliest/reference window; other regions follow later the same day).
- Live Google-Sheet-based subscription source (dotUsage already mirrors this).

sfreudenthaler:
## Summary
- New reusable workflow `cicd_comp_notify-latest-promotion.yml`, wired into `cicd_6-release.yml` right after `promote-latest` succeeds.
- Diffs the release against the previous one via the GitHub compare + commits/pulls API to find PR authors (no commit-message parsing).
- Resolves GitHub logins to Slack via the existing `utility_slack-channel-resolver.yml` (same pattern as `pr-notifier`).
- Looks up which environments are currently on the `latest` evergreen track via the dotUsage `GET /environments` API, with a generic fallback if that call fails.
- Computes the next weekday 10:00 AM America/New_York rollout window and includes it in the message.
- Gated on the existing `notify_slack == 'true'` toggle and the same CLI/LTS/non-`v` exclusion the Slack report step already uses.

## Known limitations (by design, ponytail)
- The dotUsage lookup runs on the repo's only self-hosted runner (`[self-hosted, linux, x64, ubuntu-server]`), since `usage.dotcms.cloud` sits behind a corporate-IP WAF allowlist that GitHub-hosted runners aren't on. **Not yet verified that this runner's egress IP is on that allowlist** — if it isn't, the lookup step fails soft and the message falls back to generic wording ("the `latest`-track environments") rather than naming customers.
- `DOTUSAGE_TOKEN` is a personal viewer-role token (expires ~Jan 27); fine for now per the requester, but worth swapping for a dedicated service token later.
- Previous-release lookup assumes a release's diff is under 250 commits (GitHub compare API single-page cap) — a non-issue for a single release's worth of commits.

Closes: #36859

### DIFF
--- a/.github/workflows/cicd_6-release.yml
+++ b/.github/workflows/cicd_6-release.yml
@@ -308,6 +308,32 @@ jobs:
             uv run evergreen-tracks promote --repo "${repo}" --tracks latest --apply
           done
 
+  # Notify latest-track authors - tell devs their change is live on `latest`
+  # and when evergreen-tracks will roll it out to subscribed environments.
+  # Needs `release` (not just release-prepare/promote-latest) so the GitHub
+  # Release object for this tag is guaranteed published before we look up the
+  # previous release to diff against - `release` runs in parallel with
+  # promote-latest, not before it, so skipping this dependency would race it.
+  # Reuses the same notify_slack toggle and CLI/LTS/non-v exclusion as the
+  # Slack report step below.
+  notify-latest-authors:
+    name: Notify Latest Authors
+    needs: [ release-prepare, promote-latest, release ]
+    if: >-
+      needs.promote-latest.result == 'success'
+      && github.event.inputs.notify_slack == 'true'
+      && !contains(needs.release-prepare.outputs.release_tag, 'dotcms-cli-')
+      && !contains(needs.release-prepare.outputs.release_tag, '_lts_')
+      && startsWith(needs.release-prepare.outputs.release_tag, 'v')
+    uses: ./.github/workflows/cicd_comp_notify-latest-promotion.yml
+    with:
+      release_tag: ${{ needs.release-prepare.outputs.release_tag }}
+    secrets:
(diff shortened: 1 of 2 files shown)


sfreudenthaler:
Superseded — the release-time approach here required either loosening dotUsage's WAF or a self-hosted GitHub Actions runner, both out of bounds. Reworked as an in-cluster reconciler feature instead: see dotCMS/infrastructure-as-code#8157.

Issue #36802, opened 2026-07-29
PGListener lifecycle: unbounded rebuilds with no backoff, leak window before Thread.start(), stop() allocates, flag-based idempotency guard


dsolistorres:
## Problem Statement

The `PGListener` lifecycle inside `JDBCPubSubImpl` has four independent defects that together turned a transient connection failure into the unbounded rebuild storm analysed in spike #36544 (~3,687 listener instantiations in one 600s window).

**1. Rebuilds are unbounded — no backoff, no rate limit, no cap.** `listener()` constructs a fresh `PGListener` whenever the current one is not listening, and six call sites reach it, including `publish()` (`:291`) — so ordinary cache-invalidation traffic drives reconstruction. `PostgresPubSubImpl` already has the right shape: linear backoff `min(delay + 1000, 10000)`, reset on success (`:189`, `:99`).

**2. A real leak window between the constructor and `Thread.start()`.** The connection is opened in the constructor (`:128`) but the thread only starts at `:69`. Since `run()`'s `finally { stopListening(); }` (`:177-184`) is what closes the connection, anything throwing in that window leaks it permanently — e.g. `unwrap(PGConnection.class)` failing at `:122`, or `Thread.start()` failing under a thread storm.

> Note for anyone reading the original hypothesis in #36544: the *ordinary* death paths do **not** leak. `runInternal()`'s `return` on `!connectionAlive()` and the `KILL_ON_FAILURES` throw are both covered by that `finally`. `subscribeTopic()` failure is also safe — it closes before re-throwing (`:155`). This narrow pre-`start()` window is the only genuine leak.

**3. `stop()` can construct a listener in order to destroy one.** `stop()` is `listener().stopListening()` (`:102`). If the listener is already dead, `listener()` builds a complete replacement — borrow + `LISTEN` every topic + `Thread.start()` — purely to stop it again. Shutdown paths must not allocate.

**4. The idempotency guard tests a flag, not the resource.** `isListening()` reads `runstate` (`:172-174`), which lags reality by up to `SLEEP_BETWEEN_RUNS` (500ms) and diverges entirely when the connection dies underneath the thread. `PostgresPubSubImpl` checks the real resource: `connection != null && !connection.isClosed()` (`:146`).

**Impact:** any clustered install on the default provider. Amplifies #36544's pool exhaustion; independently causes connection churn and wasted `LISTEN` traffic even on a healthy cluster.

## Steps to Reproduce

[...]

Issue #36801, opened 2026-07-29
Pub/sub listener holds a permanent connection from the shared JDBC pool, exhausting jdbc/dotCMSPool


dsolistorres:
## Problem Statement

`JDBCPubSubImpl` — the default pub/sub provider since 24.01.26 — borrows its permanently-held Postgres `LISTEN` connection from the **shared application pool** (`jdbc/dotCMSPool`):

```java
// JDBCPubSubImpl.java:121
private final Lazy<Connection> connection =
    Lazy.of(() -> Try.of(() -> DbConnectionFactory.getDataSource().getConnection())...);
```

That single design decision is the root cause behind spike #36544. Because Postgres `LISTEN/NOTIFY` delivers notifications only to the specific backend session that subscribed, and because vanilla pgjdbc has no async notification callback, the connection must be held open **and kept perpetually busy** (a `SELECT 1` poll every 500ms, `:197-201`). It can therefore never be returned to the pool, even momentarily.

The consequence in production (customer `firstmac`, pod `dotcms-firstmac-prod-1-1`, 26.07.06-01, 2026-07-08): repeated listener rebuilds each borrowed a fresh pooled connection — **~3,687 borrows in a single 600s window** — until the pool was exhausted and every DB consumer starved simultaneously (`DbConnectionFactory`, `RulesEngine`, `PostgresJobQueue`, `VelocityServlet`). Startup never completed and pods were SIGKILLed in a crash-loop.

Cluster messaging must not compete with page rendering for connections. Note the sibling implementation `PostgresPubSubImpl` — the default until December 2023 — already gets this right, acquiring a dedicated connection outside the pool (`:226`) with a `POSTGRES_PUBSUB_JDBC_URL` override (`:217`).

**Impact:** any clustered install on the default provider. `PubSubCacheTransport` is the only functioning `CacheTransport` in the tree, so all cluster cache coherence flows through this one connection.

## Steps to Reproduce

Full harness on branch `issue-36544-pubsub-connection-churn-spike`:

```bash
git fetch origin issue-36544-pubsub-connection-churn-spike
git checkout issue-36544-pubsub-connection-churn-spike
cd docker/docker-compose-examples/pubsub-connection-churn
docker compose up -d          # wait for "Startup completed" on node 1
./churn.sh                    # 30 kill cycles
```

[...]

Issue #36544, opened 2026-07-13
Spike: Cluster re-wire re-inits Postgres pub/sub transport and leaks DB connections → pool exhaustion + startup crash-loop


dsolistorres:
## Research Question

During a startup crash-loop on a large, clustered production deployment, does a repeated `ClusterFactory.rewireClusterIfNeeded()` → pub/sub cache-transport re-init cycle **leak DB connections** — because `JDBCPubSubImpl` does not close its dedicated listener connection when the `PGListener` dies/restarts — until `jdbc/dotCMSPool` exhausts and startup never completes?

Secondary question: what is the true emitter of the `select count(*) as test from inode` storm (3,571× in a single 600s startup, ~88% of startup time)? Is it the HikariCP validation query set via the `DOT_`-prefixed `DOT_DB_VALIDATION_QUERY` (not the plain `DB_VALIDATION_QUERY`, which the customer reports empty), or another source?

Can we reproduce this on a local docker-compose cluster to confirm root cause and validate fixes?

## Timebox

8h

## Acceptance Criteria

- [ ] Stand up a local 2-node dotCMS docker-compose cluster (v26.07.06 image) + shared Postgres + LB, cache-cluster autowire on, and attempt to reproduce the rewire → pub/sub re-init → connection-leak → pool-exhaustion cycle.
- [ ] Confirm or refute: `JDBCPubSubImpl` leaks its dedicated connection when the `PGListener` dies/restarts. The death paths — `runInternal()`'s `return` on `!connectionAlive()` and the `KILL_ON_FAILURES=100` throw — exit without closing the connection; only `stopListening()` closes it. Provide HikariCP pool-stat / `pg_stat_activity` evidence of connection count climbing per listener restart.
- [ ] Determine what drives repeated `rewireClusterIfNeeded()`: alive-server set instability, the current server missing from its own alive set, stale `cluster_server` rows from SIGKILLed pods that never deregister, and/or heartbeat writes failing under pool pressure.
- [ ] Identify the true emitter of `select count(*) as test from inode` (×3,571). Verify whether `DOT_DB_VALIDATION_QUERY` (or a `-D` JVM property) is set to it; if so, confirm the HikariCP `connectionTestQuery` behavior described in #34840. If not, trace the actual source.
- [ ] Characterize how the `maxLifetime=60s` bug (#34921) amplifies pub/sub connection churn (a 60s `maxLifetime` retires the "held-forever" LISTEN connection every minute, forcing listener restarts).
[...]

PR #36864, opened 2026-08-03
fix(clustering): surface silent cache-transport failures and make rewire retryable (#36803)

### LINKED ISSUE #36803: Cluster cache invalidations are dropped silently when the pub/sub transport fails to initialize
## Problem Statement

When the cluster cache transport fails to initialise, dotCMS **silently stops propagating cache invalidations** and keeps reporting itself healthy. This is why spike #36544's incident went undetected until pods crash-looped, and it is independently dangerous: nodes serve stale content with no signal at all.

**1. Invalidations are dropped with no log and no exception.** `PubSubCacheTransport.send()` opens with:

```java
// PubSubCacheTransport.java:58-60
if (!this.initialized.get()) {
    return;
}
```

Every cache invalidation is discarded. No warning, no counter, no error. `PubSubCacheTransport` is the only functioning `CacheTransport` in the tree (the alternative is `NullTransport`), so on a clustered install this means **all cache coherence is silently lost**.

**2. The rewire failure is swallowed.** `ClusterFactory.addMeToCacheIfNeeded()` catches everything from `setCluster()` / `testCluster()` and only logs it (`:381-383`), so a persistently failing rewire never escalates — and `KNOWN_SERVERS` is still updated afterwards as if it had succeeded.

**3. Re-init is not idempotent, guaranteeing a retry loop.** `init()` sets `initialized = true` only as its **last** statement (`:52`), after `pubsub.start()` and `pubsub.subscribe()`. If either throws — which they do once the pool is drained — the flag stays `false`, so `ChainableCacheAdministratorImpl.setCluster()` re-enters `init()` on every subsequent rewire (`:190`). Combined with 1 and 2, the system retries forever, drops every invalidation in the meantime, and says nothing.

**Impact:** any clustered install. The failure mode is silent staleness, which is worse than a loud failure — during the #36544 incident the cluster was not merely slow, it was quietly incoherent.

## Steps to Reproduce

1. Start the 2-node harness on branch `issue-36544-pubsub-connection-churn-spike`:

```bash
git fetch origin issue-36544-pubsub-connection-churn-spike
git checkout issue-36544-pubsub-connection-churn-spike
cd docker/docker-compose-examples/pubsub-connection-churn
docker compose up -d
```

2. Drive the pool to exhaustion so `pubsub.start()` fails inside `init()` (the harness config does this: pool of 15, 30s borrow timeout, a ~148ms validation query):

```bash
CYCLES=100 INTERVAL=2 ./churn.sh
```

3. Edit a contentlet on node 1 and read it on node 2.

**Expected:** either the invalidation propagates, or the failure is visible — a warning, a metric, a degraded health check.
**Actual:** node 2 serves stale content. Nothing is logged by `send()`. `/dotmgt/livez` and `/dotmgt/readyz` continue to report healthy. The only hint is `ClusterFactory`'s swallowed error, buried among unrelated startup noise.

## Acceptance Criteria

- [ ] `PubSubCacheTransport.send()` no longer discards invalidations silently. At minimum it increments a dropped-message counter and logs at WARN, rate-limited so a sustained outage cannot flood the log.
- [ ] A dropped-invalidation metric is exposed (e.g. `dotcms.cachetransport.invalidations.dropped`) so staleness is detectable from monitoring rather than by comparing content across nodes.
- [ ] A cache-transport / pub/sub health check is registered in `CoreHealthCheckProvider` and reports **degraded** when the transport is uninitialised or the listener is not listening. `com.dotcms.health` already provides `HealthCheckBase`, tolerance configuration and failure windows, and there is currently no such check.
- [ ] Decide and document whether an uninitialised cache transport should affect readiness. A node that cannot receive invalidations arguably should not serve traffic in a cluster — but this must not cause a cold-start deadlock, so the tolerance window matters.
- [ ] `PubSubCacheTransport.init()` is idempotent: a failed `init()` leaves no partially-initialised state, and repeated calls neither leak resources nor duplicate subscriptions.
- [ ] A persistently failing rewire in `ClusterFactory.addMeToCacheIfNeeded()` (`:381-383`) escalates rather than being logged and forgotten — surfaced via the health check above and/or an error counter, not a bare `Logger.error`.
- [ ] Verified with the harness: with the transport failing, the health endpoint reports degraded within a bounded interval, and the dropped-invalidation counter is non-zero.

## dotCMS Version

Observed during the #36544 incident on **26.07.06-01** (commit `5efb47a`). The `send()` early-return and the swallowed rewire exception are long-standing and present on `main`; they are not specific to the December 2023 provider change.

## Severity

## Links

- Spike with full RCA: #36544 — see `docs/core/incidents/36544-pubsub-connection-churn.md` on branch `issue-36544-pubsub-connection-churn-spike`
- Parent epic: #34837
- [Freshdesk ticket #38172](https://dotcms.freshdesk.com/a/tickets/38172)
- Related: #34923 (pool sizing and observability)

wezell:
### Proposed Changes

Fixes #36803 — cluster cache invalidations were dropped **silently** when the pub/sub transport failed, and a persistently failing rewire was logged and forgotten (root cause analysis in #36544).

* **`PubSubCacheTransport.send()`** no longer discards invalidations silently: increments a dropped-message counter and logs at WARN, rate-limited (`CACHE_TRANSPORT_DROP_WARN_INTERVAL_MILLIS`, default 30s) so a sustained outage cannot flood the log.
* **`PubSubCacheTransport.init()` is idempotent**: a rewire on an already-healthy transport is a no-op instead of a tear-down/rebuild of the pub/sub listener — this removes the rebuild-per-rewire-pass churn measured in #36544 (`LISTEN cluster_actions` ×3,687).
* **`ClusterFactory.addMeToCacheIfNeeded()`** now reports success/failure instead of swallowing exceptions. `KNOWN_SERVERS` is only updated on success, and `rewireClusterIfNeeded()` retries whenever there is a pending failure, so a failed transport init is always retried on the next heartbeat. Consecutive failures are tracked and exposed via `ClusterFactory.getRewireFailures()` (reset on success).
* **New `cache-transport` health check** (`CacheTransportHealthCheck`, registered in `CoreHealthCheckProvider`): reports unhealthy when the transport is uninitialized or rewires are failing persistently; exposes `initialized` / `droppedInvalidations` / `startupDroppedInvalidations` / `failedInvalidations` / `rewireFailures` / `awaitingInitialization` as structured data. Nodes with no real transport always report healthy, and a node whose transport has not come up yet is treated as still starting for a grace period rather than reported unhealthy.
* **Micrometer gauges** in the existing `CacheMetrics` binder:
  * `dotcms.cache.transport.invalidations.dropped`
  * `dotcms.cache.transport.invalidations.dropped.startup`
  * `dotcms.cache.transport.invalidations.failed`
  * `dotcms.cache.transport.initialized`
  * `dotcms.cache.transport.rewire.failures`
* `CacheTransport` gains `default long getDroppedMessages()`, `default long getStartupDroppedMessages()` and `default long getFailedMessages()` (0 for all other transports).

### Detail in comments

Two sections were moved out of this description to keep it readable:

* [**Review follow-ups**](https://github.com/dotCMS/core/pull/36864#issuecomment-5255377503) — what changed in each of the seven post-review commits and why, including the two suggestions that were deliberately not implemented as written.
* [**Deployment impact**](https://github.com/dotCMS/core/pull/36864#issuecomment-5255377848) — measured log and health-status noise per node boot, before and after the startup-grace change. Read this one if you care whether shipping this generates alerts.

### Scope note for reviewers

The second follow-up commit widens this PR past the cache subsystem the title implies. `QueuingPubSubWrapper` and `DotPubSubProvider` are shared by every pub/sub topic in the JVM — cache invalidation, OSGi restart (`OsgiRestartTopic`), and cluster management (`ClusterManagementTopic`) — so those paths now execute the new code too.

The changes there are additive and behaviour-preserving:

* `DotPubSubProvider.getFailedPublishCount(String)` is a new `default` method returning 0, so no existing provider changes behaviour or needs updating.
* In `QueuingPubSubWrapper.publish()`, the submitted task's body moved into `publishAndRecordOutcome()`. The publish call itself, the dedupe cache, the submitter, and the unconditional `true` return are all unchanged — the only additions are counting a `false` return and catching a throw that was previously discarded by the submitter.
* Failure counts are keyed per topic, so a non-cache topic's failures cannot inflate the cache transport's metric (covered by `test_failures_are_attributed_per_topic`).

Worth a second pair of eyes on that file specifically, since a regression there would affect OSGi restarts and cluster management rather than just cache invalidation.

### Readiness and alerting decisions (AC: "decide and document")

The health check defaults to **MONITOR_MODE**: it reports degradation but never fails readiness probes. A node that cannot send invalidations can still serve traffic, and gating readiness on the transport risks a cold-start deadlock (transport init happens during cluster wiring). Operators who prefer to drain such nodes can opt in with `health.check.cache-transport.mode=PRODUCTION`. It is never a liveness check — restarting pods on transport failure is what amplified the #36544 incident.

**Failed invalidations deliberately do not make the check unhealthy on their own.** The count is cumulative and never resets, so alarming on "greater than zero" would pin a node DOWN forever after one transient publish error — the same false-positive class fixed for rewire failures. Alert on the rate of increase of `dotcms.cache.transport.invalidations.failed` instead. Rewire failures *do* flip the check, but only past a threshold, because that counter resets on success and the rewire is retried every heartbeat.

### Testing

* `PubSubCacheTransportTest` (unit, 10 cases): drops are counted and not published before init; publish works after init; `init()` is idempotent (single `start()` across repeated calls, re-inits after `shutdown()`); a synchronous publish failure is counted as failed rather than dropped; a successful publish counts neither; an async provider's self-reported failures surface without being double counted; the first `init()` retires pre-init drops into the startup counter; drops after the first `init()` survive a later re-init instead of being laundered into it. Plus the atomic init guard: 16 threads racing into `init()` produce exactly one `start()`/`subscribe()`, and a throwing `start()` leaves the transport retryable.
* `ClusterFactoryRewireTest` (unit, 5 cases, new): steady state does not rewire; a pending failure forces a retry even when membership looks unchanged; membership join and leave both rewire; a node missing from its own alive set rewires; the first heartbeat rewires.
* `QueuingPubSubWrapperTest` (unit, 5 cases): a `false` return from the wrapped provider is counted even though `publish()` reported success; a thrown failure is counted rather than lost on the submitter thread; failures are attributed to their own topic and do not inflate the cache topic; successful publishes count nothing and an unused topic reports 0.
* `CacheTransportHealthCheckTest` (unit, 10 cases, new): see `5aecea7292` in [Review follow-ups](https://github.com/dotCMS/core/pull/36864#issuecomment-5255377503).
* All 30 pass locally. **Every test written for a review finding was confirmed to fail without its fix**, so none of them is a test that passes either way: the concurrency one observes 3 `start()` calls with `synchronized` removed (3/3 runs), the case-normalization one reports `expected:<1> but was:<0>`, and the retry one fails when the `pendingFailures` clause is dropped. Note that running `:dotcms-core` unit tests requires `-Dmaven.build.cache.enabled=false` — the build-cache extension skips `dependency:properties`, which populates `${net.bytebuddy:byte-buddy-agent:jar}`, leaving a literal `-javaagent:` path that crashes the surefire fork. Pre-existing and unrelated to this PR (an untouched `DotPubSubEventTest` fails identically).
* **Two-node cluster, built from this branch.** Cluster forms and both nodes report UP; PING/PONG bidirectional; a cache invalidation made on node 1 is observed on node 2 (node 2 served `ORIGINAL` from cache, then `UPDATED-VIA-NODE1` after the invalidation); exactly one `initing PubSubCacheTransport` per node, confirming the idempotent `init()` in the real rewire loop; an uninitialized transport is detected and MONITOR_MODE keeps readiness passing; the drop WARN is throttled as intended (2 warnings for 2,841 drops); a `NullTransport` node reports UP with the per-transport fields omitted; message and structured data agree on every poll.
* **The post-init failure path was verified live**, which is the novel part of `343865b39e`: cutting node 2's link to the database produced `failedInvalidations: 2` with `initialized: true` and `droppedInvalidations` unchanged, matching exactly two `Unable to send pubsub` log lines — no double counting, correct per-topic attribution, node 1 unaffected at `failed: 0`. This ran through the default queued path (`DOT_PUBSUB_USE_QUEUE` defaults true). Node 2 recovered fully on reconnect.
* **Re-verified on a rebuilt image after the grace period was added** — zero DEGRADED `cache-transport` polls across 6,110 samples on two boots, with the counter split reading `dropped: 0 / startupDropped: 2841` and `0 / 6`. Full numbers in [Deployment impact](https://github.com/dotCMS/core/pull/36864#issuecomment-5255377848).
* **Not verified live: the rewire counter.** `ServerHeartbeatJob.execute()` calls `LicenseUtil.updateLicenseHeartbeat()` before `ClusterFactory.rewireClusterIfNeeded()`, and the former throws when the database is down, so `REWIRE_FAILURES` cannot move during a database outage. It only moves when the database is healthy and `setCluster()`/`testCluster()` fails — a narrow window I could not reproduce locally. The threshold and reset logic are covered by unit tests instead.

### Known gaps

* `RedisStreamsPubSubImpl.publish()` always returns `true` (fire-and-forget async `xadd`), so it reports no failures. Left as-is; its own design decision.
* This surfaces silent transport failures; it does not fix the crash-loop or the missing already-ran guard in `StartupTasksExecutor` — #36801 / #36802 still own those.
* Three **pre-existing** problems were found while verifying this branch on a local cluster, and filed separately. None is caused by this PR, and none blocks it — the health check is unaffected by all three — but the first two do mean the gauges this PR adds are not yet consumable:
  * #36995 — the Micrometer metrics subsystem never initializes, so no `dotcms.*` meter is ever registered.
  * #36996 — `/dotmgt/metrics` has no servlet mapping and returns 404, breaking `docker/docker-compose-examples/single-node-metrics-monitoring`.
  * #36997 — `DOT_PUBSUB_PROVIDER_OVERRIDE` cannot be set by environment variable, so a container meant to use Redis pub/sub can silently run on JDBC.
* Startup drops are *reported* here, not eliminated. That invalidations are broadcast at all before the transport is up is arguably its own gap (a booting node in an existing cluster has nothing to send them through), but fixing boot ordering is well outside this PR.

https://claude.ai/code/session_01XFuXwYdNsb7irpMEeSGTyj

This PR fixes: #36803

### DIFF
--- a/dotCMS/src/enterprise/java/com/dotcms/enterprise/cluster/ClusterFactory.java
+++ b/dotCMS/src/enterprise/java/com/dotcms/enterprise/cluster/ClusterFactory.java
@@ -56,6 +56,7 @@ import java.util.Enumeration;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import org.elasticsearch.common.settings.Settings.Builder;
 
@@ -361,14 +362,17 @@ public class ClusterFactory {
     public static void rewireCluster() throws Exception {
 
         if(clusterReady()) {
-            addMeToCacheIfNeeded();
-            KNOWN_SERVERS  = APILocator.getServerAPI().getAliveServers();
+            if (addMeToCacheIfNeeded()) {
+                // only remember the alive-server set when the rewire actually succeeded,
+                // otherwise a failed transport init would never be retried
+                KNOWN_SERVERS = APILocator.getServerAPI().getAliveServers();
+            }
         }else {
             Logger.info(ClusterFactory.class, "Cluster not yet active. Not rewiring");
         }
     }
 
-     private static void addMeToCacheIfNeeded() throws DotDataException  {
+     private static boolean addMeToCacheIfNeeded() throws DotDataException  {
          if(isEnterprise()) {
 
              final Server localServer = APILocator.getServerAPI().getOrCreateMyServer();
@@ -378,15 +382,31 @@ public class ClusterFactory {
                      getImplementationObject()).setCluster(localServer);
                  ((ChainableCacheAdministratorImpl) CacheLocator.getCacheAdministrator().
                      getImplementationObject()).testCluster();
+                 REWIRE_FAILURES.set(0);
+                 return true;
              } catch (Exception e) {
-                 Logger.error(ClusterFactory.class, e.getMessage(), e);
+                 final long failures = REWIRE_FAILURES.incrementAndGet();
+                 Logger.error(ClusterFactory.class, "Unable to (re)initialize the cluster cache transport"
+                         + " (consecutive failures: " + failures
+                         + "). Cache invalidations are NOT reaching other nodes: " + e.getMessage(), e);
+                 return false;
              }
 
-                 
-             
          } else {
-             CacheLocator.getCacheAdministrator().getTransport().shutdown(); 
+             CacheLocator.getCacheAdministrator().getTransport().shutdown();
+             return true;
          }
+    }
+
+    /**
+     * Consecutive cluster cache-transport rewire failures. Reset to zero on the first
+     * successful rewire. Surfaced by the cache-transport health check and metrics so a
+     * persistently failing rewire is visible instead of being logged and forgotten.
+     */
+    private static final AtomicLong REWIRE_FAILURES = new AtomicLong(0);
+
+    public static long getRewireFailures() {
+        return REWIRE_FAILURES.get();
     }
      
      private static boolean isEnterprise() {
--- a/dotCMS/src/enterprise/java/com/dotcms/enterprise/cluster/ClusterFactory.java
+++ b/dotCMS/src/enterprise/java/com/dotcms/enterprise/cluster/ClusterFactory.java
@@ -65,7 +65,7 @@ public class ClusterFactory {
 
 
     private static boolean CLUSTER_INITED=false;
-	private static List<Server> KNOWN_SERVERS=Collections.EMPTY_LIST;
+	private static List<Server> KNOWN_SERVERS=Collections.emptyList();
 
     /**
      * Consecutive cluster cache-transport rewire failures. Reset to zero on the first
@@ -351,13 +351,8 @@ public class ClusterFactory {
 		try{
 			List<Server> aliveServers = APILocator.getServerAPI().getAliveServers();
 
-            // A pending failure must always be retried. The membership comparison below only
-            // fires when the alive-server set changes, so without this a rewire that failed
-            // while membership then settled back to KNOWN_SERVERS would never be attempted
-            // again: the transport would stay broken and REWIRE_FAILURES could never return to
-            // zero, leaving the health check reporting DOWN indefinitely.
-            if (REWIRE_FAILURES.get() > 0 || !aliveServers.equals(KNOWN_SERVERS) || !aliveServers
-                    .contains(APILocator.getServerAPI().getCurrentServer()) ) {
+            if (shouldRewire(aliveServers, KNOWN_SERVERS,
+                    APILocator.getServerAPI().getCurrentServer(), REWIRE_FAILURES.get())) {
 
 				rewireCluster();
 			}
@@ -365,10 +360,42 @@ public class ClusterFactory {
 		catch(Exception e){
 			Logger.error(ClusterFactory.class, "Unable to rewire cluster:" + e.getMessage());
 			Logger.error(ClusterFactory.class, "servers:" + KNOWN_SERVERS);
-			
-			
+
+
 			throw new DotRuntimeException(e);
 		}
+    }
+
+    /**
+     * Decides whether {@link #rewireCluster()} should run on this heartbeat.
+     *
+     * Extracted from {@link #rewireClusterIfNeeded()} so it can be unit tested without the
+     * enterprise statics, {@code APILocator} and license state that method needs. Pure function of
+     * its arguments; behaviour is unchanged from the inline condition it replaces.
+     *
+     * A rewire is needed when any of the following holds:
+     *
+     * - {@code pendingFailures > 0} -- a previous rewire failed and must be retried. This is the
+     *   clause added for issue #36803. The membership comparison below only fires when the
+     *   alive-server set *changes*, so without this a rewire that failed while membership then
+     *   settled back to {@code knownServers} would never be attempted again: the transport would
+     *   stay broken, {@code REWIRE_FAILURES} could never return to zero, and the cache-transport
+     *   health check would report unhealthy indefinitely.
+     * - cluster membership changed since the last successful rewire.
+     * - this server is missing from the alive set, so its own registration needs redoing.
+     *
+     * @param aliveServers   servers currently seen as alive
+     * @param knownServers   membership as of the last *successful* rewire
+     * @param currentServer  this server
+     * @param pendingFailures consecutive rewire failures not yet cleared by a success
+     */
+    @VisibleForTesting
+    static boolean shouldRewire(final List<Server> aliveServers, final List<Server> knownServers,
+                                final Server currentServer, final long pendingFailures) {
+
+        return pendingFailures > 0
+                || !aliveServers.equals(knownServers)
+                || !aliveServers.contains(currentServer);
     }
     
     public static void rewireCluster() throws Exception {
--- a/dotCMS/src/enterprise/java/com/dotcms/enterprise/cluster/ClusterFactory.java
+++ b/dotCMS/src/enterprise/java/com/dotcms/enterprise/cluster/ClusterFactory.java
@@ -67,6 +67,13 @@ public class ClusterFactory {
     private static boolean CLUSTER_INITED=false;
 	private static List<Server> KNOWN_SERVERS=Collections.EMPTY_LIST;
 
+    /**
+     * Consecutive cluster cache-transport rewire failures. Reset to zero on the first
+     * successful rewire. Surfaced by the cache-transport health check and metrics so a
+     * persistently failing rewire is visible instead of being logged and forgotten.
+     */
+    private static final AtomicLong REWIRE_FAILURES = new AtomicLong(0);
+
     private static final String VERIFY_REQUIRED_TABLES =
             "SELECT dc.cluster_id, dc.cluster_salt, cs.server_id, cs.cluster_id, cs.name, "
                     + "cs.ip_address, cs.host, cs.cache_port, cs.es_transport_tcp_port, "
@@ -344,7 +351,12 @@ public class ClusterFactory {
 		try{
 			List<Server> aliveServers = APILocator.getServerAPI().getAliveServers();
 
-            if (!aliveServers.equals(KNOWN_SERVERS) || !aliveServers
+            // A pending failure must always be retried. The membership comparison below only
+            // fires when the alive-server set changes, so without this a rewire that failed
+            // while membership then settled back to KNOWN_SERVERS would never be attempted
+            // again: the transport would stay broken and REWIRE_FAILURES could never return to
+            // zero, leaving the health check reporting DOWN indefinitely.
+            if (REWIRE_FAILURES.get() > 0 || !aliveServers.equals(KNOWN_SERVERS) || !aliveServers
                     .contains(APILocator.getServerAPI().getCurrentServer()) ) {
 
 				rewireCluster();
@@ -399,12 +411,10 @@ public class ClusterFactory {
     }
 
     /**
-     * Consecutive cluster cache-transport rewire failures. Reset to zero on the first
-     * successful rewire. Surfaced by the cache-transport health check and metrics so a
-     * persistently failing rewire is visible instead of being logged and forgotten.
+     * Consecutive cluster cache-transport rewire failures, zero when the last rewire succeeded.
+     *
+     * @see #REWIRE_FAILURES
      */
-    private static final AtomicLong REWIRE_FAILURES = new AtomicLong(0);
-
     public static long getRewireFailures() {
         return REWIRE_FAILURES.get();
     }
--- a/dotCMS/src/main/java/com/dotcms/cache/transport/PubSubCacheTransport.java
+++ b/dotCMS/src/main/java/com/dotcms/cache/transport/PubSubCacheTransport.java
@@ -36,7 +36,16 @@ public class PubSubCacheTransport implements CacheTransport {
 
     final AtomicLong droppedMessages = new AtomicLong(0);
 
+    /**
+     * Sends that were attempted and reported as failed by a synchronous provider. Failures from
+     * an asynchronous provider are counted by the provider itself and read back in
+     * {@link #getFailedMessages()}; see {@link DotPubSubProvider#getFailedPublishCount(String)}.
+     */
+    final AtomicLong failedMessages = new AtomicLong(0);
+
     private final AtomicLong lastDropWarnAt = new AtomicLong(0);
+
+    private final AtomicLong lastFailWarnAt = new AtomicLong(0);
 
     private static final long DROP_WARN_INTERVAL_MILLIS =
             Config.getLongProperty("CACHE_TRANSPORT_DROP_WARN_INTERVAL_MILLIS", 30000);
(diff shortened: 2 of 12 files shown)


dsolistorres:
## Review follow-ups

Moved out of the PR description to keep it readable. Seven commits added after review; full rationale in each commit message.

Seven commits added after review. Details in the commit messages.

**`b7aa686d45` — remove false positives from the monitoring**

* **`CacheMetrics` had no `NullTransport` guard.** `NullTransport.isInitialized()` is `false` once it has been shut down, so a node with no real transport published `transport.initialized=0` and alerted as though it were dropping invalidations — while the health check reported the same node healthy. Both now share one `activeTransport()` helper and agree.
* **Both call sites resolved the transport by casting `getImplementationObject()` to `ChainableCacheAdministratorImpl`.** `getTransport()` is on the `DotCacheAdministrator` interface and `CommitListenerCacheWrapper` delegates it, whereas the cast throws `ClassCastException` for any other administrator — `NullCacheAdministrator.getImplementationObject()` returns itself, which is the unit-test path — swallowed into a `null` that happened to give the right answer. Both now use the interface method, as `ClusterResource` does. The cast in `addMeToCacheIfNeeded` stays: `setCluster()`/`testCluster()` are genuinely not on the interface.
* **A single rewire failure no longer reports DOWN.** `testCluster()` can throw on a momentary database hiccup while the transport stays initialized and invalidations keep flowing. The check now tolerates `health.check.cache-transport.rewire-failure-threshold` (default 3) consecutive failures, and always reports the count.
* **The rewire counter now has a guaranteed path back to zero.** `rewireClusterIfNeeded()` only fired when the alive-server set *changed*, so a failure followed by membership settling back to the stale `KNOWN_SERVERS` was never retried — the transport stayed broken and the check would have reported DOWN indefinitely. It now also retries whenever `REWIRE_FAILURES > 0`.
* **`performCheck()` and `buildStructuredData()` share one immutable snapshot.** Each previously resolved the transport and re-read the counters independently, which let a single health response contradict itself — a message saying the transport is initialized next to structured data saying it is not.

**`343865b39e` — count invalidations that fail *after* init**

The original change instrumented only the pre-`init()` drop path. A transport that initialized successfully and then started failing every publish stayed completely invisible: no drops recorded, `isInitialized()` still true, `rewireFailures` 0, health check green, other nodes serving stale content.

* Every provider signals a failed send by **returning `false` rather than throwing** — `JDBCPubSubImpl` (the default) and `PostgresPubSubImpl` on an exception or an `execute()` that returns false, `RedisPubSubImpl` when stopped — and `send()` discarded that boolean.
* Checking it in `send()` is necessary but not sufficient. `DOT_PUBSUB_USE_QUEUE` **defaults to true**, so the provider is normally a `QueuingPubSubWrapper` whose `publish()` returns `true` immediately and completes the real send on a submitter thread, discarding the result — `send()` would read `true` 100% of the time in the default configuration. The wrapper now records the outcome of the task it submits, and `PubSubCacheTransport` sums its own synchronous failures with the provider-reported ones. Exactly one of the two counts any given attempt, so nothing is double counted.
* Counted **per topic**, because one provider instance is shared by every topic in the JVM (cache invalidation, OSGi restart, cluster management) and a JVM-wide total could not be attributed to the cache transport.
* `getFailedMessages()` is kept **distinct from `getDroppedMessages()`**: dropped means "the transport is down", failed means "the transport believes it is up but sends are erroring". Both lose invalidations, but only the first is visible from `isInitialized()`.

**`6c19c292eb` — stop reporting startup drops as failures**

Found by running a two-node cluster from this branch (see *Deployment impact* below). Boot order guarantees dropped invalidations: caches are invalidated by startup tasks and the starter import long before `ClusterFactory` wires the cluster and calls `init()`. A first boot against an empty database dropped **2,841**, and the cumulative counter carried that burst for the life of the node.

* Drops from before the **first** successful `init()` are retired into a separate `startupDroppedMessages` counter, so `getDroppedMessages()` reports only invalidations lost while the transport was expected to be carrying them. Retiring is first-init-only on purpose: a transport that came up, went down, lost invalidations and recovered has lost real ones, and retiring on every re-init would launder genuine loss into the benign startup bucket.
* The health check treats a **never-yet-initialized** transport as still starting up for `health.check.cache-transport.initialization-grace-period-seconds` (default 120). The grace period ends permanently once the check has seen an initialized transport — losing one that had been working is reported immediately.
* `dotcms.cache.transport.initialized` is left **raw**, with no grace period: a gauge states the current fact, and alert rules add the duration clause (`for: 2m`). Alert on the health check if you want the grace period applied for you.

**`5aecea7292` — unit-test the health check**

Addresses the review's medium finding: the largest new file in the PR had no coverage. Ten cases over `NullTransport`, unresolvable cache layer, the grace-period boundaries in both directions, the rewire threshold either side, structured-data keys, the MONITOR_MODE default, and liveness exclusion. Writing them caught a real defect in `6c19c292eb` — `retireStartupDrops()` ran on every `init()` — which was fixed before that commit landed.

**`47f54bf7fc` — make the init guard atomic**

The review's second medium: `init()` guarded with a check-then-act, so two threads could both read `initialized == false` and each run `start()` + `subscribe()`.

The reviewer asked for verification that concurrent invocation is possible before changing anything. It is not, today — `init()` is reached only through `ChainableCacheAdministratorImpl.setCluster()` ← `addMeToCacheIfNeeded()` ← `rewireCluster()` ← `ClusterFactory.rewireClusterIfNeeded()`, each with exactly one call site, and the last is `static synchronized`. Every entry point (`ServerHeartbeatJob`, `FreeServerFromClusterJob`, `ClusterFactory.initialize()` from `LicenseManager` and two startup tasks) funnels through that monitor. Fixed anyway, because `rewireCluster()` is `public static` and not itself synchronized, so the guarantee is not local to this class and a future caller could bypass it.

`init()` and `shutdown()` are now `synchronized` — once per cluster rewire, so the monitor costs nothing.

**Deliberately not the `compareAndSet` form suggested in review.** Claiming the flag up front (`if (!initialized.compareAndSet(false, true)) return;` with the work after) means a thrown `start()` stays flagged as initialized: `isInitialized()` returns true, `shouldReinit()` returns false, `setCluster()` never retries, and the new health check reports a healthy transport that never subscribed — reintroducing the exact silent failure this issue removes. The flag is set only after `start()` and `subscribe()` return. `test_failed_init_leaves_the_transport_retryable` is the test that would fail under the suggested form.

**`633580ee3f` — normalize the failure-count topic key**

Writes into `QueuingPubSubWrapper.failedByTopic` and reads out of it arrive by routes that disagree on case: `recordFailure()` keys on the event's topic, which `DotPubSubEvent.Builder.withTopic` has already lowercased, while `PubSubCacheTransport.getFailedMessages()` looks up `this.topic.getTopic()` — a bare `String.valueOf(getKey())` with no normalization. They match only because every current key is already lowercase (`CACHE_TOPIC` is `"dotcache_topic"`). One uppercase character in a future key and the lookup misses silently, reporting zero failures while invalidations are lost.

Normalized on both sides inside the wrapper, which owns the map, rather than with a `.toLowerCase()` at the call site as review suggested — that would leave every caller needing to know the wrapper's key convention, and would fix only the cache topic. `DotPubSubProvider` now documents that the key is matched case-insensitively. Latent fragility, not a live defect.

**`fb8db88d58` — cover the rewire-retry decision**

The rewire retry was the least-protected change in this PR: only the health check's *consumption* of `REWIRE_FAILURES` was tested, not the producer, and it is the logic that guarantees a failed transport init keeps being retried.

The decision was an inline condition behind `clusterReady()`, `isEnterprise()` and two `APILocator` calls, so it is extracted into a pure package-private `shouldRewire(aliveServers, knownServers, currentServer, pendingFailures)`. Behaviour unchanged; the rationale that was a comment is now the method's javadoc. Five cases, including the one that motivated the clause — a pending failure forces a retry even when membership has settled back to the stale `KNOWN_SERVERS`, where the membership comparison alone reports "nothing changed". That test asserts its own precondition first so it cannot pass for the wrong reason.

dsolistorres:
## Deployment impact

Moved out of the PR description to keep it readable. This is the evidence for the alerting-noise question: what an operator actually sees on a node boot, measured before and after the startup-grace change.

Verified on a local two-node cluster (PostgreSQL + OpenSearch, `pubsub` transport) built from this branch, because the failure mode this PR adds monitoring for is exactly the kind that generates deployment noise if it fires spuriously.

**Measured, before the grace period was added** — two node boots, one against an empty database and one against a populated one:

| Signal | Fresh boot | Populated DB |
|---|---|---|
| Init window (first poll → transport initialized) | 11s | 3s |
| `cache-transport` WARN from the MONITOR_MODE conversion | 1 | 1 |
| Throttled drop WARN (at a 5s test interval; production default is 30s) | 2 | 1 |
| Invalidations dropped before init | 2,841 | 6 |
| Overall `/dotmgt/health` status during the window | DEGRADED | DEGRADED |
| Readiness/liveness HTTP status | **unaffected** | **unaffected** |

The conversion WARN is one per boot, not one per poll — the check is polled less often than the window lasts. HTTP status was unaffected throughout: `HealthStateManager` treats DEGRADED as ready and the check is never a liveness check, so no probe failure and no restart. But `deriveOverallStatus()` propagates any DEGRADED component to the overall status, so a monitor alerting on `status != UP` would have gone yellow on every pod start.

**Measured again after the fix**, on a rebuilt image and a fresh two-node cluster (`/dotmgt/health` polled once a second from before the container started, 6,110 samples carrying a `cache-transport` component):

| | Fresh boot | Populated DB / joining node |
|---|---|---|
| `cache-transport` DEGRADED polls | **0** of 4,304 | **0** of 1,806 |
| Polls inside the grace window, reported UP with `awaitingInitialization: true` | 17 (11:31:46 → 11:32:06) | 24 (13:26:07 → 13:26:33) |
| `cache-transport` WARN from the MONITOR_MODE conversion | **0** | **0** |
| Throttled drop WARN (5s test interval) | 2 | 1 |
| `droppedInvalidations` in steady state | **0** | **0** |
| `startupDroppedInvalidations` | 2,841 | 6 |
| `initing PubSubCacheTransport` occurrences | 1 | 1 |

Both counters now read exactly as intended: the alertable one is 0 on a healthy node and the boot burst is parked in the startup bucket, matching the one-off INFO the transport logs (`2841 cache invalidation(s) were dropped before it came up`).

Two things the re-measurement corrected in the estimate above. The suppressed window is **17–24 consecutive polls**, not the 3–11s inferred from log timestamps — the health result is cached and served for a while after `init()` succeeds, so the uninitialized verdict persists past the log line. And each of those polls would have been a DEGRADED `cache-transport` component under the previous commit, so the avoided noise is meaningfully larger than first estimated.

Cluster function was re-verified on the same run: bidirectional PING/PONG between both server IDs, one `init()` per node, and a cluster-wide cache flush issued on node 1 returned 200 with `failedInvalidations: 0` and `droppedInvalidations: 0` on both nodes afterwards.

The pre-existing gaps listed under *Known gaps* do not change this. The Micrometer registry is never initialized in the current build, so none of the gauges above are exported yet and no gauge-based alert can fire regardless of what this PR registers.

dsolistorres (approved):

PR #36884, opened 2026-08-04
feat(content-drive): preview a workflow action's items before executing

### LINKED ISSUE #36819: Workflow Center: preview the items a workflow action will run on
### Description

Before a workflow action runs, show the user exactly which contentlets it will apply to, and let them drop some of them.

Choosing an action in the Workflow Center opens a preview: a checkable table of the affected contentlets with enough columns to recognise a row — thumbnail, title, publish status and content type — with every row checked by default. Unchecking a row removes it from the run. Executing fires only the checked rows.

**Scoping the preview to the items the action really applies to.** The bulk endpoint reports counts but never says *which* contentlets sit behind a count, so a single lookup over a mixed selection cannot answer "which of these does Translate apply to". The preview reconstructs it client-side: the selection is grouped by content type, `POST /api/v1/workflow/contentlet/actions/bulk` is called once per group, and each action records which content types contributed it. Because workflow schemes are assigned per content type, that mapping is exact at scheme level — enough to stop a Blog action from listing a `VtlInclude`.

What it deliberately does **not** resolve: an item's step position within a scheme, and Velocity conditions (`conditionPresent`), neither of which the endpoint exposes per item. So the preview can be a superset of what actually succeeds, and the backend still skips what does not apply — reported through `skippedCount` / `skipReason`.

Rows are keyed on **inode**, not identifier. Language variants of one contentlet share an identifier but have distinct inodes, and inodes are what gets fired, so keying on identifier would collapse two variants into one selection entry.

Per-item **execution results** — the spinner / check / error state and retry — are not part of this. They need backend changes to the SSE endpoint and are tracked separately.

### Acceptance Criteria

- [ ] Choosing a workflow action opens a preview of the contentlets it will run on, with a way back to the action list
- [ ] The preview lists only contentlets whose content type can run the chosen action, not the whole selection
- [ ] Every row starts checked; unchecking a row excludes it from the run
- [ ] Executing fires only the checked rows, and the execute affordance reflects how many that is
- [ ] Rows are identified by inode, so two language variants of the same contentlet appear as separate rows
- [ ] Columns are limited to what identifies a row — thumbnail, title, publish status, content type
- [ ] The table paginates only once the list outgrows a page, so a small confirmation step carries no paginator
- [ ] The table body scrolls with its header pinned and the paginator outside the scroll region; the dialog itself does not grow
- [ ] Checkboxes are frozen while an action is in flight
- [ ] Behaviour is defined when every row is unchecked — execution is not possible
- [ ] A selection whose content types produce no grouped result falls back to the whole selection rather than rendering an empty table
- [ ] Unit tests cover the eligibility filtering, the per-content-type merge, and the preview's own selection behaviour

### Priority

Medium

### Additional Context

**Why per-content-type lookups rather than one call.** `POST /api/v1/workflow/contentlet/actions/bulk` aggregates on the `wfstep` field and returns counts per action. There is no response field carrying the identifiers behind a count, and no endpoint that returns "eligible inodes for action X". Grouping by content type is the closest exact answer available without backend work, because scheme assignment is a per-content-type property.

The cost is one request per distinct content type in the selection instead of one request overall. For a typical selection that is a small number; a selection spanning many content types will fan out.

**Naming:** the dialog is presented to users as **Workflow Center** (`content-drive.action-center.header`), while the code and i18n keys keep the `action-center` prefix.

Sibling issues: #36817 (the dialog itself), #36818 (spike on multiple actions per run), #36844 (Lock/Unlock), #36845 (Refresh).

rjvelazco:
Partially addresses #36819.

Adds the **drill-down** half of that issue: picking a workflow action now opens a preview of the contentlets it is about to run on, with a checkbox per row, so the payload can be reviewed and trimmed before firing. The **execution-results** half (per-item spinner / success / error with retry) is **not** in this PR — see Additional Info.

### Proposed Changes

* **Preview step before firing a workflow action.** Selecting an action and hitting *Continue* opens a 4-column table (checkbox / Title / Status / Type) of the contentlets it applies to. *Execute* then fires only the rows still checked. Quick actions are unchanged — they still fire on click with the existing confirm.
* **Firing now honours the preview.** Workflow actions previously sent the **entire** selection and relied on the backend's `skippedCount`. They now send only the checked inodes, which also makes them consistent with quick actions (already sending just their eligible subset).
* **Per-action eligibility resolved by content type.** The bulk endpoint reports counts but never says *which* contentlets an action matches, so the preview used to list every selected item — a Blog action showing a `VtlInclude` row the server was always going to skip. Since workflow schemes are assigned per content type, the lookup now runs **once per content type** in the selection and each action records the types that can run it. Cost is bounded by distinct content types (typically 1–2), not by contentlet count, and the requests run in parallel.
* **One dialog header.** The shared dialog's header lives in the shell, so the preview rendering its own title produced a visibly duplicated header. The body now publishes a drill-down header through the store (`dialogDrillDown`), and the shell's single header retitles to the action name with a count that tracks the checked rows. Only the back control stays in the body, where it can disable itself while an action is in flight.
* **Long titles truncate.** A long title widened the table past the dialog and pushed the Type column behind a horizontal scrollbar. Fixed table layout plus `min-w-0` on the title span; full text kept as the hover title.

Reuses existing building blocks rather than adding new ones — `DotContentletStatusBadgeComponent` and `DotContentThumbnailComponent` from `@dotcms/ui`, PrimeNG table checkboxes, and content-drive's existing 20-rows-per-page convention.

### Checklist
- [x] Tests
- [x] Translations

### Additional Info

**Out of scope here — the other half of #36819.** No per-item execution results. Execute still fires, toasts, refreshes the grid and closes the dialog. Adding per-row status/retry needs a decision on how per-item outcomes are reported (the synchronous `bulkFire` returns aggregate counts; the SSE `_bulkfire` endpoint streams progress but needs a shim, since native `EventSource` cannot POST a body). Worth splitting into its own issue.

**Known limitation: scheme-level, not step-level.** Content-type grouping makes eligibility exact at scheme level, but two contentlets of the same type can sit on different workflow steps and only one step may expose the action. The endpoint still won't say which. The action's `count` remains the authority, so the preview keeps a partial-match warning that fires only on that genuine step-level gap. Making rows individually accurate needs a backend change (returning matching contentlet ids per action).

**Rows are keyed on `inode`, not `identifier`** as the main grid does. Language variants of one contentlet share an identifier but have distinct inodes, and inodes are what gets fired — keying on identifier would collapse two variants into one selection entry. Pinned by a test.

**Reviewers: new i18n keys need a dotCMS restart.** `Language.properties` is read by the backend at startup, so the five new `content-drive.action-center.*` keys render as raw keys against an already-running instance until it restarts (or the language cache is flushed). Not a frontend bug.

**One commit is a revert.** `bd8560ffbc` reverts `0304def518`, a Vue-template reformatting commit that had reformatted 29 unrelated `libs/sdk/vue` files against the repo's Prettier config (`prettier --check` flagged all 29; clean after the revert). The two cancel out, so they contribute nothing to this PR's diff. Root cause worth noting separately: `lint-staged.config.mjs` globs cover `ts,js,mjs,cjs,tsx,jsx` and `json,html,css,scss,md,yaml,yml` — **`.vue` is in neither**, so the pre-commit hook never formats Vue files while an unscoped `nx format:write` does.

**Branch history is noisy.** It carries the pre-squash commits of #36821 plus a merge, all redundant with main's squashed `d595cd8e3c`. The *diff* is unaffected (15 files, all listed above); only the commit list is long.

**Verification**

```
pnpm nx test portlets-content-drive   # 939/939, 30 suites
pnpm nx test data-access              # 751 passed, 11 skipped
pnpm nx lint portlets-content-drive   # clean
pnpm nx format:check                  # clean
pnpm nx build dotcms-ui               # AOT build passes (templates type-check)
```

Also exercised manually against a local backend on Angular 22 / TypeScript 6.0: mixed-type selection, unchecking rows, >20-item pagination, back navigation, and that quick actions still fire on a single click.

### Videos

https://github.com/user-attachments/assets/07481a85-89ed-492d-9bca-f883720e27fa

https://github.com/user-attachments/assets/5c88c503-703b-4ed6-a652-f9f4ec58ce47

This PR fixes: #36819


rjvelazco:
@claude review this PR and check what unit test are valid, what can be removed and propuse new unit tests

hmoreras (approved):

PR #36886, opened 2026-08-04
fix(migration): Site Search highlights, pointer preservation and phase-neutral index tests (#36360)

### LINKED ISSUE #36360: [QA-G17] OpenSearch Migration — Site Search Functional Validation (Phases 0–3, $sitesearch viewtool)

fabrizzio-dotCMS:
## Summary

The weekly [OpenSearch Phase Sweep](https://github.com/dotCMS/core/actions/runs/30736864056) (Sunday 2026-08-02, `main` @ `ce9fd437`) failed 6 of 18 jobs. Triaging those failures turned up three real product defects in the ES→OpenSearch migration — lost Site Search highlights, an opaque activation failure, and content-index writes silently deactivating the Site Search index — while the rest were tests asserting Elasticsearch-shaped facts rather than actual regressions. This fixes the defects and makes the affected assertions meaningful in every migration phase instead of skipping them.

Issue #36360. Includes #36899 (`SearchHit` sealed split), which was stacked on this branch and merged into it.

## Changes

**Backend — Site Search highlights (`SearchHit` and its two shapes, `OSSiteSearchAPI`)**
- Search-result snippets lost their `<em>` emphasis wherever OpenSearch serves reads. The OpenSearch path already asked the engine to highlight the `content` field, but the neutral `SearchHit` had nowhere to carry the fragments, so they were dropped with an explicit TODO.
- `SearchHit` becomes a **sealed interface** over two records (#36899): `ContentSearchHit`, the lean six-component shape every content query produces, and `SiteSearchHit`, the same six plus the per-field highlight fragments. `Builder.build()` picks the shape from the data, so no caller passes a "do I want highlights" flag down through `rawSearch` → `ContentSearchResponse.from` → `SearchHits.from`. Highlights are read through `highlightsFor(field)`, which defaults to an empty list on the interface, so a caller holding a `SearchHit` never has to know which shape it got.
- Both adapters populate it; Elasticsearch's `HighlightField` is flattened from its `Text[]` inside the adapter so the vendor type does not leak. Deserializing the interface goes through a `@JsonCreator` factory that re-uses the same builder rule, so a round-trip through the caches preserves the fragments instead of silently dropping them.
- **Both query shapes are highlighted.** `ESSiteSearchAPI` adds a default `content` highlighter to a caller-supplied JSON body that carries no `highlight` clause; the OpenSearch path only did so for the text-query shape, so `$sitesearch.search($alias, $jsonQuery, ...)` — the path `SiteSearchWebAPI` explicitly branches on — kept losing its emphasis under OpenSearch reads. Body construction moved to `OSSiteSearchAPI.searchBody(...)` and both shapes now get a clause. The check reads the *parsed* body rather than grepping the raw string as the ES path does, so a document value that merely contains the word "highlight" no longer suppresses it; a body that already declares a clause is left untouched, which is why the default leaves the engine's fragment size (the ES branch it mirrors passes none either).

**Backend — activation diagnostics (`ContentletIndexAPIImpl.activateIndex`)**
- A name matching no content slot used to fall through every branch: a silent no-op in phases 0/1/2, and in Phase 3 it saved an *empty* `VersionedIndices`, so the caller got `At least one index must be specified when saving versioned indices for version: 3.X` from deep inside the store — naming neither the input nor the problem.
- Rejected up front with a message stating what was passed and what is accepted. `ESIndexResource` already maps `DotStateException` to a 400, and no production caller reaches here with such a name — `IndexAjaxAction` routes site-search names to `SiteSearchAPI`, and `ESIndexResource.modIndex` 404s on anything outside `listDotCMSIndices()` before the call.

**Backend — Site Search pointer preservation (`ContentletIndexAPIImpl`, `VersionedIndicesImpl`, `OSSiteSearchAPI`)**
- The `siteSearch` slot shares the OpenSearch version row with the content pointers, and `saveIndices()` is a delete-by-version followed by a re-insert. Eight of the ten places that rebuild a `VersionedIndices` for content slots omitted `siteSearch`, which does not leave the slot alone — it **erases** it. So a reindex start, switchover, abort, activate or deactivate of a *content* index silently deactivated the active Site Search index.
- Re-adding the slot at each of those eight sites would have fixed the ten rebuilds that exist today while leaving the contract as "re-list every slot or lose it" — the eleventh rebuild reintroduces the bug with no signal. So the default is inverted instead: `VersionedIndicesImpl.builder(VersionedIndices)` (previously dead code, typed to the impl rather than the interface) seeds a builder from the existing record, and the `Builder` interface exposes the `Optional` overloads the generated builder already had, so **clearing a slot has to be said out loud** and forgetting one preserves it. Every rebuild now goes through a single `rebuildOsStore(existing)`.
- Falls out of that: the two byte-identical `activateIndex` blocks (Phase 3 and the Phase 1/2 mirror) collapse into `mirrorActivateToOsStore`, and `OSSiteSearchAPI`'s hand-rolled five-slot copies — the mirror-image risk, the site-search side dropping the content pointers — use the same copy-builder.
- Behavior is unchanged at every site: the switchover paths still replace the content pointers wholesale, and copying `version()` is a no-op because `loadDefaultVersionedIndices()` only ever loads the `3.X` row.

**Tests — phase-neutral assertions (5 integration classes)**
- Alias lookups go through `SiteSearchAPI.getAliasToIndexMap()`. The content-index router keys OpenSearch entries by their `.os`-tagged physical name and deliberately kept that contract in #36797, which rerouted the six production callers but left the tests behind.
- Default-index checks use `SiteSearchAPI.isDefaultIndex()`; content pointer checks use the phase-aware `getActiveIndexName()`. `IndiciesAPI.loadIndicies()` only ever reads the legacy non-versioned rows, so it reports nothing once activation writes the versioned store.
- Name comparisons strip the vendor tag (`IndexTag.strip`) rather than comparing raw — reporting `.os` from Phase 3 is the intended display contract, and `endsWith(logicalName)` can never match a `.os`-suffixed name.
- `SiteSearchJobImplTest` clears indices through `SiteSearchAPI.deleteIndex()` (idempotent per engine) instead of the content router. This was the actual cause of the "Site Search returns 0 results in Phase 3" failures: `OSIndexAPIImpl.delete()` applies the cluster prefix but **not** the `.os` tag and treats index-not-found as a successful no-op, so the old loop deleted nothing while reporting success, and the test then searched a stale leftover index.
- `optimize` receives the physical names the engine holds, matching production callers that feed it names from `getIndices()` — it dispatches by vendor tag.
- `ESIndexAPITest.testDeleteOldIndices` normalizes the retention listing to logical names. Retention itself is correct: each provider preserves its active set from its own store.
- `test_createSiteSearchIndex_shouldBePossibleToAddMoreThan100` now deletes its 115 indices; leaving them behind polluted every later alias assertion in the same JVM. The two reindex tests build their index name with the `_` separator that production and every other test use.

**Tests — new unit coverage**
- `SearchHitTest` (9): the highlight mapping, the empty-map default, the shape the builder picks, a field explicitly mapped to `null`, and the Jackson round-trip through the interface in both shapes.
- `ContentletIndexAPIImplPhaseTest` (+4): the site-search slot survives activate (Phase 3 and the Phase 2 mirror) and deactivate, including the edge where the content slot was the last populated one — that used to drop the whole version row and take the active Site Search index with it. The fake's `removeVersion` throws, so that regression fails loudly rather than quietly.
- `OSSiteSearchBodyTest` (4): the four body shapes, including a JSON body that mentions "highlight" only inside a document value.

## Testing

Verified against a live stack (PostgreSQL + Elasticsearch 7.10 + OpenSearch 3.4) — the five affected classes, 61 tests, under **each** migration phase, re-run on the current head:

| | Phase 1 | Phase 2 | Phase 3 |
|---|---|---|---|
| Failures | 0 | 0 | 0 |
| Skipped | 0 | 0 | 5 |

The 5 skipped in Phase 3 are the ES-retention tests the class itself skips there (Elasticsearch is decommissioned); they all run and pass in Phase 2, so that fix is covered.

To reproduce, per phase:

```bash
just test-integration-phase 3 ContentletIndexAPIImplTest,ESIndexAPITest,ESSiteSearchAPITest,SiteSearchJobImplTest,ContentTypesPaginatorTest
```

Unit tests: 261 green across the content-index, site-search and reindex packages, including the new coverage above —
`./mvnw test -pl :dotcms-core -Dtest=SearchHitTest,ContentletIndexAPIImplPhaseTest,OSSiteSearchBodyTest`.

One pre-existing flake, unrelated to these changes, surfaces intermittently in phases 2 and 3: `ContentletIndexAPIImplTest.test_that_live_and_working_content_makes_it_into_the_index` times out after 10s waiting for a document to become visible, and passes on rerun. Possibly a refresh/visibility gap on the OpenSearch write path — not investigated here.

## Breaking Changes

**None.** `SearchHit` becomes a sealed interface, so it is no longer directly constructible; the only construction sites are its own builder and the two vendor adapters, all in this repo, and the serialized JSON shape is unchanged for content hits (a highlighted hit gains a `highlights` key). `VersionedIndicesImpl.builder(VersionedIndicesImpl)` widens to `builder(VersionedIndices)` — a method that had no callers. `activateIndex` now throws `DotStateException` for a name that maps to no content slot, where it previously did nothing; the REST layer already renders that as a 400, and no production caller passes such a name.

## Deliberately out of scope

- **A bare-name delete in Phase 3 reports success without deleting anything** (`OSIndexAPIImpl.delete` omits the `.os` tag). That silence is worse than the Phase 1 exception, and it is the same bug family as the open #36825 — changing it in parallel would conflict.
- **`removeClusterIdFromName(null)` returns `""`**, turning "no index" into an index named the empty string. This is what made the activation failure reachable; changing the contract touches every caller and cannot be validated without the full battery. The new guard makes the symptom loud instead of silent.
- **A site-search index name without the `_` separator never resolves its alias** — the index-name timestamp parsing keys off it, and that same parsing feeds retention. Production always generates `sitesearch_<timestamp>`, so this is only reachable with an arbitrary name.
- **The load-then-save window in the store rebuilds.** Both sides read the version row and write it back whole, so a site-search activation landing between a content rebuild's read and its write is still lost. Strictly better than before (where the slot was erased unconditionally), but the real fix is a per-slot update rather than delete-and-reinsert.

This PR fixes: #36360


dotCMS/src/main/java/com/dotcms/content/index/domain/SearchHit.java
```diff
@@ -51,7 +56,8 @@ public record SearchHit(
         @JsonProperty("sourceAsMap") Map<String, Object> getSourceAsMap,
         @JsonProperty("score") float getScore,
         @JsonProperty("fields") Map<String, Object> getFields,
-        @JsonProperty("sortValues") List<Object> getSortValues) {
+        @JsonProperty("sortValues") List<Object> getSortValues,
+        @JsonProperty("highlights") Map<String, List<String>> getHighlights) {
```
wezell: Is it for all search results, not just site search results? It feels to me like this could lead to a performance regression/issues. We don't care about highlighted fields when pulling content results only when pulling site search results (nor do we care about some of the other fields, really).  Wondering if we could make the highlight fields lazy loaded on-demand rather than populating them at instantiation?

wezell (changes requested):
See my comment.

wezell:
For what it’s worth, I think we could separate the content search results from the site search results. Content search results are super important and should be small, performant, and fast, almost records. Site search results are different and can be heavier.  Really two different functions

fabrizzio-dotCMS:
Thanks — you're right about the direction, so I built it: **#36899** separates the two shapes, stacked on this branch. Details below, plus what the field actually costs today and why the lazy variant would have gone the wrong way.

### The separation (#36899)

`SearchHit` becomes a sealed interface over two records:

```java
public sealed interface SearchHit permits ContentSearchHit, SiteSearchHit {
    // ... the six neutral accessors ...
    default List<String> highlightsFor(String field) { return List.of(); }
}
```

- **`ContentSearchHit`** — six components, no highlight state at all. What every content query produces.
- **`SiteSearchHit`** — the same six plus the fragments.

The `default` is what makes it cheap: the concern is declared once on the contract, costs zero bytes per instance, and a caller holding a plain `SearchHit` never has to know or ask which shape it got.

Callers don't choose the shape either — `Builder.build()` picks it from the data, returning the lean record unless the engine actually returned fragments. That detail matters more than it looks: Site Search and content search reach the neutral layer through the *same* factory (`ContentSearchResponse.from`, since Site Search reads via `rawSearch` after #36398), so a "do I want highlights" flag would have had to be threaded through `rawSearch` → `ContentSearchResponse.from` → `SearchHits.from` → the hit adapter. Deciding from the response keeps it in one place.

Blast radius turned out to be nil outside the domain package: `compile -pl :dotcms-core` passes with zero changes anywhere else, and the enterprise consumer really does go through the contract (`invokeinterface SearchHit.highlightsFor` in the bytecode). `ContentSearchToolTest` — the VTL guard that evaluates customer templates through the real Velocity engine — is 11/11 green against it.

### What the field costs today

Worth recording since it shaped how I scoped this: on the content path the shared field was already free rather than a regression. Highlighting only produces data if the query asks for it, and the only three places that ask are all Site Search (`ESSiteSearchAPI:251`, `:257`, `OSSiteSearchAPI:344`). With no highlight section in the response the adapter returns the shared `Map.of()` singleton, so a content hit carried one reference field and one `isEmpty()` check — no allocation, nothing extra parsed or on the wire. It also never reached the query cache: `ESQueryCache` stores Elasticsearch's own `SearchHits`, not the neutral type.

So #36899 is worth doing for the reason you gave second — cohesion, content results not advertising a Site Search concern — rather than to recover measurable overhead. Side benefit: `ContentSearchHit` no longer serializes an empty `"highlights":{}`.

### Why not lazy

The fragments only exist inside the vendor hit (`org.elasticsearch.search.SearchHit` / OpenSearch's `Hit`). Deferring the read means holding a reference to that object inside the neutral record, which leaks the vendor type the neutral layer exists to hide and keeps the engine's whole response graph reachable — more retained heap, not less. And there's no second round-trip to defer: highlights arrive in the same response or not at all.

### Two things I'd flag on the split

- **Jackson.** `SearchHit`/`SearchHits` are deserialized and back caches/REST paths, and an interface can't be instantiated. I used `@JsonDeserialize(as = ContentSearchHit.class)`, which keeps `readValue(json, SearchHit.class)` working; the tradeoff is that a serialized Site Search hit would round-trip back as a content hit and lose its fragments. Nothing serializes one today. The alternative, polymorphic `@JsonTypeInfo`, would add a type discriminator to a JSON shape those caches depend on, which seemed the worse trade.
- **`SearchHit` goes from a record to an interface**, so anything compiled against the concrete type needs recompiling. Nothing in the repo does, and the type is internal to the neutral search layer — but flagging it in case you know of plugin code that shouldn't be.

Also worth noting on your framing: the split is already half-there in the direction you describe. The heavy carrier is `SiteSearchResult` (a `HashMap`-backed bag) and content results never pass through it — what was shared is only the engine adapter, which is what #36899 splits.

Happy either way on sequencing: land this PR and #36899 behind it, or if you'd rather the field never land at all, I can pull the highlights commit out of here and ship the two pointer fixes first. The tradeoff on that second path is that the Site Search phase-2/3 test stays red until the split merges.

dotCMS/src/main/java/com/dotcms/content/index/domain/SearchHit.java
wezell: Nice, this is resolved with the new work.

wezell (approved):
Looks better with the change

PR #36889, opened 2026-08-04
fix(content-drive): expose folder permissions and edit-dialog fields on folder search #36595

### LINKED ISSUE #36595: [BUG] Content Drive: Sidebar folders don't open a right-click context menu like the table
## Problem Statement

In the **Content Drive** portlet, right-clicking a folder in the left **side-panel folder tree** does nothing. On the main **table**, right-clicking a folder (or using its kebab menu) opens a context menu with folder actions (**Edit folder**, **Edit permissions**). Users reasonably expect the same right-click menu on the sidebar tree, but there is no equivalent wiring, so folder actions are unreachable from the tree.

**Impact:** Usability gap / inconsistency — folder actions are only reachable from the table, not from the sidebar tree where users spend most of their navigation time. Front-end (Angular admin UI).

### Root cause / blocker

The two views are backed by different data:

| View | Source | Folder object |
|---|---|---|
| Table (`dot-folder-list-view`) | `POST /api/v1/drive/search` | Full `DotContentDriveFolder` **including `permissions`** (+ `title`, `sortOrder`, `filesMasks`, `defaultFileType`, `showOnMenu`, …) |
| Sidebar tree (`dot-tree-folder` / `dot-content-drive-sidebar`) | `POST /api/v1/folder/search` | `FolderSearchResultView` — only `id`, `inode`, `path`, `hostName`, `addChildrenAllowed` — **no permissions, none of the full folder fields** |

The shared context menu (`dot-folder-list-context-menu`) is driven by the store (`patchContextMenu({ triggeredEvent, contentlet })`) and **gates** its items on `contentlet.permissions` (EDIT → *Edit folder*, EDIT_PERMISSIONS → *Edit permissions*). The *Edit folder* dialog also needs the full folder fields as its payload.

So simply adding a `(contextmenu)` handler to the tree node is **not enough** — with the byPath data the menu would render empty (permission gate fails) and the edit dialog would be missing fields.

### Proposed resolution options (team to decide)

1. **Backend:** add `permissions` (and the folder fields the edit dialog needs) to `POST /api/v1/folder/byPath` / `FolderSearchResultView`, giving the sidebar parity with the table.
2. **Frontend:** on right-click, fetch the full folder + permissions before opening the menu (extra request per right-click; needs an endpoint that returns permissions for a single folder).
3. **Frontend:** show the folder actions ungated in the sidebar and let the backend enforce permissions when the action runs (diverges from the table's permission-aware menu).

## Steps to Reproduce

1. Open **Content Drive**.
2. In the left side-panel, expand the folder tree.
3. **Right-click** any folder in the tree.
4. **Expected:** a context menu appears with the same folder actions as the table (Edit folder, Edit permissions).
5. **Actual:** nothing happens — no context menu.

For comparison: right-clicking a folder **row in the table** opens the context menu as expected.

## Acceptance Criteria

- [ ] Right-clicking a folder in the side-panel tree opens the folder context menu.
- [ ] The menu offers the same folder actions available from the table (Edit folder, Edit permissions), subject to the user's permissions on that folder.
- [ ] Menu items respect the user's permissions (an approach for sourcing folder permissions in the sidebar is chosen from the options above).
- [ ] Selecting **Edit folder** opens the folder dialog pre-populated with that folder's data.
- [ ] Behavior is consistent with the table's context menu (same items, same gating).

## dotCMS Version

Latest from `main` branch.

## Severity

Low - Minor issue or cosmetic

## Notes

Split out from the same work as the "New button — create content in the current folder" fix (Content Drive). Relevant files:
- `core-web/libs/portlets/dot-content-drive/ui/src/lib/dot-tree-folder/`
- `core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-sidebar/`
- `core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-folder-list-context-menu/`
- `dotCMS/.../rest/api/v1/folder/FolderResource#byPath` / `FolderSearchResultView`

ihoffmann-dot:
### Parent Issue

#36595

### Summary

Backend half of #36595. Right-clicking a folder in the Content Drive **sidebar tree** does nothing, while right-clicking a folder **row in the table** opens a context menu (Edit folder / Edit permissions). The shared `dot-folder-list-context-menu` component gates those items on `contentlet.permissions` and pre-populates the "Edit folder" dialog from the folder object it is handed — but the sidebar is backed by `GET /api/v1/folder/search` → `FolderSearchView`, which carried neither.

This PR enriches that view so the sidebar can gate the menu and pre-populate the dialog exactly the way the table does. **No FE changes here** — the FE leg is the other half of the same issue, and the contract obligations it inherits are written up in [a comment on #36595](https://github.com/dotCMS/core/issues/36595).

> The issue's "Proposed resolution options" prose names `POST /api/v1/folder/byPath` / `FolderSearchResultView`. That endpoint is deprecated (`forRemoval = true`) with zero references in `core-web`; the sidebar actually calls `GET /api/v1/folder/search`, which is what this PR extends. The comparison table in the issue body already says this correctly.

### What changed

**Always present (no flag)** — the fields the "Edit folder" dialog reads, all read straight off the `Folder` already loaded by the existing `SELECT folder.*`, so zero additional DB cost:

`title`, `sortOrder`, `filesMasks`, `defaultFileType`, `showOnMenu`

**Opt-in via `?includePermissions=true`:**

`permissions: List<String>` — the permission types the requesting user holds, drawn from `READ, EDIT, PUBLISH, EDIT_PERMISSIONS, CAN_ADD_CHILDREN`. Same field name, same 5-type meaning and spelling as the table's `DotContentDriveFolder.permissions`, so one gating implementation serves both views. `null` when not requested — deliberately not `[]`, so the FE can distinguish "not fetched" from "fetched, no grants".

Returning the full 5-type set rather than only the two the menu reads today costs **3** extra batch calls instead of 2 (`READ` is implicit — the folder passed the pre-pagination READ filter to be here; `CAN_ADD_CHILDREN` reuses the set already computed for `addChildrenAllowed`). That one extra call buys away a whole ambiguity class: if `permissions` meant 5 types on a table-sourced folder and 2 on a sidebar-sourced one, a future `PUBLISH` check would silently evaluate `false` depending on provenance, and the symptom is a **missing menu action**, not an error.

**New 400 guardrail.** `PaginationUtil` puts no upper bound on `perPage`, and `PermissionBitFactoryImpl.getPermittedIds` chunks ids by 500 — so `?perPage=10000&includePermissions=true` was reachable by any authenticated backend user at 60 extra queries. With the flag on, `perPage` is now capped by `content.drive.folder.search.permissions.max.per.page` (default 200); above it the request is rejected with a 400 naming the cap. Rejecting rather than silently dropping the flag is the point: dropping it yields `permissions: null` → "no grants" → an empty menu, indistinguishable from a user who genuinely has none. **With the flag off, any `perPage` behaves exactly as before** — there is a test asserting precisely that.

All new batch calls are scoped to `page`, the post-pagination slice, matching the existing `CAN_ADD_CHILDREN` call. With the flag off they issue no query at all.

### OpenAPI

`FolderSearchView` was **undocumented** — 0 occurrences in `openapi.yaml`, because the endpoint declared the generic `ResponseEntityPaginatedDataView` whose `entity` is a bare `type: object`. Regenerating alone would have produced no field diff at all.

This PR adds a concrete `ResponseEntityFolderSearchView`, references it from the `@ApiResponse`, and adds `@Schema` descriptions to every field. The regenerated yaml now publishes the full 13-field schema plus the `includePermissions` parameter — a real contract for the FE leg to build against.

### Rollback

**M-3 (REST API contract change), 🟢 LOW.** Purely additive: new response fields plus a new query param whose default reproduces current behavior exactly. No DB schema change, no migration, no index change. Nothing calls `includePermissions=true` until the FE leg ships, so there is no mixed-version window for this change on its own.

### Payload cost — measured

Six new fields ride every row, and `"permissions": null` serializes even with the flag off (the default mapper has no `NON_NULL` inclusion). Measured on a representative deep-link hydration payload — 10,000 folders, flag off — serialized with `DotObjectMapperProvider.createDefaultMapper()`:

| shape | raw | gzip |
|---|---|---|
| before (7 fields) | 2.19 MB · 229 B/row | 0.49 MB · 51.7 B/row |
| **after (13 fields)** | 3.66 MB · 384 B/row · **+67.5%** | 0.54 MB · 56.9 B/row · **+9.9%** |
| alt: detail fields gated behind the flag | 2.37 MB · +8.3% | 0.50 MB · +0.5% |
| alt: `@JsonInclude(NON_NULL)` on the record | 3.09 MB · +41.4% | 0.53 MB · +7.6% |

The raw +67.5% collapses to **+9.9% (+52 KB)** over the wire — the added fields are highly repetitive and compress well. Not material, so "detail fields always present" stands. Alternatives priced above in case a reviewer reads the raw number differently; note `NON_NULL` would also stop serializing `defaultBaseType: null`, a contract change on an existing field.

### Testing

54 tests green locally.

| Class | Result |
|---|---|
| `FolderSearchPaginatorTest` (unit) | 10/10 |
| `FolderResourceSearchTest` | 21/21 |
| `FolderAPIImplFilterTest` | 16/16 |
| `FolderFactoryImplFilterTest` | 7/7 |

Plus 4 new Postman requests on `/v1/folder/search` (flag on, flag omitted, over-cap 400, over-cap without the flag).

**Every permission-gating test runs as a limited, non-admin user.** `filterCollection` short-circuits for CMS Admin and the system user (`PermissionBitAPIImpl.java:1433-1434`), so the same assertions run as an admin would exercise none of this logic and would pass against an implementation that computes nothing. Coverage: all 5 types granted; READ + EDIT only with `PUBLISH`/`EDIT_PERMISSIONS`/`CAN_ADD_CHILDREN` asserted **absent**; READ only; permissions correct for page 2 of a paginated result rather than page 1's; read-your-writes after a grant with no reindex; exact string assertions on every type name.

### Two pre-existing test failures fixed here

Both were red on `main` before this PR and are unrelated to it, but one lives in a file this PR edits and would have read as a regression from this change:

- **`FolderAPIImplFilterTest.test_searchFolders_withPathScope_returnsOnlyDescendants`** — created a folder named `assets` directly under a site; `assets` is in the default `RESERVEDFOLDERNAMES` (`FolderAPIImpl:110`), so the save threw `InvalidFolderNameException`. Renamed to `site-assets`. Introduced by 27191bc6a6.
- **`FolderFactoryImplFilterTest`** — all 7 tests threw `NullPointerException: user is required`. The class never passed `.user(...)`, and `FolderSearchParams.Builder.build()` began requiring one in 6c92187cfc. Added the system user, with a comment noting the factory resolves in SQL and never reads it — permission filtering happens a layer up in `FolderAPIImpl`.

### Notable landmines documented in code

- **`PermissionAPI.Type.CANONICAL_TYPES` cannot be used to derive these names.** It is `{READ, WRITE, PUBLISH, EDIT_PERMISSIONS, CAN_ADD_CHILDREN}` — the `WRITE` alias, not `EDIT`. Deriving from it emits `"WRITE"`, the FE's `EDIT` check silently fails, and the symptom is a missing menu item. The implementation uses explicit `Type.EDIT.name()` etc., and a test asserts `WRITE` is never emitted.
- **`HttpStatusCodeException` runs `String.format` over the message you hand it**, so pre-formatting would format twice and throw on a literal `%`. The cap error uses the varargs form.

### Explicitly out of scope

- Any FE change (the other half of #36595).
- `DotFolderTransformerImpl.contentDriveView`'s per-folder N+1 permission loop — real inefficiency, different code path.
- `POST /api/v1/drive/search` / `BrowserAPIImpl` — untouched.
- The deprecated `POST /api/v1/folder/byPath` — not extended.
- The pre-existing `READ` filter running over the full pre-pagination set (`FolderAPIImpl:819`), which makes endpoint load scale with site size rather than `perPage`. It predates this ticket and is the endpoint's dominant cost; worth its own ticket if folder-search latency shows up in profiling.

This PR fixes: #36595

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/util/pagination/FolderSearchPaginator.java
+++ b/dotCMS/src/main/java/com/dotcms/util/pagination/FolderSearchPaginator.java
@@ -6,9 +6,9 @@ import com.dotmarketing.portlets.folders.business.FolderAPI;
 import com.dotmarketing.portlets.folders.business.FolderSearchParams;
 import com.dotmarketing.util.PaginatedArrayList;
 import com.dotmarketing.util.UtilMethods;
-import com.google.common.annotations.VisibleForTesting;
 import com.liferay.portal.model.User;
 import java.util.Map;
+import org.apache.commons.lang3.BooleanUtils;
 
 /**
  * {@link PaginatorOrdered} implementation for the unified folder search endpoint.
@@ -18,20 +18,25 @@ import java.util.Map;
  * <p>Extra params expected in the map (keyed by the query param names defined in
  * the calling resource): {@code "siteId"}, {@code "path"}, {@code "recursive"},
  * {@code "includePermissions"}.
+ *
+ * @param folderAPI the API the search is delegated to; injectable for testing
  */
-public class FolderSearchPaginator implements PaginatorOrdered<FolderSearchView> {
+public record FolderSearchPaginator(FolderAPI folderAPI)
+        implements PaginatorOrdered<FolderSearchView> {
 
     private static final String DEFAULT_ORDER_BY_COLUMN = "folder.name";
 
-    private final FolderAPI folderAPI;
+    private static final String SITE_ID_PARAM = "siteId";
+    private static final String PATH_PARAM = "path";
+    private static final String RECURSIVE_PARAM = "recursive";
+    private static final String INCLUDE_PERMISSIONS_PARAM = "includePermissions";
 
+    /**
+     * Production entry point — the canonical constructor taking a {@link FolderAPI} exists for
+     * tests to inject a mock.
+     */
     public FolderSearchPaginator() {
         this(APILocator.getFolderAPI());
-    }
-
-    @VisibleForTesting
-    public FolderSearchPaginator(final FolderAPI folderAPI) {
-        this.folderAPI = folderAPI;
     }
 
     @Override
@@ -40,11 +45,11 @@ public class FolderSearchPaginator implements PaginatorOrdered<FolderSearchView>
             final OrderDirection direction, final Map<String, Object> extraParams)
             throws PaginationException {
 
-        final var ep = extraParams != null ? extraParams : Map.of();
-        final String siteId   = (String) ep.get("siteId");
-        final String path     = (String) ep.getOrDefault("path", "/");
-        final boolean recursive = Boolean.TRUE.equals(ep.get("recursive"));
-        final boolean includePermissions = Boolean.TRUE.equals(ep.get("includePermissions"));
+        final Map<String, Object> ep = extraParams != null ? extraParams : Map.of();
+        final String siteId = (String) ep.get(SITE_ID_PARAM);
+        final String path = (String) ep.getOrDefault(PATH_PARAM, "/");
+        final boolean recursive = toBoolean(ep.get(RECURSIVE_PARAM));
+        final boolean includePermissions = toBoolean(ep.get(INCLUDE_PERMISSIONS_PARAM));
 
         final String orderByColumn = switch (orderBy) {
             case "mod_date" -> "folder.mod_date";
@@ -69,5 +74,20 @@ public class FolderSearchPaginator implements PaginatorOrdered<FolderSearchView>
         } catch (final Exception e) {
             throw new PaginationException(e);
         }
+    }
+
+    /**
+     * Reads a boolean out of the untyped extra-params map.
+     *
+     * <p>The resource puts real {@link Boolean} values in, but the map is {@code Map<String, Object>}
+     * and other callers may hand over the raw query string, so string forms are accepted too —
+     * {@code BooleanUtils.toBoolean(String)} covers "true"/"yes"/"on"/"1" and their variants.
+     * Anything unrecognised, including {@code null}, reads as {@code false}.
+     */
+    private static boolean toBoolean(final Object value) {
+        if (value instanceof Boolean booleanValue) {
+            return booleanValue;
+        }
+        return value != null && BooleanUtils.toBoolean(String.valueOf(value));
     }
 }
--- a/dotCMS/src/main/java/com/dotcms/util/pagination/FolderSearchPaginator.java
+++ b/dotCMS/src/main/java/com/dotcms/util/pagination/FolderSearchPaginator.java
@@ -16,7 +16,8 @@ import java.util.Map;
  * path scoping, and recursive depth control.
  *
  * <p>Extra params expected in the map (keyed by the query param names defined in
- * the calling resource): {@code "siteId"}, {@code "path"}, {@code "recursive"}.
+ * the calling resource): {@code "siteId"}, {@code "path"}, {@code "recursive"},
+ * {@code "includePermissions"}.
  */
 public class FolderSearchPaginator implements PaginatorOrdered<FolderSearchView> {
 
@@ -43,6 +44,7 @@ public class FolderSearchPaginator implements PaginatorOrdered<FolderSearchView>
         final String siteId   = (String) ep.get("siteId");
         final String path     = (String) ep.getOrDefault("path", "/");
         final boolean recursive = Boolean.TRUE.equals(ep.get("recursive"));
+        final boolean includePermissions = Boolean.TRUE.equals(ep.get("includePermissions"));
 
         final String orderByColumn = switch (orderBy) {
             case "mod_date" -> "folder.mod_date";
@@ -61,6 +63,7 @@ public class FolderSearchPaginator implements PaginatorOrdered<FolderSearchView>
                     .offset(offset)
                     .orderBy(orderByColumn)
                     .orderDirection(orderDirection)
+                    .includePermissions(includePermissions)
                     .build();
             return folderAPI.searchFolders(params);
         } catch (final Exception e) {
--- a/dotCMS/src/main/java/com/dotmarketing/portlets/folders/business/FolderAPIImpl.java
+++ b/dotCMS/src/main/java/com/dotmarketing/portlets/folders/business/FolderAPIImpl.java
@@ -811,78 +811,119 @@ public class FolderAPIImpl implements FolderAPI  {
 	public PaginatedArrayList<FolderSearchView> searchFolders(final FolderSearchParams params)
 			throws DotDataException, DotSecurityException {
 
-		// 1. Load all matching folders (no LIMIT — pagination happens in Java so that
-		//    permission filtering does not produce short pages for limited users).
-		final var all = folderFactory.searchFolders(params);
+		// Load all matching folders (no LIMIT — pagination happens in Java so that permission
+		// filtering does not produce short pages for limited users), then batch-filter by READ in a
+		// single SQL round-trip for the whole collection.
+		final List<Folder> readable = permissionAPI.filterCollection(
+				folderFactory.searchFolders(params), PermissionAPI.PERMISSION_READ,
+				params.user(), params.respectFrontendRoles());
 
-		// 2. Batch READ filter — one SQL round-trip for the entire collection.
-		final var filtered = permissionAPI.filterCollection(
-				all, PermissionAPI.PERMISSION_READ, params.user(), params.respectFrontendRoles());
+		final int total = readable.size();
+		final List<Folder> page = paginate(readable, params);
 
-		// 3. Java pagination — applied after permission filtering for correctness.
-		final int total = filtered.size();
-		final int from  = Math.min(params.offset(), total);
-		final int to    = params.limit() < 0 ? total : Math.min(params.offset() + params.limit(), total);
-		final var page  = filtered.subList(from, to);
-
-		// 4. Batch CAN_ADD_CHILDREN check — only on the page (≤ perPage items).
-		final Set<String> canAddIds = permissionAPI.filterCollection(
-				page, PermissionAPI.PERMISSION_CAN_ADD_CHILDREN, params.user(), params.respectFrontendRoles())
-				.stream().map(Permissionable::getPermissionId).collect(Collectors.toSet());
-
-		// 5. Opt-in permission types, each a batch call scoped to the page. READ needs no call (the
-		//    folder is here because it passed the READ filter in step 2) and CAN_ADD_CHILDREN reuses
-		//    step 4, so the full 5-type set costs 3 extra calls.
-		final Set<String> editableIds        = filterIdsByPermission(page, PermissionAPI.PERMISSION_EDIT, params);
-		final Set<String> publishableIds     = filterIdsByPermission(page, PermissionAPI.PERMISSION_PUBLISH, params);
-		final Set<String> editPermissionsIds = filterIdsByPermission(page, PermissionAPI.PERMISSION_EDIT_PERMISSIONS, params);
-
-		// 6. Determine which folders have permission-visible children.
+		final PagePermissions permissions = resolvePagePermissions(page, params);
 		final Set<String> parentPathsWithVisibleChildren =
 				findParentPathsWithVisibleChildren(page, params);
 
-		// 7. Map to views.
-		final var views = page.stream()
-				.map(folder -> {
-					final var fullPath   = folder.getPath();
-					final var parentPath = fullPath.substring(0, fullPath.length() - folder.getName().length() - 1);
-					final List<String> permissions = params.includePermissions()
-							? permissionNames(folder, canAddIds, editableIds, publishableIds, editPermissionsIds)
-							: null;
-					return new FolderSearchView(
-							folder.getIdentifier(), folder.getInode(), folder.getName(),
-							parentPath, canAddIds.contains(folder.getPermissionId()),
-							parentPathsWithVisibleChildren.contains(folder.getPath()),
-							folder.getDefaultBaseType(),
-							folder.getTitle(), folder.getSortOrder(), folder.getFilesMasks(),
-							folder.getDefaultFileType(), folder.isShowOnMenu(),
-							permissions);
-				})
-				.toList();
-
-		final var result = new PaginatedArrayList<FolderSearchView>();
+		final PaginatedArrayList<FolderSearchView> result = new PaginatedArrayList<>();
 		result.setTotalResults(total);
-		result.addAll(views);
+		result.addAll(toViews(page, permissions, parentPathsWithVisibleChildren, params));
 		return result;
 	}
 
 	/**
-	 * Batch-resolves which folders in {@code page} the requesting user holds {@code permission} on,
-	 * returning their permission ids. Returns an empty set — issuing no query at all — when
-	 * {@link FolderSearchParams#includePermissions()} is {@code false}.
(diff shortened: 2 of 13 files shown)


dario-daza (approved):

dotCMS/src/main/java/com/dotmarketing/portlets/folders/business/FolderAPIImpl.java
```diff
@@ -830,20 +830,33 @@ public PaginatedArrayList<FolderSearchView> searchFolders(final FolderSearchPara
...
-		// 6. Map to views.
+		// 7. Map to views.
 		final var views = page.stream()
 				.map(folder -> {
 					final var fullPath   = folder.getPath();
 					final var parentPath = fullPath.substring(0, fullPath.length() - folder.getName().length() - 1);
+					final List<String> permissions = params.includePermissions()
+							? permissionNames(folder, canAddIds, editableIds, publishableIds, editPermissionsIds)
+							: null;
 					return new FolderSearchView(
```
fabrizzio-dotCMS: Too many params are usually a code smell. It's okay, but a builder pattern would work better here.

dotCMS/src/main/java/com/dotcms/util/pagination/FolderSearchPaginator.java
```diff
@@ -43,6 +44,7 @@ public PaginatedArrayList<FolderSearchView> getItems(final User user, final Stri
         final String siteId   = (String) ep.get("siteId");
         final String path     = (String) ep.getOrDefault("path", "/");
         final boolean recursive = Boolean.TRUE.equals(ep.get("recursive"));
+        final boolean includePermissions = Boolean.TRUE.equals(ep.get("includePermissions"));
```
fabrizzio-dotCMS: I like org.apache.commons.lang3.BooleanUtils.toBoolean(String) deals with so many cases for you

dotCMS/src/main/java/com/dotcms/util/pagination/FolderSearchPaginator.java
```diff
@@ -16,7 +16,8 @@
  * path scoping, and recursive depth control.
  *
  * <p>Extra params expected in the map (keyed by the query param names defined in
- * the calling resource): {@code "siteId"}, {@code "path"}, {@code "recursive"}.
+ * the calling resource): {@code "siteId"}, {@code "path"}, {@code "recursive"},
+ * {@code "includePermissions"}.
  */
 public class FolderSearchPaginator implements PaginatorOrdered<FolderSearchView> {
```
fabrizzio-dotCMS, line 26: FolderSearchPaginator can be a record

fabrizzio-dotCMS (commented):
Reviewed the full change (Java, OpenAPI, integration/unit tests, Postman) and traced the claims in the description. **No correctness bugs found** — the implementation does what it says. Four findings below; only the first is worth acting on before merge.

---

### 🟠 1. The most valuable new tests never run in CI

`FolderAPIImplFilterTest` and `FolderFactoryImplFilterTest` are not registered in any `@SuiteClasses` suite, and the failsafe config in `dotcms-integration/pom.xml:485-489` only includes `**/MainSuite1*.java`, `**/MainSuite2*.java`, `**/MainSuite3*.java` and `**/Junit5Suite*.java`. An integration test class outside a suite is simply never executed.

That is where 5 of the new tests live — precisely the ones that justify the design:

- `test_searchFolders_includePermissions_allTypesGranted` (the `EDIT` vs `WRITE` parity assertion)
- `test_searchFolders_includePermissions_readAndEditOnly` (the explicit-absence assertions the FE gating depends on)
- `test_searchFolders_includePermissions_readOnly`
- `test_searchFolders_includePermissions_scopedToRequestedPage`
- `test_searchFolders_includePermissions_readYourWrites`
- plus `test_searchFolders_detailFields_alwaysPresent`

This is also the reason the two "pre-existing failures" fixed in this PR went unnoticed on `main`: those classes have never been executed by CI, so nothing could have reported them red.

Suggested fix: register both classes in `MainSuite2b`, where `FolderResourceSearchTest` already lives (`dotcms-integration/src/test/java/com/dotcms/MainSuite2b.java:530`).

### 🟡 2. The `perPage` cap is validated against the raw parameter, not the effective one

`FolderResource.searchFolders` checks `perPage > permissionsMaxPerPage` on the value as received. `PaginationUtil` then normalizes it: `final int perPageValue = utilParams.perPage() <= 0 ? perPageDefault : utilParams.perPage()` (`PaginationUtil.java:246`), where `perPageDefault = Config.getIntProperty(DOTCMS_PAGINATION_ROWS, 10)`.

So `?includePermissions=true&per_page=0` (or a negative value) passes the guard and then pages at `DOTCMS_PAGINATION_ROWS`. With the shipped default of 10 this is harmless, but the invariant the guard exists to enforce — "permissions are never resolved for a page larger than the cap" — ends up depending on an unrelated configuration property rather than being guaranteed. A deployment that raises `DOTCMS_PAGINATION_ROWS` above 200 silently reopens the hole the cap was added to close.

Validating the effective value instead (`perPage <= 0 ? Config.getIntProperty(DOTCMS_PAGINATION_ROWS, 10) : perPage`) makes the guard self-contained. Worth noting the automated review above asserted this path "can never exceed the cap" — that holds only under the default configuration.

Related, lower priority: the cap lives in the resource, so any other caller of `FolderAPI.searchFolders` with `includePermissions(true)` and a large `limit` is unbounded. If the bound is a property of the operation rather than of the HTTP surface, the API layer is the more durable place for it.

### 🔵 3. The 400 is thrown before authentication

The cap check runs ahead of `new WebResource.InitBuilder(...).init()`, so an unauthenticated caller gets a 400 — not the documented 401 — and the message discloses the configured cap value. This matches the existing ordering for the `siteId`/`name` validations, so it is consistent with the file; the difference is that this message leaks a configuration value. Moving the check below `init()` costs nothing.

### 🔵 4. Four sequential `filterCollection` calls per page

`PERMISSION_READ` (pre-pagination), `CAN_ADD_CHILDREN`, `EDIT`, `PUBLISH` and `EDIT_PERMISSIONS` each go through `filterCollection`, and each call independently re-resolves the user's roles (`loadRolesForUser`) before issuing its own `getPermittedIds` query. A single batch returning the permission bitmask per id would collapse all of them into one round-trip.

The description already prices this honestly and the cap bounds it, so this is not blocking — flagging it as the natural follow-up, since it would also remove the need for the `perPage` cap to exist in the first place.

---

### Verified correct

For the record, I checked the load-bearing claims in the description rather than taking them at face value:

- **`EDIT` vs `WRITE` parity holds.** The table's `DotFolderTransformerImpl.contentDriveView` maps bits via `Type.findById(int)`, which returns the first enum constant whose bit matches; `EDIT` is declared before `WRITE` and both are `2` (`PermissionAPI.java:54-55`), so the table emits `EDIT`. The FE's `PERMISSIONS_TYPE` in `dot-content-drive.model.ts:48-54` spells the same five names. Using `getCanonicalTypes()` would indeed have emitted `WRITE`.
- **`READ` is genuinely implicit.** `page` is a sublist of the collection already filtered by `PERMISSION_READ`, so every folder in it holds READ.
- **Flag-off is a true no-op.** `filterIdsByPermission` returns `Set.of()` without touching the DB, and `permissions` is set to `null`.
- **`permissions: null` really does serialize.** There is no `NON_NULL` inclusion on `ResponseEntityView` or in the default mapper, so the "not requested" vs "requested, no grants" distinction survives the wire as documented.
- **CMS Admin behaviour matches between the two views.** `getPermissionIdsFromRoles` returns the full permission-type set for the admin role, and `filterCollection` short-circuits for it — both paths yield all five names.
- **OpenAPI regeneration is consistent** with the annotations, and `respectFrontendRoles` is correctly left at `false` for this internal path.

dotCMS/src/main/java/com/dotmarketing/portlets/folders/business/FolderAPIImpl.java
```diff
@@ -830,20 +830,33 @@ public PaginatedArrayList<FolderSearchView> searchFolders(final FolderSearchPara
...
 				.stream().map(Permissionable::getPermissionId).collect(Collectors.toSet());
 
-		// 5. Determine which folders have permission-visible children.
+		// 5. Opt-in permission types, each a batch call scoped to the page. READ needs no call (the
+		//    folder is here because it passed the READ filter in step 2) and CAN_ADD_CHILDREN reuses
+		//    step 4, so the full 5-type set costs 3 extra calls.
+		final Set<String> editableIds        = filterIdsByPermission(page, PermissionAPI.PERMISSION_EDIT, params);
+		final Set<String> publishableIds     = filterIdsByPermission(page, PermissionAPI.PERMISSION_PUBLISH, params);
+		final Set<String> editPermissionsIds = filterIdsByPermission(page, PermissionAPI.PERMISSION_EDIT_PERMISSIONS, params);
+
```
fabrizzio-dotCMS: All of these: 
```
final Set<String> editableIds        = filterIdsByPermission(page, PermissionAPI.PERMISSION_EDIT, params);
		final Set<String> publishableIds     = filterIdsByPermission(page, PermissionAPI.PERMISSION_PUBLISH, params);
		final Set<String> editPermissionsIds = filterIdsByPermission(page, PermissionAPI.PERMISSION_EDIT_PERMISSIONS, params);
```

Could be resolved in one single iteration

fabrizzio-dotCMS: Split this gigantic method, searchFolders, into more manageable subroutines

dotCMS/src/main/java/com/dotmarketing/portlets/folders/business/FolderAPIImpl.java
ihoffmann-dot: Done! FolderSearchView now has a builder and the mapping sets every field by name

dotCMS/src/main/java/com/dotcms/util/pagination/FolderSearchPaginator.java
ihoffmann-dot: Done!

dotCMS/src/main/java/com/dotcms/util/pagination/FolderSearchPaginator.java
ihoffmann-dot: I agree. Done!

dotCMS/src/main/java/com/dotmarketing/portlets/folders/business/FolderAPIImpl.java
ihoffmann-dot: Done! searchFolders is now orchestration only

ihoffmann-dot: Agreed, and it's the same thing as finding #4 in your summary. Taking it as [#36940](https://github.com/dotCMS/core/issues/36940) rather than working on it as part of this PR

fabrizzio-dotCMS (approved):

PR #36899, opened 2026-08-05
refactor(search): split SearchHit into a lean content shape and a Site Search one (#36360)

fabrizzio-dotCMS:
## Summary

Follow-up to @wezell's review on #36886: content search results should stay small and fast, and
highlight fragments are a Site Search concern no other caller reads. Instead of every hit carrying
a field only one caller uses, `SearchHit` becomes a **sealed interface over two records**.

**Stacked on #36886** (base is `issue-36360-phase-sweep-fixes`, not `main`) — the field this splits
only exists on that branch. Retarget to `main` once #36886 merges.

## The shape

```java
public sealed interface SearchHit permits ContentSearchHit, SiteSearchHit {
    // ... the six neutral accessors ...
    default List<String> highlightsFor(String field) { return List.of(); }
}
```

- **`ContentSearchHit`** — six components, no highlight state at all. What every content query produces.
- **`SiteSearchHit`** — the same six plus the fragments, overriding `highlightsFor`.

The `default` is what makes it work: the concern is declared once on the contract, costs zero bytes
per instance, and a caller holding a plain `SearchHit` never has to know or ask which shape it got.

## Callers never choose the shape

`Builder.build()` picks it from the data — the lean record unless the engine actually returned
fragments. That matters because Site Search and content search reach the neutral layer through the
*same* factory (`ContentSearchResponse.from`, since Site Search reads via `rawSearch` after #36398),
so a "do I want highlights" flag would have to be threaded through `rawSearch` →
`ContentSearchResponse.from` → `SearchHits.from` → the hit adapter. Deciding from the response keeps
it in one place. Since the only queries that request highlighting are the two Site Search ones, a
content hit is always the lean shape.

## Blast radius: none outside the domain package

`./mvnw compile -pl :dotcms-core` passes with **zero changes anywhere else** — the ten
`SearchHits.from` / `SearchHit::from` call sites, `ContentSearchResponse`, `Aggregation`,
`ESContentResourcePortlet` and `OSSiteSearchAPI` all compile untouched. Verified the enterprise
consumer really does resolve through the contract:

```
javap OSSiteSearchAPI.class
  429: invokeinterface  // SearchHit.highlightsFor:(Ljava/lang/String;)Ljava/util/List;
```

## Jackson

An interface is not instantiable, so deserialization goes through a `@JsonCreator` static factory that
rebuilds via `Builder.build()`. The serialized form already carries the `highlights` key when there were
fragments, so the same rule that picks the shape coming from the engine picks it coming from JSON — no
polymorphic `@JsonTypeInfo`, so the JSON shape the query caches depend on is unchanged.

The first cut of this PR instead pinned deserialization to `ContentSearchHit`, which made a serialized
`SiteSearchHit` come back without its fragments and without an error. Nothing serializes one today, but
that is a silent-loss landmine of the same class as #36026, so it is fixed rather than documented.

Side benefit: `ContentSearchHit` no longer serializes an empty `"highlights":{}`.

## Velocity

No template reads highlights (zero `.vtl` references in the repo), and Site Search results are not
exposed through a viewtool at all — those templates consume `SiteSearchResult`, which this does not
touch. The at-risk surface is the *content* hit accessors reachable as `$hit.id` / `$hit.sourceAsMap` /
`$hit.sortValues`, and those are guarded by `ContentSearchToolTest`, which evaluates real customer VTL
through the dotCMS Velocity engine.

## Testing

All green against this branch. **25 unit + 83 integration tests**, covering every family that consumes
the neutral hit:

| Test | Covers | Result |
|---|---|---|
| `SearchHitTest` | shape decision, OS `sort()` unwrap, highlights, JSON round-trip both ways, field mapped to `null` | 9/9 |
| `AggregationDomainTest` | Jackson round-trip of `SearchHit` / `SearchHits` | 16/16 |
| `ContentSearchToolTest` | real Velocity: `$hit.id` / `.index` / `.sourceAsMap` / `.sortValues` on `search()` and `raw()`, plus the nested `top_hits` walk | 11/11 |
| `ESContentResourcePortletTest` | the `/api/es/search` and `/api/es/raw` wire shape | 14/14 |
| `ContentletIndexAPIImplTest` | content indexing + search, and the Site Search highlight assertions | 16/16 |
| `ESContentFactoryImplTest` | `searchHits` on the Elasticsearch factory path | 42/42 (1 pre-existing skip) |

Verified the ITs ran against the refactor rather than a stale artifact — the integration module resolves
`dotcms-core` from `~/.m2`, so core was reinstalled first and the installed jar carries `SearchHit` as an
interface alongside the two records.

**Coverage gap, stated plainly:** the runs above are phase 0, where Site Search reads are served by
`ESSiteSearchAPI` — which uses Elasticsearch's own `SearchHit` and never touches the neutral type. So
`SiteSearchHit` end-to-end (`OSSiteSearchAPI` → neutral hit → fragments on the result) is currently
covered by unit tests only. Exercising it needs a phase 2/3 run, where OpenSearch serves reads.

## Breaking Changes

None at the source level. `SearchHit` changes from a record to a sealed interface, so anything
compiled against the concrete type would need recompiling — nothing in the repo does, and the type is
internal to the neutral search layer.

This PR fixes: #36360

This PR fixes: #36360

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/content/index/domain/ContentSearchHit.java
+++ b/dotCMS/src/main/java/com/dotcms/content/index/domain/ContentSearchHit.java
@@ -1,0 +1,49 @@
+package com.dotcms.content.index.domain;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * The lean {@link SearchHit} shape: everything a content search result needs and nothing else.
+ *
+ * <p>This is the hit every content query produces. It deliberately carries no highlight state — not
+ * even an empty map reference — because highlighting is a Site Search concern and content queries
+ * never request it. {@link SearchHit#highlightsFor(String)} therefore answers with an empty list here,
+ * inherited straight from the interface, so a caller holding a {@code SearchHit} needs no type check.
+ *
+ * <p>Instances come from {@link SearchHit.Builder#build()}, which selects this shape whenever the
+ * engine response carried no highlight fragments. It is also the deserialization target for the
+ * {@link SearchHit} interface.</p>
+ *
+ * @param getId          the unique identifier of this search hit (the document ID)
+ * @param getIndex       the index name where this search hit was found, or {@code null} if not available
+ * @param getSourceAsMap the source document as a map of field names to values
+ * @param getScore       the search relevance score for this hit
+ * @param getFields      the document fields retrieved by the search query (additional fields beyond
+ *                       the source document that were explicitly requested), empty if none were requested
+ * @param getSortValues  the per-hit sort values the engine returns when the query sorts by a field
+ *                       (e.g. the computed distance for a {@code _geo_distance} sort), in the order the
+ *                       {@code sort} clause declared; empty for relevance-only (unsorted) queries
+ * @author Fabrizio Araya
+ * @see SearchHit
+ * @see SiteSearchHit
+ */
+public record ContentSearchHit(
+        @JsonProperty("id") String getId,
+        @JsonProperty("index") String getIndex,
+        @JsonProperty("sourceAsMap") Map<String, Object> getSourceAsMap,
+        @JsonProperty("score") float getScore,
+        @JsonProperty("fields") Map<String, Object> getFields,
+        @JsonProperty("sortValues") List<Object> getSortValues) implements SearchHit {
+
+    /**
+     * Canonical constructor. Collection components default to an empty map/list when {@code null} so
+     * the accessors never return {@code null} (mirrors the previous Immutables collection defaults).
+     */
+    public ContentSearchHit {
+        getSourceAsMap = getSourceAsMap == null ? Map.of() : getSourceAsMap;
+        getFields = getFields == null ? Map.of() : getFields;
+        getSortValues = getSortValues == null ? List.of() : getSortValues;
+    }
+}
--- a/dotCMS/src/main/java/com/dotcms/content/index/domain/SearchHit.java
+++ b/dotCMS/src/main/java/com/dotcms/content/index/domain/SearchHit.java
@@ -1,6 +1,6 @@
 package com.dotcms.content.index.domain;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import java.util.Arrays;
 import java.util.HashMap;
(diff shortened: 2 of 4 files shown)


fabrizzio-dotCMS:
Closing: I built this to address @wezell's cohesion point on #36886, then measured what it actually bought and concluded it should not land. Full reasoning in https://github.com/dotCMS/core/pull/36886#issuecomment-5196514994.

Short version: the shared contract still carries `highlightsFor(field)`, `Builder.highlights(...)` and the ES `Text[]` flattening adapter, so all this really moved is which of the two records holds the map — a net saving of one reference field per content hit that already pointed at a shared `Map.of()` singleton. Against that it introduces a record→interface binary break and a new silent-data-loss path (`@JsonDeserialize(as = ContentSearchHit.class)` means a serialized `SiteSearchHit` round-trips back without its fragments).

The branch `issue-36360-searchhit-sealed-split` stays up. If we decide the separation is still worth it, the honest version is the one *without* `highlightsFor` on the interface, which trades the saving for a type check at the Site Search call site.

The one piece worth keeping landed on #36886 instead: two `ContentSearchToolTest` assertions matched only a literal template prefix and would have stayed green on a broken accessor.

fabrizzio-dotCMS:
Reopened — @wezell is good with the shape:

```java
public sealed interface SearchHit permits ContentSearchHit, SiteSearchHit {
    // ... the six neutral accessors ...
    default List<String> highlightsFor(String field) { return List.of(); }
}
```

One thing changed since I closed it. My objection had two parts: that the cohesion gain was small, which is a judgement call and yours to make, and that the first cut introduced a *new* silent-data-loss path, which was a real defect rather than a trade-off. `d74911be8e` fixes the second.

Deserialization no longer pins the interface to `ContentSearchHit` (which would have made a serialized `SiteSearchHit` come back without its fragments, silently). It goes through a `@JsonCreator` static factory that rebuilds via `Builder.build()`, so the same rule that picks the shape coming from the engine picks it coming from JSON. No polymorphic type discriminator, so the JSON shape the query caches depend on is unchanged. Covered both ways — 24/24 unit tests green.

Still worth flagging, since it is a trade-off rather than a bug: `SearchHit` goes from a record to an interface, so anything compiled against the concrete type needs recompiling. Nothing in the repo does.

wezell:
On the Record --> Interface  change - we should check our customer plugin who leverages the base ES classes to make sure we don't break it.

Issue #35991, opened 2026-06-05
Java 25 Performance Improvements


wezell:
## Description

Now that dotCMS runs on **Java 25** (runtime `25.0.2-ms`, default bytecode 25 — see #35914), a set of language and JVM capabilities become available that let us improve runtime performance, throughput, and resource efficiency across the platform. This Epic is the **umbrella / tracking issue** for that performance work.

The goal is not a single feature but a coordinated sweep of targeted improvements that take advantage of Java 25, each delivered as its own child PR/issue so they can be reviewed, benchmarked, and rolled back independently.

### Why now

- **JEP 491 (Java 24+): `synchronized` no longer pins virtual threads.** The classic "virtual threads + JDBC/legacy `synchronized` ⇒ carrier pinning" objection no longer applies, so we can safely migrate blocking I/O thread pools to virtual threads.
- Modern GC, scoped values, and other Java 21→25 additions open further throughput/latency wins.
- Much of dotCMS's hottest code is I/O-bound (cache persistence, DB access, HTTP), which is exactly where virtual threads and reduced thread-pool contention pay off.

## Desired Outcome

We will consider this Epic successful when:

- Hot I/O-bound thread pools are migrated to virtual threads where it measurably helps, with no regression in correctness, memory footprint, or shutdown behavior.
- Each improvement is backed by a before/after benchmark or a clear architectural rationale.
- No increase in error rates (e.g. connection-pool timeouts) or cache rebuild storms under load.
- Changes are individually flag-/config-gated or independently revertable.

## Candidate child work

- [ ] **H22 cache async commits → virtual threads** — replace the fixed 5-thread `ThreadPoolExecutor` with a virtual-thread-per-task executor; preserve backpressure and bound concurrent DB writes against the Hikari pool. *(first child PR)*
- [ ] Audit other blocking thread pools / `ExecutorService` usages for virtual-thread suitability.
- [ ] Evaluate GC / JVM flag tuning for the Java 25 runtime.
- [ ] Identify and benchmark additional I/O-bound hotspots (REST, ES, DB access layers).
- [ ] Review opportunities from newer concurrency APIs (e.g. scoped values, structured concurrency) where stable.

> This list is expected to grow; new child issues/PRs should reference this Epic.

## Links

- Java 25 migration: #35914
- CI variant follow-up: #35915

wezell:
[...]

PR #36900, opened 2026-08-05
Revert "perf(cache): migrate H22 cache async commits to virtual threads" (#35992)

### LINKED ISSUE #36892: H22 cache sync fallback blocks startup indefinitely on the MVStore fair lock
## Problem

An instance cannot complete startup. The `main` thread is parked indefinitely inside an H22 (H2 on-disk cache) write, driven from `InitServlet.init()`:

```
"main" ... waiting on condition
   - parking to wait for <0x00000007fe4a3718> (a java.util.concurrent.locks.ReentrantLock$FairSync)
	at org.h2.mvstore.MVStore.commit(MVStore.java:775)
	at org.h2.mvstore.MVStore.beforeWrite(MVStore.java:1273)
	...
	at com.dotmarketing.business.cache.provider.h22.H22Cache.doUpsert(H22Cache.java:650)
	at com.dotmarketing.business.cache.provider.h22.H22Cache.put(H22Cache.java:176)
	...
	at com.dotcms.content.elasticsearch.business.ESContentFactoryImpl.lambda$findContentlets$0(ESContentFactoryImpl.java:1306)
	at com.dotcms.vanityurl.business.VanityUrlAPIImpl.populateAllVanityURLsCache(VanityUrlAPIImpl.java:128)
	at com.dotmarketing.servlets.InitServlet.init(InitServlet.java:113)
```

Startup never finishes. A cache write should never be able to block startup.

## Root cause

`H22Cache.java:176` is the **synchronous** branch of `put()`. `shouldAsync()` returned false, so the caller performs the H2 write itself. Nothing bounds that write:

- `MVStore.commit` takes one **fair** `ReentrantLock` per store (confirmed by `ReentrantLock$FairSync` in the dump). Embedded H2 is single-writer; fair means strict FIFO with no barging.
- `cache.h22.db.poolsize.max` defaults to **500**, so up to 500 threads can hold a connection and queue on that single lock.
- Hikari's `connectionTimeout=1000` bounds *acquiring* a connection, not *executing* one. No `setQueryTimeout` is set on any H22 statement, so the wait is unbounded.

### The overflow valve is what causes the stall

`isAllocationWithinTolerance()` flips at `inFlightTasks >= 0.98 x 10000` = **9800**. Past that threshold, `shouldAsync()` returns false and **every** caller becomes a writer. Concurrency goes from 5 (capped by `dbWorkPermits`) to potentially hundreds, all queued on one fair lock. Throughput collapses, the backlog never drains, and the tolerance never clears. It latches instead of shedding load.

`ESContentFactoryImpl:1305-1308` (`result.forEach(... contentletCache.add ...)`, 200 rows per batch, no pacing) reaches that threshold quickly on a site with many vanity URLs.

### `cache_h22_async_task_queue` no longer describes anything

Since #35992 moved async commits to `Executors.newThreadPerTaskExecutor`, there is no bounded queue. The real concurrency bound is `dbWorkPermits = cache_h22_async_threads` (default 5). `cache_h22_async_task_queue=10000` is now only the threshold at which the provider flips to caller-runs. Meanwhile 9800 parked virtual threads each pin their serialized value, adding heap pressure.

## Fix

1. **In async mode, never fall back to running the write on the caller.** On overflow, drop the write. This is already safe and is the existing error semantic: `put()` marks the key in `DONT_CACHE_ME` (H22Cache.java:167) *before* writing, `doUpsert` only clears it on success (:655), and `doSelect` honors `exclude()` (:664) — so a dropped put reads as a miss. Caller-runs remains only when `cache_h22_async=false` is set explicitly.
2. **Bound overflow with the permit itself** — `dbWorkPermits.tryAcquire(timeout)` and drop on failure. This replaces the phantom queue/tolerance calculation, which can then be deleted.
3. **`setQueryTimeout` on H22 statements** as a backstop so a wedged statement eventually fails instead of parking forever.

### Puts are droppable, deletes are not

`DONT_CACHE_ME` expires after 20s (H22Cache.java:73). A dropped **put** is safe indefinitely — after expiry there is simply no row to find. A dropped **delete** is not: after 20s the stale row becomes visible again and serves stale content. So deletes must still block on the permit (on a virtual thread, never on the caller).

This also means `submitAsync`'s current `RejectedExecutionException` handling is a latent stale-read bug, since it silently drops deletes as well as puts.

## Immediate mitigation (config only, no build)

```
DOT_CACHE_H22_DB_POOLSIZE_MAX=10
```

Caps the fair-lock queue. The 11th caller hits `connectionTimeout` and fails in ~1s into `handleError`, so startup proceeds with degraded caching instead of hanging.

Or drop the H22 tier entirely — it is a disk cache, so this costs read performance, not correctness:

```
DOT_CACHE_CONTENTLETCACHE_CHAIN=com.dotmarketing.business.cache.provider.caffine.CaffineCache
```

## Acceptance criteria

- [ ] With `cache_h22_async=true`, no caller thread ever executes `doUpsert`/`doDelete`; verified by the absence of a synchronous path from `put()`/`remove()`.
- [ ] Saturating the async path sheds puts (counted/logged) rather than running them on the caller, and startup completes.
- [ ] Deletes are never silently dropped while the provider is running.
- [ ] H22 statements carry a query timeout.
- [ ] `cache_h22_async_task_queue` / `cache_h22_async_tolerance` are removed, along with the already-`@Ignore`d `Test_Exhaust_Thread_Pool` that covered them.

wezell:
Reverts #35992.

Original issue: #35991 (Java 25 Performance Improvements)
Startup failure this fixes: #36892

## The problem

A customer instance could not complete startup. `main` parked indefinitely in `MVStore.commit` under `H22Cache.put` → `doUpsert`, driven from `InitServlet.init()` → `populateAllVanityURLsCache` → `ESContentFactoryImpl.findContentlets`:

```
"main" ... waiting on condition
   - parking to wait for <0x00000007fe4a3718> (a java.util.concurrent.locks.ReentrantLock$FairSync)
	at org.h2.mvstore.MVStore.commit(MVStore.java:775)
	at org.h2.mvstore.MVStore.beforeWrite(MVStore.java:1273)
	...
	at com.dotmarketing.business.cache.provider.h22.H22Cache.doUpsert(H22Cache.java:650)
	at com.dotmarketing.business.cache.provider.h22.H22Cache.put(H22Cache.java:176)
```

`H22Cache.java:176` is the *synchronous* branch of `put()` — the caller was doing the H2 write itself.

## Why the migration caused it

#35992 rested on this premise:

> JEP 491 means `synchronized` blocks in Hikari/H2 no longer pin the carrier thread, so the classic "virtual threads + JDBC pinning" concern does not apply.

That is true for `synchronized`, but it is not the pinning that matters here. **Virtual threads do not unmount on file I/O, and embedded H2 is file I/O, not socket I/O.** Every commit holding a `dbWorkPermits` permit pins its carrier for the duration of the MVStore write, so the 5 concurrent writers contend for a carrier pool sized to `availableProcessors()`. The old model used 5 dedicated *platform* threads, which are unaffected by carrier availability.

The submission path also changed cost profile: `newThreadPerTaskExecutor` starts a virtual thread immediately on submit, so a backlogged put became a live thread parked on a semaphore rather than a `LinkedBlockingQueue` node.

## Measurements

3000 upserts of a 4KB payload through 5 concurrent writers, H2 2.2.224, using the same statement shape H22 uses (`MERGE INTO … key(cache_id)`):

| carriers | platform pool (old) | virtual+semaphore (new) | ratio |
|---|---|---|---|
| 2 (small container) | 7042 puts/sec | **20 puts/sec** | 350x |
| 4 | 6977 puts/sec | 889 puts/sec | 7.8x |
| 8 | 6000 puts/sec | 1304 puts/sec | 4.6x |
| 24 (default) | 7059 puts/sec | 1293 puts/sec | 5.5x |

Platform throughput is flat regardless of carrier count. The virtual-thread path is 5x slower even with 24 carriers, and collapses on a 2-CPU container.

## Why that throughput loss becomes a hang

The caller-runs fallback in `isAllocationWithinTolerance()` is **not** a regression from #35992 — it dates to `e3b58b3694` (2020-09-16) with the same `10000` / `0.98` defaults. It was simply unreachable at ~7000 puts/sec, because the backlog never approached 9800.

At 20–1300 puts/sec it is reached routinely during startup. Once crossed, every caller writes inline, and hundreds of threads queue on H2's single-writer fair `ReentrantLock`. Throughput drops further, the backlog never drains, and the threshold never clears — it latches rather than sheds.

## Scope

- Pure revert. `H22Cache.java` is byte-identical to `98f8d66fae^`, verified by diff.
- Nothing landed in the h22 package since the migration, so there was nothing to reconcile.
- No test changes — h22 unit tests unmodified and passing.

## Follow-up (deliberately not in this PR)

Tracked in #36892, kept out to keep the revert clean and reviewable:

- overflow should shed puts rather than run them on the caller — already safe, since `put` marks the key in `DONT_CACHE_ME` before writing and `doSelect` honors `exclude()`, so a dropped put reads as a miss
- deletes must never be shed — a dropped delete resurfaces as stale content once the 20s `DONT_CACHE_ME` entry expires
- `setQueryTimeout` on the fail-safe statements
- retire `cache_h22_async_task_queue` / `cache_h22_async_tolerance`, which no longer describe a real queue

## Note for #35991

Any future virtual-thread migration in this codebase should distinguish socket I/O from file I/O. JEP 491 removed `synchronized` pinning; it did not make file I/O unmount. Workloads that block on the filesystem — embedded H2, local disk, FUSE mounts — still hold their carrier and should stay on platform threads.

## Verification

- `./mvnw test-compile -pl :dotcms-core` — clean
- `H22CacheTest` — OK (4 tests)
- Benchmark reproducible standalone against H2 2.2.224

This PR fixes: #36892

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -104,6 +104,7 @@ When editing ANY code, improve incrementally:
 
 ### Backend Development (Java/Maven)
 - [Java Standards](docs/backend/JAVA_STANDARDS.md) — Coding patterns, immutables, exceptions, utilities
+- [When to Use Virtual Threads](docs/backend/VIRTUAL_THREADS.md) — Socket I/O yes, file I/O no; carrier pinning
 - [REST API Patterns](docs/backend/REST_API_PATTERNS.md) — JAX-RS, Swagger, @Schema rules
 - [Maven Build System](docs/backend/MAVEN_BUILD_SYSTEM.md) — Dependency management
 - [Configuration Patterns](docs/backend/CONFIGURATION_PATTERNS.md) — Config.getProperty() usage
--- a/docs/backend/VIRTUAL_THREADS.md
+++ b/docs/backend/VIRTUAL_THREADS.md
@@ -1,0 +1,94 @@
+# When to Use Virtual Threads
+
+Java 25 gives us virtual threads. They are not a drop-in upgrade for every thread pool — the win depends
+entirely on **what the thread blocks on**.
+
+## The one rule
+
+> Virtual threads unmount on **socket** I/O. They do **not** unmount on **file** I/O.
+
+A virtual thread blocked on a socket releases its carrier thread, and the carrier goes off and runs
+other work. A virtual thread blocked on the filesystem **holds its carrier for the whole call**. The
+carrier pool defaults to `Runtime.getRuntime().availableProcessors()` — often 2–4 in a container — so a
+handful of concurrent file writes can starve the scheduler for every virtual thread in the JVM.
+
+## ✅ Good candidates
+
+- HTTP / REST clients, webhooks, outbound API calls
+- Elasticsearch / OpenSearch queries
+- PostgreSQL and other **networked** JDBC (`jdbc:postgresql:…`)
+- Redis / pub-sub / message queues
+- Anything with a socket at the bottom of the stack, especially high fan-out with idle waits
+
+## ❌ Poor candidates — keep these on platform threads
+
+- **Embedded H2** (`jdbc:h2:/path/…`) — MVStore writes are file I/O
+- Local disk reads/writes, `FileChannel`, `Files.*`, `File.exists()`, fsync
+- Network-backed filesystems (NFS, EFS, FUSE/geesefs) — file I/O *and* slow
+- Asset/binary storage on a mounted volume
+- CPU-bound work — virtual threads give nothing here; use a sized pool
+
+## Worked example: why we reverted the H22 cache migration
+
+PR #35992 moved the H22 (embedded H2 on-disk cache) async commit executor from a fixed pool of 5
+platform threads to `Executors.newThreadPerTaskExecutor(Thread.ofVirtual()…)`. It was reverted in
+#36900 after causing a startup hang (#36892).
+
+The reasoning in that PR looked sound:
+
+> JEP 491 means `synchronized` blocks in Hikari/H2 no longer pin the carrier thread, so the classic
+> "virtual threads + JDBC pinning" concern does not apply.
+
+That statement is **correct about `synchronized`** and still the wrong conclusion. JEP 491 removed one
+pinning mechanism; it did not make file I/O unmount. Embedded H2 writes to a file, so every writer kept
+its carrier.
+
+Measured — 3000 upserts of a 4KB payload, 5 concurrent writers, H2 2.2.224:
+
+| carriers | platform pool | virtual threads | ratio |
+|---|---|---|---|
+| 2 (small container) | 7042 puts/sec | **20 puts/sec** | 350x |
+| 4 | 6977 puts/sec | 889 puts/sec | 7.8x |
+| 8 | 6000 puts/sec | 1304 puts/sec | 4.6x |
+| 24 (default) | 7059 puts/sec | 1293 puts/sec | 5.5x |
+
+Platform throughput is flat regardless of carrier count. Virtual threads were 5x slower even with 24
+carriers available, and collapsed on a 2-CPU container.
+
+The throughput loss then exposed a second-order failure: a pre-existing overflow path that ran cache
+writes on the calling thread, which was unreachable at 7000 puts/sec, started firing during startup and
+parked the `main` thread on H2's single-writer lock. **A latent fallback that was harmless at high
+throughput became a hang at low throughput.**
+
+## Before you migrate a pool
+
+1. **Identify what it blocks on.** Socket, file, or CPU? If file or CPU, stop.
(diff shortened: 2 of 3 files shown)


wezell:
@erickgonzalez added this:

https://github.com/dotCMS/core/pull/36900/changes#diff-5c0fdd8add40b0f3b720de2f1f2c9411c5dba47c23bff370fda1be5fdfe2f566

fabrizzio-dotCMS (approved):

PR #36902, opened 2026-08-05
feat(evergreen-tracks): daily 06:00 ET cron for standard/trailing, gate kept as break-glass

### LINKED ISSUE #36705: Consider switching evergreen tag-promotion from manual kickoff back to a daily cron (once cluster reconciler crons are trusted)
## Idea
The floating evergreen track tags (latest/standard/trailing) are advanced by a **manual-kickoff** action (`.github/workflows/cicd_evergreen-tracks-promote.yml`, `workflow_dispatch`-only). Once we trust the in-cluster reconciler CronJobs (biweekly, now armed per the infra epic), we could move standard/trailing promotion **back to a daily cadence**.

## NOT a blind revert of #36652 — keep its improvements
#36652 ("manual-only standard/trailing promotion, drop daily cron") did two things: (a) dropped the daily `schedule:` cron, and (b) added genuine improvements we want to KEEP:
- **`plan` → `apply` two-stage flow** gated on the `evergreen-tracks-apply` **required-reviewer** environment (human reviews the dry-run before tags move).
- **Regression test** pinning `--tracks latest` so a release cut moves only `latest`, never standard/trailing.
- Refreshed `cli.py`/`cicd_6-release.yml` comments + README.

So this issue is **re-add the daily `schedule:` trigger only**, preserving the plan/apply structure, the isolation test, and the docs — a surgical change, not `git revert #36652`.

## Design nuance to resolve
The `apply` job waits on a **required-reviewer gate**, so a cron-triggered run would still block for human approval — i.e. it wouldn't be truly unattended. Decide the cron path's interaction with that gate:
- Option A: cron runs **plan-only** (keeps manual `apply`) — daily visibility, human still approves moves.
- Option B: cron runs plan+apply **unattended**, bypassing/auto-approving the gate for the scheduled path — true daily promotion. Safe because the infra reconciler only rolls in-window, on-parity, with immutable `@sha256` pins (this is why we can revisit unattended now).

## Why safe now
The infra reconciler pins every roll to an immutable `@sha256` digest and only acts in its biweekly maintenance window (parity gate + eligibility + no-downgrade). Daily source-tag promotion just keeps the floating tags fresh; it doesn't change rollout cadence or safety.

## Gate
Only after the cluster reconciler crons have run armed for a cycle or two and are trusted.

Parent epic: #35693. Re-adds the cron dropped in #36652 (without reverting its other changes).

sfreudenthaler:
## What

`standard`/`trailing` promotion is now **end-to-end unattended**: a daily cron at **06:00 ET** runs `plan` → `apply` with no human in the loop. Per #36705 this is **Option B**, and it's surgical — the plan/apply structure, drift check, isolation test, and docs from #36652 all stay.

**Manual dispatch is retained as the break-glass path**: `plan` → `gate` → `apply`, where `gate` still waits on the `evergreen-tracks-apply` required-reviewer environment. Use it to review a plan before it lands, or to move tags off-cycle.

`latest` is untouched — the release pipeline moves it on every GA cut.

## How the two paths diverge

| | trigger | gate | applies |
|---|---|---|---|
| Daily | `schedule` 10:00 UTC | **skipped** (`if: github.event_name == 'workflow_dispatch'`) | unattended |
| Break-glass | `workflow_dispatch` | waits for a dotDevelopers reviewer | after approval |

`apply` accepts a **skipped** gate but not a failed or rejected one — hence explicit result checks rather than a bare `always()`.

## Why unattended daily promotion is safe

Moving a floating tag **deploys nothing**. Verified against `dotCMS/infrastructure-as-code` at HEAD:

- Every customer manifest pins an immutable `<version>@sha256:<digest>`; **zero** floating-tag references exist anywhere in the repo.
- The in-cluster reconciler is the only thing that rewrites those pins, and it resolves the track tag **only at run time**, inside its biweekly on-parity Wednesday window (`timeZone: America/New_York`). A tag moved on a Tuesday has no effect until that window.
- No continuous watcher: no Argo CD Image Updater, no Keel/Flux, no `imagePullPolicy: Always` on customer pods.
- The reconciler keeps no cross-run state and its NO-DOWNGRADE gate is a pure local compare, so intervening tag moves are irrelevant — it simply jumps to wherever the tag points now.
- Reconciler is armed on all four prod clusters (dotcms/infrastructure-as-code#8086, dotcms/infrastructure-as-code#8112) and has been rolling envs successfully.

Also: **daily promotion ≠ daily tag movement.** The planner is forward-only and age-gated (14d / 28d), so a track moves only on the day a release actually crosses its threshold.

## Bug this would have shipped without

A `schedule` event carries **no inputs**, so `github.event.inputs.*` is empty on the cron path — #36652 had removed the fallbacks when it dropped the old cron. Restored in both jobs; without them the nightly run would have invoked the engine with an empty `--repo` and empty day thresholds.

## Verification

- All **six** `plan`/`gate` result combinations resolve correctly — cron applies; rejected, cancelled, or failed-plan paths do not.
- End-to-end **break-glass dispatch** against `dotcms/dotcms-test` ([run 31037723308](https://github.com/dotCMS/core/actions/runs/31037723308)): gate waited → approved → `plan unchanged; applying` → tag moved. Plan correctly excluded `latest`.
- 66 unit tests green. Both workflow files parse.

## Docs

README cadence section rewritten (daily cron + break-glass + the safety rationale); RUNBOOK intro and quick-reference updated — routine promotion no longer needs an operator, so the taint/hold controls are framed as the intervention levers.

Closes: #36705

### DIFF
--- a/.github/actions/core-cicd/evergreen-tracks/README.md
+++ b/.github/actions/core-cicd/evergreen-tracks/README.md
@@ -35,8 +35,8 @@ The workflow posts to **#dot-releases**, deliberately only when there's somethin
 
 | Outcome | Posts |
 |---|---|
-| `plan` or `apply` failed | 🚨 yes — the tags did **not** move (unchanged, not half-moved) |
-| A track advanced | 🌲 yes — with what moved where |
+| `plan` or `apply` failed | 🚨 yes — tags apply one at a time, so it points at the run log for which actually landed |
+| A tag moved — a track advanced, or a held track was reconciled back to its hold marker | 🌲 yes — with what moved where |
 | Nothing to do (most days) | nothing |
 
 A rejected break-glass approval leaves `apply` *skipped* rather than failed, so declining a
--- a/.github/actions/core-cicd/evergreen-tracks/README.md
+++ b/.github/actions/core-cicd/evergreen-tracks/README.md
@@ -29,6 +29,20 @@ tags plus `<version>_tainted` / `<track>_hold` markers).
   the drift check. Only `apply` takes the shared registry-mutation lock, so a pending
   approval on the manual path never blocks the release from moving `latest`.
 
+### Notifications
+
+The workflow posts to **#dot-releases**, deliberately only when there's something to say:
+
+| Outcome | Posts |
+|---|---|
+| `plan` or `apply` failed | 🚨 yes — the tags did **not** move (unchanged, not half-moved) |
+| A track advanced | 🌲 yes — with what moved where |
+| Nothing to do (most days) | nothing |
+
+A rejected break-glass approval leaves `apply` *skipped* rather than failed, so declining a
+plan never pages the channel. Notification failures are `continue-on-error` — Slack being
+down never fails a promotion.
+
 ### Why unattended daily promotion is safe
 
 Moving a floating tag deploys nothing. Every customer manifest in
--- a/.github/actions/core-cicd/evergreen-tracks/README.md
+++ b/.github/actions/core-cicd/evergreen-tracks/README.md
@@ -8,26 +8,36 @@ tags plus `<version>_tainted` / `<track>_hold` markers).
 
 - **`latest`** moves automatically on every GA release cut (the release pipeline calls
   `promote --tracks latest --apply`).
-- **`standard` / `trailing`** move **only when an operator manually dispatches** the
-  `evergreen-tracks-promote` GitHub Action. There is no cron — a human running the action
-  at a maintenance-window tag update is the cadence gate, so tracks never re-point
-  off-window. Automatic (e.g. daily) promotion stays off until the customer-experience
-  team green-lights it. When run, each track lands on the newest GA older than its age
-  threshold (`--standard-days` 14, `--trailing-days` 28).
+- **`standard` / `trailing`** advance on a **daily cron at 06:00 ET**, unattended, via the
+  `evergreen-tracks-promote` GitHub Action (`10:00 UTC` — GitHub cron has no DST, so it's
+  05:00 ET in winter). Each track lands on the newest GA older than its age threshold
+  (`--standard-days` 14, `--trailing-days` 28). Daily promotion does not mean daily tag
+  movement: the planner is forward-only and age-gated, so a track moves only on the day a
+  release actually crosses its threshold.
 
-  The dispatch runs in three jobs: `plan` prints the intended moves (dry-run), `gate`
-  waits on the `evergreen-tracks-apply` environment's required-reviewer gate, and `apply`
-  runs after approval — nothing moves until a human reviews the plan and approves.
-  (One-time repo setup: Settings > Environments > `evergreen-tracks-apply` > Required
-  reviewers.) The dispatch is scoped to `--tracks standard,trailing` — it never moves
-  `latest` (the release pipeline owns that). `apply` re-derives its plan from live
-  registry state at run time and **fails if it no longer matches the approved plan**
-  (e.g. a hold/taint changed, or a release aged past a threshold during a long approval),
-  so it can never move tags nobody reviewed — just re-dispatch to review the new plan.
-  Because the plan excludes `latest`, an unattended `latest` move by the release pipeline
-  mid-approval doesn't trip the drift check. Only `apply` takes the shared
-  registry-mutation lock, so a pending approval never blocks the release from moving
-  `latest`.
+  The same workflow can be **dispatched manually as the break-glass path** — to review a
+  plan before it lands, or to move tags off-cycle. On that path a `gate` job waits on the
+  `evergreen-tracks-apply` environment's required-reviewer rule before `apply` runs. The
+  scheduled path skips `gate` entirely, which is what makes it unattended. (One-time repo
+  setup: Settings > Environments > `evergreen-tracks-apply` > Required reviewers.)
+
+  Both paths are scoped to `--tracks standard,trailing` — this workflow never moves
+  `latest` (the release pipeline owns that). `apply` re-derives its plan from live registry
+  state and **fails if it no longer matches what `plan` produced** (e.g. a hold/taint
+  changed), so it can never move tags nobody planned; the next morning's run re-plans.
+  Because the plan excludes `latest`, a `latest` move by the release pipeline never trips
+  the drift check. Only `apply` takes the shared registry-mutation lock, so a pending
+  approval on the manual path never blocks the release from moving `latest`.
+
+### Why unattended daily promotion is safe
+
+Moving a floating tag deploys nothing. Every customer manifest in
+`dotCMS/infrastructure-as-code` pins an immutable `<version>@sha256:<digest>`, and the
+in-cluster evergreen-tracks reconciler is the only thing that rewrites those pins — it
+resolves the track tag at run time, inside its biweekly on-parity Wednesday maintenance
+window. A tag moved on a Tuesday has no effect until that window. Nothing in the clusters
+watches the tags continuously (no Argo CD Image Updater / Keel, no floating-tag references,
+no `imagePullPolicy: Always` on customer pods).
 
 ## Operator procedures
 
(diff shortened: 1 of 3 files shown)


sfreudenthaler:
### Added: #dot-releases notifications (commit 4f584fe849)

Wired up using the repo's existing `core-cicd/notification/notify-slack` action and the channel ID already in use for `#dot-releases` (`CE1TBQU00`) — no new pattern invented.

Designed to stay quiet, since a releases channel shouldn't get a daily "nothing to do" ping:

| Outcome | Posts |
|---|---|
| `plan` or `apply` failed | 🚨 yes — notes the tags did **not** move (unchanged, not half-moved) and that envs are unaffected until the next reconciler window |
| A track advanced | 🌲 yes — with what moved where, and a reminder that envs pick it up at their next window, not immediately |
| Nothing to do (most days) | **nothing** |

Two details worth calling out:
- **A rejected break-glass approval never pages the channel.** Declining leaves `apply` *skipped*, not failed, and the failure condition keys on `== 'failure'` — so a deliberate "no" is silent by construction.
- Both notify steps are `continue-on-error`, so Slack being down can never fail a promotion.

Verified the summary extraction against both-moved / one-moved / no-moves / held-track-only plans. YAML parses, 66 tests still green.

sfreudenthaler:
### ⚠️ Slack smoke test found a real gap — one Slack action needed before this lands

Dispatched the full path against `dotcms/dotcms-test` ([run 31056700820](https://github.com/dotCMS/core/actions/runs/31056700820)). Job routing was **exactly right** — `Notify failure` skipped, `Notify moves` fired, with a correctly rendered payload:

> 🌲 *Evergreen tracks advanced* on `dotcms/dotcms-test` — standard -> 26.03.12-01; trailing -> 26.03.12-01. Environments on these tracks pick this up at their next reconciler maintenance window, not immediately.

**But the message never reached the channel.** The Slack API returned:

```
Error: An API error occurred: not_in_channel
```

The bot behind `SLACK_BOT_TOKEN` is **`cicd`** (`U0164AW6YKC`) — it posts fine to `#log-docker`, `#log-sdk-libs`, etc., but it is **not a member of `#dot-releases`**. The release announcements already in that channel come from a *different* app (`dotBot`), which makes the channel look wired up when it isn't for this token.

**Fix — one command in Slack, no code change:**
```
/invite @cicd
```
run in `#dot-releases`. I'll re-run the smoke test afterwards to confirm.

#### This is a pre-existing bug, not one this PR introduces

`cicd_comp_changelog-site-publish-phase.yml` defaults to the same channel (`CE1TBQU00`) with the same token, and its notice fires only on failure/skip — so **changelog-publish failures have almost certainly been silently swallowed** the same way. Same `/invite` fixes both. Worth a look from whoever owns that workflow.

#### Why the job still reported success

Both notify steps are `continue-on-error: true`, which is deliberate — Slack being down must never fail a tag promotion — but it does mean a misconfigured channel fails quietly. That tradeoff is why this smoke test was worth running rather than trusting the green check.

sfreudenthaler:
### ✅ Slack verified after the `@cicd` invite

Re-ran the full path ([run 31057056392](https://github.com/dotCMS/core/actions/runs/31057056392)) — no `not_in_channel`, and the message landed in `#dot-releases` at 19:40 EDT:

> 🌲 *Evergreen tracks advanced* on `dotcms/dotcms-test` — standard -> 26.03.12-01; trailing -> 26.03.12-01. Environments on these tracks pick this up at their next reconciler maintenance window, not immediately. <View run>

Routing confirmed correct on a real run: `Notify failure` **skipped**, `Notify moves` **fired**.

**Everything on this PR is now verified end-to-end** — cron wiring, all six plan/gate outcome combinations, break-glass approval flow, drift check, and Slack delivery. Ready for review.

Two notes for after merge:
- The **first scheduled run** is the one path that can't be exercised pre-merge (scheduled triggers only fire from the default branch). It'll run 10:00 UTC / 06:00 ET the morning after merge; the `gate` job should show as skipped and `apply` should proceed unattended.
- The pre-existing `#dot-releases` gap in `cicd_comp_changelog-site-publish-phase.yml` is now also fixed by the same invite — those failure notices had been silently dropped and should work from here on.

sfreudenthaler:
### Both Medium findings confirmed and fixed (commit 434251da2b)

Checked each against the engine rather than taking them on faith — both were real.

**🟡 "not half-moved" was a false guarantee.** Confirmed: `cmd_promote` applies moves in a per-track loop (`cli.py:81-84`) and `point_tag` runs `imagetools create` with `check=True` (`executor.py:27-30`), so `standard` succeeding then `trailing` failing leaves a genuinely half-moved registry — the alert asserted the opposite of reality, which is the worst thing an incident message can do. Reworded to point responders at the run log for which tracks actually landed. Kept the "environments unaffected until the next reconciler window" line, which holds either way.

**🟡 Held-track reconciles were silently dropped.** Confirmed: the engine logs `standard (held) -> reconcile to standard_hold (…)` (`cli.py:75`) and then really does re-point that tag — but the anchored grep `^(standard|trailing) -> ` skips it. Demonstrated the miss, then fixed with one optional group: `^(standard|trailing) (\(held\) )?-> `. Answering the reviewer's open question: **yes, this should announce** — an unattended cron physically moving a tag with no notice is exactly the silent-change case this notification exists to prevent.

Headline generalized `*Evergreen tracks advanced*` → `*Evergreen track tags moved*`, since a held reconcile is a correction, not an advance.

**Verified across five plan shapes:**

| Plan | `moved` | Posts |
|---|---|---|
| both tracks advanced | `true` | ✅ |
| no track moves needed (most days) | `false` | quiet |
| held reconcile only — **previously silent** | `true` | ✅ now fires |
| held reconcile + advance | `true` | ✅ both listed |
| held but already consistent (no tag touched) | `false` | quiet — correct |

That last row matters: when a held track already matches its marker the engine logs `held at …, skipping promotion` and calls no `point_tag`, so staying quiet is right — the fix doesn't over-trigger.

66 tests green; YAML parses. The DST note the reviewer flagged is already documented inline and in the README.

sfreudenthaler:
semgrep doens't like the mutable tag but it's acceptable here because we want to pull the latest action of that major.  minors and patch updates in rapidly devloping tech are worth the security boost we get from auto-patching on a floating major tag.

sfreudenthaler:
### 🟡 Gate-failure alert: won't fix — the suggested condition would be harmful (commit 131bad7acf)

The finding's own "what to verify" was the right question, so I tested it rather than reasoned about it. **Dispatched a run and rejected the deployment** ([run 31065188863](https://github.com/dotCMS/core/actions/runs/31065188863)):

| job | result |
|---|---|
| `plan` | success |
| `gate` | **failure** ← a *rejection*, not an infra error |
| `apply` | skipped |
| `notify` | success — both Slack steps **skipped**, nothing posted |

So the reviewer's stated assumption is wrong in the decisive way: **GitHub reports a reviewer rejection as `gate.result == 'failure'`** — the identical value a gate infrastructure failure produces. The two states are indistinguishable from `needs.gate.result`.

Adding `|| needs.gate.result == 'failure'` would therefore **page #dot-releases on every deliberate rejection**, converting a correct, intentional silence into recurring noise. Telling them apart would mean calling the deployment-review API (`/actions/runs/{id}/approvals`) and parsing whether a human recorded a rejection — a real chunk of complexity for the residual case.

And that residual case is thin:
- It **cannot occur on the unattended cron** — `gate` is dispatch-only, so it never runs there. The cron is the entire reason this alert exists.
- On the manual path an operator just clicked *Run workflow*, and a gate failure leaves a **red run** (`run_conclusion: failure`) that GitHub already notifies them about.

No logic change. Added a comment recording the verified behavior and the reasoning, so the next reader doesn't "fix" it back into a noise source.

**Bonus:** this also empirically confirms a claim this PR has been making without proof — *a deliberate "no" never pages the channel*. Verified: both notify steps skipped, and no message appeared in #dot-releases.

yolabingo (approved):

PR #36903, opened 2026-08-05
fix(opensearch): surface a shadow store that stopped accepting writes (#36222)

### LINKED ISSUE #36222: [QA-G16] OpenSearch Migration — Limited (Non-Admin) OpenSearch User (TC-053–TC-058)

fabrizzio-dotCMS:
## What

In dual-write phases OpenSearch is a shadow store and its write failures are fire-and-forget. The reindex listener logged **one WARN per failed item** and nothing else, which is right for a single malformed contentlet and wrong for the case QA hit while exercising #36222 (TC-056): the OS user lost a write permission, so *every* document of *every* batch was rejected with the same message.

Two consequences, both visible in the QA log:

1. **The log became unusable.** Batches of 185/188/183/191/163 items produced ~900 identical `security_exception` warnings in one minute. A real site's reindex emits hundreds of thousands, burying every other entry — including the actionable ones we added in #36783.
2. **Nothing reported that the shadow store had stopped receiving writes at all.** The reindex reported success, the UI showed no error, and OpenSearch received zero documents. Advancing the migration phase from there would switch reads to an index frozen at the moment the permission broke.

`#36783` covered the same failure class on **index creation** (`handleOsBootstrapFailure`); this is the write half of it.

## How

- **`BulkProcessorListener`** — aggregate a batch's failures into one line per distinct message with its count (`23 of 185 item(s) in this batch — …`), keep per-item detail at `DEBUG`, and emit a single actionable `ERROR` when the batch was rejected **in full** for a systemic cause: it names the cause and remediation, states that the shadow store is diverging and must not be promoted, and keeps the verbatim rejection for support. Rate-limited by `REINDEX_SHADOW_FAILURE_ESCALATION_MINUTES` (default 10) — the throttle is static because a fresh listener is built per batch.
- **`OSIndexAPIImpl`** — classify a bare failure *message*, not only an exception chain (a rejected bulk item carries text, never a status code), and recognise the security plugin's own wording (`security_exception`, `no permissions for`). The matching core is now shared by both entry points, so `classifyConnectionError(Throwable)` behaviour is unchanged.
- **`IndexConfigHelper`** — neutral seam (`systemicFailureRemediation`) so the reindex listener escalates without importing the OpenSearch adapter's classifier.

### What deliberately does NOT happen

- **The phase is not reset.** Unlike the bootstrap path, a bulk rejection can be scoped to a single index, so auto-halting on it has too wide a blast radius. The operator gets the ERROR; the decision stays theirs.
- **Partial failures stay silent**, and so does a batch that failed entirely with an unclassifiable message (a shared mapping conflict). Escalating those would flag every malformed contentlet as a migration blocker.

## Testing

```bash
./mvnw test -pl :dotcms-core -Dmaven.build.cache.enabled=false \
    -Dtest='BulkProcessorListenerShadowFailureTest,OSIndexAPIImplConnectionClassifyTest'
```

`Tests run: 26, Failures: 0, Errors: 0` — 10 new in `BulkProcessorListenerShadowFailureTest` (aggregation, dominant-cause selection, and every case that must *not* escalate), 3 new in `OSIndexAPIImplConnectionClassifyTest` (bulk-item message classification). Both are pure-function unit tests: no container, cluster or config, so they belong to no integration suite.

Not covered by an automated test: the rate-limit itself (wall-clock dependent) and the log output shape.

## Follow-up, not in this PR

The readiness endpoint (#36360) is the natural place to report "the OpenSearch copy is not receiving writes" as a hard blocker for advancing a phase. This PR only makes that state detectable and named in the log.

This PR fixes: #36222

### DIFF
--- a/dotCMS/src/main/java/com/dotmarketing/common/reindex/BulkProcessorListener.java
+++ b/dotCMS/src/main/java/com/dotmarketing/common/reindex/BulkProcessorListener.java
@@ -12,13 +12,17 @@ import com.dotmarketing.business.CacheLocator;
 import com.dotmarketing.exception.DotDataException;
 import com.dotmarketing.util.Config;
 import com.dotmarketing.util.Logger;
+import com.dotmarketing.util.UtilMethods;
 import com.liferay.util.StringPool;
 import io.vavr.control.Try;
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
 
 /**
  * {@link IndexBulkListener} that handles the business logic before/after reindexing content.
@@ -35,6 +39,22 @@ public class BulkProcessorListener implements IndexBulkListener {
     final Map<String, ReindexEntry> workingRecords;
 
     static final List<String> RESERVED_IDS = List.of(Host.SYSTEM_HOST);
+
+    /**
+     * Config key: minutes between two systemic-rejection escalations
+     * ({@link #systemicFailureEscalation(IndexTag, int, Map)}). Default: 10.
+     */
+    static final String SHADOW_ESCALATION_INTERVAL_KEY =
+            "REINDEX_SHADOW_FAILURE_ESCALATION_MINUTES";
+
+    /** Stand-in used when the vendor reports a failed item without any message. */
+    static final String NO_FAILURE_MESSAGE = "(no failure message reported)";
+
+    /**
+     * Epoch millis of the last systemic-rejection escalation. Static on purpose: a fresh listener is
+     * created for every batch, so per-instance state could not rate-limit anything.
+     */
+    private static final AtomicLong LAST_SYSTEMIC_ESCALATION = new AtomicLong(0L);
 
     private volatile long contentletsIndexed;
     private int lastBatchSize;
@@ -110,11 +130,12 @@ public class BulkProcessorListener implements IndexBulkListener {
     @Override
     public void afterBulk(final long executionId, final List<IndexBulkItemResult> results) {
         if (shadow) {
-            // OS shadow — fire-and-forget; log individual failures for observability only
-            results.stream()
-                    .filter(IndexBulkItemResult::failed)
-                    .forEach(r -> logShadowWriteFailure(this.getClass(),
-                            "[OS] Index failure (fire-and-forget): " + r.failureMessage(), null));
+            // OS shadow — fire-and-forget, but summarised: a systemic rejection (permissions, TLS,
+            // unreachable cluster) repeats verbatim for every document in the batch, so logging one
+            // line per failed item buried the log — a real reindex emits hundreds of thousands of
+            // identical entries, hiding every other line including the actionable one below
+            // (observed on issue #36222, TC-056: ~900 identical WARNs in one minute).
+            logShadowBatchFailures(results);
             return;
         }
         Logger.debug(this.getClass(), "Bulk process completed");
@@ -163,6 +184,113 @@ public class BulkProcessorListener implements IndexBulkListener {
         }
         Logger.error(ReindexThread.class, "Bulk process failed entirely: " + msg, failure);
         workingRecords.values().forEach(idx -> handleFailure(idx, msg));
+    }
+
+    /**
+     * Logs the failed items of a shadow bulk batch as one aggregated entry per distinct failure
+     * message, keeps the per-item detail at {@code DEBUG}, and escalates once when the whole batch
+     * was rejected for a systemic reason.
+     *
+     * @param results every item result of the completed batch, successful ones included
+     */
+    private void logShadowBatchFailures(final List<IndexBulkItemResult> results) {
+        final Map<String, Long> failuresByMessage = summarizeFailures(results);
+        if (failuresByMessage.isEmpty()) {
+            return;
+        }
+        final String tag = "[" + provider.name() + "] ";
+        failuresByMessage.forEach((message, count) -> logShadowWriteFailure(this.getClass(),
+                tag + "Index failure (fire-and-forget): " + count + " of " + results.size()
+                        + " item(s) in this batch — " + message, null));
+
+        if (Logger.isDebugEnabled(this.getClass())) {
+            results.stream()
+                    .filter(IndexBulkItemResult::failed)
+                    .forEach(result -> Logger.debug(this.getClass(), tag + "Failed item id="
+                            + result.id() + ": " + result.failureMessage()));
+        }
+
+        final Optional<String> escalation =
+                systemicFailureEscalation(provider, results.size(), failuresByMessage);
+        if (escalation.isPresent() && shouldEscalateNow()) {
+            Logger.error(this.getClass(), escalation.get());
+        }
+    }
+
+    /**
+     * Counts the failed items of a batch grouped by failure message, preserving first-seen order.
+     * A systemic cause produces a single entry whose count equals the batch size; a per-document
+     * cause produces several entries, or one entry that covers only part of the batch.
+     *
+     * @param results every item result of the completed batch
+     * @return failure message → number of items that failed with it; empty when nothing failed
+     */
+    static Map<String, Long> summarizeFailures(final List<IndexBulkItemResult> results) {
+        final Map<String, Long> failuresByMessage = new LinkedHashMap<>();
+        for (final IndexBulkItemResult result : results) {
+            if (result.failed()) {
+                final String message = result.failureMessage();
+                failuresByMessage.merge(UtilMethods.isSet(message) ? message : NO_FAILURE_MESSAGE,
+                        1L, Long::sum);
+            }
+        }
+        return failuresByMessage;
+    }
+
+    /**
+     * Builds the actionable message for a shadow batch that was rejected <em>in full</em> for a
+     * systemic reason — the state where the shadow store silently stops receiving writes and
+     * diverges from the authoritative one, which must never be promoted by advancing the migration
+     * phase (issue #36222 follow-up: index creation was already covered by
+     * {@code ContentletIndexAPIImpl.handleOsBootstrapFailure}, the write path was not).
+     *
+     * <p>Deliberately silent unless <em>every</em> item failed: a partial failure is a per-document
+     * problem, and unclassifiable messages (mapping conflicts) are not migration blockers either.
+     * The phase is not reset here — a bulk rejection can be scoped to a single index, so the
+     * decision to halt is left to the operator, with this line and the readiness report as the
+     * signal.</p>
+     *
+     * @param provider          the shadow provider that rejected the batch
+     * @param batchSize         total items in the batch, successful ones included
+     * @param failuresByMessage output of {@link #summarizeFailures(List)}
+     * @return the escalation message, or empty when this batch does not warrant one
+     */
+    static Optional<String> systemicFailureEscalation(final IndexTag provider, final int batchSize,
+            final Map<String, Long> failuresByMessage) {
+        final long failed = failuresByMessage.values().stream().mapToLong(Long::longValue).sum();
+        if (batchSize <= 0 || failed < batchSize) {
+            return Optional.empty();
+        }
+        return failuresByMessage.entrySet().stream()
+                .max(Map.Entry.comparingByValue())
+                .flatMap(dominant -> IndexConfigHelper
+                        .systemicFailureRemediation(dominant.getKey())
+                        .map(remediation -> "[" + provider.name() + "] EVERY document in this bulk"
+                                + " batch (" + batchSize + " item(s)) was rejected by "
+                                + provider.name() + " — likelyCause=" + remediation
+                                + ". The shadow store is NOT receiving writes and is diverging from"
+                                + " the authoritative store, so it must not be promoted: advancing"
+                                + " the migration phase would serve reads from a stale index."
+                                + " Fix the cause above, then run a full reindex to resynchronise."
+                                + " Rejection: " + dominant.getKey()));
+    }
+
+    /**
+     * Rate-limits the systemic escalation to one entry per
+     * {@value #SHADOW_ESCALATION_INTERVAL_KEY} minutes. Static because a fresh listener is built
+     * for every batch, so per-instance state would not throttle anything.
+     *
+     * @return {@code true} when this caller won the right to log the escalation now
+     */
+    private static boolean shouldEscalateNow() {
+        final long intervalMs = TimeUnit.MINUTES.toMillis(
+                Config.getIntProperty(SHADOW_ESCALATION_INTERVAL_KEY, 10));
+        final long now = System.currentTimeMillis();
+        final long previous = LAST_SYSTEMIC_ESCALATION.get();
+        if (previous != 0L && now - previous < intervalMs) {
+            return false;
+        }
+        return LAST_SYSTEMIC_ESCALATION.compareAndSet(previous, now);
     }
 
     static String getMatchingReservedIdIfAny(final String id) {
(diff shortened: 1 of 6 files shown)


dotCMS/src/main/java/com/dotmarketing/common/reindex/BulkProcessorListener.java
```diff
@@ -110,11 +123,12 @@ public void beforeBulk(final long executionId, final int actionCount) {
...
-            results.stream()
-                    .filter(IndexBulkItemResult::failed)
-                    .forEach(r -> logShadowWriteFailure(this.getClass(),
-                            "[OS] Index failure (fire-and-forget): " + r.failureMessage(), null));
+            // OS shadow — fire-and-forget, but summarised: a systemic rejection (permissions, TLS,
+            // unreachable cluster) repeats verbatim for every document in the batch, so logging one
+            // line per failed item buried the log — a real reindex emits hundreds of thousands of
+            // identical entries, hiding every other line including the actionable one below
+            // (observed on issue #36222, TC-056: ~900 identical WARNs in one minute).
+            logShadowBatchFailures(results);
```
wezell: should we wrap this in a `Try.of` or `try/catch` and log if there are any exceptions before rethrowing?  I know exceptions can happen in the reactor class which basically take it down without any messages (older issue, probably fixed), just want this to be as robust as possible?

fabrizzio-dotCMS: Done — wrapped in the last commit.

Two notes on the shape:

- **Catching `Throwable`, not `Exception`.** The escalation path initialises `OSIndexAPIImpl` (via `IndexConfigHelper.systemicFailureRemediation`), so the plausible failure is a `LinkageError`. `Try.of` / `catch (Exception)` would miss exactly the case worth containing — and so would `CompositeBulkProcessor.close()`, which catches `Exception`, skipping its shadow-isolation branch.
- **Logging without rethrowing.** `ContentletIndexOperationsOS.flush()` already calls this callback inside its own `try`, so a rethrow comes back as `afterBulk(executionId, throwable)` and gets logged as *"Bulk process failed entirely"* — a defect in the reporting would then look exactly like the shadow store having stopped accepting writes, which is the signal used to decide whether it can be promoted. The recovery log is a plain `Logger.warn` on purpose so it cannot re-enter the classifier that may have just failed.

On the reactor: agreed it is a real failure mode, but it sits on the ES leg — `ContentletIndexOperationsES` hands `afterBulk` to the RHLC `BulkProcessor`, whose callbacks run on the httpasyncclient I/O reactor threads (the `"reactor status: STOPPED"` we still recover from in `ContentletFactory.rebuildRestHighLevelClientIfNeeded`). This shadow path calls `client.bulk(...)` synchronously, so the callback runs on the ReindexThread. The wrap still earns its place for the two reasons above.

Covered by `BulkProcessorListenerShadowFailureTest#aFailureWhileReportingIsContained_andNeverReachesTheCaller`; it fails without the wrap.

wezell (approved):
Nice!  Not 100% sure it is needed but makes me happy. Thanks Fabro

PR #36920, opened 2026-08-06
fix(security): obfuscate system config overrides in JVM info endpoint (#36919)

### LINKED ISSUE #36919: Maintenance portlet: system config overrides displayed without obfuscation
## Problem

The JVM Info screen in the maintenance portlet (`GET /api/v1/jvm`) obfuscates sensitive values in the **System Properties** and **Environment Variables** sections via `JVMInfoResource.obfuscateIfNeeded()` (masks values whose key matches `passw|pass|passwd|secret|key|token`, plus the configurable `OBFUSCATE_SYSTEM_ENVIRONMENTAL_VARIABLES` pattern).

The **configOverrides** section does not: `getDBOverrides()` copies `SystemTable.all()` straight into the response, so any secret stored as a system-table config override (API keys, tokens, passwords set via the UI/API) is rendered in plaintext to anyone with maintenance-portlet access.

## Fix

Apply the same `obfuscateIfNeeded(key, value)` rules to the config-overrides map — one line in `JVMInfoResource.getDBOverrides()`.

## Security note

Read access already requires a backend user with the maintenance portlet, so exposure is limited to authenticated admins — but the whole point of the existing obfuscation is that even admins shouldn't see raw secrets on a screen that gets screenshotted into support tickets.

wezell:
### Proposed Changes

One-line fix in `JVMInfoResource.getDBOverrides()`: run each system-table config override through the same `obfuscateIfNeeded(key, value)` used by the System Properties and Environment Variables sections, instead of copying `SystemTable.all()` into the response unmasked.

Before: `GET /api/v1/jvm` → `configOverrides` returned every override in plaintext, including keys matching `passw|pass|passwd|secret|key|token` that the other two sections would have masked as `a*********z`.

After: identical masking rules across all three sections, including the `OBFUSCATE_SYSTEM_ENVIRONMENTAL_VARIABLES` custom pattern.

No behavior change for non-sensitive keys. Endpoint already requires a backend user with the maintenance portlet; this is defense-in-depth for screens that end up in screenshots and support tickets.

### Checklist
- [x] Security Implications Contemplated (this is a security hardening fix)
- [ ] Translations (n/a)
- [x] Tests (existing endpoint coverage; masking logic itself unchanged and already exercised)

This PR fixes: #36919

This PR fixes: #36919


dotCMS-Machine-User (approved):

Issue #36894, opened 2026-08-05
Workflow Center: per-item execution results, and enrich _bulkfire to support them


rjvelazco:
### Description

Show the outcome of a bulk workflow action per item — succeeded, failed with a reason, skipped — and let the user retry the failures. Doing that needs the SSE endpoint to carry information it currently drops, so this issue covers both the backend change and the view built on it.

#### Part 1 — Enrich `POST /api/v1/workflow/contentlet/actions/_bulkfire`

Progress is already emitted per contentlet. `WorkflowAPIImpl.fireBulkActionTask` calls the callbacks once per item:

```java
successConsumer.accept(1L);                        // no inode — just "+1"
failConsumer.accept(contentlet.getInode(), e);     // inode AND the exception
```

Three things are lost between there and the wire:

1. **The failure reason.** `WorkflowResource` receives the exception and discards it: `eventBuilder.data(Map.class, Map.of("failure", inode))`. The message is already in hand — `ActionFail.newInstance(user, inode, e)` shows how the synchronous path localises it.
2. **Which item succeeded.** The `success` event carries only `{"success": 1}`, so a running total is possible but attributing success to a specific row is not. Fixing this means the `Consumer<Long>` success callback carrying the inode as well.
3. **Skipped items.** `distributeWorkAndProcessNoReturn` — the SSE path — never calls `computeSkippedContentletsCount`, so skipped is structurally absent from the stream. The synchronous path computes it up front and returns `skippedCount` plus a `skipReason`.

#### Part 2 — Results view

With those three in place, the view can show per-row state during the run and a final summary. Until then a client has to choose between live counters with no reasons (`_bulkfire`) or reasons only at the end (`PUT .../bulk/fire`, which returns `fails[]` with `errorMessage`, capped by `workflow.action.bulk.maxexceptions`, default 1000). It cannot have both from one run, because firing twice would run the action twice.

**Evidence this is a real gap, not a preference.** The legacy dialog appears to offer failure details but does not. In `view_bulk_actions_inc.jsp` the SSE path initialises `var entity = {skippedCount: 0, successCount: 0, fails: 0}` — `fails` is a number — and the SSE listeners only increment a local counter and write it to the DOM. `entity.fails` stays `0`, so `toggleFailDetails()` always renders "No failed actions" even when failures occurred, and the skipped row is permanently `0`. The counters are live; the drill-down behind them is dead.

### Acceptance Criteria

**Backend**

[...]

PR #36921, opened 2026-08-06
Content Drive: Add Lock and Unlock quick actions

### LINKED ISSUE #36844: Workflow Center: add bulk Lock and Unlock quick actions
### Description

Add **Lock** and **Unlock** to the Action Center's Quick Actions. Both appear in the design but were left out of the first release because neither has a bulk REST endpoint — see the linked task.

#### Current state

- **Single item already works** in Content Drive: the row context menu locks/unlocks via `PUT /api/v1/content/_lock|_unlock/{inodeOrIdentifier}`, with `GET /api/v1/content/_canlock/{inodeOrIdentifier}` deciding what to offer.
- **`LOCK` / `UNLOCK` are not `SystemAction` values**, so the multi-contentlet endpoint `POST /api/v1/workflow/actions/default/fire/{systemAction}` — which powers every other quick action — cannot fire them.
- **Legacy has bulk unlock only, and not over REST**: `cmd=full_unlock_list` → `EditContentletAction._batchUnlock`, which loops the inodes on a background thread. There is no bulk *lock* anywhere, including legacy.

#### Two routes — decide before building

| Route | Notes |
|---|---|
| **Client-side loop** | N calls to the existing single-item endpoints with a concurrency limit, aggregated into the same result shape the other quick actions report. This is what legacy does; it just moves the loop across the network. |
| **New REST endpoints** | `_bulklock` / `_bulkunlock` taking `{inodes: []}` and returning `BulkActionsResultView`. The `_batchUnlock` logic already exists to port, and permission checks stay server-side where they belong. |

#### Semantics to settle

Locking is **per user**, which makes bulk unlock different from the other quick actions:

- Unlocking a contentlet locked by *another* user requires permission. `_canlock` reports this per item, so a bulk operation needs an explicit rule: skip those items and report them, or refuse the whole batch.
- Eligibility counts come from row state — `!locked` for Lock, `locked` for Unlock — but that says nothing about *who* holds the lock. A count of 3 "unlockable" items may include one the user is not allowed to unlock.

### Acceptance Criteria

- [ ] Decide client-side loop vs new `_bulklock` / `_bulkunlock` endpoints, and record the decision on this issue
- [ ] Lock appears in Quick Actions with the number of selected contentlets that are not currently locked
- [ ] Unlock appears with the number that are currently locked
- [ ] Neither action is listed when it applies to zero selected contentlets, consistent with the other quick actions
- [ ] Locks held by other users are handled by a documented rule and reported back to the user rather than failing silently
- [ ] Results are reported in the same shape as the other quick actions (succeeded / skipped / failed)
- [ ] The grid refreshes and the selection clears after a successful run
- [ ] Folders in the selection are excluded, matching the rest of the dialog
- [ ] If new endpoints are added: permission is enforced server-side, not just in the UI, and integration tests cover the allowed and denied paths
- [ ] Unit tests cover the eligibility counts and the execute path

### Priority

Medium

### Additional Context

Endpoints in play today:

| Purpose | Endpoint | Scope |
|---|---|---|
| Lock one item | `PUT /api/v1/content/_lock/{inodeOrIdentifier}` | single |
| Unlock one item | `PUT /api/v1/content/_unlock/{inodeOrIdentifier}` | single |
| Can the user lock/unlock it | `GET /api/v1/content/_canlock/{inodeOrIdentifier}` | single |
| Legacy bulk unlock | `cmd=full_unlock_list` (Struts, not REST) | batch, loops server-side |

**Add to Bundle** is the other quick action excluded from the first release, for a similar reason — it is only reachable through the legacy `RemotePublishAjaxAction` servlet rather than a REST endpoint. It is not covered by this issue and is currently untracked.

rjvelazco:
Closes #36844

Adds **Lock** and **Unlock** to the Content Drive Action Center's Quick Actions, and — because firing them exposed two rough edges in the surrounding dialog — routes every quick action through the preview screen and moves execution out of the dialog so a run survives the dialog closing.

Built TDD: the first commit is tests only (Red), the rest make them pass.

---

### Route decision (AC #1)

**Lock/Unlock become `SystemAction` values, fired through the existing multi-contentlet endpoint `POST /api/v1/workflow/actions/default/fire/{systemAction}`.** No new `_bulklock` / `_bulkunlock` endpoint, and no shipped workflow action.

That works because `SystemActionApiFireCommandFactory` already provides a direct-API fallback for a system action with no workflow mapping — the same mechanism that lets `PUBLISH` work on a content type with no scheme. Locking is **per user** rather than a workflow transition, so `LOCK`/`UNLOCK` deliberately have no actionlet and nothing to map to; mapping them to a workflow action has no effect, and the enum documents that.

The alternative — a pair of bulk endpoints — would have duplicated a request/response shape the multi-contentlet fire endpoint already has (`{results, summary}`), for no behaviour the fallback does not already give us.

### Proposed Changes

**Backend**

* `SystemAction` gains `LOCK` and `UNLOCK`, documented as the two members with no actionlet, served by the API-call fallback.
* `SystemActionApiFireCommandFactory` gains fire commands for both, calling `ContentletAPI#lock` / `#unlock` directly.
* Permission is enforced server-side, independent of what the UI offered.
* `openapi.yaml` regenerated from the annotations.

**Quick Actions**

* Lock counts selected contentlets that are not locked; Unlock counts those that are. Neither is listed at a count of zero, and folders are excluded — consistent with the rest of the dialog.
* New `warnWhen` / `warningCount` / `warningHint` on a quick action. Unlock uses it to flag locks held by other users, detected from the row's `contentEditable` flag (the server's answer to "is this locked by me?").
* **Every locked row is still fired.** Only a CMS Administrator may release someone else's lock and the client cannot know whether the caller is one, so failures are reported rather than pre-filtered.
* Result counts now come from `summary.successCount` / `summary.failCount` on the response instead of the number of inodes sent. The endpoint answers `200` with per-item failures inside, so this corrects **every** quick action, not just these two.

**Quick actions now use the preview screen**

Previously only workflow actions drilled into the preview; a quick action fired straight from its row. Clicking Publish (12) published twelve items with no chance to drop one. The set was knowable but not *changeable* — which matters most on Unlock, where the row warns that some locks belong to other users and unchecking those rows is the only way to act on the warning.

**Execution moved into the store**

* New `withActionExecution` store feature owns both fire paths, the in-flight state and the result. Holding the subscription in the store makes surviving a dialog close a deliberate property rather than a leaked subscription — and gives a reopened dialog truthful state, which is what stops the same rows being fired twice.
* Firing hands off immediately: the dialog closes, the toolbar reports `Applying <action> to N item(s)…`, and a toast reports the outcome. Partial failure downgrades the toast to a warning.
* The toast moved to the shell, which owns `<p-toast>` and outlives every dialog, so it can report a result whose originating dialog is already gone. It also refreshes the grid and clears the selection.
* The folders-ignored notice became a plain element instead of `p-message`: PrimeNG 21 animates a Message's height from zero over 300ms with no way to opt out (`motionOptions` is declared but never consumed, the duration is not a design token, there is no global config), which read as the notice arriving late and shoving the action list down as it expanded.

### Scope note for reviewers

The last two sections go beyond #36844's acceptance criteria. They are here because Unlock is what surfaced them: it is the first quick action where the user needs to *edit* the set before firing, and the first where a partial failure is the expected outcome rather than an error. Happy to split them out if you would rather review them separately.

### Checklist

- [x] Tests
- [x] Translations — new keys in `Language.properties`: `content-drive.action-center.unlock.locked-by-others`, `content-drive.action-center.applying`
- [x] Security Implications Contemplated

Permission for lock/unlock is enforced server-side in the fire command, not only in the UI, and integration tests cover the allowed and denied paths. Releasing another user's lock stays restricted to CMS Administrators — the client offers the attempt and reports the failure, it does not bypass the check.

### Additional Info

**Testing**

* Integration (`WorkflowResourceLockUnlockIntegrationTest`, 8 tests): single and bulk happy paths; a lock held by another user reported per item with the batch not refused and the lock surviving; mixed ownership yielding a partial result rather than all-or-nothing; server-side permission enforcement.
* Frontend: 1,003 tests passing in `portlets-content-drive`, covering eligibility counts, the warning count, the execute path, delta accumulation and the toast's three outcomes.
* `pnpm nx build dotcms-ui` succeeds.

**One thing worth knowing:** `tsconfig.spec.json` sets `isolatedModules: true`, so Jest transpiles per-file with no cross-file type checking. A store-composition type error passed 1,000 green tests and only failed at `nx build`. Green tests are not a typecheck for anything touching `signalStore` composition.

**Follow-up, not in this PR:** the same progress/result contract would let bulk publish/unpublish/archive report per-item outcomes, but those run on `DotConcurrentFactory` rather than the job queue and have no `jobId` to monitor — see #36894.

### Screenshots

This PR fixes: #36844


ihoffmann-dot (approved):

rjvelazco:
@claude Review this code and look for: Angular anti-patterns, defects, possible bugs, flaky tests, etc. For each issue found, also propose the correct way to fix it.

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/workflow/WorkflowResourceLockUnlockIntegrationTest.java
```diff
@@ -0,0 +1,428 @@
...
+ * system-action endpoints, both for a single contentlet and for a collection.
+ *
+ * <p>Locking is <b>per user</b>, which makes these two system actions behave unlike the rest:
+ * {@code canLock} refuses to release a lock held by somebody else unless the caller holds the CMS
+ * Administrator role. The bulk endpoint therefore reports those items as failures instead of
+ * refusing the whole batch, and that contract is what most of these tests pin down.</p>
+ *
+ * @see SystemActionApiFireCommandFactory
+ */
+public class WorkflowResourceLockUnlockIntegrationTest {
```
zJaaal, line 56: **Medium — these 8 tests will never run.**

This class is not registered in any suite:

```
$ git grep -n LockUnlockIntegrationTest -- dotcms-integration
dotcms-integration/.../WorkflowResourceLockUnlockIntegrationTest.java:55
```

One match, the file itself. Its sibling is wired up — `MainSuite2a.java:71` lists `com.dotcms.rest.api.v1.workflow.WorkflowResourceIntegrationTest.class`.

Integration tests only execute through the `MainSuite*` classes, so as it stands the entire server-side proof of the contract this PR introduces — per-item failure reporting, mixed ownership yielding a partial result, `canLock` permission enforcement — does not run in CI. A regression in `LockSystemActionApiFireCommandImpl` or `UnlockSystemActionApiFireCommandImpl` would land green.

Adding the class next to its sibling in `MainSuite2a` should be all it takes.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/utils/action-center.ts
```diff
@@ -79,7 +90,38 @@ const QUICK_ACTIONS: {
...
+        icon: 'lock_open',
+        danger: false,
+        // `locked` alone would do — archive refuses locked content — but the archived exclusion is
+        // spelled out so the pair reads the same way.
+        eligibleWhen: (item) => !!item.locked && !item.archived,
+        // A lock belonging to someone else can only be released by a CMS Administrator, and the
+        // grid has no idea whether the current user holds that role. So these items are counted,
+        // fired, and reported on rather than filtered out: `contentEditable` is false on a locked
+        // row the current user does not hold.
+        warnWhen: (item) => !item.contentEditable,
```
zJaaal, line 144: **Medium — this warns the one user for whom it is false.**

The hint says *"Only an administrator can unlock those, and they will be reported as failed"*, but `contentEditable` cannot tell you whether the caller is that administrator.

`contentEditable` is derived in `BrowserAPIImpl.WfData` (`BrowserAPIImpl.java:2636`) as essentially `lockedUserId.equals(user.getUserId())` — it never consults the CMS Administrator role. The server-side gate is `ESContentletAPIImpl.canLock` (`:10358`), which returns `true` immediately for a CMS Admin.

So: an admin selects 5 locked rows, 3 held by other users. The row reads "3 … will be reported as failed", they fire it, and all 5 unlock. The warning was pure noise for the only role that can act on it.

Second, narrower false positive: a user who *holds* the lock but has EDIT at the content-type level rather than on the contentlet. `contentEditable` only checks contentlet WRITE, while `canLock` accepts either.

The comment above is honest that "the grid has no idea whether the current user holds that role" — the gap is that the copy then asserts an outcome anyway. Two ways out: soften the hint to something conditional ("may require administrator permission"), or plumb the caller's admin-ness down so `warnWhen` can return `false` outright for admins. The second also fixes the preview marker (see the template comment).

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dialogs/dot-content-drive-action-center/dot-content-drive-action-center.component.html
```diff
@@ -281,7 +303,9 @@ <h3 class="mb-3 text-[10px] font-bold tracking-widest text-surface-400 uppercase
             paginator sits below the scroll region rather than at the bottom of a scrolled page.
         -->
         <div class="flex min-h-0 flex-1 flex-col" data-testid="action-preview-body">
-            @if ($selectedAction(); as action) {
+            <!-- Workflow actions only. A quick action's count and its rows come from the same
+                 client-side filter, so it can never apply to fewer items than it lists. -->
+            @if (!$pendingQuickAction() && $selectedAction(); as action) {
```
zJaaal, line 308: **Medium — the preview cannot show the thing it was added for.**

The PR body's justification for routing quick actions through the preview is that on Unlock *"unchecking those rows is the only way to act on the warning"*. But the preview renders no per-row lock-owner information: this `@if` is gated on `!$pendingQuickAction()`, and `dot-content-drive-action-preview.component.html` has no lock marker at all — every row gets the same `dot-contentlet-status-badge`.

So the user sees "Unlock (10) ⚠3", drills in, and has no way to tell which 3 rows to uncheck. The capability the preview exists to provide is not reachable.

**Suggestion: use Content Drive's own table instead of a second hand-rolled one.**

`DotFolderListViewComponent` (`@dotcms/portlets/content-drive/ui`, exported from `ui/src/index.ts`) already renders the lock state per row:

```html
<span data-testId="lock-hint" class="text-gray-700">
    @if (!isFolder) {
        @if (item.locked) {
            <i class="pi pi-lock" data-testId="lock-icon"></i>
        } @else {
            <i class="pi pi-lock-open" data-testId="lock-open-icon"></i>
        }
    }
</span>
```

…alongside the checkbox column, `dot-content-thumbnail`, `dot-contentlet-status-badge`, type and language. The preview component's own header comment says *"Column markup follows the main grid (`dot-folder-list-view`) so a row reads the same in both places"* — it is deliberately reproducing that component, and the one column it dropped is the column this finding needs. Reusing it gets the marker for free and removes the drift risk between the two tables.

Two honest caveats, both small and both in the `ui` lib:

1. **Selection is not an input yet.** Today it is `[(selection)]="selectedItems"` against an untyped internal field (`selectedItems = []`, `:189`), with an effect at `:308` that clears it whenever `$items()` changes. The preview needs the parent to own the included set, so this wants a `selection` input (and a `disabled` input for the in-flight state) added to `DotFolderListViewComponent`. That effect would also need to not stomp a caller-provided selection.
2. **`item.locked` alone is not the right signal.** It says "locked", not "locked by someone else", so the marker has to key off whatever `warnWhen` settles on in the sibling comment above — otherwise the icon flags all 10 rows instead of the 3 that matter.

If that is more than you want in this PR, the smaller version is to add the marker to the existing preview table and keep the two in sync; but given the preview is already a copy of `dot-folder-list-view`, reusing it looks like the cheaper end state.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/utils/action-center.ts
```diff
@@ -79,7 +90,38 @@ const QUICK_ACTIONS: {
...
+        // Lock and Unlock lead the list: they are the least destructive actions here and the ones a
+        // user reaches for mid-edit, so they sit furthest from Archive and Delete.
+        id: WORKFLOW_ACTION_ID.LOCK,
+        nameKey: 'content-drive.context-menu.lock',
+        icon: 'lock',
+        danger: false,
+        // Archived content is a dead end until unarchived, and `deleteContentlets` honours
+        // `canLock`, so locking an archived item would quietly make it undeletable by anyone but
+        // the lock holder.
+        eligibleWhen: (item) => !item.locked && !item.archived
```
zJaaal, line 117: **Low/Medium — this guard is client-side only, and the state it prevents is now reachable over the API.**

The comment is right about the hazard: `deleteContentlets` honours `canLock`, so a locked archived item becomes undeletable by anyone but the holder. The problem is that `!item.archived` is the *only* thing enforcing it.

`ESContentletAPIImpl.lock` (`:3924`) checks for a blank inode and `canLock`, and never rejects archived content. `WorkflowResource.checkContentletState` only guards `isNew()`. Now that `LOCK` is a fireable `SystemAction`, a direct `POST /api/v1/workflow/actions/default/fire/LOCK` on an archived inode produces exactly the dead end the comment says must be avoided — no UI needed.

Worth mirroring the check in `LockSystemActionApiFireCommandImpl` (or in `checkContentletState`) so the invariant lives where it is actually enforceable. Pre-existing for the context-menu path, but this PR is what makes it addressable without the UI.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/features/action-execution/withActionExecution.ts
```diff
@@ -0,0 +1,176 @@
...
+                                    patchState(store, { actionExecution: undefined });
+                                    httpErrorManagerService.handle(error);
+
+                                    return EMPTY;
+                                })
+                            )
+                            .subscribe((result) =>
+                                onSettled({
+                                    actionName,
+                                    successCount: result?.summary?.successCount ?? inodes.length,
```
zJaaal: **Low — `?? inodes.length` undoes the fix the surrounding comment describes.**

The PR body is explicit that counts must come from `summary` rather than the number of inodes sent, because the endpoint answers `200` with per-item failures inside. This fallback reinstates the old behaviour for any response that parses but arrives without a `summary`.

That is not hypothetical: `printResponseEntityViewResult` (`WorkflowResource.java` ~4200) swallows `IOException` and can abort mid-stream, and `isFail` can throw `NoSuchElementException` out of the loop on an empty result map. Either way the client can get a body with `results` and no `summary`, and then reports "all N succeeded, 0 failed" with a green toast — the most reassuring possible message for the case where nothing is known to have succeeded.

`?? 0` would be consistent with the intent; treating a missing `summary` as an error and routing it through `httpErrorManagerService` would be better still. Note `failCount` below already defaults to `0`, so the two halves currently disagree about what a missing summary means.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.ts
```diff
@@ -449,6 +449,73 @@ export class DotContentDriveShellComponent {
...
+        const result = this.#store.actionExecutionResult();
+
+        if (!result) {
+            return;
+        }
+
+        const { actionName, successCount, skippedCount, failCount } = result;
+
+        const detail =
+            failCount > 0
```
zJaaal, line 481: **Low — a run that both skips and fails never reports the skips.**

The ladder is `failCount > 0` → `skippedCount > 0` → plain, so the two are mutually exclusive in the output even though the backend can return both. A bulk workflow fire over a mixed-type selection where some contentlets' schemes do not own the action (skipped) *and* some are locked (failed) shows only the failure copy.

That is worse than dropping a number: `executed-with-fails` reads as "these N could not be actioned" and attributes the whole shortfall to permissions or locks, when part of it was items that were merely on the wrong step. The user's next move (go unlock things) is then wrong.

Either add a both-present message, or compose the detail from the non-zero parts instead of picking one branch.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.ts
```diff
@@ -449,6 +449,73 @@ export class DotContentDriveShellComponent {
...
+            life: failCount > 0 ? WARNING_MESSAGE_LIFE : SUCCESS_MESSAGE_LIFE
+        });
+
+        untracked(() => {
+            // Contentlets have moved step, so the grid is stale; `loadItems` also drops the selection
+            // the run consumed.
+            this.#store.loadItems();
+            // A no-op when the user already closed the dialog, which is the common path now that
+            // firing hands off to the toolbar.
+            this.#store.closeDialog();
```
zJaaal, line 514: **Low — not a no-op in the other reachable case.**

The comment covers the common path, but there is a second one: firing hands off and closes the dialog, and the dialog is `[closable]="true"`, so the user can reopen the Action Center while a run is still in flight. When the run settles, this line closes it out from under them.

And the reopened dialog is not useful in the meantime — `$executing` disables every quick-action row and the Done button (`dot-content-drive-action-center.component.ts:222`). So the sequence is: reopen, find everything greyed out, then have the window vanish on its own.

The store already knows it is executing, which is the ingredient needed either way: block reopening while a run is in flight, or gate this `closeDialog()` on the dialog being the one that fired. The PR body frames truthful state in a reopened dialog as a feature of moving execution into the store, so this looks like a gap in that story rather than a disagreement with it.

zJaaal (commented):
Review of the full diff (26 files): the backend `SystemAction` additions, the `withActionExecution` store feature, the Action Center preview rework and the toolbar/shell reporting.

The route decision holds up. `SystemActionApiFireCommandFactory`'s direct-API fallback really does cover a system action with no workflow mapping, so adding `LOCK`/`UNLOCK` there rather than shipping `_bulklock` endpoints is the right call, and the enum documenting *why* those two have no actionlet is the sort of thing the next reader needs.

Eight findings, seven of them inline. Two I would fix before merge:

- **The integration tests never run.** `WorkflowResourceLockUnlockIntegrationTest` is not registered in any `MainSuite`, so the whole server-side proof of the contract is inert. Inline on the class declaration.
- **The Unlock warning is wrong for exactly the users it is about.** `contentEditable` does not know about the CMS Administrator role. Inline on `action-center.ts`.

Plus one design gap worth a decision rather than a fix: the preview was added so the user could uncheck locks held by others, but the preview cannot show which rows those are. Suggestion inline on the Action Center template.

---

### Not inline (line falls outside the diff hunks)

**`WorkflowResource.java:3961-3966` — `fireMergeActionDefault` still advertises `allowableValues` without `LOCK`/`UNLOCK`.**

The other three fire endpoints were updated (`3420`, `3653`, `4552`); this one was missed. `PATCH /actions/default/fire/{systemAction}` routes through the same `fireTransactionalAction` -> `mergeContentlet` -> factory fallback, so it *does* accept both at runtime. The result is that the generated `openapi.yaml` documents the same enum inconsistently across the four paths that share it, and a client generated from it cannot call this one with the new values.

---

### Checked and clean

Ruled these out rather than assuming them, in case it saves anyone re-treading:

- **DI after the move into the store**: `DotWorkflowActionsFireService` and `DotHttpErrorManagerService` are both in `apps/dotcms-ui/src/app/providers.ts`, so injecting them at store level resolves. The `dot-folder-list-context-menu` spec was correctly updated for the real store.
- `DotHttpErrorManagerService.handle()` is eager (side effects before `of(result)`), so not subscribing to it inside `catchError` is fine.
- **The shell effect cannot loop**: `loadItems` / `closeDialog` / `clearActionExecutionResult` are all inside `untracked`, and `loadItems` does set `LOADING` and clear `selectedItems` as its comment claims.
- `contentEditable` is genuinely hydrated onto drive rows (`BrowserAPIImpl.hydrate`), so it is not an undefined field that would make every Unlock row warn.
- `Contentlet.isLocked()` reads through `VersionableAPI` rather than the property map, so `unlock`'s guard works on the detached copy `createContentlet` builds.
- `fireDefaultAction`'s return-type change has exactly one caller, and `entity = {results, summary}` matches what `ResponseUtil`'s streaming wrapper emits.
- The `no-enter-motion` CSS override does win: PrimeNG 21 puts `styleClass` on the `p-message` host alongside `animate.enter`, and the theme rule is `animation: p-animate-message-enter`, so `animation: none` at higher specificity beats it.
- `content-drive.context-menu.lock` / `.unlock` already exist in `Language.properties`; the non-ASCII in that file is pre-existing.

---

*Reviewed by Claude (Claude Code), posted from @zJaaal's account.*

zJaaal (approved):

PR #36925, opened 2026-08-06
fix(uve): open file-asset links in a new tab instead of loading them as pages (#35504)

### LINKED ISSUE #35504: PDFs do not preview and instead try to open a page when in Edit mode.
### Problem Statement

When editing a page, if your content contains links to a PDF file, the link while in edit mode will attempt to open the pdf link as if it were a page, instead of opening the PDF file. 

### Steps to Reproduce

https://github.com/user-attachments/assets/86509972-5ab4-4ad2-99ec-c1c5091c59ce

### Acceptance Criteria

I think a click intercept to maybe open a new tab with the PDF file would be the best case scenario, this way it doesn't navigate the user away from the page, but shows that their link is working without having to copy the url and open the page on another tab. 

### dotCMS Version

Evergreen 26.04.11-01

### Severity

### Links

https://dotcms.freshdesk.com/a/tickets/36746

TinyMCE props I used.

{
    "menubar": false,
    "statusbar": true,
    "plugins": "advlist anchor autolink lists link image charmap hr searchreplace wordcount visualchars emoticons paste textcolor colorpicker textpattern validation dotimageclipboard",
    "height": 450,
    "resize": true,
    "image_advtab": true,
    "image_caption": true,
    "browser_spellcheck": true,
    "paste_auto_cleanup_on_paste": true,
    "paste_strip_class_attributes": "mso",
    "convert_urls": true,
    "relative_urls": true,
    "document_base_url": "/",
    "toolbar1": "undo redo | formatselect | bold italic | bullist numlist outdent indent | link unlink | image dotimageclipboard",
    "content_css": "/application/wysiwyg/wysiwyg.css",
    "file_picker_callback": "function(callback, value, meta) { cmsFileBrowser(callback, value, meta); }",
    "urlconverter_callback": "cmsURLConverter",
    "link_class_list": [
        {"title": "None", "value": ""},
        {"title": "External Link Disclaimer", "value": "link-disclaimer"}
    ]
}

dsilvam:
### Proposed Changes
* Add `isAssetPath()` to the UVE utils — a predicate that distinguishes file-asset URLs from HTMLPage URLs. It matches dotCMS asset-delivery prefixes (`/dA/`, `/dotAsset/`, `/contentAsset/`) and otherwise mirrors the backend's own extension heuristic in `Identifier#setURI`: no extension (or the configured `VELOCITY_PAGE_EXTENSION`) means a page, any other real extension means a file.
* Use it in `handleInternalNav` — a same-host href that resolves to a file asset now opens in a new tab and calls `preventDefault()`, instead of being handed to `uveStore.pageLoad()`.
* Tests for both.

### Root cause
`handleInternalNav` split anchor clicks into exactly two buckets: different hostname → open a new tab; anything else → `uveStore.pageLoad({ url: url.pathname, ... })`. There was no check for whether the same-host target was actually an HTMLPage, so a link to a file asset (`/dA/<inode>/fileAsset/doc.pdf`, `/application/files/doc.pdf`) was fed to the Page API, which cannot resolve it. The editor then rendered its "Nothing Live Here Yet" / "Page not found" state.

Because the `(internalNav)` binding is unconditional, this affected **Preview/Published mode as well as Edit mode** — the linked issue is titled edit-mode-only, so please exercise both when testing.

### Checklist
- [x] Tests
- [ ] Translations — n/a, no user-facing strings added
- [x] Security Implications Contemplated

Security note: the new branch passes the already-resolved same-origin `href` to `window.open`. The external-host branch above it is unchanged and still handles cross-origin links, so this does not widen what can be opened; it only changes *how* same-origin file links are handled. No new user input is parsed — `isAssetPath` receives a `URL.pathname` that was already constructed upstream.

### Test coverage
`utils.spec.ts` — 19 cases on `isAssetPath`:
* Assets: `/dA/.../report.pdf`, `/dA/` with no extension, `/dotAsset/`, `/contentAsset/`, `/application/files/report.pdf`, `.docx`, `.mp4`, `.tar.gz`, uppercase `.PDF`
* Pages: `/about-us/index`, `.html`, `.htm`, `.dot`, `/blog/`, `/`
* Regression guards: `/blog/release-v1.2` and `/news/2024.10` must stay pages — a naive extension check would read the trailing `2`/`10` as a file extension and break navigation to URL-map slugs
* Edge: empty and nullish input

`edit-ema-editor.component.spec.ts` — 3 cases on `handleInternalNav`: a `.pdf` link and a `/dA/` link each open a new tab, call `preventDefault`, and do not call `pageLoad`; an `.html` link still routes through `pageLoad`.

Full suite: 37/37 suites, 913 passed, 0 failures. `nx lint portlets-edit-ema-portlet` clean.

### Additional Info
Verified manually against a locally built image: clicking `/dA/<inode>/fileAsset/<name>.pdf` in edit mode now opens the PDF in a new tab and leaves the editor on the page.

Note for reviewers/QA: the Angular bundle ships from the separate `dotcms-core-web` Maven module, so `./mvnw install -pl :dotcms-core -DskipTests` **without `--am`** will silently test a stale frontend. Use `./mvnw install -pl :dotcms-core --am -DskipTests`.

Two related items deliberately left out of scope:
* The external-host branch (`edit-ema-editor.component.ts`, the `url.hostname !== window.location.hostname` case) opens a new tab but never calls `preventDefault()`, so an external link also navigates the iframe away. Same class of bug, one line, but unrelated to this issue.
* An extensionless file asset served from a folder path outside `/dA/` would still be treated as a page. Not reachable through the reported flow; a fully authoritative fix would need a backend round-trip per link click.

Refs: #35504, FD #36746

This PR fixes: #35504

This PR fixes: #35504


core-web/libs/portlets/edit-ema/portlet/src/lib/utils/index.ts
```diff
@@ -1168,3 +1168,57 @@ export const isSamePageNavigation = (incomingUrl: string, currentUrl: string): b
...
+export const isAssetPath = (pathname: string): boolean => {
+    if (!pathname) {
+        return false;
+    }
+
+    if (ASSET_PATH_PREFIXES.some((prefix) => pathname.startsWith(prefix))) {
+        return true;
+    }
+
+    const lastSegment = pathname.slice(pathname.lastIndexOf('/') + 1);
```
zJaaal, line 1242: What happens if the pathName does not have any `/` due to a malformed URL

zJaaal (dismissed):

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/edit-ema-editor.component.ts
```diff
@@ -768,6 +769,16 @@ export class EditEmaEditorComponent implements OnDestroy, AfterViewInit {
             return;
         }
 
+        // Files (PDFs, images, docs…) are not pages: the Page API cannot resolve
+        // them and the editor would show "Page not found". Open them in a new tab
+        // so the author can verify the link without leaving the editor.
+        if (isAssetPath(url.pathname)) {
+            this.window.open(href, '_blank');
```
zJaaal: The branch decision is made on `url.pathname`, which is resolved against `window.location.origin`, but the open uses the unresolved `href`. Those diverge when the click target is a child of the anchor:

```html
<a href="files/report.pdf"><span>Download</span></a>
```

`e.target` is the span, so `target.href` is `undefined` and `href` falls back to `rawHref`, the raw relative attribute. `new URL('files/report.pdf', origin)` correctly yields `/files/report.pdf` so the asset branch is entered, but `window.open('files/report.pdf')` resolves against the admin document (`/dotAdmin/...`) and opens a 404 tab. `url.href` is already computed a few lines above and is exactly what the decision was based on.

Separately, the guard above is `url.hostname !== window.location.hostname`, hostname only, so this branch can still be cross-origin on a different scheme or port. Since this adds a new `window.open`, `noopener` closes reverse-tabnabbing for the cost of one argument. (The external branch has the same gap, which you already called out as out of scope.)

```suggestion
            this.window.open(url.href, '_blank', 'noopener');
```

core-web/libs/portlets/edit-ema/portlet/src/lib/utils/index.ts
```diff
@@ -1168,3 +1168,57 @@ export const isSamePageNavigation = (incomingUrl: string, currentUrl: string): b
...
+
+/** dotCMS path prefixes that stream a binary asset instead of rendering a page. */
+const ASSET_PATH_PREFIXES = ['/dA/', '/dotAsset/', '/contentAsset/'];
+
+/**
+ * Extensions that still resolve to an HTMLPage. `html` is the default
+ * `VELOCITY_PAGE_EXTENSION`; `dot` is the legacy fallback the backend uses
+ * (see `Identifier#setURI`).
+ */
+const PAGE_PATH_EXTENSIONS = new Set(['html', 'htm', 'dot']);
```
zJaaal: `htm` is not a dotCMS page extension. `VELOCITY_PAGE_EXTENSION = html` in `dotmarketing-config.properties:91`, and `dot` is the code-level fallback in `Identifier#setURI` (`Config.getStringProperty("VELOCITY_PAGE_EXTENSION", "dot")`). The doc comment right above the Set documents only those two, so the comment and the code already disagree.

The consequence is not cosmetic: a `.htm` file uploaded as a file asset still gets handed to `pageLoad`, which is the exact bug this PR fixes. There is no page case on the other side of the trade to pay for it. The `['/about-us/index.htm', false]` case in `utils.spec.ts` should flip to `true` with this.

```suggestion
/**
 * Extensions that still resolve to an HTMLPage. `html` is the default
 * `VELOCITY_PAGE_EXTENSION`; `dot` is the legacy fallback the backend uses
 * (see `Identifier#setURI`).
 */
const PAGE_PATH_EXTENSIONS = new Set(['html', 'dot']);
```

Worth a comment noting that `VELOCITY_PAGE_EXTENSION` is configurable, so a site that overrides it will see page links open in a new tab. Not fixable client-side without plumbing the value through, but it should be written down.

zJaaal (commented):
Follow-up review on the heuristic. Three of these are behavior bugs, and two of them reintroduce the reported symptom on narrower inputs; the rest are nits.

What holds up: the three prefixes match `web.xml` exactly (`/dotAsset/*` L540, `/dA/*` L544, `/contentAsset/*` L563), so the case-sensitive `startsWith` is correct rather than an oversight (servlet url-patterns are case-sensitive). Placement after the external-host check and before `isSamePageNavigation` is right, and `preventDefault()` is what actually stops the iframe from navigating away too.

Also closing out my earlier question on `pathname.slice(pathname.lastIndexOf('/') + 1)`: a pathname with no `/` is safe. `lastIndexOf` returns `-1`, so the slice is `slice(0)`, the whole string. `'report.pdf'` classifies as an asset and `'about'` as a page. No fix needed, though it would be a cheap test case.

_Review drafted by Claude (Claude Code) on behalf of @zJaaal._

core-web/libs/portlets/edit-ema/portlet/src/lib/utils/index.ts
```diff
@@ -1168,3 +1168,57 @@ export const isSamePageNavigation = (incomingUrl: string, currentUrl: string): b
...
+
+/**
+ * Matches a plausible file extension: letter-initial, up to 8 alphanumerics.
+ * Guards URL-map slugs such as `/blog/release-v1.2`, whose trailing `2` must
+ * not be mistaken for a file extension.
+ */
+const FILE_EXTENSION_PATTERN = /^[a-z][a-z0-9]{0,7}$/;
```
zJaaal: The letter-initial rule drops real extensions that start with a digit: `7z`, `3gp`, `3ds`. I ran the function standalone and `/files/archive.7z` returns `false`, so a `.7z` download still hits "Page not found".

The thing actually protecting `/blog/release-v1.2` and `/news/2024.10` is that their trailing tokens are *all digits*, not that they start with one. Requiring at least one letter anywhere keeps both guards and fixes the digit-initial case. Verified against the full case list in this PR plus `7z`/`3gp`: `2024.10` and `release-v1.2` stay pages, `7z` becomes an asset, nothing else changes.

```suggestion
/**
 * Matches a plausible file extension: 1-8 alphanumerics containing at least
 * one letter. The letter requirement is what guards URL-map slugs such as
 * `/blog/release-v1.2` and `/news/2024.10`, whose all-digit trailing token
 * must not be mistaken for a file extension.
 */
const FILE_EXTENSION_PATTERN = /^(?=.*[a-z])[a-z0-9]{1,8}$/;
```

Worth adding `['/files/archive.7z', true]` to the asset cases so the rule is pinned.

core-web/libs/portlets/edit-ema/portlet/src/lib/utils/index.ts
```diff
@@ -1168,3 +1168,57 @@ export const isSamePageNavigation = (incomingUrl: string, currentUrl: string): b
...
+ *
+ * @param {string} pathname - The pathname to check (query and hash excluded)
+ * @returns {boolean} True when the pathname points at a file asset
+ *
+ * @example
+ * isAssetPath('/application/files/doc.pdf')  // true
+ * isAssetPath('/dA/abc123/asset/doc.pdf')    // true
+ * isAssetPath('/about-us/index')             // false
+ * isAssetPath('/about-us/index.html')        // false
+ * isAssetPath('/blog/release-v1.2')          // false
```
zJaaal, line 1231: The PR description lists two known limitations but not this one, and it is the most reachable of the three: a page whose last segment contains a dot followed by a short alpha token is classified as an asset. Both `/store/product.detail` and `/pages/about.us` return `true` and would open a new tab instead of navigating, since URL-map slugs are author-controlled and can contain dots.

No code change requested, it is inherent to an extension heuristic without a backend round-trip. But it belongs in this JSDoc next to the `release-v1.2` example so the next reader knows it was a considered trade rather than an oversight.

core-web/libs/portlets/edit-ema/portlet/src/lib/utils/utils.spec.ts
```diff
@@ -1648,4 +1649,41 @@ describe('utils functions', () => {
...
+            ['/dotAsset/abc123', true],
+            ['/contentAsset/raw-data/abc123/asset', true],
+            ['/application/files/report.pdf', true],
+            ['/files/quarterly.docx', true],
+            ['/media/promo.mp4', true],
+            ['/backups/site.tar.gz', true],
+            ['/files/REPORT.PDF', true]
+        ])('should treat %s as a file asset', (pathname, expected) => {
+            expect(isAssetPath(pathname as string)).toBe(expected);
+        });
```
zJaaal, line 1676: Nit: the `expected` column is constant within each block, and carrying it is what forces the `pathname as string` cast (the mixed tuple infers as `(string | boolean)[]`). Dropping it removes the cast and makes the block's intent read off the title.

```suggestion
        it.each([
            '/dA/abc123/asset/report.pdf',
            '/dA/abc123/asset/no-extension',
            '/dotAsset/abc123',
            '/contentAsset/raw-data/abc123/asset',
            '/application/files/report.pdf',
            '/files/quarterly.docx',
            '/media/promo.mp4',
            '/backups/site.tar.gz',
            '/files/REPORT.PDF'
        ])('should treat %s as a file asset', (pathname) => {
            expect(isAssetPath(pathname)).toBe(true);
        });
```

Same shape applies to the page block below (`.toBe(false)`).

core-web/libs/portlets/edit-ema/portlet/src/lib/utils/index.ts
zJaaal: Applied in 6cfcf1eeb5, and the evidence turned out to be stronger than what I wrote above, so recording it here.

I traced how the backend actually resolves page vs file, and it never uses an extension heuristic at all. `CMSUrlUtil#resolveResourceType` (`dotCMS/src/main/java/com/dotmarketing/filters/CMSUrlUtil.java:146-165`) goes `isFileAsset(uri, site, languageId)` → `resolvePageAssetSubtype(...)` → `isFolder(...)` then retry as `uri + CMS_INDEX_PAGE`, each one an identifier lookup against the site. The extension is never consulted. `VELOCITY_PAGE_EXTENSION` appears in three Java files repo-wide and none of them sit in request resolution: `Identifier#setURI` (legacy asset-type classification for links), `ClickstreamFilter`, and an upgrade task.

So the ask changes slightly: not just "drop `htm`" but also drop the claim that this mirrors a backend heuristic, since no such heuristic exists. The JSDoc now says this is a client-side approximation and names `CMSUrlUtil#resolveResourceType` as the authoritative path.

On `htm` specifically: the only occurrence of it as an extension anywhere in the repo is

```properties
##Allowed File assets
ASSETS_SEARCH_AND_REPLACE_ALLOWED_FILE_TYPES=vtl,css,js,txt,html,htm,xml,php
```

`dotmarketing-config.properties:634`, under a comment that reads `##Allowed File assets`. The single piece of evidence in the codebase classifies `.htm` as a file asset, which is the opposite of what the constant asserted.

Two things worth keeping as-is:
- `html` stays, because `VELOCITY_PAGE_EXTENSION = html` is what ships (line 91), even though that same list shows `.html` can also be a file asset. That ambiguity is unavoidable and resolving it in favor of pages is right.
- `CMS_INDEX_PAGE = index` (line 683) is extensionless, so the dominant page URL shape is the no-extension case the heuristic already handles correctly.

_Investigated and applied by Claude (Claude Code) on behalf of @zJaaal._

zJaaal:
@dsilvam I pushed the review fixes straight to this branch as 6cfcf1eeb5 rather than leaving you to transcribe five threads. Shout if you would rather I had left them as comments and I will happily revert.

All four findings applied, red-first: six tests failed before the change (three in `utils.spec.ts`, three in `edit-ema-editor.component.spec.ts`) and pass after.

| Change | Why |
|---|---|
| Dropped `htm` from `PAGE_PATH_EXTENSIONS` | Not a dotCMS page extension. See [the thread](https://github.com/dotCMS/core/pull/36925#discussion_r3750387519) for the backend trace; a `.htm` upload was still being fed to the Page API, which is the bug this PR fixes |
| `FILE_EXTENSION_PATTERN` → `/^(?=.*[a-z])[a-z0-9]{1,8}$/` | Digit-initial extensions (`7z`, `3gp`) are real. The slug guard only needs to reject *all-digit* trailing tokens, so `/news/2024.10` and `/blog/release-v1.2` stay pages |
| `window.open(url.href, '_blank', 'noopener')` | `href` can still be a raw relative attribute when the click lands on a child of the anchor, which opened `/dotAdmin/files/...`. Plus the host check compares hostname only, so this branch can be cross-origin on another scheme or port |
| JSDoc | Corrected the "mirrors the backend's own extension heuristic" claim (there is no such heuristic), and recorded the dotted-page-slug false positive and the `VELOCITY_PAGE_EXTENSION` config caveat |

Tests: `.htm`, `.7z` and `.3gp` cases added; a relative-href case added at the component level; the `it.each` tuples lost their constant `expected` column, which also removed the `as string` casts.

Verified locally: `nx test portlets-edit-ema-portlet` 65/65 suites and 1392 tests passing, `nx lint portlets-edit-ema-portlet` clean, `nx format:write` applied.

Still out of scope and untouched, as you flagged: the external-host branch missing `preventDefault()`, and extensionless file assets served outside `/dA/`.

_Changes authored by Claude (Claude Code) on behalf of @zJaaal._

zJaaal:
Accepting this one, and making the acceptance explicit rather than implicit. Pushed 96b01be680 (JSDoc) and 76fc64ca6d (test).

First, the finding is real, so it is worth stating why rather than waving it through. The page `url` is a plain required `TEXT` field (`HTMLPageAssetAPIImpl.java:126`) and I found no character-class sanitizer on the save path, so a dotted page slug is genuinely authorable.

The reason not to fix it is that the heuristic already fails in the correct direction, and the only real alternative would reverse that:

| Misclassification | Consequence |
|---|---|
| Page read as file (this finding) | Opens in a new tab. The page still renders, the editor keeps its state, the author sees immediately what happened |
| File read as page (the reported bug) | Handed to the Page API, which cannot resolve it, and the editor is stranded on "Page not found" |

The permissive extension test is what buys the first failure mode. Swapping it for a known-extension allowlist would fix `/store/product.detail` and simultaneously send every uncommon file type (`.dwg`, `.sketch`, `.pages`) down the second path, which is the exact defect this PR exists to close. Tightening the 8-character bound to 5 would trade one silent misclassification for another (`.numbers`, `.torrent`) on an arbitrary cutoff. Neither is an improvement.

So, two changes instead of a fix:

- **96b01be680** records the reasoning in the JSDoc, so the next reader sees a deliberate bias rather than an unhandled edge case.
- **76fc64ca6d** pins it as a test:

```ts
it.each(['/store/product.detail', '/pages/about.us', '/docs/getting.started'])(
    'should knowingly misread the page %s as a file asset',
    (pathname) => {
        expect(isAssetPath(pathname)).toBe(true);
    }
);
```

That turns the trade into a tripwire: anyone who "fixes" the false positive has to delete a test whose comment explains what they are giving up, which is the point where the allowlist bias gets reconsidered on purpose.

The authoritative fix remains a backend round-trip per link click against `CMSUrlUtil#resolveResourceType`. Worth its own issue if this ever bites in the field, not worth the per-click latency now.

_Investigated and applied by Claude (Claude Code) on behalf of @zJaaal._

core-web/libs/portlets/edit-ema/portlet/src/lib/utils/index.ts
```diff
@@ -1168,3 +1168,79 @@ export const isSamePageNavigation = (incomingUrl: string, currentUrl: string): b
...
+ * Extensions that still resolve to an HTMLPage. `html` is the shipped
+ * `VELOCITY_PAGE_EXTENSION` and `dot` is the fallback the backend defaults to
+ * when the property is unset (see `Identifier#setURI`). Deliberately excludes
+ * `htm`, which dotCMS treats as an ordinary file asset, not a page.
+ *
+ * `VELOCITY_PAGE_EXTENSION` is configurable, so a site that overrides it to
+ * something else will see its page links open in a new tab. The value is not
+ * exposed to the client, so it cannot be read here.
+ */
+const PAGE_PATH_EXTENSIONS = new Set(['html', 'dot']);
```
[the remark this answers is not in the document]
zJaaal: Correct, and correct for a reason I had already written down and then failed to apply. Fixed in 0bb949348d.

`dotmarketing-config.properties:91` ships `VELOCITY_PAGE_EXTENSION = html`, so the `"dot"` in `Config.getStringProperty("VELOCITY_PAGE_EXTENSION", "dot")` is only the absent-property fallback and is never the active value in a standard install. `.dot` is also the Word 97-2003 template extension, so listing it sent a genuine upload type to the Page API, which is the exact defect this PR exists to close.

This is the same reasoning that removed `htm` two commits earlier, and the same bias recorded on `isAssetPath`: a page misread as a file opens a tab and is recoverable, a file misread as a page strands the editor. I applied that to `htm` and not to `dot`.

`PAGE_PATH_EXTENSIONS` is now `new Set(['html'])`, with the omission of both `htm` and `dot` documented so neither gets added back. The residual case, a site that overrides `VELOCITY_PAGE_EXTENSION`, is the mild failure and is noted in place.

_Reviewed and applied by Claude (Claude Code) on behalf of @zJaaal._

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/edit-ema-editor.component.ts
```diff
@@ -800,6 +815,39 @@ export class EditEmaEditorComponent implements OnDestroy, AfterViewInit {
...
+            const doc = this.window.document;
+            const link = doc.createElement('a');
+
+            link.href = href;
+            link.target = '_blank';
+            link.rel = 'noopener';
+
+            doc.body.appendChild(link);
+            link.click();
+            link.remove();
```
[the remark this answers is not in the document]
zJaaal: Valid, fixed in 0bb949348d. `click()` is precisely the call that throws when the open is refused, which is why the catch is there at all, so the old ordering leaked an anchor into the admin document on every failed attempt.

Restructured with the declaration hoisted so cleanup is unconditional:

```ts
let link: HTMLAnchorElement | null = null;

try {
    // create, configure, append, click
} catch {
    // …
} finally {
    link?.remove();
}
```

The existing "keeps the iframe on the page when opening a new tab throws" test now also asserts `remove()` was called, so the leak cannot come back silently.

_Reviewed and applied by Claude (Claude Code) on behalf of @zJaaal._

core-web/libs/portlets/edit-ema/portlet/src/lib/utils/utils.spec.ts
```diff
@@ -1648,4 +1649,59 @@ describe('utils functions', () => {
...
+            '/backups/archive.7z',
+            '/media/clip.3gp'
+        ])('should treat %s as a file asset', (pathname) => {
+            expect(isAssetPath(pathname)).toBe(true);
+        });
+
+        it.each([
+            '/about-us/index',
+            '/about-us/index.html',
+            '/legacy/index.dot',
```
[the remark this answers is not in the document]
zJaaal: Agreed, moved in 0bb949348d. The case is now `/templates/letterhead.dot` in the asset table, named after the Word template rather than a page, since that is what a `.dot` upload actually is. Reasoning in the [thread on `PAGE_PATH_EXTENSIONS`](https://github.com/dotCMS/core/pull/36925#discussion_r3751210847).

_Reviewed and applied by Claude (Claude Code) on behalf of @zJaaal._

zJaaal:
Copilot review addressed in 0bb949348d. Both inline findings applied; recording why the suppressed one was not.

**Applied** — `dot` removed from `PAGE_PATH_EXTENSIONS`. It is only the absent-property fallback in `Config.getStringProperty("VELOCITY_PAGE_EXTENSION", "dot")`, and `dotmarketing-config.properties:91` ships the value as `html`, so `dot` is never active in a standard install. It is also the Word 97-2003 template extension, so listing it sent a real upload type to the Page API. Test case moved to the asset table.

**Applied** — `#openInNewTab` now removes the temporary anchor in a `finally`. `click()` is the call that throws when the open is refused, so the previous ordering stranded an anchor in the admin document on every failure. The existing throw test now asserts cleanup.

**Declined** — dropping the 8-character cap from `FILE_EXTENSION_PATTERN` so `.webmanifest` classifies as an asset:

* A `.webmanifest` is referenced through `<link rel="manifest">`, not an anchor an author clicks, so it does not reach `handleInternalNav`.
* Removing the cap widens the dotted-page-slug false positive that 76fc64ca6d deliberately pinned with tests. `/about.information` would flip from page to asset.
* The cap is arbitrary in both directions, so the tested and documented behaviour wins absent a real trigger.

Happy to revisit if someone produces a case where an author actually links to a long-extension asset.

Verified: `nx test portlets-edit-ema-portlet` 64/64 suites and 1397 tests passing, `nx lint` clean.

_Reviewed and applied by Claude (Claude Code) on behalf of @zJaaal._

adrianjm-dotCMS (approved):

PR #36935, opened 2026-08-06
fix(db): open the pub/sub LISTEN connection outside the Hikari pool (#36934)

### LINKED ISSUE #36934: Pub/sub PGListener holds a pooled Hikari connection for the JVM lifetime, triggering a false leak warning on every boot
## Problem Statement

Every dotCMS boot logs a `WARN` with a full stack trace that reads like a bug and is not one:

```
WARN pool.ProxyLeakTask: Connection leak detection triggered for org.postgresql.jdbc.PgConnection@… on thread main
java.lang.Exception: Apparent connection leak detected
    at com.zaxxer.hikari.HikariDataSource.getConnection(HikariDataSource.java:100)
    at com.dotcms.dotpubsub.JDBCPubSubImpl$PGListener.<init>(JDBCPubSubImpl.java:128)
    at com.dotcms.dotpubsub.JDBCPubSubImpl.listener(JDBCPubSubImpl.java:63)
    at com.dotcms.dotpubsub.JDBCPubSubImpl.start(JDBCPubSubImpl.java:88)
    at com.dotcms.dotpubsub.QueuingPubSubWrapper.start(QueuingPubSubWrapper.java:74)
    at org.apache.felix.framework.OSGIUtil.<init>(OSGIUtil.java:155)
    …
    at com.dotmarketing.startup.runalways.Task00004LoadStarter.executeUpgrade(Task00004LoadStarter.java:16)
```

`PGListener` takes a connection from the Hikari pool and holds it for the lifetime of the listener thread — which is correct for what it does, since a Postgres `LISTEN` needs a persistent connection:

```java
// JDBCPubSubImpl.java:121
private final Lazy<Connection> connection =
        Lazy.of(() -> Try.of(() -> DbConnectionFactory.getDataSource().getConnection())
                .getOrElseThrow(DotRuntimeException::new));
```

Hikari's leak detector has no way to know that, so it flags any connection held past `DB_LEAK_DETECTION_THRESHOLD` (default 300s / 60s depending on the datasource strategy). **The warning is a false positive** — the connection is released in `stop()` → `stopListening()`.

Two real consequences behind the noise:

1. **A pooled connection is permanently withdrawn** from the pool for the JVM's lifetime. On installations with a tight `max_connections` that is a slot nobody accounts for, and it does not show up as in-use work.
2. **Boot logs carry a scary stack trace that is not actionable**, which trains operators and support to ignore leak warnings — including real ones.

A `LISTEN` connection should not come from the request pool at all.

## Steps to Reproduce

1. Start any dotCMS instance backed by PostgreSQL with the default leak-detection threshold (no special configuration required — the current `dotcms-test:1.0.0-SNAPSHOT` image reproduces it).
2. Let the first boot run through the starter load (`Task00004LoadStarter`).
3. Search the log for `Apparent connection leak detected`.

**Observed:** the warning above is logged with a stack trace rooted at `JDBCPubSubImpl$PGListener.<init>`, and one pool connection stays checked out for as long as the JVM runs.

**Expected:** no leak warning, and the pool keeps all of its connections available for request work.

## Acceptance Criteria

- [ ] The pub/sub `LISTEN` connection no longer comes from the Hikari request pool — it uses a dedicated connection built from the same JDBC coordinates, so the pool keeps its full capacity.
- [ ] When those coordinates are not available (JNDI or any datasource strategy that does not expose a JDBC URL), the listener still works, falling back to the current behaviour rather than failing to start.
- [ ] A normal boot logs no `Apparent connection leak detected` warning attributable to `JDBCPubSubImpl`.
- [ ] The dedicated connection is closed on `stop()` / `stopListening()` and on listener restart, so repeated start/stop cycles do not accumulate connections.
- [ ] Pub/sub behaviour is unchanged: topics are still received after the change, including after a listener reconnect.

## dotCMS Version

Reproduced on `dotcms/dotcms-test:1.0.0-SNAPSHOT` built from `main` at 2026-08-06. The code path is long-standing and not specific to that build.

## Notes

Found incidentally while reproducing #36222 (TC-056) on a local ES→OS migration stack. Unrelated to that issue's subsystem — filed separately rather than bundled into its fix.

fabrizzio-dotCMS:
Fixes #36934

## What

`PGListener` borrowed a connection from the Hikari request pool and held it for the lifetime of the JVM:

```java
// JDBCPubSubImpl.java:121 — before
private final Lazy<Connection> connection =
        Lazy.of(() -> Try.of(() -> DbConnectionFactory.getDataSource().getConnection())…);
```

Holding it is correct — a Postgres `LISTEN` needs a persistent connection — but taking it from the request pool is not. Two consequences:

1. **The pool silently loses a slot.** It is withdrawn for the lifetime of the JVM and never shows up as in-use work, so nobody accounts for it against `max_connections`.
2. **Every boot logs `Apparent connection leak detected`** with a stack trace rooted at `PGListener.<init>`. It is a false positive — `stopListening()` does release the connection — but it reads like a bug, and it trains operators and support to ignore leak warnings, including real ones.

## How

The listener builds its own connection from the pool's JDBC coordinates through `DriverManager`, so the pool keeps its full capacity and the detector has nothing to report.

When the datasource exposes no JDBC URL — JNDI-provided or otherwise wrapped — it falls back to the previous behaviour and says so in the log. A listener that works while logging a spurious warning beats no listener at all.

Lifecycle is unchanged: `stopListening()` already closes the connection, so start/stop cycles do not accumulate connections.

## Testing

```bash
./mvnw test -pl :dotcms-core -Dmaven.build.cache.enabled=false \
    -Dtest=JDBCPubSubListenerConnectionTest
```

`Tests run: 2, Failures: 0, Errors: 0` — pure unit tests, in no suite. They pin both branches: the fallback returns the pooled connection, and a datasource that does expose coordinates is **never** asked for one (`verify(hikari, never()).getConnection()`), which is the property that actually matters.

Verified at runtime against PostgreSQL 18 + dotCMS in Docker:

```
INFO dotpubsub.JDBCPubSubImpl - Opening a dedicated connection for the Postgres pub/sub
                                listener, outside the Hikari pool.
INFO dotpubsub.JDBCPubSubImpl - Starting to listen for Postgres notifications.
```

A controlled first-boot run on an empty database — the path where the warning was originally observed, during the starter load — is in progress; I will post the result as a comment. What is already confirmed is the positive signal above: the dedicated path is taken and the listener starts and runs normally.

## Context

Found incidentally while reproducing #36222 (TC-056) on a local ES→OS migration stack. Unrelated subsystem, so it is filed and fixed separately rather than bundled into that fix.

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/dotpubsub/JDBCPubSubImpl.java
+++ b/dotCMS/src/main/java/com/dotcms/dotpubsub/JDBCPubSubImpl.java
@@ -9,14 +9,18 @@ import com.dotmarketing.util.Logger;
 import com.dotmarketing.util.StringUtils;
 import com.dotmarketing.util.UtilMethods;
 import com.google.common.annotations.VisibleForTesting;
+import com.zaxxer.hikari.HikariDataSource;
 import io.vavr.Lazy;
 import io.vavr.control.Try;
 import org.postgresql.PGConnection;
 import org.postgresql.PGNotification;
 
+import javax.sql.DataSource;
 import javax.validation.constraints.NotNull;
 import java.sql.Connection;
+import java.sql.DriverManager;
 import java.sql.PreparedStatement;
+import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.Map;
 import java.util.Set;
@@ -26,7 +30,8 @@ import java.util.regex.Pattern;
 
 /**
  * Provides notifications for the postgres pub/sub connection.
- * The secret sauce is that it borrows 1 DB connection and keeps it open forever.
+ * The secret sauce is that it opens 1 dedicated DB connection — outside the Hikari request pool,
+ * see {@code PGListener.listenerConnection()} — and keeps it open forever.
  * With this one long term connection, you "listen" to the topics on
  * postgres and continually do a "SELECT 1" in that connection to be notified
  * of any new messages to those topics.
@@ -118,7 +123,7 @@ public class JDBCPubSubImpl implements DotPubSubProvider {
 
         private RUNSTATE runstate = RUNSTATE.STARTED;
         private final Set<String> topics = ConcurrentHashMap.newKeySet();
-        private final Lazy<Connection> connection = Lazy.of(() -> Try.of(() -> DbConnectionFactory.getDataSource().getConnection()).getOrElseThrow(DotRuntimeException::new));
+        private final Lazy<Connection> connection = Lazy.of(() -> Try.of(PGListener::listenerConnection).getOrElseThrow(DotRuntimeException::new));
         private final Lazy<PGConnection> pgConnection = Lazy.of(() -> Try.of(() -> connection.get().unwrap(PGConnection.class)).getOrElseThrow(DotRuntimeException::new));
         private final Pattern validTopicRegEx = Pattern.compile("[a-z0-9_]");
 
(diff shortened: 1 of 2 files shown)


fabrizzio-dotCMS:
## Controlled first-boot verification — clean

Re-ran the exact path where the warning was originally observed: **empty database**, so `Task00004LoadStarter` runs and the starter load drives `OSGIUtil` → `QueuingPubSubWrapper.start` → `PGListener.<init>`. The patched classes were placed in the container **before** it was started, so the very first boot ran with the fix.

```
23:35:40  INFO  startup.StartupTasksExecutor  - Running Startup Tasks : Task00004LoadStarter
23:35:43  INFO  dotpubsub.JDBCPubSubImpl      - Opening a dedicated connection for the Postgres
                                                pub/sub listener, outside the Hikari pool.
23:35:53  INFO  startup.Catalina              - Server startup in [29611] milliseconds
```

After **8 minutes** of uptime — well past the 300 s default `DB_LEAK_DETECTION_THRESHOLD` that produced the original warning at roughly the 5-minute mark:

| Signal | Count |
|---|---|
| Starter load actually ran (so the case was exercised) | 1 |
| `Apparent connection leak detected` | **0** |
| Dedicated connection taken | 1 |

Before the fix, the same stack logged the warning once per starter-load boot, with the stack trace rooted at `JDBCPubSubImpl$PGListener.<init>`.

Environment: PostgreSQL 18 (`pgvector/pgvector:pg18`), dotCMS built from this branch, default leak-detection threshold — no tuning to make the result look better.

wezell:
Why are we doing this? The Hikari warning is noise. It’s not a real connection leak; it’s saying a possible connection leak.

wezell (commented):
Looks functionally correct.  What happens if the username and password are part of the `hikari.getJdbcUrl()` and not set?

Wondering if the attempt to return a connection should be wrapped in a try that falls back to the existing pool / `dataSource.getConnection()`

dotCMS/src/main/java/com/dotcms/dotpubsub/JDBCPubSubImpl.java
```diff
@@ -128,6 +133,52 @@ class PGListener extends Thread {
...
+        /**
+         * {@link #listenerConnection()} against an explicit datasource, so the choice between a
+         * dedicated and a pooled connection can be exercised without a running pool.
+         *
+         * @param dataSource the datasource to derive the connection from
+         * @return a dedicated connection when possible, a pooled one otherwise
+         * @throws SQLException if the connection cannot be opened
+         */
+        @VisibleForTesting
+        static Connection listenerConnection(final DataSource dataSource) throws SQLException {
```
ihoffmann-dot, line 166: Warn: This method opens the connection via `DriverManager.getConnection()`, which does not follows HikariCP's configured `connectionTimeout` the way pooled connection acquisition does, so an unreachable database at startup time makes the calling thread block for the default connect timeout instead of failing fast

ihoffmann-dot: Also, is it possible that this approach could drop any additional settings configured on the Hikari datasource outside the URL itself? (e.g. from db.properties)

PR #36944, opened 2026-08-07
Issue 36702 asset picker

### LINKED ISSUE #36702: Create AssetPicker component reusing Content Drive (File/Image field asset selection)
### Description

Create a new **AssetPicker** component for the new Edit Contentlet, **reusing as much as possible from Content Drive** (it is essentially a small Content Drive embedded in a picker). It is opened from **File** and **Image** fields to select an asset.

🎨 **[Design](https://aistudio.google.com/apps/d9859d8e-05a6-4745-a485-da7b57fce049?fullscreenApplet=true&showPreview=true&showAssistant=true) (from product):** 

**Key behaviors:**

1. **Upload button** behaves the same as in Content Drive.
2. **File field** entry point: **no content type** pre-selected; only the **locale** is pre-selected, using the **Contentlet's language**.
3. **Image field** entry point: pre-select the **locale**, plus base types **dotAsset** and **File Asset**; additionally apply a **mimetype filter** (e.g. `image/*`). The mimetype filter has **no chip** — it is transparent to the user.
4. **Global last-used path (localStorage):** store the **asset path** of the selected file so the picker can reopen at that location for future selections. This is **not per field** — a **single global asset path** shared across the whole system.
5. **Out of scope:** the per-field **User Searchable** and **Show in Listing** options — this is a small Content Drive, so once those work in Content Drive they auto-propagate here.
6. **Endpoint:** use **`api/v1/drive/search`** — the same endpoint used by Content Drive.
7. **Right section:** **no folders** are listed there.
8. **Content type selector (File/Image field):** only allow **base types dotAsset and File**, and content types **whose base type is one of those**. Content types with base type **Widget, Content**, etc. must **not** be selectable.

**Mimetype filter — example request payload:**

```json
{
    "assetPath": "//dotcms.com/QA/",
    "includeSystemHost": true,
    "mimeTypes": ["image/*"],
    "filters": {
        "text": "",
        "filterFolders": true
    },
    "contentCursor": 0,
    "folderCursor": 0,
    "maxResults": 20,
    "sortBy": "modDate:desc",
    "archived": false,
    "showFolders": false
}
```

### Acceptance Criteria

**Reuse & endpoint**
- [ ] The AssetPicker reuses Content Drive components/logic as much as possible (minimal duplication)
- [x] Asset browsing/searching uses the **`api/v1/drive/search`** endpoint (same as Content Drive)

**Upload**
- [x] The **upload** button behaves the same as it does in Content Drive

**File field entry point**
- [x] Opening from a **File** field pre-selects **no content type**
- [x] Opening from a File field pre-selects the **locale** using the Contentlet's language

**Image field entry point**
- [x] Opening from an **Image** field pre-selects the **locale** (Contentlet language)
- [x] Opening from an Image field pre-selects base types **dotAsset** and **File Asset**
- [x] A **mimetype** filter (e.g. `image/*`) is applied for the Image field, sent via the request payload (`mimeTypes`)
- [x] The mimetype filter is **transparent to the user** — no chip is shown for it

**Content type selector (File/Image field)**
- [x] The content type selector only offers base types **dotAsset** and **File**
- [x] It offers content types whose **base type** is dotAsset or File
- [x] Content types with base type **Widget, Content**, or any other are **not** selectable

**Global last-used path (localStorage)**
- [x] When an asset is selected, its **asset path** is stored in `localStorage`
- [x] The stored path is **global** (one value for the whole system), not per field
- [x] When the picker opens, it uses the stored global path as the starting location (when available)
- [x] Changing the asset (same field or a different field) updates the single global stored path

**Right section**
- [x] The right section lists **no folders** (assets only)

**Out of scope (explicit)**
- [x] Per-field **User Searchable** / **Show in Listing** options are NOT implemented here (they propagate from Content Drive)

### Priority

Medium

### Additional Context

Part of the new Edit Contentlet work (`dotCMS: New Edit Contentlet`); heavily reuses **Content Drive** (`dotCMS: Content Drive`). The picker is a compact Content Drive using the shared `api/v1/drive/search` endpoint. The mimetype filtering for the Image field is applied silently (no user-facing chip). Related field work: #36151 (site/folder field), #36363 (image editor in Image/File fields).

### LINKED ISSUE #36831: AssetPicker 1/7: Migrate DotFolderListView to @dotcms/ui
### Description

Migrate `DotFolderListView` (and related column models/constants) from `@dotcms/portlets/content-drive/ui` into `@dotcms/ui` so AssetPicker and Content Drive can share the same presentational list.

**Related epic:** #36702  
**Series:** AssetPicker 1/7  
**Prerequisite (tree):** #36733 (shared folder tree — separate issue; not duplicated here)

**Current location:** `core-web/libs/portlets/dot-content-drive/ui/src/lib/dot-folder-list-view/`  
**Target:** `core-web/libs/ui` — export from `@dotcms/ui`

Today the list hardcodes `selectionMode="multiple"`. AssetPicker needs single-select; Content Drive must keep multi-select as default.

### Acceptance Criteria

- [ ] `DotFolderListView` (plus column models/constants it needs) lives under `core-web/libs/ui` and is exported from `@dotcms/ui`
- [ ] Component accepts a `selectionMode` input (`single` | `multiple`) with default `multiple` so Content Drive behavior is unchanged
- [ ] Content Drive portlet/shell imports the list from `@dotcms/ui` (thin re-export from `content-drive/ui` is acceptable for compatibility)
- [ ] Existing Content Drive list unit tests are updated/passing
- [ ] Shared component has Spectator/Jest coverage for selection mode behavior
- [ ] No functional regression in Content Drive list view (pagination, sort, selection, columns)

### Priority

Medium

### Additional Context

Part of the AssetPicker work that reuses Content Drive UI inside File/Image field pickers. Do **not** create a GitHub tracked sub-issue under #36702 — this is an independent issue in the `N/7` series.

Folder tree migration is owned by #36733 and is out of scope here.

### LINKED ISSUE #36832: AssetPicker 2/7: Lift folder-tree helpers and harden drive data-access
### Description

Lift Content Drive folder-tree load helpers from the portlet into a shared location (`@dotcms/data-access` or a shared util usable by both Content Drive and AssetPicker), and harden `DotContentDriveService` so dialog hosts can provide/use it easily.

**Related epic:** #36702  
**Series:** AssetPicker 2/7  
**Depends on / align with:** #36733 (shared tree node models)

**Helpers to lift (from Content Drive portlet `utils/`):**
- `getFolderHierarchyByPath`
- `getFolderNodesByPath`
- tree node builders (`buildTreeFolderNodes` / `createTreeNode` equivalents)

**Service:** `DotContentDriveService` (`POST /api/v1/drive/search`) — today not `providedIn: 'root'`; make it easy to provide from AssetPicker dialog providers or document the required providers clearly (prefer `providedIn: 'root'` if safe).

### Acceptance Criteria

- [ ] Folder hierarchy / load-more helpers live in a shared lib (not portlet-only) and are usable by Content Drive and AssetPicker
- [ ] Helper node shapes align with the shared tree models from #36733 where applicable
- [ ] Content Drive sidebar still expands, selects, and load-mores as before (tests updated/passing)
- [ ] Unit tests cover the lifted helpers (happy path + edge cases for path hierarchy)
- [ ] `DotContentDriveService` can be used from a non-route host (dialog/picker) without ad-hoc route provider hacks — either `providedIn: 'root'` or a documented provider pattern used by Content Drive and AssetPicker
- [ ] No change to the `/api/v1/drive/search` contract

### Priority

Medium

### Additional Context

Independent issue in the AssetPicker `N/7` series for epic #36702. Not a tracked sub-issue.

### LINKED ISSUE #36833: AssetPicker 3/7: Decouple Content Drive toolbar chrome for AssetPicker reuse
### Description

Decouple picker-relevant Content Drive toolbar chrome from hard `inject(DotContentDriveStore)` so AssetPicker can reuse search, content-type filter, locale/language filter, chip filters, and upload button trigger as presentational / input-driven components.

**Related epic:** #36702  
**Series:** AssetPicker 3/7

**In scope to extract/share:**
- Search input (debounced text filter)
- Content-type filter UI
- Locale / language filter UI
- Upload button trigger (emit only; orchestration can stay wired by host)
- `DotChipFilter` / `DotFilterListItem` migration to `@dotcms/ui` if still only under `content-drive/ui` (update publishing-queue imports)

**Remain in Content Drive portlet (out of shared pieces):**
- Add New / EMA palette
- Workflow actions
- URL / query-param bound behavior
- Full store-injected toolbar shell (becomes thin adapter over shared pieces)

### Acceptance Criteria

- [ ] Shared toolbar pieces have **no** `inject(DotContentDriveStore)` — they use `input()` / `output()` (or equivalent host bindings)
- [ ] Content Drive toolbar still works end-to-end (search, filters, chips, upload trigger) via a thin store adapter
- [ ] `DotChipFilter` / `DotFilterListItem` are importable from `@dotcms/ui` (or remain shared without portlet coupling); publishing-queue (and any other consumers) imports are updated
- [ ] Add New, workflow, and URL-sync behavior remain portlet-owned and unchanged in product behavior
- [ ] Unit/component tests cover shared pieces and updated Content Drive toolbar wiring
- [ ] Pieces are suitable for composition inside AssetPicker (no router/URL dependency)

### Priority

Medium

### Additional Context

Independent issue in the AssetPicker `N/7` series for epic #36702. Upload *orchestration* parity with Content Drive is completed when the AssetPicker shell wires these triggers (later in the series).

### LINKED ISSUE #36834: AssetPicker 4/7: Headless browse store (no URL) for AssetPicker
### Description

Create a headless browse SignalStore for AssetPicker that builds the same style of `DotContentDriveSearchRequest` as Content Drive, but **without** `ActivatedRoute`, URL sync, or editor navigation.

**Related epic:** #36702  
**Series:** AssetPicker 4/7  
**Uses:** helpers/service from AssetPicker 2/7; models from `@dotcms/dotcms-models`; `DotContentDriveService` / `DotFolderService`

**Store responsibilities:**
- Site / asset path
- Sidebar folder tree state (expand / load-more via shared helpers)
- Items, pagination/sort, loading/error
- Filters: text, contentTypes, baseTypes, language, silent `mimeTypes`
- Always `showFolders: false` for picker browse requests
- Selection (single asset)

**Must NOT include:**
- Query-param / `Location.go` sync
- `DotContentDriveNavigationService` (edit content/page on double-click)
- Context menu / dragging / Add New dialogs

### Acceptance Criteria

- [ ] Picker-scoped SignalStore exists and is usable from an AssetPicker dialog host
- [ ] `$request` / search payload supports File vs Image style filters (locale, optional baseTypes, silent mimeTypes, `showFolders: false`)
- [ ] Unit tests assert request payload for: File (locale, no content type), Image (locale + baseTypes + `mimeTypes: ['image/*']`), and `showFolders: false`
- [ ] No dependency on `ActivatedRoute` or URL writing
- [ ] Does not navigate to content/page editors
- [ ] Uses shared drive search + folder APIs (no `/api/v1/browser`)

### Priority

Medium

### Additional Context

Independent issue in the AssetPicker `N/7` series for epic #36702. Example Image payload shape from the epic:

```json
{
  "assetPath": "//dotcms.com/QA/",
  "includeSystemHost": true,
  "mimeTypes": ["image/*"],
  "filters": { "text": "", "filterFolders": true },
  "showFolders": false,
  "archived": false
}
```

### LINKED ISSUE #36835: AssetPicker 5/7: Build DotAssetPicker shell in @dotcms/ui
### Description

Build `DotAssetPicker` in `@dotcms/ui`: a PrimeNG DynamicDialog shell that composes the shared folder tree, migrated folder list, shared toolbar chrome, and the headless browse store — a compact Content Drive for selecting a single asset.

**Related epic:** #36702  
**Series:** AssetPicker 5/7  

**Depends on (plain-text, not tracked sub-issues):**
- #36733 — shared folder tree in `@dotcms/ui`
- AssetPicker 1/7 — `DotFolderListView` in `@dotcms/ui`
- AssetPicker 3/7 — decoupled toolbar chrome
- AssetPicker 4/7 — headless browse store

**Behaviors:**
- Single-select assets in the right pane (`showFolders: false` — assets only)
- Confirm returns a contentlet (hydrate via existing contentlet service if needed); Cancel closes empty
- Double-click selects/confirms — does **not** navigate to the editor
- Upload button behaves like Content Drive (shared upload trigger + host orchestration)

### Acceptance Criteria

- [ ] `DotAssetPicker` (or equivalent name) lives under `core-web/libs/ui` and is exported from `@dotcms/ui`
- [ ] Dialog composes shared tree (#36733), list (1/7), toolbar chrome (3/7), and browse store (4/7)
- [ ] Right section lists **no folders** (assets only)
- [ ] Selection is single-select; Confirm enabled only when an asset is selected
- [ ] Confirm closes the dialog with a hydrated contentlet; Cancel closes without a result
- [ ] Double-click does not route to content/page editors
- [ ] Upload behaves the same as Content Drive for the selected folder/base-type flow
- [ ] Component tests cover open → select → confirm and upload happy paths
- [ ] Required providers (`DotContentDriveService`, etc.) are documented or provided at the dialog host

### Priority

Medium

### Additional Context

Independent issue in the AssetPicker `N/7` series for epic #36702. Entry-point File/Image configs and localStorage last path are handled in 6/7; Edit Content wiring in 7/7.

### LINKED ISSUE #36836: AssetPicker 6/7: File/Image entry configs, silent mime, CT limits, global last path
### Description

Implement AssetPicker entry configurations for File vs Image fields, silent mime filtering, content-type selector restrictions, and a **global** last-used asset path in `localStorage`.

**Related epic:** #36702  
**Series:** AssetPicker 6/7  

**Config API (illustrative):**
```ts
{ mode: 'file' | 'image'; languageId: number | string; initialAssetPath?: string }
```

| Mode | Locale | Content types / base types | Mime |
|---|---|---|---|
| File | Pre-select contentlet language | None pre-selected | None |
| Image | Pre-select contentlet language | Pre-select base types **dotAsset** + **File Asset** | Silent `mimeTypes: ['image/*']` — **no chip** |

**Content type selector:** only base types **dotAsset** and **File**, and content types whose base type is one of those. Widget/Content/etc. must not be selectable.

**Global last-used path:**
- On asset select, store asset path in `localStorage` (single system-wide key, e.g. `dotcms.assetPicker.lastAssetPath`)
- On open, restore that path when available/valid
- Not per-field — one global value

### Acceptance Criteria

- [ ] File mode pre-selects locale from contentlet language and pre-selects **no** content type
- [ ] Image mode pre-selects locale, pre-selects base types dotAsset + File Asset, and sends `mimeTypes: ['image/*']` on search requests
- [ ] Mime filter is transparent — **no chip** shown for mimeTypes
- [ ] Content type selector only offers base types dotAsset and File (and CTs of those base types); Widget/Content/other base types are not selectable
- [ ] Selecting an asset persists its asset path to a single global `localStorage` key
- [ ] Opening the picker uses the stored global path as the starting location when available/valid
- [ ] Changing the selected asset (same or different field) updates the same global key
- [ ] Unit/component tests cover config → request/filters, localStorage read/write, and absence of mime chip

### Priority

Medium

### Additional Context

Independent issue in the AssetPicker `N/7` series for epic #36702. Per-field User Searchable / Show in Listing remain out of scope (propagate from Content Drive later).

### LINKED ISSUE #36837: AssetPicker 7/7: Wire AssetPicker into Edit Content File/Image fields
### Description

Wire Edit Content **File** and **Image** fields to open `DotAssetPicker` instead of `DotBrowserSelector`, passing contentlet language and the correct mode.

**Related epic:** #36702  
**Series:** AssetPicker 7/7 (final integration)

**Current:** `dot-file-field` opens `DotBrowserSelectorComponent` via PrimeNG `DialogService` with `/api/v1/browser` params (`mimeTypes: ['image']` for Image, `[]` for File).

**Target:** Open AssetPicker with File/Image entry config from 6/7; browse via `/api/v1/drive/search`.

**Out of scope for this issue:**
- Block editor / edit-content-bridge consumers of `DotBrowserSelector` (keep using browser selector)
- Binary field (no existing-file picker)
- Tracked GitHub sub-issues under the epic

### Acceptance Criteria

- [ ] File field “Select existing” opens AssetPicker in **file** mode with the contentlet language
- [ ] Image field “Select existing” opens AssetPicker in **image** mode with the contentlet language
- [ ] Selecting an asset sets the field preview/value the same way as today’s browser selector close contract
- [ ] Cancel leaves the field unchanged
- [ ] `DotBrowserSelector` remains for non-File/Image consumers (block editor, bridges)
- [ ] File-field unit tests updated/passing for the new dialog open path
- [ ] Manual check: File + Image flows match epic ACs (locale, filters, upload, last path, no folders in right pane, silent mime for Image)

### Priority

Medium

### Additional Context

Closes the File/Image selection path for epic #36702. Design reference and full epic ACs live on #36702.

nicobytes:
This pull request introduces a set of improvements and refactorings to the folder tree utilities and related data-access APIs, aimed at enhancing code reuse, maintainability, and consistency across the Content Drive and Host Folder Field features. The changes include moving folder tree logic into shared utilities, updating service providers, and aligning data models.

**Core refactoring and utility extraction:**

* Introduced new shared utilities `folder-tree.utils.ts` and `folder-tree-load.utils.ts` in `@dotcms/data-access`, centralizing the logic for building, loading, and paginating folder trees. This includes new functions such as `generateAllParentPaths`, `createTreeNode`, and `buildTreeFolderNodes`, as well as utilities for paginated loading and "Load more" node handling. [[1]](diffhunk://#diff-c6bcf2c3f14bd2d62175250c71ba1c3fc4df7cb68ba70ebacaac8488e298773fR1-R128) [[2]](diffhunk://#diff-df6fb111e3b6f6713b0ecf362dfdf85da679b09414fa05a00e9fdc88d68407a7R1-R264)

**API and provider updates:**

* Updated `DotContentDriveService` to use Angular's `providedIn: 'root'` for global availability, and removed it from route-level providers. This ensures the service can be used from dialogs and other contexts without explicit injection in every route. [[1]](diffhunk://#diff-6fe4784dfb2a7ae4497316442c6c2ecdbc0a540f50c85b91f35f3897034cd225L10-R12) [[2]](diffhunk://#diff-2f24b415e1fedfc5138991280f5f1294017285073deea813aaef5222b52ee9edL3-R12)

**Imports and dependency cleanup:**

* Updated imports in various portlet files to use the new shared utilities from `@dotcms/data-access` instead of local utility definitions, and cleaned up duplicate or outdated imports for components such as `DotFolderListViewComponent`. [[1]](diffhunk://#diff-45741beedcade097fd5afb6156b294320d9e45479fbe3e26ebb2e8bed85bcb22R19) [[2]](diffhunk://#diff-45741beedcade097fd5afb6156b294320d9e45479fbe3e26ebb2e8bed85bcb22L30) [[3]](diffhunk://#diff-52568debba4600b47a2ab04ebd7bb746f891cd976cd9dae84b34378e350899fcL52-R57) [[4]](diffhunk://#diff-b3aeb1ff314c2ad9572eda3ad71f3085ee3086f6420083f53825b68cec271a7cL47-R61)

**Data model alignment:**

* Extended the `TreeNodeContentData` model to include additional metadata fields (`inode`, `defaultBaseType`, `fromTable`) for improved folder selection and upload behavior in the Content Drive and AssetPicker.

**Constants and type re-exports:**

* Refactored and re-exported shared folder tree page size constants from the new utilities, ensuring consistent pagination limits across features, and deprecated local type definitions in favor of those from `@dotcms/data-access`. [[1]](diffhunk://#diff-9c4bbae598ec2ab81e8d28bd417e5cb03ef9194cd4b8f4fa55dc20d0c446c453L1-R8) [[2]](diffhunk://#diff-9c4bbae598ec2ab81e8d28bd417e5cb03ef9194cd4b8f4fa55dc20d0c446c453L24-L37) [[3]](diffhunk://#diff-b67ceef3e85703d9a6bb3557792df13d7e08c016b36061c312163937594c206eR1-R15) [[4]](diffhunk://#diff-b67ceef3e85703d9a6bb3557792df13d7e08c016b36061c312163937594c206eL232-L243)

These changes improve code maintainability, reduce duplication, and ensure that folder tree behaviors remain consistent across all consuming features.

This PR fixes: #36702

**AssetPicker series — all seven sub-issues are resolved in this PR:**

Fixes #36831 — AssetPicker 1/7: Migrate DotFolderListView to @dotcms/ui
Fixes #36832 — AssetPicker 2/7: Lift folder-tree helpers and harden drive data-access
Fixes #36833 — AssetPicker 3/7: Decouple Content Drive toolbar chrome for AssetPicker reuse
Fixes #36834 — AssetPicker 4/7: Headless browse store (no URL) for AssetPicker
Fixes #36835 — AssetPicker 5/7: Build DotAssetPicker shell in @dotcms/ui
Fixes #36836 — AssetPicker 6/7: File/Image entry configs, silent mime, CT limits, global last path
Fixes #36837 — AssetPicker 7/7: Wire AssetPicker into Edit Content File/Image fields

This PR fixes: #36702


core-web/libs/ui/src/lib/components/dot-asset-picker/store/features/with-asset-folder-tree.feature.ts
```diff
@@ -0,0 +1,541 @@
...
+function searchFoldersInBrowsingSite(
+    roots: TreeNodeItem[],
+    term: string,
+    site: DotAssetPickerSite | undefined,
+    browsingService: DotBrowsingService
+): Observable<TreeLoadResult> {
+    const root = site ? findSiteRoot(roots, site.identifier) : undefined;
+
+    if (!site || !root) {
+        return of({ folders: roots });
```
adrianjm-dotCMS, line 149: The site being browsed can disappear from the sidebar entirely. `sitesPage()` filters the *sites* list with the same tree search term (`filter: activeSearchTerm() || '*'`). If the term matches a folder name but not the site's own name/hostname, the site gets filtered out of `roots`, `findSiteRoot` returns `undefined`, and we fall into `of({ folders: roots })` — no search runs and the current site vanishes from the tree.

Repro: while browsing "democms.com", type a folder name like "logos" that isn't part of the hostname → the site disappears and no results are shown at all, even though matching folders exist.

core-web/libs/ui/src/lib/components/dot-asset-picker/store/features/with-asset-folder-tree.feature.ts
```diff
@@ -0,0 +1,541 @@
...
+                                              store.path(),
+                                              browsingService
+                                          )
+                                ),
+                                tap(({ folders, selectedNode }) =>
+                                    patchState(store, {
+                                        folders,
+                                        selectedNode,
+                                        foldersStatus: ComponentStatus.LOADED
+                                    })
```
adrianjm-dotCMS: `TreeLoadResult` documents that an *absent* `selectedNode` means "leave the highlight alone" (distinct from `null`). But destructuring `({ folders, selectedNode })` and passing it to `patchState(store, { folders, selectedNode, ... })` turns that absent key into a present key with value `undefined` — and `patchState` merges via spread (`{...current, ...partial}`), so it overwrites the signal to `undefined` anyway. The highlight gets wiped in the exact same case as the comment above, contradicting the contract documented on `TreeLoadResult`.

Suggestion: only include `selectedNode` in the patch when it's actually present in the result (same conditional-spread pattern already used in `publish()` further down this file).

core-web/libs/ui/src/lib/components/dot-asset-picker/dot-asset-picker.component.ts
```diff
@@ -0,0 +1,465 @@
...
+        }
+
+        this.#contentletService
+            .getContentletByInodeWithContent(asset.inode)
+            .pipe(takeUntilDestroyed(this.#destroyRef))
+            .subscribe((hydrated) => {
+                // Persist on confirm, not on selection: a row the user highlights and then cancels
+                // must not move a value shared by every picker in the system.
+                writeLastAssetLocation(this.#resolveAssetLocation(hydrated));
+                this.#dialogRef.close(hydrated);
```
adrianjm-dotCMS: This `subscribe` has no error handler, unlike `#uploadByBaseType` elsewhere in this file, which shows a toast on failure. If `getContentletByInodeWithContent` errors (network blip, 404, permissions changed since selection), the error is silently swallowed, the dialog never closes, and the user gets no feedback — Confirm just looks like it did nothing.

Suggestion: add an `error:` handler (toast + keep dialog open), matching the pattern in `#uploadByBaseType`.

nicobytes:
@copilot resolve the merge conflicts in this pull request

core-web/libs/ui/src/lib/components/dot-asset-picker/dot-asset-picker.component.ts
```diff
@@ -0,0 +1,514 @@
...
+ * with nothing.
+ *
+ * It never navigates: double-clicking a row selects it rather than opening an editor, and no
+ * navigation service is injected, so routing away is impossible by construction.
+ */
+@Component({
+    selector: 'dot-asset-picker',
+    templateUrl: './dot-asset-picker.component.html',
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    providers: [DotAssetPickerStore, MessageService],
```
zJaaal: **🔴 Missing providers — the picker cannot open in the legacy Dojo binary-field host.**

`providers` covers `DotAssetPickerStore` and `MessageService`, but not:
- `DotHttpErrorManagerService` (`@Injectable()`, no `providedIn`) — injected in `store/features/with-asset-browse.feature.ts` and `with-asset-folder-tree.feature.ts`
- `DotContentTypeService` (`@Injectable()`) — injected by `dot-content-type-filter.component.ts` via `DotAssetPickerToolbarComponent`

Both are only provided in `apps/dotcms-ui/src/app/providers.ts`. `apps/dotcms-binary-field-builder/src/app/app.module.ts` — the host that bootstraps `dotcms-binary-field` — provides neither, and `DotFileFieldComponent` provides only `DialogService`.

**Repro:** legacy content editor, File/Image field with `allowExistingFile` → click "Select Existing File" → `DialogService.open(DotAssetPickerComponent)` → `DotAssetPickerStore` construction throws `NullInjectorError: No provider for DotHttpErrorManagerService` (then `DotContentTypeService`), dialog renders blank.

This is a regression: the replaced `DotBrowserSelectorComponent` used only root-provided services (`DotBrowsingService`, `DotUploadFileService`, `DotContentletService`). `dot-file-field.component.legacy-availability.spec.ts` only asserts construction, never dialog open, so nothing catches it — and `store/dot-asset-picker.store.spec.ts` mock-provides `DotHttpErrorManagerService`, which confirms the dependency is real.

Adding both to this `providers` array is the smallest fix and makes the picker self-sufficient in any host.

dotCMS/src/enterprise/java/com/dotcms/enterprise/publishing/sitesearch/SiteSearchAPIImpl.java
```diff
@@ -260,8 +260,6 @@ public SiteSearchResult getFromIndex(final String index, final String id) {
     public Optional<String> defaultIndexName() throws DotDataException {
         try {
             return router.readChecked(SiteSearchAPI::defaultIndexName);
-        } catch (DotDataException e) {
```
zJaaal, line 263: **🔴 Merge resolution dropped main's `DotDataException` passthrough (unrelated to this PR).**

`main` has `} catch (DotDataException e) { throw e; }` here, added by `ad09f0361c` (the #36983 sitesearch fix). It was removed by a merge commit on this branch, and it is still missing after the `7e1b2ae` re-merge of main — so merging this PR reverts that fix.

Consequence: any `DotDataException` from the read provider now falls through to `catch (Exception e)` and is re-wrapped as a new `DotDataException(e.getMessage(), e)`, losing the original subclass and nesting the cause one level deeper.

`isDefaultIndex()` three lines below still has the passthrough, so the two are now inconsistent — which is the clearest sign this was accidental. Please restore it.

core-web/libs/ui/src/lib/components/dot-asset-picker/store/features/with-asset-folder-tree.feature.ts
```diff
@@ -0,0 +1,552 @@
...
+                 * new object graph, so the selection has to be re-pointed by key or the highlight
+                 * silently disappears the first time a branch loads.
+                 */
+                const publish = (folders: TreeNodeItem[]): void => {
+                    const selectedKey = store.selectedNode()?.key;
+
+                    patchState(store, {
+                        folders,
+                        ...(selectedKey
+                            ? { selectedNode: findNodeByKey(folders, selectedKey) ?? null }
```
zJaaal: **🟠 `publish()` nulls the tree highlight after a sidebar search, silently retargeting uploads to the site root.**

`publish()` re-points `selectedNode` by key and sets it to `null` when `findNodeByKey` misses. But `searchFoldersInBrowsingSite` deliberately leaves `selectedNode` pointing at a node that is no longer in the published tree (documented intent), and `path` is left untouched. So the next `mutateNode` → `publish` wipes the highlight.

**Repro:** select `/images/banners/` in the tree (list scopes to it) → type "logo" in "Search sites & folders" → expand any node or click "Load more". `selectedNode` becomes `null`, so `$targetFolder` is `undefined` and `#uploadByBaseType` falls back to `hostFolder: browsingSite()?.identifier`. A dropped file is created at the **site root** while the asset list still shows `/images/banners/`. The folder's pinned `defaultBaseType` is lost too, so the user is re-prompted for Asset/File.

Either keep the previous node when the lookup misses, or clear `path` alongside it so the two cannot disagree.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-file-field/components/dot-file-field/dot-file-field.component.ts
```diff
@@ -803,42 +841,45 @@ export class DotFileFieldComponent
...
+            .subscribe({
+                next: (site) => {
+                    // Opening a picker that can't browse anything is worse than not opening it.
+                    if (site) {
+                        this.#openAssetPicker(site);
+                    }
+                },
+                // Nothing to browse and nothing to say beyond that — the picker simply doesn't open.
+                error: () => {
+                    /* noop */
```
zJaaal: **🟠 "Select Existing File" is now a completely silent no-op on failure.**

Both paths swallow it: `next` does nothing when `site` is falsy, and `error` is an empty `/* noop */`. No toast, no `Logger`, no UI message — compare line ~486 in this same file, which surfaces `SERVER_ERROR` for a comparable failure.

**Repro:** `DotSiteService.getCurrentSite()` 500s (or the session has no current site) → the user clicks "Select Existing File" and *nothing happens*, with no way to tell the click registered. Previously the dialog opened unconditionally, so this is a new dead end.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-file-field/components/dot-file-field/dot-file-field.component.ts
```diff
@@ -803,42 +841,45 @@ export class DotFileFieldComponent
...
-                sortByDesc: true
-            }
-        });
+    /** Opens the picker for a resolved site. Split out so the site lookup above stays readable. */
+    #openAssetPicker(site: DotSite) {
+        const isImage = this.$field().fieldType === INPUT_TYPES.Image;
+
+        const mode = isImage ? 'image' : 'file';
+
+        this.#dialogRef = this.#dialogService.open(
```
zJaaal, line 886: **🟠 Async dialog open can stack two pickers.**

`#openAssetPicker` assigns `this.#dialogRef` without checking whether a dialog is already open or in flight, and the open is now gated behind an HTTP round-trip that widens the race window.

**Repro:** double-click "Select Existing File" (or click again while `getCurrentSite()` is pending) → two `getCurrentSite()` requests → two `DotAssetPickerComponent` dialogs stacked, each with a live `onClose` subscription, while only the second is reachable through `#dialogRef`. The first leaks and can still write a selection. The previous synchronous open made this practically unreachable.

Guard on `#dialogRef`/an in-flight flag, or `exhaustMap` the site lookup.

zJaaal (commented):
## Review: AssetPicker / Content Drive refactor

Reviewed at `7e1b2ae` (136+ files). Focus was **bad practices and regressions in Content Drive**, plus the shared components the refactor extracted into `@dotcms/ui`. 11 findings inline, most severe first.

**Blocking, in my read:**
1. `DotAssetPickerComponent` is missing two providers, so it cannot construct in the legacy Dojo binary-field host (and now the Story Block path too).
2. A merge resolution dropped main's `DotDataException` passthrough in `SiteSearchAPIImpl` (unrelated to this PR, would land in main).
3. The folder-tree `publish()` nulls `selectedNode` after a sidebar search, so dropped files silently land at the site root.

**Content Drive specifically:** the biggest issue is `tree-folder.utils.spec.ts` going from 38 tests to 5 while `tree-folder.utils.ts` is unchanged and still live. That is not coverage moving with the code: every component that actually moved kept or grew its tests.

**Verified clean** (flagging these because they look alarming in the diff but check out):
- Removing `#fillWidth` is legitimate, not a lost fix: `title` is now unsized and absorbs the remainder, so the rescale is genuinely dead code.
- The dropzone extraction is faithful, and `resetContextMenu()` survives as the new `dragEnter` output.
- The upload-button label mapping is preserved exactly, `.toUpperCase()` included.
- The content-type filter keeps the monotonic cache, the `#cancelFetch$` race guard, the `ensure` param and the chip label fallback; the adapter keeps the same double `patchFilters`/`removeFilter` commit shape.
- The search input improved: the old placeholder was a hardcoded untranslated `"Search"`; it is now the `search` key, same text, plus value re-sync, trim and a distinct-emission guard.
- `DotContentDriveService` -> `providedIn: 'root'` is safe (stateless, `HttpClient` only).
- `ALL_FOLDER` / `SYSTEM_HOST_ID` have a single definition re-exported, so no split identity.
- Content Drive's main grid stays uncontrolled and passes no `selectionMode`/`showActions`/`titleOpensItem`, so it keeps the old defaults.
- Spec coverage carried over where things moved: folder-list-view 2417 -> 2584 lines / 183 tests, content-type filter 41 -> 52, dropzone and upload-selector 1:1.
- `[draggable]` -> `[attr.draggable]` is behavior-neutral; shell/sidebar spec changes are symbol renames with assertions intact.

_Reviewed by Claude (Claude Code), posted from @zJaaal's account._

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
```diff
@@ -272,24 +299,26 @@ export class DotFolderListViewComponent implements OnInit {
...
+     * shape the current mode expects.
      *
-     * @type {DotContentDriveItem[]}
      * @alias selectedItems
      */
-    get selectedItems(): DotContentDriveItem[] {
-        return this.$tableSelection();
+    get selectedItems(): DotContentDriveItem | DotContentDriveItem[] | null {
+        const items =
+            this.$selection() !== undefined
```
zJaaal, line 313: **🟡 Controlled/uncontrolled check regressed from `??` to `!== undefined`, so `null` now means "controlled and empty".**

The old getter was `this.$selection() ?? this.#internalSelection()`. This is now `!== undefined`, while the input type was simultaneously widened to accept `null`.

A consumer binding `[selection]="mySignal()"` where "nothing selected" is `null` — the natural shape for the `selectionMode="single"` this PR introduces — gets a permanently empty table selection, and clicking a row never visually checks it.

Both in-repo consumers pass arrays today (Content Drive's main grid is uncontrolled), so this is latent rather than live. But it is a trap for exactly the mode being added. `?? ` handled both; `!== undefined` only handles one.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/utils/tree-folder.utils.spec.ts
```diff
@@ -1,681 +1,37 @@
...
-            expect(ALL_FOLDER.expanded).toBe(true);
-        });
-
-        it('should not be a leaf node', () => {
-            expect(ALL_FOLDER.leaf).toBe(false);
-        });
-
-        it('should use a native PrimeNG folder icon', () => {
-            expect(ALL_FOLDER.icon).toBe('pi pi-folder');
+import { ALL_FOLDER } from '@dotcms/ui';
```
zJaaal, line 1: **🟠 Content Drive: the sidebar tree util suite was deleted (38 tests → 5), on production code that did not change.**

This file no longer imports `tree-folder.utils.ts` at all — the surviving 5 tests only assert the shape of the `ALL_FOLDER` constant, which moved to `@dotcms/ui`.

Deleted: `generateAllParentPaths` (8 tests), `createTreeNode` (12), `buildTreeFolderNodes` (13, including the "rootNode as selectedNode" fallback paths — empty hierarchy, no path match, empty target path).

`tree-folder.utils.ts` is **byte-identical to main** and still consumed by `store/features/sidebar/withSidebar.ts` and `utils/functions.ts`, so it still drives sidebar tree construction, parent-path expansion and selected-node resolution. That logic is now completely uncovered.

This isn't coverage moving with the code — everything that genuinely moved kept or grew its tests (folder-list-view 2417 → 2584 lines / 183 tests, content-type filter 41 → 52, dropzone 1:1). This file looks like collateral damage from the `ALL_FOLDER` relocation. Worth restoring the three suites, importing the utils from their current path.

core-web/libs/portlets/dot-content-drive/ui/src/lib/shared/models.ts
```diff
@@ -5,69 +5,13 @@ import type {
...
-/**
- * @export
- * @interface DotContentDriveUploadFiles
- * @description File and host folder for the drop zone
- */
-export interface DotContentDriveUploadFiles {
-    files: FileList;
-    targetFolder: DotFolderTreeNodeData;
-}
+export type DotContentDriveUploadFiles = DotUploadFiles;
```
zJaaal, line 14: **🟡 Content Drive: move/upload `targetFolder` lost its type guarantee.**

`DotContentDriveUploadFiles` was `{ files: FileList; targetFolder: DotFolderTreeNodeData }` — `targetFolder` **required** and folder-specific. It is now an alias of `DotUploadFiles`, whose `targetFolder` is **optional** and typed `TreeNodeData`. `DotContentDriveMoveItems = Omit<..., 'files'>` inherits that.

Meanwhile `getMoveMetadata` in `dot-content-drive-shell.component.ts` is unchanged and still reads `event.targetFolder.path` / `.hostname` with no guard. `core-web/tsconfig.base.json` sets `"strict": false`, so `strictNullChecks` cannot flag the now-optional property.

`TreeNodeLoadMoreData` is also assignable to `TreeNodeData` and carries no `inode` with optional `path`/`hostname`, so a move event built from a load-more node type-checks and yields a `pathToMove` of `//undefined/` that then goes to `bulkFire`.

To be fair on blame: the null-`targetFolder` crash predates this PR (`dot-tree-folder` `onDrop` already emitted `$activeDropNode()` unguarded). What this PR removes is the type that documented and enforced the contract. `onRequestUpload` was updated to narrow on `LOAD_MORE_NODE_TYPE`; `getMoveMetadata` was not — worth doing the same there, or keeping a Content-Drive-specific alias that requires folder data.

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
```diff
@@ -357,28 +386,28 @@ export class DotFolderListViewComponent implements OnInit {
     /**
      * The fixed columns being rendered. Filtered off `HEADER_COLUMNS` rather than off the caller's
      * list, so display order stays the table's.
+     *
+     * No width bookkeeping is needed when a column drops out: `title` is unsized (see
+     * `HEADER_COLUMNS`) and soaks up whatever the remaining sized columns leave over.
      */
     protected readonly $fixedColumns = computed<DotFolderListViewFixedColumn[]>(() => {
```
zJaaal, line 393: **🔵 Content Drive: the move stripped the comments that recorded previously-fixed bugs.**

The logic came across intact, but the "why" did not. Lost in this file:
- the `$columns` note explaining that extras anchor to where `type` sits in the **canonical** order rather than its rendered index — keying off the rendered index appended the extras when Type was hidden, putting every extra cell one heading early
- the item-title `title` attribute rationale (fixed-width column, `truncate` clips the rest)
- the `contentType` truncate rationale
- the locked-by-others color rationale

A later cleanup that "simplifies" `$columns` back to a rendered-index lookup reintroduces the misaligned-extras bug with nothing in the file warning it off. Cheap to carry these across.

core-web/libs/portlets/dot-content-drive/ui/src/index.ts
```diff
@@ -1,6 +1,18 @@
-export * from './lib/dot-folder-list-view/dot-folder-list-view.component';
+// Presentational list lives in @dotcms/ui; re-export for Content Drive consumers.
```
zJaaal, line 1: **🔵 Content Drive: two import paths now exist for the same symbols.**

This barrel re-exports `DotFolderListViewComponent`, `HEADER_COLUMNS`, `DOT_DRAG_ITEM` and the column types straight from `@dotcms/ui`, and Content Drive uses both entry points for the same classes:
- `dot-content-drive-shell.component.ts` imports `DotFolderListViewComponent` from `@dotcms/ui`
- `dot-content-drive-action-preview.component.ts` imports it (plus `DotFolderListViewColumnField`) from `@dotcms/portlets/content-drive/ui`

Same runtime class either way, but a future deprecation or signature change has to be chased through two barrels and neither path reads as canonical. Worth picking one and either dropping the re-export or marking it explicitly transitional.

core-web/libs/ui/src/lib/components/dot-workflow-push-publish/dot-workflow-push-publish.component.html
```diff
@@ -92,7 +92,7 @@
         <p class="text-sm text-surface-500">
             {{ $localTimezoneLabel() }}
             <a href="#" data-testid="push-publish-timezone-toggle" (click)="toggleTimezone($event)">
-                {{ ($showTimezone() ? 'hide' : 'Change') | dm }}
+                {{ ($showTimezone() ? 'hide' : 'change') | dm }}
```
zJaaal: **🔵 i18n key switched to one whose value is lowercase (and out of scope for this PR).**

`'Change'` → `'change'`. `Language.properties` has `Change=Change` (line 590) but `change=change` (line 589), so this timezone toggle now renders a lowercase "change" while its counterpart still renders "Hide" (`hide=Hide`). Inconsistent capitalization in the push-publish dialog.

Unrelated to Content Drive / AssetPicker — looks like an incidental edit worth reverting.

nicobytes:
## Review feedback status — rechecked at `e21bd64756`

Went through every Critical/High finding from @zJaaal, @adrianjm-dotCMS, Copilot and the `claude[bot]` pass. Below is what changed since the last review, verified against the code at this HEAD rather than from the commit messages.

### ✅ Addressed

**🟠 `SiteSearchAPIImpl` — merge dropped main's `DotDataException` passthrough** (`dfeca53acc`)
Restored the `} catch (DotDataException e) { throw e; }` in `defaultIndexName()`. The file is now **byte-identical to `origin/main`**, so merging this PR no longer reverts the #36983 fix, and it is consistent again with `isDefaultIndex()` / `getAggregations()`.

**🟠 Async dialog open can stack two pickers** (`dfeca53acc`)
Added an `#assetPickerPending` flag in `dot-file-field.component.ts`. Notes on the shape of the fix:
- `#dialogRef` could not serve as the guard: it is shared with the other three dialogs this component opens and is never reset on close, so checking it would have blocked the picker permanently after any of them ran.
- The flag is released on error, when no site resolves, and on **every** dialog close. That last one required dropping the `filter(file => !!file)` from the `onClose` subscription — with the filter in place, cancelling never released the guard and the button would have died for the rest of the session.
- Three tests added in `dot-edit-content-file-field.component.spec.ts`: double click while the site lookup is in flight, re-open after close, and retry after a failed lookup.

The same race existed on the two callers added later in this PR (`EditorModalService` for the Story Block, `DotWysiwygPluginService` for the WYSIWYG field); both already guard with an equivalent pending flag, so all three entry points are consistent now.

### ⚠️ Partially addressed — please re-read this one

**🔴 Missing providers / picker cannot construct in the legacy Dojo binary-field host**

`b27ab7557b` adds `DotHttpErrorManagerService` and `DotContentTypeService` to the picker's `providers`, as suggested. **That is necessary but not sufficient**, so this finding should stay open.

I added a regression spec that constructs `DotAssetPickerComponent` with only the provider set the custom-element host actually has (`dot-asset-picker.component.legacy-host.spec.ts`, `ec5aaccfca`). It still fails:

```
NG0201: No provider found for `DotAlertConfirmService`
```

`DotContentTypeService` is satisfied — it only needs `HttpClient`. `DotHttpErrorManagerService` is not; it transitively pulls in:

```
DotHttpErrorManagerService
  ├── DotAlertConfirmService   → ConfirmationService          ← fails here now
  ├── DotMessageDisplayService → DotRouterService → Router, DotEventsSocket
  └── DotRouterService         → Router
```

`apps/dotcms-binary-field-builder` bootstraps **without a Router** — which is the same reason `GlobalStore` was deliberately kept out of `DotFileFieldComponent`. The chain therefore cannot be closed with component-level plumbing at all: `provideRouter` returns `EnvironmentProviders` and cannot go in a component's `providers` array.

The fix that actually closes it is to drop the `DotHttpErrorManagerService` dependency from the picker's store features (6 call sites across `with-asset-browse` / `with-asset-folder-tree`) and report errors through the `MessageService` toast the picker already owns — it does exactly that for upload and confirm failures today. The spec is committed **skipped**, with the chain documented, to be un-skipped as part of that change.

### ❌ Still open

| Finding | Verified at this HEAD |
|---|---|
| 🟠 `publish()` nulls the tree highlight after a sidebar search, retargeting uploads to the site root | `with-asset-folder-tree.feature.ts:222` unchanged — still `findNodeByKey(folders, selectedKey) ?? null` |
| 🟠 `tree-folder.utils.spec.ts` gutted on unchanged production code | still 37 lines vs 681 on main, and still does not import `tree-folder.utils` |
| 🟡 "Select Existing File" gives no feedback when `getCurrentSite()` fails | still a silent return; only the new guard flag was added there |
| 🟡 `$selection() !== undefined` vs `??` | `dot-folder-list-view.component.ts:313` unchanged |
| 🟡 i18n `'change'` resolves to lowercase `change` | `dot-workflow-push-publish.component.html:95` unchanged |

### One thing worth a separate look

Two of the findings above (`SiteSearchAPIImpl`, and the `'Change'` → `'change'` i18n key) are both **merge-resolution damage from the same commit** (`5624a082d4`), not intentional changes — `main` has `'Change'` on that line. Since two independent regressions came in through one automated conflict resolution, a manual pass over the backend and CI files in `git diff origin/main` before merge seems worthwhile; I was not able to complete that sweep myself.

dotCMS/src/enterprise/java/com/dotcms/enterprise/publishing/sitesearch/SiteSearchAPIImpl.java
nicobytes: Fixed in `dfeca53acc`. Restored the `} catch (DotDataException e) { throw e; }` — the file is now **byte-identical to `origin/main`**, so merging this no longer reverts the #36983 fix and it is consistent again with `isDefaultIndex()` / `getAggregations()`. Good catch, this one was pure merge damage.

Worth noting the i18n finding further down came in through the *same* merge commit (`5624a082d4`), so a manual pass over the backend/CI files in `git diff origin/main` before merge seems warranted.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-file-field/components/dot-file-field/dot-file-field.component.ts
nicobytes: Fixed in `dfeca53acc`, with three tests in `dot-edit-content-file-field.component.spec.ts` (double click while the lookup is in flight, re-open after close, retry after a failed lookup).

Two notes on the shape, because the obvious fix does not work here:

- **`#dialogRef` cannot be the guard.** It is shared with the other three dialogs this component opens and is never reset on close, so checking it would have blocked the picker permanently after any of them ran. Used a dedicated `#assetPickerPending` flag instead.
- **The flag has to be released on `onClose` unconditionally.** That meant dropping the `filter(file => !!file)` from that subscription — with the filter in place, cancelling never released the guard and the button was dead for the rest of the session.

The same race existed on the two callers added later in this PR (`EditorModalService` for the Story Block, `DotWysiwygPluginService` for the WYSIWYG field); both already guard with an equivalent pending flag, so the three entry points are consistent now.

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
nicobytes: Done in `feb2a04429` — all four carried across.

One is reworded rather than copied verbatim: the item-title rationale said "the column is a fixed width", which is no longer true on this branch. `title` is now unsized so it absorbs whatever the sized columns leave over (that is what made `#fillWidth` dead code), so the clipping comes from the cell's `minmax(0,1fr)` track instead. Kept the *why* — `title` is what keeps truncated text reachable on hover — and updated the mechanism.

The `$columns` one is the load-bearing note you flagged, and it is back verbatim.

core-web/libs/ui/src/lib/components/dot-asset-picker/dot-asset-picker.component.ts
nicobytes: Applied in `b27ab7557b` — but **it is not enough, so please keep this open.**

I added a spec that constructs `DotAssetPickerComponent` with only the provider set the custom-element host actually has (`dot-asset-picker.component.legacy-host.spec.ts`, `ec5aaccfca`). With both providers in place it still fails:

```
NG0201: No provider found for `DotAlertConfirmService`
```

`DotContentTypeService` is satisfied — it only needs `HttpClient`. `DotHttpErrorManagerService` is not; it transitively pulls in:

```
DotHttpErrorManagerService
  ├── DotAlertConfirmService   → ConfirmationService          ← fails here now
  ├── DotMessageDisplayService → DotRouterService → Router, DotEventsSocket
  └── DotRouterService         → Router
```

`apps/dotcms-binary-field-builder` bootstraps **without a Router** — the same reason `GlobalStore` was deliberately kept out of `DotFileFieldComponent`. So the chain cannot be closed with component-level plumbing at all: `provideRouter` returns `EnvironmentProviders` and cannot go in a component's `providers` array.

What actually closes it is dropping the `DotHttpErrorManagerService` dependency from the store features (6 call sites across `with-asset-browse` / `with-asset-folder-tree`) and reporting through the `MessageService` toast the picker already owns — it does exactly that for upload and confirm failures today. The spec is committed **skipped**, with this chain documented in it, to be un-skipped as part of that change.

Happy to do that refactor here if you'd rather not defer it — it trades the repo's usual `httpErrorManager.handle()` convention for a self-sufficient leaf component, so it's worth an explicit call.

core-web/libs/ui/src/lib/components/dot-asset-picker/store/features/with-asset-folder-tree.feature.ts
nicobytes: Fixed. `searchFoldersInBrowsingSite` now re-adds the browsing site to `roots` when the site filter drops it (`with-asset-folder-tree.feature.ts:157-159`: `existingRoot ?? mapSiteToTreeNode(site)`, then `existingRoot ? roots : [root, ...roots]`), so searching a folder name that is not part of the hostname no longer makes the site vanish. Verified at `feb2a04429`.

core-web/libs/ui/src/lib/components/dot-asset-picker/store/features/with-asset-folder-tree.feature.ts
nicobytes: Fixed — the conditional spread you suggested is what shipped: `...(selectedNode !== undefined ? { selectedNode } : {})` at `with-asset-folder-tree.feature.ts:323`, so an absent key no longer becomes a present `undefined` through `patchState`'s merge. Verified at `feb2a04429`.

Note the sibling finding about `publish()` is a *different* path to the same symptom and is still open — that one re-points by key and lands on `null` when the key is absent from the published tree.

core-web/libs/ui/src/lib/components/dot-asset-picker/dot-asset-picker.component.ts
nicobytes: Fixed — `confirm()` now has an `error:` handler that toasts and leaves the dialog open (`dot-asset-picker.component.ts:240`), matching the `#uploadByBaseType` pattern you pointed at. Verified at `feb2a04429`.

core-web/libs/ui/src/lib/components/dot-asset-picker/store/features/with-asset-folder-tree.feature.ts
nicobytes: Confirmed still open at `feb2a04429` — `with-asset-folder-tree.feature.ts:222` is unchanged (`findNodeByKey(folders, selectedKey) ?? null`).

Reproduced the reasoning: `searchFoldersInBrowsingSite` deliberately leaves `selectedNode` pointing outside the published search tree, so the next `mutateNode → publish` cannot find the key and nulls it, `$targetFolder` becomes `undefined`, and `#uploadByBaseType` falls back to the site root. Silent data misplacement, so it needs fixing before merge.

Not fixed yet because your two options are not equivalent and I'd rather you pick:
1. **Keep the previous node when the lookup misses** — preserves the highlight and the upload target, but the highlight then points at a node not present in the visible tree (which is already the documented state after a search).
2. **Clear `path` alongside it** — the two can no longer disagree, but the asset list silently widens to site-wide while the user is mid-search.

My read is (1) is closer to intent, since the search is explicitly scoped to the sidebar and is not meant to change the list scope. Say the word and I'll implement it.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/utils/tree-folder.utils.spec.ts
nicobytes: Confirmed still open at `feb2a04429`. Verified the specifics: the spec is **37 lines vs 681 on `origin/main`**, no longer imports `tree-folder.utils` at all, while `tree-folder.utils.ts` is byte-identical to main and still consumed by `store/features/sidebar/withSidebar.ts` and `utils/functions.ts`.

So this is genuinely uncovered live logic, not coverage that moved. Agreed it should be restored rather than waived — it is the largest real gap in the PR. Not done yet only because it is ~640 lines of test restoration and I wanted the correctness findings landed first.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-file-field/components/dot-file-field/dot-file-field.component.ts
nicobytes: Still open — and to be precise about what changed: `dfeca53acc` touched this block, but only to release the new double-open guard. The **user-facing behaviour is unchanged**: a failed `getCurrentSite()` still returns silently with no toast and no `Logger`.

Flagging one thing for whoever decides this: the same pattern now exists in three places, since the Story Block (`EditorModalService`) and the WYSIWYG field (`DotWysiwygPluginService`) were given the identical "no site → don't open, say nothing" path later in this PR. So it is a single UX decision across all three, not a one-line fix here. If we want feedback, it should be added consistently in all three.

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
nicobytes: Still open at `feb2a04429` — `dot-folder-list-view.component.ts:313` is unchanged.

Agreed on the analysis, and worth stressing it is a trap for exactly the mode this PR adds: `selectionMode="single"` makes `null` the natural "nothing selected" value, and `!== undefined` treats that as "controlled and empty". Latent today only because the AssetPicker list is uncontrolled and Content Drive passes arrays.

Reverting to `??` is a one-liner. Only reason I have not is that it changes behaviour for a controlled consumer that deliberately passes `null` meaning "controlled, empty" — nothing in the repo does that today, so I think `??` is right, but it is your call since you traced the regression.

core-web/libs/portlets/dot-content-drive/ui/src/lib/shared/models.ts
nicobytes: Still open at `feb2a04429`.

Agreed on the framing, including your own caveat that the null-`targetFolder` crash predates this PR — what this PR removed is the type that documented the contract. `getMoveMetadata` still reads `event.targetFolder.path` / `.hostname` unguarded while `onRequestUpload` was narrowed on `LOAD_MORE_NODE_TYPE`, so the two are now inconsistent.

Leaving it for you to route: narrowing `getMoveMetadata` the same way is the small fix; keeping a Content-Drive-specific alias that requires folder data is the one that restores the compile-time guarantee. The second is better but touches the shared model this PR just unified, so it needs a call on which direction wins.

core-web/libs/portlets/dot-content-drive/ui/src/index.ts
nicobytes: Still open at `feb2a04429` — both entry points still resolve the same classes.

Fair point and worth doing, but it is a naming/ownership decision rather than a defect, and picking a canonical path means touching Content Drive imports that this PR already churned. Would rather land it as a follow-up than widen this diff further — happy to open the issue if you agree.

core-web/libs/ui/src/lib/components/dot-workflow-push-publish/dot-workflow-push-publish.component.html
nicobytes: Still open at `feb2a04429`, and confirmed it is **not** an intentional edit: `origin/main` has `'Change'` on that line, and the change came in through the merge commit `5624a082d4` — the same automated conflict resolution that dropped the `DotDataException` passthrough in `SiteSearchAPIImpl`.

So this is a one-character revert. I left it out only because it is outside this PR's surface and I did not want to bundle unrelated reverts without a nod — but given it is merge damage rather than a deliberate change, I think it belongs in this PR alongside the `SiteSearchAPIImpl` restore. Say the word and I'll revert it.

Two independent regressions from one automated merge is also the reason I'd suggest a manual pass over the backend and CI files in `git diff origin/main` before merge; I was not able to complete that sweep myself.

core-web/libs/ui/src/index.ts
```diff
@@ -7,36 +7,58 @@ export * from './lib/components/add-to-bundle/dot-add-to-bundle.component';
 export * from './lib/components/dot-action-menu-button/dot-action-menu-button.component';
 export * from './lib/components/dot-ai-image-prompt/ai-image-prompt.component';
 export * from './lib/components/dot-api-link/dot-api-link.component';
+export * from './lib/components/dot-asset-picker/dot-asset-picker.component';
+export * from './lib/components/dot-asset-picker/asset-picker-config';
```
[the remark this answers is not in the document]
nicobytes: Fixed. `DotFileFieldComponent.showSelectExistingFileDialog` now opens `DotAssetPickerComponent` (via `buildAssetPickerConfig` + `buildAssetPickerDialogConfig`), with the field mode (`image` / `file`), the contentlet language and the resolved site. The Story Block and the WYSIWYG field were wired to the same picker later in this PR, so all four entry points now go through `/api/v1/drive/search`. Verified at `feb2a04429`.

core-web/libs/ui/src/lib/components/dot-asset-picker/store/dot-asset-picker.store.ts
```diff
@@ -0,0 +1,123 @@
...
+            setSearch: (title: string): void => {
+                const filters = { ...store.filters() };
+
+                if (title) {
+                    filters.title = title;
+                } else {
+                    delete filters.title;
+                }
+
+                patchState(store, { filters, path: undefined, ...resetPaging() });
```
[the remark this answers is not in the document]
nicobytes: Fixed — `setSearch` now clears `path` **and** calls `selectRootNode()` (`dot-asset-picker.store.ts:149-153`), so the sidebar highlight follows the widened scope instead of pointing at a folder the list is no longer scoped to. That also fixes the upload target, since `$targetFolder` reads the same node. Verified at `feb2a04429`.

core-web/libs/ui/src/lib/components/dot-asset-picker/dot-asset-picker.component.ts
nicobytes: **Now actually closed** — in `852bf9ca12`. Superseding my previous reply on this thread: please disregard the "keep this open" ask.

Took the route described above rather than more provider plumbing. `DotHttpErrorManagerService` is gone from the picker's store features: the 6 call sites across `with-asset-browse` / `with-asset-folder-tree` now record the failure as `requestError: { messageKey }` state, and the picker component toasts it through the `MessageService` it already owns for upload and confirm failures. That also puts the store back in line with the repo rule that a store holds data and not UI — `handle()` was already doing UI from in there.

`DotContentTypeService` stays in `providers`; it only needs `HttpClient`.

The legacy-host spec is **un-skipped and passing**. Two things I got wrong on the first pass and fixed while doing this:

- The spec's provider set was an arbitrary subset, not the real host. Un-skipping it failed on `DotWorkflowActionsFireService`, which `apps/dotcms-binary-field-builder/src/app/app.module.ts` *does* provide. It now mirrors that module exactly, so it fails on anything router-bound being reintroduced rather than on my own omissions.
- The store spec still `mockProvider`-ed `DotHttpErrorManagerService`. Left in place it would have satisfied the dependency and kept passing if someone put it back, so it is removed.

Also added three tests for the new reporting path, including that a **second identical failure is reported again** — `requestError` is a fresh object per failure rather than a bare key precisely so the effect re-runs; otherwise a repeated outage would go silent after the first toast.

Verified: `ui` 170/170 (the legacy-host spec included), `edit-content` 219/219, `new-block-editor` 80/80, lint clean.

nicobytes:
## Two changes pushed that reviewers should know about (`852bf9ca12`, `157ef01ba6`)

### 1. The 🔴 provider finding is now genuinely closed

Details in [the thread](https://github.com/dotCMS/core/pull/36944#discussion_r3785706549). Short version: `DotHttpErrorManagerService` is out of the picker's store features — the store records failures as state, the component toasts them — and the legacy-Dojo-host spec is un-skipped and passing. My earlier "please keep this open" comment no longer applies.

### 2. I raised a bundle budget, which nobody asked for

`nx run dotcms-block-editor:build:production` was failing:

```
✘ [ERROR] bundle initial exceeded maximum budget. Budget 4.00 MB was not met by 80.43 kB with a total of 4.08 MB.
```

`apps/dotcms-block-editor/project.json` is byte-identical to `main`, so the budget was not moved — the bundle grew. **Cause is this PR:** wiring the Story Block to the picker means `EditorModalService` holds a static reference to `DotAssetPickerComponent` (`dialogService.open` needs the class), so the picker and its subtree — folder-list table, folder tree, splitter, content-type filter, Content Drive store — land in the initial chunk. It replaced `DotBrowserSelectorComponent`, which was considerably lighter.

I raised warning 3.5mb → 4.25mb and error 4mb → 4.5mb to unblock the build.

**Flagging it because it is a budget raise, not an optimization**, and it is not confined to the demo app: every consumer of `@dotcms/new-block-editor` now ships the picker in its initial bundle, `dotcms-ui` included.

The alternative is loading the picker through a dynamic `import()` in `openAssetPicker`, moving its subtree to a lazy chunk and keeping the original budget. The picker only ever opens on user action, so lazy is semantically right. I did not do it because I could not verify it is sufficient on its own — `@dotcms/ui` is already pulled in statically for `DotAIImagePromptComponent`, though esbuild tree-shakes per symbol so the picker's subtree should still move.

Two caveats on my own diagnosis, so nobody takes it as measured fact:
- I did **not** confirm whether this build passes on `main`. I attributed the growth by reasoning about the dependency graph, not by an A/B measurement — my attempt to measure it by stubbing the import did not come out clean and I reverted it rather than leave the tree dirty.
- I did not measure the picker's exact contribution.

Happy to do the lazy-import route instead if you would rather not carry the raise.

adrianjm-dotCMS (approved):

zJaaal (approved):

PR #36946, opened 2026-08-07
perf(ci): build only the frontend projects the WAR actually contains (#36945)

### LINKED ISSUE #36945: CI: artifact build compiles frontend and backend modules the WAR never contains
## Problem

The `Initial Artifact Build` is the **serial prefix** of every PR run — all ~25 test jobs wait on it. It takes ~14.2m, of which ~7.0m is Maven. A large share of that builds things the dotCMS WAR never contains.

### Frontend

`core-web` runs `nx run-many -t build` across the whole workspace (2.85m). But the WAR's `webResources` block lists exactly five paths:

```
dist/apps/dotcms-ui                    -> dotAdmin
dist/libs/dotcms-webcomponents/dist
dist/apps/dotcms-binary-field-builder
dist/apps/dotcms-block-editor
dist/libs/sdk/analytics-standalone     (its own `build-analytics` execution)
```

Everything else — `sdk-*`, `ai-evals`, `dotcdn` — is built and discarded. Both SDK publishers (`deploy-javascript-sdk`, `publish-sdk-next`) start from a clean checkout and rebuild `sdk-*` themselves; `deploy-javascript-sdk` even passes `--skip-nx-cache`.

This costs twice, because the execution is bound to the `compile` phase:
- the serial artifact build, and
- the **33m Frontend Unit Tests** job, where `mvn test -pl :dotcms-core-web` runs `compile` first — so that job is dominated by a full workspace build before `nx affected -t test` runs at all.

### Backend

`clean install` at the root builds all 24 modules unconditionally, including ones the image cannot contain:

```
 32.7s  dotcms-integration     then recompiled by all 7 IT shards anyway
 22.9s  dotcms-api-data-model  CLI
 17.6s  dotcms-cli             CLI
  7.4s  dotcms-postman
  6.6s  dotcms-ui-e2e
  1.3s  dotcms-test-karate
------
 1.48m of the 7.0m reactor (21%)
```

The CLI is 40.5s of every build and is doubly redundant: the CLI Tests job runs `verify -pl :dotcms-api-data-model,:dotcms-cli`, which recompiles those sources itself — and CLI tests are already gated behind the `cli_test` filter, so most PRs never run them.

## Scope

- [x] Narrow the frontend build to the projects the WAR contains
- [ ] Trim the Maven reactor so the artifact build skips test-only and CLI modules

## Out of scope

Remote caching (nx or WAR-level reuse across PRs) — tracked separately; it needs a storage decision.

wezell:
Fixes #36945 (frontend half; the Maven reactor trim is still open on that issue).

The artifact build ran `nx run-many -t build` over the **whole** workspace. The WAR contains four of those projects:

```
core-web/pom.xml webResources
  dist/apps/dotcms-ui                    -> dotAdmin
  dist/libs/dotcms-webcomponents/dist
  dist/apps/dotcms-binary-field-builder
  dist/apps/dotcms-block-editor
  dist/libs/sdk/analytics-standalone     <- built by its own `build-analytics` execution
```

Everything else it built — `sdk-*`, `ai-evals`, `dotcdn` — is never packaged and never reused. Both SDK publishers (`deploy-javascript-sdk`, and `publish-sdk-next` on trunk) start from a clean checkout and rebuild `sdk-*` themselves; `deploy-javascript-sdk` even passes `--skip-nx-cache`. That work was thrown away every single build.

## Why this lands twice

The execution is bound to the `compile` phase, so it runs in more than the build job:

| | before |
|---|---|
| `Initial Artifact Build` | ~2.85m of nx build, on the **serial prefix** that gates every test job |
| `Frontend Unit Tests` | 33m — `mvn test -pl :dotcms-core-web` runs `compile` first, so a full workspace build ran before `nx affected -t test` did anything |

The second one is the surprise: that job's cost is mostly the build, not the tests. `unit-test` is already `nx affected`.

## Why narrowing is safe

- `nx.json` `targetDefaults.build` has `dependsOn: ["^build"]`, so each leaf project still builds its dependency libs. Listing the four leaves is sufficient.
- `targetDefaults.test` has **no** `^build` dependency and runs against source, so unit tests are unaffected.
- The four names were read from their `project.json` files, and the list mirrors the `webResources` block directly above it in the same pom.

The list is a `<nx.build.projects>` property so a profile can widen it, with a comment tying it to `webResources` so the two stay in sync.

## Verification

`pom.xml` parses; the four project names confirmed against `apps/*/project.json` and `libs/dotcms-webcomponents/project.json`. Not run locally (the worktree has no `node_modules`) — CI is the check, and the signal to watch is that `dotAdmin` and the three sibling apps are still present in the built WAR.

## Note

I originally expected nx **caching** to be the bigger win here. It is not something to reach for blindly: `.nx/cache` is 2.0 GB on a working machine, and the GitHub Actions cache is 10 GB per repo with LRU eviction — a naive `actions/cache` would compete with the Maven repo cache and could make builds slower. Doing that properly needs a real remote cache, which is a separate decision. This PR is the part that needs no new infrastructure.


core-web/pom.xml
```diff
@@ -18,6 +18,21 @@
...
+          dotcms-binary-field-builder, dotcms-block-editor); analytics-standalone is
+          built by its own `build-analytics` execution.
+
+          Building the whole workspace here also built the sdk-* packages, which the
+          WAR does not contain - they are published separately, and both
+          deploy-javascript-sdk and publish-sdk-next rebuild them from a clean
+          checkout anyway, so that work was thrown away. nx resolves each project's
+          dependencies via `dependsOn: ["^build"]`, so listing the leaves is enough.
+
+          Keep this list in sync with the webResources block further down.
```
sfreudenthaler: don't like this and it's gonna get missed.  Gonna put a small guard in to validate

sfreudenthaler (commented):
change is good and welcome, just one improvement i want to add in

sfreudenthaler:
Pushed a follow-up commit adding the drift guard we discussed: a `validate`-phase Ant check that fails the build if a `webResources` project isn't covered by `nx.build.projects` (or has its own dedicated execution, like `analytics-standalone`).

It immediately caught a real gap: `edit-content-bridge` is packaged into the WAR (`html/js/legacy_custom_field_bridge`) but wasn't in `nx.build.projects`, and nothing else in the four leaf projects appears to import it — so it would have silently stopped building once `-p ${nx.build.projects}` landed. Added it to the property alongside the guard.

Issue #36947, opened 2026-08-07
CI: content-addressed build caching across nx, Maven and Docker


wezell:
# Proposal: content-addressed build caching across CI

Status: proposal, needs a storage decision
Context: follows the measurement work in #36942 / #36945

## The principle

Nothing should be built twice from identical inputs — across runs, across
branches, across PRs.

Today every run rebuilds everything: the whole nx workspace, all 24 Maven
modules, and the Docker image, from scratch, on a fresh runner. The
`Initial Artifact Build` that does most of this is the **serial prefix** gating
all ~25 test jobs.

The fix is content-addressed caching: hash a unit of work's real inputs, and if
that hash has been seen, fetch the output instead of recomputing it. Three
layers, three tools.

| Layer | Tool | Today |
|---|---|---|
| Frontend (`nx run-many -t build`) | nx remote cache | local only, discarded with the runner |
| Java modules (CLI, core, all 24) | Apache Maven Build Cache Extension | not used |
| Docker image | buildx registry cache | **no layer cache at all** |

The three are independent and can land in parallel.

## Why `actions/cache` cannot be the answer

GitHub Actions caches are **branch-scoped**. A run reads caches from its own
branch and the default/base branch — never a sibling branch. Two PRs building
identical code share nothing. That is a deliberate security boundary.

So `actions/cache` only supports *save-on-trunk, restore-everywhere*: a PR
benefits only from what `main` has already built.

Second problem: `.nx/cache` is **2.0 GB** on a working machine. The Actions
cache is 10 GB per repo with LRU eviction, already shared with the Maven
repository, pnpm store and Node binary caches. A large, frequently-rewritten
entry would evict the others and could make builds *slower*.

A real remote cache is keyed purely on the content hash — no branch dimension,
no 10 GB ceiling.

## Runner geography (measured, not assumed)

12 samples via PR #36948, cross-checked two ways (Azure IMDS `.location` and the
`Azure Region:` line GitHub prints in every job log header — they agreed 100%):

```
centralus       x3   Iowa
westus2         x3   Washington
eastus          x2   Virginia
northcentralus  x2   Illinois
westus          x1   California
eastus2         x1   Virginia
```

Runners are spread across **six US Azure regions**, coast to coast. No region
dominates. Consequences:

- A **single-region bucket is the wrong shape** — fast for roughly one job in
  six, slow for the rest (westus↔eastus is ~60–70ms RTT).
[...]

PR #36961, opened 2026-08-07
perf(ci): back the Maven build cache with a shared S3 bucket (#36947)

wezell:
Closes part of #36947. **Stacked on #36960** — review that one first; this PR's diff is against it.

## What

Gives the Apache Maven Build Cache Extension somewhere to keep its results: a shared S3 bucket on OVH object storage.

The extension was **already installed and enabled** (`.mvn/extensions.xml`, v1.2.0) — but local-only, so on a fresh CI runner it cached nothing. The only missing piece was shared storage.

Scoped to the **Initial Artifact Build**, the serial prefix that gates every test job. Test phases are deliberately *not* cached: a memoised green run would mean "we did not run", and this suite has measured flakes plus a hang that has burned a 122m job timeout.

## Transport

The extension speaks HTTP `PUT`/`GET`/`HEAD`; S3 needs SigV4. Rather than add a Maven S3 wagon — the available ones aren't maintained (`seahen` 1.3.3 is 2021, `gkatzioura` 2.3 is 2019, both AWS SDK v1) — [`aws-sigv4-proxy`](https://github.com/awslabs/aws-sigv4-proxy) runs as a container and signs on the way out, and Maven talks to `127.0.0.1`. **No new Maven dependency.**

## Security

A build cache untrusted code can write is a supply-chain vector: a poisoned entry is replayed as a build output on a trusted ref. This is [CVE-2025-36852](https://www.cve.org/CVERecord?id=CVE-2025-36852) ("CREEP"), which killed Nx's `@nx/s3-cache` and its siblings.

**The control is the credential, not the client** — a job holding a writable key can bypass Maven entirely with one `aws s3 cp`, so `remote.save.enabled` is defence in depth, not the boundary.

| Ref | Key | Writes |
|---|---|---|
| PR | `..._ACCESS_KEY_RO` (GetObject only) | no |
| merge queue / trunk | `..._ACCESS_KEY` | yes |
| fork PR | none — builds uncached | no |

Plus:
- `remote.save.final=true` — an existing entry is never overwritten.
- The action **asserts** its read-only key is read-only (one `PUT`, expects `403`) instead of assuming it. A writable "read-only" key looks identical to a correct setup until abused.
- Writing builds record a `provenance.json` beside each entry (ref, sha, run id, actor), first-writer-wins — nothing else in a bucket says which ref produced a hash.

## The subtle one: `alwaysRunPlugins`

Load-bearing, not tuning. On a cache hit the extension skips cached plugin executions **including `install:install`** — measured: 1 jar in `~/.m2/repository` after a cold build, **0 after a hit**. This job exists to publish that repository as the `maven-repo` artifact ~25 test jobs consume. Same story for `docker-maven-plugin:build`, which writes the `docker-build.tar` the next step uploads.

Both would have looked perfectly green on the cold populate run and broken everything on the first *warm* one.

## Verification

Against MinIO before any of this was wired:

| Check | Result |
|---|---|
| `PUT` / `GET` / `HEAD` through the proxy | `200` / `200` / `200` |
| Missing key | `404` (a `403` reads as a hard error, not a miss) |
| 10 MB body round-trip | byte-identical |
| Build with an **empty** local cache | `Found cached build, restoring … by checksum` |
| Remote unreachable | build exits `0`, logs an error, rebuilds |

And separately, because `cicd_comp_build-phase.yml` fails a PR on a dirty tree while `openapi.yaml` is a tracked file generated at compile: built `:dotcms-core --am` twice — 12 modules restored including `dotcms-core`, `openapi.yaml` md5 identical, `git status` unchanged.

## Releases are not affected

Release, LTS, nightly, manual-deploy and CLI-release workflows don't pass the secrets, and **no workflow in this repo uses `secrets: inherit`** (verified). The action sees empty credentials and exports an empty `BUILD_CACHE_ARGS`; those pipelines build from scratch exactly as today. A release is the build where "we didn't actually compile this" is least acceptable and the saving is worth least.

## Turning it off

| Scope | How |
|---|---|
| One PR | label `CI: no build cache` |
| Everything, now | repo/org variable `BUILD_CACHE_DISABLED=true` |
| Local build | `-Dmaven.build.cache.enabled=false` |
| Force rebuild, still publish | `-Dmaven.build.cache.skipCache=true` |

## Expected effect, honestly

Ceiling is the ~7.0m of Maven time inside a 14.2m build job, against a 74–103m PR wall clock. Real but modest — the twin-tail shard rebalance (#36943) is still the bigger lever. `dotcms-core-web` only participates once #36960 lands.

**The first merge-queue run is the canary**: it's what proves the OVH SigV4 handshake and the region derived from the endpoint host. If either is wrong the cache disables itself with a warning rather than failing the build.

This PR fixes: #36947


docs/core/CI_REMOTE_CACHE_PROPOSAL.md
sfreudenthaler, line 1: Suggust that we don't need this or that it move to the speckit stuff.

sfreudenthaler (approved):
fine as is, nitpik that we don't include the proposal.md or shift it into a full spec via speckit

docs/core/CI_REMOTE_CACHE_PROPOSAL.md
wezell: Agreed — removed in 7783ee78d4.

It was a third copy: the same content is the body of #36947, and the PR description already carries what a reviewer needs in front of them. Three copies of the same rationale is two too many, and they'd drift apart the first time any one of them changed.

Design rationale that should outlive the PR now lives in #36947; nothing is lost.

wezell:
Re the review findings:

**`s3:ListBucket` on the RO key — resolved.** The read-only key does grant `ListBucket`, so a `HEAD`/`GET` on a not-yet-cached key returns `404` rather than `403`. That was the right thing to question: the MinIO verification wouldn't have caught it either way, since MinIO returns `404` for missing keys largely regardless of the grant. With GetObject-only it would have turned every cache miss into a hard error and silently disabled the cache on every PR.

The remaining findings, in the order I'd weight them:

1. **`alwaysRunPlugins` on a *warm* hit** — highest blast radius and still unverified. A hit that skips `install:install` / `docker-maven-plugin:build` breaks the ~25 downstream jobs that consume the `maven-repo` artifact and the docker image, and it would look green on a cold populate run. Worth one warm run showing the module jar lands in `~/.m2/repository` and `docker-build.tar` is written *after* a `Found cached build` line.
2. **Unpinned `aws-sigv4-proxy:latest`** — it is handed the RW credential on trunk and merge-queue. Pinning to a digest is cheap and matches the trust bar the rest of the PR sets.
3. **Region derived via `cut -d. -f2`** — non-blocking, since a wrong region surfaces as a signing failure and the cache self-disables. Passing it explicitly from the workflow would make it auditable.

wezell:
Ran the warm-hit check on `alwaysRunPlugins`. **The matcher format works** — this was the highest-blast-radius finding, so here is the evidence rather than an assurance.

Test: build a module with the cache on to populate it, delete the installed jar from `~/.m2`, rebuild, and see whether a cache hit re-installs it.

```
md5 before: ac0a74ac407a9f7653b3ff6f8a460909
deleted; jar absent from ~/.m2

[INFO] Found cached build, restoring com.dotcms.core.plugins:com.dotcms.tika-api
       from cache by checksum 97c25cd239626f7e190aec64df17680ead2ce67652b7d0ebec9f8211894a9f87
[INFO] Skipping plugin execution (cached): jar:jar
[INFO] Mojo execution is forced by project property: install:install
[INFO] Installing .../com.dotcms.tika-api-1.0.0-SNAPSHOT.jar to ~/.m2/repository/...
[INFO] BUILD SUCCESS

md5 after:  ac0a74ac407a9f7653b3ff6f8a460909
```

So on a genuine cache hit the extension skips `jar:jar` but **forces `install:install`**, and the artifact lands in `~/.m2` byte-identical. The concern was that extension 1.2.0 might want `groupId:artifactId:goal` or globs — it does not; the `artifactId:goal` form in `ALWAYS_RUN` is matched, and the extension says so explicitly ("forced by project property").

**What this does not cover.** I could not exercise `docker-maven-plugin:build` locally: it only runs on `dotcms-core`, and `-pl :dotcms-core` without `--am` fails dependency resolution on the `core-web` WAR, so proving it needs a full `--am` build including the frontend. Since `docker-maven-plugin:build` uses the identical `artifactId:goal` syntax that is now proven to match, the format question is settled for both — but the docker execution itself is still unverified.

Worth confirming on the canary: that `dotCMS/target/docker-build.tar` is written **after** a `Found cached build` line for `dotcms-core`. That is one grep on the canary log, and it closes the finding completely.

PR #36732, opened 2026-07-27
feat(edit-content): open Edit Content in a slide-in side panel (#36631)

adrianjm-dotCMS:
### Proposed Changes

Opens the new Edit Contentlet in a **right-to-left slide-in side panel** (`p-drawer`, ~70% width) instead of a full-page navigation / dialog, behind the `FEATURE_FLAG_EDIT_CONTENT_SIDE_PANEL` feature flag.

This closes #36631.

#### Frontend

- **New `DotEditContentSidePanelComponent`** (`libs/edit-content`) — hosts the editor in a `p-drawer` teleported to `body` (`appendTo="body"`) so it escapes transform/overflow-clipping ancestors. Header with title, expand-to-full-screen and close (X) actions using Material Symbols icons. ESC closes the panel.
  - **Expanded/full-screen mode persists in `localStorage`**, so the next content opens in the same mode.
- **New `DotSidePanelNavController`** — collapses the main navigation (and the Content Drive tree) while a panel is open and restores it on close, only on narrower viewports (`< 1800px`). Ref-counted so stacked panels (relationship "New content") keep the nav collapsed until the last one closes; pre-panel state is kept in `localStorage` so it survives a refresh on a shared `?editContent=` link.
- **Consumers gated by the flag:**
  - **Content Drive** — edits/creates in the side panel.
  - **Query Tool** — opening a new-editor result opens the side panel instead of a new tab.
  - **UVE (`edit-ema`)** — edit + palette-drop open in the side panel (header cropping fixed).
  - **Relationship field "New content"** — opens a stacked side panel.

#### Backend

- New feature flag `FEATURE_FLAG_EDIT_CONTENT_SIDE_PANEL` (`FeatureFlagName.java`), whitelisted and exposed via `ConfigurationResource` (`/api/v1/configuration`), with a default in `dotmarketing-config.properties`.

### Checklist

- [x] Tests
- [ ] Translations
- [ ] Security implications
- [ ] Backward compatibility

> [!NOTE]
> Feature-flagged (`FEATURE_FLAG_EDIT_CONTENT_SIDE_PANEL`), on by default.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

This PR fixes: #36631

PR #36172, opened 2026-06-15
34420 unify component

nicobytes:
This pull request introduces significant updates to the handling and testing of binary fields (file/image uploads) in the content editing experience. The main focus is on unifying the binary field implementation under the new `dot-edit-content-file-field` component, updating related imports and tests, and improving Playwright E2E test coverage for both the new and legacy editors. The changes also include new and refactored E2E test helpers for binary fields, and adjustments to test data and selectors.

**Component and Module Refactoring:**

* The binary field implementation is unified under the `dot-edit-content-file-field` component, replacing the previous `dot-edit-content-binary-field` and its wrapper throughout the app and tests. This includes updating Angular module imports, dependency injection, and testbed configuration to use the new component and its bridge (`DotBinaryFieldCeBridgeComponent`). [[1]](diffhunk://#diff-dbc15f1d4256227845b1a2007977254c43f0c233786aed4b4a78e2c64870621dL10-R23) [[2]](diffhunk://#diff-dbc15f1d4256227845b1a2007977254c43f0c233786aed4b4a78e2c64870621dL33-R33) [[3]](diffhunk://#diff-8e4ebf6e3bd0e536ddd7567da45f2092c49a46801b2e8cfa22531cfd312d137fL3-R3) [[4]](diffhunk://#diff-685da2f87189fff825b7aec2666cfbabdb33c40f8f9d8c82df77ebb8652ccd4aL106-L114) [[5]](diffhunk://#diff-685da2f87189fff825b7aec2666cfbabdb33c40f8f9d8c82df77ebb8652ccd4aR156-R170) [[6]](diffhunk://#diff-685da2f87189fff825b7aec2666cfbabdb33c40f8f9d8c82df77ebb8652ccd4aR179) [[7]](diffhunk://#diff-23d2e2fd9b71e1819e10ae720cb79ecf5e4d75e51fee641b020017fe8186640fL40) [[8]](diffhunk://#diff-23d2e2fd9b71e1819e10ae720cb79ecf5e4d75e51fee641b020017fe8186640fL187-R191)

**E2E Test Enhancements:**

* Adds comprehensive Playwright E2E tests for the binary field image editor in both the new and legacy editors, including creation of content types, file import, and image editing workflows.
[...]

PR #30135, opened 2024-09-25
feat(edit-content): drag and drop functionality 

nicobytes:
### Parent Issue

#29872 

### Proposed Changes
* Create new component for Image & File component
* Create new state to handle drag and drop functionality
* Work in Preview component
* Upload file as dotAsset

### Checklist
- [x] Tests
- [x] Translations
- [x] Security Implications Contemplated (add notes if applicable)

### Screenshots

https://github.com/user-attachments/assets/90f5deb2-a416-4e7f-bfcd-db68d9d51a76

This PR fixes: #29872

PR #36965, opened 2026-08-07
fix(edit-content): side panel feedback — click-outside, 80% width, sidebar polish (#36631)

### LINKED ISSUE #36631: Open new Edit Content in a right-to-left slide-in panel (~70%) with routing and full-screen
### Description

The new Edit Content should open in a **slide-in panel** (animating **right → left**) instead of a centered dialog.

- **Default size:** **80%** of the **viewport width** (similar to how GitHub issues open within Projects).
- **Full-screen:** an option within the slide-in to **expand the editor to full screen** (and restore back to 80%).
- **Routing & navigation:** opening the editor is **deep-linkable** — the URL updates to a per-content route so it's shareable and refresh-safe, and browser **Back closes** the slide-in, returning to the previous view.
- **Feature flag:** the slide-in behavior is gated behind a **feature flag** so we can fall back to the **current full-screen** mode if needed. When the flag is off, the editor opens full screen as it does today.

> 🔀 **Supersedes #36347** (which specified a centered dialog at 80%). The still-relevant behaviors from #36347 — full-screen toggle, ESC to close, and the unsaved-changes confirmation — are carried into this issue. #36347 can be closed.

🎨 **Design:** https://claude.ai/design/p/f9eb2825-a811-4f0f-822b-1d37abfbbc17?via=share&file=Content+Drive+-+Dialog.html
💬 **Product Q&A (open questions):** https://dotcms.slack.com/archives/C0893BF7GRJ/p1784225950261049

### Acceptance Criteria

**Slide-in panel**
- [x] The new Edit Content opens as a **slide-in panel** animating from the **right to the left**
- [x] By default the panel occupies **80%** of the **viewport width**
- [x] The area behind the panel (left nav + remaining viewport) stays visible, consistent with the design

**Full-screen**
- [x] A control inside the slide-in **expands** the editor to full screen
- [x] The control **restores** the panel back to the default 80% size

**Routing & navigation**
- [x] Opening the editor updates the **URL** to a per-content route (deep-linkable / refresh-safe)
- [x] Loading that URL directly opens the editor for the corresponding content
- [x] Browser **Back** closes the slide-in and returns to the previous view
- [x] Navigation state stays consistent (no orphaned routes or stuck panels)

**Carried-over behaviors (from #36347)**
- [x] Pressing **ESC** closes the slide-in
- [x] **Clicking outside** the panel closes it, with the same behavior as ESC and the close control
- [x] Closing (ESC, click outside, close control, or Back) with **unsaved changes** shows the existing unsaved-changes confirmation; discard proceeds, keep-editing cancels
- [x] With no unsaved changes, closing dismisses immediately

**Feature flag**
- [x] The slide-in panel behavior is gated behind a **feature flag**
- [x] When the flag is **ON**, the editor opens as the right-to-left slide-in panel (this issue's behavior)
- [x] When the flag is **OFF**, the editor opens in the **current full-screen** mode (fallback with no slide-in/routing changes)
- [x] Toggling the flag switches between the two modes without requiring a code change

**Design parity**
- [ ] The slide-in matches the attached design

### Priority

High

### Additional Context

Part of the new Edit Contentlet work (`dotCMS: New Edit Contentlet`). This replaces the dialog container from #36347 with a right-to-left slide-in at 80% of the viewport width, plus routing/navigation so the editor is deep-linkable and Back-aware. Some behaviors are still pending product answers — see the linked Slack thread; update the AC as those are resolved. The design link is the visual source of truth.

adrianjm-dotCMS:
### Parent Issue

#36631 — the panel itself shipped in #36732; this PR addresses the review/QA feedback on top of it.

### What changed

**1. Click outside closes the panel (through the unsaved-changes guard)**

Clicking the area behind the panel now behaves exactly like ESC and the X button: it routes through the editor's unsaved-changes guard, so a dirty form prompts and "Keep editing" cancels the close.

PrimeNG's own `dismissible` is deliberately left **off** — it calls `close()` on mask click, which hides the drawer and emits `visibleChange(false)` immediately. That both bypasses the guard (unsaved edits lost silently) and desyncs the one-way `[visible]="!!data()"` binding. The mask click is matched at document level instead, since `appendTo="body"` moves the mask out of the component's subtree.

Edge cases covered: a click inside the panel is ignored; a drag that starts inside and ends on the mask resolves to a different target and is ignored; with stacked panels (relationship field) only the frontmost reacts, same as ESC.

**2. Default width 70% → 80%**

**3. Nav collapse threshold 1800px → 2560px**

At 80% the leftover strip is 20% of the viewport. Against the 270px expanded nav that left ~90px of content at 1800px — a useless sliver, so the nav stayed expanded for no benefit. At 2560 the strip is ~242px.

**4. Binary field preview: narrow-layout buttons no longer paint over the sticky tab bar**

`.dot-file-field-preview__actions--responsive` went from `z-index: 100` to `2`, in line with its siblings (`1` for the wide variant, `2` for remove — nothing in the component needs more than 2). `100` was the only value in the whole `edit-content` lib able to beat the form's sticky `.p-tablist` (`z-index: 10`), which carries the status chip and the permissions/references actions. Only reproduced on narrow containers, because that is the only layout where these filled blue/red variants render.

**5. Sidebar toggle icon → `dock_to_left`**

Replaces the two mirrored UVE SVGs (`left_panel_open.svg` / `left_panel_close.svg`) with a single Material Symbol. Since one glyph now covers both states, the open/closed cue moved to the button's `aria-label` and the icon is `aria-hidden` so screen readers don't announce the ligature text.

**6. Actions tab section order → Details, Locales, Workflow**

The shared `.dot-section` bottom-border override moved to Workflow (now last) so there is no divider dangling at the end of the tab. Persistence keys are unchanged, so each user's saved expand/collapse state carries over.

### Testing

Updated specs alongside each change:

- `dot-edit-content-side-panel.component.spec.ts` — 4 new tests for click-outside (closes via guard; does **not** close when the guard cancels; ignored when not frontmost; ignored on a click inside), plus the width assertions.
- `dot-side-panel-nav.service.spec.ts` — `setWideViewport()` pinned to the 2560 threshold (the comparison is exclusive, so the boundary must count as wide; bumping the threshold fails here on purpose).
- `dot-edit-content-form.component.spec.ts` — replaced the two-SVG test with one for the single `dock_to_left` glyph.
- `dot-edit-content-sidebar.component.spec.ts` — section-key order updated; noted that `queryAll` returns DOM order, so that test guards the order too.

Manual verification of the visual changes was done by the author.

This PR fixes: #36631

This PR fixes: #36631


nicobytes (approved):

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form.component.html
```diff
@@ -132,30 +132,28 @@
...
+             way). The glyph no longer changes per state, so the open/closed cue lives on the
+             button's aria-label — `aria-hidden` keeps screen readers off the ligature text. -->
         <div
             class="-mr-4 flex min-h-[52px] shrink-0 items-center justify-center"
             data-testId="sidebar-toggle">
             <p-button
                 [rounded]="true"
                 [text]="true"
                 (click)="$store.toggleSidebar()"
+                [attr.aria-label]="
```
oidacra: Scope: line added by this PR.

PrimeNG Button exposes `ariaLabel` as an `@Input` and binds it internally onto the real element with `[attr.aria-label]="ariaLabel || buttonProps?.ariaLabel"`. Putting `[attr.aria-label]` on the `<p-button>` host tag sets the attribute on the wrapper element instead, so the inner focusable `<button>` never receives it.

Combined with the new `<i>` being `aria-hidden="true"`, and the previous `` (which used to supply the accessible name) being removed, does that leave the toggle button with no accessible name at all?

Would using the component input instead work here?

```html
[ariaLabel]="(showSidebar ? 'edit.content.sidebar.close' : 'edit.content.sidebar.open') | dm"
```

Separately, and only for traceability against #36631: was the swap from the two mirrored UVE SVGs to the single `dock_to_left` glyph part of the review feedback, or an improvement you folded in along the way?

core-web/libs/edit-content/src/lib/components/dot-edit-content-side-panel/dot-edit-content-side-panel.component.ts
```diff
@@ -176,6 +183,32 @@ export class DotEditContentSidePanelComponent implements OnDestroy {
...
+     * `visibleChange(false)` the moment the mask is clicked, which both bypasses the guard (unsaved
+     * edits lost silently) and desyncs the one-way `[visible]` binding. Matching the mask ourselves
+     * keeps the panel on screen until the guard says it is safe to close.
+     *
+     * The listener is document-wide (the mask is built imperatively into `document.body`, so no
+     * template binding can reach it), which is why it filters to the mask element itself — a click
+     * inside the panel, or a drag that starts inside and ends on the mask, resolves to a different
+     * target and is ignored. As with ESC, only the frontmost stacked panel reacts.
+     */
+    protected onMaskClick(event: MouseEvent): void {
```
oidacra, line 227: Scope: handler added by this PR.

`p-drawer-mask` is a static class that every modal PrimeNG drawer applies to its own mask, so `target.classList.contains(DrawerClasses.mask)` matches any drawer mask currently in the document, not specifically this panel mask.

`dot-block-editor-sidebar.component.html` renders its own `<p-drawer>` (modal defaults to true), and it is mounted as a sibling of this panel in `edit-ema-editor.component.html` — the block editor sidebar at line 291 and the side panel at line 306, behind two independent signals. `DotSidePanelNavController.isTop()` only tracks Edit Content side panels, so it cannot distinguish the two.

If both are open at once, would clicking the block editor sidebar mask also close the Edit Content panel underneath, prompting an unsaved-changes dialog the user never asked for?

Would matching on identity rather than the shared class be safer — for example tagging this drawer mask through `[pt]="{ mask: { 'data-testid': ... } }"` and comparing the element directly?

Related: the nested-PrimeNG-overlay concern raised on #36732 for `onEscape()` looks like the same root cause. Is it worth handling both together, or tracking that separately?

One more traceability question: AC11 on #36631 enumerates only ESC, the close control, and Back as closing triggers. Should click-outside be added as an explicit AC so this behavior stays covered?

core-web/libs/edit-content/src/lib/components/dot-edit-content-side-panel/dot-edit-content-side-panel.component.html
```diff
@@ -19,8 +22,8 @@
     [pt]="{
         root: {
             'data-testid': 'edit-content-side-panel',
-            class: '!shadow-[-12px_0_24px_rgb(0_0_0_/_20%)]',
-            style: { width: $expanded() ? '100%' : '70%', transition: 'width 250ms ease' }
+            class: '!shadow-2xl',
```
oidacra, line 29: Scope: line changed by this PR.

`shadow-2xl` resolves to `0 25px 50px -12px rgb(0 0 0 / 0.25)` — purely vertical, with no horizontal offset. On a full-height right-anchored drawer, does that shadow end up falling outside the viewport rather than separating the panel from the page behind it?

Some history worth surfacing: on #36732 a reviewer proposed exactly `shadow-2xl!` on this line, and the resolution at the time was to keep the arbitrary value with a justifying comment — the rationale being that no dotCMS `$shadow-*` token is left-directional, and a right-anchored drawer needs the shadow cast to the left. That comment is also removed in this PR.

Was the left-cast requirement dropped deliberately (for example a design change), or is this an unintended visual regression? If it is deliberate, would it be worth saying so in the comment, since the surviving text now explains only why the shadow lives on `pt.root.class` and no longer why the value itself was chosen?

core-web/libs/edit-content/src/lib/components/dot-edit-content-side-panel/dot-edit-content-side-panel.component.ts
```diff
@@ -16,7 +16,11 @@ import {
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 
 import { ButtonModule } from 'primeng/button';
-import { DrawerModule } from 'primeng/drawer';
+// `DrawerClasses` carries PrimeNG's own class names for the drawer's parts. The click-outside
```
oidacra: Scope: comment added by this PR.

The claim that a PrimeNG bump "either way cannot silently disable click-to-close" looks stronger than what the import actually guarantees.

In `primeng-drawer.mjs` the mask class is declared twice, independently:

- `DrawerClasses["mask"] = "p-drawer-mask"` — the exported enum this file reads
- `mask: ({ instance }) => ['p-drawer-mask', ...]` — the style map actually applied at runtime via `addClass(this.mask, this.cx('mask'))`

They agree today, but they are two separate literals rather than one shared source. If a future PrimeNG release changed the applied literal without changing the exported enum value, would that compile cleanly and silently disable click-to-close?

The same caveat applies to the `clickOutside` helper JSDoc in the spec, which says the stand-in "cannot drift from the real mask" — it cannot drift from what the handler expects, which is a weaker guarantee, so the test would keep passing in that scenario.

Would narrowing both comments to "protects against the member being renamed or removed" be more accurate?

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form.component.spec.ts
```diff
@@ -458,15 +458,22 @@ describe('DotFormComponent', () => {
...
-            it('should render both open and close sidebar icons, hiding one per state', () => {
-                const openIcon = spectator.query(byTestId('sidebar-open-icon'));
-                const closeIcon = spectator.query(byTestId('sidebar-close-icon'));
-                expect(openIcon).toBeTruthy();
-                expect(closeIcon).toBeTruthy();
-                // One of the two icons must be hidden at any given time
-                const openHidden = openIcon?.classList.contains('hidden');
-                const closeHidden = closeIcon?.classList.contains('hidden');
-                expect(openHidden).not.toBe(closeHidden);
+            it('should render a single dock_to_left sidebar icon', () => {
```
oidacra, line 461: Scope: test added by this PR.

This test pins the glyph carefully — exact ligature text, the `material-symbols-outlined` class, `aria-hidden`, and the absence of the old markup. Since the icon is now decorative, the `aria-label` on the button is the only remaining open/closed cue for assistive tech.

Would it be worth adding an assertion that the rendered button carries the expected accessible name and that it flips between `edit.content.sidebar.close` and `edit.content.sidebar.open` as `showSidebar` changes?

Asking because a test at that level would also catch the binding question I raised on the template side.

core-web/libs/edit-content/src/lib/services/dot-side-panel-nav.service.ts
```diff
@@ -39,7 +39,7 @@ function clearPrevNavCollapsed(): void {
 }
 
 /** Collapse the navs for the side panel when the viewport is narrower than this (px). */
-const COLLAPSE_MAX_WIDTH = 1800;
+const COLLAPSE_MAX_WIDTH = 2560;
```
oidacra, line 49: Scope: value changed by this PR.

`shouldCollapse()` is `window.innerWidth < COLLAPSE_MAX_WIDTH`, so at 2560 the nav now collapses on 1440, 1920, and effectively every 2560 monitor too, since the browser viewport is always somewhat narrower than the physical display.

Is "collapse for everyone except ultra-wide and 4K" the intended behavior? The doc comment just above still frames this as collapsing when the viewport is "narrow enough", which reads as the exception rather than the common case — would it be worth updating that wording to match?

The exclusive comparison itself checks out: `2560 < 2560` is false, so pinning `setWideViewport()` to 2560 in the spec correctly treats the boundary as wide.

core-web/libs/edit-content/src/lib/components/dot-edit-content-side-panel/dot-edit-content-side-panel.component.html
```diff
@@ -19,8 +22,8 @@
...
+            style: { width: $expanded() ? '100%' : '80%', transition: 'width 250ms ease' }
```
oidacra, line 30: Scope: line changed by this PR.

AC2 on #36631 reads "~70% of the main viewport width, measured excluding the left nav bar". Because the drawer is `appendTo="body"`, this percentage is measured against the full viewport — so on a 1920px viewport with a 270px nav, the AC target works out to roughly 1155px while this renders roughly 1536px.

The nav collapse threshold bump narrows that gap by removing the nav from the equation on most widths, but it does not change the measurement basis (the full-viewport basis is pre-existing from #36732, not introduced here).

Given the PR is explicitly the feedback pass and 80% looks like a deliberate product decision, would it make sense to amend AC2 on #36631 to match, so the issue and the implementation do not drift apart? Happy to be wrong if 80% of the full viewport was already the agreed reading.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-file-field/components/dot-file-field-preview/dot-file-field-preview.component.scss
```diff
@@ -123,7 +123,11 @@
     position: absolute;
     right: 0.5rem;
     top: 0.5rem;
-    z-index: 100;
+    // Kept in the same 1-2 range as the other actions: these only need to sit above the
+    // thumbnail/code preview. A higher value let them paint over the form's sticky tab list
+    // (z-index: 10, carries the status chip + permissions/references), which was visible while
+    // scrolling on narrow containers — the only layout where these filled variants render.
+    z-index: 2;
```
oidacra, line 130: Scope: line changed by this PR.

The rationale checks out — I confirmed the sibling values are 1 for the wide variant and 2 for remove, and that the sticky `.p-tablist` in the form component is `z-index: 10`.

One piece of history worth flagging: this same value was lowered to 1 on a branch in June (`ab52b9bc7c`), and then the merge of #36172 explicitly restored 100 for the `--responsive` variant while leaving its siblings at 1. So 100 survived a specific attempt to change it, which suggests it may have been holding these buttons above something else in the narrow layout.

Since this variant only renders in narrow containers, would it be worth a targeted pass over that layout for any other overlap the 100 was covering, beyond the tab bar case this fixes?

core-web/libs/edit-content/src/lib/components/dot-edit-content-side-panel/dot-edit-content-side-panel.component.spec.ts
```diff
@@ -206,6 +206,95 @@ describe('DotEditContentSidePanelComponent', () => {
...
+        expect(confirmClose).not.toHaveBeenCalled();
+        expect(closedSpy).not.toHaveBeenCalled();
+    });
+
+    it('should NOT close on a click inside the panel', () => {
+        spectator.setInput('data', EDIT_DATA);
+        spectator.detectChanges();
+
+        const layout = spectator.query(DotEditContentLayoutComponent);
+        const confirmClose = jest.spyOn(layout, 'confirmClose');
```
oidacra, line 323: Scope: test added by this PR.

This dispatches through an optional chain and then asserts only that nothing happened. If `side-panel-title` ever stops matching — a markup rename, a teardown timing change — the click is silently never dispatched and both negative assertions still pass, so the test would go green without exercising anything.

The element does exist today, so this is not currently a false positive. Would adding a positive assertion before the dispatch (or dropping the `?.` so a missing element throws) be worth it to keep it self-checking?

core-web/libs/edit-content/src/lib/components/dot-edit-content-sidebar/dot-edit-content-sidebar.component.html
```diff
@@ -128,6 +128,20 @@
                             (actionFired)="workflowActionFired.emit($event)" />
                     }
 
+                    <!-- Content details section -->
```
oidacra, line 131: Scope: section moved by this PR.

The reorder itself is consistent and correctly tested — I verified the border-suppressing class moved off Details and onto Workflow as the new last section, and the spec pins the new DOM order.

Purely for traceability against #36631: none of the ACs cover section ordering in the Actions tab, so was this reorder explicitly requested in the review or QA feedback? Asking because persistence keys are preserved and users carry over their saved expand and collapse state, so a reordering that was not asked for would be a silent UX change for existing users.

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form.component.html
```diff
@@ -132,30 +132,28 @@
...
+                    (showSidebar ? 'edit.content.sidebar.close' : 'edit.content.sidebar.open') | dm
+                "
```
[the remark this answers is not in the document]
adrianjm-dotCMS: Fixed. Switched to PrimeNG's `[ariaLabel]` input so the name lands on the inner focusable `<button>`, and the spec now asserts that label on the descendant native button in both open/closed states.

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form.component.html
adrianjm-dotCMS: Confirmed — and worse than an omission: this PR introduced it. The previous markup had two projected `` elements whose alt text supplied the accessible name; replacing them with an `aria-hidden` glyph plus a host-level `[attr.aria-label]` left the native `<button>` with no accessible name at all.

Verified in PrimeNG 21.1.3: `Button` declares a public `ariaLabel` input and binds `[attr.aria-label]="ariaLabel || buttonProps?.ariaLabel"` on the inner `<button>`, with no host-attribute forwarding. Switched to `[ariaLabel]` here and applied the same fix to the two side-panel header buttons, which had the same defect.

On traceability: the `dock_to_left` swap was explicitly requested in the feedback pass, not folded in along the way.

core-web/libs/edit-content/src/lib/components/dot-edit-content-side-panel/dot-edit-content-side-panel.component.ts
adrianjm-dotCMS: You are right, the comment overclaimed — the applied literal and the exported enum are two independent declarations, so a rename of only the former would compile and break silently.

Moot now: the fix on the handler below replaces the class match with an identity check, so the constant and its comment are gone.

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form.component.spec.ts
adrianjm-dotCMS: Added. The spec now asserts the accessible name on the rendered native button and that it flips between the two keys as `showSidebar` changes.

core-web/libs/edit-content/src/lib/components/dot-edit-content-side-panel/dot-edit-content-side-panel.component.ts
adrianjm-dotCMS: Confirmed. The block editor sidebar is at `libs/portlets/edit-ema/portlet/src/lib/components/dot-block-editor-sidebar/` and does render a `<p-drawer>` with no `[modal]` binding, so it defaults to `true` and creates its own mask, as a sibling of this panel in the UVE template.

One correction on the suggested fix: `pt.mask` will not work. `ptm('mask')` appears nowhere in `primeng-drawer.mjs` — the mask is built imperatively and only ever receives `style`, `data-p` and the class, so the tag would never reach it.

Used identity instead. `Drawer.mask` is public (`Nullable<HTMLDivElement>`), so `viewChild(Drawer)` plus `event.target !== drawer.mask` removes the shared-class coupling entirely. Added a regression test for a foreign drawer mask.

Also took `onEscape()` in the same pass, since you were right that it is the same root cause — and unlike the mask case, that one *is* reachable today (open the image editor from a file field inside the panel, press ESC, both close). Guarded it with `ZIndexUtils`: every PrimeNG overlay registers in that shared stack on open, so comparing the top of the stack against this drawer's z-index answers "is something above me?" without matching class names. Preferred that over checking `event.target`, which breaks ESC when focus falls back to `body`.

On the AC: click-outside is now an explicit acceptance criterion on #36631, and the unsaved-changes AC lists it alongside ESC/close/Back.

core-web/libs/edit-content/src/lib/components/dot-edit-content-side-panel/dot-edit-content-side-panel.component.spec.ts
adrianjm-dotCMS: Fixed. The element is now asserted before the click is dispatched, so a markup rename fails the test instead of silently skipping the dispatch and leaving the negative assertions green.

core-web/libs/edit-content/src/lib/services/dot-side-panel-nav.service.ts
adrianjm-dotCMS: 2560 is intended — the tradeoff was weighed explicitly and, as you note, at 80% panel width the leftover strip only clears the 270px expanded nav by a useful margin on very wide screens.

You are right that the doc comment framed it backwards. Updated in three places (the constant, `shouldCollapse()`, and the class docstring) to state that collapsing is the common case, plus a note that `innerWidth` is CSS pixels so scaled or non-maximized 2560 displays land here too.

core-web/libs/edit-content/src/lib/components/dot-edit-content-side-panel/dot-edit-content-side-panel.component.html
adrianjm-dotCMS: Amended AC2 on #36631. It now reads "80% of the viewport width" and drops "measured excluding the left nav bar" — you are right that the full-viewport basis was always the real behavior, so keeping that clause would have left the AC diverging even after the number changed. Updated the Description and Additional Context to match. 80% is a deliberate product decision from this feedback pass.

core-web/libs/edit-content/src/lib/components/dot-edit-content-side-panel/dot-edit-content-side-panel.component.html
adrianjm-dotCMS: Deliberate — `shadow-2xl` is a design decision from this feedback pass, which supersedes the resolution reached on #36732.

core-web/libs/edit-content/src/lib/components/dot-edit-content-sidebar/dot-edit-content-sidebar.component.html
adrianjm-dotCMS: Explicitly requested in the feedback pass — Details / Locales / Workflow was the ordering asked for, not an incidental change.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-file-field/components/dot-file-field-preview/dot-file-field-preview.component.scss
adrianjm-dotCMS: The history does not hold up. `ab52b9bc7c` never reached `main`, and it is not an ancestor of `e3aa226313` (which is not a merge commit either) — they were parallel branches. #36172 never saw the change to `1`; it simply carried `100` forward from its own base. So `100` did not survive an attempt to change it, and there is no evidence it was holding these buttons above anything: it has been `100` on `main` continuously since #30135, unexamined.

That said, the narrow layout was exercised manually after the change and nothing else overlaps those buttons.

Issue #36971, opened 2026-08-08
N+1: DefaultTransformStrategy loads the owner user once per row of every REST/GraphQL response


wezell:
## Problem Statement

`DefaultTransformStrategy` resolves the **owner user of every contentlet individually**, producing one `SELECT` per row of every REST and GraphQL response. This is a classic N+1: a response returning 100 contentlets issues up to 100 additional user queries, each a separate Postgres round trip on the request thread.

Observed call chain:

```
DotContentletTransformerImpl.transform
  └─ AbstractTransformStrategy.apply
       └─ DefaultTransformStrategy.addCommonProperties
            └─ UserAPIImpl.loadUserById
                 └─ DotConnect.executeQuery → Postgres  (thread parked in sun.nio.ch.Net.poll)
```

The transform pipeline runs for **every** contentlet in **every** REST and GraphQL response, so the cost scales linearly with page size — exactly the dimension a client controls via `limit` / `depth` / GraphQL selection sets.

**Who it impacts:** any tenant serving content over `/api/content`, `/api/es/search`, `/api/v1/page`, or GraphQL. Worst on API-first and headless tenants with large result sets.

### Evidence

Glowroot main-thread profile, `caliber` (GraphQL-heavy tenant — GraphQL 54% of transactions, `/api/es/search` 12%, `/api/content` 10%), 7-day window ending 2026-08-07, 893 sampled stacks:

```
1.1%  AbstractTransformStrategy.apply
 1.0%  UserAPIImpl.loadUserById
  0.9%  DotConnect.executeQuery
   0.8%  sun.nio.ch.Net.poll   RUNNABLE
```

`AbstractTransformStrategy.apply` appears ~40 times across that profile as the frame immediately above per-row work. The same `loadUserById` pattern is visible in the `k8s.dairyqueen` profile under `ContentletDataFetcher` → `DotContentletTransformerImpl.lambda$transform$1`.

Note this is a *sampled* profile: it establishes that the call is hot and per-row, not an exact query count. Confirm the multiplier with SQL logging (see repro step 3).

## Steps to Reproduce

1. Create ~100 contentlets of any type, owned by a mix of users.
2. Enable statement logging on the Postgres connection, or attach a profiler to a node under load.
3. Request them in one call — e.g. `GET /api/content/query/+contentType:MyType/limit/100`, or the equivalent GraphQL query selecting a field that triggers the default transform.
4. Count `SELECT` statements issued against the user table for that single request.

[...]

Issue #36970, opened 2026-08-08
Config and App-secret reads round-trip Postgres on every request instead of using cache


wezell:
## Problem Statement

Every HTTP request performs several **Postgres round trips** to read App secrets and configuration values that should be served from cache. The reads happen on the request thread, blocked on a socket, before any application work begins.

Observed call chain:

```
SimpleWebInterceptorDelegateImpl.intercept
  └─ AppsAPIImpl.getSecrets / hasEnvBackedSecrets / hasGlobalEnvBackedSecrets
       └─ Config.getSystemTableValue
            └─ SystemTableConfigSource.getValue
                 └─ SystemTableFactoryImpl.find
                      └─ DotConnect → Postgres  (thread parked in sun.nio.ch.Net.poll)
```

Several web interceptors each resolve secrets independently on the same request — `AnalyticsTrackWebInterceptor.isAllowed`, `PreRenderSEOWebAPIImpl.prerenderIfEligible`, `DefaultAutoLoginWebInterceptor.intercept` — so the chain repeats several times per request for the same keys.

**Who it impacts:** every request on every tenant. This is fixed per-request overhead, not proportional to what the client asked for, so it is pure waste rather than a cost that scales with work done.

### Evidence

Glowroot main-thread profiles, 7-day windows ending 2026-08-07, production clusters:

| Tenant | Sampled stacks | Distinct occurrences of this chain | Approx. share of samples |
|---|---|---|---|
| `k8s.client1` | 12,126 | ~20 | ~3–4% |
| `client2` | 893 | ~25 | ~3% |
| (Velocity tenant) | 17,706 | ~15 | ~2% |

Representative frames from the `caliber` profile:

```
2.8%  io.vavr.control.Try.of
 1.7%  AppsAPIImpl.lambda$filterSitesForAppKey$0
  1.6%  AppsAPIImpl.hasEnvBackedSecrets
   0.9%  Config.getStringProperty
    0.8%  SystemTableConfigSource.getValue
     0.4%  SystemTableImpl.get
```

and from `k8s.dairyqueen`:

```
3.1%  AppsAPIImpl.getSecrets
 1.0%  AppsAPIImpl.resolveSecrets
  0.7%  Config.getSystemTableValue
   0.6%  SystemTableConfigSource.getValue
    0.3%  SystemTableFactoryImpl.find
```

Note this is a *sampled* profile: it shows time share, not exact query counts. The query count per request should be confirmed with SQL logging (see repro step 3).

[...]

PR #36978, opened 2026-08-10
fix(cost): close RequestCost instrumentation gaps and re-base the price table

### LINKED ISSUE #36977: Fix RequestCost instrumentation gaps and re-base the price table on resource-time
## Description

`@RequestCost` scores the "heaviness" of a request, token-style, so usage can be metered and rate-limited. An audit of **where the annotation is actually placed** against production behaviour — Glowroot main-thread profiles from six tenant instances, 7-day windows — found the instrumentation concentrated on the Contentlet API while the traffic was somewhere else entirely, and one class of cost being discarded outright.

### Six instances, six different cost centres

| Shape | Where the time actually went | Metered before? |
|---|---|---|
| write-heavy API | `addContentToIndex` — ~10% of all samples, blocking | **no** |
| GraphQL 60% / VTL 31% | prerender park 36%, `NavTool.getNav` 11% | **no** |
| Velocity + remote API | `CircuitBreakerUrl` outbound HTTP ~50% | yes, priced 4 |
| GraphQL 54% | GraphQL execute, transform, relationship N+1 | **no** |
| traditional Velocity | filter chain, render, binaries | partly |

A GraphQL-only customer could saturate a node and accrue almost nothing.

### Background work was billing zero

`incrementCost` returned early when no `HttpServletRequest` was on the thread — **before** touching the totals that feed the collector. Site-search reindexing, scheduled publishing, remote/push publishing and content indexing already flow through priced methods (`CONTENT_INDEX`, `CONTENT_CHECKIN`, `HTTP_FETCH`); all of it was priced correctly and then thrown away at the last step.

## Acceptance Criteria

- [ ] Every search result path is metered — GraphQL, `/api/content`, `/api/es/search`, page render all bulk-load through `ESContentFactoryImpl.findContentlets`, which bypassed `find()` and was uncosted
- [ ] Cost scales with contentlets returned, and cached content is materially cheaper than content read from the database
- [ ] Content pricing does not vary with implementation details the customer cannot influence (batch size, statement count, query plan)
- [ ] Background work reaches the collector instead of being discarded, reported separately from request cost
- [ ] `/api/es/search`, `/api/vtl/*` and `#dotParse` no longer bypass their existing charges
- [ ] Price table reflects relative resource-time; image transforms and Tika metadata generation are not priced below a cache read
- [ ] Reported token values stay in their historical range and remain integral
- [ ] Rate-limit defaults scale with the price table
- [ ] No annotation double-charges via a ByteBuddy self-invocation chain
- [ ] Existing unit tests updated for the snapshot shape change

## Additional Context

**Pricing model.** Two coherent options — meter primitives (accurate, but the bill moves with our cache hit rate and batch sizes) or meter customer-visible operations (stable, optimisable). Chosen: the latter, with the line drawn at **what the customer can influence**:

| | Whose? | Priced? |
|---|---|---|
| Did this need the database at all? | theirs — cacheable containers/pages, TTLs, query shape | **yes** |
| How many statements, what batch size, which plan? | ours | **no** |

Content is priced `CONTENT_FROM_CACHE`(1) per contentlet asked for, plus `CONTENT_FROM_DB`(10) as a surcharge on the ones read from Postgres — so 1 warm, 11 cold. The same cheap-hit/expensive-miss shape already existed for ES queries and file metadata and is now consistent across the model.

**Collector change required.** The pushed `RequestCostSnapshot` gains `windowJobTokens` and `lifetimeJobTokens` (doubles). They are reported *separately* — `windowTokens` stays request-only, total consumption is the sum. Unknown fields are dropped silently by the ingestor, so this is safe to deploy ahead of the ingestor change, but **background cost is not billed until those two columns exist**.

**Expect a step change** in reported tokens at deploy: prices were re-based and the denominator moved 1 → 10 to compensate. Roughly range-preserving for requests, but per-endpoint series shift by different amounts.

**Not validated against a running instance.** Prices are order-of-magnitude estimates from sampled time-share; placements were verified by reading call chains. The calibration loop — run a week, compare per-endpoint mean response time against mean cost, investigate outlier ratios — is what turns these into measured numbers.

**Known gaps, deliberately out of scope:**
- Embedding generation / vector DB indexing is not priced at all — no constant, nothing annotated. Likely the largest remaining gap given inference cost.
- Job cost is one undifferentiated bucket, not split by job type. Needs a job-name context set by background entry points.
- The telemetry publisher self-charges `HTTP_FETCH` for its own push (~10 tokens/window).
- Hibernate is invisible to any DB-level metering — `PermissionBitAPIImpl.getPermissionsByRole` goes through `net.sf.hibernate.loader.Loader`, never `DotConnect`.

Full analysis and evidence ship with the PR as `docs/requestcost-session-summary.md` and `docs/requestcost-placement-analysis.md`.

Related defects found during the audit: #36970, #36971.

wezell:
Fixes #36977

## Why

`@RequestCost` scores request heaviness so usage can be metered and rate-limited. Auditing **where the annotation actually sits** against Glowroot main-thread profiles from six production instances found the instrumentation concentrated on the Contentlet API while the traffic was somewhere else — and one whole class of cost being discarded.

| Instance shape | Where the time actually went | Metered before? |
|---|---|---|
| write-heavy API | `addContentToIndex` — ~10% of all samples, blocking | **no** |
| GraphQL 60% / VTL 31% | prerender park 36%, `NavTool.getNav` 11% | **no** |
| Velocity + remote API | `CircuitBreakerUrl` outbound HTTP ~50% | yes, priced 4 |
| GraphQL 54% | GraphQL execute, transform, relationship N+1 | **no** |
| traditional Velocity | filter chain, render, binaries | partly |

A GraphQL-only customer could saturate a node and accrue almost nothing.

## What changed

**Metering gaps closed**

| Where | Why it mattered |
|---|---|
| `ESContentFactoryImpl.findContentlets(List)` | The bulk loader behind *every* search result — GraphQL, `/api/content`, `/api/es/search`, page render. It bypasses `find()`, so nothing metered it. Now per contentlet: base fee, plus a surcharge on rows read from the DB |
| `incrementCost` background path | Returned **before** the totals when no request was on the thread. Reindexing, scheduled publishing, remote publishing and content indexing were priced correctly and then thrown away |
| `enterprise.priv.ESSearchAPIImpl.esSearchRaw` | `/api/es/search` bypassed the existing `ES_QUERY` charge entirely |
| `VTLResource.evalVelocity` | `/api/vtl/*` called the engine directly, bypassing the annotated merge |
| `DotDirective.render` | Velocity **rendering** was not metered anywhere on the page path — neither `#dotParse` / `#parseContainer` nor the top-level page merge, since all three mode handlers call `Template.merge()` directly rather than an annotated `VelocityUtil` method. (What each container *fetched* was metered, via `find()` and `cachedIndexSearch`; the work of assembling the page was not.) Charged past the `getFromCache()` short-circuit, so cacheable containers stay cheap |
| `VelocityLiveMode` / `EditMode` / `PreviewMode` | The page's **own** template merge was unmetered too. Charged at the merge site, not on `serve()` — LiveMode on `writePage()`, which the two page-cache hit paths return before reaching, so a cached page pays no merge |
| `PreRenderSEOWebAPIImpl` | Own `HttpClient`, so `CircuitBreakerUrl`'s charge never fired |
| `ContentletIndexAPIImpl`, `NavTool`, `DotGraphQLHttpServlet` | New `CONTENT_INDEX`, `NAV_BUILD`, `GRAPHQL_QUERY` |
| `ESContentletAPIImpl` terminals | `CONTENT_CHECKIN` / `CHECKOUT` / `COPY` / `MOVE` existed in the enum with nothing charging them |

**Pricing model.** Meter customer-visible operations, with the line drawn at what the customer can influence:

| | Whose? | Priced? |
|---|---|---|
| Did this need the database at all? | theirs — cacheable containers/pages, TTLs, query shape | **yes** |
| How many statements, what batch size, which plan? | ours | **no** |

Content is `CONTENT_FROM_CACHE`(1) per contentlet asked for plus `CONTENT_FROM_DB`(10) on the ones read from Postgres — 1 warm, 11 cold, an 11× incentive to cache. The surcharge is per missed *row*, not per SQL statement; the 200-row batching is our detail. There is deliberately no generic `DB_QUERY`.

**Price table re-based on resource-time**, then the DB tier compressed — what is metered is capacity consumed *on this node*, not wall time. A query parks the thread and burns cycles on Postgres. At `DB_QUERY` 25 the table said DB was 50% of a page render; the profile said template rendering dominated. At 10 it reads merges 59% / DB 31% / ES 10%, which matches.

`IMAGE_FILTER_TRANSFORM` was **2** and `FILE_METADATA_GENERATE` was **3** — decoding an image or running Tika burns a core and a large buffer. Both now 50.

**Reporting.** `REQUEST_COST_DENOMINATOR` 1 → 10 keeps reported tokens in their historical range. Header and snapshot totals are rounded — both have only ever carried whole numbers, and a collector parsing them as ints would break on a fraction. Per-request averages stay fractional, as always. Rate-limit defaults scaled with the table (still off by default).

## Collector change required

`RequestCostSnapshot` gains `windowJobTokens` and `lifetimeJobTokens` (doubles), reported **separately** — `windowTokens` stays request-only, total consumption is the sum. Unknown fields are dropped silently by the ingestor, so this is safe to deploy first, **but background cost is not billed until those two columns exist.**

Expect a step change in reported tokens at deploy: prices re-based, denominator moved. Roughly range-preserving for requests, but per-endpoint series shift by different amounts.

## Testing

`RequestCostSnapshotTest` and `RequestCostPublisherTest` updated for the two new fields, including the "exactly N fields on the wire" guard.

Verified with `javac` against the built classpath — error count identical to the baseline. The full Maven build fails in `dotcms-core-web`'s nx step, pre-existing and unrelated.

**Not validated against a running instance.** Prices are order-of-magnitude estimates from sampled time-share; placements were verified by reading call chains, not by watching a request. The calibration loop — run a week, compare per-endpoint mean response time against mean cost, investigate outlier ratios — is what turns these into measured numbers.

## Known gaps, deliberately out of scope

- **Embedding generation / vector DB indexing is not priced at all** — no constant, nothing annotated. Likely the largest remaining gap given inference cost.
- Job cost is one undifferentiated bucket, not split by job type.
- The telemetry publisher self-charges `HTTP_FETCH` for its own push (~10 tokens/window).
- Hibernate is invisible to DB-level metering — `PermissionBitAPIImpl.getPermissionsByRole` goes through `net.sf.hibernate.loader.Loader`, never `DotConnect`.

## Reviewer notes

Three traps worth knowing if you touch this:

1. **The enum is not an index of what is metered.** `ES_QUERY`, `ES_COUNT`, `WORKFLOW_ACTION_RUN` and `IMAGE_FILTER_TRANSFORM` are charged with direct `incrementCost(...)` calls, invisible to a `@RequestCost` grep. Grep both forms.
2. **ByteBuddy weaves bytecode, so self-invocation fires the advice** — unlike CDI proxies. `checkin` has 12 overloads chaining into each other; annotating two in one chain double-charges. Every placement here targets a verified terminal.
3. **Annotations are not inherited.** A `@RequestCost` on an interface default does nothing for an override — this had already left `CONTENT_FROM_DB` dead on the common single-find path.

Docs ship with the change: `docs/requestcost-session-summary.md` (narrative) and `docs/requestcost-placement-analysis.md` (evidence and backlog). Tenant identities are anonymised.

Defects found during the audit, filed separately: #36970, #36971.

This PR fixes: #36977


docs/requestcost-placement-analysis.md
sfreudenthaler, line 1: checked and glad to see no specific details here that leak customer info.  Fine to check-in but probably limited long term value.

docs/requestcost-session-summary.md
sfreudenthaler, line 1: @wezell suggest not checking this in.  value today but not long term I think.  maybe attach as a comment to the PR?

dotCMS/src/main/java/com/dotcms/cost/LeakyTokenBucketImpl.java
```diff
@@ -66,8 +66,15 @@ public class LeakyTokenBucketImpl implements LeakyTokenBucket {
...
                 Config.getBooleanProperty("RATE_LIMIT_ENABLED", false),
-                Config.getLongProperty("RATE_LIMIT_REFILL_PER_SECOND", 500),
-                Config.getLongProperty("RATE_LIMIT_MAX_BUCKET_SIZE", 10000)
+                // Scaled with the Price table when it was re-based on resource-time
+                // (see RequestPrices.Price): one DB round trip is 10 units, one remote
+                // HTTP call is 100, so the old 500/10000 defaults would now throttle a
+                // handful of requests per second. Ratio to the old defaults is unchanged.
+                // NOTE: the bucket drains in RAW Price units, not the denominated ones
+                // reported in the x-dotrequest-cost header - a limit set here is on the
+                // Price scale (remote HTTP = 100), not the reported-token scale.
```
sfreudenthaler: no need for comment

```suggestion
```

sfreudenthaler (approved):
@wezell i’m good with the changes.  I’d like to see the comments ripped out or much less verbose in most cases, but not blocking a merge on it

docs/requestcost-placement-analysis.md
wezell: done

Issue #36720, opened 2026-07-23
Users portlet: API Tokens tab


hmoreras:
## Goal

Deliver the API Tokens tab of the Edit User dialog.

## Scope

- Tokens table: Token ID (truncated), Label, Issued, Expires, Requested By, Network CIDR chip, per-row `Revoke`
- `Show inactive` filter
- `Request New Token` action opens a sub-dialog:
  - Label
  - Expires Date (default ~3 years)
  - Allow Network CIDR (default `0.0.0.0/0`)
- On successful creation, show the raw token once with a copy affordance
- Revoke prompts confirmation

## Acceptance criteria

- [x] Tokens list loads with correct fields
- [x] `Show inactive` toggles revoked/expired token visibility
- [x] New token creation returns and displays the raw token exactly once
- [x] Copy affordance copies the raw token to clipboard
- [x] Revoke removes the token from the active list
- [x] Unit + integration tests via Spectator

## QA

- [ ] Manual: create token, authenticate against `/api/v1/users` with it, revoke, verify 401
- [ ] Manual: CIDR restriction enforced (request from disallowed range returns 401)
- [ ] Manual: expired-only view via `Show inactive`

## Design references

Interactive prototype: https://claude.ai/design/p/324a0e09-34dc-4925-a7fa-24f39fdafa54?via=share&file=Users.html

hmoreras:
Work done, final QA / feedback, will be done in the follow-up task that includes an endpoint update from BE work, https://github.com/dotCMS/core/issues/37298

Issue #36719, opened 2026-07-23
Users portlet: Permissions tab (iframe embed)


hmoreras:
## Goal

Render the existing permissions UI inside the Users dialog via an iframe, as the interim solution while dedicated Angular designs are prepared.

## Scope

- Iframe hosted inside the Permissions tab panel
- URL + query params confirmed by the Spike
- Loading state; graceful failure if the iframe endpoint is unreachable
- Height sized to fit the dialog content area

## Not in scope

- Reimplementing the permissions UI in Angular

## Acceptance criteria

- [ ] Iframe loads the correct permissions view for the selected user
- [ ] Dialog chrome (tabs, header, footer) remains functional while the iframe is loaded
- [ ] Save Changes footer does not intercept iframe form submissions
- [ ] a11y: iframe has a descriptive `title` attribute

## QA

- [ ] Manual: assign permissions via the iframe and verify they persist
- [ ] Manual: verify no double-scrollbar or clipping on standard viewports

## Design references

Interactive prototype (list + other tabs): https://claude.ai/design/p/324a0e09-34dc-4925-a7fa-24f39fdafa54?via=share&file=Users.html — the Permissions tab has no dedicated Angular design; this task ships an iframe embed of the current UI.

hmoreras:
Work done, final QA / feedback, will be done in the follow-up task that includes an endpoint update from BE work, https://github.com/dotCMS/core/issues/37298

Issue #36718, opened 2026-07-23
Users portlet: Roles tab (dual-list)


hmoreras:
## Goal

Deliver the Roles tab of the Edit User dialog — dual-list with hierarchical role tree on the Available side and flat granted list.

## Scope

- Dual panels: `AVAILABLE` (hierarchical tree) and `GRANTED` (flat list)
- Independent filter input on each panel
- Transfer buttons to move roles between panels
- Tree nodes: expandable folders + selectable role rows
- Granted list shows role name + origin (e.g. `System`, custom role folder)
- Save persists granted role set atomically

## Acceptance criteria

- [ ] Tree renders the full role hierarchy from the endpoint
- [ ] Filter narrows visible nodes without collapsing expanded state
- [ ] Multi-select + transfer works in both directions
- [ ] Granted list reflects saved state after reload
- [ ] Unit + integration tests via Spectator

## QA

- [ ] Manual: assign / unassign roles and verify user's effective access
- [ ] Manual: large role trees (100+ nodes) remain responsive

## Design references

Interactive prototype: https://claude.ai/design/p/324a0e09-34dc-4925-a7fa-24f39fdafa54?via=share&file=Users.html

hmoreras:
Work done, final QA / feedback, will be done in the follow-up task that includes an endpoint update from BE work, https://github.com/dotCMS/core/issues/37298

PR #36990, opened 2026-08-10
feat(users): Profile tab, dialog shell & list CRUD wiring (#36717)

### LINKED ISSUE #36717: Users portlet: Profile tab (Create and Edit dialog)
## Goal

Deliver the Profile tab of the Create / Edit User dialog, plus the dialog shell that hosts all four tabs.

### Dialog shell

- 4 tabs: Profile, Roles, Permissions, API Tokens
- Header, footer, Save Changes / Create User CTA
- Sibling tasks deliver Roles, Permissions, and API Tokens tab content

### Profile tab

- Header: avatar + name + Active status chip; `NU` (initials) placeholder in create mode
- Metadata card: User ID, Last login (timestamp + IP), Failed logins — hidden in create mode
- **Account** section:
  - First name*, Last name*
  - Email* (format validated)
  - Password / Confirm password with `Generate secure password` action
  - "Leave blank to keep current password" helper text in edit mode
  - Active toggle
- **Additional information** section: Prefix, Suffix, Title, Company, Website
- **Access** toggles: CMS Administrator, Back-end User, Front-end User, Can Login to Admin UI, Show Getting Started
- **Delete User** destructive action — edit mode only, requires typed confirmation

## Acceptance criteria

- [ ] Required fields validated; email format enforced
- [ ] Password + Confirm password match required when set
- [ ] "Leave blank to keep current password" behavior works in edit mode
- [ ] `Generate secure password` populates both fields
- [ ] Delete User visible only in edit mode; requires confirmation
- [ ] Access toggles map to endpoints identified by the Spike
- [ ] All fields persist and reload correctly after save
- [ ] Unit + integration tests via Spectator

## QA

- [ ] Manual: create, edit, delete flow end-to-end
- [ ] Manual: verify Active toggle prevents login for the affected user
- [ ] Manual: verify each Access toggle produces expected effective access

## Design references

AP2300:
Closes #36717

## Summary

- Ships the Users portlet Create/Edit dialog **shell** (4-tab strip, header, footer) with the **Profile tab** fully wired against the backend and inline "Coming soon" placeholders for the Roles, Permissions, and API Tokens tabs (delivered by sibling issues #36718, #36719, #36720)
- Adds the missing user CRUD to the list flow: `GET /api/v1/users/{userId}` hydrates the edit dialog, `POST /api/v1/users` creates, `PUT /api/v1/users` updates (roles field intentionally omitted so backend preserves membership — see `UserResource#processRoles`)
- Delete flow — both from the dialog footer and the list bulk toolbar — now includes a **required replacement-user picker** (server-backed autocomplete over `/api/v1/users/filter`, excludes the users being deleted). Previously deletes silently reassigned ownership to the logged-in admin.
- Fixes `/users` routing so the beta portlet actually lands on this UI instead of bouncing to Getting Started (`PORTLET_ID_RESOLVERS.users` → `users-beta`).

## Notable non-obvious calls

- **Access section is display-only** in this PR: `CMS Admin`, `Back-end`, `Front-end`, `Can Login` map to backend **roles** (not UserForm fields), and safely editing them requires knowing the user's full role list. That's a follow-up (DWR wrapper vs. small REST endpoint decision).
- **Password stays optional on update.** Empty password field means "keep current" — matches how `UserForm` handles password already; no client-side hack needed.
- **`Show Getting Started` toggle** has no backend field; kept in the UI as informational, not persisted. To be revisited if the value is needed elsewhere.

## Test plan

- [x] Create user via the "New" button — required fields validated, password ≥ 6 chars, confirm-match enforced; on Save a new row appears in the list
- [x] Edit user via row click — dialog hydrates from list-item data instantly, then background-fetches `additionalInfo` (prefix/suffix/title/company/website); on Save, changes persist and the list reloads
- [x] Toggle Active on/off via the Profile field — persists on save
- [x] Delete user via the dialog footer — replacement picker required, cannot pick the same user; typed-email confirmation required; on confirm, list reloads
- [x] Bulk-delete N users from the toolbar — replacement picker required, cannot pick any of the selected; on confirm, list reloads
- [x] Placeholder tabs render "Coming soon" and don't interfere with Profile Save
- [x] `pnpm nx test portlets-dot-users-portlet` — 56 passing
- [x] `pnpm nx test data-access --testPathPatterns=dot-router` — 46 passing
- [x] `pnpm nx lint portlets-dot-users-portlet` — clean

## Follow-ups (each on its own branch/PR already prepared)

- #36718 — Roles tab (dual-list shuttle)
- #36719 — Permissions tab (iframe embed + JSP bridge)
- #36720 — API Tokens tab (table + request sub-dialog)

This PR fixes: #36717


core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.ts
```diff
@@ -1,26 +1,584 @@
...
+        const password = (account.password ?? '').trim();
+        if (password) {
+            payload.password = password;
+        }
+
+        if (Object.keys(additionalInfo).length > 0) {
+            payload.additionalInfo = additionalInfo;
+        }
+
+        payload.roles = this.mergeRoleKeysForSave(access);
```
hmoreras, line 504: We're sending `roles` on every save and the backend replaces the whole list instead of merging — it removes every direct role, then re-adds only what we sent. So anything we don't send gets dropped: roles with no `roleKey` (filtered out at L440), and inherited roles come back as direct grants. If any role in that set has `editUsers = false`, `addRoleToUser` throws and the whole PUT fails.

The description says we deliberately omit `roles` so the backend preserves membership — that's the right call, we just need to actually not send it here.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.ts
```diff
@@ -1,26 +1,584 @@
...
+            .pipe(
+                take(1),
+                finalize(() => this.isLoading.set(false)),
+                takeUntilDestroyed(this.destroyRef)
+            )
+            .subscribe({
+                next: ({ user, userRoles, gettingStarted }) => {
+                    const roleKeys = userRoles
+                        .map((role) => role.roleKey)
+                        .filter((key): key is string => !!key);
```
hmoreras: `roleKey` is optional and roles created from the Roles portlet often don't have one. Those get dropped here, never reach the payload, and the backend then deletes them (see L535). A user with custom roles loses all of them on a plain save.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.ts
```diff
@@ -1,26 +1,584 @@
...
+        if (this.user?.userId) {
+            payload.userId = this.user.userId;
+        }
+
+        const password = (account.password ?? '').trim();
+        if (password) {
+            payload.password = password;
+        }
+
+        if (Object.keys(additionalInfo).length > 0) {
```
hmoreras: The backend replaces `additionalInfo` wholesale rather than merging. Two things fall out: any key outside the five we manage gets wiped, and clearing all five sends nothing so the old values stick — user clears a field, saves, reopens, it's back.

Spreading the fetched `additionalInfo` and overlaying our five keys (sending `''` instead of omitting) fixes both.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.ts
```diff
@@ -1,26 +1,584 @@
...
+        TabsModule,
+        TagModule,
+        DotMessagePipe,
+        DotUsersReplacementPickerComponent,
+        DotUsersProfileTabComponent
+    ],
     templateUrl: './dot-users-create.component.html',
-    changeDetection: ChangeDetectionStrategy.OnPush
+    styleUrl: './dot-users-create.component.scss',
+    changeDetection: ChangeDetectionStrategy.OnPush,
```
hmoreras, line 105: `OnPush` is the framework default now, so `changeDetection` doesn't need to be here — and `standalone: true` is implied too. Same idea for the signals in this class: they should be `$`-prefixed, and the injected deps `#` instead of `private`. `dot-users-filter-by` next door already does all three if you want a reference.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.ts
```diff
@@ -1,26 +1,584 @@
...
+    host: { class: 'flex h-full min-h-0 flex-col block' }
 })
 export class DotUsersCreateComponent {
     private readonly dialogRef = inject(DynamicDialogRef);
+    private readonly config = inject<DynamicDialogConfig<DialogData>>(DynamicDialogConfig);
+    private readonly fb = inject(FormBuilder);
```
hmoreras: Two bigger ones, both worth an actual decision rather than leaving as-is:

- New forms should be Signal Forms. This one is brand new and #36718–#36720 will all build on it, so it's a good moment to pick.
- HTTP belongs in the store, not the dialog — this component injects the service and runs a three-call `forkJoin`. A small dialog-scoped store would keep it out of here.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.ts
```diff
@@ -1,26 +1,584 @@
...
+                            cmsAdmin: roleKeySet.has(ACCESS_ROLE_KEYS.cmsAdmin),
+                            backend: roleKeySet.has(ACCESS_ROLE_KEYS.backend),
+                            frontend: roleKeySet.has(ACCESS_ROLE_KEYS.frontend),
+                            showGettingStarted: gettingStarted
+                        }
+                    });
+                    this.form.markAsPristine();
+                    this.dataReady.set(true);
+                },
+                error: (error) => this.httpErrorManager.handle(error)
```
hmoreras: `dataReady` only flips in the success branch, so if this call fails the dialog stays open with Save disabled forever and no hint why. Worth an error state that either re-enables Save or closes the dialog.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.ts
```diff
@@ -1,26 +1,584 @@
...
+    protected readonly replacementUser = signal<DotUserListItem | null>(null);
+    /**
+     * Flips on when the user clicks Delete without a valid form.
+     * Used to reveal a footer validation message; the button itself
+     * stays enabled per design convention (no disabled buttons in UI).
+     * Reset every time the picker or email input changes so the user
+     * gets a fresh state after they course-correct.
+     */
+    protected readonly deleteAttempted = signal(false);
+    protected readonly isLoading = signal(false);
```
hmoreras: `isLoading` is set but never read in the template, so the edit dialog shows nothing while it hydrates — just a disabled Save button. Either bind it to a spinner/skeleton or drop the signal.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.ts
```diff
@@ -1,26 +1,584 @@
...
+                        }
+                    });
+                    this.form.markAsPristine();
+                    this.dataReady.set(true);
+                },
+                error: (error) => this.httpErrorManager.handle(error)
+            });
+    }
+
+    private enableCreatePasswordValidators(): void {
```
hmoreras, line 444: `minLength(6)` only gets applied in create mode, so on edit a one-character password sails through. Worth attaching it unconditionally — empty already means "keep current" and the group validator covers that.

core-web/libs/portlets/dot-users/src/lib/components/dot-users-replacement-picker/dot-users-replacement-picker.component.ts
```diff
@@ -0,0 +1,117 @@
...
+ * suggestion list on the client — the backend also rejects invalid
+ * replacements, but pre-filtering avoids showing picks that would
+ * fail on submit.
+ */
+@Component({
+    selector: 'dot-users-replacement-picker',
+    standalone: true,
+    imports: [FormsModule, AutoCompleteModule, DotMessagePipe],
+    templateUrl: './dot-users-replacement-picker.component.html',
+    changeDetection: ChangeDetectionStrategy.OnPush,
```
hmoreras, line 37: Same decorator/naming notes as the shell: drop `changeDetection` and `standalone: true`, `$`-prefix the signals, `#` for the injected deps. Also missing a `.scss` and a spec — the exclusion filtering, the `displayName` fallback chain, the error path and the `selectionChange` output are all worth covering.

One more: `dot-roles/dot-role-users-tab` already has a hand-rolled user autocomplete against this same endpoint with the same item template. Nothing shared exists today so building one here is fine, but putting it where both can use it would avoid two divergent user pickers sitting side by side.

core-web/libs/portlets/dot-users/src/lib/components/dot-users-replacement-picker/dot-users-replacement-picker.component.ts
```diff
@@ -0,0 +1,117 @@
...
+            })
+            .pipe(take(1), takeUntilDestroyed(this.destroyRef))
+            .subscribe({
+                next: (response) => {
+                    const excluded = new Set(this.excludedUserIds());
+                    this.suggestions.set(
+                        response.entity.filter((user) => !excluded.has(user.userId))
+                    );
+                },
+                error: () => this.suggestions.set([])
```
hmoreras: Swallowing the error makes a 500 look identical to "no matches". There's also no loading or empty state for the suggestions, and a new subscription per keystroke with no cancellation — the 300ms delay narrows the race but a slow early response can still overwrite a newer one. A `Subject` + `switchMap` handles all three.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-profile-tab/dot-users-profile-tab.component.ts
```diff
@@ -0,0 +1,136 @@
...
+        InputIconModule,
+        InputTextModule,
+        PasswordModule,
+        ToggleSwitchModule,
+        DotFieldRequiredDirective,
+        DotMessagePipe
+    ],
+    templateUrl: './dot-users-profile-tab.component.html',
+    styleUrl: './dot-users-profile-tab.component.scss',
+    changeDetection: ChangeDetectionStrategy.OnPush,
```
hmoreras, line 68: Same decorator/naming notes here (`changeDetection`, `standalone`, `$` on the signals).

A few others in this file: `form` is an untyped `FormGroup` while `dot-users-form.model.ts` defines interfaces nobody imports — worth wiring those up or deleting them. `showPassword` on L90 is dead, `p-password` already has `toggleMask`. And there's no spec for 136 TS + 376 HTML lines, `deleteRequested` included.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-profile-tab/dot-users-profile-tab.component.html
```diff
@@ -0,0 +1,376 @@
...
+                }
+            </div>
+            <div class="field col-span-2 flex flex-col gap-2">
+                <label for="users-profile-email" dotFieldRequired>
+                    {{ 'users.dialog.field.email' | dm }}
+                </label>
+                <p-iconField iconPosition="left" class="w-full">
+                    <p-inputIcon>
+                        <span
+                            class="material-symbols-rounded text-lg! leading-none"
```
hmoreras, line 111: `text-lg!` / `text-base!` — four `!important`s in this file. If the PrimeNG icon sizing genuinely needs them, a short comment saying why would help; otherwise they can go.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-profile-tab/dot-users-profile-tab.component.html
```diff
@@ -0,0 +1,376 @@
...
+            <span class="text-sm" data-testid="users-profile-metadata-failed-logins">
+                {{ u.failedLoginAttempts ?? 0 }}
+            </span>
+        </div>
+    </div>
+}
+
+<form [formGroup]="form()" class="form flex flex-col gap-6">
+    <section
+        [formGroup]="accountGroup()"
```
hmoreras: Nesting `[formGroup]` inside the outer one gives you three separate root form directives instead of a group tree. It works because they share the model, but `formGroupName="account"` is the usual shape and keeps `dotFieldRequired`'s group lookup predictable.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.html
```diff
@@ -1,13 +1,183 @@
...
+                </p-tab>
+                <p-tab [value]="3" data-testid="users-dialog-tab-api-tokens">
+                    {{ 'users.dialog.tabs.api-tokens' | dm }}
+                </p-tab>
+            </p-tablist>
+        </div>
+
+        <p-tabpanels class="flex min-h-0 flex-1 flex-col">
+            <p-tabpanel [value]="0" class="block min-h-0 flex-1 overflow-auto">
+                <div class="px-7 pt-[22px] pb-7">
```
hmoreras: `pt-[22px]` is an arbitrary value and it's repeated four times here — `pt-6` would do.

Also, all four tab panels render eagerly. Fine now since three are one-liners, but `@defer (when …)` would stop #36719's Permissions iframe from instantiating on every dialog open.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.html
```diff
@@ -1,13 +1,183 @@
...
+    [header]="'users.dialog.delete-confirm.header' | dm"
+    data-testid="users-delete-confirm-dialog">
+    <div class="flex flex-col gap-4">
+        <p class="m-0 text-color">
+            {{ 'users.dialog.delete-confirm.message' | dm }}
+        </p>
+        <div class="field flex flex-col gap-2">
+            <label for="users-delete-replacement-picker">
+                {{ 'users.dialog.delete-confirm.replacement-label' | dm }}
+                <span class="text-red-500" aria-hidden="true">*</span>
```
hmoreras: The required asterisk is hand-rolled here while the Profile tab uses the `dotFieldRequired` directive. Using the directive in both keeps the colour on the theme token instead of `text-red-500`, and the control has no `aria-required` right now.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.scss
```diff
@@ -0,0 +1,30 @@
+:host {
+    display: flex;
```
hmoreras: This duplicates what `host: { class: 'flex h-full min-h-0 flex-col block' }` already sets — pick one. The `::ng-deep p-tabpanel` part below is fine, that's a real PrimeNG override and well explained.

That host class also has both `flex` and `block`, which both set `display`, so whichever Tailwind emits last wins. Same in the profile tab — dropping `block` clears it up.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-form.model.ts
```diff
@@ -0,0 +1,71 @@
...
+ * the profile tab and the create component agree on the same seed.
+ */
+export function generateSecurePassword(): string {
+    const upper = 'ABCDEFGHJKLMNPQRSTUVWXYZ';
+    const lower = 'abcdefghijkmnopqrstuvwxyz';
+    const digits = '23456789';
+    const symbols = '!@#$%^&*';
+    const all = upper + lower + digits + symbols;
+
+    const pick = (source: string) => source.charAt(Math.floor(Math.random() * source.length));
```
hmoreras: `Math.random()` isn't a CSPRNG and this output becomes a real password. `sort(() => Math.random() - 0.5)` is a biased shuffle on top of that. `crypto.getRandomValues` with rejection sampling plus Fisher–Yates fixes both — the 12-char / four-class composition is good as-is.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-form.model.ts
```diff
@@ -0,0 +1,71 @@
+import { AbstractControl, FormGroup, ValidationErrors } from '@angular/forms';
+
+export interface DotUsersAccountForm {
```
hmoreras, line 9: These four interfaces aren't imported anywhere — the shell builds the form inline with `fb.nonNullable.group`. Either use them to type the form (and the profile tab's `form` input) or drop them.

core-web/libs/portlets/dot-users/src/lib/dot-users-list/store/dot-users-list.store.ts
```diff
@@ -136,15 +188,137 @@ export const DotUsersListStore = signalStore(
...
+             * surfaced through the shared HTTP error manager; the list stays
+             * in `loaded` state so the user can retry from the same dialog.
+             *
+             * `gettingStartedChange` optionally chains a toolgroup PUT after
+             * user creation succeeds. A "getting started" failure is soft —
+             * we log it via the shared error manager but the user creation
+             * is still considered successful (the toolgroup can be toggled
+             * again from the same dialog).
+             */
+            createUser(payload: DotUserFormPayload, gettingStartedChange?: 'add' | 'remove') {
```
hmoreras: `createUser` / `updateUser` / `deleteSingleUser` subscribe manually while `loadUsers` just above is a proper `rxMethod`. Beyond consistency, `rxMethod` gives per-call teardown — these currently outlive the component until the response lands — and `switchMap` cancellation, so a double-click on Save wouldn't fire two PUTs.

core-web/libs/portlets/dot-users/src/lib/dot-users-list/store/dot-users-list.store.ts
```diff
@@ -76,16 +89,55 @@ export const DotUsersListStore = signalStore(
...
-                                    status: 'loaded'
+                                    status: 'loaded',
+                                    userRoles: {}
                                 });
+
+                                if (response.entity.length === 0) {
+                                    return of(null);
+                                }
+
+                                const roleFetches = response.entity.map((user) =>
```
hmoreras, line 108: This is 1 + 20 requests per list load, and it re-runs on every sort change and debounced keystroke, each one resolving a role hierarchy server-side. The comment this replaced already suggested a batch endpoint — if that's out of scope here, a linked follow-up would be enough.

core-web/libs/portlets/dot-users/src/lib/dot-users-list/dot-users-list.component.ts
```diff
@@ -38,31 +48,79 @@ import { DotUserListItem } from '../services/dot-users.service';
...
     readonly store = inject(DotUsersListStore);
 
     private readonly dialogService = inject(DialogService);
-    private readonly confirmationService = inject(ConfirmationService);
     private readonly dotMessageService = inject(DotMessageService);
     private readonly destroyRef = inject(DestroyRef);
 
     private readonly searchSubject = new Subject<string>();
 
+    protected readonly bulkDeleteVisible = signal(false);
```
hmoreras: The new signals here need the `$` prefix, and while you're in the file the `private readonly` injections just above could be `#`.

core-web/libs/portlets/dot-users/src/lib/dot-users-list/dot-users-list.component.ts
```diff
@@ -89,53 +147,94 @@ export class DotUsersListComponent {
...
+        this.store.deleteSelectedUsers(replacement.userId);
+    }
+
+    /**
+     * Formats the Roles column: first two role names comma-separated,
+     * followed by `and N more` when the user carries more. Returns
+     * `null` while the per-user role fetch is still in flight so the
+     * cell renders empty instead of a misleading `and -2 more`.
+     */
+    formatRoles(userId: string): { visible: string; more: number } | null {
```
hmoreras: Called from the template, so it runs per row per CD pass and allocates a new object each time. A `computed()` map keyed by `userId` — or formatting it into the store's `userRoles` shape — makes it a single derivation.

core-web/libs/portlets/dot-users/src/lib/services/dot-users.service.ts
```diff
@@ -83,4 +181,74 @@ export class DotUsersService {
...
+     * We rely on this for two things:
+     *   1. Hydrating the Access section toggles by checking for
+     *      well-known role keys (CMS Administrator, DOTCMS_BACK_END_USER,
+     *      DOTCMS_FRONT_END_USER).
+     *   2. Preserving the user's other role memberships on save. The
+     *      backend `PUT /api/v1/users` replaces the full role list, so
+     *      we must send back every role key the user already had, minus
+     *      the access-role keys that are now toggled off.
+     */
+    getUserRoles(userIdOrEmail: string): Observable<DotRoleView[]> {
```
hmoreras, line 230: Worth noting on this method: the endpoint uses the `includeImplicitRoles = true` overload, so the response includes inherited roles alongside directly-granted ones. That's what turns the echo-back at `dot-users-create.component.ts:535` into inherited-becomes-direct.

core-web/libs/portlets/dot-users/src/lib/services/dot-users.service.ts
```diff
@@ -30,6 +32,102 @@ export interface DotUserListItem {
...
+}
+
+/**
+ * Payload accepted by both POST /api/v1/users (create) and PUT
+ * /api/v1/users (update). Mirrors com.dotcms.rest.api.v1.user.UserForm.
+ *
+ * Contract notes:
+ * - Create requires `password`; update leaves it optional and skips the
+ *   password mutation when omitted.
+ * - Omitting `roles` on update means "do not touch role membership".
```
hmoreras: This note is right and it's the behaviour we want — but `dot-users-create.component.ts:535` sets `payload.roles` on every save, so the code does the opposite. Once that's fixed it'd be worth saying outright that the FE never sends `roles` on update.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.spec.ts
```diff
@@ -0,0 +1,324 @@
...
+            expect(spectator.component.initials()).toBe('NU');
+        });
+
+        it('should mark password as required in create mode', () => {
+            const password = spectator.component.form.controls.account.controls.password;
+            expect(password.hasError('required')).toBe(true);
+        });
+
+        it('should not close on save when the form is invalid', () => {
+            spectator.component['save']();
```
hmoreras: These reach into protected members (`save`, `canConfirmDelete`, `onReplacementSelect`, plus `bulkReplacementUser` in the list spec). Every button already has a `data-testid`, so `spectator.click(byTestId('users-dialog-save-btn'))` gets the same assertions and covers the disabled state and `markAllAsTouched` too.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.spec.ts
```diff
@@ -0,0 +1,324 @@
...
+import { DynamicDialogConfig, DynamicDialogRef } from 'primeng/dynamicdialog';
+
+import { DotHttpErrorManagerService, DotMessageService } from '@dotcms/data-access';
+import { MockDotMessageService } from '@dotcms/utils-testing';
+
+import { DotUsersCreateComponent } from './dot-users-create.component';
+
+import { DotUserDetail, DotUserListItem, DotUsersService } from '../services/dot-users.service';
+
+const MOCK_USER: DotUserListItem = {
```
hmoreras: `MOCK_USER` / `MOCK_USER_DETAIL` are now duplicated across this spec, the list spec and the store spec. There's no `createFakeUser` in `@dotcms/utils-testing` yet — adding one would cover all three plus the `dot-roles` user tab.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.ts
AP2300: sending all to update the roles of the user is ok

core-web/libs/portlets/dot-users/src/lib/components/dot-users-replacement-picker/dot-users-replacement-picker.component.ts
AP2300: Addressed the hygiene sub-items on this branch:

- Dropped `standalone: true` (kept `changeDetection: OnPush` on purpose — app runs zoned, dropping it regresses CD perf on signal-heavy components; `dot-users-filter-by` next door does the same).
- Renamed `suggestions` → `$suggestions` and the two injected deps to `#`-prefix.
- Added a spec (`dot-users-replacement-picker.component.spec.ts`, 11 tests → 100% stmts) covering **exclusion filtering**, **displayName fallback chain**, **error path**, and **selectionChange output**.
- Skipped the empty `.scss` — codebase convention (`dot-users-filter-by`, `dot-tags-list`) is to add it only when there's real styling, and this component uses inline utilities only. Happy to add one if you'd rather have the file for consistency.

**Held as a follow-up (needs team convo):**
- Extracting the picker to a shared location so `dot-roles/dot-role-users-tab` can drop its hand-rolled autocomplete and reuse this. Best done **after** #36718 lands, in a dedicated PR, so both consumers are visible in one tree and the shared home (`libs/ui/` vs `libs/portlets/shared/`) can be decided in context.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.html
AP2300: Replaced all four `pt-[22px]` with `pt-6`.

On the `@defer` bullet — holding for #36719: the three sibling tab panels are still placeholder text on this branch, so eager render costs nothing today. Best applied when the real permissions iframe lands on that branch, where the win is measurable.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.html
AP2300: Swapped the hand-rolled `<span>*</span>` on both labels for `class="p-label-input-required"` — same theme-token treatment PrimeNG applies elsewhere in the codebase (`dot-locales`, `dot-form-file-editor`).

Couldn't use the `dotFieldRequired` directive here directly: it injects `FormGroupDirective` as a hard dependency, and the delete-confirm dialog isn't inside a `[formGroup]`. The raw class is what the other portlets that also live outside a form context do.

Added `[attr.aria-required]="true"` on the confirm input. The picker wrapper doesn't take a11y attributes through the component interface yet — if you want that wired up too, easy follow-up on the picker component.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.scss
AP2300: Dropped the `:host { display: flex ... }` block — layout comes from the tailwind host class now. Also dropped the redundant `block` from the host class on both the shell and the profile tab (`flex` was already winning; `block` was dead).

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-profile-tab/dot-users-profile-tab.component.html
AP2300: They're genuinely needed — added an HTML comment at the first occurrence (line 103) explaining it, per your suggestion.

Root cause: Google Material Symbols ships `.material-symbols-rounded { font-size: 24px }`, which ties tailwind's `text-*` classes on specificity. The `!` suffix (which tailwind compiles to `!important`) is the only clean way to force the tailwind size to win. Tried scoping via `:host .material-symbols-rounded { font-size: 1em }` first, but that has higher specificity than `text-lg` and just moves the problem.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-profile-tab/dot-users-profile-tab.component.html
AP2300: Swapped the three nested `[formGroup]` bindings to `formGroupName="account"`, `"additionalInfo"`, `"access"`. Now a single form root + a proper group tree, so `dotFieldRequired`'s parent lookup finds its controls cleanly.

core-web/libs/portlets/dot-users/src/lib/dot-users-list/dot-users-list.component.ts
AP2300: Replaced the template-called `formatRoles(userId)` with a `$rolesByUserId` computed that derives the full map once per `store.userRoles()` change. Template is now a plain `$rolesByUserId()[user.userId]` lookup — one derivation, zero per-row allocations.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.spec.ts
AP2300: Extracted `createFakeUser` + `createFakeUserDetail` to `libs/portlets/dot-users/src/lib/testing/dot-user.mock.ts`. All 5 spec sites now use the shared factory (create, list, store, service, picker).

Kept it inside the portlet rather than `@dotcms/utils-testing` because the `DotUserListItem` / `DotUserDetail` types are still declared in this portlet's service — moving the factory to the shared testing lib requires promoting those types to `@dotcms/dotcms-models` first. Filed as a follow-up so it can be picked up when `dot-roles` needs the same fixture.

core-web/libs/portlets/dot-users/src/lib/dot-users-list/store/dot-users-list.store.ts
AP2300: > Folow up by backend requiered to define a new endpoint to retrieve this infomarion 

Held as a follow-up per your suggestion. Fix needs backend cooperation (bulk `POST /api/v1/roles/users/_batch` accepting a `userId[]`, or enriching the list-filter response with role names) so it's out of scope for this frontend PR.

Tracked alongside the scoped access-endpoint proposal — both wait on a Java-side counterpart to land on a separate branch before this store can drop the per-row `forkJoin`.

core-web/libs/portlets/dot-users/src/lib/components/dot-users-replacement-picker/dot-users-replacement-picker.component.ts
AP2300: Refactored to `Subject + switchMap`:
- Per-keystroke `.subscribe` gone — one long-lived subscription owned by `takeUntilDestroyed`, `switchMap` cancels the in-flight request when a newer query arrives.
- `$isLoading` + `$hasError` signals distinguish loading / error / empty-result from each other; `pTemplate="empty"` swaps copy based on state (new i18n keys `users.dialog.replacement-picker.{loading,error,empty}`).
- Added spec coverage for the three states + the switchMap cancellation race (stale response arriving after a newer query is ignored).

core-web/libs/portlets/dot-users/src/lib/dot-users-list/store/dot-users-list.store.ts
AP2300: Converted `createUser`, `updateUser`, `deleteSingleUser` to `rxMethod`. Each now takes a single object argument (`{ payload, gettingStartedChange? }` / `{ userId, replacementUserId? }`) and pipes through `switchMap` — so a double-click on Save cancels the previous in-flight PUT instead of firing two, and subscription lifetime is bounded by the store's lifecycle.

Updated the two call sites in the list component and the four assertion points in the store spec. Left `deleteSelectedUsers` alone for now — it operates on the bulk-selection with a per-user `forkJoin` and the manual-subscribe pattern is arguably a better fit there (each row's failure is surfaced individually); happy to convert that too if you'd prefer symmetry.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.spec.ts
AP2300: Converted the four `spectator.component['save']()` calls to `spectator.click(saveButton(spectator))` — added a small helper that reaches through p-button's host to the inner native `<button>`. Also added a Save-enabled render check for create mode and an assertion that `markAllAsTouched()` fires when Save is clicked with an invalid form (the required first-name control becomes touched).

Kept the delete-flow bracket accesses (`openDeleteConfirm`, `confirmDelete`, `onReplacementSelect`) as-is on purpose — driving those through the DOM would need us to render the p-autoComplete overlay and simulate keyboard picking, which is fragile against PrimeNG internals. The current tests are unit-style and exercise the intent (state transitions + dialogRef.close payload) which is the concern that would regress. Same reasoning for the bulk-delete flow in the list spec.

core-web/libs/portlets/dot-users/src/lib/services/dot-users.service.ts
AP2300: Added the note to the `getUserRoles` JSDoc — spells out that `/api/v1/roles/users/{id}` uses the `includeImplicitRoles = true` overload, and that this is why echoing the list back at L535 promotes inherited → direct.

Left the `payload.roles` fix itself out (still open on L535 — depends on the team decision about where access-toggle writes go).

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.ts
AP2300: Addressed the **store-extraction** half of this comment:

- New `libs/portlets/dot-users/src/lib/dot-users-create/store/dot-users-create.store.ts` — dialog-scoped signalStore that owns `loadUserDetail` (`getUser` + `getUserRoles` + `getGettingStartedState` via `forkJoin`), the hydration status (`idle` → `loading` → `loaded`/`error`), and the derived data (`detail`, `roleKeys`, `additionalInfo`, `gettingStarted`).
- Shell provides the store on itself (`providers: [DotUsersCreateStore]`) so each dialog instance gets its own, and reads state through computeds (`$isLoading`, `$dataReady`).
- An `effect()` watches the status: transitions to `loaded` patch the form from the store, transitions to `error` close the dialog after the store surfaces its toast.
- Shell no longer injects `DotUsersService` / `DotHttpErrorManagerService` / `DestroyRef`, and `loadUserDetail` is gone from the component body.
- Existing spec still passes untouched (75/75) — the store is created with the mocked service so mocks flow through transparently.

**Signal Forms** decided to go to a follow up issue after this one is closed and resolved

core-web/libs/portlets/dot-users/src/lib/services/dot-users.service.ts
AP2300: Team confirmed the FE's current behavior — echo back the full role list on every save — is intentional and staying. Since `getUserRoles` returns direct + implicit roles, sending the full list back preserves membership without needing an "omit roles" path.

Updated the JSDoc on `DotUserFormPayload` to say that outright: the FE always sends the full role list with Access-toggle deltas applied, and the alternative "omit to leave untouched" mode is documented as not used by this UI. No more contradiction between the doc and the code.

core-web/libs/portlets/dot-users/src/lib/dot-users-list/store/dot-users-list.store.ts
hmoreras: follow in other task, as we need desgin conversation.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.ts
hmoreras: will be solved in a follow up conversation on why we use the roleKey.

hmoreras (approved):

Issue #36909, opened 2026-08-05
Dojo to Angular: Roles Portlet


hmoreras:
## Context

The Roles and Tools portlet (role hierarchy + role-to-user membership + per-role permissions + per-role Tool Group assignment) currently ships as a Dojo + JSP portlet built around `dijit.Tree`, `dijit.layout.TabContainer`, `dojox.grid.DataGrid`, `dotcms.dijit.form.HostFolderFilteringSelect`, and `RoleAjax` DWR calls. This epic migrates it to Angular + PrimeNG and reshapes the information architecture per the new design.

The migration is decomposed into a consolidated spike (current state + backend API + target UI mapping + iframe wrapper design), a single FE + BE implementation task (Angular library + `portlet.xml` registration as an opt-in Beta portlet), and a docs task. A follow-up swap-to-primary task is filed as a placeholder for after the Beta bakes in production.

The Users tab ships as a full Angular implementation. The Permissions and Tools tabs ship as **iframe wrappers** hosting the existing Dojo UI unchanged, because the Permissions Angular redesign is out of scope for this Epic and the Tools redesign is still being defined. Each iframe wrapper is intentionally thin so the eventual Angular replacements can be dropped in without touching the surrounding Angular shell — those replacements are tracked as follow-up `[Task]` issues.

Design reference: https://claude.ai/design/p/20548219-a454-4b14-becf-5dd4dc0b84a1?file=Roles.dc.html&via=share (internal — auth required).

## Current State (Dojo)

| Aspect | Value |
| --- | --- |
| Entry point | Portlet `roles` registered in `dotCMS/src/main/webapp/WEB-INF/portlet.xml` (lines ~227-235) |
| Portlet class | `com.liferay.portlet.JSPPortlet` |
| Shell JSP | `dotCMS/src/main/webapp/html/portlet/ext/roleadmin/view_roles.jsp` |
| Tab JSPs / includes | `view_roles_js_inc.jsp` (~1637 lines — tree + role CRUD + Users + Tool Groups), `view_role_permissions_inc.jsp` + `view_role_permissions_js_inc.jsp` (~980 lines — Permissions accordion) |
| Templates | `host_folder_accordion_title.html`, `host_folder_accordion_entry.html`, `system_host_accordion_title.html`, `system_host_folder_accordion_entry.html` |
| Tab framework | Dojo `dijit.layout.TabContainer` with tabs for role tree, Users, Permissions, Layouts (Tool Groups) |
[...]

PR #37012, opened 2026-08-11
#36936: feat(roles): add PUT /v1/roles/{roleId} for role update

### LINKED ISSUE #36936: [Task] Roles API: add `PUT /v1/roles/{roleId}` for role update + reparent
## Context

The Angular Roles and Tools portlet (FE task #36930) needs to update existing roles — including changing the parent (drag-to-reparent in the roles tree, and the Parent field in the Edit Role dialog). The current REST surface only offers `POST /v1/roles` for **creating** roles. The only path to update an existing role today is DWR `RoleAjax#updateRole` (`dotCMS/src/main/java/com/dotmarketing/business/ajax/RoleAjax.java`, ~line 411), which the Angular portlet is dropping.

Add a `PUT /v1/roles/{roleId}` endpoint that mirrors the existing `POST /v1/roles` body shape (`RoleForm`), so the Angular Edit Role dialog and drag-to-reparent can go through v1 REST end-to-end.

## Endpoint Spec

| Field | Value |
|---|---|
| HTTP | `PUT` |
| Path | `/v1/roles/{roleId}` |
| Class#method | `com.dotcms.rest.api.v1.system.role.RoleResource#updateRole` (new) |
| Path param | `roleId` — existing role ID |
| Request body | `RoleForm` — `roleName*`, `roleKey`, `parentRoleId`, `canEditUsers`, `canEditPermissions`, `canEditLayouts`, `description` |
| Response | `RoleResponseEntityView` (role.toMap()) — same shape POST already returns |
| Auth gates | `requiredBackendUser=true`, `requiredPortlet("roles")`, admin check via `roleAPI.doesUserHaveRole(user, cmsAdminRole)` |
| OpenAPI | `@Operation` annotation required; `swagger-maven-plugin` regenerated `openapi.yaml` committed |

## Behavior

- Load the role by `roleId`; 404 if not found
- Update fields from the form
- Handle `parentRoleId`:
  - Non-null → validate parent exists (`roleAPI.loadRoleById(parentRoleId)`), reject with 400 if parent is a descendant of the role being edited (would create a cycle), then `role.setParent(parentRole.getId())`
  - Null → `role.setParent(role.getId())` (root)
- Persist via `roleAPI.save(role)`
- Surface these existing exceptions as structured errors:
  - `DuplicateRoleKeyException` → 409 with field-level error on `roleKey`
  - `DuplicateRoleException` / `RoleNameException` → 409 with field-level error on `roleName`
- Do not touch `system` or `locked` flags — system roles rejected with 403

## Acceptance Criteria

- [ ] `PUT /v1/roles/{roleId}` implemented with the spec above
- [ ] `@Operation` OpenAPI doc added on the new method
- [ ] `swagger-maven-plugin` regenerated `dotCMS/src/main/webapp/WEB-INF/openapi/openapi.yaml` committed
- [ ] Integration test in `dotcms-integration/src/test/.../RoleResourceIntegrationTest.java` (or nearest existing) covering: happy-path update, reparent to a different parent, reparent to root (null), attempt to create a cycle (→ 400), attempt to edit a system role (→ 403), duplicate key (→ 409), duplicate name (→ 409), missing role (→ 404), non-admin caller (→ 403), missing portlet gate (→ 403)
- [ ] License + role gating preserved
- [ ] Existing `POST /v1/roles` and DWR `updateRole` remain functional (no breaking changes)

## References

- Spike audit: #36929 (comment: https://github.com/dotCMS/core/issues/36929#issuecomment-5210287464)
- Parent epic: #36909
- FE consumer blocked by this: #36930
- Current DWR fallback: `RoleAjax#updateRole` in `dotCMS/src/main/java/com/dotmarketing/business/ajax/RoleAjax.java`
- Reference implementation: `com.dotcms.rest.api.v1.system.role.RoleResource#addNewRole` (POST — same `RoleForm` shape, same auth pattern)

hassandotcms:
### Proposed Changes
  * `PUT /api/v1/roles/{roleId}` — update role name, key, description, can-grant flags,
    and parent. Replaces DWR `RoleAjax#updateRole` (Angular portlet migration, #36909)
  * `parentRoleId: null` → role becomes root (DWR parity)
  * 404 missing role/parent, 403 system/locked, 400 invalid name, 409 duplicate key/name
  * 400 on reparent cycles — new guard, legacy had none server-side
  * Auth: backend user + roles portlet + CMS admin (same gate as `POST /v1/roles`; stricter
    than DWR's users-portlet check — intentional). Gate extracted, shared with POST
  * Response = `ResponseEntityRoleDetailView`, same shape as `GET /v1/roles/{roleid}`
  * `RoleHelper` → `@ApplicationScoped` CDI; update logic `@WrapInTransaction`
  * Regenerated `openapi.yaml`

  ### Checklist
  - [x] 15 integration tests in `RoleResourceIntegrationTest` (MainSuite3a): all field updates,
        reparent to other/root/self/cycle, 403/404/409 cases, POST create regression
  - [x] No breaking changes: additive endpoint; DWR and POST behavior unchanged

  Fixes #36936

This PR fixes: #36936


dotCMS/src/main/java/com/dotcms/rest/api/v1/system/role/RoleHelper.java
```diff
@@ -4,25 +4,152 @@
...
+            throw new DoesNotExistException("Role not found: " + roleId);
+        }
+
+        if (role.isSystem() || role.isLocked()) {
+            throw new DotSecurityException(
+                    String.format("Role '%s' (%s) is a system or locked role and cannot be updated",
+                            role.getName(), role.getId()));
+        }
+
+        role.setName(roleForm.getRoleName());
```
fabrizzio-dotCMS: **In-place mutation of the cached `Role` — poisons the cache when the update is rejected.**

`roleAPI.loadRoleById()` returns the *cached instance*, not a copy: `RoleFactoryImpl.getRoleById` does `r = rc.get(roleId)` and `RoleCacheImpl.get` returns the reference straight out of the cache region. So these setters mutate the object that every other reader on this node will get.

Failure scenario: `PUT` a `roleKey` that already belongs to another role. `roleAPI.save` throws `DuplicateRoleKeyException`, the transaction rolls back, the client correctly gets a 409 — but the in-memory `Role` keeps the rejected name/key, so a subsequent `GET /v1/roles/{id}` on that node returns phantom values until a cache flush. Worse, `RoleCacheImpl.add` also indexes the role under `keyGroup + roleKey`, so the cache ends up holding a role advertising a `roleKey` that doesn't exist in the DB. Same applies to the 400 (invalid name), 400 (cycle) and 404 (missing parent) paths — note the parent guards below run *after* name/key/description have already been mutated.

Suggested fix: validate everything first, then mutate a detached copy (`new Role()` + setters, or `BeanUtils.cloneBean`) and pass that to `save`. This is exactly what `RoleFactoryImpl.save` already does internally — it loads a fresh Hibernate instance and copies properties onto it rather than persisting the caller's object. A `CacheLocator.getRoleCache().remove(role)` in the catch blocks would be a minimum mitigation, but working on a copy is the clean fix.

fabrizzio-dotCMS (commented):
Solid PR overall — good docs, real integration coverage, OpenAPI regenerated, and the auth gate extraction doesn't relax `POST` (it already required roles portlet + CMS admin) while improving the `SecurityLogger` call site. I also verified the two things I was most worried about and they're fine: reparenting descendants is handled (`RoleFactoryImpl.save` recomputes the FQN of children/grandchildren), and `@WrapInTransaction` still applies even though `RoleResource` instantiates the helper with `new` (it's woven by ByteBuddy per package, not a CDI interceptor).

One blocking issue though: the update mutates the **cached** `Role` instance before validating, so a rejected request leaves poisoned state in the local cache. Details inline, plus a few smaller notes.

dotCMS/src/main/java/com/dotcms/rest/api/v1/system/role/RoleHelper.java
```diff
@@ -4,25 +4,152 @@
...
+
+            final Role parentRole = this.roleAPI.loadRoleById(parentRoleId);
+            if (null == parentRole || !UtilMethods.isSet(parentRole.getId())) {
+                throw new DoesNotExistException("Parent role not found: " + parentRoleId);
+            }
+
+            // findRoleHierarchy walks getParent() up to the root, so it returns the proposed
+            // parent and all its ancestors — if the edited role is among them, the reparent
+            // would create a cycle
+            for (final Role ancestor : this.roleAPI.findRoleHierarchy(parentRole)) {
```
fabrizzio-dotCMS: **Reuse the existing hierarchy API.** This loop is exactly `roleAPI.isParentRole(role, parentRole)` (`RoleAPIImpl:583`) — same `findRoleHierarchy` walk, already `@CloseDBIfOpened`. Since self-parenting is guarded a few lines above, `isParentRole` excluding the child itself is not a problem here.

Also worth knowing about the primitive you're building on: `findRoleHierarchy` does `while(!role.getParent().equals(role.getId()) && i < 100)` — it NPEs if any ancestor has a null `parent` (the factory has `if(UtilMethods.isSet(r.getParent()))` checks, so unset parents are considered possible), and it silently truncates at 100 levels, which means on an already-corrupted hierarchy this guard passes without detecting anything. Not introduced by this PR, but the new endpoint is now the main caller.

dotCMS/src/main/java/com/dotcms/rest/api/v1/system/role/RoleHelper.java
```diff
@@ -4,25 +4,152 @@
...
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
 /**
  * Helper to encapsulate Roles logic
  * @author jsanca
  */
+@ApplicationScoped
```
fabrizzio-dotCMS: **Half-applied CDI.** `RoleHelper` becomes `@ApplicationScoped` with an `@Inject` constructor, but `RoleResource` still holds `private final RoleHelper roleHelper = new RoleHelper();`, so the bean is never resolved through CDI and the annotation is effectively dead.

To be clear, nothing is broken: `RoleAPI` does have a producer (`APILocatorProducers#getRoleAPI`) so the bean would be satisfiable, and `@WrapInTransaction` still takes effect because it's woven by the ByteBuddy agent over the `com.dotcms` package (`ByteBuddyFactory`), not by a CDI interceptor. But either inject the helper into the resource or drop the CDI annotations, otherwise the next reader will assume proxying is in play.

dotCMS/src/main/java/com/dotcms/rest/api/v1/system/role/RoleResource.java
```diff
@@ -281,60 +282,144 @@ public Response addNewRole(
...
+		final User user = this.initRequireRolesPortletAndCmsAdmin(request, response);
+
+		final Role updatedRole = this.roleHelper.updateRole(roleId, roleForm, user);
+
+		// same response shape as GET /v1/roles/{roleid}
+		final List<RoleView> childrenRoles = new ArrayList<>();
+		final List<String> roleChildrenIdList = null != updatedRole.getRoleChildren()
+				? updatedRole.getRoleChildren() : new ArrayList<>();
+		for (final String childRoleId : roleChildrenIdList) {
+			childrenRoles.add(new RoleView(this.roleAPI.loadRoleById(childRoleId), new ArrayList<>()));
```
fabrizzio-dotCMS, line 419: Minor: N+1 `loadRoleById` to build the children views. It mirrors what `GET /{roleid}` already does so it's consistent, and it's cache-backed, but for a role with many children this is one call per child on every update.

dotCMS/src/main/java/com/dotcms/rest/api/v1/system/role/RoleResource.java
```diff
@@ -281,60 +282,144 @@ public Response addNewRole(
...
+					content = @Content(mediaType = "application/json")),
+		@ApiResponse(responseCode = "404",
+					description = "Role or parent role not found",
+					content = @Content(mediaType = "application/json")),
+		@ApiResponse(responseCode = "409",
+					description = "Conflict - duplicate role key, or duplicate role name under the same parent",
+					content = @Content(mediaType = "application/json"))
+	})
+	@PUT
+	@Path("/{roleId}")
```
fabrizzio-dotCMS: Nit: the existing `GET` on this resource uses `@Path("/{roleid}")` (lowercase `id`) while this uses `{roleId}`. Harmless for JAX-RS matching, but it makes the two endpoints look like different paths at a glance — worth aligning.

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/system/role/RoleResourceIntegrationTest.java
```diff
@@ -0,0 +1,458 @@
...
+            roleAPI.unLock(role);
+        }
+    }
+
+    /**
+     * Given Scenario: A role is updated to use another role's roleKey.
+     * Expected Result: 409 ConflictException (DuplicateRoleKeyException from RoleAPIImpl.save).
+     */
+    @Test(expected = ConflictException.class)
+    public void testUpdateRole_duplicateKey_conflict() throws Exception {
```
fabrizzio-dotCMS, line 287: **Coverage gap that maps to the cache issue above.** The rejection tests assert the exception type only (or, for the cycle test, only `parent` — which happens to be mutated *after* the throw, so it passes). None of them re-read the role and assert the *other* fields survived a rejected update, which is precisely how the cached-instance mutation slips through.

Suggest adding to `testUpdateRole_duplicateKey_conflict` / `testUpdateRole_invalidName_badRequest` / `testUpdateRole_missingParent_notFound`: after the expected exception, `roleAPI.loadRoleById(role.getId())` and assert name, key and description still match the original. Those assertions should fail on the current implementation.

dotcms-integration/src/test/java/com/dotcms/MainSuite3a.java
```diff
@@ -102,6 +102,7 @@
         ContentToStringUtilTest.class,
         CacheResourceIntegrationTest.class,
         InodeExistenceCheckIntegrationTest.class,
+        com.dotcms.rest.api.v1.system.role.RoleResourceIntegrationTest.class,
```
fabrizzio-dotCMS: Nit: every other entry in this suite is referenced by simple name with an import at the top; this one is a fully-qualified inline reference. Add the import for consistency.

dotCMS/src/main/java/com/dotcms/rest/api/v1/system/role/RoleHelper.java
hassandotcms: this is insightful. fixing it.

dotCMS/src/main/java/com/dotcms/rest/api/v1/system/role/RoleHelper.java
hassandotcms: makes sense

dotCMS/src/main/java/com/dotcms/rest/api/v1/system/role/RoleResource.java
hassandotcms: we aim to change GET as well, because there is a new version of it coming in.

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/system/role/RoleResourceIntegrationTest.java
hassandotcms: resolved all the feedback.

fabrizzio-dotCMS (approved):

PR #37023, opened 2026-08-11
fix(content-drive): admin-aware Unlock warning, lock-owner visibility in the action preview, and combined skip/fail reporting

### LINKED ISSUE #36975: Workflow Center: admin-aware Unlock warning, lock-owner visibility in preview, and combined skip/fail reporting
### Description

Three findings from the code review of #36921 (*Content Drive: Add Lock and Unlock quick actions*) that were triaged as follow-up rather than fixed in that PR. All three are frontend-only. (1) and (2) share the same plumbing and should be done together; (3) is independent and much smaller.

---

#### 1. The Unlock warning is wrong for administrators

`warnWhen` in `utils/action-center.ts` flags locked rows using `contentEditable`, which `BrowserAPIImpl.WfData` derives as `lockedBy == currentUser` — it never consults the CMS Administrator role. The server-side gate, `ESContentletAPIImpl.canLock`, returns `true` for a CMS Admin *before* it looks at the lock owner.

So an administrator selects 5 locked rows, 3 held by other users, is told "3 of these are locked by another user…", fires it, and all 5 unlock. The warning is pure noise for the only role that can act on it.

> #36921 softened the copy to "may require administrator permission" so it no longer *asserts* a false outcome, but the warning still shows for admins.

A second, narrower false positive: a user holding EDIT at the **content-type** level rather than on the contentlet. `contentEditable` checks contentlet WRITE only, while `canLock` accepts either.

#### 2. The preview cannot show which rows hold foreign locks

Quick actions were routed through the preview table in #36921 specifically so a user could uncheck locks held by others. But the preview renders no per-row lock information — every row gets the same `dot-contentlet-status-badge`.

The user sees `Unlock (10) ⚠3`, drills in, and has no way to tell which 3 rows to uncheck. The capability the preview exists to provide is unreachable.

#### 3. Skips and failures are reported as mutually exclusive

The toast ladder in `dot-content-drive-shell.component.ts` is `failCount > 0` → `skippedCount > 0` → plain, but the backend can return both. A bulk workflow fire over a mixed-type selection where some contentlets' schemes don't own the action (**skipped**) *and* some are locked (**failed**) reports only the failure copy.

That is worse than dropping a number: `executed-with-fails` reads as "these N could not be actioned" and attributes the whole shortfall to permissions or locks, when part of it was items merely sitting on the wrong workflow step. The user's next move — go unlock things — is then wrong.

---

### Acceptance Criteria

**Admin-aware Unlock warning**

- [ ] The current user's admin flag is sourced from `DotCurrentUserService.getCurrentUser()` (`DotCurrentUser.admin`) and loaded once into the Content Drive store rather than per dialog open
- [ ] `warnWhen` returns `false` for every row when the current user is a CMS Administrator, so no warning icon and no `warningCount` appear on the Unlock row
- [ ] A non-admin still sees the warning, counting only locked rows the user does not hold
- [ ] The warning count remains `0` when every locked row is the user's own
- [ ] Unlock still fires **all** locked rows regardless of the warning — the warning does not filter the payload
- [ ] Unit tests cover the admin and non-admin paths, including the case where the admin flag has not resolved yet (must not warn incorrectly while loading)

**Lock-owner visibility in the preview**

- [ ] The preview table reuses `DotFolderListViewComponent` from `@dotcms/portlets/content-drive/ui` rather than the hand-rolled table in `dot-content-drive-action-preview`
- [ ] `DotFolderListViewComponent` accepts a `selection` input so the parent owns the included set, and a `disabled` input for the in-flight state
- [ ] Its existing selection-clearing effect no longer discards a caller-provided selection
- [ ] Rows whose lock is held by another user are visually marked, distinct from rows locked by the current user
- [ ] Marked rows stay **checked** by default — the user chooses whether to uncheck them, and the fired count continues to match what the row advertised
- [ ] The marker keys off the same signal as `warnWhen`, so an administrator sees no rows marked
- [ ] The main Content Drive grid is unaffected by the new inputs (existing behaviour and specs still pass)
- [ ] Unit tests cover: marker shown for a foreign lock, absent for an own lock, absent for an admin, and unchecking a marked row removes it from the payload

**Combined skip/fail reporting**

- [ ] A result carrying both `skippedCount > 0` and `failCount > 0` reports both numbers to the user
- [ ] The wording distinguishes the two causes — skipped items are on a workflow step the action does not own; failed items were refused (permission or lock)
- [ ] Results with only skips, only failures, or neither keep their current copy
- [ ] Toast severity remains `warn` whenever anything failed
- [ ] Unit tests cover all four combinations (neither / skips only / fails only / both)

### Priority

Medium

### Additional Context

**Origin.** Review of #36921 by @zJaaal — findings 2, 3 and 6 of 8. The other five were addressed in that PR: the integration suite registration, the misleading unlock copy, the `?? inodes.length` fallback that reported every item as a success when the response carried no summary, the Action Center reopen guard, and escaping the action name in the toolbar's `[innerHTML]` label.

**Decisions already taken** (so the implementer does not re-litigate them):

| Question | Decision |
|---|---|
| Preview implementation | Reuse `DotFolderListViewComponent` — it already renders a per-row lock icon, and the preview component's own header comment says it deliberately copies that grid. Removes the drift risk between two near-identical tables. |
| Marker behaviour | Mark but leave checked. Keeps the attempt-all contract and keeps the fired count equal to the advertised count. |
| Admin flag source | `DotCurrentUserService.getCurrentUser()`. A server-side alternative (having the drive search return per-row unlockability, using the same `canLock` the server enforces) would be more accurate but needs a `BrowserAPIImpl` change — out of scope here, worth its own issue if the client-side signal proves insufficient. |

**Deliberately not in scope.** Locking archived content is refused by the Action Center client-side only; `ContentletAPI.lock` has no archived guard, so `PUT /api/v1/content/_lock/{inode}`, a direct `LOCK` system-action fire, and the row context menu all still allow it. Making that a real invariant means changing `ContentletAPI.lock` for every caller — a behaviour change to a long-standing API that belongs in its own issue.

**Key files**

| Concern | File |
|---|---|
| `warnWhen` / quick-action config | `core-web/libs/portlets/dot-content-drive/portlet/src/lib/utils/action-center.ts` |
| Preview table | `.../components/dialogs/dot-content-drive-action-center/components/dot-content-drive-action-preview/` |
| Grid component to reuse | `core-web/libs/portlets/dot-content-drive/ui/src/lib/dot-folder-list-view/` |
| Toast ladder | `.../lib/dot-content-drive-shell/dot-content-drive-shell.component.ts` |
| `contentEditable` origin | `dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java` (`WfData`) |
| Server-side lock gate | `dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ESContentletAPIImpl.java` (`canLock`) |

Related: #36844, PR #36921.

rjvelazco:
Resolves #36975 — the three follow-up findings from the review of #36921, all frontend.

### Proposed Changes

**1. The Unlock warning no longer fires at administrators**

`warnWhen` flagged locked rows using `contentEditable`, which `BrowserAPIImpl.WfData` derives as `lockedBy == currentUser` — it never consults the CMS Administrator role. `ESContentletAPIImpl.canLock` returns `true` for a CMS Admin *before* it looks at the lock owner, so an admin was told "3 of these are locked by another user", fired, and unlocked all of them. The warning was noise for the only role that could act on it.

`DotCurrentUser.admin` is now resolved once per portlet load into the Content Drive store and fed to the quick-action predicates. The check moved into an exported `isLockedByAnotherUser`, which the preview's per-row marker also uses — so the count on the row and the rows marked in the preview cannot disagree. An unresolved flag counts as a non-admin: the hint says a foreign lock *may* require administrator permission, so over-warning is honest, while staying quiet would let a non-admin fire with no heads-up.

**2. The preview shows which rows hold foreign locks**

Quick actions were routed through a preview in #36921 specifically so a user could uncheck locks held by others — but every row rendered identically, so the capability was unreachable. The preview's hand-rolled table (a deliberate copy of the grid's markup, per its own header comment) is replaced by `DotFolderListViewComponent`, which already renders a per-row lock icon. Foreign locks render amber with a "Locked by another user" tooltip; they stay **checked**, so the fired count keeps matching what the action row advertised and dropping one remains the user's decision.

Six new inputs on the grid, each defaulting to today's behaviour so the main Content Drive grid is untouched:

| Input | Why |
|---|---|
| `selection` | Parent owns the checked set; the table reports rather than applies |
| `disabled` | Freezes the selection while a run is in flight |
| `dataKey` | Language variants share an `identifier`; inodes are what gets fired |
| `lazy` | The preview holds every row already and pages in memory |
| `lockedByOthers` | Which rows to mark — the caller decides, since it needs the admin role |
| `readOnly` | No drag, context menu, open-on-click, kebab or column sorting |
| `visibleColumns` | The full set overflows a dialog; the preview keeps checkbox / title / status / type |

**3. Skips and failures are reported together**

The toast ladder checked `failCount`, then `skippedCount`, then neither — but the backend returns both. A bulk fire over a mixed-type selection can skip items whose scheme does not own the action *and* be refused on items that are locked; the mixed result showed the failure copy alone, blaming permissions or locks for the whole shortfall and sending the user off to unlock content that was never the problem. Anything short of a clean run now reports all three numbers, each beside its own cause.

### Checklist
- [x] Tests
- [x] Translations
- [x] Security Implications Contemplated — none. No new endpoints or payloads; the admin flag only softens a client-side warning and never gates an action. Every item the user selected is still submitted, and the server remains the sole authority on who may unlock what.

### Additional Info

**Built test-first.** Every change went red before green, in six cycles: controlled selection → cell overflow → `dataKey`/`lazy` → `lockedByOthers`/`readOnly` → `disabled` → the swap itself. Two cycles caught real defects that would otherwise have shipped: a first pass at `visibleColumns` nested the `extraColumns` loop inside the Type column, so hiding Type would have silently dropped every caller-provided extra column; and an early `lazy` test passed before the feature existed, because `dataToRender` slices to `rows` whether or not lazy — it was rewritten to assert something that actually discriminates.

**Three deviations from the acceptance criteria**, all deliberate:

1. *Skips-only now warns*, where the AC says severity stays `warn` whenever anything **failed**. A skipped item did not get the action either, so a green toast overstated the outcome.
2. *A fails-only run renders "0 skipped"* (and vice versa), where the AC says only-skips and only-failures keep their current copy. One shape for every non-clean result means naming each cause and its number, which is what keeps the message honest.
3. *The lock marker applies to every action's preview*, not only Unlock — a lock held by somebody else fails a Publish or an Archive just as readily. Still keyed off the same predicate as `warnWhen`, so administrators see nothing marked.

**Two removed i18n keys.** `content-drive.action-center.toast.executed-with-skips` and `…executed-with-fails` are replaced by `…executed-partial`. Both became unreferenced and neither has shipped — they were added in #36921.

**Not in scope, still open.** The preview's row density and thumbnail size still come from the grid (a compact list would need a density input), and the header typography is the shared grid's. Locking archived content remains refused client-side only, as the issue records.

### Testing

`content-drive-ui` 227 passed / 4 suites · `portlets-content-drive` 1064 passed / 30 suites · lint clean on both.

### Video

https://github.com/user-attachments/assets/4f4e67f4-4893-4236-878d-717c38ba6973

This PR fixes: #36975


rjvelazco:
Thanks — both checked, one confirmed and fixed, one that does not reproduce. Pushed as 0473b2bba4.

### 🟠 High — non-lazy pagination stuck on page 1: **confirmed, fixed**

Reproduced before touching the code, with a test that clicks the paginator on a non-lazy table with **no `offset` bound** — the way the preview actually wires it:

```
● should reach page two without the caller binding an offset
    Expected: 5
    Received: 20
```

Twenty rows after clicking "next" — it advanced and snapped straight back. Fixed exactly as suggested:

```ts
protected onFirstChange() {
    if (!this.$lazy()) {
        return;
    }
    const dataTable = this.dataTable();
    if (dataTable) {
        dataTable.first = this.$offset();
    }
}
```

Your read on why the old coverage missed it is right, and it is the more useful half of the finding: `should honour the offset when paging an in-memory list` injected `offset=20` to *simulate* page two, so it asserted a state no real non-lazy caller ever produces. I have kept it — an explicitly-bound offset still positions the table — and added the realistic one alongside, which fails without the guard.

### 🟡 Medium — rows re-sorted by `modDate`: **does not reproduce**

The assumption is that PrimeNG applies the initial `sortField` to non-lazy data. It does not, at least in PrimeNG 21 on this path. Rendering a non-lazy table with two rows whose `modDate` order is the reverse of their given order:

```ts
const chosenOrder = [
    { …, title: 'Chosen first',  modDate: '2020-01-01T00:00:00Z' },
    { …, title: 'Chosen second', modDate: '2030-01-01T00:00:00Z' }
];
```

`sortField="modDate" [sortOrder]="-1"` would put *Chosen second* on top. The rendered order is the given one, and the test passed on its first run — no fix needed. It stays as a regression guard, since nothing previously stated that a non-lazy caller's ordering is preserved, and that matters here: the preview's rows are the user's selection in the order the grid showed it.

Both suites still green — `content-drive-ui` 229 / 4 suites, `portlets-content-drive` 1064 / 30 — and lint clean.

rjvelazco:
@claude review this code to find bugs

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/dot-content-drive.store.ts
```diff
@@ -332,6 +336,22 @@ export const DotContentDriveStore = signalStore(
...
                 patchState(store, { selectedItems: items });
             },
+            /**
+             * Resolves the logged-in user's CMS Administrator role, once per portlet load.
+             *
+             * A failure leaves the flag at its `false` default rather than surfacing an error: the
+             * role only softens a warning, so a portlet that cannot answer "is this an admin?" should
+             * still work — it just keeps warning, which is what it did before the flag existed.
+             */
+            loadCurrentUserIsAdmin() {
```
zJaaal, line 346: should this be a rxMethod?

rjvelazco:
Both applied — pushed as 7a02ec8ee5.

### 🟡 Medium — extra columns misaligned when `visibleColumns` hides Type: **confirmed, fixed**

Real, and my own comment on that line documented the wrong behaviour ("Appends when the type column is hidden, so extras never fall off the end silently"). The body anchors the extra-column loop at the Type slot whether or not Type renders; the header keyed off the *rendered* index, so a hidden `contentType` made `findIndex` return `-1` and the extras were appended to the end.

Reproduced before touching it — `visibleColumns: ['title', 'modUser']` with one extra column:

```
- Expected     + Received
  "name",
- "Author",
  "Edited-By",
+ "Author",
```

Header said `name, Edited-By, Author`; the body rendered `name, Author, Edited-By`. So an author value would have sat under the Edited-By heading.

Fixed by anchoring on the canonical position rather than the rendered one:

```ts
const afterType = columns.findIndex((column) => column.order > TYPE_COLUMN_ORDER);
const insertAt = afterType === -1 ? columns.length : afterType;
```

`TYPE_COLUMN_ORDER` is read off `HEADER_COLUMNS` rather than hardcoded, so the header and body cannot drift apart if the column order changes.

Your note on why the existing coverage missed it is the useful half again: `should still append caller-provided extra columns` hides everything after the title, which makes both orderings coincide. It takes a visible column *after* Type to tell them apart. The new test asserts the header labels **and** the body cell sequence together, which is the pair that actually catches a mismatch.

One correction for the record: this was not raised in the earlier round — that pass covered the non-lazy pagination bug and the `modDate` ordering question. No matter either way, it was worth raising.

### Coverage gap — "applies to every action": **added**

Fair. Added a component-level case that opens a **Publish** preview over a foreign-locked row and asserts exactly one marker across two rows, so the deviation-#3 claim rests on something more than the shared predicate's unit tests.

---

`content-drive-ui` 230 / 4 suites · `portlets-content-drive` 1065 / 30 · lint clean · `nx format:check --base=origin/main` clean (that last one is what the earlier red build was — a Prettier failure surfacing under a job named "Frontend Unit Tests", fixed in 7c89db0f0b).

zJaaal (commented):
Code review of the three changes. Six findings, all frontend, ordered by severity. Nothing here blocks the intent of the PR; finding 1 is the one I would fix before merge.

The column ordering concern I initially had (extras landing after `actions` when Type is hidden) is already handled by `TYPE_COLUMN_ORDER`, so it is not listed.

_Reviewed by Claude (Opus 5) on @zJaaal's behalf._

core-web/libs/portlets/dot-content-drive/ui/src/lib/dot-folder-list-view/dot-folder-list-view.component.html
```diff
@@ -2,23 +2,23 @@
...
+    (selectionChange)="onSelectionChange($event)"
     (onPage)="onPage($event)"
     (onSort)="onSort($event)"
     [first]="$offset()"
     (firstChange)="onFirstChange()"
-    dataKey="identifier"
+    [dataKey]="$dataKey()"
     sortMode="single"
     selectionMode="multiple"
     sortField="modDate"
```
zJaaal, line 24: **Hardcoded `sortField` / `sortOrder` now sort, and mutate, the caller's array.**

These two are static attributes, so with the new `[lazy]="false"` path PrimeNG takes the `!this.lazy` branch in `Table.onChanges` on every `value` change and calls `sortSingle()`, which does `this.value.sort(...)` **in place** (`primeng-table.mjs:1443-1445`, `:1652`).

Three consequences for the action preview:

1. The confirmation list renders newest-modDate-first instead of the order the user selected. That is exactly the re-ordering the `readOnly` rationale says it is avoiding; `readOnly` only hides the sortable headers, it does not stop the sort itself.
2. It mutates the array behind the `$previewItems()` computed, which is the *same object* the parent stored via `$includedItems.set(previewItems)` (action-center component, lines 374 and 483). A memoized signal value and the live selection array both get reordered from outside Angular.
3. `sortSingle()` emits `onSort`, which this component forwards as its `sort` output. Any future non-lazy consumer binding `(sort)` gets a spurious sort event on every data change.

Suggested fix: bind `[sortField]` / `[sortOrder]` and leave them unset when `lazy` is false, or expose them as inputs alongside the other six.

_Reviewed by Claude (Opus 5) on @zJaaal's behalf._

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/dot-content-drive.store.ts
```diff
@@ -332,6 +336,22 @@ export const DotContentDriveStore = signalStore(
...
+                dotCurrentUserService
+                    .getCurrentUser()
+                    .pipe(
+                        take(1),
+                        catchError(() => EMPTY)
+                    )
+                    .subscribe(({ admin }) => patchState(store, { currentUserIsAdmin: !!admin }));
```
zJaaal: **Destructuring the response can throw past the `catchError`.**

`catchError(() => EMPTY)` sits upstream of `subscribe`, so it only covers observable errors. The destructure `({ admin })` runs inside the subscriber: if `/api/v1/users/current/` answers with an empty or `null` body (a 204, a proxy stripping the body, a session-expired gateway returning no JSON), this throws a `TypeError` that RxJS rethrows as an unhandled error, during the store's `onInit`, for a flag the doc comment explicitly describes as non-essential.

```ts
.subscribe((user) => patchState(store, { currentUserIsAdmin: !!user?.admin }));
```

That also keeps the documented "a failure leaves the flag at its `false` default" behaviour true for a malformed response, not just for a failed request.

_Reviewed by Claude (Opus 5) on @zJaaal's behalf._

core-web/libs/portlets/dot-content-drive/ui/src/lib/dot-folder-list-view/dot-folder-list-view.component.html
```diff
@@ -63,73 +63,100 @@
                 [class.text-gray-900]="state.dragOverRowId() === item.identifier"
                 [class.bg-primary-100]="state.dragOverRowId() === item.identifier"
                 [pSelectableRow]="item"
+                [pSelectableRowDisabled]="$disabled()"
```
zJaaal: **A stray click on a preview row silently drops it from the payload.**

`pSelectableRowDisabled` is bound only to `$disabled()`, not to `$readOnly()`, and PrimeNG's `metaKeySelection` defaults to `false`, so a plain click toggles the row.

The checkbox cell, the title and the thumbnail all call `stopPropagation()`, but the status cell and the content-type cell do not. Clicking either one while reading the confirmation list unchecks that row and removes its inode from what Execute fires, with no affordance suggesting the row is clickable. The hand-rolled table this replaces only toggled via the checkbox.

If row-click selection is not wanted in a confirmation list, `readOnly` should feed `pSelectableRowDisabled` too: `[pSelectableRowDisabled]="$disabled() || $readOnly()"`.

_Reviewed by Claude (Opus 5) on @zJaaal's behalf._

core-web/libs/portlets/dot-content-drive/ui/src/lib/dot-folder-list-view/dot-folder-list-view.component.ts
```diff
@@ -429,10 +598,16 @@ export class DotFolderListViewComponent implements OnInit {
...
+     * Handles selection changes in the table and emits selected items.
+     *
+     * Records the set for the uncontrolled case and always reports it. A controlled table renders
+     * the caller's set until the caller echoes this back, so the parent can decline a change.
+     *
+     * @param items The table's new selection
      */
-    onSelectionChange() {
-        this.selectionChange.emit(this.selectedItems);
+    onSelectionChange(items: DotContentDriveItem[]) {
```
zJaaal, line 673: **The "controlled" contract described here is not enforceable.**

The doc says a controlled table "renders the caller's set until the caller echoes this back, so the parent can decline a change". PrimeNG has already written `_selection` and `selectionKeys` before it emits: both `toggleRowWithCheckbox` and `toggleRowsWithCheckbox` set `preventSelectionSetterPropagation = true` and update internal state first.

So if the parent declines, either by leaving its signal untouched or by returning an array that happens to be reference-equal, `[selection]` never changes, Angular never re-writes the input, and the table keeps the user's change on screen while the parent's state says otherwise.

No current caller declines, so this is latent. Worth either softening the comment or having the parent force a new array reference when it rejects a change.

_Reviewed by Claude (Opus 5) on @zJaaal's behalf._

core-web/libs/portlets/dot-content-drive/ui/src/lib/dot-folder-list-view/dot-folder-list-view.component.ts
```diff
@@ -610,6 +791,14 @@ export class DotFolderListViewComponent implements OnInit {
      * https://github.com/primefaces/primeng/issues/11898#issuecomment-1831076132
      */
     protected onFirstChange() {
+        // Lazy only. `$offset` is the parent's cursor into a server-paged list; a non-lazy caller
+        // holds every row, has no cursor and never binds it — so re-pinning `first` to a permanent
+        // `0` snapped the table back to page one on every page change, leaving everything past the
+        // first page unreachable.
+        if (!this.$lazy()) {
```
zJaaal, line 866: **The same lazy guard is missing in `onPage` (line 594).**

For a non-lazy table an in-memory page change still emits the `paginate` output, whose contract is "fetch me this page from the server", a request a caller holding every row cannot satisfy. It also resets `$loadingRows` on each page turn.

Harmless for the preview today only because it does not bind `(paginate)`, but the output now lies for non-lazy consumers. The guard added here belongs at the top of `onPage` as well.

_Reviewed by Claude (Opus 5) on @zJaaal's behalf._

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dialogs/dot-content-drive-action-center/components/dot-content-drive-action-preview/dot-content-drive-action-preview.component.html
```diff
@@ -1,77 +1,29 @@
...
+    - `visibleColumns` — just enough to recognise a row. The dialog is far narrower than the
+      portlet, and the grid's full set overflows it: the title collapses to an ellipsis and the
+      trailing columns disappear behind the dialog edge.
 
     Selection is a one-way `[selection]` binding plus `(selectionChange)`, rather than `[(selection)]`,
     so the parent stays the single owner of the included set.
 -->
-<p-table
-    [value]="items()"
+<dot-folder-list-view
```
zJaaal, line 19: **The preview fires a redundant languages request on every open.**

`DotFolderListViewComponent.ngOnInit` (line 521) unconditionally calls `dotLanguagesService.get()` to build `state.languagesMap`. This preview is a second instance of that grid, created and destroyed each time the user drills into an action, and `visibleColumns` deliberately excludes `languageId`, so the map it fetches is never read.

Net effect: one wasted `/api/v2/languages` call per drill-in. Guarding the fetch on `$visibleColumnSet().has('languageId')` would cover it without touching the main grid.

_Reviewed by Claude (Opus 5) on @zJaaal's behalf._

core-web/libs/portlets/dot-content-drive/ui/src/lib/dot-folder-list-view/dot-folder-list-view.component.html
rjvelazco: **Not reproducible on PrimeNG 21 — no sort, no mutation.** Verified before changing anything, because the mechanism you describe is exactly right for the code you quote.

That branch lives in `Table.onChanges(simpleChange)`. In this version that method has **zero call sites**, and `ngOnChanges` does not exist on the class at all:

```
$ grep -c "ngOnChanges" primeng-table.mjs        -> 0
$ grep -n "this.onChanges(\|\.onChanges("        -> (no matches)
```

It is dead legacy code, so the `!this.lazy` → `sortSingle()` path never runs. Confirmed with a spy on a non-lazy table configured exactly as the preview is:

```
sortSingle calls= 0 | sortField= modDate | sortOrder= -1 | lazy= false | sortMode= single
givenOrder= A,B | rendered= A,B
```

Two tests now pin all three of your consequences: rows render in the order the caller passed (`should keep an in-memory list in the order it was given`) and the caller's array is not reordered in place (`should not reorder the array the caller passed in`) — the second aimed squarely at the `$previewItems()` / `$includedItems` aliasing you flagged, which is the part that would have been worst.

Left as-is rather than bound to inputs, since there is nothing to suppress. If PrimeNG revives that path in a later version, the two tests fail rather than the preview quietly reordering.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/dot-content-drive.store.ts
rjvelazco: **Fixed** in 6a95b13754.

```ts
.subscribe((user) => patchState(store, { currentUserIsAdmin: !!user?.admin }));
```

One correction on the mechanism: it does not surface as a synchronous throw. RxJS 7 routes an error thrown in a `next` handler to `reportUnhandledError`, so it is an async unhandled error rather than one that propagates out of `next()`. My test for it (`should stay false when the response carries no body`) therefore passes both before and after the change — it pins the documented contract, not the bug, and I would rather say so than imply it caught something.

The fix is still right for the reason you give: the flag documents "a failure leaves it at `false`", and that should hold for a successful response with no body just as much as for a failed request.

core-web/libs/portlets/dot-content-drive/ui/src/lib/dot-folder-list-view/dot-folder-list-view.component.html
rjvelazco: **Fixed** in 6a95b13754 — `[pSelectableRowDisabled]="$disabled() || $readOnly()"`.

This is the finding I am most glad you caught: it is a straight regression against the table this PR deleted, which could only be toggled from its checkbox. The status and content-type cells are the two that do not `stopPropagation`, so reading a status badge could silently drop an inode from what Execute fires.

Covered by `should not toggle a row when a cell without its own click handler is clicked`, which clicks the status cell on a read-only table and asserts nothing is emitted. It fails without the guard.

Note the main grid keeps row-click selection — it is a browsing affordance there, which is the same line `readOnly` already draws for drag, context menu and open-on-click.

core-web/libs/portlets/dot-content-drive/ui/src/lib/dot-folder-list-view/dot-folder-list-view.component.ts
rjvelazco: **Already addressed** — in efc2896437, pushed about ten minutes after this review was written, so the diff you were reading predates it.

You are right that the comment described something the code could not deliver, and the mechanism you name is the one: PrimeNG writes `_selection` and `selectionKeys` before it emits, and sets `preventSelectionSetterPropagation`, so a declining parent leaves the row visually changed.

Rather than soften the comment, the contract is now enforced. An effect re-asserts the effective selection onto the table, triggered by a revision counter as well as by the value — a decline changes no value, so a value-only dependency would never fire — and it has to run after the handler returns, since PrimeNG deletes the row key *after* emitting and would undo anything restored inline.

Two tests cover it, and both fail with the effect disabled: `should stay on the parent's set when the parent declines a change` and `should follow the parent when it narrows the set to something else`.

Worth flagging that this is exactly the kind of framework-internals coupling that argues for the layer split — filed as #37024.

core-web/libs/portlets/dot-content-drive/ui/src/lib/dot-folder-list-view/dot-folder-list-view.component.ts
rjvelazco: **Fixed** in 6a95b13754 — same guard at the top of `onPage`.

Agreed on the reasoning: for a non-lazy table `paginate` is a request the caller cannot satisfy, and it also primed `$loadingRows` for skeletons that never render. Two tests: `should not ask the parent to fetch a page it already holds` (clicks the real paginator on a non-lazy table) and `should still ask the parent to fetch a page while lazy`, so the guard cannot be over-applied.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dialogs/dot-content-drive-action-center/components/dot-content-drive-action-preview/dot-content-drive-action-preview.component.html
rjvelazco: **Fixed** in 6a95b13754 — the fetch is now gated on the locale column being visible.

Correct on all counts: the map feeds only the locale column, `visibleColumns` excludes it, and the preview is a fresh instance of the grid on every drill-in — so it was one wasted `/api/v2/languages` call per open. The main grid passes no `visibleColumns`, so it still fetches exactly as before.

Two tests: `should not fetch languages when the locale column is hidden` and `should fetch languages when the locale column is shown`.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/dot-content-drive.store.ts
rjvelazco: Good question. I left it as a one-shot `take(1)` + `subscribe` deliberately, but the trade-off is real.

`rxMethod` earns its keep when a call is driven by changing inputs and needs switch/cancel semantics — it wires the source to the store's injector scope and manages the subscription for you. This call has none of that: it fires once from `onInit`, takes no argument, and the role cannot change within a session. `core-web/libs/portlets/CLAUDE.md` also points the other way for this shape — "`take(1)` on one-shot HTTP calls", and explicitly "do NOT add `take(1)` inside `rxMethod`".

The one genuine advantage would be teardown: a plain `subscribe` is not tied to the store's lifecycle, so navigating away mid-flight leaves the request in flight until it completes. With `take(1)` that is bounded to a single request that then completes on its own, and the only side effect is a `patchState` on a store nobody is reading — so the cost is a wasted write, not a leak.

Happy to switch it if you would rather have it uniform with the other store methods — say the word and I will.

KevinDavilaDotCMS (approved):

dario-daza (approved):

PR #37027, opened 2026-08-11
fix(evergreen-tracks): authenticate Hub tag reads — anonymous pagination now blocked past offset 1000

### LINKED ISSUE #37025: evergreen-tracks: track promotion fails — Docker Hub now blocks anonymous tag pagination past offset 1000
## Summary

The daily `evergreen-tracks-promote` cron failed on 2026-08-11 ([run 31481784007](https://github.com/dotCMS/core/actions/runs/31481784007)) and will keep failing every day. The same defect also breaks the **`latest` promote in the release pipeline** (`cicd_6-release.yml`), so the next GA cut would not move `dotcms/dotcms:latest`.

Nothing was mis-applied: the run died in the `plan` (dry-run) job, before `gate`/`apply`, so no tag moved.

## Root cause

`registry.list_tags()` reads the Docker Hub API **anonymously** — the module docstring says *"Public repos need no auth"* — and walks every page of the repo's tags:

https://github.com/dotCMS/core/blob/main/.github/actions/core-cicd/evergreen-tracks/src/evergreen_tracks/registry.py

Docker Hub now caps **anonymous** pagination at offset 1000:

```
GET https://hub.docker.com/v2/namespaces/dotcms/repositories/dotcms/tags?page=11&page_size=100
403 {"message":"pagination offset too large for anonymous requests; sign in to page further"}
```

`dotcms/dotcms` has **7,831 tags** (79 pages), so the walk dies on page 11 every time.

This is a Docker-side change, not ours — the 2026-08-10 run walked all 79 pages successfully in 44s; the 2026-08-11 run failed after 6s. No evergreen-tracks commit landed in between.

Reproduced locally at the failing commit (`4d814eb`):

```
requests.exceptions.HTTPError: 403 Client Error: Forbidden for url:
https://hub.docker.com/v2/namespaces/dotcms/repositories/dotcms/tags?page=11&page_size=100
```

## Secondary bug: the failure was undiagnosable

Both the `plan` and `apply` steps capture the engine with `2>/dev/null` ("drop uv's stderr chatter"). That discarded the entire Python traceback, so CI reported a bare `Process completed with exit code 1` and the Slack alert carried no cause. The root cause had to be re-derived by running the tool locally.

## Fix

1. Authenticate the read path. `executor.hub_login()` already mints a Hub JWT and every calling workflow already has `DOCKER_USERNAME` / `DOCKER_TOKEN` for `docker/login-action` — the read path just never used them. Authenticated requests are not subject to the offset cap.
2. Export those secrets to the promote (plan + apply) and release `latest`-promote steps. The admin workflow already exports them.
3. Stop swallowing stderr — surface the engine's error on failure.

## Impact

- Daily `standard` / `trailing` promotion: broken, fails every morning.
- Release pipeline `latest` promote: broken on the next GA cut.
- `evergreen-tracks-admin` (hold/taint): also broken on the read path today; fixed by (1) since it already exports the creds.
- Customer environments: unaffected. Manifests pin immutable digests and only the in-cluster reconciler rewrites them, at its biweekly window.

sfreudenthaler:
## What happened

The daily `evergreen-tracks-promote` cron failed on 2026-08-11 ([run 31481784007](https://github.com/dotCMS/core/actions/runs/31481784007)) and would have failed every morning from here on. It died in the `plan` (dry-run) job, before `gate`/`apply`, so **no tag was moved** — the alert's "one track may have moved" caveat did not apply this time.

## Root cause

Docker Hub now refuses **anonymous** pagination past offset 1000:

```
GET https://hub.docker.com/v2/namespaces/dotcms/repositories/dotcms/tags?page=11&page_size=100
403 {"message":"pagination offset too large for anonymous requests; sign in to page further"}
```

`registry.list_tags()` read the Hub API with no auth at all (*"Public repos need no auth"*) and walked every page. `dotcms/dotcms` has **7,831 tags = 79 pages**, so it 403s on page 11.

This is a Docker-side change: the 2026-08-10 run walked all 79 pages in 44s, the 2026-08-11 run died after 6s, and no evergreen-tracks commit landed in between.

Also broken, same read path — **the `latest` promote in `cicd_6-release.yml`**, which would have failed on the next GA cut, and the admin (hold/taint) workflow.

## The fix

`executor.hub_login()` already minted a Hub JWT for deletes, and every calling workflow already had `DOCKER_USERNAME`/`DOCKER_TOKEN` for `docker/login-action` — the read path just never used them. Authenticated requests aren't subject to the offset cap.

- `registry.py` — one login per walk, JWT on every page request. Auth stays optional so tests and small repos need no creds; a creds-less 403 now raises a message naming the cause instead of a bare `HTTPError`.
- promote (`plan` + `apply`) and release `latest`-promote — export the two secrets. The admin workflow already did, so it self-heals.
- `docker/login-action` does **not** cover the Hub API, which is why being "logged in" wasn't enough.

## Secondary fix: the failure was undiagnosable

Both promote steps captured the engine with `2>/dev/null` ("drop uv's stderr chatter"), which discarded the whole traceback. CI showed a bare `Process completed with exit code 1` and the Slack alert carried no cause — the root cause had to be re-derived by running the tool locally. stderr now goes to a file (it must stay off stdout, which is the plan text `apply` diffs) and is printed on failure.

## Verification

- `uv run pytest` — 68 passed, including two new tests: every tag request carries the JWT when creds are set, and a creds-less 403 explains itself.
- **Live CI run on this branch** ([31534540024](https://github.com/dotCMS/core/actions/runs/31534540024)): the `plan` job walked all 79 pages authenticated in ~29s and produced a real plan — `standard -> 26.07.28-01`. Cancelled at the approval gate so nothing applied.

Merging this makes tomorrow's 10:00 UTC cron pass and apply that pending `standard` move.

Closes: #37025

This PR fixes: #37025

### DIFF
--- a/.github/actions/core-cicd/evergreen-tracks/src/evergreen_tracks/registry.py
+++ b/.github/actions/core-cicd/evergreen-tracks/src/evergreen_tracks/registry.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 
 import os
 from dataclasses import dataclass
+from functools import lru_cache
 
 import requests
 
@@ -42,19 +43,33 @@ def _hub_message(resp: requests.Response) -> str:
         return ""
 
 
+@lru_cache(maxsize=4)
+def _jwt(username: str, password: str) -> str:
+    """Cached per credential pair: one login per process, not one per request."""
+    return hub_login(username, password)
+
+
 def _auth_headers() -> dict[str, str]:
     """Hub JWT header, or {} when no creds are in the environment."""
     username = os.environ.get("DOCKER_USERNAME")
     password = os.environ.get("DOCKER_TOKEN")
     if not username or not password:
         return {}
-    return {"Authorization": f"JWT {hub_login(username, password)}"}
+    return {"Authorization": f"JWT {_jwt(username, password)}"}
 
 
-def list_tags(repo: str) -> list[Tag]:
-    """All tags in the repo with their manifest digests, following pagination."""
+def list_tags(repo: str, *, name_filter: str = "") -> list[Tag]:
+    """Tags in the repo with their manifest digests, following pagination.
+
+    `name_filter` applies Hub's SUBSTRING filter (`?name=`), which is what keeps the
+    common path off the ~79-page full listing. It is observed behaviour rather than a
+    documented contract; if Hub ever drops it the filtered queries simply return the
+    whole repo and callers still get a correct — merely slower — answer.
+    """
     namespace, name = repo.split("/", 1)
     url = f"{_HUB}/namespaces/{namespace}/repositories/{name}/tags?page_size=100"
+    if name_filter:
+        url += f"&name={name_filter}"
     # One login per walk, not per page: the JWT outlives a full 79-page listing.
     headers = _auth_headers()
     out: list[Tag] = []
--- a/.github/actions/core-cicd/evergreen-tracks/src/evergreen_tracks/registry.py
+++ b/.github/actions/core-cicd/evergreen-tracks/src/evergreen_tracks/registry.py
@@ -1,9 +1,19 @@
-"""Read side: list tags and digests from the Docker Hub API. Public repos need no auth."""
+"""Read side: list tags and digests from the Docker Hub API.
+
+Reads are authenticated when DOCKER_USERNAME/DOCKER_TOKEN are set. Hub refuses
+ANONYMOUS pagination past offset 1000 ("pagination offset too large for anonymous
+requests; sign in to page further"), and dotcms/dotcms is at ~7.8k tags / 79 pages,
+so an unauthenticated walk 403s on page 11. Auth is therefore required in practice
+for the real repos; it stays optional so tests and small repos need no creds.
+"""
 from __future__ import annotations
 
+import os
 from dataclasses import dataclass
 
 import requests
+
+from .executor import hub_login
 
 _HUB = "https://hub.docker.com/v2"
 _TIMEOUT = 30
@@ -24,13 +34,30 @@ def _digest_of(result: dict) -> str | None:
     return None
 
 
+def _auth_headers() -> dict[str, str]:
+    """Hub JWT header, or {} when no creds are in the environment."""
+    username = os.environ.get("DOCKER_USERNAME")
+    password = os.environ.get("DOCKER_TOKEN")
+    if not username or not password:
+        return {}
+    return {"Authorization": f"JWT {hub_login(username, password)}"}
+
+
 def list_tags(repo: str) -> list[Tag]:
     """All tags in the repo with their manifest digests, following pagination."""
     namespace, name = repo.split("/", 1)
     url = f"{_HUB}/namespaces/{namespace}/repositories/{name}/tags?page_size=100"
+    # One login per walk, not per page: the JWT outlives a full 79-page listing.
+    headers = _auth_headers()
     out: list[Tag] = []
     while url:
-        resp = requests.get(url, timeout=_TIMEOUT)
+        resp = requests.get(url, headers=headers, timeout=_TIMEOUT)
+        if resp.status_code == 403 and not headers:
+            raise RuntimeError(
+                f"Docker Hub refused anonymous pagination of {repo} at {url}. "
+                "Set DOCKER_USERNAME and DOCKER_TOKEN — Hub caps anonymous "
+                "requests at offset 1000 and this repo has more tags than that."
+            )
         resp.raise_for_status()
         body = resp.json()
         for result in body.get("results", []):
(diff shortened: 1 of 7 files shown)


sfreudenthaler (commented):
ready for human review.  

> [!NOTE]  
> Critical content demanding immediate user attention due to potential risks.
Note that #37028 stacks on top of this to drastically reduce the calls and time spent.  But still we’d like to use authenticated calls after talking with @wezell.

> [!WARNING]  
> This whole approach side steps the cleanup and pruning of the dockerhub dotcms/dotcms container registry.  it’s crufty and a lot of it could be cleaned up as a separate concern.

sfreudenthaler:
Both findings were real. Reproduced each, fixed, and mutation-tested the fixes.

**1. Test isolation — confirmed and fixed.** Reproduced exactly as described:

```
$ DOCKER_USERNAME=someuser DOCKER_TOKEN=sometoken uv run pytest tests/test_registry.py
E   Request: POST https://hub.docker.com/v2/users/login
E   Available matches: GET .../tags Method does not match
FAILED tests/test_registry.py::test_list_tags_paginates_and_returns_name_digest
```

I first put the autouse fixture in `test_registry.py` as suggested, then checked whether that was enough — it wasn't. A module-level fixture only protects its own module, so I added a second test file exercising the same env-reading path and it still failed with ambient creds. Moved to `tests/conftest.py`, which covers every module including ones not written yet. Suite now passes with and without creds exported; deleting the conftest makes 3 tests fail again.

Checked the rest of the suite for the same exposure while I was in there — `_delete_marker` also reads those vars, but the admin tests patch their I/O, so `test_registry.py` was the only module doing real HTTP through env-reading code.

**2. 403 message accuracy — confirmed, and worse than described.** Your assumption checked out, and the cause is more specific than I'd claimed. Measured:

| request | result |
|---|---|
| nonexistent repo, page 1 | **404** `object not found` |
| nonexistent namespace, page 1 | **404** `object not found` |
| `dotcms/dotcms` page 10 | 200 |
| `dotcms/dotcms` page 11 | **403** `pagination offset too large for anonymous requests` |
| `library/hello-world` page 11 (3 tags!) | **403** same message |

So 403 is unambiguous today — missing repos 404. But the cap keys off the **offset alone**, not the tag count: even a 3-tag repo 403s on page 11. My "this repo has more tags than that" was an invented cause. The message now quotes Hub's own `message` field, with a status-code fallback when the body isn't JSON (Hub can serve HTML from a gateway):

```
Docker Hub refused an anonymous read of dotcms/dotcms at .../tags?page=11&page_size=100
— pagination offset too large for anonymous requests; sign in to page further.
Set DOCKER_USERNAME and DOCKER_TOKEN: anonymous requests cannot page past offset 1000.
```

**On your first non-blocking note** — the one-login-per-walk JWT lifetime. Worth knowing that #37028 removes the concern rather than mitigating it: the common path never walks 79 pages, so the longest authenticated read drops from ~29s to ~3s. The full listing survives only as the `admin` path and a fallback.

Your isolation finding also led me to a second instance of the same bug in #37028, which you couldn't have seen from this diff: `_jwt`'s `lru_cache` is process-global and outlives a test. Two tests with identical creds and different tokens — the second received the first's JWT. Same fixture now calls `_jwt.cache_clear()`.

69 tests pass here (2 new), 74 on the stacked branch, both with and without ambient creds.

.github/actions/core-cicd/evergreen-tracks/src/evergreen_tracks/registry.py
```diff
@@ -24,13 +34,44 @@ def _digest_of(result: dict) -> str | None:
...
+    password = os.environ.get("DOCKER_TOKEN")
+    if not username or not password:
+        return {}
+    return {"Authorization": f"JWT {hub_login(username, password)}"}
+
+
 def list_tags(repo: str) -> list[Tag]:
     """All tags in the repo with their manifest digests, following pagination."""
     namespace, name = repo.split("/", 1)
     url = f"{_HUB}/namespaces/{namespace}/repositories/{name}/tags?page_size=100"
```
wezell, line 70: Is there another filter parameter that we could send to limit the results?  Even if it is "images that start with v*"

wezell: If not, this is fine.

wezell (approved):
Approving with comments

.github/actions/core-cicd/evergreen-tracks/src/evergreen_tracks/registry.py
sfreudenthaler: that’s in #37030 stacked on this

sfreudenthaler: pulled into this change.

Issue #36940, opened 2026-08-07
Batch folder permission resolution into a single round-trip per page


ihoffmann-dot:
### Problem Statement

`FolderAPIImpl.searchFolders` resolves permissions with one `PermissionAPI.filterCollection` call per permission type. For a single page of folders that is up to five sequential calls:

| Call | Scope |
|---|---|
| `PERMISSION_READ` | the full pre-pagination collection |
| `PERMISSION_CAN_ADD_CHILDREN` | the page |
| `PERMISSION_EDIT` | the page (only when `includePermissions=true`) |
| `PERMISSION_PUBLISH` | the page (only when `includePermissions=true`) |
| `PERMISSION_EDIT_PERMISSIONS` | the page (only when `includePermissions=true`) |

Each call independently re-resolves the requesting user's roles (`loadRolesForUser`) before issuing its own `getPermittedIds` query, and `getPermittedIds` chunks the id list by 500. So the work scales with *number of permission types* × *ceil(page size / 500)*, when the underlying data — the permission bitmask per folder id — could be fetched once.

A single batch returning the bitmask per id would collapse all of them into one round-trip, after which the per-type sets are pure bit tests.

Raised by @fabrizzio-dotCMS while reviewing #36889 ([review comment](https://github.com/dotCMS/core/pull/36889), plus an inline note on the three adjacent calls).

### Why it wasn't done in #36889

That PR added the three opt-in types and was already a REST contract change; reworking `PermissionAPI`'s batch surface is a different blast radius. It was explicitly deferred rather than folded in.

### Why it's worth doing

Beyond the round-trips, this would **remove the reason the `perPage` cap exists**. #36889 had to add `content.drive.folder.search.permissions.max.per.page` (default 200, 400 above it) precisely because per-type resolution makes large pages expensive. With one batch the cost stops scaling with the type count, and the cap — a user-visible 400 that the frontend has to code around — could plausibly be relaxed or dropped.

### Suggested approach

1. Add a batch method to `PermissionAPI` returning `Map<String, Integer>` (permission id → bitmask) for a collection of `Permissionable`s, resolving the user's roles once.
2. Rewrite `FolderAPIImpl.resolvePagePermissions` to call it once and derive the four id sets by bit test. `PermissionAPI.Type.fromBitsAsNames(int)` already exists but returns the **canonical** names, which include `WRITE` instead of `EDIT` — see the caveat below.
[...]

PR #37029, opened 2026-08-11
fix(content-drive): open the folder context menu from the sidebar tree #36595

### LINKED ISSUE #36595: [BUG] Content Drive: Sidebar folders don't open a right-click context menu like the table

ihoffmann-dot:
### Parent Issue

#36595

### Summary

Frontend half of #36595, on top of the backend half (#36889). Right-clicking a folder in the Content Drive **sidebar tree** did nothing, while right-clicking a folder **row in the table** opened a context menu (Edit folder / Edit permissions).

The shared `dot-folder-list-context-menu` is already rendered unconditionally in the shell and driven by the store, so the sidebar does not need its own menu. It needs to publish the right-clicked folder to the store in the shape the menu and the "Edit folder" dialog consume, and every node has to arrive carrying its permissions.

### The constraint that shapes this PR

`includePermissions=true` caps `per_page` at `content.drive.folder.search.permissions.max.per.page` (default 200) and rejects anything larger with a **400**. The deep-link hierarchy load fetches one large page per ancestor level, so it could not ask for permissions. That left right-click dead on exactly the folders visible when the portlet opens, which is the issue's original symptom.

Verified against a running instance rather than assumed:

| Request | Result |
|---|---|
| `per_page=10000&includePermissions=true` | **400** |
| `per_page=201&includePermissions=true` | **400** |
| `per_page=200&includePermissions=true` | 200, permissions populated |

### What changed

**Cold load carries permissions.** `FOLDER_TREE_HIERARCHY_PAGE_SIZE` drops from 10000 to 200, the cap, and every level requests `includePermissions`. No node reaches the tree without them, whichever way it got there, so right-click stays synchronous and opens immediately.

**A deep-linked ancestor past that page is fetched directly and pinned.** One name-narrowed request matched on exact path, not a page-until-found walk: we already know the folder's path, so paging to rediscover it is wasted round-trips. It is pinned to the **top** of its level rather than appended, since dropped at the end it reads as the next folder in sort order and sits next to "Load more" where it is easy to miss.

**`mergeFolderNodePage` handles the resulting duplicate.** Page far enough and the pinned folder arrives again in its real position. The already-rendered node wins on identity but takes the incoming position, so it settles where it belongs with its expanded subtree and selection intact.

**`nextPage` is threaded through `FolderTreeHierarchyLevel`,** replacing a hardcoded `2`. The hierarchy pages at 200 while load-more pages at 40, so a fixed resume point re-requested folders already on screen. It is derived from folders *fetched*, never the rendered count, so a pinned node cannot shift it. Confirmed end to end: page 6 at 40/page lands exactly on the first unseen folder, no gap and no overlap.

**Right-click covers the whole row,** not just the label text — the toggler, the indent gutter and the space past a short name all respond, as the table's `<tr>` does. Resolved from the DOM via `.p-tree-node-content` (a node's own row, excluding its children, so a child row can never resolve to its parent), mirroring the drag-and-drop already in that component. `dot-tree-folder` stays presentational: it emits folder data, it does not search its own input for a matching node.

**The tree now scrolls to the folder the drive opened on.** The hierarchy already expands down to it, but a level can hold hundreds of folders, so on a deep link it was drawn far below the fold with the viewport at the top. The reveal already existed for table navigation and was gated on a flag only the table sets; both paths now share `#revealNode`, deferred through `afterNextRender` because at the moment the store publishes a cold-loaded selection the tree is not on screen yet. That deferral also fixes the table's path, which scrolled before a still-loading branch had rendered its rows.

**`loadFolders` sets `sidebarLoading` back to `true`.** Only the initial state ever set it, so every later cold load (a site change) left the previous site's tree on screen while its replacement was fetched, with no indication anything was happening.

**Gating no longer throws on a folder without permissions.** `contentlet.permissions.includes(...)` had no optional chaining, so a folder reaching the menu without the field threw a `TypeError` instead of degrading to "no actions".

### Bug fixed here: the Edit-folder dialog built the wrong path

`#getAssetPath` anchored on `store.path()` — the folder currently **open** — rather than the folder being edited. From the table the two coincide, because the table lists the open folder's children. From the sidebar they do not: any folder at any depth can be edited. Editing `/campaigns/2026/marketing-assets/` while `/documents/` was open would have PUT to `//host/documents/marketing-assets/`: a 404, or a **silent overwrite of a same-named folder under the open path**.

Create still anchors on the open folder (that is where a new folder lands); edit derives the parent from the edited folder's own `path`. Both sources already carry a full path, so one fix serves both views.

### What happens when the folder cannot be resolved

Left unpinned, and the selection falls back to the root. The tree still renders everything the user *can* see.

That is also what a folder the user lacks READ on looks like. `FolderAPIImpl.searchFolders` calls `permissionAPI.filterCollection(…, PERMISSION_READ, …)` and applies no permission check to the `path` scope, so an inaccessible folder is a **filtered row, not a rejected request** — confirmed: unreachable paths return `200` with an empty list, never a 403. Two things follow. The tree never collapses, and the pin cannot reveal a folder you have no rights to, because the lookup runs the same permission-filtered query. Existence is not disclosed.

The pin request itself is wrapped in `catchError` so a transient failure degrades to "unpinned" too. It runs inside a `forkJoin`, so without that an unlucky 500 on a best-effort extra request would have rejected the whole hierarchy load and collapsed the tree to its root — costing every readable folder to save one pin. Caught in review.

### Typing: `DotContentDriveActionableFolder`

The menu and dialog need 11 folder fields; `DotContentDriveFolder` (the table's row) has 21. The 10 extras are not returned by `/folder/search`. Rather than fabricate them for sidebar folders or loosen the table's type, both views converge on a narrow `DotContentDriveActionableFolder` that `DotContentDriveFolder` extends. `isFolder()` narrows to it, so a `DotContentDriveItem` still narrows to the full `DotContentDriveFolder` at the table's call sites. No table behavior changes.

### Testing

`portlets-content-drive` 1244/1244, `content-drive-ui` 245/245, `data-access` 754/754, lint clean on all three.

New coverage concentrates on the parts most likely to break silently:

- **`mergeFolderNodePage`** — dedupe, repositioning, and that the retained node keeps its children and expansion. Plus, separately, that `onLoadMore` *actually calls it* on both the nested and root branches: removing it from both call sites had left every test green, so the duplicate-render could have been reintroduced unnoticed.
- **Ancestor pinning** — top of level not appended, nested and root levels, no lookup when already on page 1, level untouched when unresolvable, tree survives a failed pin request, and `nextPage` unaffected by the pin.
- **The cold-load reveal** — scrolls on load, not while loading, not before render, and not for a load-more selection.
- **Right-click** — whole row, nested node not its parent, and the "All folders" / "Load more" / empty-space cases that must keep the native menu.

Each of these was mutation-checked rather than trusted green: inverting the prepend, swapping the merge for a naive concat, and dropping the `catchError` each fail the intended tests.

Manually verified against a local instance with a seeded 253-folder level, both nested and at a site root.

### Notes for the reviewer

- The `name` filter needs 2+ characters and is a *contains* match, so a one-character folder name or a level with more same-substring siblings than fit one page can leave a deep-linked folder unpinned. Both need a level already wider than 200. An exact-match or identifier filter on `/folder/search` would close it.
- Cold load now resolves permissions for every folder it fetches, and re-runs on every site change via `getSiteFoldersEffect`. #36940 (batch folder permission resolution) removes the per-folder loop that makes this cost scale.
- The tree writes `node.data` into a `data-json-node` DOM attribute, which the right-click handler now reads alongside drag-and-drop.

### Explicitly out of scope

- Any backend change — #36889 shipped that half.
- `DotFolderTransformerImpl.contentDriveView`'s per-folder N+1 permission loop (tracked by #36940).
- Moving the sidebar's loading state out of `dot-tree-folder` into the shell. Worth doing, but it changes a required input and every spec that sets it.
- The table's own context menu behavior, which is unchanged.

This PR fixes: #36595

This PR fixes: #36595

### DIFF
--- a/core-web/libs/data-access/src/lib/dot-folder/dot-folder.service.spec.ts
+++ b/core-web/libs/data-access/src/lib/dot-folder/dot-folder.service.spec.ts
@@ -210,6 +210,29 @@ describe('DotFolderService', () => {
             const req = spectator.expectOne(url, HttpMethod.GET);
             req.flush({ entity: [], pagination: { currentPage: 1, perPage: 40, totalEntries: 0 } });
         });
+
+        it('should send includePermissions when the caller opts in', () => {
+            spectator.service
+                .searchFolders({ siteId: 'site-1', includePermissions: true })
+                .subscribe();
+
+            const url =
+                '/api/v1/folder/search?siteId=site-1&includePermissions=true&page=1&per_page=40';
+            const req = spectator.expectOne(url, HttpMethod.GET);
+            req.flush({ entity: [], pagination: mockPagination });
+        });
+
+        it('should omit includePermissions entirely when not requested', () => {
+            // The backend defaults it to false, so sending `includePermissions=false` on every tree
+            // request would be pure noise.
+            spectator.service
+                .searchFolders({ siteId: 'site-1', includePermissions: false })
+                .subscribe();
+
+            const url = '/api/v1/folder/search?siteId=site-1&page=1&per_page=40';
+            const req = spectator.expectOne(url, HttpMethod.GET);
+            req.flush({ entity: [], pagination: mockPagination });
+        });
     });
 
     describe('createFolder', () => {
--- a/core-web/libs/data-access/src/lib/dot-folder/dot-folder.service.ts
+++ b/core-web/libs/data-access/src/lib/dot-folder/dot-folder.service.ts
@@ -96,6 +96,12 @@ export class DotFolderService {
             httpParams = httpParams.set('direction', params.direction);
         }
 
+        // Only sent when opting in: the backend defaults it to false and caps `per_page` when it is
+        // true, so an always-on `includePermissions=false` would be noise on every tree request.
+        if (params.includePermissions) {
+            httpParams = httpParams.set('includePermissions', 'true');
+        }
+
         httpParams = httpParams.set('page', String(params.page ?? DEFAULT_FOLDER_SEARCH_PAGE));
         httpParams = httpParams.set(
             'per_page',
--- a/core-web/libs/dotcms-models/src/lib/dot-content-drive.model.ts
+++ b/core-web/libs/dotcms-models/src/lib/dot-content-drive.model.ts
@@ -17,33 +17,58 @@ export interface DotContentDriveLazyLoadEvent {
     forceUpdate?: () => void;
 }
 
-export interface DotContentDriveFolder {
-    __icon__: 'folderIcon';
+/**
+ * The folder fields required to drive the shared folder actions — the context menu's permission
+ * gating and the "Edit folder" dialog's payload.
+ *
+ * Deliberately narrower than {@link DotContentDriveFolder}: the table sources folders from
+ * `POST /api/v1/drive/search` (a full folder row), while the sidebar tree sources them from
+ * `GET /api/v1/folder/search`, which returns only the fields listed here. Rather than fabricate
+ * the table-only fields (`modDate`, `owner`, `iDate`, …) for sidebar folders, both views converge
+ * on this contract, and only consumers that genuinely need the full row ask for
+ * `DotContentDriveFolder`.
+ */
+export interface DotContentDriveActionableFolder {
+    type: 'folder';
+    identifier: string;
+    /** The folder's own name (last path segment). */
+    name: string;
+    /** The folder's own full path, e.g. `/application/blog/`. */
+    path: string;
+    title: string;
+    sortOrder: number;
+    showOnMenu: boolean;
+    /** Comma-separated file-name masks allowed in this folder, e.g. `*.jpg,*.png`. */
+    filesMasks: string;
     defaultFileType: string;
     /**
      * Folder upload preference: `DOTASSET`/`FILEASSET` forces every upload to that base type,
      * `null`/`undefined` means "ask each time" (no preference). Backed by #35577.
      */
     defaultBaseType?: string | null;
+    /**
+     * Permission types the requesting user holds on this folder.
+     *
+     * Required, and always an array by the time a folder reaches an action: whoever builds this
+     * object resolves the folder's permissions first (or substitutes `[]` when they cannot be
+     * resolved), so gating never runs against `null`/`undefined`. The "not yet resolved" state
+     * lives upstream, on `DotFolder.permissions` / the tree node's data.
+     */
+    permissions: PermissionType[];
+}
+
+export interface DotContentDriveFolder extends DotContentDriveActionableFolder {
+    __icon__: 'folderIcon';
     description: string;
     extension: 'folder';
-    filesMasks: string;
     hasTitleImage: boolean;
     hostId: string;
     iDate: number;
-    identifier: string;
     inode: string;
     mimeType: string;
     modDate: number;
-    name: string;
     owner: string | null;
     parent: string;
-    path: string;
-    permissions: PermissionType[];
-    showOnMenu: boolean;
-    sortOrder: number;
-    title: string;
-    type: 'folder';
 }
 
 export const PERMISSIONS_TYPE = {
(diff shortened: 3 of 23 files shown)


zJaaal:
@claude Fixing. Will be in the next commit

dario-daza (approved):

PR #37030, opened 2026-08-11
perf(evergreen-tracks): plan promotions from filtered reads instead of a 79-call full tag scan

### LINKED ISSUE #37028: evergreen-tracks: promoter does a 79-call full tag scan to move one tag — port the reconciler's name= month filter
## Problem

`registry.list_tags()` paginates the **entire** repo — 7,832 tags, `page_size` hard-capped at 100 by Hub, so **79 calls / ~45s** — every time the engine runs, just to move one floating tag. Most of that is noise: 1,946 `*SNAPSHOT` branch-build tags and 677 `*_lts` tags. The promoter needs a handful of GA CalVer tags and three marker names.

Not urgent — it works, and 79 < the 180-req/60s per-IP Hub budget. But:

- The promoter has **no 429 handling at all**. One rate-limited page mid-scan is a hard failure and (since #37027) a Slack page. The reconciler already carries retry/backoff because it hit this.
- The full scan is O(tags) against a hard 100/page ceiling. It stops fitting the budget at ~18,000 tags. We're at 7,832 and the SNAPSHOT tags churn per branch.

## The technique already exists in our own code

`kubernetes/evergreen-tracks-reconciler/reconcile.py::resolve_track` in dotCMS/infrastructure-as-code solved this exact problem — its docstring quantifies the old cost as *"paginating all ~7,700 tags (79 calls, ~45s)"*. The Hub tags endpoint accepts a `name=` **substring** filter, so it asks only for the month the image was pushed in:

```
1. GET /tags/<track>          -> digest + push month     (1 call)
2. GET /tags?name=<YY.MM>     -> that month only          (1 call)
3. miss -> walk back a month, bounded, then full scan
```

Verified against the live API today:

| query | tags returned | pages |
|---|---|---|
| full listing | 7,832 | **79** |
| `name=26.08` | 6 | 1 |
| `name=26.07` | 27 | 1 |
| `name=26.06` | 28 | 1 |

A month is always one page. `page_size` above 100 is silently ignored, so there is no lazier fix.

## Applying it to the promoter

`_state()` currently derives four things from the full list. Each has a bounded query:

- **GA release candidates** — `name=<YY.MM>` for the months covering the threshold window (trailing = 28d, so current + 1–2 previous).
- **`<version>_tainted` markers** — free, they share the version's month prefix and land in the same filtered page.
- **`<track>_hold` markers** — one `name=_hold` call (confirmed working; returns 0 today, and all three markers 404 individually).
- **Current version per track** — the reverse digest lookup, which is the expensive part. Exactly what `resolve_track` does in 2 calls.

**~6–7 calls instead of 79**, ~45s → ~3s.

Small extra win: `cmd_promote` builds `TrackState` for all three tracks and resolves `_current_version` for each *before* `--tracks` filters them. Free against a prefetched dict, wasted work once lookups are lazy — resolve only the requested tracks.

## Why this is not a one-liner

The fast path needs the fallbacks or it trades call count for wrong tag moves:

1. **Reverse-lookup miss → `current_version=None`**, which makes `planner.plan` skip its forward-only guard entirely. Needs the bounded month walk-back *and* the terminal full scan, not just the fast path.
2. **Quiet month or all-candidates-tainted** — must keep walking back rather than concluding "no eligible release".
3. **The two paths must agree.** The reconciler hit precisely this: see `_pick_version`, where a version's suffixed and un-suffixed spellings straddling a page boundary made the fast and slow paths disagree and re-pin each other forever.

This engine mutates production tags, so it wants its own PR and tests rather than a rider on the #37025 outage fix.

Related: #37025 / #37027 (Hub auth fix — unrelated cause, same read path).

sfreudenthaler:
> **Stacked on #37027** (the Hub auth outage fix). Review that one first — this branches off it, so the diff here is only the efficiency work.

## Problem

Moving one floating tag paged the entire repo. `dotcms/dotcms` has 7,832 tags and Hub caps `page_size` at 100 (250 and 1000 are silently ignored), so every run was **79 calls / ~45s** — against a Hub budget of **180 requests / 60s per IP**. The promoter has no 429 handling, so a single rate-limited page is a hard failure and, since #37027, a Slack page.

Most of that listing is noise: 1,946 `*SNAPSHOT` branch builds, 677 `*_lts`. Promotion needs a few GA CalVer tags and three marker names.

## Approach

Same technique the in-cluster reconciler already uses in dotCMS/infrastructure-as-code (`reconcile.py::resolve_track`) — Hub's `name=` **substring** filter:

- **one call per requested track** → the `<track>` digest, and its `<track>_hold` marker for free (both match the track name as a substring)
- **one call per month of history** → that month's GA releases *and* their `<version>_tainted` markers, which share the version's month prefix

A month is one page (6–28 tags).

`--tracks` is now resolved *before* any registry read, so the release pipeline's latest-only promote no longer pays to resolve `standard` and `trailing`.

## Measured against the live registry

| path | before | after |
|---|---|---|
| daily `standard,trailing` | 79 calls | **4** |
| `--tracks latest` (release pipeline) | 79 calls | **2** |
| all three tracks | 79 calls | **5** |
| CI plan job wall-clock | 29s | **2.9s** |

Plans are byte-identical to the full scan's — `standard -> 26.07.28-01`, digest `c2526c56…`, matching the authenticated full-scan run [31534540024](https://github.com/dotCMS/core/actions/runs/31534540024). Verified in CI on this branch: [31544087963](https://github.com/dotCMS/core/actions/runs/31544087963) (`read registry state in 4 filtered call(s)`), cancelled at the approval gate so nothing applied.

## The correctness argument worth reviewing

Everything rests on fetching months **newest-first and contiguously**.

`planner.plan`'s forward-only guard only bites when a track already sits on something *newer* than the newest eligible release. Any such version lives in a month at or above the one the walk stopped on — so it has necessarily already been read. A track on something *older* resolves to "move forward", which is the same decision the full listing produces. **Nothing here can move a track backwards that the full scan would have held.**

The tempting "optimization" that breaks this is starting the walk at the threshold cutoff month instead of today (why read months newer than the cutoff?). `test_promote_narrow_reads_still_refuse_to_move_a_track_backwards` fails on exactly that mutation.

I originally also ported the reconciler's push-month lookup to name each track's current version. Mutation testing showed removing it changed no promotion outcome — the argument above is why it was dead weight — so it's gone, along with the `Tag.month` plumbing it needed.

## Fallback

Returns to the full listing when the walk window holds nothing promotable, so a broken read path surfaces as the expensive *right* answer rather than a quiet "no moves" stall — which is precisely how tracks would silently stop advancing. `admin` deliberately keeps the full scan: rare, human-triggered, and it addresses arbitrarily old versions.

## Tests

73 pass (68 existing unchanged + 5 new). The new ones were mutation-tested — each of these breaks at least one:

| mutation | caught by |
|---|---|
| skip months newer than the cutoff | backwards-move + walk-back tests |
| never walk back a month | 3 tests |
| drop the empty-window fallback | fallback test |
| stop month-filtering | 3 tests |
| ignore `--tracks` when reading | scoping test |

`ruff` is not wired into CI for this package; I checked my changes are net-neutral against the baseline anyway (21 findings before and after, same rules, shifted line numbers).

Closes: #37028

This PR fixes: #37028

### DIFF
--- a/.github/actions/core-cicd/evergreen-tracks/src/evergreen_tracks/cli.py
+++ b/.github/actions/core-cicd/evergreen-tracks/src/evergreen_tracks/cli.py
@@ -10,7 +10,7 @@ import logging
 import os
 import sys
 
-from .calver import newest, parse_release
+from .calver import age_days, newest, parse_release
 from .executor import delete_tag, hub_login, point_tag
 from .markers import TRACKS, held_tracks, hold_tag, tainted_versions, taint_tag
 from .planner import TrackState, plan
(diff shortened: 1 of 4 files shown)


wezell (approved):
approving this rather than #37027

Issue #36988, opened 2026-08-10
Experiments API gaps for the standalone portlet: archive/restore, PATCH side effects, goal operators, variant cap


oidacra:
## Description

The new Experiments portlet (#36987) turns `/experiments/new` into a routed, autosaving Configure screen and adds list actions the current backend cannot support. Four API gaps stand in the way — all other mutations already exist (create, PATCH, start, end, cancel, archive, delete, variants, promote, results).

### 1. Archive from DRAFT + restore/unarchive

Today `archive` is only allowed from `ENDED`, and no unarchive/restore method exists at all. The approved design archives from `DRAFT` and `ENDED`, and restores an `ARCHIVED` experiment back to `DRAFT`. Add both transitions to `ExperimentsAPIImpl` + `ExperimentsResource`.

**Frontend gating note:** the matrix `AllowedActionsByExperimentStatus` (`core-web/libs/dotcms-models/src/lib/dot-experiments-constants.ts:256-287`) is read by the existing UVE list menu. Do **not** widen the existing `archive: [ENDED]` key — the new transitions get **new keys** (`archiveDraft`, `restore`) so the old UI's menu cannot change.

### 2. PATCH side effects (autosave blocker)

Every `PATCH /api/v1/experiments/{id}` currently overwrites `targetingConditions` with `[]`: `ExperimentForm` defaults it to a non-null empty list, so the resource's `!= null` guard always fires, creating an empty Rule + ConditionGroup **on every call**. The new screen autosaves on edit, so this would create one Rule per keystroke pause. Fix so an absent `targetingConditions` leaves the stored value untouched.

Related, document-only: `trafficAllocation` can never be PATCHed to `0` (`if (getTrafficAllocation() > 0)`) — harmless, the design's minimum is 1.

### 3. Goal operators: validate, then extend

The `Operator` enum has only `EQUALS` / `CONTAINS` / `EXISTS`, and operators are **never validated**: an unsupported operator is silently accepted and then silently ignored, producing a goal that never converts with no error anywhere. Validate on write (400 on unsupported). Then extend with the operators the design offers — `STARTS_WITH`, `ENDS_WITH`, `REGEX` — wired through the metric/query layer so they actually evaluate. If extending is too large for this issue, validation+reject still lands here (it is the bug) and the new operators split out; the UI then trims to the supported set.

### 4. Variant cap enforcement

[...]

Issue #36987, opened 2026-08-10
Experiments v2 — Part 2: Build the new Experiments UI


oidacra:
## Description

Today experiments exist only inside the page editor: the entry point is the UVE nav item (`experiments/${page.identifier}`), the list is per-page, and the create form is a drawer with only *name* and *description* because the page comes from the route context. A screen listing every experiment on the site has never existed, and there is no way to create an experiment by choosing a page.

This EPIC creates a new site-wide **Experiments portlet** — list, routed creation/configuration (`/experiments/new` is the Configure screen with an empty draft), and results — and modernizes the stack (NgRx Signal Store + Events plugin, replacing ComponentStore). Everything new works **without the old flow ever breaking**: both coexist in the same build, and the one place they overlap (the UVE experiments entry point) is switched by `FEATURE_FLAG_EXPERIMENTS`, off by default.

### Strategy

1. **Build the new portlet complete, then migrate and delete the old one.** The old per-page UI stays untouched throughout; a final migration issue points UVE at the new flow and deletes the old tree.
2. **Additive while both coexist.** New code lives in a new `site-wide/` subtree of `libs/portlets/dot-experiments/portlet/`. Only three existing files gain one line each (lib `index.ts`, `app.routes.ts`, `Language.properties`). Deliberate exceptions: the UVE-integration issue (behind the flag) and the migration.
3. **Two switches.** The feature flag switches UVE entry-point behavior; portlet **registration** gates the portlet itself — NOT registered for customers in this scope (a flag cannot hide a nav item: `/api/v1/menu` does not read flags and `MenuGuardService` bounces unregistered routes). Manual `cms_layouts_portlets` row on dev/QA for testing.
4. **Built against the endpoint that exists today.** The list ships with a documented client-side interim; a swap issue replaces it with the server-side contract when #36823 lands.
5. **One frontend issue per screen**, each with its own Jest specs; E2E is a dedicated issue (no experiments E2E exists today, and the migration must not run without that safety net).

### Functional parity

The design prototype describes **less** functionality than the current UI has. Twelve existing behaviors (row actions, inplace variant rename, copy-variant URL, page-lock awareness, scheduling time + duration bounds, results refresh, 10-session gate, promoted state, and more) are ported explicitly — each is an acceptance criterion in the issue that owns its screen. Nothing the current UI does is dropped silently.

### Issues

| Order | Issue | Area | Depends on |
|---|---|---|---|
[...]

Issue #36823, opened 2026-07-30
Add pagination, sorting, filtering and permission checks to the Experiments list endpoint


oidacra:
## Description

The new standalone **Experiments portlet** needs a site-wide, paginated experiments list. This issue puts the existing list endpoint onto dotCMS's canonical pagination stack, closes a security hole in it, and fixes two latent SQL bugs that the portlet's own UI hits immediately.

### What already works

`GET /api/v1/experiments` **already returns every experiment on every page and every site** — `pageId`, `name` and `status` are all optional query params, and `ExperimentsFactoryImpl.java:37` starts from `SELECT * FROM experiment WHERE 1=1`. A Postman test asserts exactly this (`listExperimentsNoFilter_shouldIncludeAll`, `Experiments_Resource.postman_collection.json:6312`).

So there is no new read path or persistence design to invent. What is missing is everything the portlet's toolbar and table need on top of it.

### What is missing

| # | Gap | Evidence |
|---|---|---|
| 1 | **Pagination** — no `page`/`per_page`, no `LIMIT`/`OFFSET` in the SQL | `ExperimentsFactoryImpl.java:94-131`, `AbstractExperimentFilter.java:20-24` |
| 2 | **Sorting** — no `orderby`/`direction`, no `ORDER BY`. Result order is whatever Postgres returns, i.e. **non-deterministic across calls** | same |
| 3 | **Total count** — no `count()` on `ExperimentsAPI`/`ExperimentsFactory`. The response's `pagination` field is **always `null`** because the views use the single-arg `ResponseEntityView(entity)` ctor | `ResponseEntityView.java:66-73`, `ResponseEntityExperimentView:8-10` |
| 4 | **Site scoping** — no `host`/`site` column on the `experiment` table and no `siteId` on the filter | `postgres.sql:2531-2547` |
| 5 | **Permission filtering** — see the security section below | `ExperimentsAPIImpl.java:504-510` |
| 6 | **No `ExperimentsPaginator`** — `com/dotcms/util/pagination/` has 23 paginators, none experiment-related | — |

### 🔴 Security: the list endpoint does no permission filtering

`ExperimentsAPIImpl.list(ExperimentFilter, User)` accepts a `User` and **never uses it**:

```java
// ExperimentsAPIImpl.java:504-510
public List<Experiment> list(ExperimentFilter filter, User user) throws DotDataException {
[...]

PR #37034, opened 2026-08-12
feat(experiments): add Experiments portlet with site-wide list screen

### LINKED ISSUE #36989: Experiments Portlet — Screen 1: portlet base + site-wide List
## Description

This is the issue that makes the portlet exist: the `site-wide/` subtree, the `/experiments` route, and the List screen from the approved design — search, status filter with counts, pagination, sorting, row actions — listing every experiment on the current site. **No backend dependency**: it ships against `GET /api/v1/experiments` as it exists today (see Interim below).

Design: [approved prototype](https://claude.ai/design/p/36aed5a6-ca42-4624-b187-a465a23ce909?file=Experiments.dc.html&via=share), `list` screen.

### Portlet base (additive — three existing files, one line each)

- New subtree `libs/portlets/dot-experiments/portlet/src/lib/site-wide/` with its own `lib.routes.ts` exporting `dotExperimentsPortletRoutes` (`''`, `new`, `:id/configuration`, `:id/results` — the last three are placeholders wired by their own issues).
- Export the new route array from the lib `src/index.ts` (the existing `dotExperimentsRoutes` export untouched; `edit-ema` keeps consuming it).
- Route entry in `apps/dotcms-ui/src/app/app.routes.ts` (`PORTLETS_ANGULAR`, copy the `tags` entry). The path **must** be `experiments`: `MenuGuardService` validates the first URL segment against `/api/v1/menu`.
- **No registration UpgradeTask and no flag guard** — the portlet is reachable only where a `cms_layouts_portlets` row was inserted manually (dev/QA): one row against an existing layout (e.g. Marketing); nothing in `portlet`, `layouts_cms_roles` or `cms_layout`. Roles & Tools on the admin role works too.

### List screen

- **Table copied from `dot-tags`** (`libs/portlets/dot-tags/src/lib/dot-tags-list/`): `p-toolbar` + `p-iconField` search (300 ms debounce), lazy `p-table` + paginator + sort, per-row skeletons, `dot-empty-container` with the `science` icon. Cross-check `dot-users` (same skeleton, newest iteration).
- **Columns**: Experiment (name + description below) · Page (path, monospace) · Goal (type label or `—`) · Variants (count) · Schedule (`Jun 25, 2026 → Jul 9, 2026` / `→ Open` / `Not scheduled`) · Status (tag) · Modified (sortable, default DESC) · Actions.
- **Status tag**: portlet-local `p-tag` + `Record<DotExperimentStatus, severity>` map (pattern: `dot-publishing-status-chip`).
- **Status filter chip**: reuse `DotChipFilterComponent` / `DotFilterListItemComponent` / `CHIP_FILTER_*_PT` from `@dotcms/portlets/content-drive/ui` (precedent: `dot-users-filter-by`). Multi-select, per-status counts, Clear/Done. **Default excludes `ARCHIVED`**.
- **Primary row action** (derived): `status ∈ {RUNNING, ENDED}` → View Results, else Configure. **Hidden until Screens 2 and 3 land** — the new list must not route into the old screens.
- **Row actions — full parity with the current UVE list** (`dot-experiments-list-store.ts:483-640`): Archive/Restore inline icons per the design (archive-from-DRAFT and restore enabled once #36988 lands, via the new `archiveDraft`/`restore` keys), plus a kebab with `cancelSchedule`, `end`, `abort`, `delete`, `pushPublish`, `addToBundle` — all gated by `AllowedActionsByExperimentStatus`, each with confirm + toast. Push-publish/add-to-bundle need `DotPushPublishEnvironmentsResolver` on the route and reuse `dot-add-to-bundle`.
- **Site scoping**: the site selector drives the list; switching site reloads it.

### Store — NgRx Signal Store + Events plugin

- `experiments-list.events.ts` with `eventGroup`; state transitions in `withReducer(on(...))`; HTTP in `withEventHandlers` with Request → Succeeded → Failed triples and `switchMap` (re-trigger cancels in-flight). Components dispatch via `injectDispatch` — no store methods for state changes.
- Errors: `catchError` → `DotHttpErrorManagerService.handle`; load failure → `status: 'error'`, CRUD failure → back to `'loaded'`.
- Reference: `libs/image-editor/src/lib/store/`. Note: the async hook is **`withEventHandlers`** in the installed `@ngrx/signals` — examples naming `withEffects` will not compile.
- **Add an events-pattern section to `libs/portlets/CLAUDE.md`** in this PR, pointing at `libs/image-editor`, so reviewers have a written standard.

### URL state

All navigable state in the URL: `page`, `per_page`, `orderby`, `direction`, `filter`, `status` (repeatable). Hydrate from `route.snapshot.queryParams` **before** the first fetch; write back via effect with `queryParamsHandling: 'merge'`, `null` clearing defaults (a bare `/experiments` is the default view), guarded against `location.path(true)` to avoid redundant history entries. Reference: `dot-content-drive-shell.component.ts:454-494`.

### Interim: client-side list operations (temporary by design)

`GET /api/v1/experiments` with no params returns every experiment, so this issue computes paging, sorting, text/status/site filtering and status counts **client-side**. Acceptable at current volumes; safe because the portlet is unregistered for customers; sidesteps two SQL bugs that only fire on server-side filtering. The follow-up "swap to the server-side contract" replaces this when #36823 lands — the URL param contract is final from day one, so the swap changes where params are applied, not what they are called. Known gap: without #36823's enriched rows the API returns `pageId`, not the page path — resolve titles client-side in one bulk lookup (fallback: show the id). Shape the data-access service so the swap changes only the service and the store's load handler.

Hazard: `dot-experiments.model.ts:204-215` runs `getComputedStyle(document.body)` at import time — stub it in the new specs (lazy-getter fix belongs to the migration issue).

## Acceptance Criteria

- [x] `/experiments` renders the site-wide list for the current site (manual layout-row insert documented in the PR).
- [x] Search filters by name and page path (substring, case-insensitive), debounced 300 ms.
- [x] Status chip shows the 5 statuses with counts for the whole filtered set; default view excludes ARCHIVED.
- [x] Table pages, sorts (default `modDate` DESC) and shows the design's columns + Modified.
- [x] Kebab exposes cancelSchedule / end / abort / delete / pushPublish / addToBundle exactly where `AllowedActionsByExperimentStatus` allows them; each has confirm + toast; the list reloads after.
- [x] Archive/Restore icons per the design (enabled once #36988's transitions exist).
- [x] Filter + page + sort state round-trips through the URL: copying the URL into a new tab restores the identical view; back/forward walks the states; a bare `/experiments` has no params.
- [x] Switching site via the site selector reloads the list.
- [x] The existing UVE experiments UI is untouched: `dotExperimentsRoutes`, the nav item and `dot-ema-shell.component.spec.ts:117-120` all unchanged.
- [ ] `libs/portlets/CLAUDE.md` documents the events-store pattern.
- [ ] Jest specs: store (dispatch events, assert signals; error paths assert `DotHttpErrorManagerService.handle`), list component, filter chip, URL sync (hydration + write-back).

## Priority

High — everything else in #36987 hangs off this.

## Additional Context

- Out of scope: Configure/Results navigation targets (own issues); server-side paging/filtering (swap issue); registration UpgradeTask/starter bump; metrics in the list (`GET /{id}/results` costs 2 CubeJS round-trips + a Monte Carlo per experiment, `@NoCache`, no batch form — the design deliberately shows none).
- E2E for this screen (including deep-links) lives in the dedicated portlet E2E issue.
- Verify: `pnpm nx lint|test portlets-dot-experiments-portlet` · `pnpm nx test portlets-edit-ema-portlet` · `pnpm nx test dotcms-ui --testPathPattern=app.routes` · `pnpm nx format:write` · `pnpm nx build dotcms-ui`

oidacra:
## Summary

Adds the Experiments portlet: a new `/experiments` route rendering a searchable, filterable, sortable and paginated list of every experiment on the current site. This is the issue that makes the portlet exist, and everything else in #36987 hangs off it.

Ships against `GET /api/v1/experiments` as it exists today. No backend dependency, no registration UpgradeTask, no feature flag.

https://github.com/user-attachments/assets/bcf6b51b-7934-45d6-952d-bcf079880253

Closes #36989

## How to read this

37 commits, rebased on `main`. Reviewing them in order is much easier than reading the combined diff — most of the later ones are single-concern fixes found by running the screen.

**The feature**

| | |
|---|---|
| `55723e4` | Move legacy UVE screens under `old/` — 110 files, **zero content change** |
| `662a762` | The portlet: routes, store, list screen, specs |
| `4412dff` | Declare the portlet server-side + Analytics health gate |
| `b6e656e` | Align the list row with the approved design |

**Three DI failures, all invisible to Jest and to the AOT build**

| | |
|---|---|
| `805d1d8` `f8caf74` | Provide `DotPushPublishEnvironmentsResolver` on the route, `DotExperimentsService` on the component |
| `859f51a` | DI smoke test so this class of bug fails in CI instead |
| `8f687c2` | Hold the list until the health check answers, so there is no layout flash |

**Shared UI** (see "Scope beyond the issue")

| | |
|---|---|
| `26aad63` `7cfc299` | Align the status filter with content-drive; move the chip primitives into `@dotcms/ui` |
| `8b82736` `a9cd41b` | Move panel styling into the theme preset, then make it global |
| `362ae0e` | **Security**: stop `DotHighlightPipe` injecting unescaped text as live markup |

**Store shape**

| | |
|---|---|
| `5faf36d` `9264d75` | Split page/API events, extract helpers to `util/`, move URL write-back into the store |
| `c223db5` | Debounce the search with Angular 22's `debounced()` |
| `2cc5718` | Adopt the shared `ComponentStatus`, converted to `as const` |

**Behaviour and layout, driven by running it**

| | |
|---|---|
| `f79a24b` `dddc0f8` | Load-error state with retry; make the loading skeleton actually render |
| `bc35ab4` `e5ea193` `2cc5718` | Actions cell is one kebab, styled like content-drive's |
| `76e8e58` | Status filter starts empty, archived opt-in |
| `4e7c853` `7ffaad0` `d5aa74a` `5d0fc66` | Column widths bounded to their content; table laid out `fixed` |
| `956f9e9` | Goal filter, generalising the chip filter behind it |
| `e53d0a3` `ee7af1e` | Search covers the description; clear control in the box |
| `920cfee` | Sorting on Experiment, Page, Goal, Schedule, Status |
| `5e6ecb3` | Paginator matches content-drive |
| `f7822c7` | Three defects from review (below) |
| `3bec27c` | Split the two empty states, centre them, soften row changes |

## Architecture notes

**Events store, split by source.** `dotExperimentsListPageEvents` is what the page asks for; `dotExperimentsApiEvents` is what the backend answered. The page dispatches only the first, handlers raise only the second, so the two halves of an async flow cannot be confused. Page events read as commands (`loadExperiments`, `archiveExperiment`) rather than `…Requested`. Both fold into one `withReducer`, so reading it tells you who caused each transition.

Handlers **return** their events via `mapResponse`; `withEventHandlers` dispatches whatever they emit, so no `Dispatcher` is injected there. Verified against the installed 21.1.1 source: auto-dispatch skips events carrying `SOURCE_TYPE`, which is stamped only on incoming events as a loop guard.

The async hook is `withEventHandlers` — `withEffects` does not exist in `@ngrx/signals` 21.1.1 and will not compile. Written down in `libs/portlets/CLAUDE.md` with the rest of the pattern.

**Site scoping fails closed.** `DotExperiment` carries no host field, only `pageId`. One bulk `DotContentSearchService` query resolves each page's `url` for the Page column and `host` for the site filter. An experiment whose page cannot be resolved is **excluded** rather than shown, so an unresolvable page can never leak another site's experiment into the list.

**The store owns the whole URL contract.** It parses on entry and on popstate, and writes back through `toQueryParams`, the inverse of `parseViewState`. Serialization used to live in the component, which split one contract across two files and invited the halves to drift.

**Client-side list operations are interim by design.** The endpoint returns the full set, so paging, sorting, filtering and counts are computed client-side until #36823 lands the server-side contract. The URL param contract is final from day one, so the swap changes only `DotExperimentsService.getAllUnfiltered()` and the store's load handler.

**Analytics health gate.** Mirrors what the legacy UVE routes get from `AnalyticsAppGuard`: only `OK` passes. The check runs before the first fetch and the list is never requested when it fails — including on a later site switch. A pending gate is its own render state, otherwise the list paints and is then replaced, which read as a flash on every entry.

**No Enterprise licence gate.** The legacy list gates push-publish and add-to-bundle behind `hasEnterpriseLicense`; deliberately not carried over.

## Why `portlet.xml` changed

The issue's setup note said a `cms_layouts_portlets` row was enough. It isn't. `MenuHelper.getMenuItems()` resolves every layout portlet id through `PortletAPI` and silently skips what it cannot find. With no `experiments` portlet declared, `findPortlet` returned null, the menu never carried the entry, `MenuGuardService` rejected the route and redirected to the first portlet. Declaring it is what makes `/experiments` reachable at all.

This is a declaration, not a registration — still no UpgradeTask and no starter change, so the portlet stays invisible to customers until someone adds it to a layout by hand.

## Defects found in review, fixed in `f7822c7`

All three were verified in the code before changing anything, and each fix has a test confirmed to fail without it.

1. **The page lookup truncated the list, silently.** It limited to the number of pages, but Elasticsearch holds one document per identifier *and* language, so a multilingual site returned several documents per page and the response truncated at HTTP 200. Truncated pages resolved to nothing, the site filter fails closed, and their experiments vanished — with a total derived from the same set agreeing with the short list. Indistinguishable from reality on screen. The limit now allows headroom for languages, and a shortfall between resolved and requested ids is reported instead of passing silently.

2. **Retry after a failed health check bypassed the gate.** `healthCheckFailed` leaves `healthStatus` null, so `isMisconfigured` stayed false and the generic error branch rendered; its Retry dispatched the list, which carries no health guard. Worse, `$isLoading` keys off that null, so the table sat on skeletons forever even once the list came back. Retry now re-runs the health check.

3. **A list where no experiment had a resolvable `pageId` stayed on LOADING** — the lookup returned `EMPTY`, so nothing settled the status. Defensive, since `pageId` is typed non-optional, but it is one line standing between a bad payload and a permanent skeleton.

## Empty states

"Nothing here" and "nothing matched" are different situations and now get different copy. Previously a list emptied by a Status filter still read "No experiments yet", which is wrong and offers nothing to do about it; the filtered case now says so and carries a **Clear filters** action. Which one shows keys off the search term and both chip selections together, so clearing one of three still reads correctly.

The message also used to render *inside* the table, so it sat in a short band under the header with the table's bottom border cutting across it while the rest of the viewport stayed blank. It now replaces the table and centres in the space the table would have filled — which also removed a paginator reading "Page 0" against no records.

Row changes fade in over 140ms instead of swapping in a single frame, dropped entirely under `prefers-reduced-motion`. Content-drive has no equivalent, so this is new rather than parity.

## Scope beyond the issue

Each of these was requested during implementation; flagging them so they can be split out if preferred.

1. **Legacy screens moved to `old/`** — this portlet ultimately replaces the per-page UVE screens, so the new code takes the canonical names. Every legacy file moved with **zero content change** (`a87445b` reverts formatting that `format:write` had dragged in). `dotExperimentsRoutes` keeps its symbol name and barrel export, so `edit-ema` is untouched.
2. **Chip-filter primitives moved to `@dotcms/ui`** — they lived in `@dotcms/portlets/content-drive/ui`, so every portlet wanting a filter chip imported from another portlet's UI lib. Seven consumers across four portlets already did.
3. **Panel styling moved into the theme preset, then made global** — audited against all 24 popover and 13 listbox consumers. This restyles **every** popover and listbox in the app, not just filter panels; the two UVE selectors setting `[unstyled]="true"` needed adjusting.
4. **A Goal filter** (`956f9e9`) that no AC asks for. Fully built, tested and URL-synced. Worth adding to #36989 for traceability, or splitting out.
5. **`DotHighlightPipe` XSS fix** (`362ae0e`) — pre-existing, in `libs/ui`, unrelated to this feature. Details below.

## Security fix

`DotHighlightPipe` interpolated author-supplied text into an HTML string and handed the result to `bypassSecurityTrustHtml`, which switches Angular's sanitizer off. Verified before fixing rather than assumed: `Alpha` rendered a live `` element with its handler intact, through the pipe's own documented `[innerHTML]` usage.

Every part is now escaped before being wrapped, leaving the highlight span as the only markup. Its one existing consumer, `dot-language-variable-selector`, renders language variable keys through it, so the same input reached it. Four regression tests, each verified to fail with the escaping removed, including one asserting against the rendered DOM rather than the string.

## Acceptance criteria

- [x] `/experiments` renders the list for the current site
- [x] Search filters name, **description** and page path, case-insensitive, debounced 300 ms
- [x] Status chip shows 5 statuses with counts over the filtered set, independent of selection; archived excluded until explicitly picked
- [x] Table pages and sorts, `modDate` DESC by default, design columns plus Modified
- [x] Kebab exposes cancelSchedule / end / abort / delete / pushPublish / addToBundle exactly where `AllowedActionsByExperimentStatus` allows, each with confirm and toast, list reloads after
- [x] Archive renders per current transitions; Restore renders disabled (no `restore` key or service method until #36988)
- [x] Filter, page and sort round-trip through the URL; defaults serialise to `null` so a bare `/experiments` has no params
- [x] Switching site reloads the list, keeping search/sort/status and resetting page to 1
- [x] Existing UVE UI untouched: `dotExperimentsRoutes` unchanged, `portlets-edit-ema-portlet` passes
- [x] `libs/portlets/CLAUDE.md` documents the events-store pattern
- [x] Jest specs: store, list component, filter chip, URL sync
- [~] `dotExperimentsPortletRoutes` exports `''` only — `new`, `:id/configuration` and `:id/results` are omitted rather than stubbed. A placeholder needs a component, and routing to the old UVE screens is forbidden by AC10 and would break anyway (they need the `:pageId` shell context). Omitted, an unimplemented deep link falls through to the app catch-all instead of rendering a blank screen.
- [~] **The Actions cell is one kebab.** The design leads it with "View Results" and "Configure", but both screens land with #36990+, so a control that cannot go anywhere is noise. This changed during review — an earlier revision rendered the primary action disabled; it is now omitted entirely, matching content-drive, where the kebab is the only affordance and reveals on row hover.

Two deviations from the issue text, both decided during implementation:

- The status filter applies each toggle immediately rather than behind a Clear/Done footer. No `dot-chip-filter` consumer in the app has a footer.
- Sorting is enabled on Experiment, Page, Goal, Schedule and Status, not only Modified.

## Setup for QA

The portlet is declared but not registered, so it must be added to a layout by hand. Easiest path, no SQL: **Roles & Tools → admin role → add Experiments to a layout** (e.g. Marketing). Or:

```sql
INSERT INTO cms_layouts_portlets (id, layout_id, portlet_id, portlet_order)
VALUES (gen_random_uuid()::text, '<layout_id>', 'experiments', 10);
```

Then open `/dotAdmin/#/experiments`. Requires a rebuilt image, since `portlet.xml` is a webapp resource.

If Analytics is not configured you will get the "Analytics not Enabled" notice instead of the list — that is the health gate working. The repo has a stack for it:

```bash
cd docker/docker-compose-examples/experiments
./start-experiments.sh --experiments-only
```

`GET /api/v1/experiments/health` returning `OK` is the single indicator that the list will render.

## Verification

| Check | Result |
|---|---|
| `pnpm nx lint portlets-dot-experiments-portlet` | pass |
| `pnpm nx test portlets-dot-experiments-portlet` | 434 tests, 39 suites |
| `pnpm nx test ui` | pass |
| `pnpm nx test portlets-edit-ema-portlet` | pass |
| `pnpm nx test portlets-content-drive` | pass |
| `pnpm nx test dotcms-ui --testPathPatterns=app.routes` | pass |
| `pnpm nx build dotcms-ui` | pass |
| `nx affected -t lint --base=origin/main --exclude=tag:skip:lint` | 32 projects, pass |
| `nx affected -t test --base=origin/main` | 29 projects, pass |
| `nx format:check --base=origin/main` | pass |

All re-run after rebasing onto `main`, not before.

The last three are the commands CI itself runs. A single-project lint had been hiding an import-order error in `app.routes.spec.ts`, which is what failed the first run on this branch — everything else in that run was cancelled by fail-fast rather than failing on its own.

## Notes for the reviewer

- Every fix in `f7822c7`, and the sort comparators, the URL write-back, the fixed table layout and the chip filter toggle, are **mutation-checked**: the change was reverted locally and the new tests confirmed to fail. Tests that pass without the code they cover are worse than none.
- `apps/dotcms-ui/src/app/app.routes.spec.ts` did not exist, so the issue's `app.routes` verify command was silently finding zero tests. Added one pinning what AC3 needs.
- Jest in this repo replaced `--testPathPattern` with `--testPathPatterns`; the issue's verify command uses the old flag.
- `archived` is defined twice in `Language.properties` (line 345 `archived=archived`, line 5774 `archived=Archived`). Pre-existing and shared with the current UVE filter — left alone.
- `abort` calls `cancelSchedule` in the store. Pre-existing behaviour in the legacy list, replicated for parity, not fixed. The specs assert the dispatched events rather than the HTTP call, so they keep passing when a real abort endpoint arrives.
- `LISTBOX_OPTION_HEIGHT = 40.6` is the measured row height and is now a shared constant rather than two hardcoded copies. Review has previously argued for an integer here; keeping the measured value was a deliberate call.


oidacra:
Thanks — the critical one was real, and worse than "very likely". Confirmed against the installed PrimeNG rather than inferred: `createLazyLoadMetadata()` (`primeng-table.mjs:2313`) always includes `sortField`/`sortOrder`, and `onPageChange` (`:1549`) emits through it. So clicking page 2 dispatched `pageChanged({page:2})` and then a sort that had not changed, which reset the page to 1 — pagination never advanced, and `?page=3` was undone by the table's own first render.

Fixed in `eedb64f`: `sortChanged` is dispatched only when the field or direction actually differs from the store. Two tests — a pagination event carrying the unchanged sort dispatches no sort, and a real direction change still does. The first fails without the guard.

Your point about why the tests missed it is the useful part. The spec comment at line 828 was written by me to explain away noise in an assertion, and it was in fact describing the bug. That is the second time on this PR that a comment I wrote to work around a symptom was pointing at a defect.

On the two notes:

**Page path search** — valid, fixed in `a260dec`. The column renders `resolvePagePath`, which falls back to the raw `pageId`, while the search compared the url alone, so a row displaying an id could not be found by typing it. Both go through the same helper now. It also made the comment above that predicate untrue, which is the part that bothered me more.

**Page-lookup headroom** — agreed it is a conscious trade-off, and left as is. The limit cannot be derived without knowing the language count, so the honest options were a generous limit or an extra round trip to count first. The shortfall warning is there to make the wrong guess visible instead of silent. #36823 removes the second request entirely, which is the actual fix.

adrianjm-dotCMS (approved):

PR #37037, opened 2026-08-12
feat(evergreen-tracks): add a tracks input so latest can be moved without cutting a release

### LINKED ISSUE #37036: evergreen-tracks: no supported way to move `latest` outside a release cut
## Problem

When the 2026-08-12 release cut failed to move `latest` (#37025 — Hub started refusing anonymous tag pagination), there was **no supported way to repoint it**:

- `cicd_evergreen-tracks-promote.yml` hardcodes `--tracks standard,trailing` and has no input to change that.
- `cicd_evergreen-tracks-admin.yml` only does taint / untaint / hold / release-hold.
- Re-running the failed release job replays the pre-fix commit, so it just fails again.

The only route left was **cutting a throwaway release to repoint a tag** — which is what we did (`26.08.12-02`, code-identical to `-01` apart from the CI fix). That works, but rebuilding artifacts, redeploying to Artifactory, regenerating release notes and publishing a changelog entry to dev.dotcms.com is a lot of blast radius for moving one floating tag, and it puts a phantom release in front of customers.

## Proposal

Add a `tracks` dispatch input to the promote workflow, defaulting to `standard,trailing`.

- The daily cron and every existing dispatch behave **exactly** as today — a `schedule` event carries no inputs at all, which the `||` fallback covers.
- `tracks=latest` becomes the break-glass path, behind the same `evergreen-tracks-apply` required-reviewer gate as any other manual move.
- The apply job still holds the `evergreen-tracks-registry` lock, so it cannot race the release pipeline's latest-promote. If a GA cut repoints `latest` while an approval sits, the existing drift check makes apply refuse rather than fight the release pipeline — the correct outcome.

## Also fixes a latent bug

The moved-summary grep matches only `^(standard|trailing)`:

```
MOVES=$(printf '%s\n' "$CURRENT_PLAN" | grep -E '^(standard|trailing) (\(held\) )?-> ' ...)
```

So a `latest` move would apply correctly but report `moved=false` and post **no Slack notice**. Silent success is the one outcome this workflow is built to avoid. Widened to include `latest`.

Related: #37025, #37028.

sfreudenthaler:
Adds a `tracks` input to the promote workflow so `latest` can be repointed without cutting a release.

## Why

When the 2026-08-12 release cut failed to move `latest` (#37025 — Hub started refusing anonymous tag pagination), there was no supported way to fix it. The promote workflow hardcoded `--tracks standard,trailing`, the admin workflow only does taint/hold, and re-running the failed release job replays the pre-fix commit. We ended up cutting `26.08.12-02` — code-identical to `-01` apart from the CI fix — to move one floating tag. That works (verified below), but it rebuilds artifacts, redeploys to Artifactory, regenerates release notes and publishes a changelog entry to dev.dotcms.com, and leaves a phantom release in front of customers.

## What changes

| | before | after |
|---|---|---|
| daily cron | `standard,trailing` | `standard,trailing` (unchanged) |
| dispatch, no input | `standard,trailing` | `standard,trailing` (unchanged) |
| dispatch `tracks=latest` | not possible | moves `latest`, behind the approval gate |

A `schedule` event carries no inputs at all, so the `|| 'standard,trailing'` fallback is what keeps the cron on its current behaviour — same pattern the existing `repo` / `*_days` inputs already use.

`latest` stays owned by the release pipeline; this is explicitly the break-glass path, and the header comment says so. The apply job still holds the `evergreen-tracks-registry` lock, so it cannot race the release pipeline's latest-promote — and if a GA cut repoints `latest` while an approval sits, the existing drift check makes apply refuse rather than fight it, which is the outcome we want.

## Latent bug fixed along the way

The moved-summary grep matched only `^(standard|trailing)`, so a `latest` move would have applied correctly but reported `moved=false` and posted **no Slack notice**. Silent success is the one outcome this workflow exists to prevent. Now `^(latest|standard|trailing)`.

Worth flagging for review: this bug was invisible until the input existed, and it would have shown up as "we moved latest and nobody heard about it" — the same shape as the original incident.

## Verification

- YAML parses; `plan` and `apply` resolve `TRACKS` to the identical expression, which they must — `apply` diffs its re-derived plan against the approved one, so any divergence would fail every run.
- All three `--tracks` call sites now read `$TRACKS` (plan, apply's re-plan, apply's mutation).
- Empty/cron path resolves to `standard,trailing`.
- The engine plans a `latest`-only scope correctly against the live registry in 2 calls.
- End-to-end evidence that the underlying path is sound: release `26.08.12-02` ([run 31628506475](https://github.com/dotCMS/core/actions/runs/31628506475)) moved `latest` on both repos in 2 filtered calls each — `dotcms/dotcms` → `07e6e29f…`, `dotcms/dotcms-dev` → `982dc0fd…`, both matching `26.08.12-02`.

Closes: #37036

This PR fixes: #37036

### DIFF
--- a/.github/workflows/cicd_evergreen-tracks-promote.yml
+++ b/.github/workflows/cicd_evergreen-tracks-promote.yml
@@ -1,6 +1,12 @@
 name: evergreen-tracks-promote
-# Advances the floating `standard` / `trailing` tags. `latest` is NOT touched here —
-# the release pipeline moves it on every GA cut (cicd_6-release.yml, --tracks latest).
+# Advances the floating `standard` / `trailing` tags. `latest` is owned by the release
+# pipeline, which moves it on every GA cut (cicd_6-release.yml, --tracks latest), so the
+# daily cron deliberately leaves it alone.
+#
+# The `tracks` dispatch input can still aim this workflow at `latest` — the break-glass
+# path for when a release cut moved everything EXCEPT latest (e.g. the 2026-08-12 Hub
+# 403, #37025), where the alternative was cutting a throwaway release to repoint a tag.
+# It is dispatch-only and lands behind the same approval gate as any other manual move.
 #
 # Two triggers, one engine:
 #   * SCHEDULE (daily, 06:00 ET) — runs plan → apply unattended. No human in the loop.
@@ -43,6 +49,10 @@ on:
         description: 'Min age (days) for the trailing track'
         required: false
         default: '28'
+      tracks:
+        description: 'Tracks to move (comma-separated: latest,standard,trailing)'
+        required: false
+        default: 'standard,trailing'
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


dcolina (approved):
LGTM. 👍🏽

PR #37039, opened 2026-08-12
feat(content-drive): add configuration steps for bulk actions that need input

### LINKED ISSUE #36979: Workflow Center: design the configuration step for actions that need input before running
### Define the Problem

The Content Drive **Action Center** dialog has two screens today:

1. **Action list** — pick a quick action (Publish, Unpublish, Archive, Delete, Unarchive, Lock, Unlock) or a workflow action, each showing how many of the selected items it applies to.
2. **Preview** — a table of the contentlets the action will run on, one checkbox per row, so the set can be trimmed before committing. Execute fires from here.

That flow only works for actions that need nothing beyond the selection. **There is no screen for an action that needs the user to supply something first**, so every such action is currently disabled in the dialog:

| Action | What it needs before it can run | State today |
|---|---|---|
| **Move** | A destination folder — browse or search the tree, pick one | Disabled (`requiresInput`) |
| **Add to Bundle** | A target bundle — pick an existing one or create a new one | Visible but permanently disabled, with a "not available yet" hint |
| **Push Publish** | Environment(s), publish date/time, expire date/time, push-publish filter, "I want to" (publish / expire / both) | Disabled (`requiresInput`) |
| **Assign / Comment** | An assignee (user or role) and/or a workflow comment | Disabled (`requiresInput`) |

The dialog folds all of these into a single `requiresInput` flag and greys the row out with a tooltip. A user selecting 20 items and reaching for **Move** is told it can't be done in bulk here — so they leave Content Drive and do it elsewhere, which is exactly the problem the Action Center was built to solve.

**These are not edge cases.** Move and Push Publish are among the most common bulk operations editors perform, and Add to Bundle is the standard path to shipping a set of content between environments.

### Who's Facing the Problem?

- **Content editors** doing bulk housekeeping in Content Drive — reorganising folders, batching content for release. They hit a disabled row and have to fall back to the legacy contentlet screen or move items one at a time.
- **Publishers / release managers** assembling a bundle from a Content Drive selection. Add to Bundle is the one path that would make this a single gesture, and it is the action that has never worked here.
- **Anyone in an approval workflow** whose scheme's actions are assignable or commentable — those actions disappear from bulk entirely, so the Action Center is silently less capable for their content type than for someone else's.

### Screenshot or Video

https://github.com/user-attachments/assets/07be86bc-5f13-4e4f-b2ee-4fb8259bb4a2

### Relevant Links and Resources

- **Design prototype (current Action Center):** https://aistudio.google.com/apps/a19812ee-4919-4dc3-a954-3a379e8c0c68?showAssistant=true&showPreview=true
- **Epic:** #33999 — [EPIC] CD: Workflow Actions
- **Action Center implementation:** #36821 (initial), #36844 / PR #36921 (Lock & Unlock)
- **Where the gate lives in code:** `core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dialogs/dot-content-drive-action-center/` — `requiresInput` is set in `utils/action-center.ts` and disables the row in the template
- **Existing single-item equivalents worth borrowing from**, rather than inventing new patterns: `DotWorkflowEventHandlerService` (push-publish + assign/comment dialogs), `DotAddToBundleComponent` (bundle picker, single identifier today), and the Move dialog's folder picker

### Assumptions & Initiation Needs

**Deliverable:** design only — mockups and an interaction spec. Implementation will be tracked separately once the design is approved, so engineering isn't blocked behind design review on this ticket.

**Scope:** all four action groups in the table above. They share one `requiresInput` flag in code, so a pattern that covers the set unblocks them together. Push Publish is deliberately included even though it is the largest form — designing the pattern around only the simple pickers risks a layout that won't stretch to it.

**Open for the designer to decide** — the ticket deliberately does not prescribe these:

- **Where the step sits.** `pick → configure → preview → execute` keeps the preview as the last thing seen before committing; `pick → preview → configure → execute` lets the user trim the selection before filling in a form. There may be a better answer than either. A recommendation with rationale is what's wanted.
- **One screen or per-action screens.** A folder picker and the push-publish form have very different shapes; whether they share a frame or each get their own is a design call.
- **How the item count is carried through.** Both existing screens report how many contentlets are affected; the configuration step needs to keep that visible without repeating the preview.

**Constraints the design has to respect:**

- It lives inside the shell's shared dialog — one dialog, one open/close path. The header is owned by the shell and retitled per step via drill-down state, so the new screen should not render a second header.
- Folders in the selection are always excluded; every bulk endpoint takes contentlet inodes only.
- One action per execute. Firing an action moves contentlets to a new workflow step, which invalidates the other actions' counts.
- A run can outlive the dialog — closing it does not cancel the action, and progress is reported in the toolbar. The configuration step should not assume the dialog stays open.

**Assumption to validate:** that these actions are wanted in *bulk* at all. Move and Add to Bundle are clearly bulk operations. Assign/Comment is less obvious — assigning 50 contentlets to one person with one comment may be a real workflow or may be something no one does. Worth a quick check with editors before designing that case in depth.

rjvelazco:
Related to #36979 — a first pass at the configuration step **in code**, so the design team can leave feedback on something clickable rather than a mockup. Deliberately not `Fixes`: the ticket stays open for that feedback.

Epic: #33999

### Proposed Changes

The Action Center could only fire actions that needed nothing beyond the selection. Everything else was greyed out behind one `requiresInput` flag. This adds a **configuration step** between picking an action and previewing it, so the flow is `pick → configure → preview → execute`, and turns on every action that needs a single kind of input.

* **Split `requiresInput` into the four flags the backend actually advertises.** `WorkflowResource.createActionInputViews()` is the sole producer of the `actionInputs[]` contract and emits exactly `assignable`, `commentable`, `pushPublish`, `moveable` — so the set is closed by the API, not by convention. The single boolean gated all four together, which meant nothing could be enabled without enabling everything.
* **Move** — destination folder picker, seeded at the folder Content Drive is currently browsing so the tree opens where you already are. Reuses `DotHostFolderFieldComponent` from `@dotcms/edit-content` (sites list, lazy folder tree, search, staged selection) rather than building a picker.
* **Add to Bundle** — bundle picker; pick an existing bundle or type a name to create one. Previously permanently disabled.
* **Assign / Comment** and **Push Publish** — two new embeddable, dialog-free step components in `libs/ui`.
* **One primary action in the footer.** The per-scheme Continue button inside each accordion panel is gone, and footer Done became a disabled-until-armed Continue. Only one action can be armed across all schemes, so a per-panel button repeated a single global decision below a scroll boundary — the radio and the button acting on it often couldn't be seen together. Every step now has its primary bottom-right: `Continue → Continue → Execute`.
* **What remains disabled is now precisely "needs more than one screen."** `requiredInputKinds()` returns the screens an action needs, in the legacy wizard's order. `assignable` + `commentable` collapse to one screen (the step renders either or both), so only genuinely multi-screen actions — e.g. assignable *and* push-publish — stay gated, and the hint says that instead of implying bulk can't collect input at all.

### Checklist
- [x] Tests
- [x] Translations

**Translations:** 13 new keys in `Language.properties`, English only — other locales fall back. Two dead keys removed (`action-center.done`, `add-to-bundle.pending`).

**Security:** No new endpoints and no new permission surface. Add to Bundle posts to the existing `/DotAjaxDirector/…/addToBundle` servlet, which already accepted several comma-separated ids, so bulk needed no backend change. Bundles are deliberately **not** license-gated, matching both the existing single-item context-menu path and the product decision that dotCMS no longer sells on license tier. Permission enforcement is unchanged — server-side, per contentlet, as before.

### Additional Info

**Two contracts meet at the move step.** The picker writes `hostname:/path` (the Site-or-Folder *content field* format); `MoveContentActionlet` reads `//hostname/path`. `toPathToMove` / `toHostFolderValue` own the conversion in both directions, with a round-trip test since they bracket the picker.

**Add to Bundle is not a workflow action.** No actionlet, no step transition, and it takes **identifiers, not inodes** — the only action here that does. A bundle holds one entry per identifier, so language versions of one contentlet are a single asset. That's deduped client-side and explained in the step *before* firing, because the server's `total` is what the result toast reports and an unexplained smaller number reads as a partial failure. It shares the execution *state* (toolbar indicator, one-at-a-time guard, result toast) so it behaves identically from the outside.

**Same-folder moves are refused.** Seeding the picker means a destination is present from the outset, so Continue gates on it having *changed* — otherwise one click would move every item to where it already is, costing a version and a reindex each for no change.

**Three deliberate divergences from the legacy forms**, which the design review may want to weigh in on:
1. No move field inside assign/comment. The legacy form renders a `dot-page-selector` there, conflating a third action input; with a dedicated move step that would be a duplicate control.
2. No `customCode` branch in push publish. The legacy form replaces itself with plugin-supplied HTML via `DotParseHtmlService`, which has no contract for size or validity inside a host's dialog.
3. Push publish shows and hides fields with `iWantTo` rather than disabling them. An expiry-only push has no publish date and no filter *at all*, and a greyed control invites the user to wonder what would enable it. **This is the most visible behaviour difference from the legacy dialog.**

**`PushPublishEnvSelectorComponent` moved from `apps/dotcms-ui` to `libs/ui`.** Reaching from a lib into the app is a project cycle (`dotcms-ui` lazy-loads the portlets, so portlet → `@components/*` → `dotcms-ui` closes the loop); the existing `dot-publishing-queue` import of the legacy push-publish form has that same cycle and this avoids adding another. The component was self-contained, and its only production consumer was the legacy form — which now shares it, so the two can't drift on how environments are chosen. Its spec stays in `apps/`, where the `DOTTestBed` harness it relies on lives.

**Known follow-ups**
* Multi-screen actions (assignable + push-publish, etc.) — `configure` renders one screen at a time. `requiredInputKinds()` already returns them ordered, so this becomes "render the list" rather than a redesign.
* Back sits in a top bar while Continue is bottom-right. A conventional wizard pairs them in the footer; the same argument that moved Continue there applies to Back. Left alone as it predates this work.
* The env selector has a dead `assetIdentifier` input — unused in both its class and template. Left in place because the legacy form still binds it.

**Verification:** `nx serve dotcms-ui` builds and serves clean. `portlets-content-drive` 1181 tests, `ui` 820, `dotcms-ui` push-publish/wizard/pages 348 — all passing; lint clean across `portlets-content-drive`, `ui`, `dotcms-models`, `dotcms-ui`, `portlets-dot-publishing-queue-portlet`.

### Video

https://github.com/user-attachments/assets/ea61ccda-ce4d-4096-9574-857c403170b4

https://github.com/user-attachments/assets/71502f10-1704-4b04-a1bc-b6fd0341f985

This PR fixes: #36979

This PR fixes: #36979


KevinDavilaDotCMS (approved):

dario-daza (approved):

zJaaal:
## Review: regressions, code standards, duplication

Solid PR overall. Splitting `requiresInput` into the four flags the API actually advertises is the right call, the `toPathToMove` / `toHostFolderValue` seam is well factored, and reusing `DotHostFolderFieldComponent` instead of building a picker is exactly right. The findings below are mostly about state that does not survive the new back-navigation, plus some cleanup.

Line numbers are at head `19d727a9`.

> **Re-verified after the `main` merge + single-screen sections refactor (`26e69b7` → `19d727a9`).** That refactor resolved finding 7 and most of 17-18 outright — marked ~~resolved~~ below rather than deleted, so the trail stays readable. Findings 1-6 and 8-16 all still reproduce on the current head. Nothing new turned up in the re-check, and none of it needs to hold the merge.

---

### Blocking

**1. A chosen move destination is silently replaced on Back-from-preview, and can fire a destination the UI never showed.**
`dot-content-drive-action-center.component.html:337` binds `[startingPath]="$currentPath()"`, which is the folder Content Drive is *browsing*, not the destination the user picked. Because `@switch ($view())` destroys the whole `configure` block on Continue, stepping back re-creates `DotContentDriveActionMoveTargetComponent` from scratch, and its `linkedSignal` re-initialises from `startingPath`.

That is not only a display reset. `DotHostFolderFieldComponent` is a `ControlValueAccessor` that pushes its own value outward on load: `host-folder-field.component.ts:192` wires `handlePathToSaveChange` to `store.pathToSave`, and `:439` calls `this.onChange(pathToSave)` whenever it becomes non-null. With `[isRequired]="true"` and no path, `loadSites` resolves a node via `getCurrentSiteAsTreeNodeItem()` (`host-folder-field.store.ts:1090`), `confirmedNode` is set, `pathToSave` computes, and `onChange` fires straight through `(ngModelChange)` into `onHostFolderValueChange` and out as `pathToMoveChange`.

Repro: arm a move action, pick `//host/blogs/`, Continue, then click "Back to destination". The picker shows the browsing folder again, `$pathToMove` is overwritten with it, and Continue goes disabled with "This is the current folder".

The worse window is before that async load settles: the parent still holds `//host/blogs/` while the picker displays something else, so Continue is armed and Execute fires a destination that was never on screen.

Fix: bind `[startingPath]="$pathToMove() || $currentPath()"`.

The spec at `dot-content-drive-action-center.component.spec.ts:1326` ("should keep the chosen destination when stepping back") passes only because `DotBrowsingService` is mocked such that no node ever confirms, so the CVA never emits. Worth a test where it does.

---

### High

**2. Assign/comment and push publish input is discarded on the same Back, and actively overwritten with empty state.**
Same `@switch` destruction (`html:313` and `:326`) plus `onBack` at `component.ts:770`. The parent hoists `$assignComment` / `$pushPublish` specifically so they survive the round trip, but nothing feeds them back down as initial values, and the re-created child emits its empty state from its constructor `effect` immediately, so the parent's copy is clobbered rather than merely stale.

Repro: type a multi-line workflow comment, Continue to the preview, press Back to uncheck one row. The comment is gone and the assignee has reset to `roles[0]`. Push publish loses environments, dates and filter the same way.

This compounds with finding 8 below: the `effect` based output propagation is what turns a re-mount into a destructive write.

**3. The same-folder gate blocks legitimate moves.**
`$destinationUnchanged` (`component.ts:304`) compares the destination against `$currentPath()` (`:291`), which is `//${hostname}${store.path() || '/'}`. When `store.path()` is unset, on initial load or with a search/filter applied, that resolves to the site root.

Repro: search for content that lives in `/blogs/`, select it, arm a move action, choose the site root as the destination. Continue is permanently disabled with "This is the current folder", even though no selected item is at the root.

The doc comment on `$currentPath` already flags this ("with a search or filter applied the selection can span folders, and no single 'current path' exists for it") and then concludes "which is all it is used for". That is no longer true: it is also the correctness gate at `:304`. Either gate on the items' actual folders, or drop the gate and let the server no-op.

**4. Add to Bundle renders `NaN` in the result toast on a server-side failure.**
`withActionExecution.ts:264` assumes `result.errors` is numeric. `RemotePublishAjaxAction.addToBundle` answers **HTTP 200** with `{"errors": "<exception message>"}` and no `total` on `DotPublisherException` (`RemotePublishAjaxAction.java:850-857`, note it is written on the normal response path, not an error status). So `catchError` never fires, `failCount` becomes a string, and `successCount = Math.max((undefined ?? 0) - "message", 0)` is `NaN`. The shell's toast (`dot-content-drive-shell.component.ts:498`) then renders a warn reading "NaN succeeded, &lt;java message&gt; failed, 0 skipped" instead of an error.

Related: when `responseMap` is null or empty the servlet writes no body at all, so `result` is null and the toast reports 0 succeeded on what may have been a success. Worth guarding both with a `typeof result?.errors === 'number'` check and routing anything else to `httpErrorManager`.

**5. A failed roles lookup is an unexplained dead end.**
`dot-workflow-assign-comment.component.ts:142` swallows the error with `catchError(() => of([]))`. `$roles` stays `[]`, `$assign` stays `''`, so `$valid` is `false` forever: an empty assignee select, a permanently disabled Continue, and only the generic "Choose an assignee to continue" hint. Reachable on a transient 5xx and also whenever `nextAssign` is empty, since `DotRolesService.get('')` requests `/api/v1/roles//rolehierarchyanduserroles` and 404s.

This is standard #8 (no silent failures: handle, surface a user-facing error state, log the detail). The same `catchError(() => of([]))` pattern is in three more places: `dot-content-drive-action-bundle-target.component.ts:118`, `dot-workflow-push-publish.component.ts:232` and `:248`. The bundle one degrades gracefully by design, which is fine, but it should still log.

---

### Medium

**6. No `takeUntilDestroyed` on any of the four new HTTP subscriptions.**
`dot-content-drive-action-bundle-target.component.ts:118`, `dot-workflow-assign-comment.component.ts:133`, `dot-workflow-push-publish.component.ts:232` and `:248`. `take(1)` completes the stream but does not cancel an in-flight request, and each callback then writes signals on a destroyed component. Given finding 1, these components are destroyed and re-created on every step transition, so this is the common path rather than an edge case. The repo checklist requires `takeUntilDestroyed(destroyRef)` on every stream.

**7. ~~`@default` in the configure switch catches `move` *and* `null`.~~ Resolved in `19d727a9`.**
`html:333`. `$configureKind()` returns `'move' | 'assignComment' | 'pushPublish' | 'bundle' | null`, and move is the fallback branch, so any future kind, or a `null` reaching the `configure` view, renders the move destination picker. The footer hint chain at `:378` has the same shape, with move as the final `@else if`. Make it `@case ('move')` with an empty `@default`. The gates guarding entry are correct today; the fallback is just the wrong thing to land on. — **Fixed:** the template is now `@for (kind of $configureKinds())` with an explicit `@case ('move')`, and `DotActionCenterConfigureKind` no longer includes `null`. The footer hint chain became `$configureHint()`, which also names the first unsatisfied section.

**8. `effect()` used to propagate state.**
`dot-workflow-assign-comment.component.ts:114` and `dot-workflow-push-publish.component.ts:185` exist solely to emit outputs. Angular's docs are explicit: *"Avoid using effects for propagation of state changes. This can result in ExpressionChangedAfterItHasBeenChecked errors, infinite circular updates, or unnecessary change detection cycles."* Both components already have `onAssignChange` / `onCommentChange` / `onIWantToChange` handlers, so emit there plus once after each async load settles. The "one effect publishes both outputs so they can never disagree" rationale holds just as well with a single `emitValue()` helper called from those handlers, and it removes the clobber in finding 2.

**9. `$publishDate` / `$expireDate` are typed `signal<Date>` but hold `null` at runtime.**
`dot-workflow-push-publish.component.ts:100-101`. The template writes `(ngModelChange)="$publishDate.set($event)"` from `p-datePicker`, where `$event` is `any` and is `null` when the field is cleared. Both `buildValue()`'s `?? new Date()` and `$valid()`'s `!this.$publishDate()` are written for exactly that case, so the code knows; the type does not. Should be `signal<Date | null>`.

**10. `changeDetection: ChangeDetectionStrategy.OnPush` on all four new components.**
`dot-workflow-assign-comment.component.ts:51`, `dot-workflow-push-publish.component.ts:82`, `dot-content-drive-action-bundle-target.component.ts:67`, `dot-content-drive-action-move-target.component.ts:41`. OnPush is the framework default on Angular 22 (this workspace is on 22.1.0; `Eager` is the opt-in you see on the legacy components in the same diff). Redundant on new components per the repo's Angular standards.

**11. 65 bracket-accesses to protected/private members across the new specs.**
For example `spectator.component['onHostFolderValueChange'](...)`, `spectator.component['$hasError']()`, `spectator.component['$view'].set('preview')`. `docs/frontend/TESTING_REVIEW_RULES.md:49` lists this as a violation: *"Implementation detail testing: Testing private methods, internal state → test user-visible behavior."* The templates already carry `data-testid` on every control, so driving `(ngModelChange)` on `move-target-picker` and `bundle-target-select` and asserting on rendered output would work. It also matters for finding 1: tests that go through the DOM would have caught the CVA emit that the internal-state tests miss.

**12. None of the new steps render loading, empty or error states** (standard #7). The bundle and roles selects are only `[disabled]` while in flight: no spinner, no "couldn't load", no empty message.

**13. `PushPublishEnvSelectorComponent` moved to `libs/ui`, spec left in `apps/dotcms-ui`.**
`libs/ui` now has a component with zero coverage under `nx test ui`, and `apps/.../dot-push-publish-env-selector/` is an orphan directory containing nothing but a spec. The `DOTTestBed` dependency is a real obstacle, but the resolution is to port the spec to Spectator alongside the component rather than strand it.

---

### Duplication

**14. `buildValue()` re-implements `DotWorkflowEventHandlerService.processWorkflowPayload`.**
`dot-workflow-push-publish.component.ts:213` versus `dot-workflow-event-handler.service.ts:163-181`: same `DotFormatDateService`, same `'yyyy-MM-dd'` and `'HH-mm'` format strings, same `environment.join()` into `whereToSend`, same `pushActionSelected` into `iWantTo`. The shape is correct, so this is not a bug today, but it is two independent copies of the push-publish wire format in a lib and a service, and they will drift the first time the backend contract shifts. Extracting the push-publish branch into a shared pure function both call would be the highest-value cleanup in the PR.

**15. The bundle sessionStorage contract is copied, not shared.**
`dot-content-drive-action-bundle-target.component.ts:26` declares `const LAST_BUNDLE_USED = 'lastSelectedBundle'`, and `libs/ui/src/lib/components/add-to-bundle/dot-add-to-bundle.component.ts:36` declares the identical literal with its own inline `setItem` and `JSON.parse`. The comments say the key is "deliberately the same key `DotAddToBundleComponent` writes" and that `rememberLastBundleUsed` is "Shared with the legacy dialog", but nothing is actually shared: they are two string literals that happen to match, so a change to one silently desyncs the two dialogs.

The new `readLastBundleUsed` / `rememberLastBundleUsed` pair is strictly better than the legacy inline code, since it tolerates corrupt storage and quota failures where the legacy `JSON.parse` throws. Moving them next to the legacy component in `libs/ui`, exporting them, and having both call them would make the sharing real. Same applies to `toBundle` versus the legacy `setBundleData`, `preselectLastUsed` versus `getDefaultBundle`, and `$placeholder` versus the legacy placeholder logic.

**16. Roles loading duplicated from the legacy assign form.**
The `DotRolesService.get(roleId, hierarchy)` plus default-to-first-role plus `{label, value}` mapping at `dot-workflow-assign-comment.component.ts:130-148`. Not rebuilding `DotCommentAndAssignFormComponent` is well justified (the `dot-page-selector` conflation is a real reason), but the roles lookup itself is the reusable part and is now written twice.

---

### Cleanup

**17. Dead code.**
- ~~`requiresMovePath`~~ — deleted in `19d727a9`, along with `hasUnsupportedInput`, `actionIsBlocked` and `workflowActionHint`.
- `requiresInput` (`utils/action-center.ts:496`, model at `dot-bulk-actions.model.ts:139`) is still computed and documented as a "convenience roll-up", but nothing reads it — and with `hasUnsupportedInput` now deleted there is not even an indirect consumer left. **Still open.**
- `pendingHint` machinery: two interface fields (`:47`, `:84`), a mapping (`:241`), a template branch, and a `quickActionHint` branch, all for something no action sets any more. The comment at `:553` says so directly ("Nothing carries one today"). Removing the last producer is the moment to remove the machinery.

**18. Stale and orphaned documentation.** Mostly resolved in `19d727a9`.
`utils/action-center.ts:384-391` is a dangling JSDoc block attached to nothing, sitting immediately above a second JSDoc for `DotActionInputKind`, and its text ("Push-publish and assign/comment still have no step, so an action needing either stays disabled") is contradicted by `requiredInputKinds` directly below it. The same claim is repeated at `dot-content-drive-action-center.component.ts:114-118` (including "promoting to a lib", which this PR did), `:92` ("Only a move target is collected today"), `:333` ("Adding a step for push-publish later means one more case here", already there), and in the `actionIsBlocked` / `workflowActionHint` docs at `:510` and `:524`. — **All of those are fixed in `19d727a9`:** the dangling block is gone and the class docs now read "Every workflow action is reachable".

One stale reference survives, and it is now a dangling one: `dot-bulk-actions.model.ts:137` still tells the reader to use "`hasUnsupportedInput` to decide whether the row is still disabled", pointing at a function this refactor deleted.

Separately, `dot-content-drive-action-bundle-target.component.ts:53-56` reads "It owns its own `p-dialog`, its own accept/cancel footer, and calls `AddToBundleService.addToBundle` itself on submit" one line before "Presentational: the parent owns the chosen bundle and fires the request". The subject is meant to be the legacy component but reads as this one, and the component's own spec asserts there is no `p-dialog`.

---

### Minor

- `disabled()` is documented on `DotWorkflowPushPublishComponent` as "Freezes every field" and is bound on every other control, but not on `dot-push-publish-env-selector` (`html:123`), which also has no `setDisabledState`, so NgModel `[disabled]` would not reach it either. Environments stay editable while an action is in flight.
- `toPathToMove` splits on the first `:` while `toHostFolderValue` splits on the first `/`, so the pair is asymmetric for a hostname carrying a port: `localhost:8080:/folder/` becomes `//localhost/8080:/folder/`, a malformed `_path_to_move`. The round-trip test hides it because the other direction parses correctly.
- The Add to Bundle drill-down header is published with `previewItems.length` (`component.ts:591`) while the step's hint reports `$bundleAssetCount()`. With two language versions of one contentlet the header says "2 items" and the body says "1 item(s) will be added", which is the exact confusion the collapse notice exists to prevent.
- `$backLabel` (`component.ts:458`) returns "Back to settings" unless the *only* section is `move`, on the reasoning that "anything else is a form, or several" — but `bundle` lands there and is a picker. The whole point of `$backLabel` was that "Back to actions" would be a lie on a move.
- `ViewEncapsulation.None` plus an unscoped `.p-multiselect { width: 100% }` (`dot-push-publish-env-selector.component.scss:6`) now ships from `libs/ui`, so the Action Center dialog injects a global width override for every `p-multiselect` in the app. Pre-existing, but the move is a good moment to scope it.
- `editable="true"` (`dot-content-drive-action-bundle-target.component.html:20`) as a string attribute where the sibling uses `[isRequired]="true"`. Works via `booleanAttribute`, just inconsistent.
- Push publish failures drop `result.errorMessages` entirely (`withActionExecution.ts:268`); the legacy path at least logged them.

---

### Verified clean

- The new `AddToBundleService` / `DotCurrentUserService` shell providers. The store injects them eagerly and only the shell provides `DotContentDriveStore`, so nothing else can construct it unprovided.
- `PushPublishEnvSelectorComponent`'s move to `libs/ui`: no `writeValue` clobber loop, NgModel's `viewModel` identity check holds, and bare `@use "variables"` is already the established precedent in that lib.
- `assignable` / `commentable` matching `WorkflowResource.createActionInputViews` one to one (`WorkflowResource.java:906-913`). (The `hasUnsupportedInput` one-screen gate originally cited here is gone; every action is now reachable, with all its sections on one screen.)
- All 13 new `Language.properties` keys plus the borrowed `hide` / `Change` / `time-zone` / `Execute` / `assignee.form.assignee` / `comment.form.comments` keys are present, and both removed keys have no remaining references.

---

*Reviewed by Claude (Opus 5) on @zJaaal's behalf. Re-verified against `19d727a9`.*

PR #36499, opened 2026-07-09
fix(reindex): bound per-entry mapping so hung storage I/O can't wedge indexing

wezell:
### Proposed Changes

Fixes #36498 — a single hung filesystem `stat` on network-backed storage (NFS/EFS/S3-FUSE) permanently wedges the single-threaded `ReindexThread`, silently halting **all** content indexing (production incident 2026-07-09, geesefs mount, [yandex-cloud/geesefs#197](https://github.com/yandex-cloud/geesefs/issues/197)).

* Run each `dist_reindex_journal` entry's load+map (`loadVersionInodes` → `toMap`, which stats binary files via `Contentlet.getBinary` → `File.exists`) under a **bounded timeout** on a small dedicated executor (`ReindexMappingRunner`), wired into `ContentletIndexAPIImpl.appendBulkRequestToProcessor` — the single choke point all three observed wedge sites funnel through.
* On timeout: loud ERROR log (identifier + phase), the entry is marked failed via the **existing** `markAsFailed` / journal retry semantics, and the loop continues with the next entry.
* Each task runs on its own **virtual thread** (Java 25). A wedged thread is **abandoned, not interrupted-and-reused** — a thread stuck in an uninterruptible native `stat(2)` ignores interrupts and pins its carrier, so in-flight tasks are capped by a semaphore whose permit is only released when the task actually finishes; exhaustion of the cap is a distinct loud alarm since it means the storage itself is down. Worker threads release their thread-local DB connection after each task.

**Config** (env-overridable via `DOT_` prefix):

| Property | Default | Meaning |
|---|---|---|
| `REINDEX_CONTENTLET_MAPPING_TIMEOUT_SECONDS` | `120` | Per-entry mapping timeout; `0` disables the guard and preserves prior behavior exactly |
| `REINDEX_CONTENTLET_MAPPING_MAX_THREADS` | `8` | Cap on concurrent (incl. wedged) mapping worker threads |

**Scope notes**

* Surgical to the reindex-processor path only; `Contentlet.getBinary`, the non-processor `appendBulkRequest` path, and the delete path (no FS access) are untouched.
* A task that is merely slow (not wedged) may complete after its timeout and still write to the bulk processor — harmless: indexing is idempotent by doc id, and the entry retry re-indexes the same content.
* The spec's optional "cheaper wins" (`getTitle()` identifier fallback, metadata reuse) were skipped — the timeout covers all three observed wedge sites; add if a narrower guard proves necessary.

### Checklist

[...]

PR #37041, opened 2026-08-12
fix(reindex): stop the mapping guard from permanently killing content indexing (#37038)

### LINKED ISSUE #36498: ReindexThread: single hung filesystem stat on network storage silently halts all content indexing
## Problem Statement

A single hung filesystem operation permanently freezes **all** content indexing for a dotCMS instance.

`ReindexThread` drains the `dist_reindex_journal` on a single thread (`dot-ReindexThreadSubmitter-0`). While mapping a contentlet for Elasticsearch, that thread performs blocking file I/O (`File.exists()` on binary fields). If the underlying storage is network-backed (NFS, EFS, S3-FUSE) and a single `stat(2)` never returns, the thread blocks forever in native code — `RUNNABLE` but frozen, no exception, no timeout, no log output. The reindex queue silently stops draining.

**Production incident (2026-07-09):** a tenant's assets live on an S3-backed FUSE mount (geesefs). A geesefs bug ([yandex-cloud/geesefs#197](https://github.com/yandex-cloud/geesefs/issues/197)) left single FUSE LOOKUP requests unanswered forever. One hung `File.exists()` froze all indexing for the instance for hours — **three separate times in one day** — with no error and no alert.

The FS bug is being pursued upstream, but dotCMS should not be architected so that one unanswered stat halts indexing forever.

### Observed wedge points (thread dumps, Java 25, `dotcms/dotcms:latest`)

All three funnel through the single-threaded reindex loop (`ReindexThread.runReindexLoop` → `appendToBulkProcessor`):

1. **Title building during ES mapping**
```
java.io.File.exists(File.java:790)
com.dotmarketing.portlets.contentlet.model.Contentlet.getBinary(Contentlet.java:1221)
com.dotmarketing.portlets.contentlet.model.Contentlet.buildName(Contentlet.java:446)
com.dotmarketing.portlets.contentlet.model.Contentlet.getTitle(Contentlet.java:347)
com.dotcms.content.elasticsearch.business.ESMappingAPIImpl.toMap(ESMappingAPIImpl.java:507)
com.dotcms.content.elasticsearch.business.ContentletIndexAPIImpl.addBulkRequestToProcessor(...)
com.dotmarketing.common.reindex.ReindexThread.runReindexLoop(ReindexThread.java:229)
```

2. **JSON→contentlet transform while loading version inodes**
```
java.io.File.exists(File.java:790)
com.dotcms.content.business.json.ContentletJsonAPIImpl.getBinary(ContentletJsonAPIImpl.java:412)
com.dotcms.content.business.json.ContentletJsonAPIImpl.getContentletMapFromImmutable(...)
com.dotmarketing.portlets.contentlet.transform.ContentletTransformer.transform(...)
com.dotcms.content.elasticsearch.business.ESContentFactoryImpl.findInDb(...)
com.dotcms.content.elasticsearch.business.ContentletIndexAPIImpl.loadVersionInodes(...)
```

3. **Metadata generation during ES mapping**
```
java.io.File.exists(File.java:790)
com.dotmarketing.portlets.contentlet.model.Contentlet.getBinary(Contentlet.java:1221)
com.dotcms.storage.FileMetadataAPIImpl.generateBasicMetadata(FileMetadataAPIImpl.java:146)
com.dotcms.storage.FileMetadataAPIImpl.generateContentletMetadata(...)
com.dotcms.content.elasticsearch.business.ESMappingAPIImpl.writeMetadata(ESMappingAPIImpl.java:720)
com.dotcms.content.elasticsearch.business.ESMappingAPIImpl.toMap(ESMappingAPIImpl.java:617)
```

## Steps to Reproduce

1. Store contentlet binary assets on a network-backed filesystem (e.g., an S3 FUSE mount such as geesefs).
2. Cause a single `stat`/LOOKUP on one asset file to hang indefinitely (the geesefs bug above does this; any wedged NFS/FUSE mount reproduces it).
3. Trigger a reindex of content that includes the affected contentlet.
4. Observe: `dot-ReindexThreadSubmitter-0` blocks forever in `File.exists()`; the `dist_reindex_journal` queue stops draining; no error is logged; all content indexing for the instance halts until restart.

## Acceptance Criteria

- [ ] A hung or slow storage operation while mapping **one** contentlet does not stop the reindex pipeline: that journal entry fails per existing `dist_reindex_journal` retry/failure semantics (`markAsFailed`), and the loop continues with subsequent entries.
- [ ] Per-contentlet mapping in the reindex path runs under a **config-driven bounded timeout** (e.g. `REINDEX_CONTENTLET_MAPPING_TIMEOUT_SECONDS`, env-overridable via `DOT_` prefix), with a generous default (60–120s); `0` disables the timeout and preserves current behavior exactly.
- [ ] On timeout, a loud ERROR is logged including the contentlet identifier, inode(s), and mapping phase.
- [ ] A thread wedged in uninterruptible native I/O is abandoned (not interrupted-and-reused); the executor is bounded so wedged threads cannot pile up unbounded, and pool exhaustion logs a clear alarm (indicates storage itself is down).
- [ ] Existing journal failure semantics are reused, not reinvented.
- [ ] Unit tests: a mapping that blocks past the timeout fails that entry while the next entry still processes; timeout=0 preserves current behavior.
- [ ] Change is surgical to the reindex path — `Contentlet.getBinary` itself is not refactored.

## dotCMS Version

Production, `dotcms/dotcms:latest` (Java 25). Code path present on `main`.

## Severity

## Links

- Upstream FS bug: https://github.com/yandex-cloud/geesefs/issues/197
- NA (no Freshdesk ticket)

### LINKED ISSUE #37038: Reindex mapping guard leaks semaphore permits, permanently killing content indexing (regression in 26.07.13-01)
## Problem Statement

**Confirmed regression in 26.07.13-01 and later.** The reindex mapping timeout guard added by PR #36499 (fix for #36498) leaks semaphore permits permanently. After 8 mapping operations fail to complete, a node stops indexing content **forever** until it is restarted — and every subsequent index update is silently parked in a terminal state that the queue loader never reads again.

The net effect is worse than the bug it fixed: #36498 froze indexing with the queue intact and recovered on restart. This freezes indexing **and destroys the backlog**.

### Mechanism

`ReindexMappingRunner` caps in-flight mapping work with a `Semaphore` of `REINDEX_CONTENTLET_MAPPING_MAX_THREADS` (default 8). The permit is released in the `finally` block of the **task body**:

```java
// ReindexMappingRunner.java:72-85
final Future<T> future = executor.submit(() -> {
    try {
        return task.call();
    } finally {
        try {
            perTaskCleanup.run();
        } finally {
            inFlight.release();   // <-- only runs if the task body runs
        }
    }
});
```

A task abandoned on timeout (`future.cancel(true)`, line 91) whose body never executes never reaches that `finally`. **The cap is therefore cumulative, not concurrent**: 8 such events over the lifetime of the JVM permanently exhaust the pool, and every subsequent entry is rejected with `Reindex mapping pool exhausted` (line 66).

Two distinct paths prevent the `finally` from running. Mere slowness is *not* one of them — a task that overruns the 120s timeout but finishes at 200s does release its permit.

1. **True wedge** — the task is blocked in an uninterruptible native `stat(2)`/`read` on network-backed storage. This is the case the guard was designed for.
2. **Carrier starvation** — the task never mounts. Tasks run on virtual threads (`ReindexMappingRunner.java:30-31`), file I/O does not unmount the carrier, and the carrier pool is `availableProcessors()` (2–4 in a container). Two genuinely wedged tasks pin every carrier; subsequent tasks queue, never execute their body, time out at 120s and are abandoned **with their permit still held** — despite their own storage being healthy.

Path 2 is the significant one: it means **~2 transient stalls are enough to leak all 8 permits**, which explains recurrence every 1–3 days on multiple nodes without a dead filesystem. Note that `maxThreads` (8) is greater than the carrier count (2–4), so the semaphore can never bind before carrier starvation does — it bounds tasks, while the resource that actually runs out is carriers.

### Secondary defects

**a) A pool rejection is charged as a content failure.** The `catch` in `ContentletIndexAPIImpl.java:2609-2611` calls `markAsFailed` without distinguishing "this entry is bad" from "the pool is broken". With `RETRY_FAILED_INDEX_TIMES=5`, `markAsFailed` (`ReindexQueueFactory.java:270-276`) increments priority by 1 up to 5 attempts, then sets `Priority.ERROR.dbValue() + priority` = `400 + 105` = **505**. The queue loader selects `priority <= 400` (`ReindexQueueFactory.java:326, 337`), so entries at 505 are never read again. The queue is drained into a terminal state at the rate the guard rejects.

**b) The guarded region includes the ES bulk enqueue.** `mapEntryForProcessor` reaches `BulkProcessor.add()`, whose backoff is `constantBackoff(20s, 10)` = up to **200s** (`ReindexThread.java:113-117`) — greater than the 120s mapping timeout. Worse, `REINDEX_THREAD_CONCURRENT_REQUESTS` defaults to **1** (`ReindexThread.java:102`), so `add()` blocks on the ES semaphore while a bulk is in flight. **Elasticsearch slowness alone is sufficient to trigger this**, with no storage problem at all.

**c) Deletes bypass the guard entirely.** `appendBulkRemoveRequestToProcessor` (`ContentletIndexAPIImpl.java:2959`) does not go through the runner; only `appendBulkRequestToProcessor` (`:2605`) does. While the guard is exhausted, deletes keep applying while writes are rejected — the index loses documents while gaining none, so divergence compounds instead of merely stalling.

**d) Leaked DB connections.** `perTaskCleanup` is `DbConnectionFactory::closeSilently` (`ContentletIndexAPIImpl.java:197`) and shares the same leaked `finally`. Each leaked permit is also a retained PostgreSQL connection for the life of the JVM.

**e) No signal.** The only outward symptom is a counter on the maintenance screen. Nothing fails a health check, nothing pauses the reindex thread, and page rendering is unaffected — so the instance looks healthy while content silently stops being indexed.

### Observed impact

Confirmed on a customer's staging and production environments (anonymized). Timeline from upgrade to first failure was **~10 minutes**:

| When | What |
|---|---|
| Night of 2026-08-05 | Upgrade 26.07.06-3 → 26.07.21-01 (26.07.06-3 does **not** contain the PR) |
| 2026-08-06 00:04–00:11 | Rolling restart, 4 JVM starts |
| 2026-08-06 00:15:40 | First failed reindex record in the journal — ever |

Over 6 days, undetected:

- **315,030** journal entries in terminal ERROR state; **303,282** distinct contentlets never indexed
- **99.3%** at priority 505 — i.e. ordinary content saves, not full reindexes
- Content is in PostgreSQL but **not** in Elasticsearch → invisible to admin content search, `dotcontent` queries, query-driven widgets and navigation, site search, and REST search. Pages still render, so the instance appears healthy.
- **5 separate node episodes across 3 of 4 nodes**, each preceded by exactly 8 timeouts (= `REINDEX_CONTENTLET_MAPPING_MAX_THREADS`), then 40,000–50,000 `pool exhausted` errors per hour until that node was restarted. Recurs every 1–3 days.

### Exposure

Any instance upgraded to 26.07.13-01 or later is exposed. To check:

```sql
SELECT count(*) FROM dist_reindex_journal WHERE priority >= 400;
```

### Mitigation (no rollback required)

PR #36499 ships an off switch: `0` disables the guard and restores the prior behavior exactly (`ReindexMappingRunner.java:62-64` runs the task inline, with no executor and no semaphore).

1. Set `DOT_REINDEX_CONTENTLET_MAPPING_TIMEOUT_SECONDS=0`
2. Restart the nodes — this applies the setting and clears the leaked permits
3. Requeue the parked entries:
   ```sql
   UPDATE dist_reindex_journal
      SET priority = priority - 400 - (priority % 100),
          index_val = NULL,
          serverid = NULL
    WHERE priority >= 400;
   ```
   (For an entry at 505 this resolves to `100` = `Priority.NORMAL`. `lastIdIndexed` resets to 0 on startup and whenever the loader query returns empty — `ReindexQueueFactory.java:310, 352` — so requeued low-id rows are picked up normally.)

## Steps to Reproduce

The failure is a race against storage or Elasticsearch latency. The deterministic reproduction is to force the mapping task to overrun its timeout without ever completing.

1. Deploy 26.07.13-01 or later with the default `REINDEX_CONTENTLET_MAPPING_TIMEOUT_SECONDS=120` and `REINDEX_CONTENTLET_MAPPING_MAX_THREADS=8`.
2. Constrain the virtual-thread carrier pool to container-like conditions: `-Djdk.virtualThreadScheduler.parallelism=2`. (Reproducing on a many-core workstation hides the problem entirely.)
3. Make binary-field storage stall in an uninterruptible way — an NFS/EFS mount pointed at an unreachable server, or a FUSE mount suspended with `SIGSTOP` on the FUSE daemon — so that `stat(2)` on an asset never returns.
4. Save content with a binary field so reindex entries flow through `appendBulkRequestToProcessor`. Confirm `Timed out after 120s mapping reindex entry with identifier '...'` in the log.
5. Restore storage to full health.
6. Continue saving content. Observe that timeouts keep accruing even though storage is now healthy, and that after 8 total timeout events **every** subsequent save is rejected with `Reindex mapping pool exhausted: all 8 in-flight tasks are busy or wedged in storage I/O`.

**Expected:** once storage recovers, indexing resumes. The cap bounds how many mapping operations run at once, not how many have ever timed out.

**Actual:** indexing is permanently dead for the life of the JVM. Every rejected entry is charged a retry until it reaches priority 505, where the queue loader never reads it again — so the backlog is destroyed rather than merely delayed.

**Shortcut for the permit-leak half:** in `ReindexMappingRunnerTest`, submit a task that blocks on an uninterruptible latch, let it time out, and assert `inFlight.availablePermits()` — it stays decremented after the timeout returns.

**Shortcut for defect (b):** leave storage healthy and instead make Elasticsearch slow enough that `BulkProcessor.add()` blocks past 120s (200s of backoff is available, and `REINDEX_THREAD_CONCURRENT_REQUESTS=1` serializes the enqueue). The same timeout and the same leak occur with no storage fault present.

### Diagnosing which path a live incident took

Take a thread dump and count the `dot-reindex-mapping-*` threads:

```bash
jcmd <pid> Thread.dump_to_file -format=json /tmp/dump.json
```

If 8 threads sit in a native `stat`/`read` frame, storage is genuinely down. If only ~2 are in native frames while the log shows 8 `Timed out after` lines, the remaining permits were leaked by tasks that never mounted — carrier starvation.

## Acceptance Criteria

**The permit leak (the core defect)**

- [ ] The cap bounds **concurrent** mapping work, not the cumulative number of timeouts observed over the JVM's lifetime.
- [ ] A task abandoned on timeout that never executes its body does not permanently consume a permit.
- [ ] A task that overruns the timeout but later completes returns its permit (this already holds — protect it with a regression test).
- [ ] After a transient storage or Elasticsearch stall that produces more than `MAX_THREADS` timeouts, indexing resumes once the underlying dependency recovers, with **no restart required**.
- [ ] Regression test: N+1 timeouts where N = `MAX_THREADS`, followed by a healthy mapping operation that must succeed.
- [ ] Whatever bounds carrier pinning is reconciled with the carrier count — a cap of 8 over 2–4 carriers cannot bind before starvation does. Either the cap tracks available parallelism, or the mechanism no longer depends on virtual-thread carriers.

**Queue integrity**

- [ ] A pool rejection does **not** increment the entry's error count and does not push it toward `Priority.ERROR`. Rejection is an infrastructure failure, not a content failure.
- [ ] No code path can drain the reindex queue into a terminal state faster than it can process it.
- [ ] Entries already parked at `priority > 400` by this defect can be recovered (documented SQL is acceptable; a maintenance action is better).

**Observability and failure mode**

- [ ] On pool exhaustion the reindex thread pauses and a health check fails, instead of draining the queue into ERROR.
- [ ] The exhaustion log distinguishes "storage is wedged" from "tasks never got a carrier" so the on-call path is not misdirected — the current message asserts storage failure unconditionally.
- [ ] Permanently dead content indexing is visible without inspecting `dist_reindex_journal` by hand.

**Guarded region**

- [ ] The Elasticsearch bulk enqueue is outside the guarded region; the guard covers only the storage-touching load/map work it was introduced for.
- [ ] Elasticsearch slowness alone cannot trigger the mapping timeout. If a timeout must remain around work with its own retry budget, the timeout exceeds that budget (currently 120s < 200s).

**Deletes**

- [ ] While mapping is degraded, deletes do not continue applying unilaterally — the index must not lose documents while gaining none. Either both paths are gated together, or the divergence is bounded and reported.

## dotCMS Version

- **Introduced in:** 26.07.13-01 (commit `32fd108`, PR #36499, merged 2026-07-13)
- **Confirmed on:** 26.07.21-01
- **Last unaffected version:** 26.07.06-3
- **Still present on:** `main`
- **Affected:** every release from 26.07.13-01 onward, plus any LTS line that took #36499

## Severity

Critical - System unusable

A core subsystem dies permanently and silently on a shipped release, discards the work queue irrecoverably without manual SQL, and recurs every 1–3 days across multiple nodes. Content operations — admin search, `dotcontent` queries, query-driven widgets and navigation, site search, REST search — are all broken. Page rendering is unaffected, which is precisely why it went undetected for 6 days.

## Links

- Regression introduced by: #36499
- Original defect it fixed: #36498
- Freshdesk ticket: FD-38711

fabrizzio-dotCMS:
## What

Fixes #37038 — a regression shipped in 26.07.13-01 by PR #36499 (the fix for #36498).

The reindex mapping guard leaks its semaphore permits permanently. After 8 abandoned mapping tasks, a node stops indexing content until it is restarted, and every entry rejected in the meantime is charged a retry until it parks at priority 505 — past the `priority <= 400` the queue loader reads. The net effect was worse than the bug it fixed: #36498 froze indexing with the queue intact and recovered on restart; this froze indexing **and destroyed the backlog**.

Confirmed on a customer's staging and production environments: 315,030 journal entries in terminal ERROR state across 6 days, 303,282 distinct contentlets never indexed, 99.3% of them ordinary content saves.

## Root cause

`inFlight.release()` lived inside the `finally` of the submitted task — which is exactly the code a task that never runs can never reach:

```java
executor.submit(() -> {
    try {
        return task.call();
    } finally {
        inFlight.release();   // never reached if the body never runs
    }
});
```

That placement was not sloppiness, it was forced. Tasks ran on **virtual** threads, and file I/O does not unmount a virtual thread from its carrier. The carrier pool is `availableProcessors()` — 2–4 in a container, shared with every other virtual thread in the JVM — so an abandoned task keeps consuming a globally scarce resource. The cap therefore *had* to count abandoned work, which forced its release into the task body, which is the leak. And with `maxThreads` at 8 over 2–4 carriers, the cap sat above the point at which the carrier pool was already fully pinned, so it could never bind before starvation did.

### 1. Two counters, and platform threads to make that safe

| Counter | Released by | Bounds |
|---|---|---|
| Concurrency (`REINDEX_CONTENTLET_MAPPING_MAX_THREADS`, 8) | the **caller**, in a `finally` that always runs | how many mappings run at once |
| Abandonment (`REINDEX_CONTENTLET_MAPPING_MAX_ABANDONED`, 32) | only a task that actually finishes | how many are stuck; the ceiling opens a circuit breaker |

Moving the tasks to platform threads is what makes caller-side release safe: an abandoned platform thread costs only its own stack (~1MB) and steals nothing from anyone else, so releasing the concurrency permit on timeout cannot let a wedge go unaccounted for. Both counters are bounded, so the worst-case live thread count is `maxConcurrent + maxAbandoned`. Threads are daemons, so a thread wedged in native I/O can never hold up JVM shutdown.

Not keeping virtual threads was deliberate. The original argument for them — abandoning one is free, and a hung native `stat` ignores `interrupt()` so abandoning is the only move — is real, but it buys little once the cap is 8 anyway, and it costs a shared carrier pool. See `docs/backend/VIRTUAL_THREADS.md` (#36900) for the socket-vs-file rule.

**Because the abandonment counter heals when a stuck task eventually returns, recovery no longer requires a restart.**

### 2. A pool rejection is never charged to the entry

`ReindexPoolExhaustedException` (infrastructure) is now distinct from `ReindexMappingTimeoutException` (this entry's own mapping overran). Only the latter reaches `markAsFailed`. A rejected entry is left byte-for-byte untouched — same priority, no failure cause — so it is retried later instead of marching toward `Priority.ERROR`.

A momentarily saturated pool defers just that entry; an open circuit aborts the batch so the loop can back off.

### 3. The failure is now visible and self-limiting

- New `reindex-mapping` health check (`MONITOR_MODE` by default, not a liveness check since the guard self-heals) reporting `abandonedMappings` / `indexingStalled`.
- An open circuit makes the reindex thread sleep `REINDEX_THREAD_SLEEP_WHEN_DEGRADED` (30s) instead of spinning the journal at `SLEEP_ON_ERROR` (500ms) speed.
- The exhaustion log is throttled to once a minute — it used to emit 40–50K identical lines an hour — and no longer asserts "the underlying storage may be down" unconditionally, which misdirected the on-call path when the real cause was carrier starvation.

### 4. The index enqueue moved out of the guarded region

`BulkProcessor.add()` blocks on the client's in-flight semaphore (`REINDEX_THREAD_CONCURRENT_REQUESTS` defaults to **1**) and its backoff budget is `20s × 10 = 200s` — greater than the 120s mapping timeout. **Elasticsearch slowness alone was sufficient to trigger the guard**, with no storage fault at all.

Mapping now returns `MappedDocument`s (the storage-touching half) and `enqueueMappedDocuments` runs on the caller thread, outside the timeout.

## Tests

71 unit tests green (`ReindexMappingRunnerTest`, `ContentletIndexAPIImplMappingTimeoutTest`, `ReindexThreadUnitTest`, and the rest of the `ContentletIndexAPIImpl*Test` family).

New coverage for the regression specifically:

- `moreTimeoutsThanTheConcurrencyCapDoesNotDisableTheRunner` — `cap + 3` timeouts, then a healthy task must still succeed. This is the test that fails on the old code.
- `abandonmentCeilingOpensTheCircuit` / `circuitClosesWhenAbandonedTasksReturn` — the ceiling trips, then heals with no restart.
- `interruptibleSlowTaskIsFreedAndNotCountedAsAbandoned` — merely-slow work is not booked as abandoned.
- `asyncTaskRunsOnNamedPlatformThread` — asserts *not* virtual, so a future well-meaning swap back is caught.
- `poolRejectionNeverFailsTheJournalEntry`, `poolRecoversWhenAbandonedTaskFinishes` — wiring level.
- `ContentletIndexAPIImplMappingTimeoutIT#poolRejectionLeavesTheJournalRowUntouched` — the data-loss half against real SQL: priority unchanged, `index_val` null. Registered in `MainSuite2b`.

## Deploying to an already-affected instance

This PR removes the need for the workaround, but it does not clean up rows already parked by it. Any instance that ran 26.07.13-01 or later still needs the requeue:

```sql
SELECT count(*) FROM dist_reindex_journal WHERE priority >= 400;   -- exposure

UPDATE dist_reindex_journal
   SET priority = priority - 400 - (priority % 100),
       index_val = NULL,
       serverid = NULL
 WHERE priority >= 400;
```

`DOT_REINDEX_CONTENTLET_MAPPING_TIMEOUT_SECONDS=0` remains a full off switch (runs the mapping inline, no executor, no semaphore) and still works with an open circuit, so it stays available as a mitigation.

## New configuration

| Property | Default | Purpose |
|---|---|---|
| `REINDEX_CONTENTLET_MAPPING_MAX_ABANDONED` | 32 | abandoned-task ceiling that opens the circuit |
| `REINDEX_THREAD_SLEEP_WHEN_DEGRADED` | 30000 | back-off while the guard is refusing work |

## Not included

Recovering parked rows through a maintenance action rather than manual SQL. It is in the issue's acceptance criteria and is worth doing, but it is a UI/maintenance-portlet change rather than part of this defect's mechanism.

This PR fixes: #37038


wezell (approved):
Tested against 400,000 contentlets.  3 legitimately failed, as they should have. Looks good.  Added a logging change to prevent log span.

PR #37048, opened 2026-08-13
feat(dotAI): Implement new dotAI config page

### LINKED ISSUE #36826: dotAI: Provider Configuration UI
## What we need

## Why it matters

Three concrete failures:

## What "done" looks like

## Acceptance criteria

## Out of scope

## Notes

## Open questions

## References

KevinDavilaDotCMS:
https://github.com/user-attachments/assets/080ae9cf-a2fb-48b5-bbc1-bb99c2917661

## Summary

Redesigns the dotAI provider configuration page (per-site Chat/Embeddings/Image capabilities plus shared settings) and adds a real "Test Connection" flow, driven entirely by provider metadata from the backend so new providers need no frontend changes.

### Backend
- New `POST /v1/ai/providers/test/{capability}` endpoint (`AiProviderResource`) that builds the real LangChain4j model for a capability and issues one minimal live call.
- Generic `requiredUnless` field metadata (`ProviderField`) so a provider can declare "either this field or that sibling field is required" (e.g. Azure's `model`/`deploymentName`) without any provider-specific frontend code.
- Guards against exfiltrating a masked credential (`*****`) to an attacker-controlled `endpoint`/`provider` during connection testing (`targetsStoredDestination`).
- Capability-aware default timeout for the test call (longer for `IMAGE`, since real generation routinely exceeds the chat/embeddings default).
- Extracted shared host-resolution logic (lenient vs. strict) into `AiHostResolver`, used consistently by both the save and test-connection endpoints.

### Frontend
- New `dot-ai-config-detail` page: white background, per-capability cards (Chat/Embeddings/Image), a shared settings card, and an "Additional properties" escape hatch for provider fields not yet modeled.
- Dynamic field rendering (`dot-ai-dynamic-field`) purely from provider metadata — text/number/secret inputs, required-field markers, and cross-field `requiredUnless` validation.
- A `SECRET` field with an already-saved value renders as read-only plain text (no reveal toggle, since the real secret never reaches the browser); editing it switches back to a masked password input so a newly-typed secret isn't shown in clear text.
- "Test Connection" per capability, with the Cancel button appearing only when the form actually differs from the last-saved state.
- All new UI text localized via `Language.properties` (`apps.ai.*` keys).

### Fixes from review
Addressed the review feedback that was in-scope for this PR (see PR comments for full discussion): additional-properties no longer silently override a real field with the same name and are cleared on provider switch; non-string saved values (e.g. Vertex's JSON credentials) round-trip without `String()` corruption; the connection-test capability parsing is locale-independent; the `requiredUnless` cross-field validity fix no longer depends on incidental Angular re-render timing; the page now shows the site it's configuring; and the page blocks Save (instead of silently saving an empty/default config) when the initial load fails.

### Testing
- New/updated Jest specs for `dot-ai-capability-card`, `dot-ai-dynamic-field`, `dot-ai-config-detail`, and `dot-ai-config.constants`.
- New/updated JUnit tests for `AiProviderResource`, `ProviderConnectionTester`, and provider metadata.

### Checklist
- [x] Tests
- [x] Translations

This PR fixes: #36826


dotCMS/src/main/java/com/dotcms/ai/client/langchain4j/ProviderConnectionTester.java
```diff
@@ -0,0 +1,79 @@
...
+
+    private static String testEmbeddings(final ProviderConfig config) {
+        final EmbeddingModel model = LangChain4jModelFactory.buildEmbeddingModel(config);
+        model.embed(TextSegment.from(TEST_EMBEDDING_INPUT)).content();
+        return "Connection successful.";
+    }
+
+    private static String testImage(final ProviderConfig config) {
+        final ImageModel model = LangChain4jModelFactory.buildImageModel(config);
+        model.generate(TEST_IMAGE_PROMPT).content();
```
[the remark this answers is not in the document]
KevinDavilaDotCMS: /ar Langchain4j is the intentional core dependency this PR introduces — it's the architecture the whole dotAI provider-configuration feature is built on (see the accompanying OpenAI/Azure/Bedrock/Vertex/Anthropic/OpenRouter/Google AI ModelProviderStrategy classes added in this same PR), not an incidental or hidden usage. This specific call site (ProviderConnectionTester) only sends a fixed, hardcoded test prompt to verify provider connectivity — no untrusted or user-controlled input flows into the model call, consistent with Semgrep's own note that this is broad source/sink matching rather than a real taint issue.

dotCMS/src/main/java/com/dotcms/ai/client/langchain4j/ProviderConnectionTester.java
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Verifies that a {@link ProviderConfig} actually works by building the LangChain4J model for the
+ * requested {@link Capability} and issuing one minimal, real request against the provider.
+ *
+ * <p>Building the model already validates required fields (see {@link ModelProviderStrategy}),
+ * so a missing {@code apiKey}/{@code model} fails fast with a clear message before any network
+ * call is made. Anything past that — bad credentials, unreachable endpoint, unknown model — only
+ * surfaces once the provider actually answers, hence the real call.
+ */
+public final class ProviderConnectionTester {
```
dario-daza, line 22: I'm not sure whether it 1.validates the provider's capabilities or 2. is meant to trim error messages.

If it's the first option, consider adding a happy‑path test to the validations of the ProviderConnectionTesterTest class. Otherwise, rename the class to something more related to error handling.

dotCMS/src/test/java/com/dotcms/ai/rest/AiProviderResourceTest.java
```diff
@@ -0,0 +1,302 @@
...
+
+/**
+ * Regression tests for the masked-credential exfiltration/SSRF guard in
+ * {@link AiProviderResource#testConnection}: a masked credential (e.g. {@code "apiKey":
+ * "*****"}) must only resolve to the real stored secret when the posted {@code provider} and
+ * {@code endpoint} match what's actually stored — otherwise a caller could pair a masked field
+ * (obtainable from any {@code GET}) with an attacker-controlled {@code endpoint} and have the
+ * server send the real secret there.
+ */
+public class AiProviderResourceTest {
```
dario-daza, line 19: I'm not sure if it should be included in the main suite to be executed. ask @nollymar or @fabrizzio-dotCMS

dotCMS/src/main/java/com/dotcms/ai/client/langchain4j/ProviderConnectionTester.java
KevinDavilaDotCMS: Fair question — the class does test the connection (that's its main job: build the model for the given capability and issue one real, minimal call). `friendlyMessage()` is just the error-formatting half of returning that result, not a separate responsibility, so I'd rather keep it here than split it into its own class.

A true happy-path unit test for `test()` would need injecting the model builder (it's currently a direct static call to `LangChain4jModelFactory`), which felt like more machinery than this pass-through logic (build model → call it → fixed success string) warrants on its own. The model-building side is already covered by `LangChain4jModelFactoryTest` and the per-provider strategy tests; `ProviderConnectionTesterTest` covers the `friendlyMessage` truncation logic added here. Happy to add the DI refactor if you feel it's worth it for the coverage.

dotCMS/src/test/java/com/dotcms/ai/rest/AiProviderResourceTest.java
KevinDavilaDotCMS: Good to double check — this is a plain JUnit4 unit test under `dotCMS/src/test/java`, so it's picked up automatically by Surefire's default `**/*Test.java` pattern in the 'JVM Unit Tests' CI job. No Suite registration needed — that requirement only applies to the integration tests under `dotcms-integration` (`MainSuite`, `Junit5 Suite`, etc.), which is a different module/test type. I ran it locally with `mvn test -Dtest=AiProviderResourceTest` and it passes; the JVM Unit Tests job on this PR is green too, consistent with it being picked up.

dario-daza (approved):

rjvelazco (commented):
## Issues

### Important

**1. `onlySelf: true` leaves the parent FormGroup stale-INVALID**

* **File:** `dot-ai-capability-card.component.ts:~320` (the `requiredUnlessFields` recheck).
* **Mechanism:** `model` is constructed before `deploymentName` exists.
* **Result:** its validator sees no parent, returns an error, and the group caches INVALID.
* **The recheck** flips `model` to VALID but `onlySelf: true` skips parent recalculation.
* **User impact:** an Azure config saved with only `deploymentName` loads as invalid.
* **Consequence:** Save and Test Connection stay blocked until the user touches a field.
* **Contradiction:** the spec at line 254 asserts the opposite, so that test should be failing.
* **Verify:** `pnpm nx test dotcms-ui --testPathPattern=dot-ai-capability-card`.
* **Fix:** drop `onlySelf: true`, keep `emitEvent: false`.

**2. Bedrock and Vertex clients are never closed**

* **File:** `ProviderConnectionTester.java:~87` (`test()`).
* **Issue:** each call builds a model wrapping an `AutoCloseable` SDK client, then discards it.
* **Affected:** `BedrockRuntimeClient` / `BedrockRuntimeAsyncClient` and `VertexAI`.
* **Why it matters:** repeated admin "Test connection" clicks accumulate connection pools and threads.
* **Verify:** whether LangChain4j closes these itself or registers a shutdown hook.
* **Fix:** close the client after the test call if it does not.

**3. Test path uses lenient host resolution, save uses strict**

* **File:** `AiProviderResource.java` (`testConnection` calls `AiHostResolver.resolveHost`).
* **Behavior:** an unresolvable `siteId` silently falls back to the current host.
* **Effect:** masked credentials then resolve against a different site's stored config.
* **Risk:** an admin gets a misleading "success" for a site they never tested.
* **Fix:** use `resolveHostStrict` for consistency with `PUT /v1/ai/completions/config`.

### Minor

* **Locale-sensitive enum parsing:** `capability.toUpperCase()` breaks under `tr_TR`.
* **Fix:** pass `Locale.ROOT` to both `toUpperCase()` and `toLowerCase()`.
* **Dynamic field labels bypass i18n:** `humanizeFieldName()` derives English from field names.
* **Consequence:** every provider field label ships untranslated, unlike the rest of the page.
* **Raw provider errors reach the UI:** `friendlyMessage()` forwards SDK exception text verbatim.
* **Risk:** some SDKs embed the request URL, which can carry the API key as a query param.
* **Mixed message semantics:** `testResult.message` holds i18n keys locally, raw English from the server.
* **Fragility:** this only works because `DotMessageService.get()` echoes unknown keys.
* **Unrelated file:** `core-web/.sdkmanrc` pins Java 25 inside the frontend directory.
* **Question:** does this belong in this PR, or at the repo root?
* **Stale subscriptions:** `rebuildFieldsGroup` re-subscribes on every provider switch.
* **Impact:** harmless in practice, since discarded groups never emit again.
* **Check unused imports** in `CompletionsResource` after the four private methods were removed.

## Recommendations

* **Add a spec for `dot-ai-config-detail.component.ts`:** dirty tracking and payload assembly are untested.
* **Confirm clearing behavior:** `buildPayloadSection` drops empty values from the payload.
* **Open question:** can a user actually unset a previously-saved optional field like `temperature`?
* **Consider a contract test** asserting every strategy's `configFields` throws only for unsupported capabilities.
* **Untick items:** the PR body still has placeholder text ("change 1", "original screenshot") and unchecked boxes.

rjvelazco (approved):

core-web/apps/dotcms-ui/src/app/portlets/dot-apps/components/dot-ai-config-detail/dot-ai-config-detail.component.ts
```diff
@@ -21,130 +24,157 @@ import {
...
+
+        this.dirty.set(!isEqual(this.buildCurrentPayload(), this.savedPayload));
+    }
+
+    cancel(): void {
+        const key = this.app()?.key ?? 'dotAI';
+        this.dotRouterService.goToAppsConfiguration(key);
+    }
+
+    save(): void {
```
zJaaal, line 158: **HIGH - data loss.** `save()` is never gated on a successful load, and the Save button (`dot-ai-config-detail.component.html:41`) has no `[disabled]`.

Scenario: the `forkJoin` at line 103 errors (either `GET /v1/ai/providers` or `GET /v1/ai/completions/config` - `forkJoin` also couples them, so one failing kills both). The error handler sets `loading` to `false` while `providers()` and `initialSections()` stay empty, so all three cards render as "not configured". `buildCurrentPayload()` then returns `{}` / `{settings: {...}}`, and `CompletionsResource.saveConfig` replaces `providerConfig` wholesale - one click on Save destroys every chat/embeddings/image section and its stored credentials. The same happens if the user clicks Save before the requests resolve (payload is `{}` because no cards exist yet).

Suggested fix: keep a distinct `loadFailed` state, and disable Save while `loading()` or after a failed load.

core-web/apps/dotcms-ui/src/app/portlets/dot-apps/components/dot-ai-config-detail/components/dot-ai-settings-card/dot-ai-settings-card.component.ts
```diff
@@ -0,0 +1,161 @@
...
+                new FormControl(field.type === 'checkbox' ? (field.defaultValue ?? false) : null)
+            );
+        });
+
+        const initial = this.initialValue();
+        if (initial) {
+            this.form.patchValue(initial, { emitEvent: false });
+            this.advancedForm.patchValue(initial, { emitEvent: false });
+
+            const knownKeys = new Set([
```
zJaaal, line 89: **HIGH.** `knownKeys` is built from `SETTINGS_COMMON_FIELDS` + `SETTINGS_ADVANCED_FIELDS`, but `imageSize` exists only in the fixed `form` (line ~70) and in neither constant.

So a saved `settings.imageSize` is patched into the select **and** pushed as an "Additional properties" row. Because `buildPayloadSection` applies additional properties **last** (line 138), changing Image size in the select is silently overwritten by the stale row on save. The user also sees a confusing duplicate `imageSize` row.

Fix: add `imageSize` to `SETTINGS_COMMON_FIELDS` (its docstring already says the keys exist to exclude these from the additional-properties list).

core-web/apps/dotcms-ui/src/app/portlets/dot-apps/components/dot-ai-config-detail/components/dot-ai-capability-card/dot-ai-capability-card.component.ts
```diff
@@ -0,0 +1,360 @@
...
+
+        Object.entries(this.fieldsGroup().value as Record<string, unknown>).forEach(
+            ([key, value]) => {
+                if (value !== null && value !== undefined && value !== '') {
+                    section[key] = value;
+                }
+            }
+        );
+
+        this.additionalProperties.controls.forEach((group) => {
```
zJaaal, line 255: **MEDIUM.** Additional properties are merged *after* the provider fields, so a row whose key collides with a real provider field silently overrides that field's value on save (e.g. an extra row named `model` beats the Model input).

Compounding it: `selectProvider` (line 179) rebuilds the fields group but never clears `additionalProperties`. Concrete case - an existing config written with the old raw-JSON editor has `chat.maxCompletionTokens` (a real `ProviderConfig` field that no strategy declares in `configFields`, so it always hydrates as an extra). Switching the Chat provider from OpenAI to Anthropic resets the fields but still persists `maxCompletionTokens` into the Anthropic section, contradicting this method's own "nothing carries over" contract.

dotCMS/src/main/java/com/dotcms/ai/client/langchain4j/ProviderConnectionTester.java
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * Upper bound applied to the test call when the posted config doesn't set {@code timeout}.
+     * Every provider strategy honors {@code timeout} once set (see each strategy's {@code build*}
+     * methods) except Vertex AI, which ignores it outright regardless of source — so this default
+     * only ever narrows an otherwise-unbounded provider-SDK default, never overrides an explicit
+     * value the caller supplied. Without this, an unreachable or slow {@code endpoint} could hold
+     * the request thread open indefinitely, since the SDKs' own defaults vary by provider and
+     * aren't all finite.
+     */
+    private static final int DEFAULT_TEST_TIMEOUT_SECONDS = 10;
```
zJaaal, line 46: **MEDIUM.** The 10s default is applied to all three capabilities, and it does reach the image path (`OpenAiModelProviderStrategy.buildImageModel` -> `applyCommonConfig(..., builder::timeout)`).

Real image generation (`gpt-image-1`, `dall-e-3`, Azure Foundry, Gemini image) routinely takes 15-60s, so "Test Connection" on the Image card will report a spurious timeout failure for a perfectly valid configuration whenever the admin has not set an explicit `timeout`. Consider a per-capability default (e.g. 10s for chat/embeddings, 60s+ for image).

core-web/apps/dotcms-ui/src/app/portlets/dot-apps/components/dot-ai-config-detail/components/dot-ai-capability-card/dot-ai-capability-card.component.ts
```diff
@@ -0,0 +1,360 @@
...
+        this.rebuildFieldsGroup(this.providerId(), initial);
+
+        Object.entries(initial).forEach(([key, value]) => {
+            if (key === 'provider' || knownNames.has(key)) {
+                return;
+            }
+            this.additionalProperties.push(
+                new FormGroup({
+                    key: new FormControl(key, { nonNullable: true }),
+                    value: new FormControl(value == null ? '' : String(value), {
```
zJaaal: **MEDIUM-LOW.** Unknown values are hydrated with `String(value)`, unlike `dot-ai-settings-card` which uses `JSON.stringify` on the way in and `parseIfJson` on the way out.

A nested object under a capability section - entirely possible, since the UI this PR replaces was a free-form JSON textarea - round-trips to the literal string `"[object Object]"` and is then persisted by `buildPayloadSection`, destroying the value. Reusing the settings card's stringify/parse pair here would make the two consistent.

dotCMS/src/main/java/com/dotcms/ai/rest/AiProviderResource.java
```diff
@@ -0,0 +1,259 @@
...
+
+        if (!user.isAdmin()) {
+            return Response.status(Response.Status.FORBIDDEN)
+                    .entity(Map.of(AiKeys.ERROR, "Only CMS admins can test the AI provider connection"))
+                    .build();
+        }
+
+        final Capability parsedCapability;
+        try {
+            parsedCapability = Capability.valueOf(capability.toUpperCase());
```
zJaaal: **LOW.** `capability.toUpperCase()` here (and `capability.toLowerCase()` on line 183) use the JVM default locale. Under a Turkish default locale `"image"` uppercases to `İMAGE`, `Capability.valueOf` throws, and a valid request gets a 400 "Unknown capability". Use `Locale.ROOT` for both.

core-web/apps/dotcms-ui/src/app/portlets/dot-apps/components/dot-ai-config-detail/components/dot-ai-dynamic-field/dot-ai-dynamic-field.component.html
```diff
@@ -0,0 +1,52 @@
...
+            <p-inputnumber
+                [inputId]="field().name"
+                [formControlName]="field().name"
+                [useGrouping]="false"
+                [invalid]="isInvalid()"
+                styleClass="w-full"
+                inputStyleClass="w-full text-sm" />
+        }
+        @case (DotAiProviderFieldType.SECRET) {
+            @if (isMaskedSecret()) {
```
zJaaal, line 20: **LOW - security.** When a secret is already saved, the value equals `*****` so this branch renders a plain `<input type="text">`, and `isMaskedSecret()` is deliberately not recomputed per keystroke.

Net effect: typing a brand-new API key over the placeholder shows the real secret in clear text (and in the DOM value) for the rest of the field's lifetime, whereas a first-time secret gets the masked `p-password` treatment. Consider flipping to `p-password` as soon as the control becomes dirty.

core-web/apps/dotcms-ui/src/app/portlets/dot-apps/components/dot-ai-config-detail/dot-ai-config-detail.component.html
```diff
@@ -1,38 +1,42 @@
...
+
+            <dot-ai-settings-card [initialValue]="initialSettings()" (changed)="onAnyChanged()" />
+        }
     </div>
-}
+</div>
+
+<div
+    class="flex shrink-0 items-center justify-end gap-4 border-t border-gray-200 bg-white px-6 py-3">
+    @if (dirty()) {
```
zJaaal, line 42: **LOW.** Cancel is inside `@if (dirty())`, and `dot-apps-configuration-header` was dropped from this template, so on a clean page there is no in-page way back to the Apps list. The previous version always rendered Cancel plus the app header with the site name.

core-web/.sdkmanrc
```diff
@@ -0,0 +1,3 @@
+# Enable auto-env through the sdkman_auto_env config
+# Add key=value pairs of SDKs to use below
+java=25.0.2-ms
```
zJaaal: **LOW.** This duplicates the root `.sdkmanrc` (`java=25.0.2-ms`) inside the frontend directory and is unrelated to the dotAI config page. Looks like an accidentally committed local convenience file - worth dropping from the PR.

zJaaal (commented):
### Review: correctness pass on the new dotAI config page

Read every hunk plus surrounding context. 9 inline findings below, ordered roughly by severity - the two flagged **HIGH** are both silent-data-loss paths in the save flow, so they are the ones worth looking at first.

One more finding that has no good inline anchor:

**LOW - `dotCMS/src/main/webapp/WEB-INF/openapi/openapi.yaml`** (`/v1/ai/providers/test/{capability}`): the generated `requestBody` schema is `type: string`, taken from the `String body` parameter, even though the endpoint expects a JSON object. The `@RequestBody(content = @Content(schema = @Schema(implementation = Map.class)))` annotation is not winning over the parameter type, so clients generated from the spec will send a JSON-encoded *string*. A typed request DTO would fix both the spec and the CLAUDE.md "@Schema must match the actual type" rule.

<details>
<summary>Checked and confirmed <em>not</em> bugs (so they do not get re-litigated)</summary>

- The `updateValueAndValidity({ onlySelf: true })` recheck in `rebuildFieldsGroup` does leave the parent group's aggregate status stale (verified with a throwaway jest probe: the child flips to valid, the group does not). It is harmless in practice because `FormGroupDirective.addControl` re-validates each control *with* parent propagation on render, and PrimeNG's `p-panel` projects its content even when collapsed - so every `requiredUnless` field is always registered and the group status is repaired before any user action. Worth knowing it depends on that, though.
- `AiProviderResource:199` does **not** leak the resolved secret through the Jackson 400 message: Jackson 2.16+ disables `INCLUDE_SOURCE_IN_LOCATION` by default (verified against 2.17.2, the version in `bom/application/pom.xml`), so the source snippet is redacted.
- `API_ENDPOINT` + `/providers` resolves correctly to `/api/v1/ai/providers`.
- `pInputText` does expose an `invalid` input in PrimeNG 21.1.3, so the `[invalid]` bindings compile.
- Every key in `SETTINGS_ADVANCED_FIELDS` matches an `AppKeys` `settingsKey`.
- `ProviderConfigMerger.containsMasked` / `containsMaskedCredential` exist with the semantics the new resource assumes, and `ProviderConfig` is `@JsonIgnoreProperties(ignoreUnknown = true)`, so the free-form additional properties do not break deserialization.

</details>

---
*Automated review authored by Claude (Claude Code), posted from @zJaaal's account.*

core-web/apps/dotcms-ui/src/app/portlets/dot-apps/components/dot-ai-config-detail/dot-ai-config.constants.ts
```diff
@@ -0,0 +1,248 @@
...
+     *  untouched checkbox doesn't silently save a wrong value when the key was never set. */
+    defaultValue?: boolean;
+}
+
+/**
+ * Always-visible shared settings — surfaced across every capability. Only `key` is consumed
+ * (to exclude these from the additional-properties list); `label`/`hint` mirror the message keys
+ * the fixed markup in `dot-ai-settings-card.component.html` renders directly for these fields.
+ */
+export const SETTINGS_COMMON_FIELDS: DotAiSettingsField[] = [
```
ihoffmann-dot, line 187: `imageSize` is a control in the settings form but is missing from `SETTINGS_COMMON_FIELDS`, whose only purpose is to exclude known keys from the additional properties list, so a saved `settings.imageSize` is hydrated into both the dropdown and a duplicate additional property row. Since `buildPayloadSection()` applies additional properties last, the duplicate overwrites the dropdown value and the user's Image size change is silently dropped on save

ihoffmann-dot: Btw I saw Jal already reported this

ihoffmann-dot:
It looks good overall! In addition to Jal's feedback I found these:

🟡 _**The page no longer shows which site's configuration is being edited**_

`providerConfig` is per-site and the route carries a site id, but the removed header was what displayed app.sites?.[0]?.name. On a multi-site instance this invites a mistake: pasting a production API key into another site's configuration without noticing.

**Suggestion:** surface the site name in the page heading

🟢 _**Non-blocking: Test Connection silently ignores additional properties**_

The additional properties escape hatch lets users add keys the form doesn't model, and those keys are persisted on save. But the test endpoint deserializes the body into `ProviderConfig`, which is `@JsonIgnoreProperties`, so they're dropped before the provider client is built. A green "Connection successful" therefore doesn't reflect the configuration that will actually be stored, and if the extra property was the one that made it work, the test gives false confidence.

**Suggestion:** Just flagging it. Options if you want to address it later: reject unknown keys with a notice, or state in the UI that the test ignores them.

ihoffmann-dot (changes requested):

ihoffmann-dot (approved):

core-web/apps/dotcms-ui/src/app/portlets/dot-apps/components/dot-ai-config-detail/dot-ai-config-detail.component.html
KevinDavilaDotCMS: Fair point, but leaving as-is for this PR: Cancel is meant as "discard my unsaved edits," so it only needs to show once there's something to discard — that part's intentional. The actual gap you're describing (no way back to the Apps list on a clean page) is a byproduct of dropping `dot-apps-configuration-header` in the redesign, not something this button should fix. I'd rather add that back explicitly (e.g. a breadcrumb/back link) than repurpose Cancel to always render, which would blur what the button means. Tracking as a small follow-up rather than blocking this PR on it.

KevinDavilaDotCMS:
Went through the rest of the feedback and fixed everything actionable in the last commit (data-loss on load failure, additional-properties override/clear-on-switch, imageSize duplication, image test timeout, locale-sensitive parsing, secret re-masking, missing site name, host-resolution consistency). A few remaining items I'm intentionally leaving as-is for this PR — flagging why instead of silently skipping them:

**Bedrock/Vertex client leak on Test Connection** (@rjvelazco, also flagged by the automated review) — real concern, but it's a pre-existing gap in how `LangChain4jModelFactory` builds these clients generally, not something introduced by the test-connection endpoint. Scoping a proper fix (closing/pooling `BedrockRuntimeClient`/`VertexAI`) to this PR would mean touching the shared model-building path for a feature this PR doesn't otherwise change. Worth its own follow-up.

**`openapi.yaml` schema is `type: string` instead of an object** (@zJaaal) — correct, the `@Schema(implementation = Map.class)` annotation isn't winning over the raw `String body` parameter type. Fixing it means introducing a typed request DTO for the test-connection body, which is more surface area than this PR's scope; the endpoint's actual behavior (accepts a JSON object) isn't affected, just the generated spec's accuracy for downstream client generation.

**Test Connection ignores additional properties** (@ihoffmann-dot) — accurate: `ProviderConfig` is `@JsonIgnoreProperties(ignoreUnknown = true)`, so an extra-property key that happens to matter isn't exercised by the test call. Given how narrow that case is (an unmodeled field that's also load-bearing for connectivity), I'd rather leave it as a known limitation than add scope now; happy to open a follow-up issue if this turns out to bite someone in practice.

**Mixed i18n semantics in `testResult.message`** (@rjvelazco) — also accurate: some messages are local i18n keys, others are raw strings from the backend, and it only reads correctly because `DotMessageService.get()` echoes unknown keys back unchanged. Untangling that cleanly means deciding how (or whether) to localize arbitrary provider error text, which felt like a separate, bigger conversation than this PR's scope.

Issue #35592, opened 2026-05-06
App configurations silently wiped out (dotSecretsStore.p12 integrity failure → empty store regenerated)


syedATdot:
### Research Question

In multi-node Cloud (and previously in self-hosted internal) environments, the App Configuration store (dotSecretsStore.p12) is intermittently rendered unreadable. When this happens, dotCMS's recovery code path silently backs up the corrupted store, deletes it, and creates a fresh empty one — wiping every customer-configured App (dotCDN, SAML, OAuth, Analytics, Push Publish, etc.) on that node, with no error visible at the application UI and no customer-actionable trace in the logs beyond a single PKCS12 integrity check failed warning.

The behavior is recurring across at least three customers and one internal environment over an 18-month window (Oct 2024 → Apr 2026). Each time, support and Cloud Engineering manually re-add the lost App configs as a workaround. There is currently no permanent fix and no diagnostic telemetry that explains why the keystore becomes unreadable.

This is an investigation spike — the goal is to identify the root cause(s) of PKCS12 integrity check failed in production and the safety/blast radius of the silent "delete and regenerate empty store" recovery path, so an implementation issue can be scoped.

### Timebox

8h

### Acceptance Criteria

The deliverable is a written report (commit to this issue or a linked Notion doc) that answers each of the following — most do not require a fix, only investigation:

Reproduce the failure in a controlled environment. Set up a 2-pod cluster locally (or in a sandbox) and force one of the following: rotate cluster_salt mid-run; kill a pod between the tmp write and the rename in saveSecretsStore(); publish a CompanyKeyResetEvent from a script. Document which of these produces a PKCS12 integrity check failed on a subsequent read.
Trace cluster_salt on read vs. write. Add temporary debug logging (or attach a debugger) and confirm whether the passwordSupplier returns the same digest on every node and across pod restarts. If not, find where it diverges. Open question for Cloud Eng: is cluster_salt ever rewritten by infrastructure (e.g. statefulset bootstrap)?

### Context

All Apps appear disabled / empty in the dotCMS Apps portlet (/dotAdmin/#/apps), even though they were correctly configured the day before.
In multi-node clusters, only one node at a time is affected — Apps appear correctly to users routed to the healthy node and disappear for users routed to the unhealthy one. Manuel Miranda summarized it as "like they are taking turns" (Caliber, Apr 30 2026).
The customer's UI shows no error — the Apps list is just empty, or shows the entries with all secret fields cleared.
The only application log line of interest is:

```
[...]

PR #37053, opened 2026-08-13
fix(apps): stop silently wiping the App secrets store (#36724)

### LINKED ISSUE #36724: Fix multi-node race that silently wipes App secrets store (dotSecretsStore.p12)
## Description

Implementation follow-up to spike #35592 (root cause is **known and reproduced** there — this issue is for the fix, not re-investigation).

The shared App secrets keystore `dotSecretsStore.p12` lives on shared cluster assets storage (`<real-assets-root>/server/secrets/`) and **every node reads/writes the same physical file with no locking**. Three compounding defects in `com.dotcms.security.apps.SecretsKeyStoreHelper` let a concurrent multi-node access silently wipe every configured App secret (dotCDN, SAML, OAuth, Analytics, Push Publish, …):

- **A — self-heal-on-any-failure.** `getSecretsStore()` catches *every* load `IOException` and calls `handleStorageLoadException()`, which backs up + deletes + regenerates an **empty** store (only a WARN log + a non-blocking admin notification). It does not distinguish a transient torn read from real corruption or a wrong password.
- **B1 — torn write.** `saveSecretsStore()` writes a tmp file then `FileUtil.copyFile(tmp, dest)`. On shared network storage (NFS/EFS) where hard links are unavailable, `copyFile` truncates the destination to 0 and streams the bytes back in — a concurrent reader on another node observes a partial/empty PKCS12 → integrity failure.
- **B2 — delete-before-create window.** `FileUtil.handleHardLinks` deletes the destination *before* `Files.createLink`, and `createStoreIfNeeded()` is an unsynchronized check-then-act; a concurrent node then regenerates an empty store with **no error at all** (explains field incidents with no log line, just a shrunken/empty file).

Secondary trigger (path A only): `cluster_salt`-derived password divergence — a per-node `SECRETS_KEYSTORE_PASSWORD_KEY` override, or a stale in-JVM salt after rotation.

### Proposed fix directions (evaluate — not prescriptive)
- **Atomic publish:** write tmp then `Files.move(tmp, dest, ATOMIC_MOVE)` instead of truncate+stream copy; eliminate the delete-before-create window.
- **Cross-node/process locking** around save + `createStoreIfNeeded` (e.g. a file lock) so two nodes can't write concurrently.
- **Never wipe on transient/unknown errors:** retry/backoff first; only back up + recreate when the store is *provably* unrecoverable; escalate logging to ERROR and make it actionable.
- Optionally a config toggle to disable auto-recreate.

### Reproduction artifacts (parked, no PR)
Branch **`issue-35592-secrets-store-concurrent-wipe`** — https://github.com/dotCMS/core/tree/issue-35592-secrets-store-concurrent-wipe
- Integration test: `dotcms-integration/src/test/java/com/dotcms/security/apps/SecretsStoreConcurrentWriteRaceTest.java` — two helpers = two nodes on one file with the torn-write path forced; fails reliably on `master`.
- 2-node docker repro: `docker/docker-compose-examples/issue-35592-secrets-store-race/` — the boot race alone produced a **103-byte** integrity-failing store (the exact field signature), no license needed.

## Acceptance Criteria

- [ ] A **deterministic** regression integration test (no timing/thread races) that fails on today's code and passes after the fix. Base it on the spike's `SecretsStoreConcurrentWriteRaceTest`, reshaped to deterministically simulate the failure window: save a canary secret → delete/truncate the store file mid-flight → trigger a read on the "other node" → assert the canary **survives / is recovered** rather than silently wiped.
- [ ] Keep the spike's concurrent stress test as a companion.
- [ ] No silent data loss: a torn/partial read must never result in an emptied store.
- [ ] Concurrent writes from two nodes to the shared store cannot corrupt it (atomic publish and/or locking).
- [ ] Recovery only wipes when the store is provably unrecoverable; transient errors are retried and logged at ERROR with an actionable message.

## Additional Context

Root cause, code trace, and both reproductions are documented in spike #35592 and its findings comment. Same area/team as #35592 (backend, Team : Maintenance).

## Severity

High
 
## dotCMS Version

Current Release (dotEvergreen)
[26.07.13-01](https://dev.dotcms.com/docs/changelogs?v=26.07.13-01)

### Links

https://dotcms.freshdesk.com/a/tickets/31904
https://dotcms.freshdesk.com/a/tickets/36885

### LINKED ISSUE #37061: Apps portlet list shows 0 configurations when the secrets store is unreadable, instead of reporting the error
### Description

The Apps portlet **list** reports every app as `0 configurations` when the secrets store cannot be read, while opening an app's **detail** raises an error. An administrator sees "nothing is configured", clicks in, and gets a failure — two surfaces telling them different things about the same store.

This is a **layering problem, not a missing throw.** The portlet is answering an *administrative* question with a *runtime* API:

| Caller | Question being asked | Right answer when the store is unreadable |
|---|---|---|
| `EMAWebInterceptor`, SAML, Velocity `$secrets`, analytics, telemetry | "Is this app configured for this site?" | **Degrade.** Treat as not configured and keep serving the request. |
| Apps portlet list | "What is the state of App configuration?" | **Say so.** "I cannot read the store" is the single most important thing to report. |

Both go through `AppsAPIImpl.filterSitesForAppKey`, which `AppsHelper:124` calls to compute `configurationsCount`. That method deliberately catches `Exception` and returns "no secrets for this site", because it runs on **every `/api/* request`** via `EMAWebInterceptor` — making it raise returns a 500 for the entire API, which is a regression that was already found and fixed once in PR #37053 (`94ea3579e6`).

So the fix is **not** to make `filterSitesForAppKey` throw. It is to give the admin surface an explicit store-health signal instead of inferring state from a count that is designed to degrade.

### Current behaviour

With a store this node cannot decrypt (cluster-salt rotation, or a per-node `SECRETS_KEYSTORE_PASSWORD_KEY`):

```
GET /api/v1/apps                                   -> 200, every app "configurationsCount": 0
GET /api/v1/apps/{key}/{siteId}                    -> 500
```

Observed on a container built from PR #37053.

### Impact

Misinformation, not data loss. The admin cannot destroy anything from this state — saving also goes through `getSecretsStore()`, which raises, so a "re-enter the credentials" reaction fails loudly rather than overwriting an intact store (verified: the store's checksum was unchanged after a save attempt). The cost is that the list actively points away from the real problem, which is a mis-set salt or password, at exactly the moment someone is trying to diagnose it.

Not a regression: before #37053 the store was silently wiped and recreated empty, so the list also showed 0 — having just destroyed the credentials.

### Acceptance Criteria

- [ ] The Apps portlet list surfaces "the App secrets store cannot be read" as an explicit condition, rather than rendering it as `0 configurations`.
- [ ] The signal is carried on its own channel — `getAvailableDescriptorViewsWithErrors` already returns a `List<AppDescriptorLoadError>` alongside the views, which is the natural place, or an equivalent field.
- [ ] `AppsAPIImpl.filterSitesForAppKey` still does **not** raise. Its runtime callers (`EMAWebInterceptor` on every `/api/*` request, `InitRunner.isUveEnabled()`, SAML, telemetry) must keep degrading. A regression here returns a 500 for the whole API.
- [ ] The UI distinguishes "0 configurations" from "state unknown" — an administrator must not read an unreadable store as an empty one.
- [ ] The message points at the actual cause: the store could not be decrypted, usually a changed cluster salt or a per-node `SECRETS_KEYSTORE_PASSWORD_KEY`. The wording in `SecretsKeyStoreHelper.handleUnrecoverableLoad` is a good starting point.
- [ ] The detail view's existing hard error is preserved.
- [ ] Test coverage for the list against an unreadable store.

### Additional Context

Raised in review of PR #37053 (issue #36724), which stopped an unreadable secrets store from being silently wiped. That PR documents the split at the decision point in `AppsAPIImpl.filterSitesForAppKey` and deliberately left closing it to a follow-up, because surfacing the condition in the list is a UI-facing change with its own signal path.

Frontend work is likely required alongside the backend signal.

### Severity

Low - Cosmetic or minor issue

dsolistorres:
### Proposed Changes

Fixes #36724 (implementation of spike #35592). Also fixes #37061, filed from this PR's own review and closed here.

The shared `dotSecretsStore.p12` lives on cluster assets storage and every node reads and writes the same physical file with no lock — but **the concurrency was only the trigger. The data loss was done by a single-threaded path.** `getSecretsStore()` treated every `IOException` from `KeyStore.load` as unrecoverable corruption, so it backed up and **deleted** the store; the retry loop then recreated it empty, loaded it cleanly, and logged *"KeyStore loaded successfully after 2 tries."*

There are three ways in, and the third needs no concurrency at all: a torn read, the hard-link delete window, or simply a wrong password — the password is `digest(cluster_salt)`, so a salt rotation or a per-node `SECRETS_KEYSTORE_PASSWORD_KEY` override wiped the store on the next read. The backup left behind was encrypted with the password that had just failed, which is why customers report that restoring it does not restore access.

**Four changes:**

1. **Never destroy on read failure.** The retry now retries the **read** with backoff instead of retrying after a delete, and the store is created at most once per call so the loop cannot manufacture an empty one. Integrity failures (wrong password → stop and preserve) are distinguished from torn reads (retry), verified against the JDK rather than assumed. Recreating is opt-in via `SECRETS_STORE_AUTO_RECREATE`, default **false**.
2. **Atomic publish.** Write-temp → fsync → `ATOMIC_MOVE`, replacing `FileUtil.copyFile`, which truncates the destination and streams the bytes back. The tmp file is restricted to **0600** before the move and always removed in a `finally`.
3. **Startup cannot be taken down by the new raise.** `Task00050LoadAppsSecrets.forceRun()` tolerates an unreadable store and skips the import — booting with Apps broken and the credentials intact beats booting with them destroyed.
4. **A node that cannot open the store still works.** Four read paths that previously degraded only *by accident* — because the old code wiped the store and returned nothing — now degrade on purpose, matching the cause chain with `ExceptionUtil.causedBy` since the failure arrives re-wrapped as a bare `DotRuntimeException`. Writes and the App detail views still raise, so nobody saves a blank form over a store that is intact but merely unreadable by this node.

The Apps portlet **list** degrades to zero counts and says why — a dedicated `app-secrets-store-unreadable` entry in the 200 response's `errors` field — rather than showing a bare zero that reads as "your secrets are gone". The guard sits at that call site and deliberately not inside `appKeysByHost()`, because `removeApp()`, `removeSecretsForSite()` and `exportSecrets()` read it too, and an empty map there would make `exportSecrets()` write an **empty backup that reports success**.

Net effect: **reads degrade, writes are protected.** The instance stays usable, the encrypted data survives, and the configuration returns once the salt is corrected.

📋 **[Root cause, the four changes in full, container verification, testing and known gaps →](https://github.com/dotCMS/core/pull/37053#issuecomment-5296013552)**
📋 **[Review round 8: the High fix and both Mediums verified →](https://github.com/dotCMS/core/pull/37053#issuecomment-5295936589)**

### Deliberately not included: cluster-wide locking

With atomic publish a reader can no longer observe a torn file, so a lock is **not** needed to prevent corruption. What it would add is protection against **lost updates** — two nodes each adding a different secret, last writer wins. That is a distinct bug, it puts the database in the secrets write path, and reusing the existing `ClusterLockManager` requires handling ShedLock's skip-if-locked semantics first: `ClusterLockManagerImpl.tryLock` never checks `wasExecuted()`, so a contended lock silently returns `null`. For a save that means "we didn't persist your secret and told you nothing" — worse than the bug being fixed. Filed as #37054.

This PR fixes: #36724

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/ai/app/ConfigService.java
+++ b/dotCMS/src/main/java/com/dotcms/ai/app/ConfigService.java
@@ -1,11 +1,16 @@
 package com.dotcms.ai.app;
 
 import com.dotcms.api.web.HttpServletRequestThreadLocal;
+import com.dotcms.exception.ExceptionUtil;
 import com.dotcms.security.apps.AppSecrets;
 import com.dotcms.security.apps.AppsAPI;
+import com.dotcms.security.apps.SecretsStoreUnreadableException;
 import com.dotmarketing.beans.Host;
 import com.dotmarketing.business.APILocator;
 import com.dotmarketing.business.web.WebAPILocator;
+import com.dotmarketing.exception.DotDataException;
+import com.dotmarketing.exception.DotRuntimeException;
+import com.dotmarketing.exception.DotSecurityException;
 import com.dotmarketing.util.Logger;
 import com.google.common.annotations.VisibleForTesting;
 import com.liferay.portal.model.User;
@@ -86,16 +91,33 @@ public class ConfigService {
     }
 
     private Optional<AppSecrets> getAiSecrets(final Host host, final User systemUser) {
-        // getOrElse, not get(): Try.get() unwraps by RETHROWING the failure, so the Try was doing
-        // nothing here. That went unnoticed while an unreadable App secrets store silently wiped
-        // itself and returned no secrets; since issue #36724 it raises instead, and this method
-        // would have propagated that into dotAI's config resolution rather than reporting the app
-        // as unconfigured -- unlike every other consumer, which degrades.
-        return Try
-                .of(() -> appsAPI.getSecrets(AppKeys.APP_KEY, false, host, systemUser))
-                .onFailure(e -> Logger.warnAndDebug(ConfigService.class,
-                        "Unable to read dotAI secrets; treating the app as unconfigured: "
-                                + e.getMessage(), e))
-                .getOrElse(Optional.empty());
+        // Only an unreadable secrets store degrades here; everything else keeps propagating.
+        //
+        // Try.get() rethrows the failure rather than swallowing it, and that is load-bearing:
+        // AppsAPIImpl raises InvalidLicenseException for dotAI on an unlicensed instance, and
+        // ConfigServiceTest.test_invalidLicense asserts it reaches the caller. An earlier revision of
+        // this PR replaced get() with getOrElse(Optional.empty()) on the mistaken reading that the
+        // Try was inert, which turned an invalid license into "the app is unconfigured".
+        //
+        // The reason for touching this at all: since issue #36724 an unreadable store raises where it
+        // previously wiped itself and returned nothing, so without the catch below that condition
+        // would propagate into dotAI's config resolution instead of reporting the app as
+        // unconfigured -- unlike every other consumer, which degrades.
+        try {
+            return appsAPI.getSecrets(AppKeys.APP_KEY, false, host, systemUser);
+        } catch (final DotRuntimeException e) {
+            if (!ExceptionUtil.causedBy(e, SecretsStoreUnreadableException.class)) {
+                throw e;
+            }
+            Logger.warnAndDebug(ConfigService.class,
+                    "App secrets store is unreadable; treating dotAI as unconfigured: "
+                            + e.getMessage(), e);
+            return Optional.empty();
+        } catch (final DotDataException | DotSecurityException e) {
+            Logger.warnAndDebug(ConfigService.class,
+                    "Unable to read dotAI secrets; treating the app as unconfigured: "
+                            + e.getMessage(), e);
+            return Optional.empty();
+        }
     }
 }
--- a/dotCMS/src/main/java/com/dotcms/ai/app/ConfigService.java
+++ b/dotCMS/src/main/java/com/dotcms/ai/app/ConfigService.java
@@ -86,8 +86,16 @@ public class ConfigService {
     }
 
     private Optional<AppSecrets> getAiSecrets(final Host host, final User systemUser) {
+        // getOrElse, not get(): Try.get() unwraps by RETHROWING the failure, so the Try was doing
+        // nothing here. That went unnoticed while an unreadable App secrets store silently wiped
+        // itself and returned no secrets; since issue #36724 it raises instead, and this method
+        // would have propagated that into dotAI's config resolution rather than reporting the app
+        // as unconfigured -- unlike every other consumer, which degrades.
         return Try
                 .of(() -> appsAPI.getSecrets(AppKeys.APP_KEY, false, host, systemUser))
-                .get();
+                .onFailure(e -> Logger.warnAndDebug(ConfigService.class,
+                        "Unable to read dotAI secrets; treating the app as unconfigured: "
+                                + e.getMessage(), e))
+                .getOrElse(Optional.empty());
     }
 }
--- a/dotCMS/src/main/java/com/dotcms/jitsu/validators/SiteAuthValidator.java
+++ b/dotCMS/src/main/java/com/dotcms/jitsu/validators/SiteAuthValidator.java
@@ -58,10 +58,18 @@ public class SiteAuthValidator implements AnalyticsValidator {
             } else {
                 Logger.warn(this, "HTTP Request object could not be retrieved");
             }
-            // Deliberately Exception, not just the checked DotDataException/DotSecurityException.
-            // Since issue #36724 the secrets store raises a DotRuntimeException when it cannot be
-            // read rather than silently wiping itself; that must surface as a normal validation
-            // failure on this analytics collection path, not as a raw runtime exception.
+            // An AnalyticsValidationException is this pipeline's own signal -- ContentAnalyticsUtil
+            // raises it when the site cannot be resolved from the Origin/Referer headers -- and it
+            // already carries the message the caller is meant to see. Rethrow it untouched; the
+            // broader catch below would otherwise re-wrap it as "Site Auth for Site 'null' could
+            // not be verified: <original>" and change that message.
+        } catch (final AnalyticsValidationException e) {
+            throw e;
+            // Otherwise deliberately Exception, not just the checked DotDataException /
+            // DotSecurityException. Since issue #36724 the secrets store raises a
+            // DotRuntimeException when it cannot be read rather than silently wiping itself; that
+            // must surface as a normal validation failure on this analytics collection path, not
+            // as a raw runtime exception escaping into the request.
         } catch (final Exception e) {
             final String errorMsg = String.format("Site Auth for Site '%s' could not be verified: %s",
                     null != currentSite ? currentSite.getHostname() : BLANK, ExceptionUtil.getErrorMessage(e));
--- a/dotCMS/src/main/java/com/dotcms/jitsu/validators/SiteAuthValidator.java
+++ b/dotCMS/src/main/java/com/dotcms/jitsu/validators/SiteAuthValidator.java
@@ -6,6 +6,7 @@ import com.dotcms.rest.api.v1.analytics.content.util.ContentAnalyticsUtil;
 import com.dotcms.security.apps.AppSecrets;
 import com.dotcms.security.apps.Secret;
 import com.dotcms.security.apps.SecretsStoreUnreadableException;
+import com.dotmarketing.exception.DotRuntimeException;
 import com.dotmarketing.exception.DotDataException;
 import com.dotmarketing.exception.DotSecurityException;
 import com.dotmarketing.beans.Host;
@@ -61,18 +62,28 @@ public class SiteAuthValidator implements AnalyticsValidator {
             } else {
                 Logger.warn(this, "HTTP Request object could not be retrieved");
             }
-            // SecretsStoreUnreadableException is the addition to the original two. Since issue
-            // #36724 the secrets store raises when it cannot be read rather than silently wiping
-            // itself, and that must surface as an ordinary validation failure here rather than as a
-            // raw runtime exception escaping onto the analytics collection path.
+            // Matched on the CAUSE CHAIN rather than the thrown type. Since issue #36724 an
+            // unreadable secrets store raises instead of silently wiping itself, and that must
+            // surface as an ordinary validation failure here rather than escaping raw onto the
+            // analytics collection path. But SecretsStoreUnreadableException does not survive the
+            // journey: several `catch (Exception e) { throw new DotRuntimeException(e); }` layers
+            // sit between where it is raised and here -- SecretsKeyStoreHelper.loadValueFromStore
+            // being the one on this path -- so catching the specific type alone never matches in
+            // production.
             //
-            // Deliberately that exact type. Catching Exception also swallowed
-            // AnalyticsValidationException -- this pipeline's own signal, raised by
-            // ContentAnalyticsUtil when the site cannot be resolved from the Origin/Referer headers
-            // -- and re-wrapped it as "Site Auth for Site 'null' could not be verified: <original>",
-            // changing the message callers see. Naming the condition means that signal propagates
-            // on its own, with no rethrow clause needed to protect it.
-        } catch (final DotDataException | DotSecurityException | SecretsStoreUnreadableException e) {
+            // AnalyticsValidationException is deliberately NOT caught: it is this pipeline's own
+            // signal, raised by ContentAnalyticsUtil when the site cannot be resolved from the
+            // Origin/Referer headers, and already carries the message callers are meant to see.
+            // Catching Exception here swallowed it and re-wrapped it, which regressed
+            // AnalyticsValidatorUtilTest.
+        } catch (final DotDataException | DotSecurityException | DotRuntimeException e) {
+
+            if (e instanceof DotRuntimeException
+                    && !ExceptionUtil.causedBy(e, SecretsStoreUnreadableException.class)) {
+                // A database blip, a cache failure, a programming error -- stays loud rather than
+                // being reported as an invalid site auth.
+                throw (DotRuntimeException) e;
+            }
             final String errorMsg = String.format("Site Auth for Site '%s' could not be verified: %s",
                     null != currentSite ? currentSite.getHostname() : BLANK, ExceptionUtil.getErrorMessage(e));
             Logger.warnAndDebug(SiteAuthValidator.class, errorMsg, e);
--- a/dotCMS/src/main/java/com/dotcms/jitsu/validators/SiteAuthValidator.java
+++ b/dotCMS/src/main/java/com/dotcms/jitsu/validators/SiteAuthValidator.java
@@ -5,6 +5,9 @@ import com.dotcms.exception.ExceptionUtil;
 import com.dotcms.rest.api.v1.analytics.content.util.ContentAnalyticsUtil;
 import com.dotcms.security.apps.AppSecrets;
 import com.dotcms.security.apps.Secret;
+import com.dotcms.security.apps.SecretsStoreUnreadableException;
+import com.dotmarketing.exception.DotDataException;
+import com.dotmarketing.exception.DotSecurityException;
 import com.dotmarketing.beans.Host;
 import com.dotmarketing.business.APILocator;
 import com.dotmarketing.util.Logger;
@@ -58,19 +61,18 @@ public class SiteAuthValidator implements AnalyticsValidator {
             } else {
                 Logger.warn(this, "HTTP Request object could not be retrieved");
             }
-            // An AnalyticsValidationException is this pipeline's own signal -- ContentAnalyticsUtil
-            // raises it when the site cannot be resolved from the Origin/Referer headers -- and it
-            // already carries the message the caller is meant to see. Rethrow it untouched; the
-            // broader catch below would otherwise re-wrap it as "Site Auth for Site 'null' could
-            // not be verified: <original>" and change that message.
-        } catch (final AnalyticsValidationException e) {
-            throw e;
-            // Otherwise deliberately Exception, not just the checked DotDataException /
-            // DotSecurityException. Since issue #36724 the secrets store raises a
-            // DotRuntimeException when it cannot be read rather than silently wiping itself; that
-            // must surface as a normal validation failure on this analytics collection path, not
-            // as a raw runtime exception escaping into the request.
-        } catch (final Exception e) {
+            // SecretsStoreUnreadableException is the addition to the original two. Since issue
+            // #36724 the secrets store raises when it cannot be read rather than silently wiping
+            // itself, and that must surface as an ordinary validation failure here rather than as a
+            // raw runtime exception escaping onto the analytics collection path.
+            //
+            // Deliberately that exact type. Catching Exception also swallowed
+            // AnalyticsValidationException -- this pipeline's own signal, raised by
+            // ContentAnalyticsUtil when the site cannot be resolved from the Origin/Referer headers
+            // -- and re-wrapped it as "Site Auth for Site 'null' could not be verified: <original>",
+            // changing the message callers see. Naming the condition means that signal propagates
+            // on its own, with no rethrow clause needed to protect it.
+        } catch (final DotDataException | DotSecurityException | SecretsStoreUnreadableException e) {
             final String errorMsg = String.format("Site Auth for Site '%s' could not be verified: %s",
                     null != currentSite ? currentSite.getHostname() : BLANK, ExceptionUtil.getErrorMessage(e));
             Logger.warnAndDebug(SiteAuthValidator.class, errorMsg, e);
--- a/dotCMS/src/main/java/com/dotcms/jitsu/validators/SiteAuthValidator.java
+++ b/dotCMS/src/main/java/com/dotcms/jitsu/validators/SiteAuthValidator.java
@@ -7,8 +7,6 @@ import com.dotcms.security.apps.AppSecrets;
 import com.dotcms.security.apps.Secret;
 import com.dotmarketing.beans.Host;
 import com.dotmarketing.business.APILocator;
-import com.dotmarketing.exception.DotDataException;
-import com.dotmarketing.exception.DotSecurityException;
 import com.dotmarketing.util.Logger;
 import com.dotmarketing.util.UtilMethods;
 import com.dotmarketing.util.json.JSONObject;
@@ -60,7 +58,11 @@ public class SiteAuthValidator implements AnalyticsValidator {
             } else {
                 Logger.warn(this, "HTTP Request object could not be retrieved");
             }
-        } catch (final DotDataException | DotSecurityException e) {
+            // Deliberately Exception, not just the checked DotDataException/DotSecurityException.
+            // Since issue #36724 the secrets store raises a DotRuntimeException when it cannot be
+            // read rather than silently wiping itself; that must surface as a normal validation
+            // failure on this analytics collection path, not as a raw runtime exception.
+        } catch (final Exception e) {
             final String errorMsg = String.format("Site Auth for Site '%s' could not be verified: %s",
                     null != currentSite ? currentSite.getHostname() : BLANK, ExceptionUtil.getErrorMessage(e));
             Logger.warnAndDebug(SiteAuthValidator.class, errorMsg, e);
--- a/dotCMS/src/main/java/com/dotcms/rest/api/v1/analytics/content/util/ContentAnalyticsUtil.java
+++ b/dotCMS/src/main/java/com/dotcms/rest/api/v1/analytics/content/util/ContentAnalyticsUtil.java
@@ -12,10 +12,12 @@ import com.dotcms.jitsu.validators.AnalyticsValidatorUtil;
 import com.dotcms.jitsu.validators.SiteAuthValidator;
 import com.dotcms.security.apps.AppSecrets;
 import com.dotcms.security.apps.Secret;
+import com.dotcms.security.apps.SecretsStoreUnreadableException;
 import com.dotmarketing.beans.Host;
 import com.dotmarketing.business.APILocator;
 import com.dotmarketing.business.web.WebAPILocator;
 import com.dotmarketing.exception.DotDataException;
+import com.dotmarketing.exception.DotRuntimeException;
 import com.dotmarketing.exception.DotSecurityException;
 import com.dotmarketing.portlets.contentlet.business.HostAPI;
 import com.dotmarketing.util.*;
@@ -267,6 +269,21 @@ public class ContentAnalyticsUtil {
             Logger.error(ContentAnalyticsUtil.class,
                     "Error retrieving app secrets for site: " + currentSite.getIdentifier(), e);
             return Collections.emptyMap();
+        } catch (final DotRuntimeException e) {
+            // An unreadable App secrets store raises since issue #36724, where it previously wiped
+            // itself and returned nothing. This method must keep degrading: it feeds
+            // EventAnalyticsProxyHelper.buildAuthHeader(), which is called outside proxy()'s try
+            // block, so propagating here turns every analytics proxy/ingest request into a 500.
+            // The exception arrives wrapped as a plain DotRuntimeException -- SecretCachedKeyStoreImpl
+            // and friends re-wrap it -- so recognise it through the cause chain, and let every other
+            // runtime failure (DB, cache) keep propagating rather than masquerading as "no secrets".
+            if (!ExceptionUtil.causedBy(e, SecretsStoreUnreadableException.class)) {
+                throw e;
+            }
+            Logger.debug(ContentAnalyticsUtil.class, () -> String.format(
+                    "App secrets store is unreadable; treating Content Analytics as unconfigured for site '%s'",
+                    currentSite.getIdentifier()));
+            return Collections.emptyMap();
         }
     }
 
@@ -306,6 +323,16 @@ public class ContentAnalyticsUtil {
         } catch (final DotDataException | DotSecurityException e) {
             Logger.error(ContentAnalyticsUtil.class,
                     "Error retrieving site key from app secrets for site: " + currentSite.getIdentifier(), e);
+            return Optional.empty();
+        } catch (final DotRuntimeException e) {
+            // Same reasoning as getAppSecrets() above: degrade on an unreadable store (see #36724),
+            // propagate anything else.
+            if (!ExceptionUtil.causedBy(e, SecretsStoreUnreadableException.class)) {
+                throw e;
+            }
+            Logger.debug(ContentAnalyticsUtil.class, () -> String.format(
+                    "App secrets store is unreadable; no Content Analytics site key available for site '%s'",
+                    currentSite.getIdentifier()));
             return Optional.empty();
         }
     }
--- a/dotCMS/src/main/java/com/dotcms/rest/api/v1/apps/AppsHelper.java
+++ b/dotCMS/src/main/java/com/dotcms/rest/api/v1/apps/AppsHelper.java
@@ -2,6 +2,7 @@ package com.dotcms.rest.api.v1.apps;
 
 import static com.dotmarketing.util.UtilMethods.isNotSet;
 
+import com.dotcms.exception.ExceptionUtil;
 import com.dotcms.rest.api.MultiPartUtils;
 import com.dotcms.rest.api.v1.apps.view.AppView;
 import com.dotcms.rest.api.v1.apps.view.SecretView;
@@ -14,6 +15,7 @@ import com.dotcms.security.apps.AppsAPI;
 import com.dotcms.security.apps.AppsUtil;
 import com.dotcms.security.apps.ParamDescriptor;
 import com.dotcms.security.apps.Secret;
+import com.dotcms.security.apps.SecretsStoreUnreadableException;
 import com.dotcms.security.apps.Type;
 import com.dotcms.util.CollectionsUtils;
 import com.dotcms.util.PaginationUtil;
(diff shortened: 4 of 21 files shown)


dsolistorres:
## Review round: the High, and both Mediums verified

### 🟠 High — `ContentAnalyticsUtil` — confirmed and fixed (`3108727708`)

Verified in code rather than taken on trust, and every step of the reported chain holds:

* `ContentAnalyticsUtil.java:266` and `:306` caught only `DotDataException | DotSecurityException`.
* `buildAuthHeader(host)` is called at `EventAnalyticsProxyHelper.java:121`; the try block opens at `:138`. So the call really is outside it.
* `EventAnalyticsProxyResource` catches `AnalyticsValidationException`, `IOException`, `IllegalArgumentException`, `DotSecurityException`, `DotDataException` — no `RuntimeException`. Nothing between the raise and the response.

So an unreadable store returned **500 on every analytics proxy and ingest request**. Fixed with the same `ExceptionUtil.causedBy` cause-chain match already used by `filterSitesForAppKey` and `SiteAuthValidator`.

**Why the earlier caller audit missed it.** The audit checked whether each caller *had* a guard, not whether the guard's catch types could actually catch this. `ContentAnalyticsUtil` has a `try` that returns an empty map, so it was scored defensive — while naming only checked exceptions, which an unchecked `DotRuntimeException` walks straight past. That is the same mistake the bug itself is made of.

**One more site found by redoing the audit with the corrected criterion:** `GoogleTranslationService.getFallbackAPIKey()` — a catch whose only job is to fall back to the configured API key property, which the raise also escaped. Fixed in the same commit. The remaining callers are genuinely safe, and now for a stated reason: `AIUtil`, `AnalyticsApp`, `AnalyticsWebAPIImpl`, `AnalyticsInitializer`, `PreRenderSEOWebAPIImpl`, `DotVelocitySecretAppConfig`, `PageScannerResource`, `DotIdentityProviderConfigurationFactoryImpl` and `OpenAIImageTaggingContentListener` all guard with `Try.of(...)` / `Try.run(...)`, which catches `Throwable`; `RekognitionActionlet` catches `Exception`. `DotIdentityProviderConfigurationImpl` has no guard of its own but is only ever constructed inside the factory's `Try.of(...).getOrNull()`.

**Test coverage.** New `ContentAnalyticsUtilUnreadableStoreTest` (6 cases). Worth noting why the gap survived: `SiteAuthValidatorTest` and `EventAnalyticsProxyHelperTest` both `mockStatic(ContentAnalyticsUtil.class)`, so neither could ever observe this class's own exception handling. The new tests throw the **wrapped** production shape — `DotRuntimeException(SecretsStoreUnreadableException(...))` — against the real class, so they exercise the `causedBy` match rather than a mock at the boundary.

### 🟡 Medium 1 — telemetry collector — verified, no change needed

The finding's stated assumption ("the telemetry runner does not swallow per-collector exceptions") does not hold. `MetricStatsCollector` iterates collectors in a `for` loop at `:137`, and `:219` is a `catch (final Throwable e)` **inside** that loop which records a `MetricCalculationError` for the offending metric and continues to the next one.

So an unreadable store cannot take down a telemetry run: the snapshot still completes with every other metric, and this one lands in `errors`. Behaviour does change — from a silent "count 0" to a recorded per-metric error — but that is contained, and it is the more honest of the two: an unreadable store means *unknown*, not *zero sites configured*. Downgraded to informational per the finding's own instruction.

### 🟡 Medium 2 — `EMAResource.getDetails` — verified, deliberately unchanged

Confirmed the behaviour change: an unreadable store yields a 500 where it previously produced `DoesNotExistException`.

Keeping it. It is a REST-only entry point — `grep` for `EMAResource` across `dotCMS/src/main/java` returns no internal references, so it does **not** feed edit-mode rendering; that path runs through `EMAWebInterceptor` → `existsConfiguration()` → `filterSitesForAppKey`, which this PR already fixed. It also requires a backend user.

More importantly, 404 would be the wrong answer here. Telling an administrator "no EMA configuration exists on this site" when the store is merely unreadable invites them to re-enter the configuration and save over data that is still intact — precisely the hazard this PR exists to remove. A 500 says "this node cannot read the store", which is what actually happened. Consistent with the stance already documented for `getSecrets()` and the `AppsHelper` detail paths.

### Verification

* Unit: **39 tests, 0 failures** (`ContentAnalyticsUtilUnreadableStoreTest`, `EventAnalyticsProxyHelperTest`, `SiteAuthValidatorTest`, `AppsAPIImplUnreadableStoreTest`).
* Integration: **98 tests, 0 failures** (`SecretsStoreWipeRegressionTest`, `SecretsStoreConcurrentWriteRaceTest`, `SecretsStoreKeyStoreImplTest`, `AppsAPIImplTest`, `AnalyticsValidatorUtilTest`).

One note for anyone reproducing the integration run locally: it needs `-Dmaven.build.cache.enabled=false`. Without it the build cache restores the module and skips `failsafe:integration-test` outright, and the run still prints `BUILD SUCCESS` with no tests having executed.

dsolistorres:
## PR description detail

Moved out of the PR description to keep it readable. The description keeps the summary of proposed
changes and the locking rationale; the full reasoning, container verification, testing and known gaps
live here.

### Root cause in detail

The shared `dotSecretsStore.p12` lives on cluster assets storage and every node reads and writes the same physical file with no lock — but **the concurrency was only the trigger. The data loss was done by a single-threaded path.**

`getSecretsStore()` treated every `IOException` from `KeyStore.load` as unrecoverable corruption and responded by backing up and **deleting** the store. Its retry loop then called `createStoreIfNeeded()` again, which found nothing, wrote an **empty** store, loaded it cleanly, and logged *"KeyStore loaded successfully after 2 tries."*

```java
while (tryCount <= maxLoadTries) {
    final File secretStoreFile = createStoreIfNeeded();   // recreates EMPTY if gone
    try (InputStream in = Files.newInputStream(...)) {
        keyStore.load(in, passwordSupplier.get());
        Logger.info(..., "KeyStore loaded successfully after `%d` tries.");   // reports success
        break;
    } catch (IOException gse) {
        Try.of(() -> handleStorageLoadException(gse));    // backs up + DELETES
        tryCount++;
    }
}
```

The retry was what completed the loss and then reported success — which is why field incidents show a WARN followed by an INFO claiming everything is fine.

**Three ways in**, and the third needs no concurrency at all: a torn read while another node streamed the store back over the destination; a `NoSuchFileException` from the hard-link delete window; or simply a wrong password. The password is `digest(cluster_salt)`, so a salt rotation or a per-node `SECRETS_KEYSTORE_PASSWORD_KEY` override wiped the store on the next read. The backup left behind was encrypted with the password that had just failed — which is exactly why customers report that restoring it does not restore access.

### The four changes in detail

#### 1 — Never destroy on read failure

* The retry now retries the **read** with backoff instead of retrying after a delete, and the store is created at most once per call so the loop cannot manufacture an empty one.
* Integrity failures are distinguished from torn reads. **Verified against the JDK** rather than assumed:

  | Condition | Throws | Classified | Action |
  |---|---|---|---|
  | Wrong password | `IOException` ← `UnrecoverableKeyException` | integrity | stop, preserve |
  | Truncated | `EOFException`, no cause | transient | retry |
  | Zero bytes | `IOException`, no cause | transient | retry |

  Note a zero-byte file raises rather than loading as a valid empty keystore, so the truncate path cannot silently succeed.
* Terminal failures log at ERROR with an actionable message and raise. Recreating is opt-in via `SECRETS_STORE_AUTO_RECREATE` (default **false**). Because `saveValue()` loads before it mutates, raising also stops a caller from writing a nearly-empty store over a file that is intact but merely unreadable by this node.
* `getSecretsStore()` no longer returns an uninitialized `KeyStore` when every attempt fails — a latent bug that surfaced as `KeyStoreException` far from the cause.

#### 2 — Atomic publish

* `saveSecretsStore()` writes the tmp file in the destination's own directory, fsyncs, and renames it into place with `ATOMIC_MOVE` (falling back to a non-atomic replace, with a warning, where the filesystem cannot). A concurrent reader now sees either the whole old store or the whole new one.
* Replaces `FileUtil.copyFile`, which exists for content versioning and either hard-links (deleting the destination first) or truncates the destination and streams the bytes back in. Both leave the shared store empty or partial. This is the standard write-temp → fsync → rename pattern, the same thing Redis uses for RDB snapshots.
* The tmp file is restricted to **0600** before the move, so the published store inherits it. Both files were previously created with the default umask, in a directory shared across the cluster.
* The tmp file is deleted in a `finally`; the old code only removed it on the happy path, leaking a copy of every secret on any failure.

#### 3 — Startup cannot be taken down by the new raise

`Task00050LoadAppsSecrets.forceRun()` tolerates an unreadable store and skips the import. Booting with Apps broken and the credentials intact beats booting with them destroyed.

#### 4 — A node that cannot open the store still works

Traced rather than assumed, because this is the scenario a developer hits when restoring a customer's assets directory whose store was written under a different passkey.

| Path | Handling | Outcome |
|---|---|---|
| Velocity `$secrets` viewtool (page render) | `Try.of(...).getOrElse(Optional.empty())` | page renders, feature skipped |
| PreRender SEO | same | request proceeds |
| SAML IdP config (`DotIdentityProviderConfigurationFactoryImpl:85-87`) | same | no IdP configured |
| Startup task | guarded in this PR | boots |
| **`saveValue` / editing an App** | **propagates** | **loud error, store untouched** |

Most consumers already wrap their read defensively — `Try.of(...).getOrElse(...)` catches `Throwable`, so those degrade unchanged. **Four did not**, and successive review rounds caught them (`94ea3579e6`, `3108727708`):

* `AppsAPIImpl.filterSitesForAppKey()` caught only `DotDataException | DotSecurityException`, so the store's new failure escaped it. That escaped `EMAWebInterceptor.intercept()` — whose call site the interceptor chain (`SimpleWebInterceptorDelegateImpl:192`) does not guard — so an unreadable store returned **500 on every `/api/*` request**, since that interceptor's filter is `API_CALL + "*"`. A new failure mode: the old code wiped the store, `hasAnySecrets` returned false, and this degraded silently. Now catches `Exception`, matching the method's existing "treat as no secrets" contract and covering all six callers including `InitRunner.isUveEnabled()`, which has no catch of its own.
* `SiteAuthValidator` (analytics collection) had the same gap. Its handler already converts failures into `AnalyticsValidationException(INVALID_SITE_AUTH)`, so an unreadable store now becomes a normal validation failure.
* `ContentAnalyticsUtil.getAppSecrets()` / `getSiteKeyFromAppSecrets()` had it too, and this one was the worst of the four: `getAppSecrets()` feeds `EventAnalyticsProxyHelper.buildAuthHeader()`, invoked at `EventAnalyticsProxyHelper:121` **outside** `proxy()`'s try block, and neither `EventAnalyticsProxyResource.proxyGetRequest` nor `proxyEventRequest` catches `RuntimeException` — so an unreadable store meant **500 on every analytics proxy and ingest request** instead of falling back to the bootstrap token.
* `GoogleTranslationService.getFallbackAPIKey()` — a catch whose entire purpose is to fall back to the configured property, which the raise sailed past.

**How they recognise the condition — this took three attempts, so it is worth stating precisely.** The store raises a dedicated `SecretsStoreUnreadableException` (`final`, extending `DotRuntimeException`), but it does not arrive intact: nine `catch (Exception e) { throw new DotRuntimeException(e); }` layers sit between the raise and these callers, so what lands is a *bare* `DotRuntimeException` wrapping the real cause. All four callers therefore catch `DotRuntimeException` and match on the **cause chain** with `ExceptionUtil.causedBy`, rethrowing anything that is not the store:

```java
catch (DotRuntimeException e) {
    if (!ExceptionUtil.causedBy(e, SecretsStoreUnreadableException.class)) {
        throw e;   // database blip, cache failure, programming error — stays loud
    }
    // unreadable store — degrade
}
```

Catching `Exception` masked real bugs; catching the specific type alone compiled, read as the cleanest option, and **never fired in production**. Matching the chain is robust to however many layers re-wrap it. `AnalyticsValidationException` is deliberately not caught, so the analytics pipeline's own signal propagates with its original message.

`getSecrets()` and the `AppsHelper` detail paths deliberately still raise, so an administrator opening an App's configuration against an unreadable store gets a hard error rather than a blank form they might save over. The Apps **list** still reports 0 configurations, because it derives its count from `filterSitesForAppKey` — see *Known behaviour* below. Reads degrade, writes are protected — the instance is usable and the encrypted data is safe. A wrong password also skips the retry backoff entirely, since `isIntegrityFailure()` classifies it as deterministic.

Because the store is now *preserved*, that state persists until an operator fixes it and every read reaches the failure handler again. Two things therefore had to be rate-limited (`72971bf999`), neither of which was a problem while the old code wiped the store and let the next read succeed:

* The actionable ERROR was emitted **per read** — per page render, per login, per content save.
* `sendFailureNotification()` was also called per read, and it **writes an admin notification row** — a steady database write load and an unusable notification tray.

Both are now reported once per `SECRETS_STORE_LOAD_FAILURE_REPORT_INTERVAL_MILLIS` (default 60s), first failure always reported, `debug` in between, throttled per-instance. The destructive `AUTO_RECREATE` branch is **not** throttled — it happens once and must always be logged. Suppressing the report never suppresses the failure; `test_repeatedLoadFailures_areReportedOnceButAlwaysRaise` pins both halves.

### The Apps portlet list, and a contradiction review caught

**This section previously claimed the list "shows every app as 0 configurations". That was wrong, and it contradicted this PR's own container evidence** — the verification table below recorded `Apps portlet read | 500`. Review round 9 caught the disagreement. Both statements were in the description at the same time and I did not notice they disagreed.

The mechanism: `AppsHelper` calls `filterSitesForAppKey(appKey, appsAPI.appKeysByHost().keySet(), user)`. `appKeysByHost()` is evaluated as the **argument**, so it raises *before* the guarded method is entered — that method's cause-chain guard can never protect it. Hence the 500, not the documented 0.

**Now fixed properly, which closes #37061.** The listing degrades to zero counts *and reports why*, using the partial-failure channel `getAvailableDescriptorViewsWithErrors` already had: a dedicated `app-secrets-store-unreadable` entry in the 200 response's `errors` field, telling the administrator the store is unreadable and the secrets are intact. A silent 0 on the secrets screen reads as "your secrets are gone" — the exact misreading that would prompt someone to re-enter them, which is the hazard this PR exists to remove. The detail view still raises; the list is best-effort, the detail authoritative.

**The guard is at the call site in `AppsHelper`, deliberately not inside `appKeysByHost()`.** `removeApp()`, `removeSecretsForSite()` and `exportSecrets()` all read it too. Degrading it at the source would make the two removes silent no-ops and — far worse — make `exportSecrets()` write an **empty backup that reports success**, which an administrator would rely on while rescuing their secrets. That would be strictly worse than the bug being fixed, so those paths keep raising loudly.

### Verified on a container from this branch

**Re-verified after the exception-handling changes.** The first container run predated three successive narrowings of the catch clauses, so it no longer vouched for the shipped code. The second run confirms the mechanism through the real chain — the log shows the failure arriving as a bare `DotRuntimeException` re-wrapped at `SecretCachedKeyStoreImpl:60`, which `ExceptionUtil.causedBy` then matched:

```
ERROR apps.AppsAPIImpl - App secrets store is unreadable; reporting no secrets for `dotsaml-config`
com.dotmarketing.exception.DotRuntimeException: Unable to load the App secrets store...
    at SecretCachedKeyStoreImpl.containsKey(SecretCachedKeyStoreImpl.java:60)   <- the re-wrap
    at AppsAPIImpl.hasAnySecrets(AppsAPIImpl.java:343)
    at AppsAPIImpl.lambda$filterSitesForAppKey$0(AppsAPIImpl.java:426)          <- the catch
```

16 `/api/*` requests all returned **200** with the store failure demonstrably being hit (12 store-failure ERRORs, so not a cache miss), zero uncaught interceptor exceptions, the store byte-identical across two restarts, and the secret fully recovered once the password was restored. A type-only match would have missed that chain and returned 500s.



Built the image from this branch, saved two real App secrets through the REST API, then deliberately set a wrong `SECRETS_KEYSTORE_PASSWORD_KEY` and restarted.

| Check | Result |
|---|---|
| Store **byte-identical** after the break (md5 + size) | ✅ unchanged |
| Backup file created | ✅ **none** |
| Instance boots | ✅ `readyz=200`, `livez=200`, 0 restarts |
| 12 requests across 4 `/api/*` endpoints | ✅ **all 200**, no 500s |
| Save attempt while unreadable | ✅ refused (500), **store md5 unchanged** |
| Apps portlet read | 500 — **superseded**, see below |
| Secrets after restoring the password | ✅ fully recovered, `configurationsCount: 2` |
| Store permissions throughout | ✅ `-rw-------` |
| `.tmp` leftovers at any point | ✅ none |

The `Apps portlet read` row is the observed **pre-fix** behaviour and is what exposed the contradiction above. The `AppsHelper` guard changes it to a 200 with an explicit error entry; that change is covered by `AppsHelperUnreadableStoreTest` and has **not** been re-verified in a container, so the row is left as recorded rather than rewritten.

The API result is the one worth noting: that is finding #1 from review confirmed end-to-end. Without the `filterSitesForAppKey` widening, `EMAWebInterceptor` would have thrown on every one of those calls.

The run also surfaced a residual problem the tests could not: **20 admin notification rows in five minutes**. The ERROR log and the notification had been throttled together on one interval, which is right for a stream and wrong for a persisted row. Fixed in `554d44a9bb` — the notification is now a one-shot latch per JVM, the log keeps its interval.

### Testing

**98 integration tests and 39 unit tests pass, 0 failures, BUILD SUCCESS** (including all 54 in `AppsAPIImplTest`, re-run because this PR now touches `AppsAPIImpl`).

* **`AppsAPIImplUnreadableStoreTest`** (new, 3 unit cases): the `/api/*` path. A wrapped and a double-wrapped unreadable-store failure both degrade; an unrelated `DotRuntimeException` propagates. Verified to fail on both wrapped cases under a type-only match.
* **`SiteAuthValidatorTest`** (new, 5 unit cases): an unreadable store becomes `AnalyticsValidationException(INVALID_SITE_AUTH)` rather than a raw `DotRuntimeException` on the request path; a site-resolution failure passes through as the same instance with its original message; a missing secret still fails validation. Verified to fail with the exact CI signature when the rethrow is disabled.
* **`SecretsStoreWipeRegressionTest`** (new, 9 cases) — the deterministic guard the acceptance criteria asked for, with no threads or timing: a torn store is preserved and raises; a wrong password preserves the store *and the node holding the correct password still reads the canary*; the auto-recreate opt-in still works; repeated saves never truncate and leave no tmp files; the published store is not group/world readable; repeated failures are reported once but always raise.
* **`SecretsStoreConcurrentWriteRaceTest`** — ported from the spike branch as the concurrent companion. It failed reliably on `master`; it passes now.
* **`SecretsStoreKeyStoreImplTest`** — 14/14.
* **`ContentAnalyticsUtilUnreadableStoreTest`** (new, 6 unit cases): the analytics proxy/ingest path. Added after review round 8 found that `ContentAnalyticsUtil` returned 500 on every analytics request. The cases throw the wrapped production shape against the real class, because the pre-existing `SiteAuthValidatorTest` and `EventAnalyticsProxyHelperTest` both `mockStatic(ContentAnalyticsUtil.class)` and so could never observe this class's own exception handling.

Two things worth flagging to reviewers, both found by running the tests rather than by reading:

1. **The ported harness could never have gone green.** `getValue()` returns the value still *encrypted* — decryption is the caching wrapper's job — but the spike compared it directly against the plaintext canary, so `lostReason` tripped on the first read regardless of any race. Fixed while porting; it now decrypts before comparing and distinguishes *canary missing* from *canary changed*.
2. **`Test_Recovery_On_Load_Failure` asserted the old contract**, and its javadoc documented the wipe as intended: *"as the previous file just got wiped-out no secret previously stored will be available now."* So the destructive self-heal was a deliberate, tested decision — this PR reverses it on the grounds that destroying irreplaceable credentials is worse than refusing to open them, with the opt-in preserving the old behaviour. It also had a latent cleanup bug: it saved a `null` "original" password and restored that, so the random password it set was never cleared and every sibling test then ran against a store it could not open. The old auto-wipe rebuilt the store and hid it. Rewritten to be self-contained, plus a companion covering the opt-in.

Run with:

```
./mvnw verify -pl :dotcms-integration -Dcoreit.test.skip=false -Dmaven.build.cache.enabled=false \
  -Dit.test='SecretsStoreWipeRegressionTest,SecretsStoreConcurrentWriteRaceTest,SecretsStoreKeyStoreImplTest,AppsAPIImplTest,AnalyticsValidatorUtilTest'
```

`-Dmaven.build.cache.enabled=false` is not optional. Without it the build cache restores the module and
skips `failsafe:integration-test` outright — the run still prints `BUILD SUCCESS` with no tests executed.

### Known gaps

* Lost updates from concurrent read-modify-write are not addressed — #37054.
* `AtomicMoveNotSupportedException` on a customer's actual EFS mount is unverified. The fallback is a non-atomic replace that logs a warning, so the worst case is no worse than today's behaviour.
* Because SAML degrades to "no IdP configured", a salt mismatch in production means SSO silently stops applying and login falls back to local auth. That is a security-relevant degradation — but it is **identical to today's behaviour** after the store is wiped, and strictly better in that the configuration survives and returns once the salt is corrected. Worth calling out explicitly in the QA note.
* `backupAndRemoveKeyStore` still accumulates unbounded timestamped copies of secret material when the opt-in is enabled. Not newly introduced; worth capping separately.

dsilvam (approved):

dsolistorres:
## Review round 9: the Apps list Medium — fixed, and it closes #37061

This finding was correct: the listing calls
`filterSitesForAppKey(appKey, appsAPI.appKeysByHost().keySet(), user)`, so `appKeysByHost()` raises **while being evaluated as the argument** — before the guarded method is entered. Its cause-chain guard could never have helped. Verified at `AppsHelper:121` and `:268`.

**The contradiction was real.** The description claimed the list "reports 0 configurations" while this PR's own container verification table recorded `Apps portlet read | 500`. Both statements sat in the description at once. Corrected in the detail comment, and the table row is now marked superseded rather than rewritten — the fix is covered by unit tests but has **not** been re-verified in a container.

### The fix

`AppsHelper.getAvailableDescriptorViewsWithErrors` degrades to zero counts and reports the condition through the partial-failure channel the method already had, so the response is a **200** carrying an explicit error:

```json
{ "errors": [ { "errorCode": "app-secrets-store-unreadable",
                "fieldName": "dotSecretsStore.p12",
                "message": "This node cannot read the App secrets store, so configuration
                            counts are unavailable. Existing secrets are intact..." } ],
  "entity":  [ { "key": "dotAnalytics", "configurationsCount": 0 }, ... ] }
```

A silent zero on the secrets screen reads as "your secrets are gone" — the exact misreading that would prompt an administrator to re-enter them, which is the hazard this PR exists to remove. `AppDescriptorLoadError` gained an `errorCode` so this is distinguishable from a malformed descriptor; the existing two-arg constructor keeps the previous `app-descriptor-load-error` code, so YAML errors are unchanged.

### Why not guard `appKeysByHost()` itself, as the finding suggested

That would have been a mistake, and it is the part worth reviewing closely. `appKeysByHost()` also feeds `removeApp:755`, `removeSecretsForSite:933` and `exportSecrets:979` → `collectSecretsForExport:1004`. An empty map there turns both removes into silent no-ops and makes **`exportSecrets()` write an empty backup that reports success** — which an administrator would be relying on precisely while trying to rescue their secrets before fixing a salt mismatch. Strictly worse than the bug being fixed. Those paths keep raising loudly; only the listing degrades.

### The finding's two sub-claims

* **Telemetry** (`TotalSitesUsingDotaiMetricType:58`, `TotalSitesWithAutoIndexContentConfigMetricType:59`) — no change needed, and the stated assumption does not hold. `MetricStatsCollector:219` is a `catch (Throwable)` **inside** the per-metric `for` loop at `:137`: the offending metric records a `MetricCalculationError` and the run continues with every other metric.
* **`listAppKeys()`** — no production caller at all. Only the interface declaration and `AppsAPIImplTest:229-230`. No action.

### Verification

* Unit: **42 tests, 0 failures** — including new `AppsHelperUnreadableStoreTest` (3 cases: degrade-and-report, unrelated `DotRuntimeException` still propagates, readable store keeps real counts).
* Integration: **117 tests, 0 failures** — `AppsResourceTest` 19/19, `AppsAPIImplTest` 54/54, plus the three secrets-store suites and `AnalyticsValidatorUtilTest`.
* Re-run after the branch was rebased onto a newer `main`; the incoming changes were ES reindex work with no overlap with this PR's files.

One note: `collectLoadErrors()` became package-private `@VisibleForTesting` because it re-reads the YAML directories and pulls in ES infrastructure that cannot initialize under a unit test — it fails with an `Error`, which its own `catch (Exception)` does not stop. Test-only seam; no behaviour change.

dsolistorres:
## Review round 10: both Mediums addressed

Both findings were correct and were the same gap: `ConfigService` and `GoogleTranslationService` were the two most recently added degrade branches, and the only ones without unit coverage while the other four each got a dedicated `*UnreadableStoreTest`. Closed in `5f7449e683`.

Both suites throw the **wrapped** production shape, `DotRuntimeException(SecretsStoreUnreadableException(...))`, per the finding's own point — asserting on the unwrapped type would pass while production still failed.

### `ConfigServiceUnreadableStoreTest` (4 cases)

Unreadable store → app reported unconfigured; unrelated `DotRuntimeException` → propagates; readable store → secrets resolved. Plus one the finding didn't ask for: **`invalidLicense_stillPropagates`**. That is the regression CI caught in 2b, and it was covered only by an integration test — it now fails in under a second instead of after a full suite.

### `GoogleTranslationServiceUnreadableStoreTest` (3 cases)

Unreadable store → falls back to the configured `GOOGLE_TRANSLATE_SERVICE_API_KEY`; unrelated `DotRuntimeException` → propagates; readable store → the stored secret still wins.

### Two setup hazards worth flagging, because both produced misleading results first

Neither is a product bug, but both are traps in this codebase that cost real time here and are documented in the test files:

1. **`ConfigService`'s static `INSTANCE` field takes the fork down.** Touching the class runs `public static final ConfigService INSTANCE = new ConfigService()`, which resolves the *real* `AppsAPI` → `AppsAPIImpl` → `DataSource`. With no database that reaches `DbConnectionFactory`, whose `SYSTEM_EXIT_ON_STARTUP_FAILURE` path (default **true**) calls `System.exit`. Surefire then reports `Tests run: 0` and a `BUILD FAILURE` with no test-level diagnostics at all. `APILocator` has to be stubbed *before* the class is first touched.

2. **A false pass I nearly shipped.** `HostAPI.find()` left unstubbed returns `null`; `Try.of(...).getOrElse(systemHost)` does **not** fire on a null *result* (only on failure), so the host stayed null — and `any(Host.class)` does not match null in Mockito 2+. The `getSecrets` stub therefore never applied, Mockito returned its default `Optional.empty()`, and every case took the "no secrets" branch. `unreadableStore_fallsBackToConfiguredKey` passed **without the exception ever being thrown**. Only the other two cases in the same class exposed it — worth remembering that a green degrade test proves nothing unless something in the suite proves the stub actually matches.

### Verification

**59 unit tests, 0 failures**, including the pre-existing `GoogleTranslationServiceTest` (10/10) to confirm no regression there.

### On the two non-blocking notes

Agreed on both, no action taken:

* **`SecretsKeyStoreHelper:284` / non-`IOException` `GeneralSecurityException`** — correct that a `CertificateException` would surface as a plain `DotRuntimeException` and miss the guards. As the note says, this store holds only `SecretKeyEntry` with no certificates, and wrong-password/corruption arrives as `IOException(UnrecoverableKeyException)`, which is classified correctly. Not introduced here and not reachable for this store.
* **`AtomicMoveNotSupportedException` on real EFS/NFS** — still an accepted known gap, listed as such in the PR detail comment.

dsolistorres:
## Review round 11: the log-noise Medium — fixed in `619c4d8a42`

The finding is correct on every step, and I verified each in code rather than taking it on trust:

* `SecretsKeyStoreHelper.loadValueFromStore` caught `Exception` and called `Logger.error(class, e)` — a **full stack trace, un-throttled**.
* `getSecretsStore()` is `loadSecretsStore(true)` with no memoisation, so it re-attempts on every call.
* `SecretCachedKeyStoreImpl.getValue` caches only what the supplier *returns*; a supplier that throws caches nothing, so every `getSecrets()` on every request re-enters and re-logs.
* The asymmetry it points out is real: the sibling `containsKey()` path logs at `debug` (`SecretCachedKeyStoreImpl:59`) and stays quiet. Only the value path leaked.

So for the life of an outage the throttled diagnosis from `handleUnrecoverableLoad` was joined by a second, un-throttled `ERROR`-with-stack **per read** — precisely the flood this PR removes elsewhere. The line predates the PR, but preserving the store instead of wiping it is what makes it fire continuously, so it belongs here.

### Fix

An already-classified `SecretsStoreUnreadableException` now logs at `debug` — it was reported once already, with an actionable message — while anything else reaching that catch is genuinely unexpected and still logs at `ERROR` with its stack. Matched on the cause chain, consistent with the other five sites.

The exception is wrapped exactly as before (`new DotRuntimeException(e)`), deliberately: five callers match this condition with `ExceptionUtil.causedBy`, so the shape they see must not change. That is the one thing that could have gone wrong here, and it's covered — `AppsAPIImplUnreadableStoreTest`, `SiteAuthValidatorTest`, `ContentAnalyticsUtilUnreadableStoreTest`, `AppsHelperUnreadableStoreTest`, `ConfigServiceUnreadableStoreTest` and `GoogleTranslationServiceUnreadableStoreTest` all still pass, and they assert degradation through the real chain.

### Swept for the same mistake elsewhere

Since this was a case of "the guard is fine, the *noise* is the defect", I checked every `Logger.error` in the apps package for the same per-read exposure. The only other one on a read path is the `GeneralSecurityException` branch in `loadSecretsStore` (`:286`), which is not reachable for wrong-password or torn-read — both arrive as `IOException` and route through the throttled handler. That is also the branch the earlier non-blocking note called out as unrealistic for a store holding only `SecretKeyEntry`. No change needed.

### Verification

* Unit: **59 tests, 0 failures**
* Integration: **120 tests, 0 failures** — `AppsAPIImplTest` 54/54, `AppsResourceTest` 19/19, `ConfigServiceTest` 3/3, plus the three secrets-store suites and `AnalyticsValidatorUtilTest`

Issue #36791, opened 2026-07-29
Global/Content search: exact-name and mid-token terms don't match (prefix-only catchall gate)


ihoffmann-dot:
## Problem Statement

`GlobalSearchAttributeStrategy` — the shared class that builds the free-text Elasticsearch query for the **Content Search portlet's main search box**, **Content Drive's keyword search**, the **Relationships field's dynamic search dialog**, and any other consumer of the Lucene Query Builder service — has a prefix + tokenization matching gap.

The mandatory query clause is `+catchall:<term>*` (a **prefix** match on the tokenized `catchall` field). This means a term that lands **mid-token**, or a term that **spans a tokenizer boundary**, never matches — even the **exact full name of a piece of content can fail to match its own name**.

**Impact:** any user of the Search portlet, Content Drive, or the Relationships search dialog who types the exact name of the content they're looking for may not find it. This affects search relevance/completeness across the whole platform (shared code), not a single portlet. Confirmed empirically to affect both the Content Search portlet and Content Drive identically (same underlying strategy).

## Steps to Reproduce

1. Upload/create a FileAsset named `IMG_1004.jpeg`, indexed.
2. In the Content Search portlet (or Content Drive's keyword search), search for `1004` (mid-token — the tokenizer splits `img_1004.jpeg` on `.` into `img_1004` + `jpeg`, but `1004` sits mid-token inside `img_1004`).
   - **Expected:** the file is found.
   - **Actual:** no results.
3. Search for the **exact full filename** `IMG_1004.jpeg`.
   - **Expected:** the file is found (searching for a thing's exact name should always find it).
   - **Actual:** no results — the `.` boundary splits the term across two index tokens, and the query treats the whole typed string as one required prefix.
4. Search for `IMG`, `img`, `IMG_1004`, or `jpeg` (each is a genuine token-prefix).
   - **Actual:** correctly found — these work today.

This was proven with an integration test (`ContentDriveKeywordSearchTest.contentSearchPortlet_hasTheSameLimitation_asContentDrive`) that runs the real `LuceneQueryBuilder` + `ContentletAPI.search` path (the same mechanism `ContentResource.java` uses in production) against a seeded FileAsset, confirming the miss is not specific to Content Drive.

Found while working on #36688 (Content Drive keyword search fix, PR #36767) — that fix intentionally aligned Content Drive with this same shared strategy for consistency and relevance/performance reasons, which is how this pre-existing, platform-wide gap surfaced.

## Acceptance Criteria

[...]

PR #37056, opened 2026-08-13
fix(search): recover exact-name and mid-token matches in the legacy relationship dialog (#37052)

### LINKED ISSUE #37052: Legacy edit content: Relationships search dialog still misses exact-name and mid-token terms (ContentletAjax catchall-only gate)
## Problem Statement

The fix shipped in #36791 / PR #36793 corrected the exact-name and mid-token matching gap in `GlobalSearchAttributeStrategy`, which is the shared strategy used by the Content Search portlet, Content Drive's keyword search, and the **new (Angular) edit content** Relationships field dialog.

However, the **legacy (Dojo/JSP) edit content screen** does not go through that code path at all. Its "Relate" dialog is `dotcms.dijit.form.ContentSelector`, which dispatches over DWR to `ContentletAjax.searchContentletsByUser` — a separate, drifted copy of the same query-building logic that still uses a catchall-only prefix gate.

`dotCMS/src/main/java/com/dotmarketing/portlets/contentlet/ajax/ContentletAjax.java:943-974`:

```java
luceneQuery.append("+" + fieldName + ":" + fieldValueStr + "* ");   // line 956 -> +catchall:img_0004.jpeg*
...
if ("catchall".equals(fieldName)) {
    luceneQuery.append(" title:'" + fieldValueStr + "'^15 ");
    ...
    luceneQuery.append(" title_dotraw:*" + fieldValueStr + "*^5 "); // line 968 -> optional BOOST only
}
```

The `title_dotraw` wildcard is present, but only as an optional boost clause. The mandatory `+catchall:<term>*` gate has already excluded the document, so it never rescues a mid-token or boundary-spanning term.

**Impact:** users on the legacy edit content screen cannot find related content by mid-token substring or by its exact full name — the same defect #36791 fixed everywhere else. Confirmed during QA of #36791.

## Steps to Reproduce

1. Create a Content Type with a Relationships field pointing at another Content Type (e.g. `Images`).
2. Upload a File Asset named `IMG_0004.jpeg`, indexed.
3. Open a contentlet of the first Content Type in the **legacy edit content** screen and click "Relate".
4. Search `0004` (mid-token) -> **Expected:** `IMG_0004.jpeg` appears. **Actual:** 0 results.
5. Search `IMG_0004.jpeg` (exact full name) -> **Expected:** it appears. **Actual:** 0 results.
6. Search `IMG` or `jpeg` (genuine token prefixes) -> correctly found today.
7. Repeat steps 4-5 in the **new** edit content screen -> both work (fixed by #36793).

## Acceptance Criteria

- [ ] In the legacy edit content Relationships dialog, searching a mid-token substring (e.g. `0004` inside `img_0004`) finds the containing content.
- [ ] In the same dialog, searching the exact full name (e.g. `IMG_0004.jpeg`) finds the content.
- [ ] A term that exists in no field still returns 0 results — the broad, unscoped `catchall:*term*` from #36688 must NOT be reintroduced. Any broadened matching stays scoped to `title_dotraw`.
- [ ] No regression to the other `ContentletAjax.searchContentletsByUser` consumers (Content Search portlet's legacy screens, related-content filtering, date filtering). Existing `ContentletAjaxTest` stays green.
- [ ] Query-construction change only — no mapping/analyzer change, no reindex.

## dotCMS Version

Latest from `main`.

## Severity

## Links

- Follow-up to #36791 / PR #36793 — same defect, second code path not covered there.
- Legacy path: `dotCMS/src/main/java/com/dotmarketing/portlets/contentlet/ajax/ContentletAjax.java:943-974`
- Legacy UI: `dotCMS/src/main/webapp/html/js/dotcms/dijit/form/ContentSelector.js:899-923`
- Fixed reference implementation: `dotCMS/src/main/java/com/dotcms/rest/api/v1/content/search/strategies/GlobalSearchAttributeStrategy.java:37-48`
- NA (no Freshdesk ticket)

ihoffmann-dot:
## Problem

#36791 / PR #36793 fixed the exact-name and mid-token matching gap in `GlobalSearchAttributeStrategy` — the shared strategy behind the Content Search portlet, Content Drive's keyword search, and the **new (Angular)** edit content Relationships dialog.

The **legacy (Dojo/JSP)** edit content screen never goes through that code. Its "Relate" dialog is `dotcms.dijit.form.ContentSelector`, which dispatches over DWR to `ContentletAjax.searchContentletsByUser` — a separate, drifted copy of the same logic that still gates on a catchall-only prefix:

```java
luceneQuery.append("+" + fieldName + ":" + fieldValueStr + "* ");   // +catchall:img_0004.jpeg*
...
if ("catchall".equals(fieldName)) {
    ...
    luceneQuery.append(" title_dotraw:*" + fieldValueStr + "*^5 "); // optional BOOST only
}
```

The `title_dotraw` wildcard was already there, but only as an optional boost clause — the mandatory `+catchall:<term>*` had already excluded the document, so it never rescued anything. Surfaced during QA of #36791.

## Fix

Make the `catchall` gate a disjunction, mirroring `GlobalSearchAttributeStrategy`:

```java
+(catchall:<term>*^10 OR title_dotraw:*<term>*^2)
```

- `title_dotraw` is scoped to a single field, so this recovers mid-token and exact-full-value matches **without** reintroducing the broad, whole-document `catchall:*term*` removed in #36688.
- Asymmetric boosts are deliberate and match the reference implementation: a genuine token-prefix hit outranks a raw-substring hit that can land anywhere.
- Query-construction only. No mapping/analyzer change, no reindex.

Two deliberate secondary changes worth a reviewer's eye:

1. **The `catchall` branch is now evaluated before the `contains("-")` branch.** Previously a term containing a dash (e.g. `nordic-skiing`) fell into the no-wildcard branch and behaved differently. `GlobalSearchAttributeStrategy` has no such special case, so this aligns the two. `languageId` keeps its existing handling.
2. **The standalone `title_dotraw:*value*^5` boost was removed**, since the same clause is now part of the mandatory gate at `^2`. Keeping both would duplicate the clause.

## Validation

New integration test `ContentletAjaxTest#test_searchContentletsByUser_globalSearch_matchesMidTokenAndExactFullName`, seeding a File Asset named `IMG_<uniqueToken>_0004.jpeg` (tokenizes to `img_<uniqueToken>_0004` + `jpeg`):

| Search term | Before | After | Matches via |
|---|---|---|---|
| `IMG_<uniqueToken>` (genuine token prefix) | ✅ | ✅ | catchall prefix |
| `<uniqueToken>` (mid-token) | ❌ | ✅ | `title_dotraw` substring |
| `IMG_<uniqueToken>_0004.jpeg` (exact full name, spans the `.`) | ❌ | ✅ | `title_dotraw` substring |
| `zz<uniqueToken>` (in no field) | 0 results | 0 results | negative control — no broad catchall |

The term is unique per run so the assertions hold against the shared test index.

**Red/Green verified.** With the production change reverted the test fails on the mid-token assertion (`expected:<1> but was:<0>`) while the preceding token-prefix assertion passes — so it isolates this defect rather than failing for an unrelated reason. With the fix, `ContentletAjaxTest` is 10/10 green (9 pre-existing + 1 new). The class is already registered in `MainSuite2a`, so no suite changes were needed.

## Scope

Closes #37052. Follow-up to #36791 / PR #36793 — same defect, second code path.

## Out of scope / follow-ups

- Collapsing the duplicated query-building logic by having `ContentletAjax` call `GlobalSearchAttributeStrategy` directly (as `BrowserAPIImpl` already does). Removes the drift at the root, but touches more legacy surface than this fix warrants.

This PR fixes: #37052

### DIFF
--- a/dotCMS/src/main/java/com/dotmarketing/portlets/contentlet/ajax/ContentletAjax.java
+++ b/dotCMS/src/main/java/com/dotmarketing/portlets/contentlet/ajax/ContentletAjax.java
@@ -950,14 +950,29 @@ public class ContentletAjax {
 
 									fieldValueStr = fieldValueStr.replaceAll(specialCharsToEscape, "\\\\$1");
 
-							        if(fieldName.equals("languageId") || fieldValueStr.contains("-")){
+							        if("catchall".equals(fieldName)) {
+										// Mandatory gate: match either a catchall token PREFIX (fast, existing
+										// behavior) OR the raw title via wildcard. Unlike catchall (which
+										// aggregates every field of the document), title_dotraw is scoped to a
+										// single field, so this alternative recovers mid-token and
+										// exact-full-value matches -- e.g. a file named "IMG_0004.jpeg"
+										// tokenizes to "img_0004"+"jpeg", so "0004" or a full-name search never
+										// satisfies a catchall-only prefix gate -- without reintroducing an
+										// unscoped, whole-document wildcard like the old broad catchall:*value*.
+										// Boosts are deliberately asymmetric: catchall (a real token-prefix hit)
+										// outranks title_dotraw (a raw substring hit that can land anywhere,
+										// including mid-word). Mirrors GlobalSearchAttributeStrategy, which is
+										// the equivalent gate for the new Content Search / Content Drive path.
+										luceneQuery.append("+(catchall:" + fieldValueStr + "*^10 OR title_dotraw:*"
+												+ fieldValueStr + "*^2) ");
+									} else if(fieldName.equals("languageId") || fieldValueStr.contains("-")){
 										luceneQuery.append("+" + fieldName +":" + fieldValueStr + " ");
 									}else{
 										luceneQuery.append("+" + fieldName +":" + fieldValueStr + "* ");
 									}
-							        
+
 							        if("catchall".equals(fieldName)) {
-							           
+
 							            luceneQuery.append(" title:'" + fieldValueStr + "'^15 ");
                                         final String[] titleSplit = fieldValueStr.split("[,|\\s+]");
                                         if (titleSplit.length > 1) {
@@ -965,7 +980,6 @@ public class ContentletAjax {
                                                 luceneQuery.append(" title:" + term + "^5 ");
                                             }
                                         }
-							            luceneQuery.append(" title_dotraw:*" + fieldValueStr + "*^5 ");
 							        }
 							        
 							        
--- a/dotcms-integration/src/test/java/com/dotmarketing/portlets/contentlet/ajax/ContentletAjaxTest.java
+++ b/dotcms-integration/src/test/java/com/dotmarketing/portlets/contentlet/ajax/ContentletAjaxTest.java
@@ -17,6 +17,8 @@ import com.dotcms.contenttype.model.type.SimpleContentType;
 import com.dotcms.datagen.ContentTypeDataGen;
 import com.dotcms.datagen.ContentletDataGen;
 import com.dotcms.datagen.FieldDataGen;
+import com.dotcms.datagen.FileAssetDataGen;
+import com.dotcms.datagen.FolderDataGen;
 import com.dotcms.datagen.LanguageDataGen;
 import com.dotcms.languagevariable.business.LanguageVariableAPI;
 import com.dotcms.repackage.org.directwebremoting.WebContext;
@@ -33,6 +35,7 @@ import com.dotmarketing.portlets.contentlet.business.ContentletAPI;
 import com.dotmarketing.portlets.contentlet.model.Contentlet;
 import com.dotmarketing.portlets.contentlet.model.IndexPolicy;
 import com.dotmarketing.portlets.folders.business.FolderAPI;
+import com.dotmarketing.portlets.folders.model.Folder;
 import com.dotmarketing.portlets.languagesmanager.model.Language;
 import com.dotmarketing.portlets.structure.model.Relationship;
 import com.dotmarketing.portlets.structure.model.Structure;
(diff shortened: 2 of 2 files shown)


dario-daza (approved):

dotCMS/src/main/java/com/dotmarketing/portlets/contentlet/ajax/ContentletAjax.java
```diff
@@ -950,22 +950,36 @@ else if( fieldbcontentname.startsWith("date") ){
...
+										// exact-full-value matches -- e.g. a file named "IMG_0004.jpeg"
+										// tokenizes to "img_0004"+"jpeg", so "0004" or a full-name search never
+										// satisfies a catchall-only prefix gate -- without reintroducing an
+										// unscoped, whole-document wildcard like the old broad catchall:*value*.
+										// Boosts are deliberately asymmetric: catchall (a real token-prefix hit)
+										// outranks title_dotraw (a raw substring hit that can land anywhere,
+										// including mid-word). Mirrors GlobalSearchAttributeStrategy, which is
+										// the equivalent gate for the new Content Search / Content Drive path.
+										luceneQuery.append("+(catchall:" + fieldValueStr + "*^10 OR title_dotraw:*"
+												+ fieldValueStr + "*^2) ");
```
nollymar, line 967: how did you choose the ^10 and ^2?

ihoffmann-dot: Good question! Carried over from GlobalSearchAttributeStrategy ([#36793](https://github.com/dotCMS/core/issues/36793)) so both paths rank the same. Happy to change the spread if necessary, but I think it should be done in both places

ihoffmann-dot: **To add some more context:**

The values are added in what that class already had: 
^10 for a token prefix
^2 for the raw substring, which can land mid word and must never outrank a real match

That ordering is not just an assumption, GlobalSearchAttributeStrategyMatchingTest asserts it, so the numbers are arbitrary but the ranking they produce is tested, of course

nollymar (approved):

PR #37073, opened 2026-08-14
feat(content-drive): trim quick actions to the old-search bulk operations and wire Push Publish

rjvelazco:
> **Draft / spike.** One of two directions explored for Content Drive's Action Center; #37075 is the other. This is the one being taken forward. Still a draft: no linked issue yet, and no manual QA against a running instance.

## What this does

Narrows **Quick Actions** to the bulk operations the old content search offered *outside* its workflow dropdown, and wires Push Publish.

| Row | State |
|---|---|
| Lock | wired |
| Unlock | wired — keeps the "locked by another user" warning |
| Add to Bundle | wired — keeps the bundle-picker step |
| **Push Publish** | **wired** — gated on push publish environments |
| Refresh | disabled placeholder, `Work in progress` badge |
| ~~Publish / Unpublish / Archive / Unarchive / Delete~~ | **removed** |

## Why remove the five state actions

They are already offered by the **Workflow Actions** section below, as the *scheme's own* actions.

That is not a duplicate — it is a more correct answer. A quick action fires `POST /v1/workflow/actions/default/fire/{systemAction}`, which resolves through the scheme's Default Actions mapping. If a content type's scheme maps `PUBLISH` to something other than a plain publish, the quick-action row labelled "Publish" and the workflow row labelled "Publish" do different things. Two rows with the same label and different behaviour is worse than one row that resolves correctly.

Lock and Unlock stay because they are genuinely per-user state on the version info, not a workflow transition — they have no scheme action to defer to. `WorkflowAPI.SystemAction` documents this: they carry no actionlet, and mapping them has no effect.

### Gate

`PushPublishService.getEnvironments()` runs when the dialog opens. The row stays disabled with an explanatory tooltip until it answers with at least one environment.

Deliberately **not** the same state as Refresh's coming-soon badge: nothing is missing from dotCMS here, something is missing from the *configuration*, and the fix belongs to an administrator. The row says which.

Fails closed. An unresolved or failed lookup reads as "none" — offering a push with nowhere to go fails at the servlet with a message the user cannot act on.

> **One deliberate difference from the JSP.** `view_bulk_actions_inc.jsp` gates on `sendingEndpoints` (publishing endpoints existing). This gates on **environments reachable by the user's role**, which is the stricter question and the one the form itself needs: endpoints can exist with no environment this role can send to, and then the picker would open empty.

### Form

Reuses `DotWorkflowPushPublishComponent` — the same section the workflow-action path renders — so a push publish is collected identically in both dialogs. `$configureKinds` returns `['pushPublish']` for the row, putting it through the existing `pick → configure → preview → execute` flow.

### Fire

New `PushPublishService.pushPublishAssets`, posting to `RemotePublishAjaxAction/cmd/publish` — the same servlet the old search's Push Publish button hits.

Kept separate from the existing `pushPublishContent` because of **date handling, not count**. `DotPushPublishData` carries whole dates that the service splits into the servlet's `remotePublishDate` / `remotePublishTime` pair; the form emits them *already* split. Routing through the existing method would mean recombining two strings into a `Date` for the service to take apart again — losing the timezone the user picked on the way.

`DotWorkflowPushPublishValue` moved down to `@dotcms/dotcms-models` so data-access can accept it without depending on `@dotcms/ui`, and is re-exported from ui where callers have always imported it.

Sends **identifiers**, deduped, like Add to Bundle: a push sends the asset, so language versions of one contentlet are a single entry. The servlet splits `assetIdentifier` on `,` (`RemotePublishAjaxAction`: *"Support for multiple ids in the assetIdentifier parameter"*), so bulk needs no new endpoint.

The servlet answers 200 for its own failures, so a response without a numeric `errors` goes to the error handler rather than being reported as a success — the same guard Add to Bundle uses.

### What the JSP told us

- **No per-contentlet precondition.** `pushPublishSelectedContentlets()` is just `pushHandler.showDialog(getSelectedInodes())` — the whole selection, unfiltered by live/archived/locked. The row therefore counts every contentlet.
- **The enterprise-license gate is deliberately not reproduced.** That requirement no longer applies.

## Why the placeholders are rendered rather than hidden

Refresh is an action the old search does offer. Leaving it out entirely makes Content Drive look like it dropped it; a disabled row with a `Work in progress` badge and a tooltip is the honest state. It is backed by `_bulkrefresh`, which streams progress over SSE and is not job-backed, so it cannot reuse the synchronous `bulkFire` path the other quick actions run on.

It counts over the whole selection rather than sitting at `0` — a `0` would read as "does not apply to these items", a different and untrue claim from "not built yet". The row is disabled by its own flag, not by its count.

## Commits

1. **Trim the quick actions** — list, new IDs, `comingSoon` flag, template, specs.
2. **Drop the machinery the removed rows were the only users of** — `danger` (permanently `false` once the destructive rows left, plus its two red-styling branches), the entire confirmation path (`confirmMessage`, the `ConfirmationService` provider/injection, `ConfirmDialogModule`, `<p-confirmDialog>`, and `fireQuickAction`, which existed only so the confirm and direct branches could share a path), their tests, and doc comments still describing Publish and Delete as quick actions.
3. **Prettier formatting.**
4. **Wire Push Publish** — gate, form, fire, and the last two stale comments.

## i18n

Two keys added: `content-drive.action-center.coming-soon` and `content-drive.action-center.no-environments`. Labels reuse the legacy `Remote-Publish` and `Refresh` keys the old search already uses, so the parity is free in every translation. The badge reuses the existing `content-drive.work-in-progress`.

## Known leftovers — deliberate, flagging for review

Two keys are orphaned repo-wide after commit 2 (`core-web` and `dotCMS/src` both grep clean):

- `content.drive.worflow.action.delete.confirm`
- `content-drive.action-center.confirm.header`

Left in `Language.properties` on purpose: removing i18n keys is a different risk profile, since customer overrides and plugins can reference them. Happy to delete them if reviewers prefer.

## Testing

- `pnpm nx test portlets-content-drive` — **1253 passed**, 32 suites
- `pnpm nx test data-access` — **758 passed**
- `pnpm nx test ui` — **820 passed**
- `pnpm nx run-many -t lint -p portlets-content-drive data-access ui dotcms-models` — clean

New coverage: the environments gate (available / none / lookup fails / still pending), the configure step opening on the row, Continue gated on form validity, the fired payload being deduped identifiers plus the collected settings, rows unchecked in the preview being honoured, the refusal to fire without settings, the servlet body shape and its comma-joining, and the five state actions being absent from the list.

## Not done here

**Manual QA against a running instance.** Worth covering:

- an instance with no push publish environment for your role → Push Publish disabled, tooltip names the configuration
- after adding an environment → row enabled, form opens, push lands in the target environment
- a multi-language selection → one bundle entry per identifier, not per language version

## Comparison with #37075

| | This PR | #37075 |
|---|---|---|
| Publish/Unpublish/Archive/Unarchive/Delete | removed from Quick Actions | kept, gated on the workflow mapping |
| Closes the raw-API bypass | yes, by removal | yes, by gating |
| New requests per dialog open | 1 (environments) | 1 per content type + 1 per scheme |
| Backend follow-up | none | optional: a resolved-mapping endpoint |

### DIFF
--- a/core-web/libs/data-access/src/lib/push-publish/push-publish.service.spec.ts
+++ b/core-web/libs/data-access/src/lib/push-publish/push-publish.service.spec.ts
@@ -193,4 +193,79 @@ describe('PushPublishService', () => {
         );
         req.flush(mockResponse);
     });
+
+    describe('pushPublishAssets', () => {
+        const SETTINGS = {
+            whereToSend: 'env1,env2',
+            iWantTo: 'publish' as const,
+            publishDate: '2026-09-01',
+            publishTime: '10-00',
+            expireDate: '2026-10-01',
+            expireTime: '23-59',
+            filterKey: 'hol',
+            timezoneId: 'Costa Rica'
+        };
+
+        it('should post the split date fields through untouched', () => {
+            // The whole point of this method over `pushPublishContent`: the payload arrives already
+            // in the servlet's shape, so nothing is re-parsed and the chosen timezone survives.
+            spectator.service.pushPublishAssets('id-1', SETTINGS).subscribe((items) => {
+                expect(items).toEqual(mockResponse);
+            });
+
+            const req = spectator.expectOne(
+                '/DotAjaxDirector/com.dotcms.publisher.ajax.RemotePublishAjaxAction/cmd/publish',
+                HttpMethod.POST
+            );
+
+            expect(req.request.body).toBe(
+                'assetIdentifier=id-1&remotePublishDate=2026-09-01&remotePublishTime=10-00' +
+                    '&remotePublishExpireDate=2026-10-01&remotePublishExpireTime=23-59' +
+                    '&timezoneId=Costa Rica&iWantTo=publish&whoToSend=env1,env2' +
+                    '&bundleName=&bundleSelect=&filterKey=hol'
+            );
+            req.flush(mockResponse);
+        });
+
+        it('should send several identifiers comma-joined', () => {
+            // `RemotePublishAjaxAction` splits `assetIdentifier` on "," — bulk needs no new endpoint.
+            spectator.service.pushPublishAssets('id-1,id-2', SETTINGS).subscribe();
+
+            const req = spectator.expectOne(
+                '/DotAjaxDirector/com.dotcms.publisher.ajax.RemotePublishAjaxAction/cmd/publish',
+                HttpMethod.POST
+            );
+
+            expect(req.request.body).toContain('assetIdentifier=id-1%2Cid-2');
+            req.flush(mockResponse);
+        });
+
+        it('should omit the filter key when there is none', () => {
+            // Expiring takes no filter, so an empty value is a real case. The servlet reads the
+            // parameter straight into `getFilterDescriptorByKey`, so it is left out rather than
+            // sent blank.
+            spectator.service.pushPublishAssets('id-1', { ...SETTINGS, filterKey: '' }).subscribe();
+
+            const req = spectator.expectOne(
+                '/DotAjaxDirector/com.dotcms.publisher.ajax.RemotePublishAjaxAction/cmd/publish',
+                HttpMethod.POST
+            );
+
+            expect(req.request.body).not.toContain('filterKey');
+            req.flush(mockResponse);
+        });
+
+        it('should record the environments it last pushed to', () => {
+            spectator.service.pushPublishAssets('id-1', SETTINGS).subscribe();
+
+            spectator
+                .expectOne(
+                    '/DotAjaxDirector/com.dotcms.publisher.ajax.RemotePublishAjaxAction/cmd/publish',
+                    HttpMethod.POST
+                )
+                .flush(mockResponse);
+
+            expect(spectator.service.lastEnvironmentPushed).toEqual(['env1', 'env2']);
+        });
+    });
 });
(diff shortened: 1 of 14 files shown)


rjvelazco:
Route 2 is now open for comparison: #37075 — keeps the state actions and gates them on the workflow Default Action mapping, rather than removing them.

Both are drafts and are meant to be read against each other:

| | #37073 (this one) | #37075 |
|---|---|---|
| Publish/Unpublish/Archive/Unarchive/Delete | removed from Quick Actions | kept, gated on the mapping |
| Closes the raw-API bypass | yes, by removal | yes, by gating |
| New requests per dialog open | none | 1 per content type + 1 per scheme |
| Diff size | −298 / +285 | +903 / −82 |
| Backend follow-up | none | optional: a resolved-mapping endpoint would collapse the fan-out to one call |

rjvelazco:
Closing this in favor of **#37081**.

This PR did its job as a spike: it was one of two directions explored for Content Drive's Quick Actions, alongside #37075, which kept the state actions and gated them on the workflow default action mapping. @nollymar decided to go with the parity approach explored here.

Now that the decision is made, the work has been promoted to a properly tracked branch:

| | |
|---|---|
| Issue | #37080 — under the **CD: Workflow Actions** epic (#33999) |
| PR | #37081 |
| Branch | `issue-37080-workflow-center-trim-quick-actions-to-the-old-search-bulk-operations-and-wire-push-publish` |

**No code changed.** #37081 carries the same four commits from this branch, unmodified. The only differences are the branch name (now following the `issue-{number}-{description}` convention), the linked issue, and a PR description that drops the spike framing.

The review discussion belongs on #37081 from here. This branch stays as the record of the spike.

PR #37078, opened 2026-08-17
#37070 #37071: feat(roles): add GET /v1/roles/{roleid}/users and childCount/userCount on RoleView

### LINKED ISSUE #37070: [Task] Roles API: add `GET /v1/roles/{roleId}/users` returning the list of directly-granted users
## Context

The Angular Roles and Tools portlet (FE task #36930) needs to list the users granted a given role, with **email** and other user detail. Today the FE combines two paths — neither ideal:

- `GET /v1/users/filter?roleKey=X` — fast + carries email, but only accepts `roleKey`. Roles created via the UI without an explicit key (or where `roleKey` differs from `roleId`) can't be filtered that way.
- `GET /v1/roles/{roleId}/rolehierarchyanduserroles` — accepts `roleId` but returns `List<Role>` (not `List<User>`). Each user is represented by their individual user-role object, which does **not** carry `emailAddress`. The FE can identify who the user is but shows an empty email column.

The result: some rows in the Users tab render without email, purely because of endpoint shape.

A dedicated, standard REST endpoint keyed on `roleId` and returning the platform `User` DTO closes the gap without special-casing UI concerns.

## Endpoint Spec

| Field | Value |
|---|---|
| HTTP | `GET` |
| Path | `/v1/roles/{roleid}/users` (path template is lowercase `roleid`, matching the existing `GET /{roleid}` and the membership endpoints in #37077 so the OpenAPI spec keeps one template spelling per path; the real URL is unaffected) |
| Class#method | `com.dotcms.rest.api.v1.system.role.RoleResource#loadUsersByRoleId` (new) |
| Path param | `roleId` — target role id |
| Query params | `filter` (search by name / email), `page` (default 1), `per_page` (default 40), `orderBy`, `orderDirection` |
| Response | `PaginationUtil.getPageView` returning `ResponseEntityPaginatedDataView`: `entity` is the list of standard dotCMS user maps including `emailAddress`, and a `pagination` object (currentPage, perPage, totalEntries) comes in the response body. `getPage` and header-based pagination are deprecated on main, so the endpoint uses the current envelope. FE task #36930 must parse this shape. |
| Auth gates | `requiredBackendUser=true`, `requiredPortlet("roles")`. This is intentionally stricter than the sibling GETs on this resource and `/v1/users/filter` (backend user only) because the response contains user PII. Precedent: `loadUserRoles` on this resource already gates PII behind portlet access. Loosening later is backward compatible, tightening later is not, so start strict. |
| OpenAPI | `@Operation` annotation required; regenerated `openapi.yaml` committed |

### Response shape

Use the existing `User` map / DTO that `/v1/users/filter` already returns — same `userId`, `firstName`, `lastName`, `emailAddress`, `active`, `gravitar`, `admin`, etc. No new DTO required, no UI-specific fields added.

## Behavior

- Load the target role by `roleId`; 404 if not found.
- Return users **directly granted** to `roleId` — the equivalent of `RoleAPI.findUsersForRole(role, false)`, mapped through the existing `User` serialization.
- **Inheritance is intentionally NOT included** in this endpoint. Clients that need "users of this role and all descendants would inherit" (dotCMS grants flow downward — a user granted an ancestor role is also considered a member of every descendant) walk the ancestor chain themselves via `parent` on `RoleView` and call this endpoint per role. Keeping inheritance out of the endpoint lets the API stay a pure "role → its direct members" resource, symmetric with other v1 role endpoints.
- Support `filter` / pagination / sorting via the same paginator that powers `/v1/users/filter` — one implementation, consistent behavior.
- A user-role id (`role.isUser()`) is served like any other role: direct grants as recorded in `users_cms_roles`. Every user-role gets a self-grant row at creation (`RoleFactoryImpl.addUserRole`), so the response is the user themself. No special casing, unlike `rolehierarchyanduserroles`. Covered by an integration test.

## Why this shape (and not the alternatives)

- **`GET /v1/users/filter?roleId=X`** — could work by extending the existing filter, but `roleId` filtering has different validation and error semantics than `roleKey`. A separate resource is cleaner: `/roles/{id}/users` matches the natural "children of a parent" REST idiom and doesn't overload the filter endpoint.
- **A new `List<RoleMemberView>` DTO with `grantedFromRoleId` + `grantedFromRoleName`** — was considered and rejected. Denormalizes the ancestor role name into every row (stale on rename), couples the response shape to a specific UI (the "Granted From" chip on the Angular Users tab), and leaks presentational concerns into a general-purpose API. Non-UI consumers (CLIs, integrations, reports) get fields they didn't ask for and payload they can't trust.
- **Include `inherited=true` mode in the endpoint** — was considered and rejected. Inheritance is a client concern: some consumers want direct grants only (compliance reporting, audit trails), some want the effective grant list. A single flag muddles that. The ancestor walk is O(depth) client-side and each hop is a call to *this* endpoint — the semantics stay simple.
- **Reshape `/v1/roles/{roleId}/rolehierarchyanduserroles`** — was considered and rejected. That endpoint already ships and has existing callers; changing its response shape risks silent breakage. A new endpoint is safer.

## Acceptance Criteria

- [ ] `GET /v1/roles/{roleId}/users` implemented with the spec above
- [ ] Response uses the same `User` shape (or map) that `/v1/users/filter` already exposes — no new DTO
- [ ] `@Operation` OpenAPI doc added
- [ ] `swagger-maven-plugin` regenerated `openapi.yaml` committed
- [ ] Integration test covering:
  - Role with direct grants → each `User` row carries `emailAddress`, `firstName`, `lastName`
  - Role without a roleKey with direct grants → users returned (this is the main regression this ticket fixes)
  - Role without any direct grants → returns an empty list (users granted only via ancestors are NOT in the response — inheritance is a client concern)
  - `filter` search matches on name / email
  - Pagination boundaries
  - Missing role (→ 404), non-admin caller (→ 403), missing portlet gate (→ 403)
- [ ] License + role gating preserved
- [ ] Reuse `UserPaginator`. This requires one change first: `UserFactoryImpl.appendRoleFilter` matches on `cms_role.role_key` and binds `role.getRoleKey()`, so a role without a key binds null and matches nothing. Without this change the new endpoint returns an empty list for the exact roles this ticket is about. Change the EXISTS subquery to `ur.role_id IN (...)` binding `role.getId()`. Keys are unique (enforced at save) and both existing callers pass DB-loaded roles, so behavior is preserved. The same helpers serve `getCountUsersByName`, so one change covers list and count. Bind defensively: use the id when set, otherwise resolve the key to an id, to protect external `UserAPI` callers.
- [ ] Regression integration tests for both existing consumers of the changed filter (`GET /v1/users/filter?roleKey=` and `GET /v1/users/loginAsData`), written and green against main before the change, then carried unchanged across it
- [ ] FE task #36930 updated to consume `roleId` instead of `roleKey` for the ancestor walk; the `loadRoleMembersById` fallback (which today uses `rolehierarchyanduserroles` and returns Roles instead of Users) is deleted

## References

- Spike audit: #36929
- Parent epic: #36909
- FE consumer: #36930
- Related BE tasks: #36936 (PUT role), #36937 (grant user), #36938 (bulk remove), #36939 (delete role)
- Reference logic already in the codebase: `RoleAPIImpl.findUsersForRole(role, false)`, `UserPaginator` (accepts `List<Role>` via `ROLES_PARAM`)

### LINKED ISSUE #37071: [Task] Roles API: expose `childCount` and `userCount` on `RoleView` for scalable tree rendering
## Context

The Angular Roles and Tools portlet (FE task #36930) renders a `p-tree` of the entire role hierarchy. Two UI signals depend on data that isn't in the `RoleView` response today:

- **Folder vs. shield icon** — parents (roles with children) render with a folder icon, leaves with a shield. We currently know a node has children only when its `roleChildren` array is populated in the response. Since `GET /v1/roles?loadChildrenRoles=true` only hydrates two levels deep, any node at level 2 or beyond starts as an unknown (chevron shown, actual leaf status only confirmed after a lazy-load click). Users see a chevron they can click, expand, and get an empty branch — poor UX.
- **User-count badge (`👤 N`)** — the design shows the count of users granted each role directly. There is no endpoint that returns this metadata today; the badge is currently hidden in the FE for that reason.

Adding two small integer fields to `RoleView` (`childCount`, `userCount`) closes both gaps at scale.

## Scale considerations

This ticket is driven by customers with **hundreds of roles**. The spec below is designed so nothing here degrades with role count:

- `childCount` costs **zero additional queries** — `Role.roleChildren` (a `List<String>` of child ids) is already populated on the domain model. Serializing its `.size()` is free.
- `userCount` costs **one aggregated query per response**, not one per role. Implement via a single grouped `SELECT` over `users_cms_roles` filtered by the set of role ids being serialized. Result is materialized into a `Map<String, Integer>` and each `RoleView` picks its count out of the map at serialization time. Scales linearly with rows returned to the client, not with total roles in the system.

These fields also unlock an optimization on the FE: today the portlet loads roots + level 1 with `loadChildrenRoles=true`, materializing thousands of nodes in the initial payload for large customers. Once each root exposes `childCount`, the FE can switch to `loadChildrenRoles=false` for the initial call and lazy-load subtrees only when the admin expands them — the icon still renders correctly upfront because it can be derived from `childCount > 0`.

## Fields Spec

Add to `com.dotcms.rest.api.v1.system.role.RoleView`:

| Field | Type | Semantics |
|---|---|---|
| `childCount` | `int` | Number of direct child roles. Equals `role.getRoleChildren().size()`. Never null. |
| `userCount` | `int` | Number of users directly granted this role. Does not include users inherited via ancestor grants (that is a client-side computation and would require walking the hierarchy — see #37070 for the related discussion). Also excludes users hidden from the user listing (the system and anonymous users, the default user, and users flagged delete_in_progress), so the tree badge always equals the Users tab total returned by `GET /v1/roles/{roleid}/users`. Never null. |

Both fields are always populated on every endpoint that returns `RoleView`. There are exactly three: `GET /v1/roles`, `GET /v1/roles/{roleId}`, and `GET /v1/roles/users/{userIdOrEmail}`. Note: `GET /v1/roles/_search` returns `SmallRoleView` and `POST /v1/roles` returns raw `role.toMap()`; both are out of scope, since migrating them onto `RoleView` would be a breaking response-shape change. If the in-flight PUT endpoint (PR #37012) ends up returning `RoleView`, it inherits the fields automatically; coordinate rebases.

## Implementation notes

- `childCount`: derive from `role.getRoleChildren()`. If the list is null, treat as 0. No new queries.
- `userCount`: batch resolution. When a request is going to serialize N `RoleView` instances, issue **one** query grouped by role id restricted to those N ids. Pass the resulting `Map<String, Integer>` to `RoleView`'s constructor (or to a builder / mapper) so each view picks up its own count without additional round trips.
- OpenAPI: the schema for `RoleView` needs both fields declared as required (`int32`, minimum 0). Regenerate `openapi.yaml`.
- Do **not** compute `userCount` inheritance-aware — that would require walking `findRoleHierarchy` per role and running per-hierarchy queries, which does not scale. Inheritance is a client concern.
- The three endpoints currently build `RoleView`s inline while iterating. Restructure to: collect all role ids in the response (parents plus children), run one `countUsersByRoleIds` query, then build the views from the resulting map. Add the aggregate as a new `RoleAPI`/`RoleFactory` method: `Map<String, Integer> countUsersByRoleIds(Collection<String>)`, implemented as `count(distinct user_id) ... group by role_id` with chunked IN lists, following the factory's existing chunking convention.
- Do not store counts on cached `Role` objects. `RoleCache` entries are long-lived; counts are computed per response.
- `userCount` for a user-role is 1 (the self-grant row created by `addUserRole`). This does not affect the tree, since `findRootRoles` excludes the Users branch, but it is the documented behavior for other consumers.

## Acceptance Criteria

- [ ] `RoleView` exposes `childCount: int` and `userCount: int`
- [ ] `childCount` derived from `Role.roleChildren.size()`. No additional queries by construction (the list is already on the loaded `Role` object); the integration test asserts correct values
- [ ] `userCount` populated via a single aggregated query per response. Enforced by making `countUsersByRoleIds` the only code path (verified in code review) plus a functional integration test asserting correct counts across a response with many roles. A literal query-count assertion is not practical; the integration test setup has no SQL counting harness.
- [ ] Both fields present on every endpoint that returns `RoleView`: `GET /v1/roles`, `GET /v1/roles/{roleId}`, `GET /v1/roles/users/{userIdOrEmail}` (`_search` and `POST /v1/roles` do not return `RoleView` and are out of scope)
- [ ] `@Schema` annotations added; `swagger-maven-plugin` regenerated `openapi.yaml` committed
- [ ] Integration test coverage:
  - Role with N children → `childCount === N`
  - Role with M direct user grants → `userCount === M`
  - `userCount` does not include users inherited from ancestor roles (grants on child roles do not count for the parent's `userCount`)
  - Response containing many `RoleView`s returns correct `userCount` for every row (batching itself is enforced in code review, see the AC above)
  - Empty children / empty user list → count fields are `0` (not null)
- [ ] License + role gating unchanged
- [ ] FE task #36930 updated to: (1) render the folder / shield icon from `childCount > 0` instead of the local `#expandedRoles` heuristic, (2) render the `👤 N` badge from `userCount`, and (3) switch the initial `loadRootRoles` call to `loadChildrenRoles=false` for scalable large-customer loads

## References

- Spike audit: #36929
- Parent epic: #36909
- FE consumer: #36930
- Related BE tasks: #36936, #36937, #36938, #36939, #37070
- Reference model: `com.dotmarketing.business.Role#getRoleChildren` (source of `childCount`), `com.dotmarketing.business.RoleAPIImpl#findUsersForRole(role, false)` semantics (source of `userCount`)

hassandotcms:
### What
Backend for the Angular Roles and Tools portlet (epic #36909):

- `GET /v1/roles/{roleid}/users` (#37070): paginated list of users directly granted a role, standard user maps with `emailAddress`. Direct grants only; inheritance stays a client
concern via the ancestor walk on `RoleView.parent`. Gated by backend user + roles portlet (response carries PII). 404 on missing role.
- `childCount` and `userCount` on `RoleView` (#37071): present on the three endpoints that return this DTO (`GET /v1/roles`, `GET /v1/roles/{roleid}`, `GET /v1/roles/users/{userIdOrEmail}`). `childCount` derives from the already-loaded `Role.roleChildren` (no extra queries). `userCount` resolves through the new `RoleAPI.countUsersByRoleIds`: one grouped `users_cms_roles` query per response, chunked IN lists, never stored on cached `Role` objects.

### Prerequisite fix

`UserFactoryImpl.appendRoleFilter` matched roles by `role_key`, so a role without a key matched no users. The EXISTS subquery now matches `ur.role_id`, with a defensive key-to-id fallback for hand-built `Role` objects from external `UserAPI` callers. Affects `/v1/users/filter?roleKey=` and `/v1/users/loginAsData`: behavior preservation is demonstrated, not assumed. Regression tests were written and green against the previous SQL, then carried unchanged across the change.

### Tests

22 new integration tests (TDD, red before each implementation), all green:
- `RoleResourceCountsIntegrationTest` (8): counts per endpoint, direct-only semantics, batch aggregate
- `RoleResourceUsersIntegrationTest` (8): emails present, keyless role (the headline regression), no inheritance, user-role self-grant, filter, pagination, 404/403
- role-filter tests folded into `UserAPITest` and `UserResourceIntegrationTest` next to existing coverage
- pre-existing `testGetUsersByNameFilteredByRole` re-validated against the new SQL

Fixes #37070, Fixes #37071

This PR fixes: #37070


hassandotcms:
> Semgrep found **2** [`CUSTOM_INJECTION-2`](https://semgrep.dev/playground/r/ZRTldRY/gitlab.find_sec_bugs.CUSTOM_INJECTION-2?utm_campaign=finding_notification&utm_medium=review_comment&utm_source=github&utm_content=rule) findings:
> 
> * dotCMS/src/main/java/com/dotmarketing/business/RoleFactoryImpl.java
>   
>   * [L596-597](https://github.com/dotCMS/core/blob/954959530ce735ae8064e00f6cd01413a8e84353/dotCMS/src/main/java/com/dotmarketing/business/RoleFactoryImpl.java#L596-597) - [Triage](https://semgrep.dev/orgs/dotcms1/findings/930928077)
>   * [L596-597](https://github.com/dotCMS/core/blob/954959530ce735ae8064e00f6cd01413a8e84353/dotCMS/src/main/java/com/dotmarketing/business/RoleFactoryImpl.java#L596-597) - [Triage](https://semgrep.dev/orgs/dotcms1/findings/930928078)
> 
> The method identified is susceptible to injection. The input should be validated and properly escaped.
> 
> If this is a critical or high severity finding, please also link this issue in the #security channel in Slack.

False positive: the concatenated parts are only `?` placeholders from `map(id -> "?")` and the constant getDBFalse(); all values are bound via addParam ->preparedStatement.setObject (DotConnect.java:1229).

dotcms-integration/src/test/java/com/dotmarketing/business/UserAPITest.java
```diff
@@ -1606,4 +1607,59 @@ public void testGetUsersByNameFilteredByRole() throws DotDataException {
...
+				.roles(keylessRole).nextPersisted();
+		final User plainUser = new UserDataGen().firstName("keylessFilter" + unique + "plain").nextPersisted();
+		try {
+			final List<User> matches = userAPI.getUsersByName("keylessFilter" + unique,
+					List.of(keylessRole), 0, 40);
+			assertEquals(1, matches.size());
+			assertEquals(granted.getUserId(), matches.get(0).getUserId());
+			assertEquals(1, userAPI.getCountUsersByName("keylessFilter" + unique, List.of(keylessRole)));
+		} finally {
+			UserDataGen.remove(granted);
```
dario-daza, line 1634: I think you should also apply this cleanup to the roles/users in tests you created above:

- RoleResourceCountsIntegrationTest
- RoleResourceUsersIntegrationTest
- UserResourceIntegrationTest

dario-daza (dismissed):

dotcms-integration/src/test/java/com/dotmarketing/business/UserAPITest.java
hassandotcms: done. i did that in cleanup methods

dario-daza (approved):

Issue #36845, opened 2026-07-31
Workflow Center: add a Refresh quick action that reindexes the selection


rjvelazco:
### Description

Add a **Refresh** action to the Action Center's Quick Actions that sends every selected contentlet to be reindexed.

Unlike the other quick actions, Refresh does not depend on content state. Publish only applies to items that are not live, Delete only to archived ones — Refresh applies to the whole selection regardless of live / archived / locked status, so it is always listed whenever the selection contains at least one contentlet.

**What it does.** Clears the contentlet cache and reindexes the item — the same effect as the existing single-item endpoint `PUT /api/v1/content/_refresh/{identifierOrInode}`, which performs `CacheLocator.getContentletCache().remove(identifier)` followed by `contentletAPI.refresh(contentlet)`. Useful when content is correct in the database but stale or missing in search results.

**There is no bulk endpoint.** Two things it must not be confused with:

- `POST /api/v1/esindex/reindex` rebuilds the **entire index** (optionally filtered by content type). It is not a per-selection operation and must not be used here.
- Legacy does have a batch path, but not over REST: `cmd=full_reindex_list` → `EditContentletAction._batchReindex`, which loops the inodes on a background thread.

So this needs either N calls to `_refresh` from the client, or a new bulk endpoint. That decision should be made before building.

#### Permission question to settle

The legacy equivalent of this button is **not** shown to everyone. `view_contentlets.jsp` gates it:

```java
final boolean canReindexContentlets =
    APILocator.getRoleAPI().doesUserHaveRole(user, loadRoleByKey(Role.CMS_POWER_USER))
    || APILocator.getRoleAPI().doesUserHaveRole(user, loadCMSAdminRole());
```

"Always available" in this issue means *not gated by content state*. Whether it is additionally gated by role is a separate decision, and reindexing is an expensive operation, so matching legacy (Power User or Admin) is the conservative default. This needs an explicit answer rather than being inherited by accident.

### Acceptance Criteria

Updated to match what shipped in #37131. Original wording kept wherever it still describes the
behaviour; reworded only where the implementation went somewhere the AC did not anticipate.

[...]

PR #37081, opened 2026-08-17
feat(content-drive): trim quick actions to the old-search bulk operations and wire Push Publish

### LINKED ISSUE #37080: Workflow Center: trim Quick Actions to the old-search bulk operations and wire Push Publish
### Description

The Workflow Center dialog (multi-item selection in Content Drive) currently offers a **Quick Actions** section containing Publish, Unpublish, Archive, Unarchive, Delete, Lock and Unlock. Five of those rows duplicate — and can silently contradict — the **Workflow Actions** section rendered directly below them.

A quick action fires `POST /v1/workflow/actions/default/fire/{systemAction}`, which resolves through the content type scheme's **Default Actions** mapping. If a scheme maps `PUBLISH` to something other than a plain publish, the quick-action row labelled "Publish" and the workflow row labelled "Publish" do different things. Two rows with the same label and different behaviour is worse than one row that resolves correctly.

This task narrows Quick Actions to the bulk operations the legacy content search offered *outside* its workflow dropdown, and wires the one that is missing: **Push Publish**.

#### Target state

| Row | State after this task |
|---|---|
| Lock | wired |
| Unlock | wired — keeps the "locked by another user" warning |
| Add to Bundle | wired — keeps the bundle-picker step |
| **Push Publish** | **wired** — gated on push publish environments |
| Refresh | disabled placeholder with a `Work in progress` badge (behaviour tracked by #36845) |
| ~~Publish / Unpublish / Archive / Unarchive / Delete~~ | **removed** — reachable through the Workflow Actions section, as the scheme's own actions |

**Why Lock and Unlock stay:** they are per-user state on the version info, not a workflow transition, so they have no scheme action to defer to. `WorkflowAPI.SystemAction` documents this — they carry no actionlet and mapping them has no effect.

#### Push Publish design notes

- **Gate.** `PushPublishService.getEnvironments()` runs when the dialog opens. The row stays disabled with an explanatory tooltip until at least one environment answers. This is deliberately *not* the same state as Refresh's coming-soon badge: nothing is missing from dotCMS, something is missing from the *configuration*, and the fix belongs to an administrator.
- **Fails closed.** An unresolved or failed lookup reads as "none" — offering a push with nowhere to go fails at the servlet with a message the user cannot act on.
- **One deliberate difference from the JSP.** `view_bulk_actions_inc.jsp` gates on `sendingEndpoints` (publishing endpoints existing). This gates on **environments reachable by the user's role**, which is the stricter question and the one the form itself needs: endpoints can exist with no environment this role can send to, and the picker would then open empty.
- **Form.** Reuses `DotWorkflowPushPublishComponent` — the same section the workflow-action path renders — so a push publish is collected identically in both dialogs.
- **Fire.** New `PushPublishService.pushPublishAssets` posting to `RemotePublishAjaxAction/cmd/publish`, the same servlet the legacy search's Push Publish button hits. Kept separate from the existing `pushPublishContent` because of **date handling, not count**: `DotPushPublishData` carries whole dates that the service splits into `remotePublishDate` / `remotePublishTime`, while the form emits them already split — recombining would lose the timezone the user picked.
- **Payload.** Sends deduped **identifiers**, like Add to Bundle: a push sends the asset, so language versions of one contentlet are a single entry. The servlet splits `assetIdentifier` on `,`, so bulk needs no new endpoint.
- **No per-contentlet precondition.** The legacy `pushPublishSelectedContentlets()` is just `pushHandler.showDialog(getSelectedInodes())` — the whole selection, unfiltered by live/archived/locked. The row therefore counts every selected contentlet.
- **The enterprise-license gate is deliberately not reproduced** — that requirement no longer applies.

### Acceptance Criteria

**Quick Actions list**

- [ ] Quick Actions contains exactly five rows: Lock, Unlock, Add to Bundle, Push Publish, Refresh
- [ ] Publish, Unpublish, Archive, Unarchive and Delete are absent from Quick Actions and remain reachable from the Workflow Actions section
- [ ] Lock, Unlock and Add to Bundle keep their existing behaviour, including the "locked by another user" warning and the bundle-picker step
- [ ] Refresh renders as a disabled row with a `Work in progress` badge and a tooltip; it counts the whole selection rather than showing `0`, and is disabled by its own flag rather than by its count

**Push Publish — gate**

- [ ] The Push Publish row is enabled when `PushPublishService.getEnvironments()` returns at least one environment reachable by the user's role
- [ ] With no reachable environment, the row is disabled and its tooltip names the missing *configuration* (distinct from Refresh's coming-soon wording)
- [ ] A failed or still-pending environment lookup is treated as "none" — the row stays disabled (fails closed)

**Push Publish — form and execution**

- [ ] Selecting the row opens the configure step rendering `DotWorkflowPushPublishComponent`, the same form the workflow-action path uses
- [ ] Continue is gated on form validity; the action cannot fire without collected settings
- [ ] Firing posts to `RemotePublishAjaxAction/cmd/publish` with **deduped identifiers** — one entry per contentlet identifier, not per language version — comma-joined in `assetIdentifier`
- [ ] Publish date and time reach the servlet as the `remotePublishDate` / `remotePublishTime` pair the form emits, with the user's chosen timezone preserved
- [ ] Rows unchecked in the preview step are excluded from the fired payload
- [ ] A `200` response without a numeric `errors` value is routed to the error handler rather than reported as success (same guard Add to Bundle uses)

**Cleanup**

- [ ] Machinery whose only users were the removed rows is deleted: the `danger` flag and its two red-styling branches, the confirmation path (`confirmMessage`, the `ConfirmationService` provider/injection, `ConfirmDialogModule`, `<p-confirmDialog>`), and `fireQuickAction`
- [ ] Doc comments no longer describe Publish or Delete as quick actions
- [ ] `content.drive.worflow.action.delete.confirm` and `content-drive.action-center.confirm.header` are **retained** in `Language.properties` even though they are orphaned repo-wide — removing i18n keys is a different risk profile, since customer overrides and plugins can reference them
- [ ] `DotWorkflowPushPublishValue` lives in `@dotcms/dotcms-models` so data-access can consume it without depending on `@dotcms/ui`, and is re-exported from `@dotcms/ui` for existing callers

**i18n**

- [ ] `content-drive.action-center.coming-soon` and `content-drive.action-center.no-environments` are added
- [ ] Push Publish and Refresh labels reuse the existing `Remote-Publish` and `Refresh` keys so every translation carries over; the badge reuses `content-drive.work-in-progress`

**Automated tests**

- [ ] Unit coverage for the environments gate (available / none / lookup fails / still pending), the configure step, Continue gated on validity, the deduped-identifier payload, preview exclusions, refusal to fire without settings, the servlet body shape and its comma-joining, and the absence of the five removed actions
- [ ] `portlets-content-drive`, `data-access` and `ui` suites pass; lint and `format:check` are clean

**Manual QA against a running instance**

- [ ] An instance with no push publish environment for your role → Push Publish disabled, tooltip names the configuration
- [ ] After adding an environment → row enabled, form opens, and the push lands in the target environment
- [ ] A multi-language selection → one entry per identifier, not per language version
- [ ] Lock, Unlock and Add to Bundle behave unchanged throughout

### Priority

Medium

### Additional Context

**Implementation:** PR #37073.

**Direction chosen.** Two approaches to the duplicate rows were spiked. #37075 is the alternative: keep the five state actions in Quick Actions and gate each on the scheme's resolved workflow mapping. This ticket takes the removal path.

| | This ticket (#37073) | Alternative (#37075) |
|---|---|---|
| Publish / Unpublish / Archive / Unarchive / Delete | removed from Quick Actions | kept, gated on the workflow mapping |
| Closes the raw-API bypass | yes, by removal | yes, by gating |
| New requests per dialog open | 1 (environments) | 1 per content type + 1 per scheme |
| Backend follow-up | none | optional: a resolved-mapping endpoint |

**Why Refresh cannot be wired here.** It is backed by `_bulkrefresh`, which streams progress over SSE and is not job-backed, so it cannot reuse the synchronous `bulkFire` path the other quick actions run on. Tracked by #36845.

**Naming.** The dialog is presented to users as **Workflow Center**; the code and i18n keys keep the `action-center` prefix.

**Reference:** `dotCMS/src/main/webapp/html/portlet/ext/contentlet/view_bulk_actions_inc.jsp` — the legacy implementation hitting the same endpoints.

rjvelazco:
Closes #37080

> Supersedes #37073, the spike this branch was promoted from. #37075 explored the alternative direction (keeping the state actions, gated on the workflow mapping); @nollymar decided on the parity approach implemented here.

## What this does

Narrows **Quick Actions** to the bulk operations the old content search offered *outside* its workflow dropdown, and wires Push Publish.

| Row | State |
|---|---|
| Lock | wired |
| Unlock | wired — keeps the "locked by another user" warning |
| Add to Bundle | wired — keeps the bundle-picker step |
| **Push Publish** | **wired** — gated on push publish environments |
| Refresh | disabled placeholder, `Work in progress` badge |
| ~~Publish / Unpublish / Archive / Unarchive / Delete~~ | **removed** |

## Why remove the five state actions

They are already offered by the **Workflow Actions** section below, as the *scheme's own* actions.

That is not a duplicate — it is a more correct answer. A quick action fires `POST /v1/workflow/actions/default/fire/{systemAction}`, which resolves through the scheme's Default Actions mapping. If a content type's scheme maps `PUBLISH` to something other than a plain publish, the quick-action row labelled "Publish" and the workflow row labelled "Publish" do different things. Two rows with the same label and different behaviour is worse than one row that resolves correctly.

Lock and Unlock stay because they are genuinely per-user state on the version info, not a workflow transition — they have no scheme action to defer to. `WorkflowAPI.SystemAction` documents this: they carry no actionlet, and mapping them has no effect.

### Gate

`PushPublishService.getEnvironments()` runs when the dialog opens. The row stays disabled with an explanatory tooltip until it answers with at least one environment.

Deliberately **not** the same state as Refresh's coming-soon badge: nothing is missing from dotCMS here, something is missing from the *configuration*, and the fix belongs to an administrator. The row says which.

Fails closed. An unresolved or failed lookup reads as "none" — offering a push with nowhere to go fails at the servlet with a message the user cannot act on.

> **One deliberate difference from the JSP.** `view_bulk_actions_inc.jsp` gates on `sendingEndpoints` (publishing endpoints existing). This gates on **environments reachable by the user's role**, which is the stricter question and the one the form itself needs: endpoints can exist with no environment this role can send to, and then the picker would open empty.

### Form

Reuses `DotWorkflowPushPublishComponent` — the same section the workflow-action path renders — so a push publish is collected identically in both dialogs. `$configureKinds` returns `['pushPublish']` for the row, putting it through the existing `pick → configure → preview → execute` flow.

### Fire

New `PushPublishService.pushPublishAssets`, posting to `RemotePublishAjaxAction/cmd/publish` — the same servlet the old search's Push Publish button hits.

Kept separate from the existing `pushPublishContent` because of **date handling, not count**. `DotPushPublishData` carries whole dates that the service splits into the servlet's `remotePublishDate` / `remotePublishTime` pair; the form emits them *already* split. Routing through the existing method would mean recombining two strings into a `Date` for the service to take apart again — losing the timezone the user picked on the way.

`DotWorkflowPushPublishValue` moved down to `@dotcms/dotcms-models` so data-access can accept it without depending on `@dotcms/ui`, and is re-exported from ui where callers have always imported it.

Sends **identifiers**, deduped, like Add to Bundle: a push sends the asset, so language versions of one contentlet are a single entry. The servlet splits `assetIdentifier` on `,` (`RemotePublishAjaxAction`: *"Support for multiple ids in the assetIdentifier parameter"*), so bulk needs no new endpoint.

The servlet answers 200 for its own failures, so a response without a numeric `errors` goes to the error handler rather than being reported as a success — the same guard Add to Bundle uses.

### What the JSP told us

- **No per-contentlet precondition.** `pushPublishSelectedContentlets()` is just `pushHandler.showDialog(getSelectedInodes())` — the whole selection, unfiltered by live/archived/locked. The row therefore counts every contentlet.
- **The enterprise-license gate is deliberately not reproduced.** That requirement no longer applies.

## Why the placeholders are rendered rather than hidden

Refresh is an action the old search does offer. Leaving it out entirely makes Content Drive look like it dropped it; a disabled row with a `Work in progress` badge and a tooltip is the honest state. It is backed by `_bulkrefresh`, which streams progress over SSE and is not job-backed, so it cannot reuse the synchronous `bulkFire` path the other quick actions run on. Its behaviour is tracked by #36845.

It counts over the whole selection rather than sitting at `0` — a `0` would read as "does not apply to these items", a different and untrue claim from "not built yet". The row is disabled by its own flag, not by its count.

## Commits

1. **Trim the quick actions** — list, new IDs, `comingSoon` flag, template, specs.
2. **Drop the machinery the removed rows were the only users of** — `danger` (permanently `false` once the destructive rows left, plus its two red-styling branches), the entire confirmation path (`confirmMessage`, the `ConfirmationService` provider/injection, `ConfirmDialogModule`, `<p-confirmDialog>`, and `fireQuickAction`, which existed only so the confirm and direct branches could share a path), their tests, and doc comments still describing Publish and Delete as quick actions.
3. **Prettier formatting.**
4. **Wire Push Publish** — gate, form, fire, and the last two stale comments.

## i18n

Two keys added: `content-drive.action-center.coming-soon` and `content-drive.action-center.no-environments`. Labels reuse the legacy `Remote-Publish` and `Refresh` keys the old search already uses, so the parity is free in every translation. The badge reuses the existing `content-drive.work-in-progress`.

## Known leftovers — deliberate, flagging for review

Two keys are orphaned repo-wide after commit 2 (`core-web` and `dotCMS/src` both grep clean):

- `content.drive.worflow.action.delete.confirm`
- `content-drive.action-center.confirm.header`

Left in `Language.properties` on purpose: removing i18n keys is a different risk profile, since customer overrides and plugins can reference them. This was confirmed as the intended outcome when #37080 was written, so the AC there requires retaining them.

## Testing

- `pnpm nx test portlets-content-drive` — **1253 passed**, 32 suites
- `pnpm nx test data-access` — **758 passed**
- `pnpm nx test ui` — **820 passed**
- `pnpm nx run-many -t lint -p portlets-content-drive data-access ui dotcms-models` — clean

New coverage: the environments gate (available / none / lookup fails / still pending), the configure step opening on the row, Continue gated on form validity, the fired payload being deduped identifiers plus the collected settings, rows unchecked in the preview being honoured, the refusal to fire without settings, the servlet body shape and its comma-joining, and the five state actions being absent from the list.

## Video

https://github.com/user-attachments/assets/4ccbe7c2-02ec-4cf6-a51c-52156cc18c96

## Alternative considered — the v1 REST path

There is a REST route to the same outcome, and it lost on two counts rather than on capability:

`POST /api/v1/bundles/assets` to resolve or create a bundle from `assetIds`, then `POST /api/v1/publishing/push/{bundleId}` with `{operation, publishDate, expireDate, environments, filterKey}`.

- **It leaves a named bundle behind.** `/v1/bundles/assets` needs a `bundleName` to auto-create, so every quick-action push would deposit a bundle in the Bundles list. The servlet creates an anonymous one.
- **It is two non-atomic round trips.** A push that fails after the bundle is created leaves an orphan.

Given the goal is parity with the old search — whose Push Publish button hits the servlet — staying on the servlet keeps behaviour identical and adds nothing to clean up.

## Manual QA

Acceptance criteria on #37080:

- an instance with no push publish environment for your role → Push Publish disabled, tooltip names the configuration
- after adding an environment → row enabled, form opens, push lands in the target environment
- a multi-language selection → one bundle entry per identifier, not per language version
- **legacy push publish dialog, two or more environments and a non-default filter** → arrives in every target. The encoding fix below changes `getPublishEnvironmentData`'s output, and that method's only caller is the dialog every legacy push in the product funnels through (content types, templates, containers, pages, rules). `whoToSend` moved from a literal `env1,env2` to `env1%2Cenv2`; the container percent-decodes before `getParameter` and `RemotePublishAjaxAction` splits on `,` afterwards, so both forms should arrive identically — but this path has never been exercised, and it is the kind of thing that fails silently.
- **`publishexpire`** → the toast under-reports. `RemotePublishAjaxAction` creates one bundle for the publish half and a second for the expire half, and the second `responseMap` overwrites the first, so the counts describe the expire half alone. Not a regression: `PushPublishActionlet.doPushPublish` splits the same way. Documented on `executePushPublish`.

## Review follow-ups

- `getPublishEnvironmentData` now escapes `timezoneId`, `iWantTo`, `whoToSend` and `filterKey` — a raw `&` or `=` in a form-encoded value silently swallows every parameter after it. Fixed rather than documented, with two tests covering the `&`/`=` case that the earlier assertions did not.
- `executeAddToBundle` and `executePushPublish` collapsed onto one `fireLegacyServletBulk` runner, so the "servlet answers 200 for its own failures" guard has a single definition.
- The equivalent duplication one layer down — `pushPublishAssets` and `getPublishEnvironmentData` building the same eleven parameters — is tracked as **#37090**, deferred because it changes a shared service with callers outside this PR's feature area.

This PR fixes: #37080


rjvelazco:
Both review notes addressed in `73eb4d4`.

## 1. `executePushPublish` store coverage

Added a `describe` block mirroring `executeAddToBundle`'s — comma-joining, server-count reporting, the `Math.max(total - errors, 0)` clamp, the in-flight marker, the replay guard, the empty-identifiers no-op, and `catchError`.

Plus the two cases **neither** method had: a string `errors` and a missing body. That was the right call-out — it is the branch that justifies the separate code path, since the servlet answers 200 for its own failures and reporting that as a result yields `NaN` successes on a push that never happened.

Verified non-vacuous by mutation rather than by the tests going green: stubbing the guard to `if (false)` fails **exactly** those two tests and nothing else.

> Worth noting the same guard is untested on `executeAddToBundle`. Left alone here — it is pre-existing and outside this PR — but it is a real gap if anyone wants a follow-up.

## 2. Parameter encoding

Every value now goes through `encodeURIComponent`, not just `assetIdentifier`.

The assessment was right that no breakage was expected — environment ids are UUIDs, `iWantTo` is a fixed enum, dates arrive pre-formatted, `timezoneId` is a Java zone id. But `filterKey` is a descriptor key with no such guarantee, and one unescaped `&` or `=` silently swallows every parameter after it: the push fires with the wrong filter and nothing errors. Cheap to remove the question entirely. `getPublishEnvironmentData` still escapes only the identifier; this deliberately does not copy that.

Also fixed the fixture: it used `timezoneId: 'Costa Rica'`, which is not a real Java zone id, so the assertion described a body that could not occur. Now `America/Costa_Rica`, with a new test covering a `filterKey` carrying `&`, `=` and a space.

## Checks

- `portlets-content-drive` — **1263 passed**, 32 suites (+10)
- `data-access` — **759 passed** (+1)
- lint and `format:check` — clean

core-web/libs/dotcms-models/src/lib/dot-push-publish-data.model.ts
```diff
@@ -11,3 +11,29 @@ export interface DotPushPublishData {
...
 }
+
+/** What the user can ask a push publish for. Values are the backend's `iWantTo` vocabulary. */
+export type DotWorkflowPushPublishAction = 'publish' | 'expire' | 'publishexpire';
+
+/**
+ * The push publish payload, in the shape the backend's `PushPublishBean` expects.
+ *
+ * Already converted — dates split into `yyyy-MM-dd` + `HH-mm` pairs, environments comma-joined into
+ * `whereToSend` — so a consumer can hand it straight to a fire request without repeating the
```
zJaaal: This sentence is what sends a reader looking for a fire request that does not exist. The first consumer added in this PR does **not** hand the value to a fire request, it hand-encodes a form body for `RemotePublishAjaxAction`.

Please say why the fire request is not reachable: `bulkFire` needs a `workflowActionId` wiring `PushPublishActionlet`, and there is no `PUSH_PUBLISH` in `WorkflowAPI.SystemAction`. As written, the comment advertises a capability the codebase does not have, and the very next commit contradicts it.

core-web/libs/data-access/src/lib/push-publish/push-publish.service.ts
```diff
@@ -82,6 +83,68 @@ export class PushPublishService {
...
 
+    /**
+     * Push publishes one or more assets, from a payload already in the servlet's own shape.
+     *
+     * Backing endpoint: the same `RemotePublishAjaxAction/cmd/publish` servlet
+     * {@link pushPublishContent} uses. `assetIdentifier` accepts several **identifiers**
+     * comma-joined — the action splits on "," (`RemotePublishAjaxAction`, "Support for multiple ids
+     * in the assetIdentifier parameter"), so bulk needs no new endpoint.
+     *
+     * Separate from {@link pushPublishContent} because of the date handling, not the count.
```
zJaaal: Date handling is the smaller half of the reason, and leading with it reads as "we could have reused `pushPublishContent` but did not want to convert dates," which invites a reviewer to ask for exactly that refactor.

The load-bearing reason is that there is no fire endpoint to route this through at all: no `PUSH_PUBLISH` in `WorkflowAPI.SystemAction`, and `bulkFire` needs a concrete scheme action that wires `PushPublishActionlet`, which a mixed-content-type selection cannot supply. Please state that first, then the date detail as the reason the two methods do not share a body.

Separately: `pushPublishAssets` and `getPublishEnvironmentData` now build the same eleven-parameter body twice. A private `post(assetIdentifier, params)` with two adapters, one per input shape, would keep the date-handling difference where it belongs and leave one definition of the servlet's contract.

core-web/libs/data-access/src/lib/push-publish/push-publish.service.ts
```diff
@@ -82,6 +83,68 @@ export class PushPublishService {
...
+
+        const headers = new HttpHeaders({
+            'Content-Type': 'application/x-www-form-urlencoded'
+        });
+
+        // Every value is encoded, not just the identifier. In practice these are all URL-safe —
+        // environment ids are UUIDs, `iWantTo` is a fixed enum, dates arrive pre-formatted and
+        // `timezoneId` is a Java zone id — but `filterKey` is a descriptor key with no such
+        // guarantee, and one unescaped `&` or `=` in a form-encoded body silently corrupts every
+        // parameter after it. `getPublishEnvironmentData` encodes only the identifier; this does not
```
zJaaal: This comment correctly identifies a live bug in `getPublishEnvironmentData` (`filterKey` interpolated unencoded, so one `&` or `=` in a descriptor key silently corrupts every parameter after it) and then leaves it in the method directly below.

Since the PR is already in this file, please fix it rather than document it. It is `encodeURIComponent` on the remaining interpolations, and it removes the need for most of this paragraph.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/features/action-execution/withActionExecution.ts
```diff
@@ -294,6 +300,68 @@ export function withActionExecution() {
...
+                     * Push publishes the given identifiers to the chosen environments.
+                     *
+                     * Identifiers, not inodes: push publish sends the *asset*, so every language
+                     * version of a contentlet is one entry — the same collapse Add to Bundle makes.
+                     *
+                     * Shares the legacy servlet's response shape with Add to Bundle, and the same
+                     * caveat: it answers 200 for its own failures, so a body without a numeric
+                     * `errors` is routed to the error handler rather than reported as a success.
+                     */
+                    executePushPublish: (
```
zJaaal, line 351: Heads-up on `publishexpire`, which the QA list on #37080 does not cover.

In `RemotePublishAjaxAction.publish` the publish branch and the expire branch each create their **own** bundle, and only the second `responseMap` is written to the response. So for `iWantTo: 'publishexpire'` the `successCount` computed here reflects only the expire half, and the toast under-reports.

Shared with the workflow path (`PushPublishActionlet.doPushPublish` does the same two-bundle split), so not a regression from this PR. Worth adding to the manual QA list, and worth a line here so the next reader does not assume the count covers both operations.

zJaaal (commented):
_Review authored by Claude, posted from @zJaaal's account._

The code is correct as far as I can trace it, but the PR explains the wrong reason for its central design decision, and that is what makes the diff hard to read. The first question it provokes is "why an AJAX trigger when we already have a push publish component?"

## The justification is misfiled

The PR body and `pushPublishAssets`' doc block both justify the new method on **date handling**. That is real but minor. The reason that actually forces the design is absent: the workflow push publish path is unreachable from here.

- `bulkFire` (`PUT /v1/workflow/contentlet/actions/bulk/fire`) requires a `workflowActionId` that is a real action UUID belonging to the contentlets' scheme, and it only push publishes if that action wires `PushPublishActionlet`. `FireBulkActionsForm.java:29` documents the skip behaviour when the scheme does not match.
- There is no `PUSH_PUBLISH` in `WorkflowAPI.SystemAction` (`WorkflowAPI.java:1229`), so there is no `default/fire/{systemAction}` route either, the way Lock and Unlock get one.
- The quick action fires over an arbitrary, possibly multi-content-type selection with no action chosen, so there is nothing to put in `workflowActionId`.

Worth adding that routing through the workflow path would buy nothing anyway: `PushPublishActionlet.doPushPublish` (`PushPublishActionlet.java:129`) and `RemotePublishAjaxAction.publish` (`RemotePublishAjaxAction.java:195`) are duplicate implementations of the same sequence (`new Bundle(null, publishDate, ...)`, `saveBundle(bundle, envsToSendTo)`, `addContentsToPublish` / `addContentsToUnpublish`), and the actionlet imports the servlet's own `DIALOG_ACTION_*` constants. Please put this in the PR body and in the method's doc comment. It turns "why not reuse what we have" into "there is nothing to reuse."

## One alternative is not weighed

There is a REST path for this now, and the PR does not mention it: `POST /api/v1/bundles/assets` (`BundleManagementResource.java:171`) to resolve or create a bundle from `assetIds`, then `POST /api/v1/publishing/push/{bundleId}` (`PublishingResource.java:733`) with `{operation, publishDate, expireDate, environments, filterKey}`. Steps 7 to 10 of `pushBundle` are functionally what the servlet does.

I think staying on the servlet is defensible, for two reasons the PR should state rather than leave implicit: `/v1/bundles/assets` needs a `bundleName` to auto-create, so the quick action would leave a **named bundle in the Bundles list** where the servlet creates an anonymous one; and it is two non-atomic round trips, so a failed push leaves an orphan bundle. Given the PR's goal is parity with the old search, the reader needs to see that this was considered and why it lost.

## Verified as correct

So the following do not need changes, noting them so nobody re-checks:

- `errors` and `total` come back as ints from `PublisherAPIImpl` (lines 328, 330), so the `typeof result?.errors !== 'number'` guard cannot misfire on a successful push.
- Omitting `filterKey` falls through to the default filter in `PublisherAPIImpl.getFilterDescriptorByKey` (line 306), no NPE.
- The component emits `yyyy-MM-dd` and `HH-mm`, which is what the servlet's `STANDARD_DATE_FORMAT` parse expects.

## Process note

The body still carries a "Still to do before this leaves draft" section listing manual QA as acceptance criteria on #37080, but the PR is open for review and not a draft. Either that QA is done and the section should be updated, or this should go back to draft.

Inline comments follow.

rjvelazco:
@claude Review code looking for anti-patterns and dirty code

rjvelazco:
Fixed in `e8e55df` — all four inline points were right.

**Justification was misfiled.** Both docs now lead with the real reason: there is no fire path to route through. `bulkFire` needs a real action UUID from the contentlets' own scheme wiring `PushPublishActionlet`, there is no `PUSH_PUBLISH` in `WorkflowAPI.SystemAction`, and a mixed-content-type selection with no action chosen cannot supply one. Date handling is now stated second, as the reason the two methods do not share a body. Also added that routing through the workflow path would buy nothing — the actionlet and the servlet are duplicate implementations.

**`dot-push-publish-data.model.ts`** — "hand it straight to a fire request" was untrue of the consumer this PR adds. Now names both destinations.

**`getPublishEnvironmentData`** — fair hit, fixed rather than documented. `timezoneId`, `iWantTo`, `whoToSend`, `filterKey` now encoded. Five existing assertions changed from `timezoneId=Costa Rica` to `Costa%20Rica` — they were asserting the bug.

**`publishexpire`** — verified at `RemotePublishAjaxAction.java:258-275`: the expire branch's `responseMap` overwrites the publish branch's, so the counts describe the expire half alone. Documented on `executePushPublish`, noting it is shared with `doPushPublish` and so not a regression. Added to the QA list on #37080.

**Not done — the shared body builder.** `pushPublishAssets` and `getPublishEnvironmentData` do now build the same eleven parameters twice. Skipping it here: extracting a common builder means reworking the method that five existing tests pin by exact body string, for no behaviour change, in a PR whose scope is the quick actions. Happy to take it as a follow-up if you would rather it land now.

**Alternative not weighed** — agreed, `/v1/bundles/assets` + `/v1/publishing/push/{bundleId}` should have been in the body with the two reasons it lost (a named bundle left in the Bundles list, and two non-atomic round trips leaving an orphan on failure). Going into the PR description with the corrected justification.

**Process note** — correct, the PR is not a draft. Updating the body now.

Checks: content-drive **1263**, data-access **780**, lint and format clean.

core-web/libs/data-access/src/lib/push-publish/push-publish.service.ts
rjvelazco: Agreed on the duplication — after the encoding fix these two build the same eleven parameters identically.

**Tracked as #37090**, to be done in its own PR.

**Correcting myself:** I first said the six exact-body-string assertions made this risky to refactor. That's backwards. For a pure string builder, pinning the exact wire format is the *strongest* regression detector available — any drift fails loudly and points at the character. Not a reason to defer.

## Effort — small, ~30–45 min

The param lists are identical in set and order: `assetIdentifier · remotePublishDate · remotePublishTime · remotePublishExpireDate · remotePublishExpireTime · timezoneId · iWantTo · whoToSend · bundleName= · bundleSelect= · filterKey?`

Only the input shapes differ — dates (`Date` → formatted vs already split), `whoToSend` (`string[]` vs pre-joined), and the action field name (`pushActionSelected` vs `iWantTo`). So exactly what you proposed: one `post(assetIdentifier, params)` over a flat record, two thin adapters. Net ≈ −20 lines.

## Risk — modest, and not where I said

1. **Blast radius.** `pushPublishContent` has two production callers, both outside this PR's feature area: `dot-push-publish-dialog.component.ts` (the single-item dialog, reached from many screens) and `dot-content-types.component.ts` (push publish a content type from the listing). Neither is exercised by content-drive tests.
2. **The date-defaulting branch** — `publishDate ? new Date(publishDate) : new Date()`, four times. That's the subtle part, not the concatenation. Mitigated by keeping the derivation inside its adapter and refactoring only the assembly.

## Why separate

Not because it is hard, but because it modifies a shared service with two user-facing callers this PR has nothing to do with. Bundling it means a reviewer of a quick-actions PR is implicitly signing off on the content-types listing and the push publish dialog, and those want their own manual QA — which #37090 carries as acceptance criteria.

rjvelazco:
Thanks — took the first one, deferring the second.

## 🟡 `executeAddToBundle` / `executePushPublish` duplication — fixed in `086a0fe`

Right, and it was mine: I wrote `executePushPublish` against the bundle method as a template. Extracted `fireLegacyServletBulk(actionName, identifiers, request, message)`; both methods are now a single call, request passed lazily so nothing posts when the guards refuse. −102/+97 lines, and the "answers 200 for its own failures" guard now has one definition instead of two.

Behaviour unchanged — the 105 existing store tests pass untouched, and re-running the mutation check *through the helper* still fails exactly the two tests that assert the guard and nothing else.

**One correction:** `onUnknownOutcome` isn't an unused helper the two methods ignore — it's used at `withActionExecution.ts:143` by the `bulkFire` path, which answers with a `summary` object rather than `{errors, total}`. Different shape, different message, so it was never a drop-in here. Noted that in the new helper's doc so the next reader doesn't try to merge them.

## 🟡 `push-publish.service.ts` duplication — deferring

Same category, different risk. Already proposed a follow-up ticket in the inline thread ([r3799314063](https://github.com/dotCMS/core/pull/37081#discussion_r3799314063)); the distinction is that the store tests assert *behaviour*, so they caught a bad refactor, while five service tests pin `getPublishEnvironmentData` by exact body string — refactoring the body construction there is a different proposition from collapsing two identical control flows.

Agreed on the direction (`private post(assetIdentifier, params)` with two shape-adapters), just not in this PR.

Checks: content-drive **1263**, data-access **780**, lint and format clean.

core-web/libs/data-access/src/lib/push-publish/push-publish.service.ts
```diff
@@ -93,6 +166,8 @@ export class PushPublishService {
             timezoneId
         }: DotPushPublishData
     ): string {
+        // Encoded for the same reason as `pushPublishAssets`: an unescaped `&` or `=` in any value
```
zJaaal, line 169: Right call fixing it here, and thank you for not just documenting it.

Flagging the reach, since it argues against the #37090 deferral more than for it: the only direct caller of this method is `dot-push-publish-dialog.component.ts`, and that dialog is the single funnel for every legacy push publish in the product (content types, templates, containers, pages, rules). So this commit already changes the wire format on all of those, including `whoToSend` moving from `env1,env2` to `env1%2Cenv2` for a multi-environment push.

I believe that is correct (the container percent-decodes before `getParameter`, and the servlet splits on `,` after decoding), so no change requested to the code. Please add the corresponding manual QA line to the PR body: one push from the legacy dialog with two or more environments and a non-default filter.

core-web/libs/data-access/src/lib/push-publish/push-publish.service.spec.ts
```diff
@@ -110,7 +110,7 @@ describe('PushPublishService', () => {
         );
 
         expect(req.request.body).toBe(
-            `assetIdentifier=${assetIdentifierEncoded}&remotePublishDate=2020-07-08&remotePublishTime=10-10&remotePublishExpireDate=${currentDateStr}&remotePublishExpireTime=${currentTimeStr}&timezoneId=Costa Rica&iWantTo=publish&whoToSend=env1&bundleName=&bundleSelect=&filterKey=hol`
+            `assetIdentifier=${assetIdentifierEncoded}&remotePublishDate=2020-07-08&remotePublishTime=10-10&remotePublishExpireDate=${currentDateStr}&remotePublishExpireTime=${currentTimeStr}&timezoneId=Costa%20Rica&iWantTo=publish&whoToSend=env1&bundleName=&bundleSelect=&filterKey=hol`
```
zJaaal, line 113: **No test here would have failed before the fix.** All five updated assertions moved `timezoneId=Costa Rica` to `Costa%20Rica`, which pins *space* encoding. The bug that was fixed is `&` and `=` corruption in `filterKey`, and nothing in this describe covers it.

The sibling block already has exactly the right test — `should escape a value carrying a reserved character`, asserting `filterKey=a%26b%3Dc%20d`. Two `it`s copied down here would close it:

1. `filterKey: 'a&b=c d'` through `pushPublishContent`, asserting the escaped form.
2. `environment: ['env1', 'env2']`, asserting `whoToSend=env1%2Cenv2`. Every assertion in this block uses a single environment, so the one value whose wire format materially changed is the one nothing checks.

Nit while you are in the file: the two tests at lines 137-155 and 157-175 are byte-for-byte identical, same name, same `formValue`, same assertion. Pre-existing, but this PR touches both, so deleting one is free.

zJaaal (commented):
_Review authored by Claude, posted from @zJaaal's account. Second pass, on 086a0fe._

Everything from the first pass is addressed, and the runner extraction is better than what I asked for. One new consequence to own before this merges, and one test gap.

## Verified locally, not just read

Checked out 086a0fe and ran it:

- `data-access --testPathPatterns=push-publish.service.spec.ts` — **11/11 pass** (all 5 `pushPublishAssets` tests execute)
- `portlets-content-drive` store + action-center + `action-center.spec` — **388/388 pass, 6 suites**
- `lint` on `data-access`, `dotcms-models`, `ui`, `portlets-content-drive` — clean. The one warning is a pre-existing unused eslint-disable in `libs/dotcms-models/jest.config.ts`, untouched by this PR.

## Resolved

- **Justification.** Both docs now lead with the unreachable fire path (no `PUSH_PUBLISH` in `WorkflowAPI.SystemAction`, `bulkFire` needs a scheme action wiring `PushPublishActionlet`), with the date detail second. The model's doc naming both destinations is the right fix.
- **Encoding.** Fixed in `getPublishEnvironmentData` rather than documented. Making `environment.join(',')` explicit instead of leaning on implicit array stringification is a good extra.
- **`publishexpire`.** Documented on `executePushPublish` with the two-bundle mechanism and the not-a-regression note.
- **`fireLegacyServletBulk`.** Better than the `post(assetIdentifier, params)` I suggested: it collapses the *store-side* duplication, so the "answers 200 for its own failures" guard has one definition instead of two, and the lazy `request()` thunk correctly keeps the post from firing when the guards refuse. Passing `bundle` in via the closure rather than through the runner keeps the runner free of either action's specifics.

## Deferral to #37090 accepted, with one consequence

The reasoning holds: a shared service with callers outside this PR's feature area wants its own QA, and bundling it makes a quick-actions reviewer sign off on unrelated screens.

But that argument now applies to something already **in** this PR. The encoding fix changes `getPublishEnvironmentData`'s output for `timezoneId`, `iWantTo`, `whoToSend` and `filterKey`, and the wire format for a multi-environment push moved from literal `env1,env2` to `env1%2Cenv2`. Its only direct caller is `dot-push-publish-dialog.component.ts`, which is the funnel every legacy push publish in the product goes through: content types, templates, containers, pages, rules. That is a *wider* blast radius than the refactor being deferred, not a narrower one.

I am not asking to revert it, it fixes a real bug. I am asking this PR to own the QA it implies. One line on the manual list:

> Push publish a single asset from the legacy dialog with **two or more environments** selected and a **non-default filter**, and confirm it arrives in every target.

The percent-decode happens in the container before `getParameter`, and `RemotePublishAjaxAction` splits on `,` after decoding, so I expect this to pass. It has just never been exercised, and it is the kind of thing that fails silently rather than loudly.

## PR body still stale

Unchanged from the first pass: the body still carries a "Still to do before this leaves draft" section while the PR is open for review and not a draft, and it still does not mention the `POST /v1/bundles/assets` + `POST /v1/publishing/push/{bundleId}` alternative or why the servlet won. The named-bundle and non-atomic-two-call reasons are good ones and belong in the body, otherwise the next reader repeats this whole thread.

Inline comments follow.

zJaaal (approved):

rjvelazco:
Second pass addressed in `cda2b25`, plus the PR body.

## The test gap — you were right, no test there would have failed

Fixed. The five assertions I'd updated pinned `timezoneId` space encoding, which is the harmless case; the `&`/`=` corruption the fix actually closed had no coverage in that block. Two tests added:

- `filterKey: 'a&b=c d'` → `filterKey=a%26b%3Dc%20d`
- `environment: ['env1', 'env2']` → `whoToSend=env1%2Cenv2` — the parameter whose wire format actually moved, and the one every existing assertion missed by using a single environment

Verified non-vacuous: reverting **only** `getPublishEnvironmentData`'s encoding fails exactly these two and nothing else. 12/12 with the fix in place.

Duplicate test deleted too — confirmed byte-identical before removing.

## The blast-radius consequence — owned

Fair, and the inconsistency was mine: I deferred #37090 on blast-radius grounds while landing an encoding change with a *wider* reach in the same PR. Your QA line is in the PR body verbatim, and on #37080 with the reasoning: https://github.com/dotCMS/core/issues/37080#issuecomment-5330252471

## PR body — updated

- **Stale draft section gone.** Renamed to `## Manual QA`; the PR has not been a draft for a while and the heading was lying about it.
- **Alternative now weighed.** Added `/v1/bundles/assets` + `/v1/publishing/push/{bundleId}` with the two reasons it lost — the named bundle it leaves in the Bundles list, and two non-atomic round trips leaving an orphan on failure. You were right that the next reader would otherwise repeat this thread.
- **Review follow-ups section** added, so the encoding fix, the runner extraction and the #37090 deferral are visible without reading the commit log.
- **`publishexpire`** on the QA list as well as in the code comment.

Checks: data-access **12/12** on the push publish spec, lint and `format:check` clean.

PR #37082, opened 2026-08-17
feat(users): Roles tab wired to real endpoints (#36718)

### LINKED ISSUE #36718: Users portlet: Roles tab (dual-list)

AP2300:
Closes #36718

Depends on #36990 (Profile tab shell). This PR's diff currently includes the shell commits; once #36990 is merged to main, this branch will be rebased onto the new main and the diff will show only the Roles tab changes.

## Summary

- Replaces the placeholder \"Coming soon\" Roles panel with a **shuttle-style picker**: Available tree on the left, Granted list on the right, arrows in the middle.
- Available side renders the full N-level role hierarchy from `GET /api/v1/roles?loadChildrenRoles=true` + per-node fan-out on non-root roles. Parent nodes get a bulk-select checkbox that toggles every grantable leaf underneath.
- Granted list is seeded from `GET /api/v1/roles/users/{userId}` and refreshed on every move; the parent dialog picks the change up via `grantedChange` and folds it into `PUT /api/v1/users` as the `roles` field.
- Non-grantable branches (Publisher / Legal, isolated organizational nodes) render without checkboxes.
- Empty roots are pruned from the Available panel once every grantable descendant has been moved to Granted (no dead containers).

## Notable non-obvious calls

- **Personal-role filter on save** (`roleKey === userId`, `editUsers=false`). The backend's \`UserResource#processRoles\` calls \`removeAllRolesFromUser\` then loops \`addRoleToUser\`; the personal role trips the \`editUsers\` guard and rolls the save back with \"Cannot alter users on this role\". Filtering it in the outbound payload sidesteps the guard. Proper fix belongs on the backend.
- **Empty roles list is currently a no-op on the backend** — \`UtilMethods.isSet(roles)\` treats \`[]\` as \"don't touch\". Not fixed here; documented in the Profile branch PR.

## Test plan

- [x] Open the edit dialog on a user; Roles tab hydrates with the current granted set on the right
- [x] Move a leaf across via the arrow; save → reopen → the leaf sticks
- [x] Check a parent → all grantable leaves get selected; move them across; the parent disappears from Available
- [x] Check Publisher / Legal roots render without a checkbox
- [x] Save with a personal-role-only user; no \"Cannot alter users on this role\" error


core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.ts
```diff
@@ -1,26 +1,450 @@
...
+                            frontend: user.frontendUser ?? false,
+                            canLogin: user.hasConsoleAccess ?? false,
+                            showGettingStarted: false
+                        }
+                    });
+                    this.form.markAsPristine();
+
+                    const roleKeys = roles
+                        .map((role) => role.roleKey)
+                        .filter((key): key is string => !!key);
```
hmoreras: Dropping keyless roles here means they're invisible in the Granted panel *and* revoked on save. `roleKey` is optional — roles created from the Roles portlet often have none (`dot-roles-add.component.ts` defaults it to `''`). The backend's `processRoles` wipes every direct role and then re-adds only the keys we send, so a role that never made it into this list is gone after a save the admin thought was a no-op on that role.

The identity we send has to be able to name a keyless role, otherwise the Roles tab can't be safe. Worth pairing with the `grantIdentifier` thread in the tab.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-roles-tab/dot-users-roles-tab.component.ts
```diff
@@ -0,0 +1,488 @@
...
+        return !this.rolesWithChildren().has(role.id);
+    }
+
+    /**
+     * The identifier we hand up on `grantedChange` for a given role.
+     * `roleKey` when the backend has one; the role `id` otherwise
+     * so the shell can still preserve/save the assignment.
+     */
+    private grantIdentifier(role: RoleOption): string {
+        return role.roleKey || role.id;
```
hmoreras, line 157: Falling back to `role.id` doesn't reach the backend: `loadRoleByKey` runs `from Role where role_key = ?`, so a UUID resolves to null and `UserHelper.addRole` logs "does NOT exist… Ignoring it" and skips. Because `processRoles` has already wiped every role by then, the grant just doesn't happen.

So the note above is right about the mechanism but reads too gently — the user moves a keyless role across, saves, reopens, and it isn't there. That's test-plan item 2 failing for exactly the roles that most need this tab.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-roles-tab/dot-users-roles-tab.component.ts
```diff
@@ -0,0 +1,488 @@
...
+                    this.isLoading.set(false);
+                }
+            });
+    }
+}
+
+function toRoleOption(role: DotRoleView): RoleOption {
+    return {
+        id: role.id,
+        roleKey: (role.roleKey ?? '').trim(),
```
hmoreras, line 499: `editUsers` is on `DotRoleView` but dropped here, and `isGrantableLeaf` only asks "does this role have children" — so the claim up at the `grantableLeavesByRole` doc ("non-user-assignable roles contribute nothing to the map") isn't actually implemented anywhere.

Consequence: any `editUsers=false` leaf gets a checkbox and can be moved across, and `addRoleToUser` throws `Cannot alter users on this role` on save — the whole PUT is transactional, so the profile edit rolls back too. Carrying `editUsers` through into `RoleOption` and folding it into `isGrantableLeaf` makes the comment true.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-roles-tab/dot-users-roles-tab.component.html
```diff
@@ -0,0 +1,169 @@
...
+                [placeholder]="'users.dialog.roles.filter.placeholder' | dm"
+                data-testid="users-roles-granted-filter" />
+        </div>
+        <div class="max-h-96 overflow-auto p-2" data-testid="users-roles-granted-body">
+            @if (grantedList().length === 0) {
+                <p class="text-color-secondary p-2 text-sm" data-testid="users-roles-granted-empty">
+                    {{ 'users.dialog.roles.granted.empty' | dm }}
+                </p>
+            } @else {
+                @for (role of grantedList(); track role.roleKey) {
```
hmoreras: `track role.roleKey` is empty for keyless roles (`toRoleOption` normalises them to `''`), so granting two of them gives `@for` duplicate keys — Angular flags that (NG0955) and the rows can render wrong. `role.id` is unique and already on the shape. Same for the `data-testid` a few lines down.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.ts
```diff
@@ -1,26 +1,450 @@
...
+            // that loop throws `Cannot alter users on this role` when
+            // it hits a role with `editUsers=false`, and the personal
+            // role always has that flag. Filtering it here keeps the
+            // save from tripping the guard. Proper fix belongs in the
+            // backend (skip editUsers=false roles inside processRoles).
+            const personalRoleKey = this.user?.userId ?? '';
+            const outbound = personalRoleKey
+                ? roleKeys.filter((key) => key !== personalRoleKey)
+                : roleKeys;
+            if (outbound.length > 0) {
```
hmoreras: Revoking everything is a silent no-op: `outbound` is empty so `roles` is left off the payload, and the backend treats a missing (or empty) list as "don't touch". The admin clears the Granted panel, saves, gets a success toast, reopens, and every role is back.

Given the backend can't express "remove all" through this field, the honest frontend move is to stop the save and say so — something like blocking submit while Granted is empty rather than shipping a request that quietly does nothing.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-roles-tab/dot-users-roles-tab.component.ts
```diff
@@ -0,0 +1,488 @@
...
+
+        return list;
+    });
+
+    protected readonly grantedCount = computed(() => this.granted().length);
+    protected readonly canGrant = computed(() => this.selectedAvailable().length > 0);
+    protected readonly canRevoke = computed(() => this.selectedGranted().length > 0);
+
+    constructor() {
+        this.loadRoles();
```
hmoreras, line 365: `loadRoles()` in the constructor means the fan-out fires whenever the dialog opens, not when someone visits the Roles tab — the shell renders all four `p-tabpanel`s eagerly, so editing a name pays for the whole role tree too.

Worth noting the fan-out itself is unavoidable: `loadChildrenRoles=true` only returns one level (each child comes back with an empty children list), so walking the depth per node is the only option. Gating the load on tab activation is the part that's in our hands.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-roles-tab/dot-users-roles-tab.component.ts
```diff
@@ -0,0 +1,488 @@
...
+        return count > 0 && count < leaves.length;
+    }
+
+    protected readonly granted = signal<string[]>([]);
+    protected readonly selectedAvailable = signal<string[]>([]);
+    protected readonly selectedGranted = signal<string[]>([]);
+    protected readonly availableFilter = signal('');
+    protected readonly grantedFilter = signal('');
+    protected readonly collapsed = signal<Record<string, boolean>>({});
+    protected readonly isLoading = signal(false);
```
hmoreras: `isLoading` never gets read in the template, so while the tree is being walked the Available panel shows the "no roles" empty message instead — a slow load and a genuinely empty tree look the same. Same for the error path: `httpErrorManager.handle` fires but the panel still reads as empty.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-roles-tab/dot-users-roles-tab.component.ts
```diff
@@ -0,0 +1,488 @@
...
+        NgTemplateOutlet,
+        FormsModule,
+        ButtonModule,
+        CheckboxModule,
+        InputTextModule,
+        DotMessagePipe
+    ],
+    templateUrl: './dot-users-roles-tab.component.html',
+    styleUrl: './dot-users-roles-tab.component.scss',
+    changeDetection: ChangeDetectionStrategy.OnPush,
```
hmoreras, line 76: `changeDetection` can go (OnPush is the framework default now) and `standalone: true` is implied. The signals want the `$` prefix and the injected deps `#` instead of `private` — `dot-users-filter-by` in this portlet already does all three. Also `host: 'flex flex-col gap-4 block'` sets `display` twice; drop `block`.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-roles-tab/dot-users-roles-tab.component.ts
```diff
@@ -0,0 +1,488 @@
...
+    host: { class: 'flex flex-col gap-4 block' }
+})
+export class DotUsersRolesTabComponent {
```
hmoreras, line 79: 488 lines and no spec. The tree logic is the part that really wants one — `grantableLeavesByRole` memoisation, `isPartiallyChecked` on a multi-level parent, and the pruning in `availableTree` (a root disappearing once its last grantable leaf moves across) are all easy to break and impossible to eyeball.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-roles-tab/dot-users-roles-tab.component.html
```diff
@@ -0,0 +1,169 @@
...
+                    <button
+                        type="button"
+                        class="flex w-full items-start gap-2 rounded-sm p-2 text-left hover:bg-surface-100"
+                        [class.bg-primary-50]="isSelectedGranted(role.roleKey)"
+                        (click)="toggleGrantedSelection(role.roleKey)"
+                        [attr.data-testid]="'users-roles-granted-item-' + role.roleKey">
+                        <span
+                            class="material-symbols-rounded text-base! text-surface-500"
```
hmoreras, line 129: `text-base!` — three `!important`s in this file with no reason beside them. A short comment if the Material Symbols sizing needs it, otherwise plain `text-base` should hold.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-roles-tab/dot-users-roles-tab.component.html
AP2300: Swapped `@for track role.roleKey` → `track role.id` (unique). The click handler + `isSelectedGranted` still use the grant identifier (`roleKey || id`) because that's what the internal `granted` signal is keyed on, and `data-testid` uses `role.id` so keyless-role testids stay unique too. Added an inline HTML comment spelling out why.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-roles-tab/dot-users-roles-tab.component.ts
AP2300: Carried `editUsers` through into `RoleOption` (defaults to `true` when absent so a missing flag stays permissive) and folded it into `isGrantableLeaf` — a leaf with `editUsers === false` now returns `false`, so `grantableLeavesByRole` skips it, `canSelectRole` returns false, and no checkbox is ever shown. That prevents `addRoleToUser` from throwing the `Cannot alter users on this role` guard and dragging the profile edit into a rolled-back transaction.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-roles-tab/dot-users-roles-tab.component.ts
AP2300: Applied the hygiene pass:
- Dropped `standalone: true` (implied).
- `$`-prefix all signals + computeds (`$allRoles`, `$grantableLeavesByRole`, `$availableTree`, `$grantedList`, etc.); template updated in lockstep.
- `#`-prefix native-private for the three injections (`#usersService`, `#httpErrorManager`, `#destroyRef`).
- Dropped `block` from the host class — `flex` and `block` both set `display`.

Kept `changeDetection: OnPush` explicit — app is zoned (not zoneless) so removing it flips CD back to Default and degrades perf on signal-heavy components. Matches the decision on the profile branch (`dot-users-filter-by` next door keeps it too).

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-roles-tab/dot-users-roles-tab.component.ts
AP2300: Wrapped the shell's roles `p-tabpanel` content in `@defer (when activeTab() === 1)` — the roles tab (and its `loadRoles()` call) only instantiates the first time the user opens the tab, not on every dialog open. Once activated it stays instantiated so subsequent tab switches don't re-fetch.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-roles-tab/dot-users-roles-tab.component.ts
AP2300: Bound `$isLoading` in the Available panel's template — the empty-message branch now only fires when the tree is genuinely empty, and a loading state gets its own `data-testid="users-roles-available-loading"`. Error path still routes through the shared `DotHttpErrorManagerService` toast; if that ever needs an inline hint too, the same slot is the right place.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-roles-tab/dot-users-roles-tab.component.html
AP2300: Added an HTML comment above the first `text-base!` explaining the reason — Google Material Symbols ships `.material-symbols-rounded { font-size: 24px }`, which ties tailwind's `text-*` classes on specificity, so `!important` (`!` suffix) is the only clean override. Same rationale applies to the other two occurrences.

Matches how we resolved the equivalent thread on the profile branch (`dot-users-profile-tab.component.html`).

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-roles-tab/dot-users-roles-tab.component.ts
AP2300: Added `dot-users-roles-tab.component.spec.ts` — 15 tests, 82% stmts coverage on the tab. Covers exactly the three flagged risk areas:

- **grantableLeavesByRole memoisation** — via `canSelectRole`: roots with grantable descendants surface checkboxes, `editUsers=false` leaves do not, mixed-descendant parents still do as long as one grantable leaf exists.
- **isPartiallyChecked on multi-level parents** — partial when Sub A1's two leaves are half-selected; partial on Root A when only Sub A1 leaves are picked but Leaf A2 isn't; fully-checked flip when every leaf is in the selection.
- **availableTree pruning** — Root B disappears once granted; Root A disappears once all three of its grantable leaves are on the granted side; Root A stays while any leaf is still ungranted.

Plus grant/revoke flows and the load-error path (surfaces to `DotHttpErrorManagerService`, panel stays empty).

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.ts
AP2300: Same root cause as the profile PR's L440 thread — the `PUT /api/v1/users` endpoint looks up each entry via `loadRoleByKey` (never by id), so keyless roles can't be represented in the payload and the write-through-replace contract drops them.

Held for the same team convo: backend needs either an id-accepting variant on the endpoint, a parallel `roleIds` field, or a scoped access endpoint. Tracked as a follow-up on the profile PR (#36990) — flagging here so the two branches move together.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-roles-tab/dot-users-roles-tab.component.ts
AP2300: You're right — the mechanism note in the code ("worst case backend loadRoleByKey silently no-ops") reads too gently. In practice the guard flow is exactly what you described: `processRoles` wipes every direct role, `UserHelper.addRole` skips the UUID because `loadRoleByKey` returns null, and the role is gone after a save the admin thought was a no-op.

The frontend can't paper over this cleanly — as long as the endpoint keys by `roleKey`, keyless roles can't survive a save. Held with the L348 chain for the team-side fix (id-accepting endpoint / `roleIds` field / scoped access endpoint). Tracked as a follow-up on the profile PR #36990.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.ts
AP2300: Related to the same backend-contract chain (L348 / L141) — the empty-`roles` no-op is really `UtilMethods.isSet(Collection)` on the backend treating `[]` the same as null ("don't touch role membership"). Two paths once the team decides:

- If the endpoint gains "remove all" semantics (e.g. an explicit `clearRoles` flag or a dedicated DELETE), we can lift the current omit and just wire the empty-Granted case through.
- Otherwise the honest frontend fallback you suggested — block Save while Granted is empty with an inline message — is the right stopgap.

Holding both for the team convo alongside the roleKey question.

AP2300: Backend fix landed on main via #37109 (`fix(users): honor empty roles array on PUT /v1/users`) — `PUT /api/v1/users` now treats an explicit `"roles": []` as "remove all" instead of collapsing it to null via `UtilMethods.isSet(Collection)`.

Our FE payload builder already always sets `payload.roles = mergeRoleKeysForSave(...)` (including the empty array case), so no code change was needed here. Added a spec case pinning it: when the Roles tab clears everything and access toggles are all off, the outgoing payload carries `roles: []` and the backend now honors it as intended.

hmoreras (approved):

PR #37083, opened 2026-08-17
feat(users): Permissions tab (legacy iframe embed) (#36719)

### LINKED ISSUE #36719: Users portlet: Permissions tab (iframe embed)

AP2300:
Closes #36719

Depends on #36990 (Profile tab shell). This PR's diff currently includes the shell commits; once #36990 is merged to main, this branch will be rebased onto the new main and the diff will show only the Permissions tab changes.

## Summary

Ships the Permissions tab of the Users dialog as an **iframe embed of the legacy permissions surface**. Full-native reimplementation is deferred — the legacy UI already covers the acceptance criteria and reimplementing it would multiply the scope of the users portlet migration.

- Loads the legacy `/html/portlet/ext/useradmin/*` permissions view scoped to the currently-edited user id.
- Iframe is sized to the tab pane; scroll lives inside the iframe so the dialog chrome stays fixed.
- Placeholder retired: the tab no longer renders \"Coming soon\".

## Notable non-obvious calls

- **Kept iframe scope narrow** to the permissions sub-view specifically, not the entire legacy user editor, so the modernized Profile/Roles/API Tokens tabs aren't shadowed by a stale copy.
- **Auth**: the iframe inherits the parent session cookie, no separate auth handshake needed.

## Test plan

- [ ] Open an existing user; Permissions tab loads the legacy view keyed to the correct userId
- [ ] Grant/revoke a permission via the iframe → refresh the tab → change persists
- [ ] Confirm the iframe doesn't overflow the dialog on smaller viewports

This PR fixes: #36719


core-web/libs/ui/src/lib/components/dot-permissions-iframe-dialog/dot-permissions-iframe.component.ts
```diff
@@ -0,0 +1,65 @@
...
+     */
+    readonly height = input<string | null>(null);
+
+    readonly $iframeSrc = computed<SafeResourceUrl | null>(() => {
+        const url = this.url();
+        if (!url) return null;
+        // Only allow same-origin relative paths; reject absolute URLs,
+        // protocol-relative URLs, and dangerous schemes (javascript:,
+        // data:, http:, etc.)
+        if (!url.startsWith('/') || url.startsWith('//')) return null;
```
hmoreras: This guard came over from the dialog unchanged, so it isn't something the PR introduced — but it's worth tightening now that it's the shared entry point. `/\evil.com` passes both checks and browsers resolve it as protocol-relative: the URL parser treats `\` as `/` in the relative-slash state, so the host becomes `evil.com` and the frame loads cross-origin. Since `bypassSecurityTrustResourceUrl` turns Angular's sanitizer off, this line is the only thing standing there — rejecting a second character of `/` or `\` closes it.

core-web/libs/ui/src/lib/components/dot-permissions-iframe-dialog/dot-permissions-iframe.component.ts
```diff
@@ -0,0 +1,65 @@
...
+ * adapter over this component; see that file for the modal wiring.
+ */
+@Component({
+    selector: 'dot-permissions-iframe',
+    imports: [DotMessagePipe],
+    template: `
+        @let src = $iframeSrc();
+
+        @if (src) {
+            <iframe
```
hmoreras: The iframe renders with no load or error handling, and the JSP behind it emits an empty body when the viewer lacks `users` portlet access or the id doesn't resolve — so a permission failure and a slow load look identical from here: a blank white pane. A `(load)` handler switching off a skeleton, plus a fallback message when nothing arrives, would make the difference visible.

Small aside on the same element: `title="Permissions"` is hardcoded English in a shared lib, so it reaches screen readers untranslated. An input or an i18n key travels better.

core-web/libs/ui/src/lib/components/dot-permissions-iframe-dialog/dot-permissions-iframe.component.ts
```diff
@@ -0,0 +1,65 @@
...
+ * are responsible for building the URL. Only same-origin relative paths
+ * are accepted so a caller cannot navigate the parent frame to an
+ * arbitrary origin.
+ *
+ * Usage as an inline tab:
+ *   <dot-permissions-iframe
+ *     [url]="/html/portlet/ext/useradmin/permissions.jsp?userId=…&popup=true"
+ *     minHeight="34rem" />
+ *
+ * The dialog variant `DotPermissionsIframeDialogComponent` is a thin
```
hmoreras: Good extraction — the presentational/dialog split is what lets the tab embed this without dragging in the dialog machinery. Worth one more step while it's fresh: `dot-roles/dot-role-permissions-iframe` is a fourth hand-rolled copy of this same `bypassSecurityTrustResourceUrl` dance, and its own comment says it follows the dialog's pattern. It could become a thin wrapper passing `[url]` from the store, which would also mean the guard fix above covers it.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-permissions-tab/dot-users-permissions-tab.component.ts
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * In create mode the userId does not exist yet, so we render an empty
+ * state prompting the user to save first.
+ */
+@Component({
+    selector: 'dot-users-permissions-tab',
+    standalone: true,
+    imports: [DotMessagePipe, DotPermissionsIframeComponent],
+    templateUrl: './dot-users-permissions-tab.component.html',
+    changeDetection: ChangeDetectionStrategy.OnPush,
```
hmoreras: `changeDetection` can go (OnPush is the framework default now) and `standalone: true` is implied — the two libs/ui components in this same PR already omit both and use `#config` / `$url`, so this file ends up the odd one out. `userId` and `permissionsUrl` want the `$` prefix for the same reason. Also `host: 'flex h-full min-h-0 flex-col block'` sets `display` twice; dropping `block` removes the ambiguity.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-permissions-tab/dot-users-permissions-tab.component.ts
```diff
@@ -0,0 +1,39 @@
...
+    host: { class: 'flex h-full min-h-0 flex-col block' }
+})
+export class DotUsersPermissionsTabComponent {
```
hmoreras, line 57: No spec for this one, while both libs/ui components in the PR got theirs. Two cases would cover it: the create-mode empty state, and that the built URL carries the userId plus `popup=true` — your comment says that second param is load-bearing, which is exactly what a test should pin down.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-permissions-tab/dot-users-permissions-tab.component.html
```diff
@@ -0,0 +1,17 @@
+@if (permissionsUrl()) {
+    <dot-permissions-iframe
+        [url]="permissionsUrl()"
+        height="100%"
+        class="flex min-h-0 flex-1 flex-col" />
+} @else {
+    <div
+        class="flex h-full min-h-96 flex-col items-center justify-center gap-2 py-8"
+        data-testid="users-permissions-tab-empty">
+        <span class="material-symbols-rounded text-5xl! text-surface-400" aria-hidden="true">
```
hmoreras, line 49: `text-5xl!` is an `!important` with no reason beside it. If the Material Symbols sizing really needs it a short comment would do; otherwise plain `text-5xl` should hold.

core-web/libs/ui/src/lib/components/dot-permissions-iframe-dialog/dot-permissions-iframe.component.ts
AP2300: Tightened the guard: reject when the second char is either `/` (existing) OR `\\` (new). Backslash-relative URLs (`/\evil.com/path`) now bounce with an empty state instead of reaching `bypassSecurityTrustResourceUrl`. Added the corresponding spec case.

core-web/libs/ui/src/lib/components/dot-permissions-iframe-dialog/dot-permissions-iframe.component.ts
AP2300: Landed both parts:

**Load/error handling** — added a `$state` signal that starts at `loading` and flips via a `(load)` handler:
- `loading` — skeleton overlay while the iframe fetches; iframe is opacity-0 so paint flashes don't leak through.
- `loaded` — iframe fades in.
- `empty` — same-origin means we can read `contentDocument.body.innerHTML`; if it lands empty (blank top_inc.jsp when the viewer lacks permission or the id doesn't resolve) we surface a fallback `data-testid="permissions-empty-body"` message instead of a silent blank pane. New i18n key `dot.permissions.iframe.dialog.empty-body`.

**Title** — new `iframeTitle` input (default `Permissions` for backwards compat). The consumer passes the translated `users.dialog.permissions.iframe.title` in — the shared lib no longer bakes in the English string. Added a spec case for the custom title path.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-permissions-tab/dot-users-permissions-tab.component.ts
AP2300: Hygiene pass done:
- Dropped `standalone: true` (implied).
- `$`-prefix on the computed (`$permissionsUrl`). Left `userId` unprefixed on the `input()` because Angular ESLint's `@angular-eslint/no-input-rename` forbids the alias needed to expose `$userId` as a public binding under the name `userId` — matches the convention across the codebase where `input()` names stay canonical and only derived signals get the `$`.
- Dropped `block` from the host class — `flex` and `block` both set `display`. Also fixed the shell's tabpanel wrapper for the same reason.

Kept `changeDetection: OnPush` explicit — app runs zoned, matches the decision on the profile / roles branches (contradicts the verbatim reviewer wording but avoids a CD-perf regression on signal-heavy components).

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-permissions-tab/dot-users-permissions-tab.component.html
AP2300: Added an inline HTML comment above the icon — Google Material Symbols ships `.material-symbols-rounded { font-size: 24px }`, which ties tailwind's `text-*` classes on specificity, so the `!` suffix (compiled to `!important`) is the only clean override. Same rationale as the equivalent threads on the profile / roles branches.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-permissions-tab/dot-users-permissions-tab.component.ts
AP2300: Added `dot-users-permissions-tab.component.spec.ts` — 9 tests split across create mode / edit mode:

**create mode** — no userId: renders `users-permissions-tab-empty` state, no iframe, `$permissionsUrl() === ''`.
**edit mode** — with a userId: iframe renders, URL targets `/html/portlet/ext/useradmin/permissions.jsp` with the userId query param, `popup=true` present (calling out that top_inc.jsp renders blank without it), reserved chars in userId are URL-encoded via `URLSearchParams`, and the translated `users.dialog.permissions.iframe.title` reaches the iframe's `title` attribute.

core-web/libs/ui/src/lib/components/dot-permissions-iframe-dialog/dot-permissions-iframe.component.ts
AP2300: Held for team convo — same class of cross-branch dedupe as the shared user picker follow-up on the profile PR (`dot-users-replacement-picker` vs `dot-roles/dot-role-users-tab`'s hand-rolled autocomplete).

The `dot-roles/dot-role-permissions-iframe` consumer lives on the roles branch (#37082) — extracting the wrapper there **and** on this branch simultaneously is exactly the merge-coordination pain the profile-branch follow-up captured. Best done in a dedicated PR once #37082 lands so both consumers are visible in one tree and the shared home / API surface can be decided in context. When that happens, the L61 guard fix here will cover `dot-roles`'s copy automatically since it'll consume this same component.

Leaving open so the team convo has a place to land the decision.

core-web/libs/portlets/dot-users/src/lib/components/dot-users-replacement-picker/dot-users-replacement-picker.component.ts
```diff
@@ -0,0 +1,110 @@
...
+ * dialog) and the bulk-delete confirm (on the list toolbar).
+ *
+ * The excluded-ids input keeps deletion targets out of the
+ * suggestion list on the client — the backend also rejects invalid
+ * replacements, but pre-filtering avoids showing picks that would
+ * fail on submit.
+ */
+@Component({
+    selector: 'dot-users-replacement-picker',
+    standalone: true,
```
hmoreras: remove not needed.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-permissions-tab/dot-users-permissions-tab.component.ts
```diff
@@ -0,0 +1,39 @@
...
+ * shared `view_role_permissions_inc.jsp` fragment. `popup=true` is
+ * mandatory — without it the legacy `top_inc.jsp` renders
+ * `<body style="visibility:hidden">` and the iframe shows up blank.
+ *
+ * In create mode the userId does not exist yet, so we render an empty
+ * state prompting the user to save first.
+ */
+@Component({
+    selector: 'dot-users-permissions-tab',
+    standalone: true,
```
hmoreras: remove not needed.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-profile-tab/dot-users-profile-tab.component.ts
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * Profile tab of the Create/Edit User dialog. Owns the visuals — the
+ * shell component owns the `FormGroup` reference and dispatches saves,
+ * so this component stays purely presentational and testable in
+ * isolation.
+ */
+@Component({
+    selector: 'dot-users-profile-tab',
+    standalone: true,
```
hmoreras: remove not needed.

core-web/libs/portlets/dot-users/src/lib/components/dot-users-replacement-picker/dot-users-replacement-picker.component.ts
```diff
@@ -0,0 +1,110 @@
...
+    /** Currently selected user, or null when the picker is empty. */
+    readonly value = input<DotUserListItem | null>(null);
+
+    /** Emits every selection change (user or null when cleared). */
+    readonly selectionChange = output<DotUserListItem | null>();
+
+    protected readonly suggestions = signal<DotUserListItem[]>([]);
+
+    protected onSearch(event: AutoCompleteCompleteEvent): void {
+        this.usersService
```
hmoreras: we should handle this kind of calls directly in the store.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.scss
```diff
@@ -0,0 +1,30 @@
+:host {
```
hmoreras: lets try to avoid scss and move this to tailwinds rules.

hmoreras (changes requested):

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-permissions-tab/dot-users-permissions-tab.component.ts
AP2300: Applied on `dot-users-permissions-tab` — dropped `changeDetection: ChangeDetectionStrategy.OnPush` (default in Angular 22+) and trimmed the now-unused `ChangeDetectionStrategy` import. Also swept the shared `dot-permissions-iframe` component in `libs/ui` since it's touched by this branch. Saved as a general rule for future components (applies to `standalone: true` too — same reasoning).

core-web/libs/portlets/dot-users/src/lib/components/dot-users-replacement-picker/dot-users-replacement-picker.component.ts
AP2300: Held for a follow-up — this file lives on `main` (came in with #36990) and my rebased branch doesn't currently touch it. Same rule is in-scope wherever `dot-users-replacement-picker` gets edited next (e.g. the shared-picker extraction follow-up alongside `dot-role-users-tab`'s hand-rolled autocomplete). Saved as a general rule so we don't miss it on the next pass.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-profile-tab/dot-users-profile-tab.component.ts
AP2300: Same call as the picker thread — `dot-users-profile-tab` is on `main` from #36990 and this branch doesn't modify it after the rebase. Saved as a general rule; will apply on any branch that touches the profile tab next (e.g. the Signal Forms migration, if that ever happens).

core-web/libs/portlets/dot-users/src/lib/components/dot-users-replacement-picker/dot-users-replacement-picker.component.ts
AP2300: Held for a follow-up — the picker lives on `main` from #36990 and this branch doesn't modify it. Recorded as a **general rule** for the codebase (component MUST NOT run HTTP directly; either the surrounding feature store owns the call, or a component-scoped SignalStore does), alongside the existing dedupe follow-up against `dot-role-users-tab`'s hand-rolled autocomplete on the roles branch — the cleanest time to do both is when the shared picker gets extracted.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.scss
AP2300: Held for a follow-up — `dot-users-create.component.scss` lives on `main` from #36990 and my rebase doesn't touch it. Recorded as a **general rule** (prefer Tailwind utilities, `::ng-deep` is a red flag pointing at either wrong-element utility placement or a theme-layer fix). Concretely for this file: the layout inside `:host ::ng-deep { p-tabs, p-tabpanels, p-tabpanel { ... } }` should move up onto the host class + tabpanel wrapper classes — will land on whichever branch modifies the shell next.

hmoreras (dismissed):

hmoreras (dismissed):

hmoreras (approved):

oidacra (approved):

PR #37084, opened 2026-08-17
feat(users): API Tokens tab wired to real endpoints (#36720)

### LINKED ISSUE #36720: Users portlet: API Tokens tab

AP2300:
Closes #36720

Depends on #36990 (Profile tab shell). This PR's diff currently includes the shell commits; once #36990 is merged to main, this branch will be rebased onto the new main and the diff will show only the API Tokens tab changes.

## Summary

Ships the **API Tokens tab** wired against the modern `ApiTokenResource` (`/api/v1/apitoken/*`) — visual and behavioral parity with the legacy admin's token surface.

- **List** — `GET /api/v1/apitoken/{userId}/tokens?showRevoked={bool}` with a \"Show revoked\" toggle that mirrors the legacy checkbox.
- **Create** — a small sub-dialog collects Label + Expires Date + Allow Network, converts the date to \`expirationSeconds\`, POSTs to `/api/v1/apitoken` and returns `{jwt, token}`.
- **Reveal** — clicking a valid row opens a modal that calls `GET /api/v1/apitoken/{id}/jwt`. Backend mints a fresh signed JWT on every call (\"Generates a new JWT for an existing token\" / \"overwrites the JWT value\"), so re-revealing is safe by design — no exposure risk from a persisted token value.
- **Revoke** — `PUT /api/v1/apitoken/{id}/revoke` behind a `p-confirmDialog`. Row stays in the list for audit.
- **Revoked / expired rows** — muted (`text-surface-400 opacity-60`), non-clickable, action cell shows a small `Revoked` pill instead of a button.

## Notable non-obvious calls

- **Dropped legacy \"Get Token\" separate button.** A single click on the row does the same thing.
- **Dropped delete affordance** for now — revoked rows show a static \"Revoked\" indicator only. The DELETE endpoint still exists; we can add a Delete button later if audits build up.
- **Create-mode users see a hint**, not the table. A userId is required to key any of the token endpoints.

## Test plan

- [ ] Open an existing user with tokens — table lists them; toggling \"Show revoked\" refetches
- [ ] Click a valid row → reveal dialog opens, JWT lands, Copy button works
- [ ] Close and re-open the reveal → a fresh JWT string arrives (different signature)
- [ ] Request a new token → dialog closes, list refreshes with the new row, reveal dialog opens with the returned JWT
- [ ] Revoke a valid token → confirmation modal → on accept the row flips to muted \"Revoked\", becomes non-clickable
- [ ] Open the dialog in create mode → tab body renders the \"save the user first\" hint

This PR fixes: #36720


core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-api-tokens-tab.component.html
```diff
@@ -0,0 +1,173 @@
...
+                            class="cursor-pointer border-0 bg-transparent p-0 text-sm font-medium text-red-500 hover:underline"
+                            (click)="revoke(token, $event)"
+                            [attr.data-testid]="'users-api-tokens-revoke-' + token.id">
+                            {{ 'users.dialog.tokens.revoke' | dm }}
+                        </button>
+                    } @else {
+                        <span
+                            class="inline-block rounded-md bg-surface-100 px-2 py-1 text-xs text-surface-400"
+                            [attr.data-testid]="'users-api-tokens-revoked-pill-' + token.id">
+                            {{ 'users.dialog.tokens.revoked' | dm }}
```
hmoreras: `statusOf` distinguishes `expired` from `revoked`, but the template collapses both into the "Revoked" pill — a token that simply aged out shows up as Revoked, which isn't true. Worth a second branch (or deriving the label key from `status`) so expired rows say Expired; right now the `'expired'` arm of `TokenStatus` never reaches the UI.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-api-tokens-tab.component.html
```diff
@@ -0,0 +1,173 @@
...
+                        [class.bg-surface-100]="!isInactive"
+                        [class.bg-surface-50]="isInactive">
+                        {{ token.allowNetwork || ('users.dialog.tokens.network.any' | dm) }}
+                    </span>
+                </td>
+                <td>
+                    @if (!isInactive) {
+                        <button
+                            type="button"
+                            class="cursor-pointer border-0 bg-transparent p-0 text-sm font-medium text-red-500 hover:underline"
```
hmoreras: Hand-rolled `<button>` with a hardcoded `text-red-500`, while everything else in this file uses `p-button` — `<p-button severity="danger" [text]="true" size="small">` gets the same look off the theme. Same for the pill just below: `p-tag` is what we use for read-only status elsewhere in the portlet, like the access chips in the list.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-api-tokens-tab.component.html
```diff
@@ -0,0 +1,173 @@
...
+                        "
+                        [outlined]="!copied()"
+                        (onClick)="copyJwt()"
+                        data-testid="users-api-tokens-reveal-copy" />
+                </div>
+            } @else {
+                <div
+                    class="text-color-secondary flex items-center gap-2 text-sm"
+                    data-testid="users-api-tokens-reveal-loading">
+                    <i class="pi pi-spinner pi-spin"></i>
```
hmoreras: `pi pi-spinner` is PrimeIcons — the rest of this portlet is on Material Symbols (`material-symbols-rounded` in the list and the profile tab). `progress_activity` with `animate-spin` matches.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-api-tokens-tab.component.html
```diff
@@ -0,0 +1,173 @@
...
+        </ng-template>
+        <ng-template pTemplate="body" let-token>
+            @let status = statusOf(token);
+            @let isInactive = status !== 'valid';
+            <tr
+                [class.cursor-pointer]="!isInactive"
+                [class.text-surface-400]="isInactive"
+                [class.opacity-60]="isInactive"
+                [attr.title]="!isInactive ? ('users.dialog.tokens.row.reveal-hint' | dm) : null"
+                (click)="onRowClick(token)"
```
hmoreras, line 68: Revealing the JWT is the row's primary action but it's mouse-only — the `<tr>` isn't focusable and there's no key handler, with `title` as the only hint. Since the separate Reveal button was dropped in favour of the row click, this is now the only path to the JWT, so keyboard users have none.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-api-tokens-tab.component.html
```diff
@@ -0,0 +1,173 @@
...
+                            [attr.data-testid]="'users-api-tokens-revoked-pill-' + token.id">
+                            {{ 'users.dialog.tokens.revoked' | dm }}
+                        </span>
+                    }
+                </td>
+            </tr>
+        </ng-template>
+        <ng-template pTemplate="emptymessage">
+            <tr class="hover:bg-transparent!">
+                <td colspan="7" class="border-none p-0">
```
hmoreras, line 111: A failed fetch leaves `tokens` empty, so a 500 renders as "No active tokens." and the toast is the only sign something broke. The copy is also off when Show inactive is checked and there really are no tokens — it isn't only active ones being listed at that point.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-api-tokens-tab.component.ts
```diff
@@ -0,0 +1,287 @@
...
+        ConfirmDialogModule,
+        DialogModule,
+        InputTextModule,
+        TableModule,
+        DotMessagePipe
+    ],
+    templateUrl: './dot-users-api-tokens-tab.component.html',
+    styleUrl: './dot-users-api-tokens-tab.component.scss',
+    providers: [DialogService, ConfirmationService],
+    changeDetection: ChangeDetectionStrategy.OnPush,
```
hmoreras: `OnPush` is the framework default now so `changeDetection` can go, and `standalone: true` is implied. The signals here want the `$` prefix and the injected deps `#` instead of `private` — `dot-users-filter-by` in this portlet is the closest reference. Also `host: 'flex flex-col gap-4 block'` sets `display` twice; dropping `block` removes the ambiguity.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-api-tokens-tab.component.ts
```diff
@@ -0,0 +1,287 @@
...
+     */
+    protected readonly revealVisible = signal(false);
+    protected readonly revealJwt = signal<string | null>(null);
+    protected readonly revealTokenId = signal<string>('');
+    protected readonly copied = signal(false);
+
+    protected readonly hasUser = computed(() => !!this.userId());
+
+    constructor() {
+        effect(() => {
```
hmoreras, line 119: This effect reads `userId` / `showRevoked` and then writes `isLoading` and `tokens`, which is exactly what `untracked()` is for. `dot-role-users-tab` over in dot-roles does the same load-on-change and wraps it: `untracked(() => this.loadTokens(id, showRevoked))`.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-api-tokens-tab.component.ts
```diff
@@ -0,0 +1,287 @@
...
+        }
+
+        try {
+            await navigator.clipboard.writeText(jwt);
+            this.copied.set(true);
+        } catch {
+            // Clipboard API can be blocked (insecure context, permission
+            // policy). Fall back to selecting the field so the admin can
+            // Cmd+C manually.
+            const el = document.querySelector<HTMLInputElement>(
```
hmoreras: Reaching into the DOM by `data-testid` couples the clipboard fallback to a test hook — rename it for a spec and the fallback silently stops working. A `viewChild` on the input gives the same element with a compile-time reference.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-api-tokens-tab.component.ts
```diff
@@ -0,0 +1,287 @@
...
+type TokenStatus = 'valid' | 'revoked' | 'expired';
+
+/**
+ * API Tokens tab. Matches the legacy admin's contract:
+ *
+ * - Clicking a valid row (or Reveal button) fetches a FRESH JWT via
+ *   `GET /api/v1/apitoken/{id}/jwt`. Each call mints a new signed
+ *   value on the backend — the token record is unchanged, so there
+ *   is no "reveal only once" limitation to worry about.
+ * - Revoke (soft) keeps the row visible for audit; Delete purges it,
```
hmoreras: This says Delete purges the row and that the button only shows for revoked/expired rows, but there's no Delete button in the template — the description says it was dropped on purpose. Worth trimming the comment so the next reader doesn't go hunting for it.

core-web/libs/portlets/dot-users/src/lib/services/dot-users.service.ts
```diff
@@ -30,6 +32,128 @@ export interface DotUserListItem {
...
+export interface DotApiTokenCreatePayload {
+    userId: string;
+    expirationSeconds: number;
+    network?: string;
+    claims?: { label?: string } & Record<string, unknown>;
+}
+
+/**
+ * Response envelope for POST /api/v1/apitoken. The raw JWT string is
+ * shown to the caller exactly once (design choice) — after that the
```
hmoreras: This says the JWT is shown exactly once and can't be re-revealed, and `createApiToken` below repeats it — but `getApiTokenJwt` re-reveals on every row click, which is what the tab actually does.

The good news is the tab is right and these two comments are wrong: `ApiTokenAPI.getJWT` just calls `generateApiToken(token)`, a stateless sign over the persisted record with no DB write. Worth fixing the wording, since as written it argues against the feature.

core-web/libs/portlets/dot-users/src/lib/services/dot-users.service.ts
```diff
@@ -83,4 +207,86 @@ export class DotUsersService {
...
+    createApiToken(payload: DotApiTokenCreatePayload): Observable<DotApiTokenCreateResult> {
+        return this.#http
+            .post<DotApiTokenCreateResponse>('/api/v1/apitoken', payload)
+            .pipe(map((response) => response.entity));
+    }
+
+    /**
+     * Mints a fresh JWT for an existing token id. The token record
+     * itself is unchanged — the backend simply signs a new JWT value
+     * over the same token, which supersedes any previously-issued one.
```
hmoreras: `generateApiToken` only signs — it doesn't touch the record or invalidate anything, so every JWT previously minted for this token stays valid until it's revoked or expires. "Supersedes any previously-issued one" reads like rotation, which could leave someone thinking a re-reveal kills a leaked JWT. Revoke is the only thing that does.

core-web/libs/portlets/dot-users/src/lib/services/dot-users.service.ts
```diff
@@ -83,4 +207,86 @@ export class DotUsersService {
...
+     * Backend refuses this on revoked/expired tokens (400).
+     */
+    getApiTokenJwt(tokenId: string): Observable<string> {
+        return this.#http
+            .get<{ entity: { jwt: string } }>(
+                `/api/v1/apitoken/${encodeURIComponent(tokenId)}/jwt`
+            )
+            .pipe(map((response) => response.entity?.jwt ?? ''));
```
hmoreras: `?? ''` turns a malformed response into an empty string, and the reveal dialog treats `''` as falsy — so the spinner sits there forever instead of surfacing anything. Letting it come through as null and handling it in the error path would at least fail visibly.

core-web/libs/portlets/dot-users/src/lib/services/dot-users.service.ts
```diff
@@ -83,4 +207,86 @@ export class DotUsersService {
...
+    revokeApiToken(tokenId: string): Observable<unknown> {
+        return this.#http.put(`/api/v1/apitoken/${encodeURIComponent(tokenId)}/revoke`, null);
+    }
+
+    /**
+     * Hard-deletes a token. Backend guards this to revoked or expired
+     * tokens only — the UI mirrors the guard so the button only shows
+     * for those rows.
+     */
+    deleteApiToken(tokenId: string): Observable<unknown> {
```
hmoreras, line 494: `deleteApiToken` isn't called anywhere (only mocked in a spec), and the comment describes a guard the UI mirrors even though there's no Delete affordance — worth dropping until the button lands, or leaving a TODO.

While you're in here: `dot-users.service.spec.ts` covers the filter and delete-user URLs but none of the five new token methods. The `showRevoked` param and the two response-envelope unwraps would be cheap to pin.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-request-token-dialog.component.ts
```diff
@@ -0,0 +1,120 @@
...
+    standalone: true,
+    imports: [
+        ReactiveFormsModule,
+        ButtonModule,
+        InputTextModule,
+        DotFieldRequiredDirective,
+        DotMessagePipe
+    ],
+    templateUrl: './dot-users-request-token-dialog.component.html',
+    changeDetection: ChangeDetectionStrategy.OnPush
```
hmoreras: Same decorator/naming notes as the tab (`changeDetection`, `standalone`, `$` on the signals, `#` on the deps). No spec for this dialog either — the required-field validation, the `past` guard and the close-with-result path are the interesting bits.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-request-token-dialog.component.ts
```diff
@@ -0,0 +1,120 @@
...
+                    this.dialogRef.close(result);
+                },
+                error: (error) => {
+                    this.submitting.set(false);
+                    this.httpErrorManager.handle(error);
+                }
+            });
+    }
+
+    private expirationSecondsFrom(expiresIsoDate: string): number {
```
hmoreras, line 102: `defaultExpiryDate` emits `YYYY-MM-DD`, and `new Date('2029-08-19')` parses as UTC midnight while `Date.now()` is local — so the TTL is off by the browser's offset and the token expires at UTC midnight of the chosen date, which is the previous evening for anyone west of UTC. It also means picking today's date gets rejected as `past` in those timezones. Harmless at the three-year default; building the `Date` from the parts keeps it local.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-api-tokens-tab.component.html
AP2300: Fixed — the pill now flips between `Expired` and `Revoked` off `statusOf()`; two new i18n keys. Covered by `should label an expired-but-not-revoked row as Expired (not Revoked)` and `should label a revoked row as Revoked`.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-api-tokens-tab.component.html
AP2300: Fixed — hand-rolled Revoke `<button>` → `p-button [text]="true" severity="danger" size="small"`; hand-rolled Revoked pill → `p-tag` with the label derived from `statusLabelKey()`.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-api-tokens-tab.component.html
AP2300: Fixed — `pi pi-spinner` → `<span class="material-symbols-rounded animate-spin text-base!">progress_activity</span>` to match the portlet-wide Material Symbols family.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-api-tokens-tab.component.html
AP2300: Fixed — the row is now `role="button" tabindex="0"` with an `onRowKeydown` handler (Enter and Space open reveal, Space preventDefault, ignores keydown on inactive rows). Covered by the two keydown tests in the tab spec.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-api-tokens-tab.component.html
AP2300: Fixed — added a `$loadError` signal so the empty-message branches between the empty template and an error template with a Retry button. Also switched the empty copy between `users.dialog.tokens.empty` and `users.dialog.tokens.empty-all` based on `$showRevoked()`.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-api-tokens-tab.component.ts
AP2300: Fixed — dropped `changeDetection` and `standalone: true`, `$`-prefixed the signals, `#`-prefixed the injections, dropped `block` from the host so `display` is only set once. Kept the `userId` input unprefixed because `@angular-eslint/no-input-rename` blocks the alias needed to expose `$userId` under the `userId` binding.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-api-tokens-tab.component.ts
AP2300: Fixed — wrapped the load body in `untracked(() => …)` inside the effect. Matches the pattern in `dot-role-users-tab`.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-api-tokens-tab.component.ts
AP2300: Fixed — replaced the `document.querySelector('[data-testid=…]')` lookup with a `viewChild('revealInput')` signal query. `viewChild` doesn't allow ES-private fields (Angular compiler constraint), so the field stays `protected readonly`.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-api-tokens-tab.component.ts
AP2300: Fixed — trimmed the class-level comment. The Delete button was dropped from the template on purpose and the doc no longer implies otherwise.

core-web/libs/portlets/dot-users/src/lib/services/dot-users.service.ts
AP2300: Fixed — rewrote the JSDoc. The JWT in the create response isn't a one-shot secret; `getApiTokenJwt` re-mints an equivalent JWT any time. The old wording argued against a feature the tab actually ships.

core-web/libs/portlets/dot-users/src/lib/services/dot-users.service.ts
AP2300: Fixed — dropped the "supersedes any previously-issued one" phrase. `generateApiToken` is stateless — previous JWTs stay valid until the token is revoked or expires. Revoke is the only kill switch; comment now says so.

core-web/libs/portlets/dot-users/src/lib/services/dot-users.service.ts
AP2300: Fixed — the map now `throw`s `Malformed JWT response` instead of `?? ''`. The reveal dialog's error branch fires and the tab's `httpErrorManager` handles it. Covered by `should throw when the JWT envelope is malformed instead of returning ""`.

core-web/libs/portlets/dot-users/src/lib/services/dot-users.service.ts
AP2300: Fixed — rewrote the JSDoc to name `deleteApiToken` as 'no UI surface yet' (the guard the old note described does not exist in the FE anyway). Also added 9 new service tests to `dot-users.service.spec.ts` covering the 5 token methods — `showRevoked` param, envelope unwraps, malformed-JWT throw, revoke PUT null body, DELETE.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-request-token-dialog.component.ts
AP2300: Fixed — dropped `changeDetection` and `standalone: true`, `$`-prefixed the signals, `#`-prefixed the injections. Added `dot-users-request-token-dialog.component.spec.ts` — required-label guard, `past` guard, close-with-result, trimmed label + local-timezone TTL derivation, HTTP error surface, cancel.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-api-tokens-tab/dot-users-request-token-dialog.component.ts
AP2300: Fixed — the derived `Date` is now built from local Y/M/D parts via a regex parse (`new Date(year, monthIndex, day)`) instead of `new Date('YYYY-MM-DD')` (which is UTC midnight); `defaultExpiryDate` also emits local Y/M/D so the value round-trips through the date input in the user's timezone. Covered by the 9.5-day TTL test in the dialog spec.

hmoreras (dismissed):

rjvelazco (dismissed):

oidacra (dismissed):

hmoreras (approved):

PR #37086, opened 2026-08-17
fix(sdk): render dotImage href/target as an anchor in Block Editor renderers (#36998)

### LINKED ISSUE #36998: Block Editor SDK renderers (Angular & React) drop href/target on dotImage nodes — linked images not clickable on frontend
### Problem Statement

After #36361 / PR #36368 shipped in 26.07.13-01, the Block Editor correctly saves `href` and `target` as attributes on a `dotImage` node when an editor links an image. However, neither the `@dotcms/angular` nor the `@dotcms/react` SDK's built-in Block Editor renderer reads those attributes when rendering the node on the frontend. The image renders as a plain `<figure></figure>` with no anchor, so the link — while correctly stored and visible as "linked" inside the CMS editor — never becomes clickable on the live page.

PR #36368's own description confirms the data layer was never the gap: "The `dotImage` node already supports `href`/`target` end-to-end... the break was purely in the UI flow." That PR fixed the UI flow (editor side). This issue is the renderer-side counterpart that was never addressed.

#### Affected components

- **`@dotcms/angular`** — `DotImageBlock` component (selector `dotcms-block-editor-renderer-image`), confirmed present through the current published latest (26.8.10-1)
- **`@dotcms/react`** — `DotCMSImage` component at `DotCMSBlockEditorRenderer/components/blocks/Image.tsx`, confirmed present through the current published latest (26.8.10-1)

#### Evidence

**Angular SDK** — `DotImageBlock` template (verified directly against the published `@dotcms/angular@26.8.10-1` package, `fesm2022/dotcms-angular.mjs`):

```html
<figure [style]="$wrapperStyle()">
    <img [alt]="attrs?.['alt']" [src]="$srcURL()" />
</figure>
```

Only `src`, `alt`, `textWrap`, `textAlign` are read from `attrs`. No `href`, no `target`.

**React SDK** — `Image.tsx` (verified directly against the published `@dotcms/react@26.8.10-1` package, `lib/next/components/DotCMSBlockEditorRenderer/components/blocks/Image.esm.js`):

```js
const { src, alt, textWrap, textAlign } = node.attrs;
...
return <figure style={wrapperStyle}><img alt={alt} src={src} .../></figure>;
```

Same gap — `href`/`target` never destructured or used.

**Customer reproduction** (Freshdesk #38732) — actual frontend DOM from the customer's Chrome DevTools:

```html
<dotcms-block-editor-renderer-image>
  <figure>
    <img alt="vuso_logo.png" src="/dA/0c3abd69-299d-4bb1-b367-b76486b59c3c">
  </figure>
</dotcms-block-editor-renderer-image>
```

Confirms the Angular renderer path is exactly what's hit in production — the component name in the DOM is the literal Angular selector for `DotImageBlock`.

### Steps to Reproduce

1. In the Block Editor, insert an image into a rich text field.
2. Select the image, click the Link toolbar button, set a URL (optionally "open in new tab").
3. Save and publish.
4. Confirm in the editor that the image shows as linked (works correctly since 26.07.13-01).
5. View the published page on the frontend (Angular or React SDK-rendered site).
6. Inspect the DOM around the image — no `<a>` wrapper is present; the image is not clickable.

### Acceptance Criteria

The SDK's built-in `dotImage` renderer should wrap the `` in an `<a href="..." target="...">` when `href` is present on the node's `attrs`, mirroring what `dotImage.vtl` does server-side for Velocity-rendered pages (confirmed present in `dotCMS/src/main/webapp/WEB-INF/velocity/static/storyblock/dotImage.vtl` on `main`, including the `#if ($!item.attrs.href != "null")` branch — though note that template itself has no `target` support either, a related but separate gap).

### dotCMS Version

26.07.30-01

### Severity

### Links

https://dotcms.freshdesk.com/a/tickets/38732

adrianjm-dotCMS:
## What

Fixes #36998 — a linked image authored in the Block Editor rendered without an anchor on the frontend, so it was not clickable.

The data layer was never the gap. The Block Editor stores the link as `href`/`target` on the `dotImage` node (`image.node.ts:63-77`) and its own `renderHTML` wraps the image in an anchor — which is why the link looks correct inside the CMS. The renderers simply never read those attributes.

## Approach

Wrap the `` in `<a href target>` when `href` is present, mirroring the editor's `renderHTML` output. Absent `href` leaves the existing markup untouched, so this is additive for every unlinked image.

The guard is a truthiness check rather than a null check because the editor produces two "no link" states: `href: null` when never set, and `href: ''` after `unsetImageLink` (`image.node.ts:110`).

## Scope

The issue was reported against `@dotcms/angular` and `@dotcms/react`. The same gap existed in `@dotcms/vue` and in the Velocity template (`dotImage.vtl`, where `href` was already handled but `target` was not), so both are fixed here as well.

## Implementation notes

**`rel="noopener noreferrer"` is emitted when `target="_blank"`.** The editor's own `renderHTML` does not emit it; this is an intentional divergence to prevent reverse tabnabbing on the published page.

**Angular binds `target`/`rel` as attributes, not properties.** With `[target]`/`[rel]`, Angular renders the literal string `"null"`, so every linked image without a target would ship `rel="null"`. `[attr.*]` omits the attribute instead. Covered by tests in both Angular renderers.

## Testing

```
sdk-angular   246 passed
sdk-react     138 passed
sdk-vue        55 passed   (DotImage.spec.ts is new)
lint 3/3, build 3/3
```

Each renderer covers: anchor with correct `href`; `figure` wrapper style preserved when linked; `target` + `rel` on `_blank`; neither attribute on same-tab links; bare `` for `href: null` and `href: ''`.

Manual — each SDK built and swapped into its example app against a live instance, plus the Velocity path:

- `examples/angular` (semantic renderer) — anchor renders, image clickable
- `examples/nextjs` (React) — confirmed
- `examples/vuejs` (Vue) — confirmed
- Velocity page with three images — no link, `_blank`, and same-tab — renders no anchor, `target`+`rel`, and a bare anchor respectively

This PR fixes: #36998

### DIFF
--- a/core-web/libs/sdk/angular/src/lib/components/dotcms-block-editor-renderer-semantic/dotcms-block-editor-renderer-native.component.html
+++ b/core-web/libs/sdk/angular/src/lib/components/dotcms-block-editor-renderer-semantic/dotcms-block-editor-renderer-native.component.html
@@ -162,11 +162,23 @@
 
             @case (BLOCKS.DOT_IMAGE) {
                 <figure [style]="imageStyle(node.attrs)">
+                    @if (imageLink(node.attrs); as link) {
+                        <!-- target/rel are attribute bindings so a null value omits the
+                             attribute instead of rendering the string "null". -->
+                        <a [href]="link.href" [attr.target]="link.target" [attr.rel]="link.rel">
+                            <ng-container *ngTemplateOutlet="dotImageEl" />
+                        </a>
+                    } @else {
+                        <ng-container *ngTemplateOutlet="dotImageEl" />
+                    }
+                </figure>
+
+                <ng-template #dotImageEl>
                     <img
                         [alt]="node.attrs?.['alt']"
                         [src]="node.attrs?.['src']"
                         style="max-width: 100%; height: auto" />
-                </figure>
+                </ng-template>
             }
 
             @case (BLOCKS.DOT_VIDEO) {
--- a/core-web/libs/sdk/angular/src/lib/components/dotcms-block-editor-renderer-semantic/dotcms-block-editor-renderer-native.component.ts
+++ b/core-web/libs/sdk/angular/src/lib/components/dotcms-block-editor-renderer-semantic/dotcms-block-editor-renderer-native.component.ts
@@ -136,8 +136,9 @@ export class DotCMSBlockEditorRendererNativeComponent implements OnInit {
 
     /**
      * The link assigned to a `dotImage` in the Block Editor, stored as
-     * `href`/`target` on the node. Returns `null` when the image has no link —
-     * `href` is `null` when never set and `''` once the editor unsets it.
+     * `href`/`target` on the node. Returns `null` when the image has no link,
+     * i.e. when `href` is `null`. The truthiness check also covers the
+     * transient `''` the editor writes while unsetting a link.
      *
      * `rel` guards against reverse tabnabbing when the link opens in a new tab.
      */
--- a/core-web/libs/sdk/angular/src/lib/components/dotcms-block-editor-renderer-semantic/dotcms-block-editor-renderer-native.component.ts
+++ b/core-web/libs/sdk/angular/src/lib/components/dotcms-block-editor-renderer-semantic/dotcms-block-editor-renderer-native.component.ts
@@ -134,6 +134,27 @@ export class DotCMSBlockEditorRendererNativeComponent implements OnInit {
         return {};
     }
 
+    /**
+     * The link assigned to a `dotImage` in the Block Editor, stored as
+     * `href`/`target` on the node. Returns `null` when the image has no link —
+     * `href` is `null` when never set and `''` once the editor unsets it.
+     *
+     * `rel` guards against reverse tabnabbing when the link opens in a new tab.
+     */
+    imageLink(
+        attrs: BlockEditorNode['attrs']
+    ): { href: string; target: string | null; rel: string | null } | null {
+        const href = attrs?.['href'];
+
+        if (!href) {
+            return null;
+        }
+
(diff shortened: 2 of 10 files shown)


zJaaal (approved):

adrianjm-dotCMS:
Applied. Both occurrences now check `""` alongside `"null"`, via an `$imageHref` variable rather than a repeated expression — the opening and closing conditions have to stay identical or the template emits an unbalanced anchor.

One correction to the stated mechanism, because it matters for the record: **the editor does not persist `href: ''`.**

`unsetImageLink` does write `href: ''` into ProseMirror state (`image.node.ts:110`) — that part of the analysis is accurate, and `target` is indeed left untouched there. But that value is normalized before it reaches stored content. Verified on a live instance: after unlinking an image in the Block Editor and publishing, the content API returns

```json
"href": null, "target": null
```

and the Velocity page renders a bare `` with no anchor. So the `<a href="" target="_blank" rel="noopener noreferrer">` output is not reachable through the editor, and the `target`/`rel` emission added by this PR does not compound anything in that path.

The guard is still worth having, for a different reason than the one given: `''` can arrive from non-editor write paths (REST, imports, migrations), and all four JS renderers already guard both forms. Leaving the Velocity template laxer than the SDKs is exactly the asymmetry this PR is trying to remove — so it goes in as hardening and parity, not as a regression fix.

Re-verified after the change on a Velocity page with three images (unlinked, `_blank`, same-tab). Output is unchanged, which is the expected result for a no-op on reachable states:

```html
<img …>                                                                       <!-- unlinked -->
<a href="…" target="_blank" rel="noopener noreferrer"><img …></a>              <!-- _blank -->
<a href="…"><img …></a>                                                        <!-- same tab -->
```

Issue #36471, opened 2026-07-08
OpenSearch: phase rollback during in-flight reindex wipes OS index store; partial orphan indices later adopted, causing silent incomplete search results


swicken:
### Problem Statement

During the ES→OS migration, changing `FEATURE_FLAG_OPEN_SEARCH_PHASE` from `1` to `0` **while a full reindex is in flight** (`dist_reindex_journal` still draining) corrupts the OS side of the index store, and the damage stays invisible until later:

1. The remaining ~journal entries index to ES only (phase 0), so the OS reindex pair is left **partially populated**.
2. The phase-0 reindex switchover promotes the ES pair and **replaces the `indicies` store with ES-only rows — the active `.os` working/live rows are deleted**. dotCMS forgets the OS indices exist; the partial `.os` reindex pair is orphaned on the cluster.
3. On the still-running node the problem is masked: OS reads keep working via the surviving `VersionedIndicesCache`, so DB and cache silently disagree.
4. After any **restart** in Phase ≥ 1, boot catchup mirrors the ES index names — whose `.os` twins are exactly the partial orphans — and **adopts them as the active OS indices**. `dist_reindex_journal` is 0, so nothing heals.

Result: in Phase 2/3, search silently returns a fraction of the content with **zero errors or warnings**. Measured on the demo starter after the sequence below: `+catchall:snow*` returned **8 of 24**, `+contentType:Blog +live:true` **0 of 6**, `+contentType:webPageContent +live:true` **6 of 28**. Only another full reindex repairs it.

This contradicts the migration's documented rollback guarantee ("rollbacks among Phases 0↔1↔2 are lossless with no reindex" — `docs/backend/OPENSEARCH_MIGRATION.md`), and the window is realistic: at customer scale a full reindex runs for a long time, and rolling the phase flag back is the intuitive "abort" action if the reindex looks wrong. The runbook's rollback-risk section does not cover it.

Found during QA-G17 (#36453) adversarial testing; full evidence in that ticket's QA notes.

### Steps to Reproduce

Rig: `docker/docker-compose-examples/single-node-os-migration/` (demo starter, phase driven via system table so flips are restart-free).

1. Phase 1 with ES and OS in sync (demo starter reindexed: ES and OS both 682/681 working/live).
2. `POST /api/v1/esindex/reindex` — new reindex pair (e.g. ts `20260708132710`) is created on **both** engines.
[...]

PR #37103, opened 2026-08-18
docs(opensearch): add the ES → OpenSearch migration runbook for Support and Cloud

### LINKED ISSUE #37102: Write the definitive ES → OpenSearch migration runbook for Support and Cloud
### Description

We have architecture and QA material for the ES → OpenSearch migration (`docs/backend/OPENSEARCH_MIGRATION.md`, `OPENSEARCH_MIGRATION_TESTER_GUIDE.md`, `OPENSEARCH_MIGRATION_TEST_PLAN.md`, `OPENSEARCH_CLIENT_CONFIGURATION.md`), but nothing that answers the question a Support or Cloud engineer actually has: **"I have been handed a customer instance. How do I migrate it, start to finish, without breaking them?"**

The tester guide is written for someone exercising the migration in a lab. The architecture doc is written for someone changing the code. Neither is a procedure. Today that gap is filled by tribal knowledge, and the same operational traps keep costing us round-trips:

- Enabling Phase 1 without a restart, so the OpenSearch indices are never created and every dual-write is silently swallowed.
- Assuming a phase change backfills OpenSearch (it never does — only a full reindex does).
- A missing `cluster:monitor/main` permission on the OpenSearch role, which dotCMS cannot distinguish from an unreachable cluster, so the migration silently halts back to Phase 0 (spike #35922).
- An OpenSearch role whose index pattern does not match the cluster ID, producing a reindex that returns HTTP 200 and creates nothing (#36222).
- Running a Site Search crawl before the content reindex in Phases 2/3, producing a permanently truncated index that reindexing afterwards does not repair (#36360).
- Not knowing that `/api/v1/index/migration/readiness` exists, or that it needs both CMS Admin and the migration support role.
- OSGi plugins and VTL templates still bound to Elasticsearch SDK types, which fail at Phase 2 and not before (#37870, #37559, #36026).

We need one document that a Support or Cloud engineer can follow from zero, without reading Java.

Separately, `OPENSEARCH_MIGRATION_TESTER_GUIDE.md` still describes `OS_MIGRATION_INDEX_VISIBILITY_ROLE_KEY` as the role that reveals `.os` indices in the index portlet. That stopped being true in #36360: `MigrationIndexVisibility` is now purely phase-based (hidden in Phases 0/1/2 for everyone, shown in Phase 3) and the role key only gates the readiness endpoint. The stale text should be corrected in the same change so the doc set does not contradict itself.

### Acceptance Criteria

- [ ] A new `docs/backend/OPENSEARCH_MIGRATION_RUNBOOK.md` written for a Support / Cloud audience, assuming no knowledge of the internals
- [ ] Explains the migration concepts and vocabulary, and **why** four phases exist rather than a single cutover
- [ ] Explains the mirror strategy for the active index: what dual-write does, what it does not do (no retroactive backfill), and every source of drift
- [ ] Explains how one dotCMS supports two engines: the single routing layer, and the replacement of Elasticsearch SDK types with dotCMS domain objects
- [ ] Covers the impact surface: VTL viewTools, Lucene queries and the search REST endpoints, Site Search, and OSGi plugins — including how to audit a customer's bundles and what to do when a plugin bug is found
- [ ] Covers OpenSearch server prerequisites and security criteria: version, endpoint separation, TLS, auth, and the least-privilege service account with its exact permission set
- [ ] Covers what happens when dotCMS cannot reach OpenSearch, per phase (safe degradation in 1/2, fail-loud in 3), with the log lines to recognise
- [ ] Covers migration strategy and customer selection criteria: number and size of indices, contentlet volume, measured reindex duration, plugin exposure
- [ ] Documents rehearsing the migration on a clone of the customer instance before touching the real environment
- [ ] Documents the fast path — going straight to Phase 3 for a small customer whose reindex fits in a window — with its cost stated explicitly
- [ ] States clearly when a node restart is required after changing the migration phase, and when it is not
- [ ] Covers clustered instances: node phase consistency, rolling restarts, and traffic direction during the change
- [ ] Covers downgrading to the previous phase and what to watch out for, including coming back from Phase 3
- [ ] Documents `GET /api/v1/index/migration/readiness` as the source of truth, the role that protects it, and how to read the report
- [ ] `OPENSEARCH_MIGRATION_TESTER_GUIDE.md` corrected: the visibility role no longer gates the index portlet (#36360)
- [ ] The four existing OpenSearch docs cross-link to the runbook

### Priority

Medium

### Additional Context

Docs-only change. No code, no schema, no API surface — nothing to test in CI and nothing for QA to exercise.

The runbook is derived from the existing architecture doc plus the operational findings accumulated across #35922, #36222, #36360, #36471, #36983, #37870 and the QA epic #35476. It is intended to become the entry point of the OpenSearch doc set: the architecture doc explains *why*, the runbook explains *how*, the tester guide explains *how to exercise it in a lab*.

fabrizzio-dotCMS:
Closes #37102

### Proposed Changes

**New — `docs/backend/OPENSEARCH_MIGRATION_RUNBOOK.md`.** A step-by-step migration guide: a Support or Cloud engineer is handed a customer instance and this walks them from "I have not touched anything" to "the customer is on OpenSearch". It assumes no Java and no knowledge of the internals.

It is a *guide*, not a test plan — verification steps exist only to confirm a step landed before you move to the next one. QA coverage stays where it lives, in `OPENSEARCH_MIGRATION_TESTER_GUIDE.md` and `OPENSEARCH_MIGRATION_TEST_PLAN.md`.

**Structure — the procedure leads, the reference follows.**

* **Before you begin.** What you are about to do (one diagram), the **three rules** that prevent almost every incident (a phase change never copies data; going up needs a restart, coming down does not; the readiness endpoint is the only truth), the **standard-vs-fast-path decision**, and the access checklist.
* **Part 1 — the procedure.** Six stages, each with a goal, an elapsed estimate, the risk to the customer, numbered steps, and a **gate** you must tick before continuing.
  * **Stage 0 — Assess the customer.** Measure the corpus, grep the VTL for `esSearch`/`esRaw`, grade every OSGi bundle Level 1/2/3 with the `unzip`/`grep` commands, inventory Site Search, choose the path.
  * **Stage 1 — Prepare the OpenSearch server.** Version and separation, TLS, the least-privilege service account with its exact permission set (including the `cluster:monitor/main` gap from spike #35922), the three permission checks that catch the index-pattern-vs-cluster-id trap (#36222), and the `OS_*` settings.
  * **Stage 2 — Rehearse on a clone.** A dry run of the migration on a copy of the customer, to get the measured reindex duration and to find out whether *their* templates and plugins survive Phase 2.
  * **Stage 3 — Turn on dual-write (0 → 1).** Readiness role, baseline, set the phase, rolling restart, confirm the startup log, confirm the `.os` rows exist, confirm one write reaches both engines **by document**, run the full reindex, crawl Site Search, confirm with the readiness report, soak.
  * **Stage 4 — Move reads to OpenSearch (1 → 2).** `safeToAdvance`, phase, rolling restart, **walk the customer's site** (the moment the Stage 0 risks materialise, with a symptom→cause table), watch for read fallbacks, soak.
  * **Stage 5 — Cut over (2 → 3).** Final readiness check, OpenSearch availability posture, phase, fail-loud rolling restart, confirm, cooling-off then decommission.
  * **Fast path.** Straight to Phase 3 for a small customer, as its own sequence, with the degraded-search window and the two-window rollback stated up front.
* **Part 2 — when something goes wrong.** Go back from where you are (downgrade table by cost, the 3 → 2 → reindex → 1 sequence, build rollback during dual-write, phase flip mid-reindex #36471, the pre-migration backup index trap), emergency stop, and a symptom table that links straight to the step that fixes it.
* **Part 3 — reference (R1–R15).** Concepts, why four phases, the mirror strategy and drift, the neutral domain objects that replaced the Elasticsearch SDK types, VTL/viewTools, Lucene and the search endpoints, OSGi in detail, OpenSearch security in detail, the readiness report field by field, the phase setting and restart rule, clustering, Site Search rules, configuration, log lines, commands. The steps link into it; you do not have to read it.

**Fixed — `OPENSEARCH_MIGRATION_TESTER_GUIDE.md` was stale.** It still described `OS_MIGRATION_INDEX_VISIBILITY_ROLE_KEY` as the role that reveals `.os` indices in the index portlet and in `/api/v1/esindex`. That stopped being true in #36360: `MigrationIndexVisibility` is now purely phase-based — hidden in Phases 0/1/2 for everyone, shown in Phase 3, consulting no user or role — and the role key gates only the readiness endpoint. Corrected in the config table, the §6 callout, the "which phase am I in" bullets, and the FAQ, pointing readers at the readiness endpoint instead.

**Cross-links.** `OPENSEARCH_MIGRATION.md` and the tester guide now point to the guide.

### Checklist
- [x] Tests — n/a, docs only
- [x] Translations — n/a
- [x] Security Implications Contemplated — the guide documents the OpenSearch least-privilege service account and the readiness endpoint's role gate; it introduces no code and no new surface. Credentials shown are the published local-lab values already in `docker/docker-compose-examples/single-node-os-migration`; no customer or production data appears anywhere in it.

### Additional Info

Every claim was verified against the code on this branch rather than carried over from the older docs — the readiness model and its derived `driftPercent` / `esIndexedPercent` / `osIndexedPercent` fields, the startup validator's halt-vs-fail-loud split by phase, `MigrationPhase` and the system-table precedence over the environment variable, `MigrationIndexVisibility` being phase-only, the OpenSearch role's permission set in `opensearch.py`, the `/v1/esindex` and `/v1/system-table` endpoint shapes, and the `ESContentTool` method surface.

All 92 internal anchors and all sibling-document links resolve.

Docs-only: no code, no schema, no API contract change. Nothing for CI to test and nothing for QA to exercise.

This PR fixes: #37102


docs/backend/OPENSEARCH_MIGRATION_RUNBOOK.md
```diff
@@ -0,0 +1,2360 @@
+# ES → OpenSearch Migration Guide
+
+> **Who this is for.** A Support or Cloud engineer who has been handed a customer instance and has to
```
nollymar: It can be used by customers executing the migration by themselves

docs/backend/OPENSEARCH_MIGRATION_RUNBOOK.md
```diff
@@ -0,0 +1,2360 @@
...
+
+Because dotCMS's existing Elasticsearch client **cannot write content to an OpenSearch 3.x
+cluster**. Repointing `ES_ENDPOINTS` at OpenSearch does not work — the connection may look fine while
+content indexing fails. Talking to OpenSearch 3.x needs a different client, which dotCMS ships
+alongside the old one.
+
+**So dotCMS temporarily runs two clients at once**, one per engine, and the migration is the
+controlled handover between them:
+
+- The two engines start out completely unrelated: Elasticsearch has all the customer's content,
```
nollymar: Saying Opensearch has nothing can be confusing for someone trying to upgrade from Opensearch 1 to 3

docs/backend/OPENSEARCH_MIGRATION_RUNBOOK.md
```diff
@@ -0,0 +1,2360 @@
...
+Elasticsearch is dropped. Unlike a template, you cannot fix it yourself. Background:
+[R7](#r7-osgi-plugins-in-detail).
+
+**Do this** for each of the customer's bundle jars:
+
+```bash
+# Does the bundle import Elasticsearch packages? (the worst case)
+unzip -p <bundle>.jar META-INF/MANIFEST.MF | tr ',' '\n' | grep -i elasticsearch
+
+# Does the compiled code reference ES classes or the deprecated search methods?
```
nollymar, line 479: @yolabingo @Gerald-OG, should we execute this ourselves or ask the customers to do it themselves?

docs/backend/OPENSEARCH_MIGRATION_RUNBOOK.md
fabrizzio-dotCMS: Good point — done in dfd8f070b5. The "Who this is for" block now covers both a Support/Cloud engineer handed a customer instance and a customer migrating their own installation, plus one line saying that where the guide writes "the customer", that is you if you are running it yourself. Kept the "customer" wording throughout rather than rewriting 2,300 lines — it is the party whose search must not break either way.

docs/backend/OPENSEARCH_MIGRATION_RUNBOOK.md
fabrizzio-dotCMS: Agreed, that line was written as if every install starts on Elasticsearch. Fixed in dfd8f070b5:

- §1 now opens with a source/target table making both starting points explicit — Elasticsearch **or** OpenSearch 1.x is the *source engine*, always configured through `ES_*` (historical naming, not a statement about the product), and the target is always a new OpenSearch 3.x cluster. Same procedure either way.
- The confusing bullet is reworded: the source engine has the content, the **new 3.x cluster is empty** — even when the source is itself an OpenSearch, the target is a separate cluster with nothing in it.
- Phase 0 ("Only Elasticsearch exists") and Step 1.1 got the same treatment.
- Step 1.1 also picks up the trap only a 1.x customer hits: do not upgrade the existing cluster in place and point both `ES_*` and `OS_*` at it. The fallback path at every phase depends on the source engine staying intact in its own cluster; the 1.x one gets decommissioned at the end of Stage 5 like any other source engine.

fabrizzio-dotCMS:
Applied both findings from the automated review, in 9bdba31778.

**🟠 Precedence was inverted — confirmed, and the consequence is worse than a wording bug.**

Verified against the code: `MigrationPhase.current()` (`IndexConfigHelper.java:144`) reads through `Config.getIntProperty`, and that method short-circuits on `environmentSetKeys` (`Config.java:634`) — the `DOT_`-prefixed variables `Config` recorded from the environment at startup — returning without ever calling `getSystemTableValue`. Real order:

```
DOT_FEATURE_FLAG_OPEN_SEARCH_PHASE  >  system table  >  properties file / in-memory writes
```

The reason this had to be fixed rather than reworded: on a normal containerized node the env var *is* set, so `POST /api/v1/system-table` returns 200 and changes nothing — and the `DELETE` the emergency-stop procedure tells you to run changes nothing either. The runbook was pointing an engineer under pressure at a lever that is not connected.

Corrected in all five places, each with the operational consequence rather than just the ordering:
- **Start here / "Where the switch is"** — explicit precedence ladder plus the trap.
- **Step 3.3** — Option B is now conditional: if any node sets the env var, use Option A.
- **Emergency stop step 3** — states that the DELETE is inert on env-var nodes, that the downgrade has to be made where the node actually reads from, and that Phase 2 still has the read fallback while Phase 3 does not.
- **Troubleshooting row** — "something higher in the chain wins", with the chain.
- **R10** — precedence column, the mechanism, and the two incident consequences.

The adjacent sub-claim the review flagged as correct — that the in-memory Phase-0 reset is a no-op when a system-table entry exists — is kept as-is; that one was right.

**🟡 `§16` → `R9`** — fixed in the tester guide. Correct: the runbook has no §16, the readiness report is R9.

Also folded in while in that file: "twin" for the counterpart index replaced with the terminology the runbook and the rest of the docs already use.

Re-ran the anchor sweep across both documents after the edits — all internal anchors and cross-references resolve.

yolabingo (approved):
👍

docs/backend/OPENSEARCH_MIGRATION_RUNBOOK.md
yolabingo: Since the first pass is simply doing a grep on the jar files I think we should give it a pass. Not 100% reliable but minimal effort and certainly better than nothing.

Gerald-OG (approved):

PR #37104, opened 2026-08-18
fix(block-editor): keep embedded contentlet selectable when text precedes it

### LINKED ISSUE #36985: Block Editor: embedded contentlet cannot be selected when text or other content precedes it
## Problem

In the **new Block Editor** (Evergreen), an embedded contentlet card (`dotContent` node) **cannot be selected with the mouse when any text or other element precedes it** in the document. Clicking the card produces no NodeSelection — no selection ring, and the contentlet can't be acted on (delete / reorder / edit via toolbar).

If the Block Editor field contains **only** contentlet cards (no text/media before them), selection appears to work.

Confirmed as a bug by engineering (Freddy Montes) in [#feat-block-editor-upgrade](https://dotcms.slack.com/archives/C08MDK1BYLF/p1786386929146079): "the selection of the block contentlet is broken". A screen recording is attached to that Slack thread.

## Steps to reproduce
https://dotcms.slack.com/archives/C08MDK1BYLF/p1786386929146079

1. Create (or edit) a contentlet with a Block Editor field — e.g. a `Direct - Block` content type.
2. Type some text (or add an image) first in the Block Editor.
3. Add one or more contentlets below it (slash menu → contentlet).
4. Save & Publish, close the editor.
5. Reopen the contentlet and click the embedded contentlet card to select it.

**Expected:** the card gets a NodeSelection (selection ring, toolbar contentlet actions become available; card can be deleted/reordered).
**Actual:** the card cannot be selected. The caret stays in the surrounding text.

**Note:** the issue does not occur when the document contains only contentlet cards — likely because ProseMirror's default selection (`Selection.near` on focus) lands on the first atom node when no text position exists, which masks the broken click-to-select path.

## Impact

- Global: affects every site/content type using the Block Editor with mixed text + contentlet content.
- Editors cannot delete, reorder, or open embedded contentlets once text precedes them.
- The contentlet toolbar entry point is gated on `NodeSelection` of `dotContent` (`core-web/libs/new-block-editor/src/lib/editor/components/toolbar/editor-toolbar.store.ts` ~L98), so it is also unreachable.

## Investigation notes (support)

- Node definition: `core-web/libs/new-block-editor/src/lib/editor/extensions/nodes/contentlet/contentlet.extension.ts` — `dotContent` is `atom: true, draggable: true`, selectable (default), Angular node view via ngx-tiptap `AngularNodeViewRenderer`.
- Node view: `contentlet.component.ts` is passive (no click handlers); selection styling comes from ngx-tiptap's `selected` prop.
- No in-repo click/mousedown interception was found on the card path (checked `editor-chrome-click.ts`, `block-gutter.extension.ts`, `selection-preserve.extension.ts`, editorProps in `editor.component.ts`).
- Third-party layers checked: tiptap core 3.22.2 default `NodeView.stopEvent` passes `mousedown` through for selectable nodes; `@tiptap/extension-drag-handle` 3.22.2 only hooks `keydown/mouseleave/mousemove`; ngx-tiptap 14.0.1 defines no `stopEvent`. Root cause is likely in the interaction between the Angular node view DOM and ProseMirror's click→NodeSelection path — needs debugging in a running editor.
- Possibly related (same node, different defect): #36976 (dragging an embedded contentlet duplicates it).

## Environment

- dotCMS: Evergreen (current release), reproduced on customer staging during Evergreen update testing
- Reporter environment: Worldline, PROD 2 staging (`docs.direct.worldline-solutions.com`, "test" page in root folder)
- Reproducible: always (with text/element before the contentlet)

## Workaround

Reverting the environment to the original Block Editor (`DOT_FEATURE_FLAG_NEW_BLOCK_EDITOR=false`) restores previous behavior — same rollback used for other new-Block-Editor regressions (e.g. tickets 38284/38290).

## Related Freshdesk ticket

https://dotcms.freshdesk.com/a/tickets/38740
https://dotcms.slack.com/archives/C08MDK1BYLF/p1786386929146079

rjvelazco:
## Proposed Changes

Fixes #36985 — in the new Block Editor, an embedded contentlet card (`dotContent`) could not be selected with the mouse when any text or other block preceded it. Without a `NodeSelection` the card showed no selection ring and could not be deleted, reordered, or edited (the toolbar's contentlet actions are gated on a `dotContent` NodeSelection).

### Root cause (two parts)

1. **No click-to-select path on the atom's Angular node view.** The `dotContent` node view (`contentlet.component.ts`) was passive, so selection depended entirely on ProseMirror's default click handling.
2. **The selection was immediately clobbered.** The emitted value carries root document-stat attrs (`charCount` / `wordCount` / `readingTime`) via `withDocStats`, but the change guard (`editorContentMatchesParsed`) compared the round-tripped value against `editor.getJSON()`, which has **no** root attrs. So once there was any text the guard was permanently `false`, firing a `setContent` on every value round-trip that rebuilt the document and reset the selection.

Diagnosed with a live ProseMirror `dispatch` trace: click created the NodeSelection, then a full-document (byte-identical) `setContent` transaction reset it.

### Changes

- Add a `mousedown` handler on the contentlet node view that selects the node by its own `getPos()` — mirrors the existing `dotImage` atom pattern.
- Extract `stripDocStats` to `utils/doc-stats.utils.ts` and compare on the stats-free shape on **both** sides of `editorContentMatchesParsed`, so the guard is symmetric with what is emitted. This stops the spurious `setContent` (and the per-keystroke doc churn it caused).
- Unit tests for `stripDocStats`.

## Checklist

- [x] Tests: `pnpm nx test new-block-editor` — 62/62 pass (incl. new `doc-stats.utils.spec.ts`)
- [x] Lint + Prettier clean
- [ ] Manually verified in the Block Editor (new): click-to-select an embedded contentlet with text before it; delete/reorder/edit reachable; typing and two-way `[value]`/`valueChange` unaffected

## Video

https://github.com/user-attachments/assets/89d84647-4b18-4ff3-84c6-141e70e756a8

## Notes

Separate, pre-existing issue spotted while debugging (not in scope here): picking slash-menu "Text" can leave a literal `/` paragraph behind. Will file separately.

This PR fixes: #36985

### DIFF
--- a/core-web/libs/new-block-editor/src/lib/editor/extensions/nodes/contentlet/contentlet.component.ts
+++ b/core-web/libs/new-block-editor/src/lib/editor/extensions/nodes/contentlet/contentlet.component.ts
@@ -120,6 +120,9 @@ export class DotContentletNodeViewComponent extends AngularNodeViewComponent {
      */
     protected selectNode(event: MouseEvent): void {
         const pos = this.getPos()();
+        // Nullish-only guard on purpose: `getPos()` returns `number | undefined`, and a contentlet
+        // that is the first block resolves to position 0. `!pos` would treat that 0 as "no position"
+        // and skip selecting the first-block card — the exact only-contentlets case from #36985.
         if (pos == null) {
             return;
         }
--- a/core-web/libs/new-block-editor/src/lib/editor/extensions/nodes/contentlet/contentlet.component.ts
+++ b/core-web/libs/new-block-editor/src/lib/editor/extensions/nodes/contentlet/contentlet.component.ts
@@ -19,7 +19,8 @@ import { EditorStore } from '../../../store/editor.store';
         '[attr.data-identifier]': 'identifierAttr()',
         '[attr.data-language-id]': 'languageIdAttr()',
         '[attr.data-inode]': 'inodeAttr()',
-        '[attr.data-content-type]': 'contentTypeAttr()'
+        '[attr.data-content-type]': 'contentTypeAttr()',
+        '(mousedown)': 'selectNode($event)'
     },
     template: `
         @if (data(); as d) {
@@ -103,4 +104,27 @@ export class DotContentletNodeViewComponent extends AngularNodeViewComponent {
         const ct = this.data()?.contentType;
         return ct ? String(ct) : null;
     });
+
+    /**
+     * Select the embedded contentlet card on mousedown.
+     *
+     * The card is an `atom` rendered through this Angular node view. When text or another block
+     * precedes it, the browser's default mousedown drops a text caret into that preceding block
+     * before the `click` fires — so ProseMirror resolves the click to a text position, never to
+     * the atom, and the card gets no NodeSelection (no selection ring; the toolbar contentlet
+     * actions, gated on a `dotContent` NodeSelection, stay unreachable) — #36985.
+     *
+     * Selecting the node by its own position (`getPos()`, always accurate regardless of what
+     * surrounds it) and suppressing the default mousedown fixes click-to-select in every layout.
+     * Reordering is unaffected: it runs off the block-gutter drag handle, a separate element.
+     */
+    protected selectNode(event: MouseEvent): void {
+        const pos = this.getPos()();
+        if (pos == null) {
+            return;
+        }
+
+        event.preventDefault();
+        this.editor().chain().focus().setNodeSelection(pos).run();
+    }
 }
--- a/core-web/libs/new-block-editor/src/lib/editor/editor.component.ts
+++ b/core-web/libs/new-block-editor/src/lib/editor/editor.component.ts
@@ -52,12 +52,13 @@ import { DotUploadService } from './services/dot-upload.service';
 import { EditorModalService } from './services/editor-modal.service';
 import { EditorPopoverService } from './services/editor-popover.service';
 import { EditorStore } from './store/editor.store';
+import { stripDocStats } from './utils/doc-stats.utils';
 import { loadRemoteExtensions, parseCustomBlocksField } from './utils/remote-extensions.loader';
 import { preserveUnknownBlockNodes, restoreUnknownBlockNodes } from './utils/unknown-block.utils';
 
 /** Stringifies the editor document for form output (plain ProseMirror JSON, no extra attrs). */
 function editorDocumentJsonText(editor: Editor): string {
-    return JSON.stringify(editor.getJSON());
+    return JSON.stringify(stripDocStats(editor.getJSON()));
 }
 
 /**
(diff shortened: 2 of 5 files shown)


core-web/libs/new-block-editor/src/lib/editor/extensions/nodes/contentlet/contentlet.component.ts
```diff
@@ -103,4 +104,27 @@ export class DotContentletNodeViewComponent extends AngularNodeViewComponent {
...
+     * the atom, and the card gets no NodeSelection (no selection ring; the toolbar contentlet
+     * actions, gated on a `dotContent` NodeSelection, stay unreachable) — #36985.
+     *
+     * Selecting the node by its own position (`getPos()`, always accurate regardless of what
+     * surrounds it) and suppressing the default mousedown fixes click-to-select in every layout.
+     * Reordering is unaffected: it runs off the block-gutter drag handle, a separate element.
+     */
+    protected selectNode(event: MouseEvent): void {
+        const pos = this.getPos()();
+        if (pos == null) {
```
KevinDavilaDotCMS, line 126: if (!post) {
  return;
}

KevinDavilaDotCMS (approved):

rjvelazco:
Thanks for the review. Addressed the one actionable item:

- 🟡 **Missing test for click-to-select** → added `contentlet.component.spec.ts` (`de72507c95`): a Spectator test on `DotContentletNodeViewComponent` asserting a host `mousedown` calls `editor.chain().focus().setNodeSelection(getPos())` and `preventDefault`s, plus a no-op case when `getPos()` is unresolved. Full lib suite now 64/64.

The two "verify" notes need no change:
- `getPos()()` double-call is the ngx-tiptap 14 signal-wrapped node-view prop convention (same as `node()`/`editor()`/`selected()` in this file).
- `mousedown` → `preventDefault()` for any button matches the `dotImage` precedent (`image.extension.ts:185`).

core-web/libs/new-block-editor/src/lib/editor/extensions/nodes/contentlet/contentlet.component.ts
rjvelazco: Good catch to scrutinize, but I kept it as `pos == null` on purpose. `getPos()` returns `number | undefined`, and a contentlet that's the **first block** resolves to position `0`. `if (!pos)` would treat that `0` as falsy and bail — skipping selection for a first-block card, which is exactly the only-contentlets case #36985 is about.

Added an inline comment explaining this (df0371fa91) so it's clear for next time. Let me know if you'd prefer `pos === undefined` for precision.

rjvelazco: It's a position and it can be 0. I can not use `if (!post)` here

PR #37112, opened 2026-08-19
36991 support showlinks in the content drive search api apiv1drivesearch

### LINKED ISSUE #36991: Support showLinks in the Content Drive search API (/api/v1/drive/search)
## Description

`POST /api/v1/drive/search` cannot return menu Links, so any client built on it can only select folders and contentlets. Menu Links are reachable today only through the legacy `POST /api/v1/browser`.

This surfaced while migrating custom-field VTLs to the new editor. `redirect_custom_field_new.vtl` picks a redirect target through `DotCustomFieldApi.openBrowserModal()`:

```js
params: {
  showLinks: true,
  showPages: true,
  showDotAssets: false,
  showFiles: false,
  showFolders: false,
  showWorking: false,
  showArchived: false,
  sortByDesc: true
}
```

That call resolves to the legacy browser (`DotBrowserSelectorComponent` → `/api/v1/browser`). Every one of those flags maps onto the drive search **except `showLinks`**:

| Legacy `ContentByFolderParams` | Drive search equivalent | Status |
|---|---|---|
| `showPages: true` | `baseTypes: ["HTMLPAGE"]` | ✅ exists |
| `showFiles` / `showDotAssets` | `baseTypes: ["FILEASSET"\|"DOTASSET"]` | ✅ exists |
| `showFolders: false` | `showFolders: false` | ✅ exists |
| `showArchived: false` | `archived: false` | ✅ exists |
| `showWorking: false` | `live: true` | ✅ exists (`BrowserAPIImpl:857` — `final boolean live = !browserQuery.showWorking;`) |
| `sortByDesc: true` | `sortBy: "modDate:desc"` | ✅ exists |
| `mimeTypes` | `mimeTypes` | ✅ exists |
| **`showLinks: true`** | — | ❌ **missing** |

Until Links are supported, that custom field (and any other client needing link selection) cannot move off the legacy browser.

## Why this is smaller than it looks

The drive search is **not** a separate Elasticsearch query. `ContentDriveHelper` builds a `BrowserQuery` and calls the same `BrowserAPI` the legacy browser uses — Elasticsearch is only an optional *filtering* strategy, enabled when a text filter is present.

| Location | Today |
|---|---|
| `dotCMS/src/main/java/com/dotcms/rest/api/v1/drive/ContentDriveHelper.java:147` | pins `.showLinks(false)` |
| `dotCMS/src/main/java/com/dotcms/rest/api/v1/drive/AbstractDriveRequestForm.java` | has no `showLinks` field |
| `dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java` `getPaginatedContents` (~L1600) | slices **folders** and **contentlets** only — no link branch |

`BrowserAPI` already knows how to fetch links — `BrowserAPIImpl:1536`, `:1557` and `:2523` (`getLinks(...)`), used by `getFolderContent` / `getFolderContentList`. What is missing is a link branch in the **paginated** path, plus the request/response fields to drive it.

## Open design question (backend team's call)

`getPaginatedContents` currently pages each source independently: folders consume `maxResults` first from `folderCursor`, then contentlets consume the remainder from `contentCursor`, and the response reports `folderCount` / `contentCount` / `hasMoreFolders` / `hasMoreContent` / `nextFolderCursor` / `nextContentCursor`.

Links need to fit into that. Options seen from the outside, deliberately not prescribed here:

1. **A third symmetric source** — `linkCursor` in the request; `linkCount` / `hasMoreLinks` / `nextLinkCursor` in the response, mirroring the folder slice. Most predictable for clients, most new fields.
2. **Folded into the folder slice** — links and folders share a cursor (both are DB-backed and hang off the same parent folder). Fewer new fields, but two different entities behind one counter.
3. Something else the team prefers.

Whatever is chosen, clients need a deterministic way to page links to exhaustion without duplicates or gaps.

Two details worth deciding alongside it:

- **`baseTypes` interaction** — Links are not a `BaseContentType` (`BaseContentType.java` has ANY, CONTENT, WIDGET, FORM, FILEASSET, HTMLPAGE, PERSONA, VANITY_URL, KEY_VALUE, DOTASSET — no LINK). `ContentDriveHelper` derives `showContent(!baseContentTypes.isEmpty())`, so a "links only" request (`showLinks: true` with no `baseTypes`) needs a defined outcome.
- **Sorting** — `getPaginatedContents` sorts the merged list with `GenericMapFieldComparator(sortBy, sortByDesc)`. Link maps need to expose the sortable keys (`title`, `modDate`, `modUser`) or ordering will be inconsistent between links and the rest.

## Acceptance Criteria

- [ ] `POST /api/v1/drive/search` accepts a `showLinks` flag on the request form
- [ ] `showLinks` defaults to `false`, so every existing caller is unaffected (Content Drive, AssetPicker, and any other current consumer behave exactly as they do today)
- [ ] With `showLinks: true`, menu Links under the resolved asset path are returned alongside the other results
- [ ] Returned links are filtered by the requesting user's READ permission
- [ ] Links honour `live` and `archived` consistently with how the legacy `/api/v1/browser` treats them
- [ ] Links participate in `maxResults` and in pagination: paging through a folder that contains links returns every link exactly once, with no duplicates and no gaps
- [ ] The response carries whatever metadata a client needs to page links to exhaustion (exact shape decided by the chosen design above), and that shape is documented
- [ ] A request with `showLinks: true` and no `baseTypes` has a defined, documented outcome
- [ ] Result ordering honours `sortBy` consistently across links, folders and contentlets
- [ ] `openapi.yaml` is regenerated from the annotations and committed alongside the Java changes
- [ ] Integration test coverage for: links returned when enabled, absent when disabled/defaulted, permission filtering, and paging across a folder with more links than `maxResults`
- [ ] Postman coverage for the new flag on the drive search collection

## Priority

Medium

## Additional Context

**Not a blocker for the AssetPicker epic #36702** — nothing there selects Links; the picker deals with dotAsset / File Asset. This is filed separately so the AssetPicker work can ship without waiting on backend.

What it *does* block is migrating custom fields that select Links off the legacy browser, e.g. `dotCMS/src/main/webapp/WEB-INF/velocity/static/htmlpage_assets/redirect_custom_field_new.vtl` (see `.claude/skills/dot-ui-vtl-migration/references/migration-guide.md`, Rule 11 — "File and Page Browser Dialog").

Frontend consumption (exposing `showLinks` through `DotAssetPickerConfig` / `DotCustomFieldApi.openBrowserModal`) is **out of scope here** and will follow once the API contract lands.

Relevant files:
- `dotCMS/src/main/java/com/dotcms/rest/api/v1/drive/ContentDriveHelper.java`
- `dotCMS/src/main/java/com/dotcms/rest/api/v1/drive/AbstractDriveRequestForm.java`
- `dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java`
- `dotCMS/src/main/java/com/dotcms/browser/BrowserQuery.java`

nicobytes:
This pull request enhances the pagination and retrieval of menu links in the Content Drive browser API. It introduces independent, cursor-based paging for menu links (in addition to folders and contentlets), ensuring more consistent and scalable navigation when browsing assets under a folder. The changes also unify how links are mapped and returned, improving both API clarity and client-side handling.

**Pagination and API enhancements:**

* Added independent cursor-based pagination for menu links, alongside folders and contentlets, in the `getPaginatedContents` method of `BrowserAPIImpl`, including new fields (`linkCursor`, `linkCount`, `hasMoreLinks`, `nextLinkCursor`) in both the request (`BrowserQuery`) and response (`PaginatedContents`). [[1]](diffhunk://#diff-bf8151158d087844f299503dc61ea3da47dc5449ca564ea49e1d9847cc0a2ef3R1625-R1631) [[2]](diffhunk://#diff-ba08d9df93b41994d8ccf0eaf5c2dfa837a33c28463c5c08da0c8814c4ff5d82R48) [[3]](diffhunk://#diff-ba08d9df93b41994d8ccf0eaf5c2dfa837a33c28463c5c08da0c8814c4ff5d82R141) [[4]](diffhunk://#diff-ba08d9df93b41994d8ccf0eaf5c2dfa837a33c28463c5c08da0c8814c4ff5d82R265) [[5]](diffhunk://#diff-ba08d9df93b41994d8ccf0eaf5c2dfa837a33c28463c5c08da0c8814c4ff5d82R304) [[6]](diffhunk://#diff-bf8151158d087844f299503dc61ea3da47dc5449ca564ea49e1d9847cc0a2ef3L1676-R1722) [[7]](diffhunk://#diff-bf8151158d087844f299503dc61ea3da47dc5449ca564ea49e1d9847cc0a2ef3R1738-R1759) [[8]](diffhunk://#diff-bf8151158d087844f299503dc61ea3da47dc5449ca564ea49e1d9847cc0a2ef3L1664-R1695)

* Modified the documentation and method contracts in both `BrowserAPI` and `BrowserAPIImpl` to describe the new paging behavior for links, clarifying the contract for clients and the order in which page slots are filled (folders, then links, then contentlets). [[1]](diffhunk://#diff-52da51c2f9c9dbcb8d66ce3fdeec9c0322780de46025c176123d76feae6a363eL253-R278) [[2]](diffhunk://#diff-bf8151158d087844f299503dc61ea3da47dc5449ca564ea49e1d9847cc0a2ef3L1676-R1722)

**Link retrieval and mapping improvements:**

* Implemented the `linksDefaultView` method to retrieve, filter, and deterministically order menu links directly under a parent, supporting stable paging and matching legacy endpoint behavior.

* Standardized link mapping in API responses by using a shared constant for the MIME type and providing additional metadata (permissions, owner name, etc.) in the returned map, aligning with folder conventions. [[1]](diffhunk://#diff-bf8151158d087844f299503dc61ea3da47dc5449ca564ea49e1d9847cc0a2ef3R114-R115) [[2]](diffhunk://#diff-bf8151158d087844f299503dc61ea3da47dc5449ca564ea49e1d9847cc0a2ef3L2522-R2571)

**Supporting changes:**

* Added necessary imports for new functionality, including `Stream` and `UserLocalManagerUtil`. [[1]](diffhunk://#diff-bf8151158d087844f299503dc61ea3da47dc5449ca564ea49e1d9847cc0a2ef3R26) [[2]](diffhunk://#diff-bf8151158d087844f299503dc61ea3da47dc5449ca564ea49e1d9847cc0a2ef3R63) [[3]](diffhunk://#diff-bf8151158d087844f299503dc61ea3da47dc5449ca564ea49e1d9847cc0a2ef3R95)

These updates collectively provide a more robust, flexible, and user-friendly API for browsing assets, especially when dealing with large numbers of menu links.

#36991 

Fixes #36991


nicobytes:
## How to use `showLinks` — API contract and examples

Two new fields on the request, three on the response. Everything else is unchanged.

### Request

| Field | Type | Default | Purpose |
|---|---|---|---|
| `showLinks` | boolean | `false` | Include the menu Links directly under `assetPath` |
| `linkCursor` | int | `0` | Index to start paging links from |

The `false` default is what keeps every current consumer (Content Drive, AssetPicker) behaving exactly as it does today.

### Response

```json
{
  "entity": {
    "list": [ /* folders, links and contentlets merged, ordered by sortBy */ ],
    "folderCount": 2,  "hasMoreFolders": false, "nextFolderCursor": 2,
    "linkCount": 3,    "hasMoreLinks": true,    "nextLinkCursor": 3,
    "contentCount": 0, "hasMoreContent": true,  "nextContentCursor": 0
  },
  "errors": [], "messages": [], "i18nMessagesMap": {}
}
```

Links are a **third symmetric pagination source**: their own cursor, count and `hasMore` flag, mirroring the folder slice.

---

### 1. Basic — links alongside everything else

```bash
curl -u admin@dotcms.com:admin \
  -X POST 'http://localhost:8080/api/v1/drive/search' \
  -H 'Content-Type: application/json' \
  -d '{
    "assetPath": "//demo.dotcms.com/about-us/",
    "showLinks": true,
    "maxResults": 20
  }'
```

### 2. The `redirect_custom_field_new.vtl` case — links + pages

This is the direct translation of the legacy browser flags that motivated the issue:

```bash
curl -u admin@dotcms.com:admin \
  -X POST 'http://localhost:8080/api/v1/drive/search' \
  -H 'Content-Type: application/json' \
  -d '{
    "assetPath": "//demo.dotcms.com/",
    "showLinks": true,
    "baseTypes": ["HTMLPAGE"],
    "showFolders": false,
    "live": true,
    "archived": false,
    "sortBy": "modDate:desc",
    "maxResults": 20
  }'
```

### 3. Links only

Links are **not** a `BaseContentType`, so `showLinks` is orthogonal to `baseTypes`. An empty `baseTypes` array disables the content query:

```bash
curl -u admin@dotcms.com:admin \
  -X POST 'http://localhost:8080/api/v1/drive/search' \
  -H 'Content-Type: application/json' \
  -d '{
    "assetPath": "//demo.dotcms.com/about-us/",
    "showLinks": true,
    "showFolders": false,
    "baseTypes": [],
    "maxResults": 20
  }'
```

The three documented outcomes:

| Request | Result |
|---|---|
| `showLinks: true`, `baseTypes` omitted | links **plus** content of every base type |
| `showLinks: true`, `baseTypes: ["HTMLPAGE"]` | links plus pages |
| `showLinks: true`, `baseTypes: []`, `showFolders: false` | **links only** |

### 4. Paging links to exhaustion

Feed `nextLinkCursor` back as `linkCursor` and keep `offset` at `0`:

```bash
# page 1
curl -u admin@dotcms.com:admin -X POST 'http://localhost:8080/api/v1/drive/search' \
  -H 'Content-Type: application/json' \
  -d '{"assetPath":"//demo.dotcms.com/about-us/","showLinks":true,
       "showFolders":false,"baseTypes":[],"maxResults":2,"linkCursor":0}'
# -> linkCount: 2, hasMoreLinks: true, nextLinkCursor: 2

# page 2
curl ... -d '{... "maxResults":2, "linkCursor":2}'
```

Stop when `hasMoreLinks: false`; from then on send `showLinks: false` to skip the query entirely.

Budget order is **folders -> links -> contentlets**. If folders fill `maxResults`, you get `linkCount: 0` with `hasMoreLinks: true` and an unadvanced `nextLinkCursor` — that is the signal that links remain, not a bug.

---

### Link shape

Identify links by `mimeType === "application/dotlink"` (or `type === "links"`).

```json
{
  "identifier": "9f2c...",
  "type": "links",
  "title": "Contact Us Redirect",
  "name": "Contact Us Redirect",
  "mimeType": "application/dotlink",
  "extension": "link",
  "__icon__": "linkIcon",
  "url": "www.google.com",
  "protocol": "https://",
  "target": "_blank",
  "linkType": "EXTERNAL",
  "permissions": ["READ", "WRITE"],
  "modDate": "2026-08-19T17:41:28.000Z",
  "owner": "Admin User"
}
```

Note two Content Drive conventions, consistent with the folders this endpoint already returns: **no `inode`**, and `permissions` as role-type **names** rather than raw integer ids. The legacy `includeLinks()` used by `/api/v1/browser` is untouched.

### Three filters suppress links

Sending `showLinks: true` together with any of these yields `linkCount: 0`, because a Link cannot satisfy them:

- **`mimeTypes`** — a Link has no file MIME type
- **`workflow`** — a Link carries no workflow state (folders are already dropped here for the same reason)
- **`userSearchable`** — resolves against a single content type; a Link has no fields

### Two inherited limits

Both match legacy `/api/v1/browser` behaviour, which is what the AC asked for:

- Links are **direct children only** of the resolved `assetPath` — no recursion into subfolders, even at site root (where contentlets *are* gathered recursively).
- `filters.text` matches link **titles only**, applied in memory, because links are not indexed in Elasticsearch.

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
```diff
@@ -1641,6 +1649,28 @@ public PaginatedContents getPaginatedContents(final BrowserQuery browserQuery)
...
 
+        // Menu Links — cursor-based: slice starting from linkCursor, mirroring the folder
+        // slice. Links consume whatever budget folders left over, before contentlets.
+        // When hasMoreLinks=false is returned, the caller should set showLinks=false on the
+        // next request to skip this query entirely.
+        if (browserQuery.showLinks) {
+            final List<Map<String, Object>> allLinks = linksDefaultView(browserQuery, roles);
+            final int totalLinks = allLinks.size();
+            final int linkStart = Math.min(browserQuery.linkCursor, totalLinks);
+
```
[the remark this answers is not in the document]
nicobytes: Good catch, and the diagnosis is exactly right — with `linkCursor: 999` against 5 links, `nextLinkCursor` comes back as `999` rather than the documented "index of the first link not yet returned".

I'm going to leave the behaviour as-is deliberately, for one reason: **the pre-existing folder branch does exactly the same thing.** `nextFolderCursor` is initialised from the request and only reassigned inside `if (folderStart < totalFolders)` (`BrowserAPIImpl.java:1636-1650`), so an out-of-range `folderCursor` is echoed back verbatim too. That code isn't in this diff, which is why only the link copy got flagged.

Clamping `nextLinkCursor` alone would make links behave differently from folders, which cuts against the design premise of this PR — links as a *symmetric* third source with the same paging semantics. Given the practical impact is nil (`hasMoreLinks: false` already tells the client to stop, and a client following the documented loop never sends an out-of-range cursor), asymmetry seems like the worse trade.

Instead I've tightened the contract wording so the guarantee matches reality for all three cursors: `next*Cursor` is meaningful while the matching `hasMore*` is `true`. If we do want clamping, it should be a follow-up that clamps `nextFolderCursor` and `nextLinkCursor` together, with a test for both.

dotCMS/src/main/java/com/dotcms/rest/api/v1/drive/ContentDriveHelper.java
```diff
@@ -144,7 +149,7 @@ public PaginatedContents driveSearch(final DriveRequestForm requestForm, final U
             .showArchived(showArchived)
             .showWorking(!live)
             .showFolders(showFolders)
-            .showLinks(false)
+            .showLinks(showLinks)
```
oidacra, line 152: This line makes `live:true` reachable for links in the drive path for the first time. Tracing that combination: `.showWorking(!live)` sends `working=false` into `getLinks`, and because `ChildrenCondition.working` is a `Boolean`, `FolderFactoryImpl` (around line 1328) emits the predicate whenever it is non-null rather than only when true:

```java
sql += versionTable + ".working_inode" + (cond.working ? "=" : "<>") + tableName + "_1_.inode and ";
```

so the fragment becomes `link_version_info.working_inode <> links_1_.inode`. As far as I can tell that predicate never joins the version table on the link's identifier, which would make it a cross join — a published link would come back once per other non-deleted version row, old versions would surface alongside it, and with `setMaxRows(1000)` the duplicates could push distinct links out of the result entirely.

If that reading is right it would conflict with AC6's requirement that paging returns every link exactly once with no duplicates or gaps — though only for `live:true`, and the legacy `/api/v1/browser` path looks equally affected, so this seems newly exposed rather than introduced here.

Is `live:true` intended to be a supported combination for links? If so, would passing `working=true` and filtering on `hasLiveVersion` be preferable to deriving the flag from `!live`? If not, is it worth rejecting or documenting the combination so callers do not hit it silently? Either way, would an integration case for `showLinks:true, live:true` be worth adding — `testLinksHonourLiveAndArchivedLikeLegacyBrowser` compares against the legacy path, so if both share this behavior the assertion would pass regardless.

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
```diff
@@ -1641,6 +1649,28 @@ public PaginatedContents getPaginatedContents(final BrowserQuery browserQuery)
...
+            if (linkStart < totalLinks) {
+                // When folders exhausted the budget maxResults is exactly 0, so the slice is
+                // empty, the cursor stays put and hasMoreLinks correctly reports the remainder.
+                final int linkEnd = Math.min(linkStart + maxResults, totalLinks);
+                list.addAll(allLinks.subList(linkStart, linkEnd));
```
oidacra, line 1745: Question about hostile input on the new cursor. `linkCursor` does not appear to be clamped anywhere along the path — `AbstractDriveRequestForm.linkCursor()` has `@Value.Default` but no `@Value.Check`, `BrowserQuery.Builder.linkCursor(int)` stores it as-is, and `ContentDriveResource` has no `@Valid`/`@Min` on the body. Since the guard above is `if (linkStart < totalLinks)`, a negative value survives `Math.min(-1, totalLinks)` and enters the branch, so `subList(-1, linkEnd)` would throw `IndexOutOfBoundsException` — an uncaught 500 rather than a 400. The catches in `linksDefaultView` only wrap the `getLinks` call, so they would not cover it.

This is not new with links: `folderCursor` follows the same `Math.min` + `subList` shape on main, and `contentCursor` flows unclamped into the DB offset. `maxResults` is guarded in the same builder (`maxResults < 0 ? this.maxResults : maxResults`), so the asymmetry looks unintentional.

Since cursors are designed to be echoed back verbatim from a prior response, a client replaying a corrupted value seems plausible. Would you rather clamp defensively here (`Math.max(0, browserQuery.linkCursor)`), or add a `@Value.Check` covering all three cursors so it surfaces as a 400? The latter fixes the pre-existing cases too, though it widens the PR.

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
```diff
@@ -2593,6 +2648,134 @@ private List<Map<String, Object>> foldersDefaultView(final BrowserQuery browserQ
...
+        // Links are not indexed in Elasticsearch, so the text filter has to be applied here or a
+        // narrowed search would return every link under the parent.
+        final Stream<Link> filtered = UtilMethods.isSet(browserQuery.filter)
+                ? links.stream().filter(link -> null != link.getTitle() && link.getTitle()
+                        .toLowerCase().contains(browserQuery.filter.toLowerCase()))
+                : links.stream();
+
+        // A stable order is what makes the index-based linkCursor sound. Titles are not unique
+        // among sibling links, hence the identifier tiebreaker.
+        final List<Link> ordered = filtered
```
oidacra, line 2801: Question about how this interacts with `sortBy`. This comparator is fixed to title-ascending, and since the page slice is taken from `ordered` before `getPaginatedContents` applies `GenericMapFieldComparator`, the requested sort reorders the items already selected rather than influencing which ones are selected.

The case that made me look: `sortBy:"title:desc"` with `maxResults:2` over four links titled A, B, C, D. `linkStart=0`, `linkEnd=2`, so the slice is {A,B} and the final sort returns {B,A}, where I would have expected {D,C}. Here the sort key is the same key used for slicing and only the direction differs, so it is not a case of `sortBy` diverging from the slicing key.

This looks structural rather than link-specific — `getFolders` hard-codes `Folder::getName` ascending the same way, and contentlets order by `mod_date` regardless of `sortBy` — so links are arguably consistent with what folders and contentlets already do, which may satisfy AC9's "consistently across" wording.

Do you read AC9 as already satisfied by that consistency, or would you want the pre-slice order derived from `sortBy`/`sortByDesc` (keeping title+identifier as the stability tiebreaker)? If the current behavior is intended, would it be worth noting in the `showLinks()` javadoc that links page in name order independent of `sortBy`, since the `PaginatedContents` docs describe a merged sorted list?

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/drive/ContentDriveLinksTest.java
```diff
@@ -0,0 +1,589 @@
...
+        final PaginatedContents results = search(baseRequest()
+                .showLinks(true)
+                .showFolders(false)
+                .baseTypes(List.of())
+                .sortBy(sortBy)
+                .build());
+        return linksIn(results).stream().map(item -> (String) item.get("title"))
+                .collect(Collectors.toList());
+    }
+}
```
oidacra, line 793: Was an `@AfterClass` teardown intentionally left out here? Between `prepare()` and `testLinksFilteredByReadPermission` the class persists two Sites, three Folders, eight Links, a ContentType, a Role and a User, and I do not see anything deleting or archiving them. Since the class is registered in `MainSuite3a`, those fixtures would outlive it for the rest of the suite.

The sibling `ContentDriveKeywordSearchTest` in the same package keeps a field specifically for this, with the comment "Kept for @AfterClass cleanup so the suite does not accumulate sites/types/temp dirs", and `docs/testing/INTEGRATION_TESTS.md` asks for cleanup in reverse order of creation.

Would it be worth mirroring that `cleanup()` here — deleting `testType`, archiving/deleting `testSite` — and hoisting the permission test's site, folder, role and user into static fields (or deleting them in a `finally`) so they are torn down too?

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
```diff
@@ -2593,6 +2648,134 @@ private List<Map<String, Object>> foldersDefaultView(final BrowserQuery browserQ
...
+                    "User '%s' cannot read the parent of the requested links: %s",
+                    browserQuery.user.getUserId(), e.getMessage()));
+            return List.of();
+        } catch (final DotDataException e) {
+            Logger.error(this, "Could not load links : ", e);
+            return List.of();
+        }
+
+        // Links are not indexed in Elasticsearch, so the text filter has to be applied here or a
+        // narrowed search would return every link under the parent.
```
oidacra: This in-memory title filter looks like the one piece of genuinely new logic in the PR without any test behind it, and the comment above it explains exactly why it matters — links are not in Elasticsearch, so if this branch stopped working a narrowed search would quietly return every link under the parent rather than erroring.

As far as I can tell nothing exercises it: `ContentDriveLinksTest` never sets `filters(...)` on a request, and none of the six new Postman requests combines `filters.text` with `showLinks`. The path is reachable from the form via `ContentDriveHelper`'s `withFilter(requestForm.filters().text())`.

Would it be worth adding a case that searches with `filters.text` and `showLinks:true` and asserts only the matching link comes back, plus a differently-cased term to cover the `toLowerCase()` branch and a non-matching term asserting zero links? `ContentDriveKeywordSearchTest` looks like a close model for the content-side equivalent.

One related thought while looking at this: the link filter applies unconditionally, whereas the folder equivalent in `getFolders` is gated behind `filterFolderNames` (mapped from `filters.filterFolders`). Is that asymmetry intended? If so it may be worth a line in the `showLinks()` javadoc, since a caller setting `filterFolders:false` might not expect link titles to still be narrowed.

dotCMS/src/main/java/com/dotcms/rest/api/v1/drive/ContentDriveHelper.java
nicobytes: Your reading was right, and it turned out to be worse than the version you described. I reproduced it, so concretely:

`live:true, archived:false` against five links under a folder returned **35 rows** — each link seven times. A second run returned 20. The multiplier is the installation's link-version count, so it varies per database.

Three things on top of what you wrote:

1. **The `deleted` predicate is uncorrelated in that branch too.** `link_version_info.deleted='false'` with nothing joining the version row degenerates to "there exists any non-archived link anywhere", so the archived filter is a complete no-op whenever `working=false`. It only works in the `=` branch, where `working_inode = links_1_.inode` incidentally pins the row.
2. **`setMaxRows(1000)` runs with no `ORDER BY`.** `orderBy` is `null` on every links path, and the limit is applied client-side in `DotConnect` over an unordered result set. So once an installation's version-row count approaches 1000, the window can be consumed by duplicates of a single link and other links in the folder become invisible — differently on each call.
3. **`testLinksHonourLiveAndArchivedLikeLegacyBrowser` could not have caught it**, for two independent reasons: it collapsed the ids into a `HashSet`, which erases duplicates by construction, and it asserted parity against the legacy path, which shares the defect. Both halves had to change.

To your questions:

**Is `live:true` intended for links?** Yes — it is the whole point of the issue. `redirect_custom_field_new.vtl` sends `showWorking: false`, which `ContentDriveHelper` maps to exactly this combination. So the feature was broken in its motivating use case.

**Pass `working=true` and filter on `hasLiveVersion` instead of deriving the flag from `!live`?** That is what I did — your suggestion, essentially verbatim. `BrowserAPIImpl.getLinks` now always asks `FolderAPI` for the working links and resolves "live" by keeping the ones carrying a published version.

The blast radius is narrow and worth stating: `BrowserQuery` ORs `showArchived` into `showWorking`, so `live:true, archived:true` already passed `working=true`, as did the default `live:false`. **Only `live:true, archived:false` changes behaviour.** But `getLinks(BrowserQuery)` is shared with `includeLinks`, so legacy `/api/v1/browser` with `showWorking:false` is fixed by the same change — from garbage to correct. Flagging that explicitly since it is outside the stated scope of this PR.

**An integration case for `showLinks:true, live:true`?** Two, and I deliberately did not rely on legacy parity for either:

- `testLiveOnlyReturnsPublishedLinks` — `prepare()` now publishes two of the five links, so `live:true` has a genuinely smaller set to narrow to. Pins the *meaning* of the flag rather than agreement with another code path. Would have passed against a live filter that silently did nothing; does not pass against this bug.
- `testLinksHonourLiveAndArchivedLikeLegacyBrowser` — compares `List`s now, with an explicit no-duplicates assertion, before the set comparison against legacy.

Both confirmed failing on the pre-fix code (`expected:<35> but was:<5>`).

**The factory itself is untouched**, so any other caller passing `working=false` / `live=false` still hits the cross join. Filed as #37133 with the SQL, the reproduction, and the one-line fix — it needs its own PR because `getChildrenClass` is shared by `Link` and `Contentlet`, and correcting `working=false` also changes its semantics ("versions that are not the working one", which is not "the live version").

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
nicobytes: Confirmed, and reproduced as an actual `IndexOutOfBoundsException: fromIndex = -1` out of `ArrayList.subList` — a 500, exactly as you traced. You were also right that the `catch` blocks in `linksDefaultView` do not cover it: they wrap only the `getLinks` call, not the slice.

I went with the defensive clamp rather than `@Value.Check`, for two reasons:

- It fixes the pre-existing cases in the same stroke. `folderCursor` has the identical `Math.min` + `subList` shape and `offset` flows unclamped into the DB, so all four (`contentCursor`, `folderCursor`, `linkCursor`, `offset`) are now `Math.max(0, …)` in `BrowserQuery.Builder`. A form-level check would only cover the drive path and would leave the legacy browser as it is.
- It matches the idiom already in that builder — `maxResults` does `maxResults < 0 ? this.maxResults : maxResults` two methods down, which is what made the asymmetry look unintentional to you, and I think you were right about that.

`testNegativeLinkCursorIsTreatedAsZero` covers it; confirmed throwing before the change.

If we do want a 400 for garbage input rather than a silent clamp, that is a reasonable follow-up and should cover all three cursors plus `offset` in one go — happy to file it. But I would keep the clamp underneath either way, since `BrowserQuery` is reachable from more than the drive form.

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
nicobytes: Your own reading is the one I landed on: AC9 is satisfied by the consistency, and links are not the odd one out here. `getFolders` hardcodes `Folder::getName` ascending and contentlets order by `mod_date`, both regardless of `sortBy` — the pre-slice order is what makes each source's index-based cursor stable, and `sortBy` reorders the page that order selected.

Your `title:desc` example is the sharpest way to state it, so I documented it rather than left it implicit. Two places:

- `showLinks()` — links page in title-ascending order (identifier as tiebreaker) independent of `sortBy`; `sortBy` reorders the items already selected, not which items get selected. With the note that this matches the other two sources.
- `PaginatedContents` — each source slices in its *own* internal order, and `sortBy` applies to the merged list of that page only. You were right that the existing docs describing "a merged sorted list" invited the opposite reading.

I did not make links derive their pre-slice order from `sortBy`: that would make links *better* than folders and contentlets, which is asymmetry in the other direction, and it would need field mapping for `modDate`/`modUser` on the Link map. If we want it, it should land for all three sources at once.

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/drive/ContentDriveLinksTest.java
nicobytes: Not intentional — an oversight. Fixed, mirroring `ContentDriveKeywordSearchTest.cleanup()` as you suggested: one `try/catch` per resource, null-guarded, `Logger.warn` and swallow, in reverse order of creation.

And you were right that hoisting was the part that actually mattered. `testLinksFilteredByReadPermission` built its own site, folder, three links, role and user as method locals, so a class-level teardown would have missed all of them. The site, role and user are now static fields (with a comment saying why), so the folder and links go with the site and the role and user are deleted explicitly.

`@AfterClass` now tears down: the content type, then both sites (archive + delete), then the user, then the role.

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
nicobytes: Right on both counts — the filter had no coverage anywhere, and the asymmetry is real.

**Coverage.** Three cases added, modelled on `ContentDriveKeywordSearchTest`:

- `testFilterTextNarrowsLinksByTitle` — a term matching exactly one of the five links asserts only that link comes back and `linkCount == 1`. It runs the term twice, lowercase and upper-cased, which is the `toLowerCase()` branch you asked about.
- `testFilterTextWithNoMatchReturnsNoLinks` — a non-matching term asserts zero links, which is the failure mode you described: if this branch stopped working it would not error, it would quietly return every link under the parent.

Postman got two requests as well, but honestly weaker ones: menu links cannot be created over REST, so that collection has no link fixtures and can only pin that the combination is accepted, that the pagination contract survives it, and that any link returned does match the term. Said so in the request descriptions so nobody mistakes them for the real assertions.

**The `filterFolders` asymmetry.** Intended, and now documented in `showLinks()`. The reasoning: `filterFolders` exists so a client can keep folders *navigable* while narrowing the results — a folder is something you descend into, so hiding it costs you the path. A link is a selectable leaf, so there is nothing to preserve by keeping it visible, and narrowing it is what the user asked for. Also noted there is no `filterLinks` equivalent, since your point is that a caller setting `filterFolders:false` would not otherwise expect link titles to still be narrowed. Left a comment at the filter itself too, so the next reader does not "fix" the asymmetry by accident.

dotCMS/src/main/java/com/dotcms/rest/api/v1/drive/AbstractDriveRequestForm.java
```diff
@@ -360,6 +360,38 @@ public interface AbstractDriveRequestForm {
...
+     *   <li>{@code showLinks: true, baseTypes: ["HTMLPAGE"]} returns links plus pages.</li>
+     *   <li>{@code showLinks: true, baseTypes: [], showFolders: false} returns links
+     *       <i>only</i> — an empty {@code baseTypes} array disables the content query.</li>
+     * </ul>
+     * <p>
+     * Links are ignored when {@code mimeTypes} or {@code workflow} filters are present, since a
+     * Link carries neither a file MIME type nor workflow state and so could never satisfy them.
+     * Links are also always the <i>direct</i> children of the resolved path — they are never
+     * gathered recursively across subfolders, matching the legacy {@code /api/v1/browser}
+     * endpoint.
```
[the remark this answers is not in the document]
nicobytes: Fixed in 243523a29b, which landed before this thread was revisited — the Javadoc now lists `userSearchable` alongside `mimeTypes` and `workflow`. That commit also builds the `BrowserQuery` once and logs the query itself rather than the pre-override locals, which was the related point in the `ContentDriveHelper` thread.

nicobytes:
## Review feedback addressed — `725c224d14`

Thanks @oidacra, this was a good catch list. One of the five turned out to be a genuine blocker, so the headline first.

### `live:true` + `showLinks:true` was returning duplicates

Thread [#discussion_r3815730314](https://github.com/dotCMS/core/pull/37112#discussion_r3815730314) was right, and worse than stated. `FolderFactoryImpl.getChildrenClass` appends the version table to the `FROM` list with no join predicate of its own; the only correlation it ever produces is the incidental one from `working_inode = links_1_.inode`. Passing `showWorking` straight through meant `working=false` flipped that to `<>`, which correlates nothing and degenerates into a cross product.

Reproduced: **five links came back 35 times** for `live:true, archived:false`. A second run returned 20 — the multiplier is the installation's link-version count.

That is the one combination the motivating consumer actually sends: `redirect_custom_field_new.vtl` passes `showWorking: false`, which the drive maps to `live: true`. The feature was broken in the use case that prompted #36991.

`BrowserAPIImpl.getLinks` now always asks `FolderAPI` for the working links and resolves "live" by keeping the ones carrying a published version. `BrowserQuery` ORs `showArchived` into `showWorking`, so **only `live:true, archived:false` changes behaviour** — the default and `archived:true` already passed `working=true`. Since `getLinks(BrowserQuery)` is shared with `includeLinks`, legacy `/api/v1/browser` with `showWorking:false` is fixed too; calling that out explicitly as it is outside this PR's stated scope.

The factory itself is untouched, so any other caller passing `working=false` still hits the cross join. Filed as **#37133** with the SQL, the reproduction and the one-line fix — it needs its own PR because `getChildrenClass` is shared by `Link` and `Contentlet`, and correcting `working=false` also changes its semantics.

### Everything else

| Thread | Outcome |
|---|---|
| [Negative cursor → 500](https://github.com/dotCMS/core/pull/37112#discussion_r3815730948) | Clamped `contentCursor`, `folderCursor`, `linkCursor` and `offset` at zero in `BrowserQuery.Builder`, matching the clamp `maxResults` already does there. Fixes the pre-existing folder/offset cases too. |
| [`sortBy` vs pre-slice order](https://github.com/dotCMS/core/pull/37112#discussion_r3815731492) | Documented, per your own reading — folders and contentlets slice the same way, so AC9 holds by consistency. Noted in both `showLinks()` and `PaginatedContents`. |
| [Missing `@AfterClass`](https://github.com/dotCMS/core/pull/37112#discussion_r3815732837) | Added, mirroring `ContentDriveKeywordSearchTest.cleanup()`. The permission test's site, role and user are hoisted into statics so they are torn down too. |
| [Untested title filter](https://github.com/dotCMS/core/pull/37112#discussion_r3815733501) | Three cases added, including the upper-cased term for the `toLowerCase()` branch. The `filterFolders` asymmetry is intentional and now documented. |
| [Copilot: `userSearchable` in Javadoc](https://github.com/dotCMS/core/pull/37112#discussion_r3815436167) | Already fixed in `243523a29b`. |

### Test evidence

`ContentDriveLinksTest` is 14 → 18 tests, all green. The four new ones were confirmed **failing against the pre-fix code**:

```
testLiveOnlyReturnsPublishedLinks              AssertionError: expected:<35> but was:<5>
testLinksHonourLiveAndArchivedLikeLegacyBrowser AssertionError: expected:<35> but was:<5>
testNegativeLinkCursorIsTreatedAsZero          IndexOutOfBoundsException: fromIndex = -1
```

Worth noting why the existing live/archived test could not have caught this: it collapsed ids into a `HashSet`, which erases duplicates by construction, **and** asserted parity against the legacy path, which shares the defect. Both halves had to change — it now compares lists with an explicit no-duplicates assertion.

Postman got two more requests (`filters.text` and `live:true`), but weaker ones by necessity: menu links cannot be created over REST, so that collection has no link fixtures and can only pin that the combinations are accepted and the contract survives. Said so in the request descriptions.

### On AC10 (`openapi.yaml`)

Nothing to commit, and that is correct rather than an omission: `ContentDriveResource.search` is `@Hidden`, so `/v1/drive` has no generated schema at all. The single `showLinks` in the yaml belongs to `BrowserQueryForm` (the legacy browser), untouched here. Verified with `./mvnw compile -pl :dotcms-core` — the file comes back unchanged.

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
```diff
@@ -1641,6 +1649,28 @@ public PaginatedContents getPaginatedContents(final BrowserQuery browserQuery)
...
+
+            if (linkStart < totalLinks) {
+                // When folders exhausted the budget maxResults is exactly 0, so the slice is
+                // empty, the cursor stays put and hasMoreLinks correctly reports the remainder.
+                final int linkEnd = Math.min(linkStart + maxResults, totalLinks);
+                list.addAll(allLinks.subList(linkStart, linkEnd));
+                linkCount = linkEnd - linkStart;
+                maxResults -= linkCount;
+                nextLinkCursor = linkEnd;
+                hasMoreLinks = linkEnd < totalLinks;
```
freddyDOTCMS, line 1749: I think you should used PaginationUtil for pagination here

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
```diff
@@ -1641,6 +1649,28 @@ public PaginatedContents getPaginatedContents(final BrowserQuery browserQuery)
...
+            }
+            // else: linkCursor is past the end — all links already shown, add nothing
+        }
```
freddyDOTCMS, line 1752: can we do this pagination in the database query?

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
```diff
@@ -2536,20 +2597,48 @@ private List<Map<String, Object>> includeLinks(final BrowserQuery browserQuery)
...
+        } else {
+            return Collections.emptyList();
         }
-        return Collections.emptyList();
+
+        if (browserQuery.showWorking) {
+            return links;
+        }
+
+        return links.stream()
```
jcastro-dotcms: The `folderAPI.getLinks()` method should be in charge of this. Such an API should have a method that handles this scenario.

nicobytes: Agreed — fixed in 9a32593948.

You were right that this belonged in `FolderAPI`. Better still, the method you describe **already existed**: `FolderAPI.getLiveLinks(Folder, User, boolean)` (`FolderAPIImpl:1313`). And its SQL is already correct — `cond.live = true` emits `live_inode = links_1_.inode`, the correlating `=` branch, so it never hits the cross join that `working=false` produces. The only thing missing was the `Host` overload for the site-root case.

So:

- Added `FolderAPI`/`FolderAPIImpl.getLiveLinks(Host, User, boolean)`, mirroring the existing `getLinks(Host, working, deleted, …)` with `cond.live=true; cond.deleted=false`.
- `BrowserAPIImpl.getLinks` now just picks between `getLinks` and `getLiveLinks` per parent type. The `hasLiveVersion` stream is gone, and so is the javadoc paragraph that existed only to explain the workaround.

Net effect on the PR: **-28 lines in `BrowserAPIImpl`**, and the browser layer no longer knows anything about version-table join semantics.

`getLiveLinks` pinning `deleted=false` turned out to be exactly what this path needs, not a limitation: `BrowserQuery` does `showWorking = showWorking || showArchived`, so the live branch is only ever reached with `archived=false`. "Live and archived at once" is never expressible from the request.

**One behavioural difference worth declaring rather than burying:** `getLiveLinks` returns the **live version rows**, whereas the code it replaces returned the **working rows of links that have a live version**. For `live:true` I think the new behaviour is the more correct one — the caller asked for published content, not for the draft of something published. The tests assert identifiers, which are stable across versions, so they did not need changing.

Which is the part I'd point at as evidence: `ContentDriveLinksTest` is **untouched and still 18/18 green**. I deliberately did not adjust a single assertion, so the suite passing is a statement that the refactor preserves observable behaviour rather than a statement that I re-fitted the tests to the new code.

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
nicobytes: I dug into this and I don't think `PaginationUtil` fits here — but it's a fair thing to ask, so here's what I found rather than just a no.

`com.dotcms.util.PaginationUtil` is REST-layer infrastructure, and the codebase is unanimous about it: of the **25 `new PaginationUtil(...)` sites in `dotCMS/src/main/java`, all 25 are under `com.dotcms.rest.*`**. Not one business-layer `*APIImpl` uses it. That isn't an accident of style:

- It imports `javax.ws.rs.core.Response` and `javax.servlet.http.HttpServletRequest/Response`, and builds the JAX-RS `Response` itself, setting `X-Pagination-Per-Page`, `X-Pagination-Current-Page`, `X-Pagination-Total-Entries` and an RFC-5988 `Link` header.
- All five `getPage(...)` overloads are `@Deprecated`. The replacement, `getPageView(PaginationUtilParams)`, unconditionally dereferences `utilParams.request().getRequestURI()` and `utilParams.response().setHeader(...)` — passing null NPEs.

`BrowserAPIImpl.getPaginatedContents(BrowserQuery)` has no request, no response and no request URI. It returns the `PaginatedContents` POJO.

Beyond the layer, the model doesn't line up either:

- **One source, not three.** `Paginator<T>.getItems(user, limit, offset, params)` returns one `PaginatedArrayList<T>` with one `totalResults`. This block runs three heterogeneous sources against one decrementing `maxResults` budget and returns nine pagination scalars (three counts, three `hasMore`, three cursors). There's no seam in the `Paginator` contract for a second source, and nowhere in `PaginatedArrayList` to carry a `hasMore`.
- **Page numbers, not cursors.** `getMinIndex(currentPage, perPage) = (currentPage - 1) * perPage` — the offset is *derived* from a page number, so you can't hand it three independent cursors. Grepping the whole pagination package for cursor semantics returns nothing. Meanwhile this contract says "keep `offset` at 0 on every request; only the cursors change."
- **It requires a total we deliberately don't compute.** `lastPage = ceil(totalRecords / perPage)`. `PaginatedContents` has no `total` field by design — the `hasMore*` flags exist precisely to avoid counting. With `totalResults` unset, `PaginationUtil` would advertise `last` page 0 and never emit a `next` rel: actively wrong output.

`FolderSearchPaginator` is the correctly-scoped example in-repo — single source, single window, delegating to `FolderAPI.searchFolders`. That shape works; this one doesn't.

**The adjacent point I think you may actually be reaching for is real, though**, and I'd rather name it than let it hide behind my "no": `ContentDriveResource` returns `new ResponseEntityView<>(...)` rather than `ResponseEntityPaginatedDataView`, so drive responses carry no `pagination` envelope and no `X-Pagination-*` headers — the pagination state lives entirely in the entity body. That's a legitimate discussion about the *resource's* response shape. But it's a change to the `/api/v1/drive/search` contract, which is already on `main` with `folderCount`/`contentCount`/`hasMore*`/`next*Cursor` and has consumers — so not something a `showLinks` PR should do. And even there, `Pagination`'s `currentPage`/`perPage`/`totalEntries` triple still can't represent three cursors.

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
nicobytes: I looked into this properly, and the answer is no — but for a reason that surprised me and is worth writing down, because the premise doesn't hold.

**`DotConnect` does not paginate in the database at all.** In `dotCMS/src/main/java/com/dotmarketing/common/db/DotConnect.java`:

- `:698` → `// statement.setMaxRows(maxRows);` is **commented out**. On the `createStatement()` branch it's never called either. The JDBC driver receives no row hint whatsoever.
- `:790-792` → `for (int i = 0; i < startRow; i++) { rs.next(); }`. The offset is a **client-side skip** over the `ResultSet`, not SQL `OFFSET`.
- `:797` → `while (rs.next() && (maxRows <= 0 || i < maxRows))`. The limit is client-side too.

So `FolderFactoryImpl.getChildrenClass` already asks the DB for **every** link under the parent today; its `offset`/`limit` args only cap how many Java objects get built. Threading them through would satisfy the wording of the request while delivering nothing in the database. I didn't want to make that change and report back "done, it paginates in the DB now."

**And real SQL `LIMIT`/`OFFSET` still wouldn't be correct here**, because four filters shrink the set *after* the query: the READ `filterCollection` in `FolderAPIImpl`, the live filter, the in-memory title filter (links aren't in Elasticsearch), and the per-link `getPermissionIdsFromRoles` check. A DB page of 40 could return 3 rows with `hasMoreLinks: false` — silently lost links. The Java sort also runs after filtering, so a SQL `ORDER BY` would order a different set than the one the cursor indexes into, breaking the cursor invariant outright.

This question has been adjudicated twice in this codebase already, in writing:

- `FolderAPIImpl.searchFolders` (`:812`) — the newest paginated `FolderAPI` method, written for Content Drive: *"Load all matching folders (no LIMIT — pagination happens in Java so that permission filtering does not produce short pages for limited users)"*.
- `BrowserAPIImpl:184` — *"Pagination is applied **after** permission filtering. Applying DB-level pagination before filtering produces empty or sparse pages for restricted users."*

Permissions can't move into the `WHERE` clause either: `permission_reference` is a lazily and **asynchronously** populated cache (`PERMISSION_REFERENCES_UPDATE_ASYNC` defaults to `true`), so a freshly created link has no row until someone reads it. Joining it would omit *readable* links — a correctness bug, worse than a short page. That's also why `content.drive.folder.search.permissions.max.per.page` exists to **cap** page size with a 400, rather than to paginate by permission.

Contentlets do use real `LIMIT ? OFFSET ?`, but only inside the `getContentByChunks` over-fetch machinery (`BrowserAPIImpl:248-320`) that exists precisely because SQL paging plus post-hoc permission filtering is unsound on its own. Worth it for contentlets (unbounded cardinality); not for links, which are the direct children of one folder.

Lastly, folders do exactly the same in-memory slice today (`getFolders` + `allFolders.subList`, `:1636-1650`) — the link branch is a line-for-line mirror of it. Paginating only links in the DB would make the two asymmetric without buying correctness.

**Where I think your instinct is pointing at something real:** the convenience overloads of `getChildrenClass` hardcode a **1000-row ceiling on a result set with no `ORDER BY`**, so a folder with more than 1000 links is truncated non-deterministically. That's a genuine limitation of this path and I documented it in `linksDefaultView`'s javadoc rather than leave it implicit. Fixing it properly means a `searchLinks(...)`-style method modelled on `searchFolders` — permission-filter first, then slice, with a total order — not SQL pagination. Happy to file that if you think it's worth queueing.

nicobytes:
## Second review round — `9a32593948`

Three new threads, from @freddyDOTCMS and @jcastro-dotcms. All architectural rather than defects. I verified each one against the code before answering: **one was right and produced a code change; two don't hold, and I've put the evidence in the threads.** Net effect is that the PR gets smaller.

### @jcastro-dotcms — right, and the method already existed

The `hasLiveVersion` filter I'd put in `BrowserAPIImpl` was a workaround in the wrong layer. `FolderAPI.getLiveLinks(Folder, User, boolean)` already existed with the correct SQL (`cond.live=true` → `live_inode = links_1_.inode`, the correlating `=` branch — no cross join). The only gap was the `Host` overload.

- Added `FolderAPI`/`FolderAPIImpl.getLiveLinks(Host, User, boolean)`.
- `BrowserAPIImpl.getLinks` now picks between `getLinks` and `getLiveLinks`; the in-memory filter and the javadoc paragraph explaining the workaround are both deleted.

**`ContentDriveLinksTest` is untouched and still 18/18 green.** I didn't adjust a single assertion, so the suite passing is a statement that behaviour is preserved rather than that I re-fitted the tests. One behavioural difference declared in the thread: `getLiveLinks` returns the *live version rows*, where the old code returned the *working rows of links that have a live version*. For `live:true` the new one is the more correct answer.

### @freddyDOTCMS — `PaginationUtil`

Not applicable at this layer. Of the **25 `new PaginationUtil(...)` sites in `dotCMS/src/main/java`, all 25 are under `com.dotcms.rest.*`** — no business-layer `*APIImpl` uses it. It imports JAX-RS + Servlet types and writes `X-Pagination-*` headers; all five `getPage` overloads are `@Deprecated`, and the replacement NPEs without a request/response. Beyond the layer: `Paginator<T>` is one source with one `totalResults`, while this block runs three sources against one decrementing budget and returns nine pagination scalars; and it derives its offset from a page number, so it can't take three cursors.

I did flag the adjacent point that I think is genuinely worth discussing: `ContentDriveResource` returns `ResponseEntityView` rather than `ResponseEntityPaginatedDataView`, so drive responses carry no `pagination` envelope or `X-Pagination-*` headers. That's a real conversation about the resource's response shape — but it changes the `/api/v1/drive/search` contract that's already on `main` with consumers, so not for a `showLinks` PR.

### @freddyDOTCMS — pagination in the DB query

The premise doesn't hold: **`DotConnect` doesn't paginate in the database.** `statement.setMaxRows` is commented out (`DotConnect:698`), `startRow` is a client-side `rs.next()` skip (`:790`), and the limit is a client-side loop bound (`:797`). Passing `offset`/`limit` through `getChildrenClass` would satisfy the wording while changing nothing in the DB — I wasn't willing to make that change and report it as done.

Real SQL `LIMIT`/`OFFSET` wouldn't be correct either: four filters shrink the set after the query (READ `filterCollection`, live, the in-memory title filter since links aren't in ES, and the per-link `getPermissionIdsFromRoles`), so a page of 40 could return 3 with `hasMoreLinks: false`. And permissions can't move into the `WHERE` clause — `permission_reference` is a lazily, *asynchronously* populated cache, so a join there would omit readable links.

The codebase has already decided this twice in writing — `FolderAPIImpl.searchFolders:812` (*"no LIMIT — pagination happens in Java so that permission filtering does not produce short pages"*) and `BrowserAPIImpl:184`. Folders do the same in-memory slice today; the link branch mirrors them line for line.

**The real limitation inside that question**, which I'd rather name than let my "no" bury: the convenience overloads of `getChildrenClass` hardcode a **1000-row ceiling on an unordered result set**, so a folder with >1000 links truncates non-deterministically. It's documented in `linksDefaultView`'s javadoc. Fixing it means a `searchLinks(...)` method modelled on `searchFolders`, not SQL paging — happy to file that if it's worth queueing.

---

I resolved @jcastro-dotcms's thread since it's actioned, and **left both of @freddyDOTCMS's open on purpose** — they were questions, and a declined objection isn't mine to close. Push back if you disagree with either reading.

jcastro-dotcms (approved):

PR #37113, opened 2026-08-19
fix(ui): prevent ng-event listener accumulation on iframe reload

### LINKED ISSUE #34534: Allow comments on workflow action is throwing error while saving content
### Problem Statement

When a workflow action has **Allow comments** enabled and a user saves **new content**, the save fails with a unique field constraint error (e.g. on URLTitle). The same save works if Allow comments is not selected.

The failure happens because the save is triggered twice when the comment path is used (both the legacy and Angular flows call the same save endpoint), which causes a second insert and the unique constraint error.

**Impact:** Users cannot reliably save new content when using workflow actions that require a comment.

When we are selecting the allow comments option in any workflow action and then try to save a  new content we get an error on the screen related to unique field value constraint being violated. In this particular case the below error related to unique URLTitle field. It appears that insert is being performed twice as the content is already created and saved

```
[06/02/26 17:41:53:823 GMT] ERROR extratable.DBUniqueFieldValidationStrategy: The unique value 'syeddd' for the field 'urlTitle' in the Content Type 'Product' already exists
[06/02/26 17:41:53:823 GMT]  WARN business.ESContentletAPIImpl: Value of Field [urlTitle] must be unique. Contents having the same value (syeddd): Unknown/New Contentlet
[06/02/26 17:41:53:823 GMT]  WARN business.ESContentletAPIImpl: Contentlet with ID 'Unknown/New' ['Syeddd'] has invalid/missing field(s). The unique value 'syeddd' for the field 'urlTitle' in the Content Type 'Product' already exists - Fields: [UNIQUE]: URL Title (urlTitle)
[06/02/26 17:41:53:824 GMT]  WARN business.ESContentletAPIImpl: com.dotmarketing.portlets.contentlet.business.DotContentletValidationException$Builder.build(DotContentletValidationException.java:557)
[06/02/26 17:41:53:824 GMT] ERROR business.WorkflowAPIImpl: Failed to fire Workflow Action 'Save' [ceca71a0-deee-4999-bd47-b01baa1bcfc8]: Contentlet with ID 'Unknown/New' ['Syeddd'] has invalid/missing field(s). The unique value 'syeddd' for the field 'urlTitle' in the Content Type 'Product' already exists - Fields: [UNIQUE]: URL Title (urlTitle)
[06/02/26 17:41:53:824 GMT] ERROR db.HibernateUtil: ---------- DotHibernate: error on rollbackTransaction ---------------
```

### Steps to Reproduce

1. Create or edit a workflow action and enable **Allow comments**.
2. Create **new** content (e.g. with a unique URLTitle) and save using that workflow action.
3. Enter a comment in the dialog/wizard and submit.
4. Observe the unique field (e.g. URLTitle) constraint error.
 Reproducible when the DWR call `ContentletAjax.saveContent.dwr` is invoked twice (e.g. visible in browser Network tab).

With **Allow comments** off, the same content saves successfully.

https://drive.google.com/file/d/1ecfKMkBEupnpYe1ySl2zqt0sjGbjzIZ6/view?usp=sharing

### Acceptance Criteria

System should allow saving of content when allow comments is selected in the workflow action.

### dotCMS Version

Latest 26.01.26-1

### Severity

### Links

https://helpdesk.dotcms.com/a/tickets/35362

gortiz-dotcms:
## Summary

Fixes #34534 — saving new content via a workflow action with **Allow Comments** enabled fails with a unique constraint error on URLTitle (and `identifier_pkey` at the DB level).

- **Root cause:** `IframeComponent.handleIframeEvents()` called `.bind(this)` inline for both `addEventListener` and `removeEventListener`. Each `.bind()` call creates a new function object, so `removeEventListener` never matched the previously registered handler — silently accumulating an extra `ng-event` listener on the iframe document on every `onLoad` event. With two listeners, the `workflow-wizard` custom event reached Angular twice, triggering two concurrent `ContentletAjax.saveContent` DWR calls. Both calls saw an empty inode (new content) and raced to `INSERT` the same deterministic identifier, the second losing with a unique constraint violation.
- **Fix:** Bind the handler references once as class fields (`boundEmitKeyDown`, `boundEmitCustomEvent`) so the same stable object is passed to both `addEventListener` and `removeEventListener`, ensuring each `onLoad` correctly replaces the listener instead of stacking a new one.
- **Test updated:** The existing spec now asserts the same function reference is used for both add and remove, which is the invariant the fix enforces.

## Test plan

- [ ] `pnpm nx test dotcms-ui --testPathPattern=iframe.component.spec` passes
- [ ] Open the content editor for a **new** contentlet
- [ ] Use a workflow action with **Allow Comments** enabled, enter a comment, and submit
- [ ] Confirm only **one** `ContentletAjax.saveContent` call appears in the Network tab (previously two)
- [ ] Confirm the content saves without a unique constraint error

### DIFF
--- a/core-web/apps/dotcms-ui/src/app/view/components/_common/iframe/iframe-component/iframe.component.spec.ts
+++ b/core-web/apps/dotcms-ui/src/app/view/components/_common/iframe/iframe-component/iframe.component.spec.ts
@@ -219,7 +219,7 @@ describe('IframeComponent', () => {
             };
         });
 
-        it('should remove and add listener on load', () => {
+        it('should remove and add listener on load using the same stable function references', () => {
             iframeEl.triggerEventHandler('load', {
                 target: {
                     contentDocument: {
@@ -228,19 +228,26 @@ describe('IframeComponent', () => {
                 }
             });
 
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            const keyDownRef = (comp as any).boundEmitKeyDown;
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            const customEventRef = (comp as any).boundEmitCustomEvent;
+
+            // remove must receive the exact same reference as add so the browser
+            // can actually deregister the handler instead of silently failing
             expect(
                 comp.iframeElement.nativeElement.contentWindow.removeEventListener
-            ).toHaveBeenCalledWith('keydown', expect.any(Function));
+            ).toHaveBeenCalledWith('keydown', keyDownRef);
             expect(
                 comp.iframeElement.nativeElement.contentWindow.document.removeEventListener
-            ).toHaveBeenCalledWith('ng-event', expect.any(Function));
+            ).toHaveBeenCalledWith('ng-event', customEventRef);
 
             expect(
                 comp.iframeElement.nativeElement.contentWindow.addEventListener
-            ).toHaveBeenCalledWith('keydown', expect.any(Function));
+            ).toHaveBeenCalledWith('keydown', keyDownRef);
             expect(
                 comp.iframeElement.nativeElement.contentWindow.document.addEventListener
-            ).toHaveBeenCalledWith('ng-event', expect.any(Function));
+            ).toHaveBeenCalledWith('ng-event', customEventRef);
         });
 
         it('should set the colors to the jsp on load', () => {
--- a/core-web/apps/dotcms-ui/src/app/view/components/_common/iframe/iframe-component/iframe.component.ts
+++ b/core-web/apps/dotcms-ui/src/app/view/components/_common/iframe/iframe-component/iframe.component.ts
@@ -76,6 +76,13 @@ export class IframeComponent implements OnInit, OnDestroy {
 
     private destroy$: Subject<boolean> = new Subject<boolean>();
 
+    // Stable bound references required so removeEventListener can match the
+    // exact function object that was passed to addEventListener. Using
+    // .bind(this) inline creates a new object each call, making removal a
+    // no-op and causing listeners to accumulate on every iframe load event.
+    private readonly boundEmitKeyDown = this.emitKeyDown.bind(this);
+    private readonly boundEmitCustomEvent = this.emitCustonEvent.bind(this);
+
     ngOnInit(): void {
         this.iframeOverlayService.overlay
             .pipe(takeUntil(this.destroy$))
@@ -264,17 +271,11 @@ export class IframeComponent implements OnInit, OnDestroy {
     }
 
     private handleIframeEvents($event): void {
-        this.getIframeWindow().removeEventListener('keydown', this.emitKeyDown.bind(this));
-        this.getIframeWindow().document.removeEventListener(
-            'ng-event',
-            this.emitCustonEvent.bind(this)
-        );
+        this.getIframeWindow().removeEventListener('keydown', this.boundEmitKeyDown);
+        this.getIframeWindow().document.removeEventListener('ng-event', this.boundEmitCustomEvent);
 
-        this.getIframeWindow().addEventListener('keydown', this.emitKeyDown.bind(this));
-        this.getIframeWindow().document.addEventListener(
-            'ng-event',
-            this.emitCustonEvent.bind(this)
-        );
+        this.getIframeWindow().addEventListener('keydown', this.boundEmitKeyDown);
+        this.getIframeWindow().document.addEventListener('ng-event', this.boundEmitCustomEvent);
         this.charge.emit($event);
 
         const doc = this.getIframeDocument();


core-web/apps/dotcms-ui/src/app/view/components/_common/iframe/iframe-component/iframe.component.ts
```diff
@@ -264,17 +271,11 @@ export class IframeComponent implements OnInit, OnDestroy {
...
-        this.getIframeWindow().addEventListener('keydown', this.emitKeyDown.bind(this));
-        this.getIframeWindow().document.addEventListener(
-            'ng-event',
-            this.emitCustonEvent.bind(this)
-        );
+        this.getIframeWindow().removeEventListener('keydown', this.boundEmitKeyDown);
+        this.getIframeWindow().document.removeEventListener('ng-event', this.boundEmitCustomEvent);
+
+        this.getIframeWindow().addEventListener('keydown', this.boundEmitKeyDown);
+        this.getIframeWindow().document.addEventListener('ng-event', this.boundEmitCustomEvent);
```
oidacra, line 278: I traced the save path for this one and wanted to check the root cause with you. `keydown` is registered on `getIframeWindow()` (the WindowProxy, which persists across navigations, so it can genuinely accumulate), but `ng-event` is registered on `getIframeWindow().document` — and a navigation always creates a fresh `Document`. `iframe-porlet-legacy.component.ts:182-190` goes further and destroys/recreates the whole `<iframe>` via the `isLoading` gate whenever the URL changes, so the document listener starts from zero each load.

Since the duplicate save runs through `ng-event` (`RemotePublisherDialog.js:271-276` -> this listener -> `DotCustomEventHandlerService` -> `openWizard`), and `DotWizardService.open()` completes any pending subject without emitting before creating a new one, would two stacked listeners actually produce two `saveContent` calls here?

Have you been able to confirm in the Network tab that the two `ContentletAjax.saveContent.dwr` calls collapse to one? That's the unchecked item in your own test plan, and it would settle it. If it turns out the duplicate persists, would it be worth dropping the `Fixes #34534` link so the issue isn't auto-closed?

core-web/apps/dotcms-ui/src/app/view/components/_common/iframe/iframe-component/iframe.component.ts
```diff
@@ -76,6 +76,13 @@ export class IframeComponent implements OnInit, OnDestroy {
 
     private destroy$: Subject<boolean> = new Subject<boolean>();
 
+    // Stable bound references required so removeEventListener can match the
+    // exact function object that was passed to addEventListener. Using
+    // .bind(this) inline creates a new object each call, making removal a
+    // no-op and causing listeners to accumulate on every iframe load event.
```
oidacra, line 82: Small wording question on this comment. The first half is exactly right — `.bind(this)` returns a new object each call, so `removeEventListener` never matched. But the last clause, "causing listeners to accumulate on every iframe load event," reads as if it applies to both fields below it.

As far as I can tell it only holds for `keydown`, which is attached to `getIframeWindow()` (the WindowProxy keeps its identity across navigations). `ng-event` is attached to `getIframeWindow().document`, and each navigation brings a fresh `Document`, so the old listener goes away with the old document rather than piling up.

Would it be worth splitting that out so the next person reading this doesn't assume both listeners leak the same way?

core-web/apps/dotcms-ui/src/app/view/components/_common/iframe/iframe-component/iframe.component.spec.ts
```diff
@@ -219,7 +219,7 @@ describe('IframeComponent', () => {
             };
         });
 
-        it('should remove and add listener on load', () => {
+        it('should remove and add listener on load using the same stable function references', () => {
```
oidacra, line 222: Would it be worth adding a test that exercises the symptom directly? Right now the assertions stop at which references were handed to `add`/`removeEventListener` after a single `load`, so nothing covers the property that actually matters here: one dispatched `ng-event` producing exactly one emission after several reloads.

Something like triggering `load` two or three times, grabbing the handler from `contentWindow.document.addEventListener.mock.calls`, invoking it once, and asserting `custom.emit` fired once would pin the regression down. As it stands, a refactor could bring the duplicate back without turning the suite red.

core-web/apps/dotcms-ui/src/app/view/components/_common/iframe/iframe-component/iframe.component.spec.ts
```diff
@@ -228,19 +228,26 @@ describe('IframeComponent', () => {
                 }
             });
 
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+            const keyDownRef = (comp as any).boundEmitKeyDown;
```
oidacra, line 232: Would deriving these from the mock call history work just as well here? Something like:

```ts
const [, keyDownRef] = comp.iframeElement.nativeElement.contentWindow.addEventListener.mock.calls[0];
```

That asserts the same invariant without depending on the private field names, so a later rename wouldn't break the test. It would also leave the door open to switching these to `#` private fields per TYPESCRIPT_STANDARDS, which the `(comp as any)` access would otherwise block.

gortiz-dotcms:
Closing this PR. After a full investigation:

- The `ng-event` listener accumulation described in the root cause does not actually occur — each iframe navigation produces a fresh `document`, so old listeners are garbage collected with it.
- Issue #34534 was already fixed by **PR #35423** (`fix(dot-wizard): isolate output stream so cancelled wizards don't fire workflow actions`), shipped April 22, 2026. Tested on latest: only one `ContentletAjax.saveContent.dwr` call, no constraint error.
- The `keydown` listener accumulation on `contentWindow` is real but low-severity (affects Escape key behaviour in `DotIframeDialogComponent` after in-place reloads).

The `Fixes #34534` link would have auto-closed an issue that this PR did not fix, crediting the wrong commit.

PR #36807, opened 2026-07-29
fix(dev-image): pgvector reccomended installs flakes pupeline 

sfreudenthaler:
Every release is currently blocked: the dev-image arm64 (QEMU) build fails on `apt install postgresql-$PG_VERSION-pgvector` because the *recommends* chain drags in `networkd-dispatcher`/`python3-gi`, whose maintainer scripts try to wire systemd-resolved inside the container and die under emulation (newly fatal after a 2026-07-29 Ubuntu noble package update; observed on release run 30495772928 / v26.07.28-01, which then skipped Release, Release-Notes, Promote-latest, and Changelog publish).

One flag: `--no-install-recommends` — the DB container needs pgvector, not Ubuntu's network-management stack. Also slims the image.

Closes: #36806

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_012ogYXVeZsBVUHkwzdStYPJ

This PR fixes: #36806

PR #37119, opened 2026-08-20
fix(ci): stop unbounded apt-get from hanging release and PR pipelines

### LINKED ISSUE #37118: Release blocked: 'Install xmllint' apt-get update hangs for hours in CLI Deploy
## Problem

Every release attempt on 2026-08-19 stalled in `Deployment / deployment` → **CLI Deploy** → **Install xmllint**. The step runs:

```
sudo apt-get update && sudo apt-get install -y libxml2-utils
```

`azure.archive.ubuntu.com` was unreachable from the runner (every index line `Ign:`), apt fell back to `archive.ubuntu.com`, fetched the four `InRelease` files, then stalled forever on the `Packages` indices. Composite-action steps cannot carry `timeout-minutes`, so nothing bounded it — the job simply hung until a human cancelled.

Observed:

| Run | Deployment duration | Last log line before cancel |
|---|---|---|
| [32264406479](https://github.com/dotCMS/core/actions/runs/32264406479) (26.08.19-01) | 4h24m | `Get:5 …/ubuntu noble-security InRelease` @ 15:41:43, cancelled 19:15:52 |
| [32298005348](https://github.com/dotCMS/core/actions/runs/32298005348) (26.08.19-03) | 3h26m | `Get:5 …/ubuntu noble-security InRelease` @ 21:15:56, cancelled 00:02:34 |

This is unrelated to the dev-image pgvector fix in #36807 — run -01 hung here at 15:41Z, over four hours before that PR merged (20:05Z). Run -02 failed earlier, in the dev-image build, which #36807 fixed; -03 then got past the dev image and surfaced this pre-existing hang.

## Fix

`xmllint` is called exactly once in `deploy-jfrog/action.yml`, to read `distributionManagement/<repository|snapshotRepository>/url` out of `pom.xml`. Three sibling steps in the same action already read POM values with `mvn help:evaluate`. Use that here too and delete the apt step — the release path stops depending on Ubuntu package mirrors entirely.

Verified against `pom.xml` — identical output:

- `repository` → `https://repo.dotcms.com/artifactory/libs-release-local`
- `snapshotRepository` → `https://repo.dotcms.com/artifactory/libs-snapshot-local`

sfreudenthaler:
Two CI steps ran `sudo apt-get update` with nothing bounding it. When the Azure Ubuntu mirror wedged on 2026-08-19, apt fell back to `archive.ubuntu.com`, started the `InRelease` fetches, and then simply stopped making progress. A step that *hangs* sails straight past `set -e -o pipefail`, so nothing failed — the job just ran until a human cancelled it: 4h24m on [32264406479](https://github.com/dotCMS/core/actions/runs/32264406479), 3h26m on [32298005348](https://github.com/dotCMS/core/actions/runs/32298005348).

Both sites are fixed, by different means, because the dependency is different.

### 1. Release path — `deploy-jfrog/action.yml`: drop apt entirely

`xmllint` was called exactly once, to read `distributionManagement/<repository|snapshotRepository>/url` from `pom.xml`. Three sibling steps in the same action already read POM values with `mvn help:evaluate`, so use that and delete the install step. The release path now touches no Ubuntu package mirror at all.

Verified against `pom.xml` — identical to the `xmllint` xpath for both node types:

- `repository` → `https://repo.dotcms.com/artifactory/libs-release-local`
- `snapshotRepository` → `https://repo.dotcms.com/artifactory/libs-snapshot-local`

Composite-action steps can't carry `timeout-minutes`, so removing the dependency was the only real option here.

### 2. PR path — `cicd_comp_test-phase.yml`: bound it

`libvips42` / `libheif-plugin-aomenc` are genuine runtime deps for `VipsParityTest`, so this one keeps apt and gets a `timeout-minutes: 15`.

The cap is measured, not guessed. Sampling **41 successful runs** of that step across the PR, trunk, nightly and merge-queue callers:

| | |
|---|---|
| median | 18s |
| p90 | 117s |
| p95 | 216s |
| max | **521s (8m41s)** |

That max — run [32295540104](https://github.com/dotCMS/core/actions/runs/32295540104), 8m41s, conclusion `success` — is the load-bearing case, because its log says:

```
Fetched 11.4 MB in 58s (198 kB/s)
```

The mirror was **alive but throttled**, not wedged, and the install completed fine. An earlier revision of this PR used a 5m cap, which would have failed that run (2.4% of samples) and came within 36s of failing a merge-queue run at 264s ([32163267363](https://github.com/dotCMS/core/actions/runs/32163267363)) — a trip there blocks the queue, not just one PR. Since throughput degradation has no natural floor, a cap set near the slow end keeps tripping.

15m rather than 10m: the failure being caught ran 3h+, so both catch it identically, but 15m sits 1.7x above the worst real case instead of 1.2x.

Deliberately left fatal rather than `continue-on-error`: the parity tests self-skip when libvips is missing, so swallowing the failure would quietly drop that coverage instead of surfacing a retryable red X.

### Notes

Unrelated to #36807 (dev-image pgvector) — run -01 hung on this step at 15:41Z, over four hours before that PR merged at 20:05Z. What #36807 fixed was the dev-image build, which had been masking this hang further down the same job.

Ubuntu versions were not a factor: the runner was 24.04.4 on image `ubuntu-24.04` built 2026-08-10, and the Microsoft/Google third-party repos in the same `apt-get update` fetched fine in under a second.

Closes: #37118

https://claude.ai/code/session_01NX6GeNqyq1xRX1zzi1KtVn

### DIFF
--- a/.github/actions/core-cicd/deployment/deploy-jfrog/action.yml
+++ b/.github/actions/core-cicd/deployment/deploy-jfrog/action.yml
@@ -40,11 +40,6 @@ runs:
   steps:
     - uses: actions/checkout@v4
 
-    - name: Install xmllint
-      run: |
-        sudo apt-get update && sudo apt-get install -y libxml2-utils
-      shell: bash
-
     - uses: ./.github/actions/core-cicd/maven-job
       with:
         github-token: ${{ inputs.github-token }}
@@ -153,7 +148,7 @@ runs:
         echo "NODE_TYPE=$NODE_TYPE"
 
         # Extract the repository URL from the POM file
-        REPO=$(xmllint --xpath "//*[local-name()='distributionManagement']/*[local-name()='$NODE_TYPE']/*[local-name()='url']/text()" "$POM_FILE" 2>/dev/null)
+        REPO=$(mvn -f "$POM_FILE" -N help:evaluate -Dexpression=project.distributionManagement.$NODE_TYPE.url -q -DforceStdout 2>/dev/null)
         
         if [[ "$REPO" =~ ^(https://[^/]+/[^/]+)/(.+)$ ]]; then
           SERVER_URL="${BASH_REMATCH[1]}"
--- a/.github/workflows/cicd_comp_test-phase.yml
+++ b/.github/workflows/cicd_comp_test-phase.yml
@@ -254,8 +254,11 @@ jobs:
       # The libvips image engine (IMAGE_API_USE_LIBVIPS) is exercised by VipsParityTest.
       # Install native libvips on the JVM unit-test runner so those tests run instead of
       # self-skipping. Ubuntu 24.04 libvips42 bundles the pdf/heif/svg/webp/jxl delegates.
+      # Bounded: a wedged Ubuntu mirror hung this class of step for hours (#37118).
+      # Normally ~12s, so 5m fails loud and fast instead of burning the job timeout.
       - name: Install native libvips (image engine parity tests)
         if: matrix.test_type == 'jvm'
+        timeout-minutes: 5
         run: sudo apt-get update && sudo apt-get install -y --no-install-recommends libvips42 libheif-plugin-aomenc
 
       - name: Run ${{ matrix.name }}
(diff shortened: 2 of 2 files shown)


sfreudenthaler:
reguardless of how 26.08.19-04 release action plays out.  these changes are worth making

adrianjm-dotCMS (approved):

PR #37131, opened 2026-08-20
feat(content): add bulk refresh endpoint and enable the Refresh quick action

### LINKED ISSUE #36845: Workflow Center: add a Refresh quick action that reindexes the selection

rjvelazco:
Fixes #36845

Adds a **Refresh** quick action to the Content Drive Action Center that reindexes every selected contentlet, and the bulk endpoint behind it.

The single-item `PUT /api/v1/content/_refresh/{identifierOrInode}` already existed; there was no bulk equivalent over REST (only the legacy Struts `cmd=full_reindex_list`). This adds one rather than looping `_refresh` N times from the client — a client-side loop gives no cancellation, no survival across a page reload, and no server-side bound on cost.

> **This is not a full index rebuild.** `POST /api/v1/esindex/reindex` is a different operation over the whole index and is untouched.

## Endpoints

| Method | Path | Purpose |
|---|---|---|
| `POST` | `/api/v1/content/_bulkrefresh` | Submit a selection → `202` + `{jobId, statusUrl, submitted}` |
| `GET` | `/api/v1/content/_bulkrefresh/{jobId}` | Status snapshot; poll for progress and the final result |
| `POST` | `/api/v1/content/_bulkrefresh/{jobId}/cancel` | Request cancellation |

## Key decisions

**Job-backed, not synchronous.** Reindexing a 500-item selection synchronously would hold a request open for minutes. `202` rather than `200` because the work is accepted, not done — a client must not be able to read the response as "reindexed".

**Indexing is synchronous per item (`IndexPolicy.WAIT_FOR`).** The default `DEFER` only enqueues into `dist_reindex_journal` and returns, which is why the single-item endpoint can answer `true` with nothing reindexed. This endpoint exists so that "done" means done — and there's an integration test that deletes content from the index and asserts it's findable again once the job reports `SUCCESS`, which only holds if the write is synchronous.

**Per identifier, not per inode.** Content missing from search is rarely confined to one language, so three selected language rows of the same contentlet are one reindex. Each result still names every inode that resolved to it, so the client can settle each row the user selected.

**Not a workflow action or a new `SystemAction`.** Bulk fire resolves its target set by *searching the index* (`WorkflowAPIImpl` builds a `+inode:( … ) +(wfstep:…)` Lucene query), which is circular for an operation whose job is to fix the index — content missing from the index cannot be found by an index search, so the items that most need this are exactly the ones that path could never reach. The processor goes DB-first via `findAllVersions`. It would also inherit hardcoded `IndexPolicy.DEFER`, silent `wfstep` skips, and workflow side effects (actionlets, new versions, task/history rows) for an operation that changes no content.

**Unresolvable inodes are per-item failures, not request errors.** A selection can go stale between the click and the submit; one dead row must not cost the caller the rest of the batch.

**Role-gated server-side on CMS Power User or CMS Administrator,** matching the legacy `view_contentlets.jsp` check, so nobody who could press the old button loses access and nobody who could not gains it. "Always available" in the issue means *not gated by content state* — that still holds.

**Capped** by `CONTENT_BULK_REFRESH_MAX_ITEMS` (default 500), and no Lucene query is accepted: an unbounded set plus synchronous writes is a self-inflicted full reindex.

## Frontend

Refresh was already rendered but disabled. It's now selectable for any selection with ≥1 contentlet regardless of live/archived/locked state, goes straight to the preview (a reindex needs no configuration), and reports through the existing store settle path — so the toast, grid reload and selection clear all come unchanged.

**It gets its own toast copy.** The default partial-outcome message names workflow causes — *"failed (you may not have permission, or the content is locked by another user), skipped (this action is not on their workflow step)"*. Neither is why a reindex falls short, and a shortfall explained by the wrong cause sends the user off to fix something that was never the problem. So `DotContentDriveActionExecutionResult` gained an optional `partialDetailKey`; absent, the default still applies, leaving workflow actions byte-identical.

Final toast:
- Clean run → success: *"**Refresh** ran on 12 item(s)."*
- Anything short → warning: *"**Refresh**: 9 reindexed, 2 failed (the content could not be read or written to the index), 1 skipped (the reindex was cancelled before reaching them)."*

## Testing

| Suite | Result |
|---|---|
| Backend unit (processor + form) | **18/18 passing** |
| `portlets-content-drive` (full) | **1165/1165 passing** |
| `dot-bulk-refresh` service | **7/7 passing** |
| Lint — content-drive, data-access, models | clean |
| Integration (14 tests) | **written, compiling, not yet run** |
| Postman (`BulkRefreshResource`) | **written, not yet run** |

Backend unit tests were confirmed failing first (14 errors, all `UnsupportedOperationException`) before implementation, per the constitution's TDD gate.

⚠️ **The integration and Postman suites have not been executed** — no Docker daemon available locally. They need a run before merge. The integration test that matters most is `test_bulkRefresh_makesContentMissingFromTheIndexFindableAgain`, which is what proves the `WAIT_FOR` policy actually took effect.

## Video

https://github.com/user-attachments/assets/d78b1c1f-662d-4ed6-ab05-f3e7c2da499d

https://github.com/user-attachments/assets/6a8289e5-689b-4155-9f5c-45cf6ca81277

## Reviewer notes / open items

- [ ] **AC #1 is unmet**: the issue asks that the role-gating decision be *recorded on the issue*. It's implemented and documented in code, but the issue still has no comment saying so.
- [ ] **No client-side role gate.** `core-web` has no power-user awareness — `currentUserIsAdmin` comes from `getCurrentUser().admin` and there's no role list — so the only gate available would hide Refresh from exactly the users the legacy button was written for. A visible action that answers 403 through the existing error handler beats silently withholding a capability from people who have it. Flagging in case you'd rather add a lookup.
- [ ] **No live counters.** The status endpoint reports a `progress` float but nothing item-wise mid-run, so the dialog can show a percentage but not "5 of 12". Deliberate, per PO discussion. Adding them means surfacing the processor's counters via `JobQueueManagerAPI.getInstance(jobId)` on the status endpoint.
- [ ] **No SSE.** An earlier revision had an SSE progress stream; removed after PO discussion. The `SSEMonitorUtil` overload it needed was reverted, so that shared util has zero diff against `main`.
- [ ] **No cancel control in the UI.** The endpoint supports it; there's no affordance since the dialog closes on execute.
- [ ] **Unrelated pre-existing failure**: `PushPublishService` fails 4 tests (`2026-08-20` expected vs `2026-08-19` received) — its spec builds the expected date from `new Date().toISOString()` (UTC) while the service uses local time, so it fails every day after ~20:00 local. Untouched by this PR; worth its own ticket.
- `openapi.yaml` is regenerated (+132) and committed alongside the annotations, per the repo rule.

This PR fixes: #36845


dotCMS/src/main/java/com/dotcms/jobs/business/processor/impl/BulkRefreshContentletsProcessor.java
```diff
@@ -0,0 +1,395 @@
...
+            final boolean includeDependencies, final boolean recordItemResults) {
+
+        if (workItem.unresolved()) {
+            fail(workItem, workItem.resolutionError, recordItemResults);
+            return;
+        }
+
+        try {
+            // Cache first: a stale cache entry would otherwise be re-indexed as-is.
+            this.contentletCache.remove(workItem.identifier);
```
zJaaal: 🔴 **This eviction is a no-op, and it defeats the point of the operation.**

`ContentletCache` is keyed by **inode**, not identifier: `add(String inode, Contentlet)`, and `remove(Contentlet c)` delegates to `remove(c.getInode())` (`ContentletCacheImpl.java:107`, `:125`). Passing an identifier evicts nothing.

That matters more here than it would elsewhere, because the very next call reads *through* that cache: `findAllVersions` → `ESContentFactoryImpl.findAllVersions` → `findContentlets(inodes)`, which serves hits straight from `contentletCache` (`ESContentFactoryImpl.java:1294`). So when a cached version is stale, this loads the stale object and writes it to the index, which is the exact failure the comment above says it is preventing.

Mitigating: the single-item `_refresh` has the identical call (`ContentResource.java:801`), so this is faithful to precedent rather than newly wrong. But a PR whose thesis is "done should mean done" probably shouldn't inherit that one.

The fix needs the inodes before the load. Either evict `workItem.inodes` first, then evict each returned version's inode and re-read, or go through a path that bypasses the cache.

<sub>🤖 Reviewed by Claude Code, posted by @zJaaal.</sub>

dotCMS/src/main/java/com/dotcms/jobs/business/processor/impl/BulkRefreshContentletsProcessor.java
```diff
@@ -0,0 +1,395 @@
...
+            final List<Contentlet> versions =
+                    this.contentletAPI.findAllVersions(identifier, false, user, false);
+
+            for (final Contentlet version : versions) {
+                // Without this the write is only enqueued, and "done" would be a lie.
+                version.setIndexPolicy(IndexPolicy.WAIT_FOR);
+                this.contentletIndexAPI.addContentToIndex(version, includeDependencies);
+            }
+
+            final int indexed = null == versions ? 0 : versions.size();
```
zJaaal: Dead null check: the enhanced-for over `versions` two lines above would already have thrown NPE. `versions.size()` is enough.

Related, a few lines up: if version 3 of 5 throws, the item is recorded FAILED and `versionsIndexed` gets nothing, even though two versions were written. Worth incrementing inside the loop if the counter is meant to describe work done.

<sub>🤖 Reviewed by Claude Code, posted by @zJaaal.</sub>

dotCMS/src/main/java/com/dotcms/rest/api/v1/content/bulkrefresh/BulkRefreshResource.java
```diff
@@ -0,0 +1,212 @@
...
+                    + "per-item records.",
+            tags = {"Content"},
+            responses = {
+                    @ApiResponse(responseCode = "200", description = "Job status retrieved"),
+                    @ApiResponse(responseCode = "401", description = "Unauthorized - no backend user session"),
+                    @ApiResponse(responseCode = "403", description = "Forbidden - not a CMS Power User or CMS Administrator"),
+                    @ApiResponse(responseCode = "404", description = "Not Found - unknown job id"),
+                    @ApiResponse(responseCode = "500", description = "Internal Server Error")
+            })
+    public ResponseEntityView<JobView> getJobStatus(@Context final HttpServletRequest request,
```
zJaaal: 🟠 **Every status poll ships the entire submitted inode list back to the client.**

`JobContract.parameters()` has no `@JsonIgnore`, so the `JobView` returned here serializes `contentletIds` and `userId` in full. The frontend polls at 1.5s for up to 5 minutes: ~200 responses, each carrying up to 500 UUIDs (~18KB) at the configured cap. `DotBulkRefreshJob` only reads `id`, `state`, `progress` and `result`.

Interesting that the javadoc on `authorized()` already leans on this exposure as a *reason* for the role gate, so it was spotted as a security matter but not as a bandwidth one. A trimmed status view would settle both.

🟡 **Separately: neither this nor `cancelJob` checks `job.queueName()`.** `GET /api/v1/content/_bulkrefresh/{anyJobId}` returns any job's full `JobView`, and the cancel path cancels any job of any type.

Not an escalation, to be clear: `/api/v1/jobs/{jobId}/cancel` already permits exactly this to any backend user with *no* role gate at all (`JobQueueResource.java:368`), so this endpoint is strictly more restrictive than what already ships. But a one-line queue-name check makes the 404 honest and keeps the endpoint's contract to itself.

<sub>🤖 Reviewed by Claude Code, posted by @zJaaal.</sub>

dotCMS/src/main/java/com/dotcms/rest/api/v1/content/bulkrefresh/BulkRefreshResource.java
```diff
@@ -0,0 +1,212 @@
...
+    @GET
+    @Path("/{jobId}")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Operation(operationId = "getBulkRefreshStatus", summary = "Get the status of a reindex job",
+            description = "Returns the job's current state and progress while it runs, and once terminal "
+                    + "the full result including counters and — when requested at submit — the "
+                    + "per-item records.",
+            tags = {"Content"},
+            responses = {
+                    @ApiResponse(responseCode = "200", description = "Job status retrieved"),
```
zJaaal: 🟡 CLAUDE.md: *"REST @Schema: Must match actual return type."*

This returns `ResponseEntityView<JobView>` and `cancelJob` returns `ResponseEntityStringView`, but both 200s declare only a `description`, so the committed openapi has bare `"200": description:` for each. `ResponseEntityJobView` and `ResponseEntityStringView` already exist in `com.dotcms.rest`.

The `202` on submit does this correctly with `ResponseEntityBulkRefreshSubmitView`, so this reads as an omission rather than a deliberate pattern.

<sub>🤖 Reviewed by Claude Code, posted by @zJaaal.</sub>

dotCMS/src/main/java/com/dotcms/rest/api/v1/content/bulkrefresh/BulkRefreshHelper.java
```diff
@@ -0,0 +1,158 @@
...
+    /** Ceiling on inodes per submission. Synchronous indexing makes an unbounded batch expensive. */
+    public static final String MAX_ITEMS_CONFIG_PROPERTY = "CONTENT_BULK_REFRESH_MAX_ITEMS";
+
+    public static final int MAX_ITEMS_DEFAULT = 500;
+
+    private static final String STATUS_ENDPOINT = "/api/v1/content/_bulkrefresh/%s";
+
+    private final JobQueueManagerAPI jobQueueManagerAPI;
+
+    public BulkRefreshHelper() {
```
zJaaal, line 49: 🟡 **This isn't required, and it produces a broken bean.**

`ContentImportHelper`, the class cited above as the pattern this follows, is `@ApplicationScoped` with only the `@Inject` constructor and no no-arg one. Weld instantiates normal-scoped client proxies without calling a constructor, so nothing needs this.

What it does add is a public path to an instance whose `jobQueueManagerAPI` is `null`, which then NPEs on `submit`, `getJob` and `cancelJob`. Suggest deleting it.

<sub>🤖 Reviewed by Claude Code, posted by @zJaaal.</sub>

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/content/bulkrefresh/BulkRefreshResourceIntegrationTest.java
```diff
@@ -0,0 +1,490 @@
...
+
+    @BeforeEach
+    void prepare() throws Exception {
+        resource = new BulkRefreshResource(bulkRefreshHelper);
+
+        // Without this the queue accepts jobs and never runs them: every job stays PENDING and each
+        // wait below times out. ContentImportResourceIntegrationTest never noticed because it only
+        // asserts job *creation* — 24 tests in under nine seconds — so this suite had no test that
+        // actually needed a processor to execute until now.
+        if (!jobQueueManagerAPI.isStarted()) {
```
zJaaal, line 110: 🟠 **This starts the shared job queue and nothing ever stops it.**

There's no `@AfterAll` to undo it, and the class is registered in `Junit5Suite1` alongside `JobProcessorDiscoveryTest` and the job-queue tests. Once started, the queue keeps draining jobs for the rest of the JVM.

The comment right above identifies exactly this hazard for `ContentImportResourceIntegrationTest` ("only asserts job *creation*") and then leaves the queue running for it. Declaration order happens to put BulkRefresh after ContentImport, which is luck rather than isolation. An `@AfterAll` that stops the queue when this class was the one to start it would close it.

Also missing in this class: any cleanup of the contentlets and users created, and content goes on the default site rather than a dedicated one (see the repo's shared-DB isolation pattern: dedicated site + `@After` cleanup).

<sub>🤖 Reviewed by Claude Code, posted by @zJaaal.</sub>

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/content/bulkrefresh/BulkRefreshResourceIntegrationTest.java
```diff
@@ -0,0 +1,490 @@
...
+    void test_cancelJob_leavesNoItemPending() throws Exception {
+        final List<String> inodes = new ArrayList<>();
+        for (int i = 0; i < 10; i++) {
+            inodes.add(newContentlet().getInode());
+        }
+
+        final String jobId = submit(adminUser, inodes, false, true);
+        try {
+            resource.cancelJob(requestFor(adminUser), response, jobId);
+        } catch (final Exception e) {
```
zJaaal: 🟡 Swallowing every exception hides a genuine 500 behind "the race landed the other way". Narrow it to the type you actually expect from an already-terminal job, or assert on the message. As written, a `cancelJob` that starts throwing `NullPointerException` still passes this test.

<sub>🤖 Reviewed by Claude Code, posted by @zJaaal.</sub>

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/content/bulkrefresh/BulkRefreshResourceIntegrationTest.java
```diff
@@ -0,0 +1,490 @@
...
+     * {@code doesUserHaveRole(user, null)} is then silently false — so passing the lookup straight
+     * through produced a "power user" with no such role and a permission check that could only ever
+     * fail. Mirrors {@code TestUserUtils.getOrCreateAdminRole}.
+     */
+    private static Role getOrCreatePowerUserRole() throws DotDataException {
+        final Role existing = APILocator.getRoleAPI().loadRoleByKey(Role.CMS_POWER_USER);
+
+        return null != existing
+                ? existing
+                : new RoleDataGen().key(Role.CMS_POWER_USER).nextPersisted();
```
zJaaal: 🟡 This persists a fabricated role keyed `CMS Power User` into a shared test DB if the lookup misses. The branch is probably dead against a real starter, but if it ever fires it poisons every subsequent `loadRoleByKey(Role.CMS_POWER_USER)` in the JVM, including production code paths under test. Worth failing loudly instead of creating one.

<sub>🤖 Reviewed by Claude Code, posted by @zJaaal.</sub>

core-web/libs/data-access/src/lib/dot-bulk-refresh/dot-bulk-refresh.service.ts
```diff
@@ -0,0 +1,150 @@
...
+        return timer(0, DOT_BULK_REFRESH_POLL_INTERVAL_MS)
+            .pipe(
+                // exhaustMap, not switchMap: a status call slower than the interval must be allowed to
+                // finish. Cancelling and reissuing it would mean that under sustained latency no poll
+                // ever completes, nothing is ever emitted, and only the overall timeout ends the run.
+                exhaustMap(() =>
+                    this.#status(jobId).pipe(
+                        // One blip over a five-minute run is ~200 requests' worth of exposure. The job
+                        // is unaffected by a failed poll, so retrying beats abandoning it.
+                        retry({ count: DOT_BULK_REFRESH_POLL_RETRIES, delay: () => timer(1000) })
```
zJaaal: 🟡 This retries indiscriminately: a 401, 403 or 404 gets three retries a second apart before it surfaces. None of those get better by asking again, and a 404 in particular means the job id is wrong, which the user should hear immediately.

A predicate that only retries 5xx and network failures (`status >= 500 || status === 0`) would keep the "one blip over ~200 requests" reasoning intact while letting the terminal statuses through.

<sub>🤖 Reviewed by Claude Code, posted by @zJaaal.</sub>

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/features/action-execution/withActionExecution.ts
```diff
@@ -233,6 +236,101 @@ export function withActionExecution() {
...
+                                            })`
+                                        })
+                                    );
+
+                                    return;
+                                }
+
+                                onSettled({
+                                    actionName,
+                                    successCount: counts.successCount,
```
zJaaal: 🟡 **The toast reports de-duplicated identifiers as if they were selected rows.**

`counts.successCount` is per identifier, but the dialog counted eligible **inodes**. Select three language rows of one contentlet and the dialog offers "3 items" while the toast says *"Refresh ran on 1 item(s)."*

`DotBulkRefreshCounts.total` is documented honestly as post-dedup, so this is known at the model layer, but nothing reconciles it at the UI. Either submit with `includeItemResults: true` and sum `inodes.length` across the records (the server deliberately names every inode back for precisely this), or have the counters carry an inode-level figure alongside the identifier one.

<sub>🤖 Reviewed by Claude Code, posted by @zJaaal.</sub>

zJaaal (commented):
## Review

Read the full diff and cross-checked the backend against the APIs it calls. This is unusually well-documented work and the architecture calls hold up: job-backed over synchronous, per-identifier over per-inode, and the reasoning for not modelling this as a workflow action or `SystemAction` is correct (bulk fire really does resolve its target set by searching the index, which is circular here).

One finding is worth blocking on, the rest are cleanups. Inline comments have the detail.

**🔴 The cache eviction is a no-op and defeats the operation.** `contentletCache.remove(identifier)` against an inode-keyed cache evicts nothing, and `findAllVersions` then reads back through that same cache. A stale cached version gets written to the index, which is the failure this endpoint exists to fix. Same call exists in the single-item `_refresh`, so it's inherited rather than new.

**🟠 Every status poll returns the full submitted inode list** (`JobContract.parameters()` isn't `@JsonIgnore`d), ~200 times per run at up to 500 UUIDs each.

**🟠 The integration test starts the shared job queue and never stops it**, in a suite that also runs `ContentImportResourceIntegrationTest` and `JobProcessorDiscoveryTest`.

**🟡 Also:** the toast counts identifiers where the dialog counted rows; `BulkRefreshHelper`'s no-arg constructor builds a bean that NPEs; status/cancel accept any job id regardless of queue; the two 200 responses are missing `@Schema` against the repo rule; the poll retries 401/403/404.

### Verified as correct

Worth stating explicitly, since these are the load-bearing claims:

- `IndexPolicy.WAIT_FOR` genuinely bypasses the journal (`ContentletIndexAPIImpl.java:2308`). The `inTransaction()` branch below it doesn't apply here: `processNextJob` is `@CloseDBIfOpened`, not `@WrapInTransaction`.
- The "counters sit directly on `result`, not under `metadata`" comment in `dot-content-drive.model.ts` is exactly right per `OptionalJobResultSerializer`.
- `@Dependent` plus `processorInstancesByJobId.computeIfAbsent` gives one processor instance per job, so the mutable counters don't leak between runs.
- `Jdk8Module` is registered on the job queue's mapper, so `Optional<String> identifier()` round-trips through the persisted metadata.
- Path routing (matches the `/v1/content/_import` precedent), CDI discovery, and exception mapping all resolve: `IllegalArgumentException` → 400, `DotSecurityException` → 403, `DoesNotExistException` → 404, and the form's `ValueInstantiationException` → 400.
- The `FAILED` → `FAILED_PERMANENTLY` transition the frontend depends on does happen: `handleJobFailure` marks FAILED and requeues, then `processJobWithRetry` promotes it via `handleNonRetryableFailedJob`. Worth knowing the gap between the two is bounded only by the 5-minute UI timeout, so a slow promotion surfaces to the user as the 504 "stopped waiting" copy rather than as a failure.

### On the open items in the description

The two that look blocking are the ones already flagged: the integration and Postman suites have never been run, and `test_bulkRefresh_makesContentMissingFromTheIndexFindableAgain` is the only thing standing behind the central `WAIT_FOR` claim. The no-client-side-role-gate decision is defensible as argued. AC #1 is a one-comment fix on the issue.

<sub>🤖 Reviewed by Claude Code, posted by @zJaaal.</sub>

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/features/action-execution/withActionExecution.ts
```diff
@@ -233,6 +236,101 @@ export function withActionExecution() {
...
+
+                        patchState(store, {
+                            actionExecution: { actionName, total: inodes.length },
+                            actionExecutionResult: undefined
+                        });
+
+                        bulkRefreshService
+                            .refresh(inodes)
+                            .pipe(
+                                take(1),
```
zJaaal, line 302: 🟠 **This polling subscription needs `takeUntilDestroyed`.**

It's the one action in this file where the omission actually bites. The others issue a single HTTP request, so `take(1)` completes them after one round trip and the worst case on destroy is a stale toast. This one is a different shape: `timer(0, 1500)` inside the service is an infinite interval, and the `take(1)` here sits on the **outer** observable, which doesn't emit until the poll loop has already finished. It does nothing to stop the loop.

`DotContentDriveStore` is component-scoped (`dot-content-drive-shell.component.ts:120`), so it dies when the user leaves the portlet. This subscription doesn't. It keeps issuing `GET /_bulkrefresh/{jobId}` every 1.5s until the job settles or the 5-minute `timeout()` fires, up to ~200 requests against a screen nobody is on.

Two consequences past the wasted traffic:

- `httpErrorManagerService.handle(...)` is global. The 504-shaped timeout error, or the "did not report a usable outcome" 500 below, surfaces a dialog minutes after the user has navigated to a different portlet, about work they can no longer see.
- Navigate back and a fresh store is constructed with `actionExecution: undefined`, so the toolbar shows nothing in flight while the orphaned poll from the previous instance is still running and will eventually `patchState` a dead store. Two instances, one job, one of them lying.

`withMethods` already runs in an injection context (that's how the `inject(...)` default parameters on line 70 work), so `DestroyRef` goes in the same list:

```ts
withMethods(
    (
        store,
        ...
        bulkRefreshService = inject(DotBulkRefreshService),
        destroyRef = inject(DestroyRef)
    ) => {
        ...
        bulkRefreshService
            .refresh(inodes)
            .pipe(
                catchError(/* unchanged */),
                takeUntilDestroyed(destroyRef)
            )
            .subscribe(...)
```

The bare `takeUntilDestroyed()` form would throw here, since the `.pipe()` runs later inside the method rather than during store construction, hence the explicit `destroyRef`. The pattern is already established in this portlet at `dot-content-drive-field-filter-menu.component.ts:127`, so it needs no new convention.

Worth noting in a comment so it isn't later mistaken for a regression: this stops the client, not the job. The reindex continues server-side, which is the stated design intent. But that intent is about outliving the *dialog*, not the *store*, and since there's no reattach-by-jobId logic anywhere, nothing is lost by dropping the poll on destroy.

Minor, while you're in here: the `take(1)` is redundant. `last()` inside `#awaitCompletion` already guarantees a single emission before completion.

<sub>🤖 Reviewed by Claude Code, posted by @zJaaal.</sub>

dotCMS/src/main/java/com/dotcms/jobs/business/processor/impl/BulkRefreshContentletsProcessor.java
rjvelazco: Confirmed and fixed in 7639071064. Traced it: `findAllVersions` → `SELECT inode FROM contentlet WHERE identifier = ?` → `findContentlets(inodes)`, which serves cache hits at `ESContentFactoryImpl.java:1294`. So the stale object really was what got indexed, and the comment above the call described an intent the code did not achieve.

Went with your first option: evict `workItem.inodes`, read the versions to learn the rest, evict each of those through the `Contentlet` overload (which also drops the page/host/relationship caches derived from it), then read again — the second read is what gets indexed, and it is cold.

Costs a second `findAllVersions` per identifier, which seemed the right trade for indexing database truth. Added `test_process_evictsTheCacheByInodeNotByIdentifier`, which asserts `never()).remove("ident-A")` and so fails against the old code.

Left the single-item `_refresh` alone — different blast radius, and worth its own ticket rather than a drive-by here.

dotCMS/src/main/java/com/dotcms/jobs/business/processor/impl/BulkRefreshContentletsProcessor.java
rjvelazco: Both fixed. Dropped the dead null check, and `versionsIndexed` now increments inside the loop so a failure on version 3 of 5 still reports the two that landed.

dotCMS/src/main/java/com/dotcms/rest/api/v1/content/bulkrefresh/BulkRefreshResource.java
rjvelazco: Both fixed. Replaced `JobView` with a `BulkRefreshStatusView` carrying `id`, `state`, `progress` and the result metadata — the flattened `result` shape clients already read, so the Postman assertions and the TS model are unchanged. `parameters` is gone.

You are right that the `authorized()` javadoc had already spotted the exposure and treated it only as a reason to gate rather than as something to stop sending.

Queue check added to `getJob`, and `cancelJob` now resolves through it, so both answer 404 for another queue's job.

dotCMS/src/main/java/com/dotcms/rest/api/v1/content/bulkrefresh/BulkRefreshResource.java
rjvelazco: Fixed — `ResponseEntityBulkRefreshStatusView` on the status 200 and `ResponseEntityStringView` on cancel, openapi regenerated. It was an omission, not a pattern.

core-web/libs/data-access/src/lib/dot-bulk-refresh/dot-bulk-refresh.service.ts
rjvelazco: Fixed with the predicate you suggested: retries only when `status` is undefined, `0`, or `>= 500`. A 404 now surfaces on the first response.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/features/action-execution/withActionExecution.ts
rjvelazco: Confirmed and fixed. Checked the scoping you cited — `providers: [DotContentDriveStore]` at the shell's `:120` — and the mechanism is exactly as described: `take(1)` is on the outer observable, which does not emit until `last()` fires, so it never interrupts the inner `timer(0, 1500)`.

`takeUntilDestroyed(destroyRef)` added, with `DestroyRef` in the same `withMethods` injection list. Left the other actions alone since they are single-request shapes where `take(1)` genuinely completes them.

dotCMS/src/main/java/com/dotcms/rest/api/v1/content/bulkrefresh/BulkRefreshHelper.java
rjvelazco: Pushing back on the premise here: `ContentImportHelper` does have a no-arg constructor — `:93-95`, assigning `null` to the same field. So this is faithful to the cited pattern rather than a departure from it.

Your substantive point stands though: it is a public path to an instance that NPEs. I kept it because removing it risks an unproxyable-bean deployment failure for a cosmetic gain, and documented that reasoning on the constructor so the next reader does not have to re-derive it.

Happy to delete it if you would rather — it just wants a deploy to confirm Weld is fine without it, which is more than I can verify from here.

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/content/bulkrefresh/BulkRefreshResourceIntegrationTest.java
rjvelazco: Fixed the queue lifetime — `@AfterAll` stops it only when this class was the one to start it, so it hands the queue back as found. You are right that declaration order was luck rather than isolation.

Not doing the contentlet/user cleanup and dedicated site in this PR. It is a fair point about the shared-DB pattern, but it is a broader change to a test that only just started passing, and I would rather not reshape its fixtures in the same commit that fixes a correctness bug. Worth a follow-up if you want it tracked.

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/content/bulkrefresh/BulkRefreshResourceIntegrationTest.java
rjvelazco: Fixed — narrowed to `IllegalStateException | DoesNotExistException` and it logs which outcome it took. An NPE from `cancelJob` now fails the test instead of reading as "the race landed the other way".

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/content/bulkrefresh/BulkRefreshResourceIntegrationTest.java
rjvelazco: Pushing back with evidence: this branch is not dead against the CI database. Before the helper existed, `test_bulkRefresh_powerUserIsAllowed` failed with `must be a CMS Power User or a CMS Administrator` for exactly the reason you describe elsewhere — `loadRoleByKey` answered null and `doesUserHaveRole(user, null)` is silently false. So the role genuinely is absent here, and failing loudly would only reinstate that failure.

Kept the creation, mirroring `TestUserUtils.getOrCreateAdminRole`, but it now logs a warning when it fires so persisting into a shared database is visible in the output.

The interesting part is upstream of the test: production `canRefresh` inherits the same silent null from `view_contentlets.jsp:209`, so on an install without that role every Power User is denied and nothing says why. Not introduced or changed here, but it may deserve a ticket.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/features/action-execution/withActionExecution.ts
rjvelazco: Not fixing this one in this PR, deliberately — flagging rather than quietly deferring.

You are right that nothing reconciles it at the UI, and this is the third time it has come up. But both of your options change what the toast *means*, not just how it is computed, and picking between "3 rows" and "1 asset" is a copy decision I would rather not make unilaterally after already inventing the `refreshed-partial` wording.

My preference is your second option — carry an inode-level figure alongside the identifier one — since it avoids turning `includeItemResults` on just to derive a count. Left for @rjvelazco to call.

zJaaal (commented):
Code review of the bulk refresh endpoint and the Refresh quick action. 9 findings, posted inline, most severe first.

**Highest impact:** the permission gate resolves a role key that does not exist, so it degrades to admin-only, and the frontend consumes the completion event with no `jobId` correlation. That missing correlation is the root of three separate inline findings (clearing an unrelated action's guard, closing a dialog mid-edit, reacting to another tab's run); keeping the `jobId` from the 202 and filtering on it collapses all three into one fix.

Checked and found correct, so no need to re-verify: per-job processor instances (the `AtomicInteger` counters are safe), `@Dependent` scope against `JobProcessorDiscovery`, `JobCompletedEvent` firing on permanent failure, `WAIT_FOR` writing synchronously, the exception mappers producing the documented 400/403, the `refreshed-partial` placeholder order, and the `dot-folder-list-view` controlled `[selection]` mode.

_Review authored by Claude (Claude Code), posted via @zJaaal's account._

dotCMS/src/main/java/com/dotcms/rest/api/v1/content/bulkrefresh/BulkRefreshHelper.java
```diff
@@ -0,0 +1,136 @@
...
+     * whole job, parameters included — a pre-existing exposure this feature neither creates nor
+     * closes.
+     *
+     * @param user the user to check
+     * @return true for a CMS Power User or a CMS Administrator
+     */
+    public boolean canRefresh(final User user) throws DotDataException {
+
+        final RoleAPI roleAPI = APILocator.getRoleAPI();
+        return roleAPI.doesUserHaveRole(user, roleAPI.loadRoleByKey(Role.CMS_POWER_USER))
```
zJaaal: **Critical: CMS Power Users are silently locked out, the gate degrades to admin-only.**

`loadRoleByKey(Role.CMS_POWER_USER)` looks the role up by *key*, but `"CMS Power User"` appears nowhere else in the repo as a role key, and `RoleFactoryImpl.doesUserHaveRole` returns `false` (not an error) when the role is `null`.

This PR's own integration helper `getOrCreatePowerUserRole()` documents that `loadRoleByKey` "answered null here" and then *creates* the role so the test can pass, which hides the production behavior. On any database lacking that key, a real CMS Power User gets a 403 with no log line, and only CMS Administrators can bulk refresh. That contradicts the stated gate ("nobody who could press the old button loses access").

Resolve by name, or fail loudly when the role is missing.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/features/action-execution/withActionExecution.ts
```diff
@@ -233,6 +248,122 @@ export function withActionExecution() {
...
+                                    status: 500,
+                                    statusText:
+                                        'The reindex counters did not account for every item'
+                                })
+                            );
+
+                            return;
+                        }
+
+                        onSettled({
```
zJaaal: **Critical: a completed reindex clears an unrelated in-flight action's guard.**

`reportRefreshCompleted` early-returns for non-terminal states specifically so it does not touch `actionExecution`, then calls `onSettled(...)` on the SUCCESS/CANCELED path, and `onSettled` does `patchState(store, { actionExecution: undefined, ... })`.

Scenario: user fires Refresh (backgrounded, sets nothing), reselects rows, fires Lock (sets `actionExecution`, showing "Applying..." and gating the Action Center). Minutes later the refresh completion event lands, wipes `actionExecution`, the indicator disappears, and the replay guard in `executeQuickAction`/`fireLegacyServletBulk` (`if (... || store.actionExecution()) return`) opens, so Lock can be fired a second time over rows still being changed.

The store spec only covers this for `FAILED_PERMANENTLY`, which takes the early-return branch, so the real path is untested.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/features/action-execution/withActionExecution.ts
```diff
@@ -370,6 +501,25 @@ export function withActionExecution() {
...
+        ),
+        withHooks({
+            onInit(store) {
+                const eventsSocket = inject(DotEventsSocket);
+                const dotMessageService = inject(DotMessageService);
+                const destroyRef = inject(DestroyRef);
+
+                // The socket is already open app-wide, so subscribing costs nothing. This is what
+                // replaced polling: the run reports itself when it settles instead of being asked.
+                eventsSocket
```
zJaaal, line 554: **High: no correlation between the submitted job and the consumed event.**

`executeRefresh` discards the `jobId` from the 202 response entirely, and the hook reports *every* `BULK_REFRESH_COMPLETED` for the user. The event is user-scoped (`Visibility.USER`), so a refresh submitted in another browser tab, another window, or by a Login-As session reaches this store too: it toasts "Refresh ran on N item(s)" for a run this grid never started, reloads the grid, closes the dialog, and clears `actionExecution`.

Keeping the returned `jobId` and filtering on it fixes this together with the two adjacent findings.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/features/action-execution/withActionExecution.ts
```diff
@@ -370,6 +501,25 @@ export function withActionExecution() {
...
+                    .on<DotBulkRefreshCompletedEvent>(DotSystemEventType.BULK_REFRESH_COMPLETED)
+                    .pipe(takeUntilDestroyed(destroyRef))
+                    .subscribe((event) => {
+                        // Resolve the label here rather than server-side: the backend should not be
+                        // composing user-facing copy, and this keeps the wording with the rest of the
+                        // Action Center's i18n.
+                        store.reportRefreshCompleted(dotMessageService.get('Refresh'), event);
```
zJaaal, line 561: **High: the pushed completion closes a dialog the user is mid-way through configuring.**

The completion routes into the shell's `actionExecutionResultEffect`, whose `untracked` block calls `loadItems()`, `closeDialog()` and `clearActionExecutionResult()`.

Scenario: user submits Refresh, reopens the Action Center, and is filling in a Push Publish or assign-comment form. The refresh event arrives minutes later and slams the dialog shut, discarding the input, and reloads the grid underneath.

The shared settle path was written for a run whose dialog was already handed off, not for an event that can arrive at an arbitrary time.

dotCMS/src/main/java/com/dotcms/jobs/business/processor/impl/BulkRefreshContentletsProcessor.java
```diff
@@ -0,0 +1,413 @@
...
+                // Without this the write is only enqueued, and "done" would be a lie.
+                version.setIndexPolicy(IndexPolicy.WAIT_FOR);
+                this.contentletIndexAPI.addContentToIndex(version, includeDependencies);
+                // Counted per write rather than after the loop, so a failure partway still reports
+                // the versions that did land instead of silently discarding them.
+                indexed++;
+                this.versionsIndexed.incrementAndGet();
+            }
+
+            this.successCount.incrementAndGet();
```
zJaaal, line 241: **High: an item that indexed nothing is counted as a success.**

If the content is deleted between `resolve()` and `refresh()`, `identifierAPI.find()` yields an identifier with no versions and `findAllVersions` returns an empty list. The loop body never runs, yet `successCount.incrementAndGet()` fires unconditionally.

The run then reports a clean green "N reindexed" for an item where zero documents were written, which is the exact misleading success the endpoint's javadoc says it exists to remove.

Treat `versions.isEmpty()` as a failure, or at minimum log and record it.

dotCMS/src/main/java/com/dotcms/rest/api/v1/content/bulkrefresh/BulkRefreshCompletionListener.java
```diff
@@ -0,0 +1,204 @@
...
+    private void notify(final Job job, final Map<String, Object> counters, final String userId)
+            throws Exception {
+
+        final int failed = intValue(counters, EVENT_FAILED_COUNT);
+        final int skipped = intValue(counters, EVENT_SKIPPED_COUNT);
+        final int succeeded = intValue(counters, EVENT_SUCCESS_COUNT);
+
+        final String messageKey;
+        final NotificationLevel level;
+        if (JobState.SUCCESS != job.state() || (0 == succeeded && failed > 0)) {
```
zJaaal: **Medium: a user-cancelled run is reported as a failure.**

`JobState.SUCCESS != job.state()` lumps `CANCELED` in with `FAILED_PERMANENTLY` and `ABANDONED`, so a run the user themselves cancelled after 400 of 500 items succeeded records `notification.bulkrefresh.failed` ("The reindex did not complete...") at `NotificationLevel.ERROR`.

The frontend disagrees with the backend here: `reportRefreshCompleted` accepts `CANCELED` as a legitimate outcome and toasts it as a partial. Cancellation deserves its own branch, or at least WARNING.

dotCMS/src/main/java/com/dotcms/rest/api/v1/content/bulkrefresh/BulkRefreshResource.java
```diff
@@ -0,0 +1,118 @@
...
+                    @ApiResponse(responseCode = "500", description = "Internal Server Error - the job could not be created")
+            })
+    public Response bulkRefresh(@Context final HttpServletRequest request,
+            @Context final HttpServletResponse response,
+            final BulkRefreshForm form) throws DotDataException, DotSecurityException {
+
+        final User user = init(request, response).getUser();
+
+        Logger.debug(this, () -> String.format("User %s is submitting %d inode(s) to be reindexed",
+                user.getUserId(), form.getContentletIds().size()));
```
zJaaal, line 101: **Medium: an empty request body is a 500, not the documented 400.**

With `@Consumes(APPLICATION_JSON)` and no body (or a literal `null`), Jersey passes `form == null`. `BulkRefreshForm`'s bean validation never runs, and the first dereference (`form.getContentletIds()` in the debug lambda, or inside `helper.submit`) NPEs into `RuntimeExceptionMapper`, giving a 500. The `@ApiResponse` set promises 400 for "empty selection".

A null check returning `Response.Status.BAD_REQUEST` closes it.

dotCMS/src/main/java/com/dotcms/rest/api/v1/content/bulkrefresh/BulkRefreshForm.java
```diff
@@ -0,0 +1,77 @@
...
+ * Duplicates are allowed and collapsed by identifier server-side. Inodes that no longer resolve are
+ * <b>not</b> a request error — they count toward {@code failedCount} rather than rejecting the batch,
+ * because a selection can go stale between the click and the submit.
+ *
+ * @author dotCMS
+ */
+public class BulkRefreshForm extends Validated {
+
+    @NotNull(message = "A non-empty list of contentlet inodes is required")
+    @Size(min = 1, message = "A non-empty list of contentlet inodes is required")
```
zJaaal, line 27: **Low: the item cap is enforced only after full deserialization.**

`@Size(min = 1)` has no `max`, and the `CONTENT_BULK_REFRESH_MAX_ITEMS` check lives in `BulkRefreshHelper.submit` *after* `canRefresh`. A caller can POST an arbitrarily large `contentletIds` array and the whole list is materialized in the JVM before anything rejects it.

Adding `max` to `@Size` (or a body-size cap) makes the documented cap enforced at parse time.

dotCMS/src/main/java/com/dotcms/jobs/business/processor/impl/BulkRefreshContentletsProcessor.java
```diff
@@ -0,0 +1,413 @@
...
+            // because findAllVersions reads back *through* that cache
+            // (ESContentFactoryImpl.findContentlets serves hits straight from it). So an identifier-
+            // keyed eviction leaves a stale cached version to be written to the index, which is the
+            // precise failure this endpoint exists to repair.
+            //
+            // Hence: evict the inodes we know, read the versions to learn the rest, evict those too,
+            // then read again. The second read is the one whose result gets indexed, and it is cold,
+            // so what reaches the index came from the database rather than from the cache.
+            workItem.inodes.forEach(this.contentletCache::remove);
+            this.contentletAPI.findAllVersions(identifier, false, user, false)
```
zJaaal: **Low (efficiency): `findAllVersions` is executed twice per identifier purely to warm-then-evict the cache.**

The first call exists only so its results can be fed to `contentletCache::remove`. Each `findAllVersions` is a DB round trip plus per-version permission filtering, so a 500-item submission does roughly 1000 version reads instead of 500, on top of the 500 `contentletAPI.find` calls in `resolve()`.

Since the versions of an identifier are reachable from `identifier` plus `VersionableAPI`/`ESContentFactory` inode lookup, the warm-up read is avoidable.

zJaaal (approved):
Last feedback, good to merge after resolving my latests comments

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/features/action-execution/withActionExecution.ts
rjvelazco: Fixed in 0b4fa93. The success branch went through `onSettled`, which clears `actionExecution`; it now publishes only `actionExecutionResult` and leaves that field alone on every path.

Test added for the path that was untested: `should not clear an unrelated action that is still in flight on the success path` — fires Lock over a `NEVER` observable after a backgrounded reindex, then lands the SUCCESS event and asserts Lock's `actionExecution` is the same object reference.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/features/action-execution/withActionExecution.ts
rjvelazco: Fixed in 0b4fa93, and you were right that this collapses the three. `BulkRefreshCompletionListener` now puts the job id in the payload (`EVENT_JOB_ID`), the store keeps the id from its own 202 in `refreshJobIds`, and `reportRefreshCompleted` returns early for anything else — silently, since an error toast would blame the user for another tab's event.

One trade worth naming: the ids live in store state, not storage, so a page reload loses them and a run submitted before the reload settles silently. The bell notification still records it, which seemed better than reacting to somebody else's run.

Covered by `should ignore a run this store never submitted` and `should settle a given run only once`.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/features/action-execution/withActionExecution.ts
rjvelazco: Fixed in 0b4fa93. Results now carry `backgrounded`, set only by the pushed completion. The shell never calls `closeDialog()` for one, and skips `loadItems()` while a dialog is open — reloading pulls the rows out from under the form being filled in, which was the other half of this.

With no dialog open it still reloads, since that is the common case and nothing is at risk. Two shell specs cover both branches.

dotCMS/src/main/java/com/dotcms/jobs/business/processor/impl/BulkRefreshContentletsProcessor.java
rjvelazco: Fixed in 0b4fa93. `versions.isEmpty()` now routes to `fail(...)` with "The content no longer has any version to reindex" instead of falling through to `successCount.incrementAndGet()`.

`test_process_anIdentifierWithNoVersionsIsNotASuccess` asserts successCount 0, failedCount 1, versionsIndexed 0, and that `addContentToIndex` was never called.

dotCMS/src/main/java/com/dotcms/rest/api/v1/content/bulkrefresh/BulkRefreshCompletionListener.java
rjvelazco: Fixed in 0b4fa93. `JobState.CANCELED` gets its own branch at `WARNING` with a new key, `notification.bulkrefresh.cancelled` ("Reindex cancelled: {0} item(s) reindexed, {1} failed, {2} not attempted."), so the backend and the toast now agree.

`test_onJobCompleted_cancelledRunIsNotReportedAsAFailure` covers 400 of 500 cancelled.

dotCMS/src/main/java/com/dotcms/rest/api/v1/content/bulkrefresh/BulkRefreshResource.java
rjvelazco: Fixed in 0b4fa93. Null form throws `IllegalArgumentException`, which `IllegalArgumentExceptionMapper` maps to the documented 400 — same route the item-cap check already took.

Covered twice: an integration test calling the resource with a null form, and a Postman request that posts no body at all, so the real Jersey path is exercised rather than just the guard.

dotCMS/src/main/java/com/dotcms/rest/api/v1/content/bulkrefresh/BulkRefreshHelper.java
rjvelazco: Confirmed the mechanism, and it is worse than "appears nowhere else" — I unzipped the starter and listed every role key. `CMS Administrator`, `CMS Anonymous`, `CMS Owner`, `Login As`, `Scripting Developer`, `Form Editor` are all there. `CMS Power User` is not. So `loadRoleByKey` answers null, `doesUserHaveRole(user, null)` is false, and the gate is admin-only. All correct.

Where I'd push back is the conclusion that this contradicts the stated gate. `view_contentlets.jsp:209` and `view_contentlets_js_inc.jsp:17` make the **identical** call, so the legacy Refresh button is admin-only in practice too. Nobody who could press the old button loses access — the honest reading is that the old button never admitted Power Users either.

That makes "resolve by name" a widening of the gate beyond both legacy and the AC's "conservative default", which I don't think a code review should decide. So in 0b4fa93 I fixed the silence, not the outcome: the miss is now logged as a warning naming the key, and the javadoc states plainly that this resolves to CMS Administrator alone and why.

Leaving this thread open — whether real Power Users should be able to bulk reindex is a product call, and if the answer is yes it wants its own change (and probably an upgrade task seeding the role, since resolving by name would depend on an unenforced display name).

dotCMS/src/main/java/com/dotcms/rest/api/v1/content/bulkrefresh/BulkRefreshForm.java
rjvelazco: Not doing this one, because the stated benefit doesn't hold: Jackson deserializes the whole body before bean validation runs, so `@Size(max = ...)` rejects the list after it has already been materialized — exactly the cost the finding is trying to avoid. It would move the error from `IllegalArgumentException` to a constraint violation and change nothing about memory.

There's a second reason to leave it: the cap is `CONTENT_BULK_REFRESH_MAX_ITEMS`, configurable at runtime. A static `max` in the annotation would silently override it — set the property to 2000 and requests would still fail at whatever number was compiled in.

The real fix for the concern is a request body size limit at the container level, which is infrastructure rather than this form. Leaving open in case you disagree, or want that filed separately.

dotCMS/src/main/java/com/dotcms/jobs/business/processor/impl/BulkRefreshContentletsProcessor.java
rjvelazco: Fixed in 0b4fa93, though not quite as suggested — the naive version loses cache invalidation.

`ContentletCacheImpl.remove(Contentlet)` does strictly more than `remove(String)`: beyond the inode it invalidates the page cache via `PageLoader` for pages, the site cache via `SiteLoader` for hosts, and the relationship cache. Evicting only by inode from a `Versionable` would have left all three stale — a correctness regression traded for a DB round trip.

So both properties are kept: `VersionableAPI.findAllVersions(identifier)` supplies the inodes cheaply before the read (no contentlet cache, no per-version permission filtering) so the read is cold, and the `Contentlet` overload runs on the versions afterwards so the derived caches still go. One permission-filtered read per identifier instead of two, nothing lost.

`test_process_readsThePermissionFilteredVersionsOnlyOnce` pins the count, and the existing `test_process_evictsTheCacheByInodeNotByIdentifier` still asserts the `Contentlet` overload is used.

dotcms-integration/src/test/java/com/dotcms/Junit5Suite1.java
```diff
@@ -18,6 +19,7 @@
         JobQueueHelperIntegrationTest.class,
         ImportContentletsProcessorIntegrationTest.class,
         ContentImportResourceIntegrationTest.class,
+        BulkRefreshResourceIntegrationTest.class,
```
dario-daza, line 34: I'm not sure whether to include the integration test here or in one of our MainSuites, ask Nolly or Fabro.

rjvelazco: @fabrizzio-dotCMS @nollymar

dario-daza (approved):

dotcms-integration/src/test/java/com/dotcms/Junit5Suite1.java
nollymar: @dario-daza is right @rjvelazco . It must be in a MainSuite

rjvelazco: It has to be here rather than a MainSuite, and it's a framework question rather than a subject-area one — happy to have Nolly or Fabro confirm, but the code settles it.

The MainSuites are JUnit 4: `MainSuite2a` is `@RunWith(MainBaseSuite.class)` with `org.junit.runners.Suite.SuiteClasses`, and `MainBaseSuite extends org.junit.runners.Suite`. This suite is JUnit 5: `@Suite` + `@SelectClasses` from `org.junit.platform.suite.api`.

`BulkRefreshResourceIntegrationTest` is JUnit 5 — `org.junit.jupiter.api.Test`, `@EnableWeld`, `extends Junit5WeldBaseTest`. Listed in a JUnit 4 `@SuiteClasses` it would be **silently skipped**: the JUnit 4 runner finds no `@Test` it recognises, so the class contributes zero tests and nothing reports that it did not run.

That failure mode already bit this PR once. The test existed and passed locally for a while before I noticed it was in no suite at all, so CI had never executed it. A green pipeline that runs none of your new tests is the worst version of this, which is why I'd rather it stay somewhere it demonstrably runs — "Integration Tests - Junit5 Suite 1" is green on this run with it included.

Precedent too: `ContentImportResourceIntegrationTest` and `ImportContentletsProcessorIntegrationTest` are already here, and they're the same shape — a job-backed REST resource plus its processor.

Pushed a class javadoc on `Junit5Suite1` recording the rule, since the question is a fair one to have and the file gave no hint either way: framework decides the suite, both families are picked up by failsafe (`**/MainSuite*.java` and `**/Junit5Suite*.java` in dotcms-integration/pom.xml), and each becomes its own CI job.

PR #37134, opened 2026-08-20
37132 new assetpicker must not be used in the legacy dojo editor pick the picker per host wysiwyg block editor fileimage

### LINKED ISSUE #37132: New AssetPicker must not be used in the legacy Dojo editor — pick the picker per host (WYSIWYG, Block Editor, File/Image)
## Problem Statement

PR #36944 (epic #36702) wired the new **`DotAssetPickerComponent`** into every asset-selection entry point unconditionally. Three of those entry points are Angular components that the **legacy Dojo/JSP edit contentlet** also mounts as custom elements, so the new AssetPicker now opens inside the old editor — a UI that was never designed for it and does not match anything else on that page.

The choice of picker must be **host-driven**, not global:

| Host | Expected picker |
|---|---|
| New Edit Content (Angular shell / slide-in side panel) | `DotAssetPickerComponent` (new) |
| Legacy edit contentlet (Dojo/JSP, fields mounted as custom elements) | The pre-#36944 picker |

**Who it impacts:** every customer still on the legacy edit contentlet — i.e. anyone who has not enabled the new Edit Content, which is the current default for most installs. Front-end only; reproduces in all browsers.

### Legacy hosts involved

The Dojo editor mounts three Angular custom elements, and all three now reach the new picker:

| Custom element | Registered in | Mounted by |
|---|---|---|
| `<dotcms-block-editor>` / `<dotcms-old-block-editor>` | `core-web/apps/dotcms-block-editor/src/main.ts` | `dotCMS/src/main/webapp/html/portlet/ext/contentlet/field/edit_field.jsp:248` (tag chosen by `FEATURE_FLAG_NEW_BLOCK_EDITOR`); scripts at `edit_contentlet.jsp:88-91` |
| `<dotcms-binary-field>` | `core-web/apps/dotcms-binary-field-builder/src/app/app.module.ts` | scripts at `edit_contentlet.jsp:114-116` |

### Code paths that open the new picker unconditionally

| Field | File | Entry point |
|---|---|---|
| Block Editor (image / video / audio) | `core-web/libs/new-block-editor/src/lib/editor/services/editor-modal.service.ts` | `openAssetPicker()` → `mountAssetPicker()` |
| WYSIWYG (TinyMCE `dotAddImage`) | `core-web/libs/edit-content/src/lib/fields/dot-edit-content-wysiwyg-field/dot-wysiwyg-plugin/dot-wysiwyg-plugin.service.ts` | `dotImageDialog()` → `openImagePicker()` |
| File / Image | `core-web/libs/edit-content/src/lib/fields/dot-edit-content-file-field/components/dot-file-field/dot-file-field.component.ts` | `showSelectExistingFileDialog()` |

### Picker to restore per field (pre-#36944 behavior)

| Field | Component to open in the legacy host | Notes |
|---|---|---|
| WYSIWYG | `DotAssetSearchDialogComponent` (`@dotcms/ui`) | header `Insert Image`, `data: { assetType: 'image' }` |
| Block Editor | `DotBrowserSelectorComponent` (`@dotcms/ui`) | via `buildBrowserSelectorConfig({ header, mimeTypes })` — **deleted** by #36944 from `core-web/libs/new-block-editor/src/lib/editor/config.utils.ts` and must be restored |
| File / Image | `DotBrowserSelectorComponent` (`@dotcms/ui`) | headers `dot.file.field.dialog.select.existing.image.header` / `...file.header`; inline dialog config previously built in `showSelectExistingFileDialog()` |

### Proposed mechanism

Follow the precedent already in the codebase for exactly this problem — the image editor's **`IMAGE_EDITOR_LAUNCHER`** injection token (`core-web/libs/edit-content/src/lib/fields/shared/image-editor-launcher/image-editor-launcher.token.ts`), provided **only** by the Angular hosts:

- `core-web/libs/edit-content/src/lib/edit-content.shell.component.ts`
- `core-web/libs/edit-content/src/lib/components/dot-edit-content-side-panel/dot-edit-content-side-panel.component.ts`

Introduce an equivalent host-capability token (e.g. `ASSET_PICKER_LAUNCHER`) provided by those same two Angular hosts. Injected with `{ optional: true }`: **present → new AssetPicker; absent → legacy picker.** A server-side `Config` / feature flag is *not* a valid substitute here — the discriminator is which host mounted the component, not a customer setting.

## Steps to Reproduce

**Video screencast:** NA — screenshots in the epic thread / see tables below.

**Detailed steps:**

1. Use a build from `main` that contains #36944 (merged 2026-08-19).
2. Make sure the **legacy edit contentlet** is what opens (new Edit Content disabled).
3. Create a content type with a **WYSIWYG** field, a **Block Editor (Story Block)** field, and an **Image** (or File) field.
4. Edit a contentlet of that type in the legacy editor.
5. **Block Editor field** → slash menu (or toolbar) → *Image* / *Video* / *Audio*.
   - **Actual:** the new `DotAssetPickerComponent` opens (folder tree + asset table + Content Types / Locale chips + Upload).
   - **Expected:** `DotBrowserSelectorComponent`, as before #36944.
6. **Image / File field** → *Select Existing File*.
   - **Actual:** the new `DotAssetPickerComponent` opens.
   - **Expected:** `DotBrowserSelectorComponent` with the `dot.file.field.dialog.select.existing.*` header.
7. **WYSIWYG field** → *Add image* toolbar button.
   - **Actual:** the new `DotAssetPickerComponent` opens.
   - **Expected:** `DotAssetSearchDialogComponent` ("Insert Image").
8. Now open the **same content type in the new Angular Edit Content** and repeat 5–7.
   - **Expected:** the new `DotAssetPickerComponent` in all three — this path must not regress.

> **To verify during implementation — WYSIWYG legacy mount point.** The legacy editor renders WYSIWYG as a plain `textarea` plus JSP-side TinyMCE (`edit_field.jsp:465-490`, `edit_field_js.jsp`), whose `dotAddImage` button comes from `dotCMS/src/main/webapp/html/js/tinymce/js/tinymce/plugins/dotCustomButtons/plugin.min.js` and opens the Dojo `FileBrowserDialog` ("Select a file", `html/js/dotcms/dijit/FileBrowserDialog.js`). No Angular custom element for the WYSIWYG field was found. However `DotWysiwygPluginService` documents itself as being constructed by "hosts that are not the Edit Content shell (legacy Dojo pages)". Confirm whether the Angular WYSIWYG field is genuinely reachable from the legacy editor. If it is not, the WYSIWYG change is a guard for consistency rather than a live regression — implement it either way so the three entry points behave identically.

## Acceptance Criteria

**Host detection**

- [ ] A host-capability mechanism exists (injection token, mirroring `IMAGE_EDITOR_LAUNCHER`) that tells an asset-selection entry point whether it is running in the new Angular Edit Content or in a legacy host
- [ ] It is provided **only** by `edit-content.shell.component.ts` and `dot-edit-content-side-panel.component.ts`; every other host leaves it absent
- [ ] It is injected with `{ optional: true }` — a missing provider resolves to the legacy path and never throws
- [ ] Detection is host-based, not driven by a server-side `Config` key or feature flag

**Block Editor field**

- [ ] In the new Angular Edit Content, *Image* / *Video* / *Audio* open `DotAssetPickerComponent` (unchanged from today)
- [ ] In the legacy Dojo editor (`<dotcms-block-editor>` and `<dotcms-old-block-editor>`), they open `DotBrowserSelectorComponent` with the pre-#36944 mime-type scoping (`['image']`, `['video']`, `['audio']`)
- [ ] `buildBrowserSelectorConfig()` is restored in `core-web/libs/new-block-editor/src/lib/editor/config.utils.ts`, including its `OVERLAY_ABOVE_FULLSCREEN_Z_INDEX` base z-index so the picker clears the full-screen editor shell's `z-[9998]` backdrop
- [ ] Inserting an asset from the legacy picker still produces the correct `dotImage` / `dotVideo` / `dotAudio` node at the current selection

**WYSIWYG field**

- [ ] In the new Angular Edit Content, *Add image* opens `DotAssetPickerComponent` (unchanged)
- [ ] In a legacy host, it opens `DotAssetSearchDialogComponent` with `assetType: 'image'`
- [ ] Either picker inserts the image via `formatDotImageNode()` honoring the `WYSIWYG_IMAGE_URL_PATTERN` property, and returns focus to the editor on close (insert **and** dismiss)

**File / Image field**

- [ ] In the new Angular Edit Content, *Select Existing File* opens `DotAssetPickerComponent` (unchanged)
- [ ] In the legacy Dojo host (`<dotcms-binary-field>`), it opens `DotBrowserSelectorComponent` with `mimeTypes: ['image']` for Image fields and `[]` for File fields, and the `dot.file.field.dialog.select.existing.image.header` / `...file.header` header
- [ ] The selected asset is written to the store identically on both paths (preview, `fileName` / title auto-fill, and the legacy binary-field contract consumed by the JSP editor)
- [ ] It is decided and recorded whether `dot-asset-picker.component.legacy-host.spec.ts` stays as a safety net or is removed, since the picker is no longer opened from that host

**Regression / edge cases**

- [ ] Double-clicking the trigger still opens exactly one dialog on **both** paths (the pending-guard behavior added in #36944 is preserved for the new picker and does not leak into the legacy path)
- [ ] Dismissing via ✕, Esc, or the overlay mask leaves no orphan dialog ref and no stuck pending flag on either path
- [ ] Destroying the field (navigating away, closing the side panel, removing the custom element) closes any open dialog on either path
- [ ] The legacy picker requires no `Router` and no app-shell providers — it constructs in the binary-field host, which bootstraps with only `provideHttpClient`, `provideAnimations`, `DotMessageService`, `DotUploadService`, `DotWorkflowActionsFireService`
- [ ] When the current site cannot be resolved, the new-picker path still opens nothing (existing behavior) and the legacy path is unaffected, since it does not need a `DotSite`
- [ ] Unit tests cover both branches for each of the three entry points — a legacy-host factory (token absent) and an Angular-host factory (token provided) — following the pattern of `dot-file-field.component.legacy-availability.spec.ts`
- [ ] The existing Playwright specs added by #36944 (`wysiwyg-field-asset-picker.spec.ts`, `block-editor-field-asset-picker.spec.ts`, `file-field.spec.ts`, `image-field.spec.ts`) still pass, since they exercise the new Angular Edit Content

## dotCMS Version

Latest from `main` — regression introduced by #36944, merged 2026-08-19. Not present in any released version yet.

## Severity

## Links

- Epic: #36702 — Create AssetPicker component reusing Content Drive (File/Image field asset selection)
- Regressing PR: #36944 — Issue 36702 asset picker
- Freshdesk ticket: NA

nicobytes:
This pull request introduces a new Angular-based Asset Picker for file and image selection in the Edit Content experience, ensuring it is only available in modern Angular hosts while maintaining the legacy picker for older or custom element hosts. It also adds comprehensive tests to verify both picker behaviors. The changes ensure a seamless transition between the new and legacy pickers depending on the hosting context, and provide clear dependency injection for the new picker.

**Asset Picker Integration (Angular Hosts):**

- Registers the new `AngularAssetPickerLauncher` as the provider for `ASSET_PICKER_LAUNCHER` in all Angular Edit Content hosts (`dot-create-content-dialog`, side panel, and shell), ensuring the new Asset Picker is used in these contexts. [[1]](diffhunk://#diff-6b9478344893b9382bfd2c1d0e0ee915329d27cd6b17e8785065425c5a239bbaL57-R63) [[2]](diffhunk://#diff-3edcb1e4ac7b9d7d5891517e7d63ef0a828a6262699b1cc9d7a9f3e780990b03R87-R89) [[3]](diffhunk://#diff-cd2e6a94870558b0e3ab8be35a6a2488283a0cd46c02421fd4b0b81d3146d85aR27-R30)
- Updates imports and providers to include `ASSET_PICKER_LAUNCHER` and `AngularAssetPickerLauncher` where needed. [[1]](diffhunk://#diff-6b9478344893b9382bfd2c1d0e0ee915329d27cd6b17e8785065425c5a239bbaL18-R18) [[2]](diffhunk://#diff-3edcb1e4ac7b9d7d5891517e7d63ef0a828a6262699b1cc9d7a9f3e780990b03L25-R25) [[3]](diffhunk://#diff-cd2e6a94870558b0e3ab8be35a6a2488283a0cd46c02421fd4b0b81d3146d85aR8-R9) [[4]](diffhunk://#diff-5a87db878b1084c2416970b0884a73458af5bd0ee97ef02bc70ec981f13b0800R14-R23) [[5]](diffhunk://#diff-5a87db878b1084c2416970b0884a73458af5bd0ee97ef02bc70ec981f13b0800R68-R72)

**Legacy Picker Preservation (Custom Element/Legacy Hosts):**

- Modifies `DotFileFieldComponent` to inject `ASSET_PICKER_LAUNCHER` optionally; if absent, falls back to the legacy `DotBrowserSelectorComponent` for file selection, maintaining previous behavior in legacy hosts. [[1]](diffhunk://#diff-e3c1fad86db0fdf78277c6bac4feccab190111bcc28987a5785a6aef3731ae9bR42-R45) [[2]](diffhunk://#diff-e3c1fad86db0fdf78277c6bac4feccab190111bcc28987a5785a6aef3731ae9bR132-R138) [[3]](diffhunk://#diff-e3c1fad86db0fdf78277c6bac4feccab190111bcc28987a5785a6aef3731ae9bR862-R872)
- Adds a dedicated test suite (`dot-file-field.component.legacy-picker.spec.ts`) to verify the legacy picker is used and behaves as expected when the new launcher is not provided.

**Testing and Verification:**

- Expands the main file field component test to verify the new Asset Picker is opened and correctly scoped when the launcher is present (Angular hosts).
- Ensures both picker paths prevent multiple dialogs and handle dialog closure and field disabling correctly.

**Project Configuration:**

- Adds a feature directory entry for tracking this work in `.specify/feature.json`.

Closes #37132


adrianjm-dotCMS (dismissed):

core-web/libs/edit-content/src/lib/fields/dot-edit-content-file-field/components/dot-file-field/dot-file-field.component.ts
```diff
@@ -853,8 +859,17 @@ export class DotFileFieldComponent
         // The site lookup is what makes opening asynchronous, so `#dialogRef` alone is not a
         // sufficient guard — it is still null while the request is in flight, and a second click
         // would sail past it and stack a second dialog whose `onClose` stays live but unreachable.
+        // It guards the legacy path too, which is synchronous but must not stack either.
         this.#assetPickerPending = true;
 
+        // No launcher means a legacy host, which browses without a site — so it skips the lookup
+        // entirely rather than paying for a request it has no use for.
+        if (!this.#assetPickerLauncher) {
```
oidacra, line 867: The new legacy branch returns before the site lookup, which is right, but it means `#assetPickerPending` (set just above, line 863) is now also guarding a synchronous path whose only reset happens inside `#trackPicker`'s `onClose`. If `#dialogService.open(...)` or the `#dotMessageService.get(...)` header lookup throws, the exception escapes `showSelectExistingFileDialog()` before any close handler is wired and the flag stays `true` — "Select Existing File" is then dead for the rest of the session with no toast and no log.

`EditorModalService.openAssetPicker()` avoids this by never setting the pending flag on its legacy branch at all. Would it be worth either mirroring that ordering here, or wrapping the open in a `try`/`catch` that clears the flag?

Scope note: the flag itself predates this PR (#36944); the synchronous branch it now also guards is new here.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-wysiwyg-field/dot-wysiwyg-plugin/dot-wysiwyg-plugin.service.ts
```diff
@@ -126,6 +152,12 @@ export class DotWysiwygPluginService {
 
         this.imagePickerBusy = true;
 
+        if (!this.assetPickerLauncher) {
```
oidacra, line 160: Same shape as the file-field comment: `imagePickerBusy` is set at line 153, and this new legacy branch returns before the site lookup, so on this path the flag is only ever cleared inside `trackImagePicker`'s `onClose`. A synchronous throw out of `openLegacyImageDialog` would leave it `true` and permanently disable the toolbar's Add image button.

Would a `try`/`catch` around the open that resets `imagePickerBusy` be worth adding on both branches?

core-web/libs/edit-content/src/lib/fields/dot-edit-content-wysiwyg-field/dot-wysiwyg-plugin/dot-wysiwyg-plugin.service.ts
```diff
@@ -47,6 +46,27 @@ export class DotWysiwygPluginService {
      */
     private readonly editContentStore = inject(DotEditContentStore, { optional: true });
 
+    /**
+     * Present only in the Angular Edit Content, the sole host the new AssetPicker was built for.
+     * Absent in the legacy Dojo pages, where *Add image* falls back to
```
oidacra: This reads as settled fact that the legacy Dojo pages construct this service and that Add image genuinely falls back to `DotAssetSearchDialogComponent` there. But `specs/37132-picker-per-host/spec.md`'s own Assumptions section concludes the opposite: the legacy editor renders WYSIWYG as a plain textarea plus JSP-side TinyMCE, there is no Angular custom element for the field, and the change is therefore a consistency guard rather than a live regression.

Since the doc block and the spec now disagree, could the wording be softened to match the spec, e.g. noting no legacy host is currently known to construct this service without the Angular shell? The same claim appears again in the `dotImageDialog()` JSDoc below.

core-web/libs/new-block-editor/src/lib/editor/services/editor-modal.service.legacy-host.spec.ts
```diff
@@ -0,0 +1,192 @@
...
+            expect(openedConfig().data.mimeTypes).toEqual(mimeTypes);
+        });
+
+        it('should title the dialog through PrimeNG chrome, as the legacy picker has no header of its own', () => {
+            expect(openedConfig().header).toBe(titleKey);
+        });
+
+        it('should clear the fullscreen editor shell backdrop', () => {
+            // Without this the modal renders under the shell's `z-[9998]` and is unreachable.
+            expect(openedConfig().baseZIndex).toBe(OVERLAY_ABOVE_FULLSCREEN_Z_INDEX);
```
oidacra: This block asserts four things about the config (component, `mimeTypes`, `header`, `baseZIndex`), and the z-index assertion in particular is a good catch. `buildBrowserSelectorConfig()` returns a dozen-plus fields though — `closable`, `closeOnEscape`, `dismissableMask`, `draggable`, `keepInViewport`, `resizable`, `modal`, `width`, `style`, `contentStyle`, plus the `showLinks`/`showDotAssets`/`showPages`/`showFiles`/`showFolders`/`showWorking`/`showArchived`/`sortByDesc` browse flags.

Given that restoring this config exactly is one of the issue's acceptance criteria, would asserting the whole object be stronger, e.g. `expect(openedConfig()).toEqual(buildBrowserSelectorConfig({ header: titleKey, mimeTypes }))`? As written, a dropped browse flag would still pass.

core-web/libs/new-block-editor/src/lib/editor/services/editor-modal.service.legacy-host.spec.ts
```diff
@@ -0,0 +1,192 @@
...
+        expect(siteService.getCurrentSite).not.toHaveBeenCalled();
+    });
+
+    describe('inserting the picked asset', () => {
+        it.each([
+            ['openImagePicker', () => insertImage, () => [insertVideo, insertAudio]],
+            ['openVideoPicker', () => insertVideo, () => [insertImage, insertAudio]],
+            ['openAudioPicker', () => insertAudio, () => [insertImage, insertVideo]]
+        ] as const)('should insert the node %s corresponds to', (method, expected, others) => {
+            const contentlet = { identifier: 'id-1', inode: 'inode-1' } as DotCMSContentlet;
```
oidacra: Would `createFakeContentlet({ identifier: 'id-1', inode: 'inode-1' })` from `@dotcms/utils-testing` work here? The `as DotCMSContentlet` cast suppresses the fact that the literal is missing nearly all of the interface's required fields, so it would keep compiling even if the inserters later start reading another property.

core-web/libs/ui/src/lib/components/dot-asset-picker/angular-asset-picker.launcher.spec.ts
```diff
@@ -0,0 +1,84 @@
+import { createServiceFactory, mockProvider, SpectatorService } from '@openng/spectator/jest';
+
+import { DialogService, DynamicDialogRef } from 'primeng/dynamicdialog';
+
+import { DotSite } from '@dotcms/dotcms-models';
+
+import { AngularAssetPickerLauncher } from './angular-asset-picker.launcher';
+import { DotAssetPickerComponent } from './dot-asset-picker.component';
+
+const SITE: DotSite = {
```
oidacra: Minor one: `createFakeSite({ identifier: 'site-1', hostname: 'dotcms.com' })` would cover this and stay in sync if `DotSite` gains fields. This is properly typed rather than cast, so it is lower-risk than a cast — just noting the shared factory exists.

nicobytes:
Superseded by #37196.

Same changes, byte-for-byte — verified with `git diff` between this branch's tip (`b6e62ced`) and the new commit.

**Why:** this PR was opened before the `2026-08-24_incident-response` ruleset landed, so 6 of its 9 commits are unsigned. That ruleset requires verified signatures and blocks force-pushes (`non_fast_forward`), making those commits impossible to sign in place. #37196 is a single signed commit (`verified=true`) off current `main`.

Closing this one in favour of #37196.

Issue #36763, opened 2026-07-27
Experiments: A/B Testing v2


fmontes:
## Summary

Rebuild dotCMS A/B testing on the new analytics platform. Four parts, one epic:

1. **Migrate Experiments to the new analytics infrastructure** — move goal evaluation off CubeJS onto CAEM + ClickHouse
2. **Build the new Experiments UI from scratch** — per the approved design below
3. **Make the SDK and the traditional page experience work** — one analytics SDK for both rendering modes
4. **Update the docs** — rewritten against what ships

These ship together. The migration alone leaves a feature that is hard to operate; a UI rebuild alone has no working data source; SDK work alone fixes delivery for experiments that still cannot be measured or managed; docs written against current behavior go stale immediately. One acceptance bar: a user can run an experiment end to end and trust the result.

## Design

**[Approved UI prototype →](https://claude.ai/design/p/36aed5a6-ca42-4624-b187-a465a23ce909?file=Experiments.dc.html&via=share)**

The prototype is the spec for Part 2. It is interactive and covers all four screens — experiments list, configuration, UVE variant editor, results — including empty states, validation, and the configuration-lock behavior. Use the `startScreen` control to move between screens and `lockRunningConfig` to see the lock.

Two things it settles, rather than repeating them as a task list here:

- **Configuration locking is one rule.** When status is `RUNNING`, `ENDED`, or `ARCHIVED`, goal, variants, traffic, and scheduling lock as a set. Running is locked until stopped; ended and archived are read-only.
- **Four goal types, no more.** Bounce rate, exit rate, reach page, URL parameter. Reach-page and URL-parameter reveal a conditions block. The "coming soon" placeholders get deleted, not built.

This is a **redesign of existing capability**, not new behavior. Variant promotion, archive/restore, the page picker, weight-split editing, and the Bayesian result values all exist today. The work is rebuilding the surface, not inventing what sits behind it. The one genuinely new screen is the **cross-page experiments list** — today entry is per-page only (`/:pageId`), so there is no way to see which pages have experiments in draft, scheduled, running, or finished states.

## Background

Experiments evaluates every goal by querying **CubeJS**, which is being retired in favor of the CAEM + ClickHouse pipeline delivered by #35048. This is a dependency deadline, not a product preference: when CubeJS is removed, experiment reporting stops working.

Three other problems compound:

- **Reporting is slow or fails.** Reports for in-progress and completed experiments can take a long time to load, and in some environments do not load at all.
[...]

PR #37135, opened 2026-08-20
feat(experiments): view results screen (/experiments/:id/results)

### LINKED ISSUE #37004: Experiments Portlet — Screen 3: View Results
## Description

The Results screen on the portlet's stack: stat strip, Daily/Bayesian tabs, summary table, promote. The current reports screen is the cheapest to port — its store is already page-independent (state is only `{experiment, status, results}`, loaded by `experimentId` alone) — but it is rewritten on Signal Store Events like the rest of the portlet, and the charts/data plumbing are **reused, not reinvented**.

Heads-up before starting: experiment results still go through `CubeJSClient` on `main` (2 CubeJS round-trips + a 1000-sample Monte Carlo per call, `@NoCache`) — #36763 Part 1 migrates that to CAEM + ClickHouse. Confirm with the analytics team whether to build against the current contract or wait for the new one; the result model (`ExperimentResults`, `GoalResults`, `VariantResults`) is unchanged by the migration, so building now is expected to be safe.

Design: [approved prototype](https://claude.ai/design/p/36aed5a6-ca42-4624-b187-a465a23ce909?file=Experiments.dc.html&via=share), `results` screen.

### Scope

- **Route**: `/experiments/:experimentId/results` (placeholder wired by #36989).
- **Header**: back to the list · name + status tag · subline `{pageTitle} · {pagePath} · {n} Variants` · `Configuration` button (→ #37003's screen) · `Stop Experiment` when RUNNING.
- **Stat strip**: Winner (`ENDED`) / Leading Variant (otherwise) with inline `Promote` when the leader is not the control · Goal name · Period · Sessions To Date. Ported: **"no winner yet" negative state** (`block` icon + legend — the design has no negative state) and the **refresh-results button** (results are `@NoCache`; users watch this screen while the test runs).
- **Tabs**: Daily Results (conversion rate by day, one series per variant, interactive legend) · Bayesian Results (posterior distribution). **Reuse the existing Chart.js setup** (`dot-experiments-reports-chart/chartjs/` chart options) — do not hand-roll the prototype's SVG. **Bayesian summary data comes from the backend** (`bayesianResult` in the results payload): conversion rate, credibility interval, probability and risk per variant. The posterior *curve* is sampled on the client. `BayesianResult.distributionPdfs` can supply it, but it is gated behind `INCLUDE_BETA_DISTRIBUTION_SAMPLES` (off by default) and carries `BETA_DISTRIBUTION_SAMPLE_SIZE` points per variant — a thousand out of the box — on an endpoint that is `@NoCache`. Sampling a Beta locally costs less than putting those points on the wire on every read.
- **Summary table** (under both tabs): Variant (dot, name, `CONTROL`/`LEADING` chips) · Sessions · Conversions · Conversion Rate · **Lift vs Original** (percentage points, `+11.7 pts`, green/red, `—` on control — the one design addition over the current table) · Probability To Be Best · Conversion Rate Range (95%) · Promote. Ported: **minimum-10-sessions gate** (below it, empty state instead of meaningless rates) and **promoted state** (Promote buttons hidden once a variant is promoted; a variant must not be promotable twice).
- **Promote confirm**: promoting while RUNNING **auto-ends the experiment** (`ExperimentsAPIImpl.java:1392`) — the confirm must say so; the design draws a bare button.
- **Analytics health gate on this route only**: reuse `dotAnalyticsHealthCheckResolver` from `@dotcms/ui` (not the edit-page-coupled `AnalyticsAppGuard`). The list must not be gated.
- **Store**: Signal Store Events (`experiments-results.events.ts`), load/refresh/promote as Request → Succeeded → Failed triples.
- **Unhide from Screen 1**: the `View Results` primary row action in #36989 routes here.

## Acceptance Criteria

- [ ] `/experiments/:id/results` renders stat strip, tabs and summary per the design for RUNNING and ENDED experiments.
- [ ] Daily and Bayesian charts render with the existing Chart.js config; legend toggles series.
- [ ] `Lift vs Original` shows points vs the control row, green/red, `—` on control.
- [ ] Below 10 sessions the summary shows the empty state.
- [ ] Refresh re-fetches results without a full navigation.
- [ ] No suggested winner → the negative state renders, not a false "Leading Variant".
- [ ] Promote asks confirmation; RUNNING copy states the experiment will be ended; after promoting, Promote buttons disappear and the promoted row is marked.
- [ ] A misconfigured analytics app blocks only this route, not the list.
- [ ] The old UVE reports screen is untouched.
- [ ] Jest specs for the results store and summary table.

## Priority

Medium — independent of #37003; can run in parallel once #36989's shell exists.

## Additional Context

- Depends on: #36989. Independent of #37003.
- Out of scope: any change to the results backend contract; charts in the list; `bayesHeadline`/`bayesNote` narrative copy (present in prototype logic, unrendered — confirmed dropped).
- Port sources: `dot-experiments-reports/` (store shape `dot-experiments-reports-store.ts:51-55,208`, summary, details table, `chartjs/`).
- E2E (empty state, 10-session gate, promote confirm) lives in the dedicated E2E issue.

oidacra:
## What

Screen 3 of the Experiments portlet: the Results screen on `/experiments/:experimentId/results` — header, stat strip, Daily/Bayesian tabs and summary table, with promote and stop.

Closes #37004

> **Stacked on #37064.** Base is `oidacra/experiments-portlet-screen-2-create-update-exper`, so this diff shows only screen 3. Retarget to `main` after #37064 merges.

## Reuse over reimplementation

The charts, the legend plugin, the headless table and the winner/promoted helpers already existed under `old/`, which is frozen and slated for deletion. Rather than duplicating them, they are **relocated** to `shared/` and `old/` is repointed at the new paths — one source of truth, and `old/` stays deletable. Only import lines change under `old/`; git records the moves as R100/R099/R097, and `old/`'s specs pass unmodified.

The approved prototype drives visual and behavioural detail only.

## Notable decisions

- **The load is sequential, not `forkJoin`.** DRAFT and SCHEDULED settle on `getById` alone and never reach `getResults`, which is uncached and costs two analytics round-trips plus a Monte Carlo run. The branch reads `experiment.status`, not whether `results` is null.
- **A lost report does not cost the screen.** When `getById` succeeds and `getResults` fails, the header, goal and schedule still render and the failure is reported inline. The full error card is reserved for a missing experiment. Blanking here regressed against the screen this replaces — reproduced against a CubeJS schema without the `Events` cube, where `/results` 400s while `getById` returns 200.
- **The winner comes from the backend.** The prototype picks the leader by highest conversion rate with no significance gate, so it always names one. Using `bayesianResult.suggestedWinner` is what lets the "no winner yet" state exist at all.
- **Exactly one chart is mounted at a time.** The legend plugin resolves its container by walking up from the canvas and calling `querySelector`, which from a shared ancestor returns the first match in document order — two mounted charts would silently claim each other's legend, with no thrown error. `@if`/`@else` makes that impossible.
- **The session gate is experiment-wide.** Below ten total sessions the whole summary is one empty state; above it every row shows full data regardless of its own count.
- **Both Promote entry points confirm.** Promoting a RUNNING experiment ends it server-side in the same call (`ExperimentsAPIImpl.java:1409`), so the strip's inline shortcut cannot be the one path that skips the dialog.
- **Lift is keyed, not positional.** The baseline is resolved by key and `isControl` by name, so row order cannot change the arithmetic.

## Verification

| Check | Result |
|---|---|
| `nx test portlets-dot-experiments-portlet` | 52/52 suites, 914/914 tests |
| `tsc --noEmit` | 0 errors in the portlet |
| `nx affected -t lint` | pass |
| `nx format:check` | clean |

Four projects (`dotcms-js`, `block-editor`, `dotcms-block-editor`, `dotcms-webcomponents`) fail `nx affected -t test` for reasons unrelated to this branch: the only `dotcms-models` change here is 8 purely additive lines, `block-editor` has no dependency on experiments models, and `dotcms-js` fails on `Cannot find module 'karma'`.

Exercised in the browser on the routed screen: header, stat strip, tabs and summary render; tabs switch with a single legend container; the negative winner state renders; goal and period read off the experiment when the report is missing.

## Acceptance criteria

- [x] Route renders stat strip, tabs and summary for RUNNING and ENDED
- [x] Daily and Bayesian charts use the existing Chart.js config; legend toggles series
- [x] `Lift vs Original` in points vs the control, green/red, em dash on control and on a zero-rate control
- [x] Experiment-wide 10-session gate replaces the whole summary
- [x] Refresh re-fetches without a full navigation
- [x] No suggested winner renders the negative state, not a false Leading Variant
- [x] Promote confirms, states the auto-end while RUNNING, then hides every Promote and chips the promoted row
- [x] Stop Experiment confirms and re-renders the header as ENDED in place
- [x] A misconfigured analytics app blocks only this route
- [x] The old UVE reports screen is untouched
- [x] Jest specs for the results store and the summary table

## Not in this PR

- The `htmlLegendPlugin` could not be exercised end to end: without results data there is no canvas. Needs an environment with analytics data, or #36763's migration.
- New `Language.properties` keys need a dotCMS restart to load; the raw key shows until then.
- E2E (empty state, session gate, promote confirm) lives in the dedicated E2E issue.

This PR fixes: #37004


core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-list/dot-experiments-list.component.ts
```diff
@@ -432,6 +442,17 @@ export class DotExperimentsListComponent {
         this.#router.navigate(configureCommandsOf(experiment.id));
     }
 
+    /**
```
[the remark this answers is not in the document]
oidacra: Delete is still in the row menu — it moved rather than being removed.

`40da392b0a` puts it at the end of `#buildRowMenuItems`, behind a rule, so it is never hit while reaching for the entry above it. The rule carries delete's own gate, so an experiment that cannot be deleted does not end its menu on a dangling line.

```
dot-experiments-list.component.ts
552:  { separator: true, visible: isAllowed('delete', status) },
554:    id: 'experiments-delete',
563:      accept: () => this.#dispatch.deleteExperiment(experiment)
```

`deleteExperiment` is reachable from the UI, and the move is covered in both directions:

- `should close the kebab with delete behind a rule for DRAFT` / `SCHEDULED`
- `should leave no dangling rule when delete is not offered for RUNNING` / `ENDED` / `ARCHIVED`
- `should confirm DRAFT / experiments-delete before dispatching`

The finding says to ignore it if the removal was deliberate. It was, so closing this one out.

dotCMS-Machine-User (dismissed):

adrianjm-dotCMS (dismissed):

dotCMS-Machine-User (dismissed):

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-configure/components/dot-experiments-add-variant-dialog/dot-experiments-add-variant-dialog.component.html
```diff
@@ -14,14 +14,14 @@
...
                 {{
                     'experiments.configure.variants.add-dialog.name.max-length'
                         | dm: [maxNameLength.toString()]
                 }}
             </small>
         } @else {
-            <small class="text-gray-500" data-testid="add-variant-name-hint">
+            <small class="p-field-hint" data-testid="add-variant-name-hint">
                 {{ 'experiments.configure.variants.add-dialog.name.hint' | dm }}
             </small>
```
nicobytes, line 26: `p-field-error` / `p-field-hint` are not documented in STYLING_STANDARDS; the styles live in `.form .p-field-*` in `style.css`. In this same PR the change-page dialog uses `p-error`. Should we settle on `p-field-*` (paired with `.form`) or on PrimeNG's `p-error`?

core-web/libs/portlets/dot-experiments/portlet/src/lib/shared/ui/dot-experiments-reports-chart/dot-experiments-reports-chart.component.ts
```diff
@@ -0,0 +1,128 @@
...
+
+@Component({
+    selector: 'dot-experiments-reports-chart',
+    imports: [ChartModule, SkeletonModule, TagModule, DotMessagePipe],
+    templateUrl: './dot-experiments-reports-chart.component.html',
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    host: {
+        class: 'w-full'
+    }
+})
```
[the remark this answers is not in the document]
oidacra: Fixed in `a9791ec8d1`. `changeDetection: ChangeDetectionStrategy.OnPush` and the now-unused `ChangeDetectionStrategy` import are gone from the five components this PR writes, along with `standalone: true` — including the two host components in the `dotFieldRequired` spec that the separate finding covers.

Two files were left as they are, deliberately:

- `libs/ui` carries about fifty components with the same redundancy. None are touched by this PR, so cleaning them here would bury a Results screen change under an unrelated sweep.
- `dot-experiments-details-table` appears in the diff as a pure move out of `old/` — zero lines changed — so it keeps the decorator it arrived with rather than being modernised by accident.

The `standalone: false` in the spec stays: that host is declared through `declarations`, so there it is load-bearing rather than redundant.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-summary-table/dot-experiments-results-summary-table.component.ts
```diff
@@ -0,0 +1,177 @@
...
+            .sort(
+                (first, second) =>
+                    Number(second.id === DEFAULT_VARIANT_ID) -
+                    Number(first.id === DEFAULT_VARIANT_ID)
+            )
+            .map((row, index) => ({
+                ...row,
+                // The control is drawn first, so it always takes the first colour — the same one
+                // the charts give it. Past the palette's length the colours repeat.
+                color: VARIANT_COLORS[index % VARIANT_COLORS.length],
```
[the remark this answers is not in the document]
oidacra: Confirmed and fixed in `a4ecd11ba9`. The palettes really were different: the dot came from `VARIANT_COLORS` (`#0ea5e9`, `#a855f7`, …) while the lines and legend swatches come from `ChartColors` (turquoise, fuchsia, yellow). `VARIANT_COLORS` is the Configure screen's palette — its own doc comment says the Results screen has its own — so the code comment claiming the dot matched the chart was wrong.

Rather than restating the chart's values, the row now reads `RESULTS_VARIANT_COLORS`, derived from `ExperimentChartDatasetColorsVariants`. There is one set of values, so the dot and the line cannot drift apart again.

Covered by `dots each row in the colour its chart line is drawn in`, which asserts against the chart palette itself rather than against copied literals.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/dot-experiments-results.component.ts
```diff
@@ -0,0 +1,278 @@
...
+     * data" panels stacked read as two failures rather than as one Experiment that has simply not
+     * run long enough.
+     */
+    readonly $hasNothingToReport = computed<boolean>(
+        () => !this.store.$hasEnoughSessionsForTable()
+    );
+
+    readonly $emptyReportTitle = this.#dotMessageService.get('experiments.results.empty.title');
+
+    readonly $emptyReportSubtitle = computed<string>(() =>
```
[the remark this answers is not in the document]
oidacra: Confirmed and fixed in `a4ecd11ba9`. The path is exactly as described: a failed `getResults` settles as `LOADED` with `results: null` and `reportUnavailable: true`, which closes the session gate through no fault of the session count, so the body claimed "needs N sessions" directly under a banner saying the report could not be loaded.

`reportUnavailable` now takes precedence over both other reasons, with its own copy.

This also added the first spec for the results shell, covering all three reasons the body can be empty. Both new cases were checked against a deliberately reverted fix to confirm they fail without it.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-summary-table/dot-experiments-results-summary-table.component.ts
```diff
@@ -0,0 +1,177 @@
...
+export class DotExperimentsResultsSummaryTableComponent {
+    readonly store = inject(DotExperimentsResultsStore);
+
+    /**
+     * Rows as they are drawn: the control first, then the variants in the order the results name
+     * them. `$detailData` arrives in `Object.values()` order, which guarantees nothing, and the
+     * sort is presentation only — every row's lift is measured against the control by key,
+     * whichever position it happens to arrive in.
+     */
+    readonly $rows = computed<DotExperimentsSummaryTableRow[]>(() =>
```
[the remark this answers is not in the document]
oidacra: Not actioned: the code this quotes does not exist. The current code indexes by the row's position, `VARIANT_COLORS[index % VARIANT_COLORS.length]`, not by `row.id` — and `row.id` is a variant id string, so `row.id % n` would not compile in the first place. The stated concern about `DotExperimentsSummaryTableRow` missing properties does not occur either; `tsc` and the suite are clean.

The palette itself was a real problem and is fixed in `a4ecd11ba9` — see the thread on the same file.

adrianjm-dotCMS (dismissed):

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/dot-experiments-results.component.scss
```diff
@@ -0,0 +1,27 @@
...
+
+@keyframes dot-experiments-screen-enter {
+    from {
+        opacity: 0;
+    }
+
+    to {
+        opacity: 1;
+    }
+}
```
nicobytes: This can be done with the `tailwindcss-primeui` utilities we already use (`animate-fadein` in edit-content). On the `host` it would be something like `animate-fadein animate-duration-180 animate-ease-out motion-reduce:animate-none`, and this SCSS could go away (the same keyframe is copied in list and configure).

The default `animate-fadein` is 150ms linear; `animate-duration-180` + `animate-ease-out` reproduces what is here today.

core-web/libs/ui/src/lib/components/dot-radio-card/dot-radio-card.component.scss
```diff
@@ -6,11 +6,28 @@
...
+:host {
+    display: block;
+    height: 100%;
+}
+
 .radio-card {
     display: block;
+    height: 100%;
     cursor: pointer;
 }
```
nicobytes: These ones can be utilities: they don't fight PrimeNG. The comment above applies to `p-card` / `.p-card-body` (the `@layer` problem), not to the host or the `<label>`.

`host: { class: 'block h-full' }`, and on the label `class="block h-full cursor-pointer"` (the disabled `cursor: default` is already handled below via `:has(input:disabled)`). The `p-card` selected/hover rules stay in the SCSS.

core-web/libs/ui/src/lib/dot-field-required/dot-field-required.spec.ts
```diff
@@ -54,12 +60,70 @@ describe('Directive: dotFieldRequired', () => {
...
+describe('Directive: dotFieldRequired (outside any form)', () => {
+    it('should mark the label without a FormGroup above it', () => {
+        const fixture = TestBed.createComponent(NoFormHostComponent);
+        fixture.detectChanges();
+
+        expect(hasMarker(fixture.debugElement.query(By.css('[data-testid="plainLabel"]')))).toBe(
+            true
+        );
+    });
+});
```
nicobytes: Spectator is the required harness (`TESTING_FRONTEND`), so this host component + `TestBed.createComponent` shouldn't be needed.

For the bare attribute, `createDirectiveFactory` from `@openng/spectator/jest` (not `@openng/spectator`). For the signal-forms describe below, `createHostFactory` does fit, since that host needs `form()` / `signal()`.

Also `byTestId('plainLabel')` instead of `By.css('[data-testid=…]')`.

nicobytes (commented):
Angular 22 follow-up (on top of the earlier comments). Nothing below is a blocker by itself; the `FieldTree<unknown>` one is a real template type error under `strictTemplates`.

core-web/libs/ui/src/lib/dot-field-required/dot-field-required.directive.ts
```diff
@@ -1,21 +1,54 @@
...
-    private el = inject(ElementRef);
-    private renderer = inject(Renderer2);
-    private formGroupDirective = inject(FormGroupDirective);
+    /**
+     * The signal-forms field this label stands for, when there is one.
+     *
+     * Empty string rather than `undefined` as the default: the bare attribute binds no value, and
+     * the alias makes `dotFieldRequired` both the selector and this input.
+     */
+    readonly $field = input<FieldTree<unknown> | ''>('', { alias: 'dotFieldRequired' });
```
nicobytes: `FieldTree<unknown>` is not a supertype of a real field. `FieldTree<TModel>` is a callable **and** the model's subfield shape, so `unknown` collapses to `{ [Symbol.iterator](): Iterator<[string, never]> }` and a `FieldTree<string>` (e.g. `tree.name` in the spec) does not assign.

The directive only calls `field().required()`, which lives on `FieldState`. Use `Field<unknown>` — same type `dot-radio-card` already takes — and the spec template type-checks under `strictTemplates`.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-summary-table/dot-experiments-results-summary-table.component.ts
```diff
@@ -0,0 +1,176 @@
...
+        ButtonModule,
+        ConfirmDialogModule,
+        SkeletonModule,
+        TableModule,
+        TagModule,
+        DotEmptyContainerComponent,
+        DotMessagePipe
+    ],
+    templateUrl: './dot-experiments-results-summary-table.component.html',
+    // Its own instance, so this confirmation and the shell's Stop confirmation never answer for
```
nicobytes, line 69: On Angular 22 `OnPush` is already the framework default (`ANGULAR_STANDARDS`: do not set `changeDetection` on new components). Same on `dot-experiments-results-charts` and `dot-experiments-results-stat-strip`. Dropping the import and the option keeps these aligned with the results shell / header.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-summary-table/dot-experiments-results-summary-table.component.html
```diff
@@ -0,0 +1,115 @@
...
+                    </div>
+                </td>
+                <td class="text-right!" data-testid="summary-row-sessions">{{ row.sessions }}</td>
+                <td class="text-right!" data-testid="summary-row-conversions">
+                    {{ row.conversions }}
+                </td>
+                <td class="text-right!" data-testid="summary-row-conversion-rate">
+                    {{ row.conversionRate }}
+                </td>
+                <td class="text-right!" [class]="row.liftClass" data-testid="summary-row-lift">
```
nicobytes: `[class]="row.liftClass"` is a full class-string binding on an element that already has `class="text-right!"`. Prefer discrete class bindings so the alignment class cannot be replaced: `[class.text-green-800]="row.liftTone === 'positive'"` (and the other tones), which is also the `[class.x]` form in `ANGULAR_STANDARDS`.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-summary-table/dot-experiments-results-summary-table.component.html
```diff
@@ -0,0 +1,115 @@
...
+                    {{ 'experiments.reports.conversion.rate.range' | dm }}
+                </th>
+                <th class="text-right!" data-testid="summary-header-promote">
+                    {{ 'experiments.reports.promote' | dm }}
+                </th>
+            </tr>
+        </ng-template>
+
+        <ng-template pTemplate="body" let-row>
+            <tr data-testId="detail-row">
```
nicobytes: This file mixes `data-testId` (`loading-skeleton`, `detail-row`, `empty-template`) with `data-testid` on the rest of the table. The frontend testing standard is lowercase `data-testid` — `byTestId()` matches that attribute. Worth making this file consistent so tests do not have to query both.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-stat-strip/dot-experiments-results-stat-strip.component.html
```diff
@@ -0,0 +1,109 @@
...
+                <span
+                    class="text-[15px]/[1.35] font-semibold text-surface-900"
+                    [pTooltip]="$thresholdHintKey() | dm"
+                    data-testid="stat-strip-winner-legend"
+                    tooltipPosition="bottom">
+                    {{ $leadingVariantName() }}
+                </span>
+                @if ($canPromote()) {
+                    <p-button
+                        (onClick)="promoteRequested.emit($suggestedWinner()!.variantName)"
```
nicobytes: The `!` is redundant with the `@if ($canPromote())` above (that computed already requires a suggested winner). Angular 22 `@if` aliasing avoids it: `@if ($suggestedWinner(); as winner)` then `promoteRequested.emit(winner.variantName)` — no non-null assertion in the template.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-charts/dot-experiments-results-charts.component.ts
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        const metric = this.#dotMessageService.get(GOALS_METADATA_MAP[goal.type].label);
+
+        return goal.name ? `${goal.name} · ${metric}` : metric;
+    });
+
+    protected selectTab(tab: string | number): void {
+        this.$activeTab.set(tab as ResultsChartTab);
+    }
```
nicobytes, line 103: `valueChange` is typed `string | number`, so `as ResultsChartTab` will happily write a number into `$activeTab`. Narrow first (`tab === 'daily' || tab === 'bayesian'`) and ignore anything else — the `@if / @else` below assumes the union is closed.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-header/dot-experiments-results-header.component.ts
```diff
@@ -0,0 +1,183 @@
...
+     * screen: a page that cannot be resolved leaves the subline reading the variant count alone
+     * rather than blocking a report that is otherwise complete.
+     */
+    readonly #page = toSignal(
+        toObservable(computed<string | null>(() => this.store.experiment()?.pageId ?? null)).pipe(
+            distinctUntilChanged(),
+            switchMap((pageId) => this.#lookupPage(pageId))
+        ),
+        { initialValue: null }
+    );
```
nicobytes: This is an HTTP call in a presentational header, wired as `toObservable(computed(...)).pipe(switchMap(...))`. On Angular 22 that pattern is what `rxResource` / `resource()` is for, and in this portlet HTTP belongs in the store (`libs/portlets/CLAUDE.md`). A failed lookup already swallows into `null`; keeping it on the store would also mean the header does not need `DotContentSearchService` or the Lucene identifier guard.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-stat-strip/dot-experiments-results-stat-strip.component.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    /** Nothing has been measured yet, so there is no leader to name (AC10). */
+    $isWaitingForData = input<boolean>(false, { alias: 'isWaitingForData' });
+
+    /** A mutation is on the wire; Promote stays closed until it settles. */
+    $isSaving = input<boolean>(false, { alias: 'isSaving' });
+
+    /** Promote was pressed, carrying the id of the variant to publish. */
+    promoteRequested = output<string>();
+
```
nicobytes: `ANGULAR_STANDARDS` marks signal inputs/outputs as `readonly` (`readonly $name = input(...)`). These inputs and `promoteRequested` are writable properties on the class even though `input()` / `output()` already freeze the binding API.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-configure/components/dot-experiments-add-variant-dialog/dot-experiments-add-variant-dialog.component.html
oidacra: Done in dc13482fbe — and you were right that the scope question mattered more than the naming. `p-field-*` describe a *field's* hint and error, so they stay under `.form`; the change-page dialog is not a field, so it uses `p-message severity="error"` instead. STYLING_STANDARDS now has a Form Classes section covering the five classes, and records that `p-error` matches no rule anywhere.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/dot-experiments-results.component.scss
oidacra: Done in adf85220c8. `animate-fadein animate-duration-180 animate-ease-out motion-reduce:animate-none` on the host of all three screens — you were right that the keyframe was copied into list, configure and results. Checked the utilities resolve as expected before swapping: `fadein`, 0.18s, `cubic-bezier(0,0,0.2,1)`, same as the hand-rolled rule. The results stylesheet held nothing else, so it is gone.

core-web/libs/ui/src/lib/components/dot-radio-card/dot-radio-card.component.scss
oidacra: Done in adf85220c8. `host: { class: 'block h-full' }` and `radio-card block h-full cursor-pointer` on the label; only the `p-card` rules stay in the stylesheet, and the comment now says that is what the layering note is about. Confirmed the disabled `cursor: default` still wins — an unlayered component rule beats a utility whatever the specificity.

core-web/libs/ui/src/lib/dot-field-required/dot-field-required.spec.ts
oidacra: Done in adf85220c8 — `createDirectiveFactory` for the bare attribute and `byTestId` throughout. The two form hosts use `createComponentFactory` rather than `createHostFactory`: they are the host, not a component to place in one, so a host template had nothing to wrap.

core-web/libs/ui/src/lib/dot-field-required/dot-field-required.directive.ts
oidacra: Good catch, and it was a real error rather than a style point. `tsc` reports TS2322 on the old signature: `FieldTree<string>` is not assignable to `FieldTree<unknown>`, exactly for the reason you give. The spec never saw it because Jest does not type-check templates. It is `Field<unknown>` now, matching `dot-radio-card`. Fixed in adf85220c8.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-summary-table/dot-experiments-results-summary-table.component.ts
oidacra: Already done in a9791ec8d1, just before this review landed — the option and the import are gone from the five components this PR writes. `libs/ui` carries about fifty more with the same redundancy, and `dot-experiments-details-table` shows in the diff as a pure move, so both were left alone.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-summary-table/dot-experiments-results-summary-table.component.html
oidacra: Done in adf85220c8 — discrete `[class.text-green-800]` bindings per tone. It also left `LIFT_TONE_CLASSES` and the row's `liftClass` with no reader, so those went with it.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-summary-table/dot-experiments-results-summary-table.component.html
oidacra: Done in adf85220c8 — the three `data-testId` in that file are lowercase now, so the whole table matches what `byTestId()` queries.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-stat-strip/dot-experiments-results-stat-strip.component.html
oidacra: Done in adf85220c8 — `@if ($canPromote() && $suggestedWinner(); as winner)`, and the emit reads `winner.variantName`. No assertion left in the template.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-charts/dot-experiments-results-charts.component.ts
oidacra: Done in adf85220c8 — it narrows to the two tabs and ignores anything else, rather than casting into a union the `@if / @else` treats as closed.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-header/dot-experiments-results-header.component.ts
oidacra: Done in adf85220c8 — it is a `pageResolved` event on the store now, resolved off whichever load settled with an experiment, so a failed report still frames itself with its page. The header keeps only the `$subline` computed, and no longer injects `DotContentSearchService` or carries the identifier guard.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-stat-strip/dot-experiments-results-stat-strip.component.ts
oidacra: Done in adf85220c8 — `readonly` on all of them, and on `dot-experiments-reports-chart`'s five while I was there.

nicobytes (dismissed):

dotCMS-Machine-User (approved):

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-summary-table/dot-experiments-results-summary-table.component.ts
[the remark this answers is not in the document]
oidacra: Duplicate of https://github.com/dotCMS/core/pull/37135#discussion_r3928210715, answered there. Fixed in a4ecd11ba9.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/dot-experiments-results.component.ts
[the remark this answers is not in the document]
oidacra: Duplicate of https://github.com/dotCMS/core/pull/37135#discussion_r3928210725, answered there. Fixed in a4ecd11ba9.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-summary-table/dot-experiments-results-summary-table.component.ts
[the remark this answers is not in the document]
oidacra: Duplicate of https://github.com/dotCMS/core/pull/37135#discussion_r3928210715, answered there. Fixed in a4ecd11ba9.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/dot-experiments-results.component.ts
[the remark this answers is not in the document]
oidacra: Duplicate of https://github.com/dotCMS/core/pull/37135#discussion_r3928210725, answered there. Fixed in a4ecd11ba9.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-charts/dot-experiments-results-charts.component.ts
```diff
@@ -0,0 +1,99 @@
+import { ChangeDetectionStrategy, Component, computed, inject, signal } from '@angular/core';
```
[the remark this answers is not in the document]
oidacra: Duplicate of https://github.com/dotCMS/core/pull/37135#discussion_r3936765270, answered there. Fixed in a9791ec8d1.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-empty/dot-experiments-results-empty.component.ts
```diff
@@ -0,0 +1,31 @@
+import { ChangeDetectionStrategy, Component, input } from '@angular/core';
```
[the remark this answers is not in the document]
oidacra: Duplicate of https://github.com/dotCMS/core/pull/37135#discussion_r3936765270, answered there. Fixed in a9791ec8d1.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-stat-strip/dot-experiments-results-stat-strip.component.ts
```diff
@@ -0,0 +1,101 @@
+import { DatePipe, DecimalPipe } from '@angular/common';
+import { ChangeDetectionStrategy, Component, computed, input, output } from '@angular/core';
```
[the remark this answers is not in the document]
oidacra: Duplicate of https://github.com/dotCMS/core/pull/37135#discussion_r3936765270, answered there. Fixed in a9791ec8d1.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-summary-table/dot-experiments-results-summary-table.component.ts
```diff
@@ -0,0 +1,177 @@
+import { injectDispatch } from '@ngrx/signals/events';
+
+import { ChangeDetectionStrategy, Component, computed, inject } from '@angular/core';
```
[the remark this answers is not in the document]
oidacra: Duplicate of https://github.com/dotCMS/core/pull/37135#discussion_r3936765270, answered there. Fixed in a9791ec8d1.

core-web/libs/portlets/dot-experiments/portlet/src/lib/shared/ui/dot-experiments-reports-chart/dot-experiments-reports-chart.component.ts
```diff
@@ -0,0 +1,128 @@
+import { ChartData } from 'chart.js';
+
+import { ChangeDetectionStrategy, Component, computed, effect, input, signal } from '@angular/core';
```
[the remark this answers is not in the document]
oidacra: Duplicate of https://github.com/dotCMS/core/pull/37135#discussion_r3936765270, answered there. Fixed in a9791ec8d1.

core-web/libs/ui/src/lib/dot-field-required/dot-field-required.spec.ts
```diff
@@ -54,12 +60,72 @@ describe('Directive: dotFieldRequired', () => {
...
+ * The bare attribute has to stand on its own: the screens that need it most are the ones with no
+ * `[formGroup]` anywhere above the label.
+ */
+@Component({
+    standalone: true,
+    imports: [DotFieldRequiredDirective],
+    template: `
+        <label data-testid="plainLabel" dotFieldRequired for="plain">Plain</label>
+    `
+})
```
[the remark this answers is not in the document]
oidacra: Duplicate of https://github.com/dotCMS/core/pull/37135#discussion_r3936765270, answered there. Fixed in a9791ec8d1.

core-web/libs/portlets/dot-experiments/portlet/src/lib/store/dot-experiments-results.store.ts
```diff
@@ -0,0 +1,471 @@
...
+                 * both settled loads carry an experiment — a report that failed still frames itself
+                 * with the page it ran on. Ancillary throughout, so a rejected lookup answers `null`
+                 * instead of reaching the error manager.
+                 */
+                resolvePage$: merge(
+                    events
+                        .on(apiEvents.loadSucceeded)
+                        .pipe(map(({ payload }) => payload.experiment)),
+                    events.on(apiEvents.resultsUnavailable).pipe(map(({ payload }) => payload))
+                ).pipe(
```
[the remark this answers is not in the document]
oidacra: Fixed in e71ce2dbc4. The operator is gone. Your reasoning holds: only one of `loadSucceeded`/`resultsUnavailable` fires per load, so it could never dedupe within one — it could only drop the second of two experiments sharing a page, and `enter` has already reset `page` to null by then.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-header/dot-experiments-results-header.component.ts
```diff
@@ -0,0 +1,138 @@
...
+    host: {
+        class: 'flex flex-none items-center justify-between gap-6 border-b border-surface-200 bg-white px-8 py-4'
+    }
+})
+export class DotExperimentsResultsHeaderComponent {
+    readonly store = inject(DotExperimentsResultsStore);
+
+    readonly $title = computed<string>(() => this.store.experiment()?.name ?? '');
+
+    readonly $statusSeverity = computed<TagSeverity>(
```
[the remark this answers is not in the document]
oidacra: Correct, and it was a real compile break: `tsc` reports TS2304 on it. I removed the import myself in the previous commit — a regex that dropped an unused `DotExperimentConfigurePage` took the whole statement, and `TagSeverity` was in it. Neither Jest (isolatedModules) nor ESLint type-checks, which is why the suite stayed green over a file that could not compile. Fixed in e71ce2dbc4.

core-web/libs/portlets/dot-experiments/portlet/src/lib/store/dot-experiments-results.store.ts
```diff
@@ -0,0 +1,471 @@
...
+                    map(({ pageId }) => pageId),
```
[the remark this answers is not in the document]
oidacra: Same finding as https://github.com/dotCMS/core/pull/37135#discussion_r3937394990, answered there. Fixed in e71ce2dbc4.

core-web/libs/portlets/dot-experiments/portlet/src/lib/shared/dot-experiment-results.utils.ts
```diff
@@ -0,0 +1,253 @@
...
+    if (!bayesianResult || bayesianResult.suggestedWinner === BayesianStatusResponse.NONE) {
+        return [];
+    }
+
+    // Iterate through all the variants
+    return Object.entries(variants).map(([variantId, variant], index) => {
+        // Calculate the number of successes and failures
+        const success = variant.uniqueBySession.count;
+        const failure = sessions.variants[variantId] - variant.uniqueBySession.count;
+        const label = variant.variantDescription;
```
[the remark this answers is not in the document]
oidacra: Fixed in e71ce2dbc4 — it uses `orderVariants` now, as `buildDailyChartData` already did. This also mattered for a second reason: the summary-table dot was changed in a4ecd11ba9 to take the chart palette so a row's dot matches its line, and the unordered Bayesian datasets undid that on the second tab.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-results/components/dot-experiments-results-header/dot-experiments-results-header.component.ts
```diff
@@ -0,0 +1,138 @@
...
+import { ConfirmationService } from 'primeng/api';
+import { ButtonModule } from 'primeng/button';
+import { TagModule } from 'primeng/tag';
+import { TooltipModule } from 'primeng/tooltip';
+
+import { DotMessageService } from '@dotcms/data-access';
+import { DotExperimentStatus } from '@dotcms/dotcms-models';
+import { DotMessagePipe } from '@dotcms/ui';
+
+import {
```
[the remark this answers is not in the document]
oidacra: Same finding as https://github.com/dotCMS/core/pull/37135#discussion_r3958664304, answered there. Fixed in e71ce2dbc4.

core-web/libs/portlets/dot-experiments/portlet/src/lib/store/dot-experiments-results.store.ts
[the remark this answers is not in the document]
oidacra: Same finding as https://github.com/dotCMS/core/pull/37135#discussion_r3937394990, answered there. Fixed in e71ce2dbc4.

adrianjm-dotCMS:
PR Approved ✅

PR #37137, opened 2026-08-20
feat(ci): manual changelog site publish + fix release-notes auto-detect

### LINKED ISSUE #37136: Changelog site publish has no manual trigger, so a missed release requires local credentials to fix
## Problem

`cicd_comp_changelog-site-publish-phase.yml` declares `workflow_call:` only. When the release pipeline skips it — which happens whenever an upstream job fails or is cancelled, since the job is gated on `needs: [..., release-notes]` + `success()` — there is no way to publish that release's changelog from CI.

The only remaining path is running the publisher locally, which requires an operator to know and obtain three values that otherwise never leave GitHub:

- `DOTCMS_DEVSITE_URL` (repo variable)
- `DOTCMS_DEVSITE_RELEASENOTES_TOKEN` (repo secret, Keeper)
- `DOTCMS_DEVSITE_RELEASENOTES_ACCOUNT` (repo variable)

This came up on 26.08.19: four release attempts, three of which shipped Docker images with no changelog at all, and the fourth published a changelog covering 1 of 19 commits. Recovering it needed a local checkout plus credentials from Keeper.

Related: the phase's own comments describe `--force` as "a manual operator re-run only", but no manual re-run path exists — the flag is unreachable from CI.

## Proposal

Add `cicd_manual_changelog-site-publish.yml`, a `workflow_dispatch` wrapper around the existing reusable phase — the same pattern `cicd_ai-release-notes-backfill.yml` already uses for the release-notes phase. Credentials stay in repo vars/secrets; an operator supplies only the release tag, from the Actions UI or `gh workflow run`.

The wrapper derives what the pipeline would have passed:

- `release_version` from the tag
- `docker_tags` — the sha-tagged image, looked up from Docker Hub's public API, so a manual entry matches a pipeline-written one
- `released_date` from the release's own `publishedAt`, so backfilling an older release does not stamp today's date

Two small additions to the phase workflow, both defaulting to current behavior so the release pipeline is unchanged:

- `force` (boolean, default false) — threads `--force` through for the documented override
- `released_date` (string, default empty) — falls back to `$(date -u +%F)` exactly as today

## Out of scope

The auto-detect bug that caused the wrong 26.08.19-04 changelog (`findPreviousTag()` picking the immediately-preceding tag) is a separate issue.

### LINKED ISSUE #37138: Release-notes auto-detect picks an undocumented tag, producing a changelog that covers a fraction of the release
## Problem

`findPreviousTag()` (`.github/scripts/gather-release-data/src/github.ts:57`) returns `tags[idx + 1]` unconditionally — the tag immediately preceding the target in the releases list. On a day with more than one release attempt, that preceding tag is another attempt at the *same* release, whose pipeline died before writing any notes.

The result is a changelog whose range is one attempt wide instead of one release wide.

**Observed on 26.08.19.** Four attempts were cut. `-01`, `-02` and `-03` all pushed real Docker images but died in Deployment, so their release bodies are the empty string. `-04` succeeded, auto-detect resolved its previous tag to `v26.08.19-03`, and the generated changelog described 1 of the 19 commits in the release — publishing to dev.dotcms.com that a release containing Accessibility Studio, the Experiments portlet, three roles endpoints and nine fixes "contains internal maintenance only."

That is worse than a missing changelog: it is an affirmatively false "nothing changed" statement that customers may act on when deciding whether to upgrade.

## Root cause

Three compounding defects in the same two functions:

1. **No notion of whether a predecessor was ever documented.** An empty-bodied release is treated as a valid range boundary, which strands every commit behind it.
2. **`listStandardReleaseTags()` does not filter drafts.** `listReleases` returns drafts first regardless of date, so drafts corrupt the newest-first ordering the function's own docstring promises. Six draft releases currently match `STANDARD_RELEASE_PATTERN`.
3. **`indexOf` on a duplicated tag returns the first match.** One of those drafts duplicates `v26.04.11-02`, so the index — and therefore the resolved `fromTag` — can be wrong. Latent today; filtering drafts closes it.

## Proposed fix

Carry per-release notes state alongside the tag, skip drafts when listing, and walk back past undocumented releases when resolving:

- `listStandardReleaseTags()` returns `{ tag, hasNotes }[]` and skips drafts
- `findPreviousTag()` walks backwards to the first entry with `hasNotes`, so an undocumented attempt's commits stay inside the range rather than being stranded behind its tag

Same-day attempts that *did* complete keep working exactly as they do now — the skip is conditioned on missing notes, not on the date.

## Workaround until fixed

Always pass `previous_tag` explicitly to `cicd_ai-release-notes-backfill.yml`. Never rely on auto-detect after a day with more than one attempt.

sfreudenthaler:
Closes: #37136
Closes: #37138

Two related changes: a manual publish path, and the auto-detect bug that made one necessary.

## 1. Manual trigger for changelog site publish (#37136)

Adds `cicd_manual_changelog-site-publish.yml`, a `workflow_dispatch` wrapper around the existing phase — the pattern `cicd_ai-release-notes-backfill.yml` already uses.

```bash
gh workflow run cicd_manual_changelog-site-publish.yml --repo dotCMS/core --ref main \
  -f release_tag=v26.08.19-04
```

**Why:** the phase is `workflow_call:` only and gated on `success()`, so any upstream failure or cancellation skips it with no CI path to recover. Today that means a local checkout plus three values that otherwise never leave GitHub — `DOTCMS_DEVSITE_URL`, `DOTCMS_DEVSITE_RELEASENOTES_TOKEN`, `DOTCMS_DEVSITE_RELEASENOTES_ACCOUNT`. The phase's comments also describe `--force` as "a manual operator re-run only", but no manual re-run path existed, so the flag was unreachable.

The wrapper derives what the pipeline would have passed:

| Input | Derived from |
|---|---|
| `release_version` | the tag, minus the `v` |
| `docker_tags` | Docker Hub public API, preferring the `version_sha` tag so a manual entry matches a pipeline-written one |
| `released_date` | the release's own `publishedAt`, so backfilling an older release does not stamp today's date |

The phase gains `force` and `released_date`, both defaulting to current behavior. **The release pipeline path is unchanged** — `cicd_6-release.yml` sets neither input.

### Access control

`workflow_dispatch` is limited to accounts with write access — never forks, never the public — but that is 45 accounts today, and this writes to the public docs site with a service-account token. So the publish job sits behind a no-op approval gate on the `changelog-site-publish` environment (`required_reviewers: dotDevelopers`), matching `cicd_evergreen-tracks-promote.yml`'s apply gate.

The environment has been created with that protection rule. It must keep it: GitHub auto-creates a missing environment **unprotected** on first use, which would silently remove the gate.

## 2. Auto-detect picks an undocumented tag (#37138)

`findPreviousTag()` returned `tags[idx + 1]` unconditionally. On a multi-attempt day that is another attempt at the same release, whose pipeline died before writing notes — so the changelog covers one attempt instead of the release.

This is what produced the 26.08.19-04 changelog: auto-detect resolved its predecessor to `v26.08.19-03`, and the release notes described 1 of 19 commits, publishing "contains internal maintenance only" for a release carrying Accessibility Studio, the Experiments portlet, three roles endpoints and nine fixes.

Three fixes in the same two functions:

- `listStandardReleaseTags()` returns `{ tag, hasNotes }` and **skips drafts**. `listReleases` returns drafts first regardless of date, corrupting the newest-first ordering the docstring promises — 6 drafts currently match the standard pattern, and one duplicates `v26.04.11-02`.
- `findPreviousTag()` walks back to the first release with notes, so an undocumented attempt's commits stay inside the range instead of being stranded behind its tag.
- `findIndex` on tag equality replaces `indexOf`, which returned the first match for a duplicated tag.

A same-day attempt that *did* publish notes is still a valid boundary — the skip is conditioned on missing notes, not on the date, and there is a test for exactly that (26.08.12-02 → 26.08.12-01).

## Verification

- `npx tsc --noEmit` clean; `npx jest` 35/35 passing, including three new `findPreviousTag` cases: skips undocumented predecessors, keeps a documented same-day attempt, returns undefined when every predecessor is undocumented.
- Both workflows parse as valid YAML.
- The resolve logic was run against a real tag and reproduces the manual publish exactly: docker tag → `26.08.19-04_a0181f9`, released date → `2026-08-20`, both matching the entry now live on the site.
- Draft counts above were measured against the live API, not assumed.

Not exercised end-to-end against the live site: the only release currently needing a publish was already published by hand, and re-running would trip human-edit protection. A run on the next release that needs it is the real test.

### DIFF
--- a/.github/scripts/gather-release-data/src/github.test.ts
+++ b/.github/scripts/gather-release-data/src/github.test.ts
@@ -39,7 +39,15 @@ describe('extractPRNumbers', () => {
 });
 
 describe('findPreviousTag', () => {
-  const tags = ['v26.03.13-02', 'v26.03.13-01', 'v26.03.12-01', 'v26.03.11-01'];
+  const documented = (...tags: string[]) =>
+    tags.map((tag) => ({ tag, hasNotes: true }));
+
+  const tags = documented(
+    'v26.03.13-02',
+    'v26.03.13-01',
+    'v26.03.12-01',
+    'v26.03.11-01'
+  );
 
   it('finds the tag immediately before the given tag', () => {
     expect(findPreviousTag(tags, 'v26.03.13-02')).toBe('v26.03.13-01');
@@ -53,6 +61,35 @@ describe('findPreviousTag', () => {
 
   it('returns undefined for the oldest tag', () => {
     expect(findPreviousTag(tags, 'v26.03.11-01')).toBeUndefined();
+  });
+
+  // Regression: 26.08.19 shipped four attempts. -01/-02/-03 died before writing
+  // notes, so stopping at -03 described 1 of 19 commits.
+  it('skips undocumented releases', () => {
+    const withGhosts = [
+      { tag: 'v26.08.19-04', hasNotes: true },
+      { tag: 'v26.08.19-03', hasNotes: false },
+      { tag: 'v26.08.19-02', hasNotes: false },
+      { tag: 'v26.08.19-01', hasNotes: false },
+      { tag: 'v26.08.14-01', hasNotes: true },
+    ];
+    expect(findPreviousTag(withGhosts, 'v26.08.19-04')).toBe('v26.08.14-01');
+  });
+
+  it('still returns a same-day attempt that published notes', () => {
+    const sameDay = [
+      { tag: 'v26.08.12-02', hasNotes: true },
+      { tag: 'v26.08.12-01', hasNotes: true },
+    ];
+    expect(findPreviousTag(sameDay, 'v26.08.12-02')).toBe('v26.08.12-01');
+  });
+
+  it('returns undefined when every predecessor is undocumented', () => {
+    const allGhosts = [
+      { tag: 'v26.08.19-02', hasNotes: true },
+      { tag: 'v26.08.19-01', hasNotes: false },
+    ];
+    expect(findPreviousTag(allGhosts, 'v26.08.19-02')).toBeUndefined();
   });
 });
 
--- a/.github/scripts/gather-release-data/src/github.ts
+++ b/.github/scripts/gather-release-data/src/github.ts
@@ -27,44 +27,60 @@ export function parseRepo(fullRepo: string): { owner: string; repo: string } {
   return { owner, repo };
 }
 
+/** A published standard release, and whether its notes were ever written. */
+export interface ReleaseRef {
+  tag: string;
+  /** False when the release body is empty — a cut whose pipeline died before notes. */
+  hasNotes: boolean;
+}
+
 /**
- * List standard release tags (matching vYY.MM.DD-NN pattern), sorted by
- * creation date descending (newest first).
+ * List standard releases (matching vYY.MM.DD-NN pattern), newest first.
+ *
+ * Drafts are skipped: listReleases returns them first regardless of date, so they
+ * corrupt the ordering this function promises, and a draft may duplicate a real tag.
  */
 export async function listStandardReleaseTags(
   octokit: Octokit,
   owner: string,
   repo: string
-): Promise<string[]> {
-  const tags: string[] = [];
+): Promise<ReleaseRef[]> {
+  const releases: ReleaseRef[] = [];
   for await (const response of octokit.paginate.iterator(
     octokit.repos.listReleases,
     { owner, repo, per_page: 100 }
   )) {
     for (const release of response.data) {
+      if (release.draft) continue;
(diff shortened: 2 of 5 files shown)


sfreudenthaler:
Superseded by stack #37143, which splits this into three reviewable PRs:

- #37140 — auto-detect fix (closes #37138)
- #37141 — manual changelog site publish (closes #37136)
- #37142 — gate the release notes backfill (closes #37139)

Same content, verified tree-identical to this branch before splitting. Closing in favor of the stack.

PR #37141, opened 2026-08-20
feat(ci): add a manual trigger for changelog site publish

### LINKED ISSUE #37136: Changelog site publish has no manual trigger, so a missed release requires local credentials to fix

sfreudenthaler:
Closes: #37136

Stack #37143, middle. Based on #37140.

## What

Adds `cicd_manual_changelog-site-publish.yml`, a `workflow_dispatch` wrapper around the existing changelog site publish phase — the pattern `cicd_ai-release-notes-backfill.yml` already uses for the notes phase.

```bash
gh workflow run cicd_manual_changelog-site-publish.yml --repo dotCMS/core --ref main \
  -f release_tag=v26.08.19-04
```

## Why

The phase is `workflow_call:` only and gated on `success()`, so any upstream failure or cancellation skips it with no CI path to recover. Today that means a local checkout plus three values that otherwise never leave GitHub — `DOTCMS_DEVSITE_URL`, `DOTCMS_DEVSITE_RELEASENOTES_TOKEN`, `DOTCMS_DEVSITE_RELEASENOTES_ACCOUNT`.

That is what 26.08.19 cost: four attempts, three shipping Docker images with no changelog, the fourth publishing a changelog covering 1 of 19 commits. Recovery needed Keeper and a local shell.

The phase's comments also describe `--force` as "a manual operator re-run only", but no manual re-run path existed — the flag was unreachable from CI.

## How

The wrapper derives what the pipeline would have passed:

| Input | Derived from |
|---|---|
| `release_version` | the tag, minus the `v` |
| `docker_tags` | Docker Hub public API, preferring the `version_sha` tag so a manual entry matches a pipeline-written one |
| `released_date` | the release's own `publishedAt`, so backfilling an older release does not stamp today's date |

The phase gains `force` and `released_date`, both defaulting to current behavior. **The release pipeline path is unchanged** — `cicd_6-release.yml` sets neither input.

## Access control

`workflow_dispatch` is limited to accounts with write access — never forks, never anonymous — but that is 45 accounts today, and this writes to the public docs site with a service-account token. So the publish job sits behind a no-op approval gate on the `changelog-site-publish` environment (`required_reviewers: dotDevelopers`), matching `cicd_evergreen-tracks-promote.yml`'s apply gate.

The environment has been created with that rule. It must keep it: GitHub auto-creates a missing environment **unprotected** on first use, which would silently remove the gate — it fails open, not closed.

## Verification

Both workflows parse as valid YAML. The resolve logic was run against a real tag and reproduces the manual publish exactly: docker tag → `26.08.19-04_a0181f9`, released date → `2026-08-20`, both matching the entry now live on the site.

Not exercised end-to-end against the live site: the only release currently needing a publish was already published by hand, and re-running would trip human-edit protection. A run on the next release that needs it is the real test.

### DIFF
--- a/.github/workflows/cicd_comp_changelog-site-publish-phase.yml
+++ b/.github/workflows/cicd_comp_changelog-site-publish-phase.yml
@@ -33,6 +33,16 @@ on:
         description: 'Space-separated docker image tags produced by the deployment phase'
         required: true
         type: string
+      force:
+        description: 'Override human-edit protection and update in place. Manual operator use only — the release pipeline never sets this.'
+        required: false
+        type: boolean
+        default: false
+      released_date:
+        description: 'Availability date (yyyy-MM-dd). Defaults to today, which is correct for a live release; a manual backfill of an older release should pass that release''s own date.'
+        required: false
+        type: string
+        default: ''
       allow_failure:
         description: 'If true, job failures are non-blocking (used by release pipeline). Defaults to false.'
         required: false
@@ -109,10 +119,12 @@ jobs:
           DOTCMS_DEVSITE_URL: ${{ vars.DOTCMS_DEVSITE_URL }}
           # Read once by the tool; never echoed to logs / GITHUB_OUTPUT / step summary.
           DOTCMS_DEVSITE_RELEASENOTES_TOKEN: ${{ secrets.DOTCMS_DEVSITE_RELEASENOTES_TOKEN }}
-          # Service-account modUser id for human-edit protection (FR-011). --force is never
-          # passed here — override is a manual operator re-run only.
+          # Service-account modUser id for human-edit protection (FR-011). The release
+          # pipeline never sets force; only a manual operator re-run does.
           DOTCMS_DEVSITE_RELEASENOTES_ACCOUNT: ${{ vars.DOTCMS_DEVSITE_RELEASENOTES_ACCOUNT }}
           DOCKER_TAGS: ${{ inputs.docker_tags }}
+          FORCE: ${{ inputs.force }}
+          RELEASED_DATE: ${{ inputs.released_date }}
         run: |
           # Pick the sha-tagged dotcms/dotcms image (version_sha form, e.g.
           # dotcms/dotcms:26.07.10-01_84b0486). The `_` in the glob excludes the floating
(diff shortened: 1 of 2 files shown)


sfreudenthaler:
Both findings addressed.

**1. Silent no-op on a missing `v` prefix** — fixed. `resolve` now fails fast:

```bash
case "$RELEASE_TAG" in
  v*) ;;
  *) echo "Error: release_tag must start with 'v' (got '$RELEASE_TAG')"; exit 1;;
esac
```

Good catch, and the failure mode was the bad kind: `resolve` succeeds, a human approves at the gate, `publish` is skipped by the phase's `startsWith` guard, and the run goes green having published nothing. It now dies before anyone is asked to approve. Verified both branches of the case statement.

**2. Environment provisioning** — the review assumed it is not yet provisioned; it is. `changelog-site-publish` was created before this PR was opened, and re-verified just now via `gh api /repos/dotCMS/core/environments/changelog-site-publish`:

```
required_reviewers -> dotDevelopers
```

Same for `release-notes-backfill` in #37142. This is in the PR description, but worth restating here since it cannot be checked from the diff.

The caveat in the code comment stands regardless: if either environment is ever deleted, GitHub recreates it unprotected on first use and the gate silently disappears. It fails open, and no code in this repo can defend against that.

sfreudenthaler:
Bumping — this and #37142 have been clean and awaiting review since 2026-08-22. No conflicts, no failing checks, `mergeable: MERGEABLE`; the only blocker is an approval. @dcolina you approved the base of this stack (#37140), so tagging you.

Context that raises the priority since this was opened: squash merging is now disabled on the repo, and the release pipeline's PR-number resolution was silently broken by it (#37201 / #37213 — 5 of 13 numbers on the last release cut were issue numbers that 404). Separately, #37317 tracks a report that changelog entries have been missing from dev.dotcms.com. This PR's manual `workflow_dispatch` trigger for the changelog site publish phase is exactly the recovery path that gap needs — the phase is `workflow_call:`-only and gated on `success()`, so today there is no CI route to republish a version whose publish was skipped.

Issue #37008, opened 2026-08-11
Experiments Portlet — migration: delete the old per-page experiments UI and remove the flag


oidacra:
## Description

The epic's strategy is: build the new portlet complete, then migrate and delete the old one. This is the issue where the duplication ends. It is filed up front — together with the rest of the epic — so "delete old later" is a scheduled step, not a wish that never happens.

This is deliberately the only issue that rewrites/removes existing code at scale.

**Entry condition:** every "port" row of the functional-parity inventory verified working in the new portlet on the dev/QA instance, and the E2E suite (#37006) green. Nothing is deleted before its replacement is verified.

### Delete

- The old per-page UI in `libs/portlets/dot-experiments/portlet/src/lib/`: `dot-experiments-shell/`, `dot-experiments-list/` (including the drawer create component), `dot-experiments-configuration/`, `dot-experiments-reports/`, `dot-experiments-analytic-app-misconfiguration/`, the old `shared/` tree, and the old `lib.routes.ts` (`dotExperimentsRoutes`).
- With them go: the three `@ngrx/component-store` stores, the 10 hardcoded `router.navigate(['/edit-page/…'])` calls, and the `window.location.href`-parsing `getUrl()` — no decoupling refactor needed; the code that holds them disappears.

### Rewire

- UVE entry point permanently targets the new flow; **remove `FEATURE_FLAG_EXPERIMENTS_PORTLET`** — the entry-point flag introduced by #37005 — its `withFlags` wiring, its `FeaturedFlags` entry, its `ConfigurationResource` allow-list entry, and its explicit `false` in `dotmarketing-config.properties`.
- **Do not remove `FEATURE_FLAG_EXPERIMENTS`.** Despite the name, it is not a frontend flag: it is the backend kill-switch for the whole Experiments feature, read by `ConfigExperimentUtil.isExperimentEnabled()` and gating experiment JS injection (`ExperimentWebAPIImpl:338`) and experiment resolution during page render (`HTMLPageAssetRenderedAPIImpl:176,239`). Removing it would stop experiments from serving to visitors. It keeps its name, meaning, default and consumers. The unused enum entry `FeaturedFlags.LOAD_FRONTEND_EXPERIMENTS = 'FEATURE_FLAG_EXPERIMENTS'` (`shared-models.ts:28`) never gained a consumer and *can* be deleted here — the property cannot.
[...]

Issue #37007, opened 2026-08-11
Experiments Portlet — swap the List to the server-side contract


oidacra:
## Description

#36989 ships the list against `GET /api/v1/experiments` as it exists today, computing paging, sorting, filtering, status counts and site scoping client-side — an explicit, documented interim. This issue replaces that interim with the server-side contract from #36823, and is **filed together with #36989 on purpose** so the interim cannot quietly become permanent. It lands whenever #36823 does; nothing else waits on it.

The URL param contract was designed to be final from day one, so this swap changes **where** the params are applied, not what they are called.

### Scope

- Replace client-side paging / sorting / text filtering / status filtering / site filtering / status counts with the server-side parameters from #36823 (`page`, `per_page`, `orderby`, `direction`, `filter`, `status`, `goal`, `siteId`).
- Add two new toolbar filter controls — filter by creator and filter by running date range — consuming #36823's `created_by` and `running_from`/`running_to` parameters.
- Consume the enriched rows: real `pageTitle` / `pageUrl` / site in the Page column, deleting #36989's interim client-side page lookup.
- Consume the per-status counts from the API instead of counting the loaded array.
- `totalRecords` from the response body (`pagination.totalEntries`), not the `X-Pagination-*` headers.
- Enable server-side filtering only against a build where #36823's two SQL fixes are merged (`name`+`status` malformed SQL; `pageId`+multi-status precedence).

### Boundary check

Only two places should change in existing code: the data-access service and the store's load event handler. **No existing component should need touching** — the one enumerated exception is the two new toolbar filter controls (creator, running date range), which are new components this issue adds. If any other component needs a change, #36989's service boundary was drawn wrong — flag it in review instead of patching around it.

## Acceptance Criteria

- [ ] The list issues one request per view change carrying the server-side params; no client-side slicing/sorting/filtering remains (interim utilities deleted).
- [ ] URL behavior is byte-identical for all pre-existing params: same names, same defaults-omitted rule, same deep-link restoration (the E2E deep-link spec from #37006 passes unchanged). The three new params (`created_by`, `running_from`, `running_to`) are additive and follow the same defaults-omitted rule.
- [ ] Page column shows the enriched title/path; the interim lookup is deleted.
- [ ] Status counts come from the API and reflect the whole filtered set.
- [ ] Site scoping uses the `siteId` param; switching site refetches.
- [ ] Diff touches only the data-access service, the store load handler, the two new toolbar filter controls, and deleted interim code.
[...]

Issue #37006, opened 2026-08-11
Experiments Portlet — E2E test suite (incl. flag-off regression of the old flow)


oidacra:
## Description

There is **no E2E coverage for experiments at all today** — which also means the migration issue (deleting the old UI) has no safety net without this. Build the Playwright suite for the whole portlet, plus the regression proving the old UVE flow is untouched while `FEATURE_FLAG_EXPERIMENTS` is off.

Playwright specs in `core-web/apps/dotcms-ui-e2e` (page objects + fixtures per screen), covering:

### List (#36989)
- Site-wide list renders; search by name and page path; status chip filtering with counts; ARCHIVED hidden by default, visible when selected.
- Pagination and column sort (default Modified DESC).
- **Deep-links**: a URL carrying `filter` + `status` + `page` + `orderby` restores the exact view; back/forward walks the states; a bare `/experiments` carries no params.
- Row actions per status (kebab gating follows `AllowedActionsByExperimentStatus`): archive, restore, cancel schedule, end, abort, delete — each with confirm and toast.

### Create/Update (#37003)
- `/experiments/new`: create by typing a name and selecting a page (autosave POST, URL becomes `/:id/configuration` without a history entry).
- `?url=/index` preselects the page; invalid value shows a message.
- Goal/variants/scheduling edits persist across a reload (autosave PATCH).
- Start with missing fields reveals the 8 validation rules and scrolls to the first error.
- Start → RUNNING (no date) / SCHEDULED (with date); Stop → ENDED; locked state disables the form.

### Results (#37004)
- Stats, tabs and summary for a RUNNING and an ENDED experiment.
- Below 10 sessions the table shows the empty state.
- Promote: confirm dialog (RUNNING copy mentions auto-end), promoted state hides Promote buttons.

### UVE integration (#37005)
- Flag ON: variant `Edit Content` round-trip — portlet → UVE → back to `/experiments/:id/configuration`, variant marked as edited.
- **Flag OFF regression: the old per-page UVE experiments flow works byte-identically** — nav item, per-page list, configure, reports.

### Fixtures
- Seeded experiments per status (DRAFT/SCHEDULED/RUNNING/ENDED/ARCHIVED) via the REST API in test setup; the manual `cms_layouts_portlets` registration scripted in the suite's global setup.

## Acceptance Criteria

- [ ] All scenarios above implemented and green in headless mode.
- [ ] The deep-link spec fails if URL state stops round-tripping.
[...]

Issue #37005, opened 2026-08-11
Experiments Portlet — UVE integration behind FEATURE_FLAG_EXPERIMENTS_PORTLET + variant Edit Content round-trip


oidacra:
## Description

The one place where the new flow overlaps something customers use today is the experiments entry point inside the Universal Visual Editor. This issue (a) completes the Variants card of #37003 with the Edit Content round-trip into UVE, and (b) switches the UVE entry point to the new portlet flow **behind a feature flag, off by default** — the old per-page screens keep working unchanged in every build until the flag is flipped, and it can be flipped back.

This is the only issue before the migration that modifies existing behavior.

### 1. Variant Edit Content round-trip

- `Edit Content` on a variant navigates to UVE for `{pageId, variantName}` using the existing deep-link shape: `/edit-page/content?url=…&variantName=…&experimentId=…` (current construction: `dot-uve-toolbar.component.ts:262-277`).
- Build the URL from the experiment's page data in the store — do **not** copy the current `getUrl()` that parses `window.location.href` (`dot-experiments-configuration-variants.component.ts:187-234`).
- Returning from UVE lands back **in the portlet**, on `/experiments/:id/configuration` for the experiment the editor came from — resolved by experiment identity, not by page, since a page can host more than one experiment.
- Control variant and locked experiments open UVE read-only (`Preview`), per the design.
- **No "edited" indicator.** No edited/revision signal exists on the variant model on either side of the wire (`AbstractExperimentVariant`, `Variant`), and the legacy per-page Variants card shows no edit indication at all. Do not invent one — not persisted, not derived, not session-held. The deliverable is the correct round-trip, not a badge.
- One consequence: the new Variants card renders the fixed string `No content changes yet` for every non-control variant (`experiments.configure.variants.meta.variant`). That is true today only because the Edit Content button beside it is disabled. This issue is what makes it capable of being false, so the string must be reworded to something state-neutral or dropped.

### 2. UVE entry-point switch behind the flag

**The flag.** Add a new, dedicated flag — `FEATURE_FLAG_EXPERIMENTS_PORTLET` — shipped explicitly `false` in `dotmarketing-config.properties`, with a matching `FeaturedFlags` entry and an allow-list entry in `ConfigurationResource` so the frontend can read it.

[...]

PR #37144, opened 2026-08-20
docs(experiments): specification for the Experiments portlet

### LINKED ISSUE #36987: Experiments v2 — Part 2: Build the new Experiments UI

oidacra:
## What

A descriptive specification for the standalone Experiments portlet as a whole —
across the list, create/update and results screens — rather than for any one of
them. It captures the decisions and invariants that until now lived only in
docblocks and commit messages, and it is the baseline the portlet's remaining
screens are planned against.

Refs #36987. Covers #36989, #37003, #37004 as built, and #37005, #37006, #37007,
#37008 as pending.

> **Stacked on #37135.** Base is `issue-37004-experiments-results-screen`, so this
> diff is one new file. Retarget to `main` after #37135 merges.

## Why it reads as documentation, not as a request

Most of this feature already ships. Every requirement therefore carries a `[BUILT]`
or `[PENDING]` marker, and the spec states in its opening section that the
constitution's TDD gate (Principle V) applies to the pending work and to future
changes — **not** retroactively to code that is already merged or in review. It
does not reconstruct a Red phase that never happened.

## What it records

- Five prioritized user journeys with 40 acceptance scenarios, including the two
  that are easy to lose: keeping the portlet usable when analytics is broken, and
  keeping the legacy tree deletable in one commit.
- 77 functional requirements, grouped cross-cutting / per screen / analytics
  health / coexistence.
- A separate **Architectural Invariants** section holding the eight load-bearing
  constraints in implementation terms on purpose — the events-plugin store
  contract and its `@ngrx/signals` version constraint, why results are fetched
  lazily and never polled, why significance is the backend's call rather than a
  conversion-rate comparison, why a lost report must not cost the whole screen,
  and why nothing in the new tree may import from `old/`. Paraphrasing these into
  technology-neutral language would have destroyed the constraint they impose on
  the remaining screens.
- Explicit out-of-scope list, so the exclusions already agreed on the issues are
  not reopened.
- Legacy Considerations and a traceability table mapping requirement ranges to
  issues, PRs, source files and commits.

## Follow-ups this surfaced

Planning against the spec turned up three things worth fixing on the pending
issues before they are picked up:

- **#37008** says "promote `site-wide/` to the lib root". No `site-wide/` directory
  was ever created — the new code shipped at `src/lib/` with the legacy quarantined
  in `src/lib/old/`. The migration is a deletion, not a promotion.
- **#37008** says "delete the old `shared/` tree". There are two: `old/shared/` goes,
  and `src/lib/shared/` must stay — it holds the charts, legend plugin and details
  table the Results screen renders from.
- **#37007** cannot meet its own acceptance criteria as filed: the shipped list has a
  goal filter with counts and a repeatable `goal` URL param, and #36823's contract
  has neither.

## Scope of the diff

One new file, `specs/36987-experiments-portlet/spec.md`. The plan, research and
checklist artifacts stay local, per the spec-folder commit policy in
`.specify/CUSTOMIZATIONS.md`.

## Verification

Documentation only — no source, no build config, no tests touched. Prettier is
scoped to `core-web/` and does not cover `specs/` at the repo root, so there is no
formatting step that applies. Nothing to lint or run.

This PR fixes: #36987


specs/36763-portlet-experiments/spec.md
```diff
@@ -0,0 +1,708 @@
+# Feature Specification: Experiments Portlet
```
erickgonzalez, line 1: Let's rename the folder to keep our format

erickgonzalez (dismissed):

specs/36763-portlet-experiments/spec.md
oidacra: done

erickgonzalez (dismissed):

PR #37149, opened 2026-08-20
fix(block-editor): register the Highlight mark and load array-shaped values

### LINKED ISSUE #37145: Block Editor 2.0: Highlight mark not registered — Story Block field renders blank for content containing highlight marks (parity regression with legacy editor)
### Problem Statement

The new TipTap-v3 Block Editor (`core-web/libs/new-block-editor`) **does not register the Highlight mark**. Any content carrying `highlight` marks — authored while the legacy editor was active — cannot be deserialized by the new schema, and **the entire field renders blank**.

| | Legacy editor | New editor |
|---|---|---|
| Registers Highlight | ✅ [`dot-block-editor.component.ts:35,737`](https://github.com/dotCMS/core/blob/main/core-web/libs/block-editor/src/lib/components/dot-block-editor/dot-block-editor.component.ts#L737) | ❌ absent from [`editor-extensions.ts:72-178`](https://github.com/dotCMS/core/blob/main/core-web/libs/new-block-editor/src/lib/editor/extensions/editor-extensions.ts#L72-L178) |
| Renders affected content | ✅ correctly | ❌ blank field |

#### Why the whole field disappears

TipTap deserializes the document as a **single atomic operation**, so one unknown mark aborts `Node.fromJSON` for the entire doc. With `enableContentCheck` at its default (off), TipTap logs a warning and falls back to an **empty document**:

```
[tiptap warn]: Invalid content. Passed value: {type: 'doc', attrs: {…}, content: Array(19)}
Error: RangeError: There is no mark type highlight in this schema
```

All 19 blocks are lost in the editor — **the stored JSON is intact**, the editor simply cannot open it.

#### The underlying defect: marks bypass sanitisation

The new editor already guards against unknown **nodes** — `preserveUnknownNodesInDocument` swaps them for the `dotUnsupportedBlock` placeholder. But `getKnownNodeNames` reads **only** `editor.schema.nodes`:

```ts
// core-web/libs/new-block-editor/src/lib/editor/editor.component.ts:128
function getKnownNodeNames(editor: Editor): Set<string> {
    return new Set(Object.keys(editor.schema.nodes));   // ← nodes only; marks skip this entirely
}
```

Registering Highlight fixes *this* instance. The unsanitised **mark** path is a broader hardening concern and is **out of scope for this issue** — recorded here only so the observation is not lost.

> [!NOTE]
> **Where the marks come from.** The legacy editor never exposed a Highlight **button** — there is no `toggleHighlight` anywhere in the codebase. It applies the mark as a *temporary selection indicator* in its link popover, so the author can see which text is being linked once focus moves to the search input:
> ```ts
> // libs/block-editor/.../dot-link-editor-popover.component.ts:372
> private focusSearchInput() {
>     this.editor().commands.setHighlight();   // paints the selection
>     this.searchInput?.nativeElement.focus();
> }
> ```
> When `unsetHighlight()` never runs — popover closed unexpectedly, or the contentlet saved while it was open — the mark is persisted. The reporting customer's JSON shows exactly this: a `highlight` run sitting immediately beside a `link` run. So these marks are **UI residue, not authorial intent**, and **the legacy editor is still producing them today**. The new editor solved the same UI problem with ProseMirror `Decoration`s (view-only, never serialized) in `SelectionPreserveExtension`, which is why it has no need for the mark itself.

Same class of defect as #36572 (Underline mark, shipped in `26.07.17-01`). That fix was scoped to Underline only, which is why Highlight is still unregistered.

#### Second defect found while testing: array-shaped values blank the field too

Opening the same contentlet through the **UVE side panel** blanks the field with a *different* error:

```
[tiptap warn]: Invalid content. Passed value: {0: {…}, 1: {…}, … 15: {…}, content: undefined}
Error: RangeError: Unknown node type: undefined
```

Note the shape of the value: numeric keys plus `content: undefined`. That is an **array spread into an object**. A Block Editor value reaches `setContent` either as a `{ type: 'doc', content: [...] }` document *or* as a bare array of nodes — and the new editor's helper spreads unconditionally:

```ts
// core-web/libs/new-block-editor/src/lib/editor/editor.component.ts
function preserveUnknownNodesInDocument(parsed, knownNodeNames) {
    return {
        ...parsed,                                              // array → { 0: node, 1: node, … }
        content: preserveUnknownBlockNodes(parsed.content, ...)  // array.content → undefined
    };
}
```

Spreading an array drops the document `type`, so TipTap throws `Unknown node type: undefined`. The legacy editor branches on the shape before preserving:

```ts
// libs/block-editor/.../dot-block-editor.component.ts:773
const preservedContent = Array.isArray(content)
    ? preserveUnknownBlockNodes(content, this.#knownEditorNodeNames)
    : { ...content, content: preserveUnknownBlockNodes(content.content, ...) };
```

The new editor lost that `Array.isArray` branch.

**This is independent of Highlight** — it is deterministic for any array-shaped value, whatever marks the content carries. It is also independent of the mark fix above: the two defects share only the symptom (blank field).

### Steps to Reproduce

1. Start dotCMS ≥ `26.07.04-01` (no flag overrides).
2. Create or update a contentlet so its Block Editor field contains a `highlight` mark:
   ```json
   { "type": "paragraph", "content": [
       { "type": "text", "marks": [{ "type": "highlight" }], "text": "Good Credit History:" }
   ]}
   ```
3. Confirm the mark persisted: `GET /api/v1/content/{identifier}`
4. Open the contentlet in the editor → the field renders **blank**, console shows `RangeError: There is no mark type highlight in this schema`.
5. Compare against the legacy editor (`DOT_FEATURE_FLAG_NEW_BLOCK_EDITOR=false`) → the same contentlet renders correctly, highlight applied.

### Expected Behavior

- [ ] The new Block Editor registers `@tiptap/extension-highlight` (**already a dependency** at `core-web/package.json:107`), at parity with the legacy editor.
- [ ] Existing content containing highlight marks **round-trips** through the new editor — marks are not stripped and the field is not emptied.
- [ ] Every mark registered by the legacy editor is audited for parity, not just Highlight.
- [ ] A Block Editor value passed as a bare **array of nodes** (as the UVE side panel does) loads correctly instead of being spread into an object and rejected as `Unknown node type: undefined`.

### Actual Behavior

- The whole Block Editor field renders blank; only a console warning indicates why.

> [!WARNING]
> **Risk of data loss.** The field loads via `setContent(..., { emitUpdate: false })`, so merely *opening* the contentlet does not dirty it — but the editor now holds an empty document. Any subsequent edit fires `onUpdate`, and a **Save persists the empty body over the real one**.

### dotCMS Version

26.08.03-01

### Severity

### Links

- Related: #36572 — same defect for the Underline mark

adrianjm-dotCMS:
## Proposed Changes

Two defects that both surface as a **blank Block Editor field**, found while fixing and then QA'ing the first one.

1. The **Highlight** mark was never registered, so content carrying `highlight` marks could not be deserialized.
2. `preserveUnknownNodesInDocument` spread **array-shaped** values into an object, so the UVE side panel blanked the field for *any* content.

Plus one parity follow-up: the highlight colour now matches the legacy editor.

---

## 1. Highlight mark not registered

TipTap parses a document **atomically**, so one mark missing from the schema aborts `Node.fromJSON` for the *whole* doc — and with `enableContentCheck` at its default (off), TipTap silently falls back to an **empty document**:

```
[tiptap warn]: Invalid content. Passed value: {type: 'doc', attrs: {…}, content: Array(19)}
Error: RangeError: There is no mark type highlight in this schema
```

The field renders blank over stored JSON that is perfectly intact — and because the editor then holds an empty doc, any subsequent edit + Save persists that empty body over the real one.

### Where the marks come from

Not paste, as first assumed. The **legacy editor's link popover** applies the mark as a temporary selection indicator while its search input holds focus:

```ts
// libs/block-editor/.../dot-link-editor-popover.component.ts:372
private focusSearchInput() {
    this.editor().commands.setHighlight();   // paints the selection
    this.searchInput?.nativeElement.focus();
}
```

If `unsetHighlight()` never runs — popover closed unexpectedly, or the contentlet saved while it was open — the mark is persisted. The reporting customer's JSON shows exactly this: a `highlight` run sitting immediately beside a `link` run. Two consequences:

- These marks are **UI residue, not authorial intent**. Nobody chose to highlight anything.
- **The legacy editor is still producing this content today**, so more contentlets will hit this as customers migrate.

The new editor solved the same UI problem correctly, with ProseMirror `Decoration`s (view-only, never serialized) in `SelectionPreserveExtension` — which is why it has no need for the mark itself.

### Fix

- **`editor-extensions.ts`** — registers `@tiptap/extension-highlight` (already a dependency at `core-web/package.json:107`). Unconditional, like Superscript/Subscript: `allowedBlocks` gates blocks, not marks.
- **`CLAUDE.md`** — adds `highlight` to the marks inventory.

### Why no toolbar button

Deliberate, and it differs from #36574 (a UI-only change for a mark already in the schema):

1. The legacy editor never exposed a Highlight button either — there is no `toggleHighlight` anywhere in the codebase. Given the marks are UI residue, a button would be a **new authoring feature**, not parity.
2. The SDK renderers have **no `highlight` case** (Angular `text.component.ts:125`, React `Texts.tsx:128`). Both fall to a default that returns the bare text and stops recursing, so an author-applied highlight would not render on the front end — and would drop every mark after it. Shipping a button for a mark the delivery layer cannot paint would be a new defect.

Authors can still remove existing highlights with the toolbar's clear-formatting button, and `Cmd/Ctrl+Shift+H` comes with the extension.

---

## 2. Array-shaped values blank the field

Found while QA'ing the above: opening the same contentlet through the **UVE side panel** still blanked it, with a *different* error.

```
[tiptap warn]: Invalid content. Passed value: {0: {…}, 1: {…}, … 15: {…}, content: undefined}
Error: RangeError: Unknown node type: undefined
```

Note the shape — numeric keys plus `content: undefined`. That is an **array spread into an object**. A Block Editor value reaches `setContent` either as a `{ type: 'doc', content: [...] }` document *or* as a bare array of nodes; the UVE side panel passes the latter, and the helper spread unconditionally:

```ts
return {
    ...parsed,                                              // array → { 0: node, 1: node, … }
    content: preserveUnknownBlockNodes(parsed.content, ...)  // array.content → undefined
};
```

Spreading drops the document `type`, so TipTap throws. **This is independent of Highlight** — it is deterministic for any array-shaped value, whatever marks the content carries; the two defects share only the symptom. The legacy editor branches on the shape first (`dot-block-editor.component.ts:773`); the new editor had lost that branch.

### Fix

`preserveUnknownNodesInDocument` now branches on `Array.isArray`, and moved from `editor.component.ts` to **`utils/unknown-block.utils.ts`**. It is a pure JSON transform over unknown blocks with no component dependencies, it now sits beside the helpers it wraps (`preserveUnknownBlockNodes` / `restoreUnknownBlockNodes`), and — since the component has no spec of its own — moving it is what made the fix testable at all.

---

## 3. Highlight colour: parity with the legacy editor

`#accef7`, the same colour the legacy editor sets inline via `Highlight.configure({ HTMLAttributes: { style: 'background: #accef7;' } })`, so migrated content looks unchanged to authors. Applied as a CSS rule on `.tiptap mark` rather than inline HTMLAttributes, so it lives with the rest of the editor's styles. Editor chrome only either way — the stored JSON carries no colour attribute.

---

## Testing

7 new tests, all confirmed **Red** first.

**`editor-extensions.spec.ts`** — the second failed with the exact `RangeError` from the customer's console:
- `highlight` is registered exactly once, at parity with the legacy editor.
- A legacy document with `marks: [highlight, bold]` deserializes and keeps its text.

**`unknown-block.utils.spec.ts`** — all five failed with `is not a function`:
- a bare array stays an array;
- an array is never turned into a plain object carrying an undefined `content`;
- unknown nodes inside a bare array are still preserved as placeholders;
- doc-shaped documents keep working;
- sibling doc attrs (the doc stats) survive.

## Checklist

- [x] `nx test new-block-editor` — 12 suites / 94 tests green, no regressions
- [x] `tsc --noEmit` on the lib clean
- [x] `nx lint new-block-editor` clean
- [x] `nx format:write` clean
- [x] Manually verified in Edit Content: content loads, highlight painted
- [ ] Manual re-check of the UVE side panel after the array fix (the bundle is deployed; not yet confirmed in the browser)

## Notes for reviewers

**Scope.** An earlier revision also added a generic unknown-*mark* sanitizer, so that any future missing mark would degrade gracefully instead of blanking the field. It was **split out and dropped**: ~82% of the diff, fixing nothing that was reported, and putting a permanent transform on every content-load path. Both defects here are reproduced, user-visible failures.

**Manual testing note.** The legacy JSP edit screen loads `/dotcms-block-editor/main.js` from the webapp, *not* from `nx serve` — verifying there requires building `dotcms-block-editor` and deploying the bundle into the container. Only the new Edit Content screen renders `<dot-block-editor>` from the dev server.

**Out of scope, worth its own issue:** the SDK renderers drop every mark following an unknown one, so `marks: [highlight, bold]` currently renders unbolded on the front end, and a `highlight` never paints at all. Text survives; formatting after the unknown mark does not.

Fixes #37145

This PR fixes: #37145

### DIFF
--- a/core-web/libs/new-block-editor/CLAUDE.md
+++ b/core-web/libs/new-block-editor/CLAUDE.md
@@ -180,6 +180,7 @@ What actions are available on each node type. **Slash** = appears in `/` menu (`
 | `code` | StarterKit | Inline code | any text |
 | `superscript` | `@tiptap/extension-superscript` | Sup | any text |
 | `subscript` | `@tiptap/extension-subscript` | Sub | any text |
+| `highlight` | `@tiptap/extension-highlight` | — (schema only; the legacy editor has no button either) | any text |
 | `link` | `@tiptap/extension-link` | Link popover | any text (gated by `link` allowed-block) |
 | `textAlign` | `@tiptap/extension-text-align` | Align L/C/R/Justify | configured for `paragraph` + `heading` only |
 
--- a/core-web/libs/new-block-editor/src/lib/editor/editor.component.css
+++ b/core-web/libs/new-block-editor/src/lib/editor/editor.component.css
@@ -644,6 +644,17 @@
     border-radius: 0.25rem;
 }
 
+/* ─── Highlight (mark) ──────────────────────────────────── */
+/* The Highlight extension renders a plain <mark>. Styled here rather than through the
+   extension's inline HTMLAttributes (which is what the legacy editor does) so the colour
+   stays with the rest of the editor's theme. Stored JSON is unaffected either way. */
+:host ::ng-deep .tiptap mark {
+    background-color: #fef08a;
+    color: inherit;
+    padding: 0.0625rem 0.125rem;
+    border-radius: 0.125rem;
+}
+
 /* ─── Upload placeholder ─────────────────────────────── */
 :host ::ng-deep .upload-placeholder {
     display: flex;
--- a/core-web/libs/new-block-editor/src/lib/editor/editor.component.css
+++ b/core-web/libs/new-block-editor/src/lib/editor/editor.component.css
@@ -645,11 +645,13 @@
 }
 
 /* ─── Highlight (mark) ──────────────────────────────────── */
-/* The Highlight extension renders a plain <mark>. Styled here rather than through the
-   extension's inline HTMLAttributes (which is what the legacy editor does) so the colour
-   stays with the rest of the editor's theme. Stored JSON is unaffected either way. */
+/* The Highlight extension renders a plain <mark>. Same colour as the legacy editor for
+   feature parity, so migrated content looks unchanged to authors — the legacy editor sets it
+   inline via `Highlight.configure({ HTMLAttributes: { style: 'background: #accef7;' } })`
+   (dot-block-editor.component.ts:737); here it lives with the rest of the editor's styles.
+   Either way it is editor chrome only: the stored JSON carries no colour attribute. */
 :host ::ng-deep .tiptap mark {
-    background-color: #fef08a;
+    background-color: #accef7;
     color: inherit;
     padding: 0.0625rem 0.125rem;
     border-radius: 0.125rem;
--- a/core-web/libs/new-block-editor/src/lib/editor/editor.component.ts
+++ b/core-web/libs/new-block-editor/src/lib/editor/editor.component.ts
@@ -54,7 +54,10 @@ import { EditorPopoverService } from './services/editor-popover.service';
 import { EditorStore } from './store/editor.store';
 import { stripDocStats } from './utils/doc-stats.utils';
 import { loadRemoteExtensions, parseCustomBlocksField } from './utils/remote-extensions.loader';
-import { preserveUnknownBlockNodes, restoreUnknownBlockNodes } from './utils/unknown-block.utils';
+import {
+    preserveUnknownNodesInDocument,
+    restoreUnknownBlockNodes
+} from './utils/unknown-block.utils';
 
 /** Stringifies the editor document for form output (plain ProseMirror JSON, no extra attrs). */
 function editorDocumentJsonText(editor: Editor): string {
@@ -128,16 +131,6 @@ function parseAllowedContentTypes(field: DotCMSContentTypeField | undefined): st
 
 function getKnownNodeNames(editor: Editor): Set<string> {
     return new Set(Object.keys(editor.schema.nodes));
-}
-
-function preserveUnknownNodesInDocument(
-    parsed: JSONContent,
-    knownNodeNames: Set<string>
-): JSONContent {
-    return {
-        ...parsed,
-        content: preserveUnknownBlockNodes(parsed.content, knownNodeNames)
-    };
 }
 
 /** True when {@link parsed} represents the same document already in {@link editor}. */
--- a/core-web/libs/new-block-editor/src/lib/editor/extensions/editor-extensions.spec.ts
+++ b/core-web/libs/new-block-editor/src/lib/editor/extensions/editor-extensions.spec.ts
@@ -1,6 +1,7 @@
 import type { Injector } from '@angular/core';
 
-import { Extension, flattenExtensions } from '@tiptap/core';
+import { Extension, flattenExtensions, getSchema } from '@tiptap/core';
+import { Node as PMNode } from '@tiptap/pm/model';
 
 import type { DotMessageService } from '@dotcms/data-access';
 
@@ -15,6 +16,11 @@ import type { SlashMenuService } from '../components/slash-menu/slash-menu.servi
  * feature depends on (#36646): the editor must register exactly one `link` / `underline`
  * (StarterKit v3 bundles both) and must drop remote extensions whose names collide with a
  * built-in instead of double-registering them.
+ *
+ * They also pin the mark inventory against the legacy editor (#37145). A mark the legacy
+ * editor registered but this one does not makes TipTap abort `Node.fromJSON` for the WHOLE
+ * document and fall back to an empty doc — the field renders blank even though the stored
+ * JSON is intact.
  */
 describe('createEditorExtensions', () => {
     // `allowedBlocks: ['link']` keeps DotLink but excludes table/codeBlock/image/etc., so the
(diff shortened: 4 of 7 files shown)


rjvelazco (commented):
Does the unknown block get removed when we save that block editor data to the backend?

adrianjm-dotCMS:
> Does the unknown block get removed when we save that block editor data to the backend?

 No  unknown blocks are kept. The editor can't display a block it doesn't recognise, so it shows a locked "**Unsupported block"** box instead , but it holds on to a copy of the original underneath, and writes that exact same content back when you save. Nothing gets stripped.
 
**Unknown marks** are the opposite, and that's the bug here: one doesn't just get dropped, it stops the whole document from loading. The editor ends up empty, so an edit + Save would overwrite the real body.

rjvelazco (approved):

PR #37077, opened 2026-08-16
#36937 #36938: feat(roles): add role-membership endpoints (grant user, bulk remove users)

hassandotcms:
Adds the two role-membership mutations for the Angular Roles & Tools portlet migration (epic #36909), replacing DWR `RoleAjax#addUserToRole` / `#removeUsersFromRole`.

Closes #36937, closes #36938. #36939

### POST /v1/roles/{roleid}/users/{userId} — grant user to role

Grants a direct membership. Returns `{granted, roleId, user: {userId, email, fullName}}`.

- Idempotent (legacy parity): granting an already-held role — direct or inherited — returns 200 and changes nothing. Membership inherits down the role tree, so granting an inherited role does not create a direct membership; documented in the OpenAPI description.
- `editUsers=false` role → 403 (covers workflow/system roles). Missing role/user → 404.
  
### DELETE /v1/roles/{roleid}/users — bulk remove members

JSON body `{userIds: [...]}` (same pattern as bulk category delete).

- Partial success: `{removedUserIds, skipped: [{userId, reason: not_found|inherited|error}]}`. `inherited` reports what legacy silently no-ops.
- Per-user commits, no batch transaction — a batch wrapper would roll back earlier removals when a later entry fails.
- Empty/null/blank `userIds` → 400. Missing role → 404.
  
### Both

  - Auth: backend user + Roles portlet + CMS admin (same gate as PUT/DELETE role).
  - Audit logs per mutation; typed immutable views; openapi.yaml regenerated per commit.
  
### Testing

22 new integration tests in `RoleResourceIntegrationTest` (MainSuite3a), TDD red→green.
Security review: no critical/high findings.

This PR fixes: #36937

PR #37156, opened 2026-08-21
#37109: fix(users): honor empty roles array on PUT /v1/users

### LINKED ISSUE #37109: PUT /api/v1/users silently ignores empty roles array — cannot remove all roles from a user
### Problem Statement

`PUT /api/v1/users` accepts a `roles` array whose contract (established in #33410) is full replacement: the list sent becomes the user's complete role set.

This works for every input except the empty array. Sending `"roles": []` — the admin's way of saying "this user should hold no roles" — returns **200 OK but changes nothing**. The empty list is treated as if the field was not sent.

The legacy Users portlet supports this operation: un-assigning every role and saving calls DWR `UserAjax.updateUserRoles(userId, [])`, which removes all of the user's user-assignable roles. The REST API cannot do what the legacy UI can, and it fails silently — the caller gets a success response while the user keeps CMS Administrator, Back-end User, Front-end User, etc.

**Root cause:** `UserResource#processRoles` gates on `UtilMethods.isSet(roles)`, which is `false` for both `null` and an empty list; additionally `UserForm` normalizes a missing `roles` field to an empty list at construction, erasing the distinction between "not sent" and "sent empty" before the resource can act on it.

**Expected behavior:**

- `roles` **absent** from the payload → roles untouched (current behavior, unchanged).
- `roles` **non-empty** → full replacement (current behavior, unchanged).
- `roles: []` → all of the user's **user-assignable** roles (`editUsers = true`) are removed, matching legacy `UserAjax.updateUserRoles` semantics. System-managed roles that role editing never touches (e.g. the user's individual role) are unaffected, exactly as in the legacy portlet.

### Consistency: apply the same rule to the role-side membership endpoint

Both membership-mutation surfaces must enforce the same legacy rule — only
user-assignable roles (`editUsers = true`) are manageable:

- `POST /v1/roles/{roleId}/users/{userId}` (grant) already returns **403** for
  non-assignable roles.
- `DELETE /v1/roles/{roleId}/users` (bulk remove) currently has no such gate: it
  accepts a non-assignable role id (e.g. a user's individual role, discoverable via
  `GET /v1/roles/users/{userIdOrEmail}`) and unlinks the membership. The legacy
  Roles portlet blocks this in its UI: for a role with `editUsers = false` the user
  rows render no selection checkboxes and the Remove button is hidden
  (`view_roles_js_inc.jsp`), so this removal was never reachable end-to-end.

This endpoint merged in #37077 on Aug 18, 2026 and has only shipped in the agile
releases cut since (v26.08.19-03+); it is not in any LTS and predates any client
adoption (the Angular Roles portlet that will consume it is not built yet), so
aligning it now is effectively non-breaking. The gate mirrors the grant endpoint:
**403 when `role.isEditUsers()` is false** (role-level, whole request — not a
per-user `skipped` reason, since assignability is a property of the role).

### Steps to Reproduce

1. Create a user with roles:

```
curl -X POST '/api/v1/users' -H 'Content-Type: application/json' -d '{
  "firstName": "Test", "lastName": "User", "email": "roles-test@dotcms.com",
  "password": ["1","2","3","4","5","6","7","8"],
  "roles": ["DOTCMS_BACK_END_USER", "Scripting Developer"]
}'
```

2. Update the user with an empty roles array:

```
curl -X PUT '/api/v1/users' -H 'Content-Type: application/json' -d '{
  "userId": "<userId from step 1>",
  "firstName": "Test", "lastName": "User", "email": "roles-test@dotcms.com",
  "roles": []
}'
```

3. Load the user's roles (e.g. `GET /api/v1/roles/users/<userId>`).

**Expected:** the user holds no user-assignable roles.
**Actual:** 200 OK, the user still holds Back-end User and Scripting Developer.

### Acceptance Criteria

- [ ] `PUT /api/v1/users` with `"roles": []` removes all user-assignable roles from the user
- [ ] `PUT /api/v1/users` without a `roles` field leaves roles untouched
- [ ] Non-empty `roles` behavior is unchanged
- [ ] OpenAPI description of the `roles` field documents all three cases
- [ ] Integration/Postman coverage asserts the resulting role set for each case
- [ ] `DELETE /v1/roles/{roleId}/users` returns 403 when the target role is not user-assignable (`editUsers = false`), mirroring the grant endpoint
- [ ] OpenAPI description of the DELETE endpoint documents the 403 and defines "user-assignable"

### dotCMS Version

Latest from main branch (behavior introduced with the #33410 fix, PR #33414)

### Severity

### Links

NA

hassandotcms:
Fixes #37109

**Problem**
  
`PUT /api/v1/users` treats an explicit `"roles": []` as "field not sent" and returns 200 without changing anything, so all roles cannot be removed from a user via REST. The legacy Users portlet (DWR `UserAjax#updateUserRoles`) supports this. Root cause: `UtilMethods.isSet()` conflates null and empty, and `UserForm` normalized a missing `roles` field
to an empty list at construction, erasing the distinction before the resource could act on it.

**Fix**

* `UserForm` no longer collapses null (field absent) into an empty list (explicitly sent).
* `UserResource#processRoles` now reconciles memberships against the payload, matching legacy `UserAjax#updateUserRoles`: only user-assignable roles (`editUsers=true`) are added or removed.
  * `roles` absent → roles untouched (unchanged)
  * `roles` non-empty → becomes the user's complete user-assignable role set (same net result as before)
  * `roles: []` → removes all user-assignable roles (was a silent no-op)
  * null/blank entries → 400, whole update rolls back (mirrors `RoleUsersForm` on the role-side endpoints)
* Side effects fixed by the reconcile: the old wipe-and-re-add transiently unlinked the user's individual role on every roles-bearing update, and a payload naming an `editUsers=false` role key aborted the whole request (`DotStateException`); it is now skipped, as legacy does.
* Consistency: `DELETE /v1/roles/{roleId}/users` (merged 2 days ago in #37077, not in any LTS) now rejects `editUsers=false` roles with 403, mirroring the grant endpoint. The legacy Roles portlet never allowed this removal — it hides the controls for such roles.
* OpenAPI descriptions document the roles contract; `openapi.yaml` regenerated.

### Checklist

- [x] Tests: `UserResourceIntegrationTest` (5 new: empty array removes assignable roles and preserves the individual role; absent field untouched; non-empty replaces;
unknown/non-assignable keys ignored; null entry → 400 + rollback), `RoleResourceIntegrationTest` (1 new: `editUsers=false` → 403, membership intact) — 49/49 and 7/7 green
- [x] Postman: new `UpdateUserEmptyRolesRemovesAll` folder asserting the resulting role set per case
- [x] `openapi.yaml` regenerated and committed

This PR fixes: #37109

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/rest/api/v1/system/role/RoleHelper.java
+++ b/dotCMS/src/main/java/com/dotcms/rest/api/v1/system/role/RoleHelper.java
@@ -336,9 +336,10 @@ public class RoleHelper {
     }
 
     /**
-     * Bulk-removes users from a role with PARTIAL-SUCCESS semantics: once the role resolves,
-     * the batch never fails as a whole — every removable direct membership is removed and every
-     * non-removable entry is reported in the result's {@code skipped} list (see #36938):
+     * Bulk-removes users from a role with PARTIAL-SUCCESS semantics: once the role resolves
+     * and passes the {@code editUsers} gate, the batch never fails as a whole — every removable
+     * direct membership is removed and every non-removable entry is reported in the result's
+     * {@code skipped} list (see #36938):
      * <ul>
      *   <li>{@code not_found} — no user matches the id</li>
      *   <li>{@code inherited} — the user is not a DIRECT member (holds the role only through
@@ -347,6 +348,13 @@ public class RoleHelper {
      *       portlet only offered checkboxes on direct rows — same outcome, now reported</li>
      *   <li>{@code error} — unexpected per-user failure, logged server-side</li>
      * </ul>
+     *
+     * Like the grant endpoint, the whole request is rejected with 403 when the role is not
+     * user-assignable ({@code editUsers = false}, see #37109): memberships of such roles — the
+     * user's individual role, system roles — are system-managed and were never removable through
+     * the legacy Roles portlet either (it hides the removal controls for them). Assignability is
+     * a property of the role, not of each user, so this is a request-level rejection rather than
+     * a per-user {@code skipped} reason.
      *
      * DELIBERATELY NOT {@code @WrapInTransaction}: partial-success semantics require per-user
      * commits — each {@code roleAPI.removeRoleFromUser} call is already transactional, and a
@@ -363,6 +371,12 @@ public class RoleHelper {
         final Role role = this.roleAPI.loadRoleById(roleId);
         if (null == role || !UtilMethods.isSet(role.getId())) {
             throw new DoesNotExistException("Role not found: " + roleId);
+        }
+
+        if (!role.isEditUsers()) {
+            throw new DotSecurityException(
+                    String.format("Users cannot be removed from role '%s' (%s): the role does not allow membership changes",
+                            role.getName(), role.getId()));
         }
 
         final List<String> removedUserIds = new ArrayList<>();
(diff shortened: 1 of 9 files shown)


wezell (approved):
This is the correct pattern and we use it in many places:

absent = untouched, empty set = remove all user-assignable roles, non-empty set = replace.

PR #37167, opened 2026-08-24
fix(webdav): reject document type declarations in WebDAV XML request bodies

swicken:
## Summary

The WebDAV endpoints take XML request bodies on `PROPFIND`, `PROPPATCH` and `LOCK`.
RFC 4918 defines those as plain namespaced documents, and no WebDAV client sends a
document type declaration, so there is no reason for us to accept one. This adds a
filter on `/webdav/*` that validates those bodies up front, returning `400` for
anything carrying a DTD, plus a size cap.

The check sits in a filter rather than in the parser because the WebDAV servlet comes
from a third-party library whose parsers are not configurable from our code, and one
of the three parses happens in a static method. A filter is the only place that
covers all three consistently.

## Behaviour

- Only `PROPFIND`, `PROPPATCH` and `LOCK` are inspected. Other methods are forwarded untouched.
- A body with a DTD, or malformed XML, returns `400` rather than the `500` it used to produce.
- Bodies over 512KB return `413`. Validating a body means buffering it, so it needs a bound.
- Empty bodies pass through unchanged: `PROPFIND` with no body means allprop, and `LOCK` with no body is a refresh.
- The body is replayed to the servlet byte for byte, so nothing downstream sees a consumed stream.

## Testing

Six unit tests in `WebDavXmlValidationFilterTest` cover rejection, pass-through with an
intact body, empty bodies, non-XML methods, the size cap, and detection across encodings.

Also exercised against a running instance: the existing `WebDav` Postman collection passes
all 14 of its requests, and `PROPFIND` (with and without a body), `OPTIONS`, `MKCOL`, `PUT`,
`GET`, `PROPPATCH` and `LOCK` all behave as they did before the change.

Closes dotCMS/private-issues#671

This PR fixes: #671

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/filters/WebDavXmlValidationFilter.java
+++ b/dotCMS/src/main/java/com/dotcms/filters/WebDavXmlValidationFilter.java
@@ -1,0 +1,206 @@
+package com.dotcms.filters;
+
+import com.dotmarketing.util.Logger;
+import java.io.BufferedReader;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.util.Optional;
+import java.util.Set;
+import javax.servlet.Filter;
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ReadListener;
+import javax.servlet.ServletException;
+import javax.servlet.ServletInputStream;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletRequestWrapper;
+import javax.servlet.http.HttpServletResponse;
+import javax.xml.XMLConstants;
+import javax.xml.parsers.SAXParserFactory;
+import org.xml.sax.InputSource;
+import org.xml.sax.XMLReader;
+import org.xml.sax.helpers.DefaultHandler;
+
+/**
+ * Validates the XML bodies sent to the WebDAV endpoints, rejecting any document that carries a
+ * document type declaration.
+ *
+ * <p>WebDAV clients do not send a DTD: the request bodies defined by RFC 4918 are plain namespaced
+ * XML. Accepting one only widens what the parser has to do, so this refuses them up front and
+ * keeps the parse predictable.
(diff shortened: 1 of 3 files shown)


dotCMS/src/main/java/com/dotcms/filters/WebDavXmlValidationFilter.java
```diff
@@ -0,0 +1,260 @@
...
+            reader.setErrorHandler(new DefaultHandler());
+        } catch (final Exception unconfigurable) {
+            Logger.error(WebDavXmlValidationFilter.class,
+                    "Could not build a DOCTYPE-detecting XML parser; refusing the WebDAV body",
+                    unconfigurable);
+            return true;
+        }
+
+        try {
+            reader.parse(new InputSource(new ByteArrayInputStream(body)));
```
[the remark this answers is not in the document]
wezell: /fp - ThreadLocalSaxParserFactory does these protections already

dotCMS/src/main/java/com/dotcms/filters/WebDavXmlValidationFilter.java
```diff
@@ -0,0 +1,251 @@
...
+            Logger.warn(this, String.format("Rejecting oversized WebDAV %s body from %s",
+                    request.getMethod(), request.getRemoteAddr()));
+            response.sendError(HttpServletResponse.SC_REQUEST_ENTITY_TOO_LARGE);
+            return;
+        }
+
+        // An absent body is legal: PROPFIND with no body means allprop, and LOCK with no body is
+        // a lock refresh. Only inspect when there is something to inspect.
+        if (body.get().length > 0 && declaresDoctype(body.get())) {
+            Logger.warn(this, String.format("Rejecting WebDAV %s body declaring a DOCTYPE from %s",
```
wezell: Should this use the SecurityLogger?

wezell (changes requested):
I would use the security logger here - this should never happen in normal operations.

dotCMS/src/main/java/com/dotcms/filters/WebDavXmlValidationFilter.java
swicken: done

erickgonzalez (approved):

dotCMS/src/main/java/com/dotcms/filters/WebDavXmlValidationFilter.java
wezell: nice

wezell (approved):
Good change

PR #37168, opened 2026-08-24
fix(uve): lock layout canvas during save to prevent container data loss (#36197)

### LINKED ISSUE #36197: Layout: Lose `containers`/`content` because the editing canvas is never locked during save
## Problem Statement

On the page **Layout** tab of the Universal Visual Editor (UVE), users intermittently lose work when editing the layout: containers disappear, a newly added container replaces an existing one (losing its content), and in at least one reported case an entire page's content was lost and had to be recovered from version history. The problem is worse on slower / higher-latency networks.

**Root cause (verified in code):** the editing canvas is **never locked while a save is in flight**, so the user can keep editing during the save → reload → re-hydrate window, and those edits diverge from what the server returns.

1. Every layout change emits `templateChange` → pushed to `updateTemplate$` → **debounced 5s** → `POST /api/v1/page/{id}/layout` (`edit-ema-layout.component.ts`).
2. On change, `setUveStatus(LOADING)` is set — but this **only blocks the Properties tab**, not the GridStack canvas. There is no `[disabled]`, no `pointer-events: none`, no overlay, and no `@if` gating tied to the save state on the canvas. (The comment at `edit-ema-layout.component.ts:137` literally reads *"Prevent the user to access page properties."*)
3. Because the canvas stays interactive during the 5s debounce, the in-flight `POST`, **and** the post-save `pageReload()`, any edit made in that window is dropped when `pageReload()` re-fetches the page and `updateOldRows()` merges server state over local state.
4. The resulting **divergent snapshot** is then sent on the next save to a backend that performs a **destructive diff** — any container present before but absent in the incoming payload has its `multi_tree` rows hard-deleted (`addRemoveChanges` → `removeMultiTrees`, `DELETE FROM multi_tree`). There is **no concurrency/version guard** on the backend (last-write-wins).

**Second, related problem — no save feedback.** Save status is shown only as a **1-second fire-and-forget toast**. On a slow connection the request is still pending (or has failed), but the UI has already moved on — so the user cannot tell whether their work persisted. This is the "they wait and nothing happens" report.

**Impact:** content editors across the customer's many state agencies; data loss is the most severe symptom. Browser/OS: not browser-specific (logic-level race).

## Steps to Reproduce

1. Open a page in the UVE and go to the **Layout** tab.
2. Add or move a container, then — within ~5 seconds, or while the save request is in flight — **keep editing** (add another container / move a row). Throttle the network (e.g. Chrome DevTools "Slow 3G") to widen the window and make it reliable.
3. Wait for the save + automatic page reload to complete.
4. **Observe:** edits made during the save/reload window are dropped from the canvas; on a subsequent save the backend diff deletes the now-"missing" container(s), removing their content.

> Note: because the canvas is never disabled, there is no UI signal that a save is in progress, so the user has no way to know they should stop editing.

## Acceptance Criteria

Scope: fix the data-loss race (#1) and the missing save feedback (#2). **Out of scope:** whether autosave stays, is removed, or becomes configurable — that product decision is tracked in #36165.

**Canvas lock (data-loss fix):**
- [ ] While a layout save is in progress, the Layout canvas is **non-interactive** (no drag, add, remove, resize of rows/columns/containers).
- [ ] The locked window spans the **entire** lifecycle: save trigger → state snapshot (captured atomically at save time) → `POST` → response → `pageReload()` re-fetch → `updateOldRows()` re-hydrate → unlock. The lock is **not** released on HTTP 200 before re-hydration completes.
- [ ] The two save paths (debounced save and force-save-on-leave) are serialized so two layout saves for the same page can never be in flight concurrently.
- [ ] **Regression test (required):** an e2e/integration test edits the canvas during an in-flight save and asserts **no container/content loss** after the save+reload settles. This test is the definition of done for the race being closed.

**Save-state feedback (trust fix):**
- [ ] The Layout tab shows a **persistent** save-state indicator with distinct states: `Unsaved changes` → `Saving…` → `Saved (timestamp)` → `Failed`. It does not auto-dismiss while a save is pending.
- [ ] On save failure, the indicator shows a `Failed` state with a retry affordance, and local edits are **not** discarded (the user does not lose work and is not misled into thinking the save succeeded).

## dotCMS Version

Latest from `main` branch (reproduces on current UVE Layout tab). Also reported in the customer's production/Cloud environment.

## Severity

High - Major functionality broken (data loss of page layout/content)

## Links

- Related feature request (product decision on autosave): #36165
- Slack thread: https://dotcms.slack.com/archives/C035XMEDQ4F/p1781535585069909
- Freshdesk ticket: https://helpdesk.dotcms.com/a/tickets/37840

gortiz-dotcms:
## Summary

Fixes dotCMS/core#36197 — UVE Layout tab data loss caused by the GridStack canvas never being locked while a layout save is in flight.

### Root cause

Every layout change emits `templateChange` → `updateTemplate$` → 5-second debounce → `POST /api/v1/page/{id}/layout`. The canvas remained fully interactive during the entire debounce window, the in-flight POST, and the `pageReload()` + `updateOldRows()` re-hydration cycle. Any edit made in that window was silently discarded when the server response overwrote local state. Because the backend performs a destructive diff (deletes any container absent from the incoming payload), mid-flight edits could permanently remove content.

---

### `TemplateBuilderComponent` (`template-builder.component.ts` / `.html`)

- **`@Input() disabled = false`** — new input that locks the canvas.
- **`applyGridDisabled(disabled)`** — called synchronously from `ngOnChanges` (before any render cycle):
  - `grid.disable()` / `grid.enable()` on the main grid and every subgrid — blocks drag, resize, and external drop at the GridStack event level.
  - Dispatches `keydown Escape` on `document` — GridStack's `DDDraggable` registers a document-level `keydown` handler and calls `cancel()` on Escape, removing the drag clone from the body so the overlay is visible. PrimeNG dropdowns also close on Escape, covering the open-dropdown edge case.
- **Visual overlay** — `@if (disabled)` block rendered **after** the grid content (correct DOM stacking order), with `bg-white/50 backdrop-blur-[2px]` and a centered `p-progress-spinner`, so the user clearly sees the canvas is saving.

### `EditEmaLayoutComponent` (`edit-ema-layout.component.ts` / `.html`)

- **`$isSaving` computed signal** — `computed(() => uveStore.uveStatus() === UVE_STATUS.LOADING)`, bound as `[disabled]="$isSaving()"` on the template builder.
- **Lock timing fixed** — `setUveStatus(LOADING)` moved from the `tap()` before `debounceTime` (locked on every keystroke) into `switchMap` (locks only when the POST is actually sent, i.e. when the "Saving…" toast appears). Users can edit freely during the 5-second debounce window.
- **`nextTemplateUpdate` guard** — returns early when `uveStatus === LOADING`. Covers edge cases where an in-progress action (held drag, open dropdown selection) completes after the freeze activates, preventing the late `templateChange` from resetting the debounce or corrupting the in-flight payload.

---

## Lock lifecycle

```
User edits → 5 s debounce settles
  → setUveStatus(LOADING)           [canvas locks + overlay appears]
  → POST /api/v1/page/{id}/layout
  → handleSuccessSaveTemplate()
  → pageReload()                    [LOADING held through re-fetch]
  → updateOldRows() re-hydration
  → setUveStatus(LOADED)            [canvas unlocks]
```

---

## Test plan

- [ ] Edit layout (move a row) → wait 5 s → "Saving…" toast appears → canvas shows overlay + spinner → cannot drag/add/resize during save + reload cycle → canvas unlocks after reload
- [ ] Edit layout → **do not wait** → immediately start dragging "Add Row" button → hold for 5 s → Escape cancels the drag, overlay appears → row is not added
- [ ] Edit layout → open "+" container dropdown in a box → wait 5 s → dropdown closes, overlay appears → canvas is frozen
- [ ] Edit layout → save fails (disconnect network) → canvas unlocks, `uveStatus` set to `ERROR`, user can retry
- [ ] `pnpm nx test portlets-edit-ema-portlet --testPathPattern=edit-ema-layout`
- [ ] `pnpm nx test template-builder --testPathPattern=template-builder.component`

---

https://github.com/user-attachments/assets/b77977ad-8b99-43ff-83f4-faa5705b2a4d

### DIFF
--- a/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.html
+++ b/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.html
@@ -1,4 +1,5 @@
 <dotcms-template-builder-lib
+    [disabled]="$isSaving()"
     (templateChange)="nextTemplateUpdate($event)"
     [layout]="$layoutProperties().layout"
     [template]="$layoutProperties().template"
--- a/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.spec.ts
+++ b/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.spec.ts
@@ -262,6 +262,9 @@ describe('EditEmaLayoutComponent', () => {
     describe('LOADING guard and disabled binding', () => {
         it('should drop templateChange events and not forbid navigation while uveStatus is LOADING', () => {
             store.setUveStatus(UVE_STATUS.LOADING);
+            // The mock is shared across all tests — clear accumulated calls from earlier tests
+            // before asserting this specific action had no effect.
+            (dotRouter.forbidRouteDeactivation as jest.Mock).mockClear();
 
             templateBuilder.templateChange.emit();
 
--- a/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.spec.ts
+++ b/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.spec.ts
@@ -224,6 +224,17 @@ describe('EditEmaLayoutComponent', () => {
             expect(store.isClientReady()).toBe(false);
         }));
 
+        it('should NOT set uveStatus to LOADING before the debounce fires', fakeAsync(() => {
+            const setUveStatusSpy = jest.spyOn(store, 'setUveStatus');
+
+            templateBuilder.templateChange.emit();
+            tick(1000); // Less than DEBOUNCE_TIME (5000 ms)
+
+            expect(setUveStatusSpy).not.toHaveBeenCalledWith(UVE_STATUS.LOADING);
+
+            tick(4000); // flush remaining timer to avoid pending-timer warning
+        }));
+
         it('should save right away if we request page leave before the 5 secs', () => {
             const saveTemplate = jest.spyOn(component, 'saveTemplate');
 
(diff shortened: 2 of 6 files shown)


gortiz-dotcms:
Closing in favor of #37244 which has verified commit signatures.

Issue #37067, opened 2026-08-14
Add Status multiselect filter to the Content Drive toolbar


zJaaal:
### Description

Add a **Status** filter to the Content Drive toolbar so users can narrow results to Archived, Unpublished or Locked content. This is the frontend half of the Status filter; the backend clauses are a separate ticket and this one is blocked by it.

**Design:** a **multiselect** chip filter with three options (Archived, Unpublished, Locked), sitting alongside the existing content type / base type / language / workflow filters. Nothing selected means today's behavior: archived hidden, everything else returned.

**Semantics: AND.** Unlike the existing multiselects (base type, content type, language), which are OR within the facet, the three statuses are independent boolean flags on the same content row. Selecting several narrows the result set.

**Why multiselect rather than a single select:** `Unpublished + Locked` ("drafts currently checked out by someone") is a real question content teams ask, and a single select cannot express it. The legacy Content Search portlet exposes these as a mutually exclusive dropdown (`view_contentlets.jsp:695-698`), but that was a UI simplification, not a domain constraint: its backend has always ANDed the three flags independently (`ContentletAjax.java:1010-1021`).

**Current state:** `archived: false` is hardcoded in the search request at `dot-content-drive.store.ts:135`. That pin is removed and driven by the new filter instead.

**Two behaviors worth knowing before building the UI:**

1. `Archived + Unpublished` will always return the same count as `Archived` alone. Archiving removes the live version (`ESContentletAPIImpl.java:3833`), so every archived item is already unpublished. The combination is redundant, not broken, and should not look like a bug to the user.
2. `Archived + Locked` is reachable but uncommon (it needs a self-lock or a CMS Admin archive), so an empty result there is legitimate.

**Folders carry no status.** When any status is selected, folders must drop out of the results, consistent with how the store already handles the other narrowing filters (`dot-content-drive.store.ts:137-146`) and how `ContentDriveHelper` sets `showFolders(false)` for the workflow filter.

### Acceptance Criteria

- [ ] Status multiselect renders in the Content Drive toolbar with options Archived, Unpublished and Locked
- [ ] Options use i18n keys via `DotMessagePipe`, following the `content-drive.*` key convention
- [ ] Selecting one or more statuses sends them on the drive-search request and the results narrow accordingly
- [ ] `Unpublished + Locked` returns content that is both (the driving use case for a multiselect)
- [ ] Empty selection sends no status and preserves current behavior (archived content stays hidden)
[...]

PR #36416, opened 2026-07-03
chore(speckit): install & customize Spec-Kit for dotCMS (legacy-aware, ADR-gated)

nollymar:
## What

Installs GitHub **Spec-Kit v0.12.4** (Claude Code skills) into `dotCMS/core` and customizes it for dotCMS. Standard flow: `/speckit-specify` **or** `/speckit-specify-fix` → `/speckit-plan` → `/speckit-tasks` → `/speckit-implement`.

## Why (dotCMS-specific requirements)

1. **Legacy-aware** — the codebase mixes modern `com.dotcms.*` and legacy `com.dotmarketing.*`; specs/plans must account for that, not assume greenfield.
2. **ADRs feed the plan phase** — ADRs live in the private repo `dotCMS/platform-adrs`; every `/speckit-plan` must consult relevant ones as binding input.
3. **Spec-Kit never creates ADRs** — it may only *propose* them; ADRs are authored in `platform-adrs` via its own `new-adr.sh`.
4. **Two spec flows** — separate templates/commands for new features vs. issue/bug resolution.

## How

- **Constitution** (`.specify/memory/constitution.md`): legacy-aware principle + Critical Rules + ADR law (mandatory consultation, never-create guardrail). Loaded by every skill.
- **ADR integration** (upgrade-safe, no shipped files edited): read-only `adr-context.sh` (`gh` → `platform-adrs/INDEX.md`), `/speckit-adr-context` skill, a **mandatory `before_plan` hook** in `extensions.yml`, and an **ADR Alignment gate** in the plan-template override.
- **Two flows**: `/speckit-specify` (feature) + new **`/speckit-specify-fix`** (defect-framed issue template). Both funnel into `/speckit-plan`, so fixes get the ADR + legacy gates too.
- **Upgrade-safety**: customizations live in `.specify/templates/overrides/` and net-new files; `resolve_template` prefers `overrides/`. See `.specify/CUSTOMIZATIONS.md`.

## Verification

- `resolve_template` → overrides for spec/plan, issue template for the fix flow ✅
- `adr-context.sh` returns real ADRs (e.g. ADR-0009/0018) and exits 0 on no-match ✅
- `extensions.yml` valid YAML; `create-new-feature.sh` dry-run OK ✅
- Change set is additive (only `CLAUDE.md` modified) ✅
- Remaining manual check: a live `/speckit-plan` run to watch the `before_plan` hook fire end-to-end.

## Note

[...]

Issue #34760, opened 2026-02-24
Content Drive: Optional Inclusion of System Host Content in Search


rjvelazco:
### Description

Currently, the Content Drive search functionality is hardcoded to always include content from the System Host. The front-end sends `includeSystemHost: true` to the search endpoint, regardless of user intent.

This task aims to provide users with the ability to decide whether they want to see System Host content in their results. The proposed solution involves adding a control (e.g., a checkbox or a field in an advanced search form) that allows the user to toggle this parameter before executing a search.

### Acceptance Criteria

TBD

### Priority

Medium

### Additional Context

- **Technical Note:** The current implementation hardcodes `includeSystemHost:true`. This needs to be refactored to be state-driven.
- **Design/UX:** TBD. The look and feel and the specific placement within the form are still under discussion by the product team.

zJaaal:
Closing this in favour of #36795, where the work has shipped: PR #37074.

The ticket's Acceptance Criteria and Design/UX were both "TBD", so the design decisions were taken on that branch and are recorded in the PR body under "Absorbed from #34760: optional inclusion of shared assets". In short:

- A **"Show shared assets"** toggle chip leads the Content Drive filter row, **on by default**. It is first because every other chip narrows within a set of assets, while this one decides which set is in play.
- The state rides the existing `filters` query param as `sharedAssets`, and is **seeded like the default language** rather than written only when it differs, so what is applied is always visible in the URL instead of being implied by a missing key. It therefore inherits the URL encode/decode, the back/forward guard and the legacy-editor round-trip.
- **"Clear all" restores it to on**, and now only appears at all when some filter differs from its default. Previously it was on screen permanently, because both the language and this toggle are always seeded.
- The technical note on this ticket is resolved: `includeSystemHost` is no longer hard-coded in the store, it derives from that filter. **No backend change was needed** — `AbstractDriveRequestForm.includeSystemHost` already defaulted to `true` and flows through `ContentDriveHelper` to `BrowserQuery.forceSystemHost`.

Verified against a running instance: default load sends `"includeSystemHost": true`, toggling off sends `false` and drops the System Host rows, a deep link with `sharedAssets:false` loads off, and Back re-hydrates.

*Posted by Claude Code on @zJaaal's behalf.*

PR #37170, opened 2026-08-24
docs(content-drive): spec and design for Status filter (Archived, Unpublished, Locked) (#37066)

### LINKED ISSUE #37066: Content Drive: Status filter (Archived, Unpublished, Locked) on drive search and in the toolbar
### Description

>

## Backend

|---|---|---|

|---|---|---|

```java
```

**Target predicates:**

|---|---|---|

```sql
-- [UNPUBLISHED, LOCKED]

```

## Frontend

### Acceptance Criteria

**Backend**

**Frontend**

### Priority

Medium

### Additional Context

zJaaal:
**Spec-Kit PR 1 of 2.** Carries the spec and the two design artifacts that clear the commit-policy bar in `.specify/CUSTOMIZATIONS.md`. Needs a developer **approval** (not a merge) — the implementation PR stacks on this branch rather than waiting for it to land on `main`.

Resolves the spec + plan phases of #37066.

### Proposed Changes
* `spec.md` — the reviewed contract: 5 prioritized user stories, 20 functional requirements, 7 success criteria, edge cases, and the dotCMS Legacy Considerations section.
* `data-model.md` — the `ContentStatus` enum and its column/index mapping, the `status` transport shape and validation rules, the per-selection predicate table, the `BrowserQuery` field, the frontend filter-bag entry.
* `contracts/drive-search-status.md` — the `status` field on `POST /api/v1/drive/search`.

`plan.md`, `research.md` and `quickstart.md` were generated and stay local, per the commit policy in #36416. The two committed design docs are written to stand on their own — no citations point at files outside this diff.

### ⚠️ Semantics reversed since the ticket was written

The ticket originally specified these statuses combine with **AND**. **This spec says OR** — selecting more statuses returns *more* content, exactly like the Content Type and Language filters beside it. #37066's body and acceptance criteria have been updated to match, with a comment recording why, so the ticket and the spec do not diverge.

The case for AND was that these are independent flags one item can hold at once, so intersecting them is meaningful. It is — for one of four combinations:

|---|---|---|
| `Archived + Unpublished` | **redundant** — identical to `Archived` alone, since archiving removes the live version | "everything with no live version" |
| `Archived + Locked` | **almost always empty** | "archived or checked out" |
| `Unpublished + Locked` | useful — "drafts checked out" | "drafts or checked out" |

The first draft of this spec had to document two of those as "expected, not a bug" — that disclaimer was the design objecting. The decisive point is consistency: every other chip in that toolbar row widens on selection, and nothing in the UI can convey that one of them inverts the rule. A user who checks a second box and gets *fewer* results reads that as broken. The other filters being OR is *forced* by their single-valued nature (an item has exactly one base type), but that reasoning is invisible to the user; all they learn is the pattern.

**The thing to sign off on deliberately:** "unpublished **and** locked" — drafts currently checked out — is no longer expressible anywhere in Content Drive. Called out in `spec.md` Assumptions rather than absorbed quietly. If it turns out to matter, it belongs in a later refinement that makes the combining rule explicit in the UI.

### Design decisions worth a reviewer's eye

**1. All three predicates resolve in the database.** [ADR-0018](https://github.com/dotCMS/platform-adrs/blob/main/decisions/0018-database-first-content-drive-search-with-index-deferred-text-filtering.md) routes version-info flags (archived/deleted) to the DB and says structural criteria must never be re-routed to the index for speed. All three statuses are version-info flags:

| Status | SQL disjunct | Index term |
|---|---|---|
| `ARCHIVED` | `cvi.deleted = true` | `deleted:true` |
| `UNPUBLISHED` | `cvi.live_inode is null` | `live:false` |
| `LOCKED` | `cvi.locked_by is not null` | `locked:true` |

The index terms exist only because `BROWSE_API_HEURISTIC_TYPE=PURE_ES` is a supported (non-default) configuration where the filter would otherwise silently no-op, returning a **wider** set than asked for. Under the default hybrid heuristic the SQL clauses apply with and without a keyword.

**One accepted divergence (FR-009a).** `UNPUBLISHED` means *no live version exists* — a question about the content, not about one version. The index stores `live` per version, so a published item with newer unpublished edits has a working document carrying `live:false`, which `live:false` matches and `cvi.live_inode is null` does not. Under `PURE_ES`, `UNPUBLISHED` therefore returns that item too.

There is no index-side fix short of redefining the status per-version, which would degrade the **default** path to match a limitation of a strategy nobody runs. ADR-0018 already routes structural predicates to the database for exactly this reason and states that `PURE_ES` forfeits that guarantee for every criterion; this is one instance of a limitation that decision accepted. FR-009 and SC-005 are narrowed accordingly, and `contracts/` carries a table contrasting the two predicates. `ARCHIVED` and `LOCKED` are unaffected, as is the default path.

**2. The archived baseline sits OUTSIDE the OR group.** Excluding archived content is the drive's pre-existing default, not a fourth status. The selected statuses are OR-ed together and that group is AND-ed against the baseline, which only `ARCHIVED` lifts:

```sql
  and cvi.deleted = false                                        -- baseline, unless ARCHIVED selected
```

Folding the baseline into the group is the bug to avoid — `[UNPUBLISHED, LOCKED]` would become `(deleted = false or …)` and match nearly every row: a filter that silently stops filtering.

**3. `status` is typed `List<String>`, not `List<ContentStatus>`.** FR-010 wants a 400 on an invalid value "consistent with how `userSearchable` rejects unknown keys", and that precedent is an explicit `BadRequestException` in `ContentDriveHelper`. A typed field would let the Immutables/Jackson layer reject it as an `InvalidFormatException` whose mapping to a useful 400 is not under this code's control. The typed field would read better; the deterministic error is worth more.

**4. No schema or mapping change.** `deleted`, `live_inode` and `locked_by` already exist on `contentlet_version_info` (`postgres.sql:550-552`); `locked` is already indexed (`ESMappingAPIImpl.java:527`). Nothing here is rollback-unsafe.

**5. The existing `showArchived` is left alone.** It is *inclusive* — archived content plus everything else — and the legacy Site Browser checkbox (`view_browser.jsp:145`) depends on that. The exclusive Archived behavior is added alongside (FR-008), with an integration assertion guarding the old one.

### The risk this design carries into PR 2

Two existing code paths already have an opinion about `cvi.deleted`, and both must learn about `ARCHIVED` or the filter breaks:

* The **global archived exclusion** (`BrowserAPIImpl:2006`) would otherwise emit `cvi.deleted = false` AND-ed against a group containing `= true`, so `ARCHIVED` would always return nothing.
* **Archive-target workflow steps** (`:1980`) — `appendWorkflowQuery` owns `cvi.deleted` per branch when such a step is selected, so `ARCHIVED` needs the same treatment `showArchived` already gets. `ContentDriveWorkflowArchiveStepTest` is the regression guard.

A third, quieter one: `BrowserQuery`'s `showWorking` derivation (`:151`) must also be true for `ARCHIVED`/`UNPUBLISHED`, or `selectQuery` joins on `live_inode` — which those rows never have — and the filter returns nothing with no error. See `data-model.md`.

With no `status` sent, every generated query stays byte-identical to today.

### UI placement and navigation
`shared-assets | content-type | workflow | status | locale`

The row has exactly **one** real dependency edge, and it must not be broken:

| Edge | Real? |
|---|---|
| content-type → workflow | **yes** — `getSchemesByContentTypes` derives the scheme list |
| content-type → field-filter menu | **yes** — only materializes at exactly one content type |
| shared-assets → content-type | no — the content-type filter reads only `baseType`/`contentType` |
| status → anything | no — its three options are fixed |

So Content Type and Workflow stay adjacent. Status is independent of everything, which means it goes wherever costs least: beside Workflow, so the two lifecycle questions still read as a pair, but *after* it rather than between Workflow and the selection it derives from. Content Type keeps its position (almost certainly the most-used filter) and only Locale shifts right by one.

Rejected: placing Status between Shared Assets and Content Type. It reads well as broadest-scope-first, but it separates Content Type from Workflow — the one pairing the code actually couples — and the gate→pivot relationship that would justify it doesn't exist.

The shared-assets toggle is deliberately **not** moved. Its comment justifies position 1 as "a precondition for the filters that narrow within them", but nothing in the row derives from it; its better justification is that it's a different *class* of control — the only `mode="toggle"` chip, and the only one that changes what's in the set rather than filtering within it. There's a real frequency argument for demoting a toggle almost nobody flips out of the first slot, but that's a filter this ticket doesn't own; it belongs in its own ticket.

The selection rides the shared `filters` bag rather than its own query param, so it inherits deep link, reload, folder browsing, Back/Forward and the legacy-editor round-trip identically to every other filter (FR-016).

### Checklist
- [ ] Tests — land in PR 2, written and confirmed failing first (Constitution Principle V). Strategy: unit for status parsing + the 400; integration for each status, each pair, all three, the empty default, the never-shrinks property, `PURE_ES` parity and the workflow archive-step regression; Jest/Spectator for the chip, the store payload and the URL round-trip (including the single-value decode case, where a lone `status:ARCHIVED` would otherwise decode as a string rather than an array)
- [ ] Translations — four `content-drive.status-filter.*` keys, in PR 2
- [x] Security Implications Contemplated — the only input is a closed enum, validated and rejected with a 400, so it never reaches SQL as text. Permission filtering is untouched, so the filter grants no new visibility

### Additional Info
Absorbs #37067 (closed, merged into #37066). Parent epic #33999. Shared Assets / System Host toggle out of scope (#34760). #37171 was an earlier split of the design into its own stack layer, folded back here.

`/speckit-adr-context` ran as the mandatory `before_plan` hook. ADR-0018 (proposed) governs and this design complies; ADR-0009 (accepted) constrains only the index terms, all core fields mapped identically under ES and OpenSearch; ADR-0020 surfaced but does not apply. **No new ADR proposed.**

This PR fixes: #37066

This PR fixes: #37066


zJaaal:
Both 🟡 notes addressed in `b154169`, and a third issue the review surfaced indirectly in `fe40cd34`. Note the PR has grown since that review: it now carries the design artifacts too (`plan.md`, `research.md`, `quickstart.md`, `data-model.md`, `contracts/`), which were briefly split into #37171 before being folded back here.

**1. FR-004 vs FR-006 — right that a strict reader would get this wrong.** They're not contradictory, but the spec never said why. Excluding archived content is the drive's *pre-existing baseline* (`appendExcludeArchivedQuery` runs on every request today), the statuses AND on top of it, and `ARCHIVED` is the only one that lifts it. `data-model.md` now has a per-selection predicate table making that explicit, and `contracts/` points at it. Worth noting it needs no special handling in code — the baseline is skipped only when the selection contains `ARCHIVED`, so `UNPUBLISHED`/`LOCKED` alone keep it for free.

**2. `LOCKED` + `working:true` — verified, composes cleanly.** `ContentletAjax.java:1018` appends `+locked:true` and `:1035` unconditionally appends `+working:true`, so legacy has always paired them. Documented, along with one deliberate difference: legacy *always* scopes to working, whereas here only `ARCHIVED` and `UNPUBLISHED` force it (neither can have a live version). So `live: true` + `LOCKED` yields "live content that is locked" — coherent and strictly more expressive, not a bug.

Both now have integration cases in the plan's Test Strategy. The FR-004 one carries a fixture note: it needs an item that is **both** archived and unpublished, or the assertion passes vacuously.

**3. On `feature.json` — the review passed it as "no issue", but it was wrong to commit and it's now removed.** It isn't a spec artifact; it's a per-developer pointer to the feature *you're* currently on, rewritten by `/speckit-specify`. Committed, it would aim every teammate's next `/speckit-plan` at whatever feature last merged, and two concurrent feature branches would conflict on its single line every time. It had never been committed on `main` (36605 and 36834 both shipped without it) but wasn't gitignored either — untracked by luck. Now it's a rule, documented in `.specify/CUSTOMIZATIONS.md`.

Also confirmed the `buildPureESQuery:612` point: it's `research.md` R6 and has its own `PURE_ES` parity test in the strategy.

<sub>Analysis and reply by Claude (Claude Code), posted from @zJaaal's account.</sub>

zJaaal:
Thanks — one real finding, and two that were made against a stale snapshot of the PR description.

**🟡 R-references pointing at a gitignored file — valid, fixed in `a74385f`.** Citing `research.md` R2/R4/R7 from a committed document, in a PR that reaffirms gitignoring `research.md`, is exactly the tension you describe. Did both things you suggested: the reasoning is inlined in full **and** the R-numbers are gone, so nothing in the committed docs points outside the diff.

The `showWorking` passage gained the most, since it was the least verifiable. It now names the two call sites that make the flag load-bearing — `BrowserAPIImpl:1947` picking the joined inode column, and `:615` choosing `+working:true` vs `+live:true` — explains why the join can never match for archived/unpublished rows, and adds two things that were only in `research.md` before: the drive path is safe today *only* because the form's `live()` defaults to `false` (a coincidence in one caller, not a property of `BrowserQuery`), and `LOCKED` alone does **not** need the flag, since a locked item may well have a live version. Also dropped the last dangling pointer from the PR body's test-strategy line.

**🟡 "PR description vs committed spec — stale AND/OR narrative" — already fixed before the review ran.** The body was rewritten at 14:32 UTC and no longer contains the "The one thing to check in the spec" section you quote. It now leads with **⚠️ Semantics reversed since the ticket was written**, the same four-combination comparison table, and the explicit trade-off callout. I think the review read an earlier revision — worth knowing the description can move under a review that takes a couple of minutes.

To answer the substantive half anyway, since it deserves a direct answer: **yes, OR is deliberate**, and yes it flips the ticket's original requirement. That AND wording was my own earlier rewrite of #37066, not a team decision, and it did not survive planning — under AND, two of three pairs are degenerate and the first draft of the spec had to document them as "expected, not a bug," which was the design objecting. The decisive point is that every other chip in that toolbar row widens on selection and nothing in the UI can convey that one of them inverts the rule. #37066's body and acceptance criteria are now updated to match, with a comment recording why, so the ticket and the spec no longer diverge. The `Unpublished AND Locked` trade-off is called out in `spec.md` Assumptions precisely so it gets signed off deliberately rather than absorbed.

**Minor — "PR body still lists plan.md/research.md/quickstart.md under Proposed Changes"** — same stale snapshot. Proposed Changes lists only the four committed items; the three process artifacts appear in one line stating they stay local per #36416.

<sub>Analysis and reply by Claude (Claude Code), posted from @zJaaal's account.</sub>

specs/37066-content-drive-status-filter/spec.md
```diff
@@ -0,0 +1,302 @@
...
+  hidden, everything else returned — with no change to result counts, ordering or pagination.
+- **Archived is the only status that reveals archived content.** Selecting Unpublished or Locked
+  alone must not surface archived items, even though every archived item is technically also
+  unpublished. Hiding archived content is the drive's standing default, and only Archived lifts it.
+- **An unrecognized status value** submitted directly to the search endpoint is rejected with a
+  clear client error naming the accepted values, rather than being silently ignored (which would
+  return a wider result set than the caller asked for).
+- **Status combined with a workflow filter.** Content Drive can already filter by workflow, including
+  steps that archive content. A status selection combined with such a workflow filter must return a
+  coherent result, not an empty one caused by two rules contradicting each other about archived
```
nollymar: in this case, will a workflow filter + a status become an AND?

specs/37066-content-drive-status-filter/contracts/drive-search-status.md
```diff
@@ -0,0 +1,135 @@
...
+---
+
+## Query-path guarantees
+
+The same `status` selection must return the same set regardless of which search strategy the
+environment runs (FR-009 / SC-005). Both paths are in scope:
+
+| `BROWSE_API_HEURISTIC_TYPE` | Path | How `status` applies |
+|---|---|---|
+| `HYBRID_SINGLE_CHUNKED_QUERY_ES` (**default**) | `BrowserAPIImpl.selectQuery` supplies the candidate set; the index only narrows by text | SQL clauses — so they apply with **and** without a keyword |
```
nollymar, line 105: we have to make sure that an empty status does not change these query path guarantees. The status field must be ignored

specs/37066-content-drive-status-filter/data-model.md
```diff
@@ -0,0 +1,232 @@
...
+---
+
+## Entity: `ContentStatus` (new)
+
+`dotCMS/src/main/java/com/dotcms/browser/ContentStatus.java`
+
+A closed enum of three independent states a contentlet version can hold.
+
+| Constant | Meaning | Backing column (`contentlet_version_info`) | Index term |
+|---|---|---|---|
```
nollymar, line 22: Important to consider: the OR in Lucene must be explicit when multiple statuses are added in the query. Otherwise, `+deleted:true +live:false` will be translated to `+deleted:true AND +live:false`

specs/37066-content-drive-status-filter/spec.md
```diff
@@ -0,0 +1,306 @@
...
+### Key Entities
+
+- **Content Status**: An independent state an item can hold, from a closed set of three — *Archived*
+  (removed from circulation but recoverable), *Unpublished* (no live version), and *Locked* (checked
+  out by a user). An item may hold several at once, but the filter asks whether an item is in *any*
+  selected state, not all of them.
+- **Status Selection**: The set of statuses the user has chosen. Empty by default; every member
+  widens the result set.
+
+## Success Criteria *(mandatory)*
```
dario-daza, line 278: Should we describe that we need to include unit/integration/Postman tests to meet these success criteria, or are these instructions already in the local plan file?

specs/37066-content-drive-status-filter/data-model.md
zJaaal: Good catch — this was a real bug in the doc, and the one most likely to have shipped wrong. Fixed in `c07bea47`.

The enum table was listing the terms as `+deleted:true` / `+live:false` / `+locked:true`. Since `+` is REQUIRED, concatenating them is exactly the AND you describe — which would have silently reverted the OR semantics this spec spent its review cycle settling on. The terms are now **bare** in the table, with a callout not to prefix them, and the composed forms spelled out:

| Selection | Index query |
|---|---|
| `[UNPUBLISHED, LOCKED]` | `+deleted:false +(live:false OR locked:true)` |
| `[ARCHIVED, LOCKED]` | `+(deleted:true OR locked:true)` |

The archived baseline stays its own required clause outside the group, mirroring the SQL shape (`and cvi.deleted = false and ( … or … )`).

Worth noting the convention is already in the exact method being changed — `BrowserAPIImpl:630` writes `+(conhost:<id> OR conhost:SYSTEM_HOST)` — so the fix is to follow what `buildPureESQuery` already does, not to invent anything.

One deliberate exception: the `ContentletAjax.java:1018` reference keeps its `+locked:true`, because legacy genuinely *does* AND its status flags. I added a parenthetical there so nobody copies that form into the new code.

There's also a new test row for it: assert under `PURE_ES` that two statuses return the union, which is what catches a concatenated-`+` regression.

<sub>Analysis and reply by Claude (Claude Code), posted from @zJaaal's account.</sub>

specs/37066-content-drive-status-filter/spec.md
zJaaal: **Yes — AND.** OR applies *within* the status filter only; separate filters still AND with each other, exactly as they do today.

So `workflow=[schemeA]` + `status=[UNPUBLISHED, LOCKED]` means *governed by scheme A* **and** *(unpublished **or** locked)*. Adding a status can never loosen the workflow filter.

Verified rather than assumed: each filter appends its own `and ( … )` clause — `BrowserAPIImpl:2338` is the workflow one — so they compose the same way `contentType` and `languageId` already do.

The spec never stated this, which it got away with while every filter was OR-only. Now that OR means something specific *inside* status, the between-filters rule needs saying out loud. Added as **FR-006a** in `c07bea47`, plus a sharpened edge case (the workflow one previously only said the pairing must be "coherent" without saying how it combines), and an integration test row asserting a workflow + two-status query returns the intersection of the workflow set with the status union.

<sub>Analysis and reply by Claude (Claude Code), posted from @zJaaal's account.</sub>

specs/37066-content-drive-status-filter/contracts/drive-search-status.md
zJaaal: Agreed, and it wasn't actually guaranteed anywhere — fixed in `c07bea47`.

FR-002 said an empty set "preserves today's behavior exactly" but never said the field is **skipped entirely**. That gap matters concretely: a builder that opens a group and then has nothing to fill emits `and ( )`, which is a SQL syntax error, or `+()`, which is invalid Lucene. So it has to be an early return, not an empty group.

Three changes:
- **FR-002** now says the filter must be skipped entirely, not translated into a vacuous "matches anything" condition — and calls out that this is the **default path, not an edge case**, since every drive search that exists today sends no status.
- The contract gained an *"An empty `status` MUST be ignored entirely"* section under Query-path guarantees, covering both paths.
- A test row asserting the empty-status result set matches a no-status baseline exactly, and that neither builder emits a vacuous clause.

That's also what makes the plan's "byte-identical to today" claim literally true rather than approximately true.

<sub>Analysis and reply by Claude (Claude Code), posted from @zJaaal's account.</sub>

specs/37066-content-drive-status-filter/spec.md
zJaaal: They were in the local plan file — which is precisely the problem, since `plan.md` is gitignored by the commit policy in #36416, so you can't see it from this PR. You shouldn't have had to ask. Added a short note under Success Criteria in `c07bea47`.

For the record, the strategy is:

| Layer | Type | Covers |
|---|---|---|
| `ContentDriveHelper` status parsing | **Unit** (JUnit) | valid values map to the enum; unknown value → 400 |
| Drive search semantics | **Integration** | each status alone, each pair, all three, the empty default, the never-shrinks property, the explicit-OR-group check under `PURE_ES`, status × workflow, and the `ContentDriveWorkflowArchiveStepTest` regression |
| Toolbar chip + store | **Jest/Spectator** | single/multiple selection, clearing, folders excluded, URL round-trip (including the single-value decode case) |

**Postman is deliberately omitted**, and I'd rather state that than let it look like an oversight. The behavior needs seeded archived + locked fixtures, which a collection can't construct against a shared environment — the integration tests hit the same endpoint and can build exactly the state each assertion needs. Constitution Principle V treats "no tests of type X" as requiring an explicit reason rather than silence, so that's the reason. If you'd still like a Postman smoke case for the 400 on an invalid status value, that one *is* cheap and stateless — happy to add it.

Note the spec deliberately stays technology-agnostic (Spec-Kit's template convention), so the test types live in the plan rather than in `spec.md` itself; the new note points at them instead of inlining the whole table. Per Principle V they also land **before** implementation — `/speckit-tasks` orders every story as tests → developer-approval gate → confirmed-failing (Red) gate → implementation.

<sub>Analysis and reply by Claude (Claude Code), posted from @zJaaal's account.</sub>

dario-daza (dismissed):

hmoreras (approved):

Issue #37176, opened 2026-08-24
Allow changing the page of a draft experiment that has no variants


oidacra:
### Description

The page of an experiment is fixed the moment the draft is created. `$isPageImmutable` is `!!store.experiment()`, so the Select button is disabled for every experiment that already exists.

That is stricter than the data requires, and the underlying reason is a gap rather than a decision: `ExperimentsResource.patchExperiment` copies `name`, `description`, `trafficAllocation`, `trafficProportion`, `scheduling`, `goals`, `targetingConditions` and `lookbackWindow` from the form. It never reads `pageId`, even though `ExperimentForm` carries the field. A PATCH with a page change is discarded silently. `ExperimentsAPIImpl.save()` does not forbid the change either — it only validates the page exists and the user may edit it.

**Why variants are the real constraint.** `addVariant` copies the page's MultiTrees into each new variant, scoped to `experimentPage.getIdentifier()`:

```java
if (!DEFAULT_VARIANT.name().equals(experimentVariant.id())) {
    copyMultiTrees(persistedExperiment, experimentVariant);
}
```

A variant's content is a copy of *that page's* layout. Change the page and every copy is orphaned. The control is explicitly exempt from that copy — it holds no duplicated content, it *is* the page. So the swap is safe exactly while the control is the only variant.

**The detail that is easy to miss.** `createExperimentVariant` bakes the page URL into every variant, control included: `.url(page.getURI() + "?variantName=" + variantName)`. Even in the permitted case the swap must regenerate the control's stored `url`, or its "copy preview URL" button hands back a link to the previous page.

### Acceptance Criteria

**Backend**

- [ ] `patchExperiment` reads `pageId` from `ExperimentForm` and applies it.
- [ ] It is applied only when status is `DRAFT` **and** `trafficProportion.variants` holds nothing but `DEFAULT`.
- [ ] A `pageId` that differs from the stored one and fails either condition is rejected with **HTTP 400** and a message naming which condition failed. Silently dropping it is the current bug and must not survive.
- [ ] A `pageId` equal to the stored one is a no-op, not an error.
- [ ] On a successful change, the control variant's `url` is regenerated from the new page.
- [ ] Every field PATCH already supports keeps working unchanged.
- [ ] Schedule conflicts on the target page are **not** re-validated here. `start()` already owns that rule — a page may host several experiments as long as their schedules do not overlap — and duplicating it would be stricter than the platform.

**Frontend**

[...]

PR #37180, opened 2026-08-24
docs(experiments): spec for changing the page of a draft experiment (#37176)

oidacra:
## Spec only — PR 1 of 2

This PR carries `spec.md` alone, per the repo's spec-driven flow. It needs another dev's
**approval** (not merge) before `/speckit-plan` runs; the implementation lands in PR 2, branched
from here.

Refs #37176 — epic #36763 (Experiments: A/B Testing v2).

## What the spec covers

The page an experiment tests is frozen the moment the draft is created, so an editor who picked the
wrong page has one remedy: delete the experiment and rebuild it. That restriction is broader than
the data requires, and it rests on a gap rather than a decision — a submitted page is accepted by
the request contract and then never read.

The real constraint is variants: creating one copies the page's layout into it, so the copy only
means anything for the page it came from. The control is exempt from that copy — it owns no
duplicated layout, it *is* the page. So the swap is safe in exactly one case: still a draft, and
the control is the only variant.

The spec covers that case and the honest refusal of every other one.

| | |
|---|---|
| User stories | 4 (2xP1, 1xP2, 1xP3) |
| Functional requirements | 13 |
| Success criteria | 5 |
| `[NEEDS CLARIFICATION]` markers | 0 |
| Quality checklist | 16/16 pass (2 iterations) |

- **US1 (P1)** correct the page on a clean draft, control's preview URL follows
- **US2 (P1)** ineligible changes are refused, never silently dropped — this is the actual defect
- **US3 (P2)** the variants tooltip says what to delete
- **US4 (P3)** a read-only experiment keeps its existing lock reason

## Three things worth a reviewer's attention

**1. A backward-compatibility consequence the issue does not name.** A field that is ignored today
becomes applicable-or-rejectable, so any caller already sending a page changes behavior. FR-004
(equal value is a no-op) is what contains it: a client echoing the experiment back is unaffected.
Only a caller sending a genuinely different page at an ineligible experiment sees a new 400 — and
that caller is currently being told its change succeeded when it did not. Written up under Legacy
Considerations, since it is the reason that requirement exists rather than mere ergonomics.

**2. The control variant is identified two different ways in the codebase.** `ExperimentsAPIImpl`
detects it by its `Original` description; the frontend uses `id === 'DEFAULT' || name === 'Original'`.
The spec assumes a single check (Assumptions, item 1). Flagged for `/speckit-plan` to pick one and
decide whether the divergence deserves its own issue.

**3. One deliberate deviation from technology-agnostic wording.** FR-003 names HTTP 400 explicitly.
The issue makes the status code part of the contract — it is what distinguishes this fix from the
silent drop — so dropping it would have removed a real acceptance condition. It is confined to that
requirement; no success criterion depends on it.

## Not in this PR

`checklists/requirements.md` stays local: `specs/*/checklists/` is gitignored as a process artifact
per `.specify/CUSTOMIZATIONS.md`. The second commit adds `.specify/feature.json` to `.gitignore` —
a machine-local pointer rewritten on every `/speckit-specify` run, which Spec-Kit 0.16.1 will handle
through a managed ignore file once this repo moves off 0.12.4.

## Out of scope

Changing the page of a running, scheduled, ended, or archived experiment; migrating variant content
between pages; relaxing the variants rule by moving or re-copying layouts.

### DIFF
--- a/specs/37176-draft-experiment-page-change/spec.md
+++ b/specs/37176-draft-experiment-page-change/spec.md
@@ -1,6 +1,6 @@
 # Feature Specification: Changing the Page of a Draft Experiment
 
-**Feature Branch**: `issue-37176-draft-experiment-page-change`
+**Feature Branch**: `oidacra/allow-changing-the-page-of-a-draft-experiment-th`
 
 **Created**: 2026-08-24
 
--- a/specs/37176-draft-experiment-page-change/spec.md
+++ b/specs/37176-draft-experiment-page-change/spec.md
@@ -1,6 +1,6 @@
 # Feature Specification: Changing the Page of a Draft Experiment
 
-**Feature Branch**: `oidacra/allow-changing-the-page-of-a-draft-experiment-th`
+**Feature Branch**: `issue-37176-draft-experiment-page-change`
 
 **Created**: 2026-08-24
 
@@ -35,6 +35,26 @@ experiment is still a draft and the control is its only variant.
 This spec covers that situation and the honest refusal of every other one. It does **not** cover
 migrating variant content between pages, nor changing the page of a running, scheduled, ended, or
 archived experiment.
+
+**Why DRAFT and not also SCHEDULED.** A scheduled experiment has not started and has collected no
+data, so it looks like it should qualify. It does not, and the reason is not about data — it is
+about a clearance that has already been granted. Starting an experiment validates that its page is
+free of overlapping experiments, and it validates that against *the page the experiment had at that
+moment*. A scheduled experiment has passed that gate and is committed to starting on its own. Let it
+move to a different page and the clearance is silently stale: the new page may already host a
+scheduled or running experiment with an overlapping window, and nothing re-checks before the
+experiment fires.
+
+That is exactly why FR-007 can say schedule conflicts are not re-validated here. For a draft the
+rule is safe, because a draft has never been cleared and will be checked when it starts. For a
+scheduled experiment it would not be. DRAFT is the last state in which the page is still
+un-cleared, which is what makes it the boundary.
+
+Two consequences worth stating plainly. Admitting SCHEDULED would mean either accepting that hole or
+re-validating conflicts on the page change — the second contradicts FR-007 and is a larger change
+than this one. And in practice a scheduled experiment is very unlikely to qualify anyway: starting
+one requires more than the control variant, so the eligibility rule below would refuse it on the
+variants condition even if status allowed it.
 
 ---
 
@@ -247,6 +267,7 @@ assert the read-only reason is the one shown.
   page. No separate permission rule is introduced.
 - **Only the control's preview URL is derived from the page.** Nothing else stored on the experiment
   (goals, targeting, traffic split) is re-derived on a page change.
-- **Out of scope**: changing the page of a running, scheduled, ended, or archived experiment;
-  migrating variant content between pages; relaxing the variants rule by moving or re-copying
-  layouts.
+- **Out of scope**: changing the page of a running, scheduled, ended, or archived experiment —
+  scheduled included, for the clearance reason given in the Scope Note, not merely because it is
+  "already started"; migrating variant content between pages; relaxing the variants rule by moving
+  or re-copying layouts.
--- a/specs/37176-draft-experiment-page-change/spec.md
+++ b/specs/37176-draft-experiment-page-change/spec.md
@@ -1,0 +1,252 @@
+# Feature Specification: Changing the Page of a Draft Experiment
+
+**Feature Branch**: `oidacra/allow-changing-the-page-of-a-draft-experiment-th`
+
+**Created**: 2026-08-24
+
+**Status**: Draft
+
+**Type**: Task (behavior enablement + silent-drop defect)
+
+**Epic**: [#36763 — Experiments: A/B Testing v2](https://github.com/dotCMS/core/issues/36763)
+
+**Work item**: [dotCMS/core#37176 — Allow changing the page of a draft experiment that has no variants](https://github.com/dotCMS/core/issues/37176)
+
+**Input**: User description: "Allow changing the page of a draft experiment that has no variants" — taken from issue #37176.
+
+---
+
+## Scope Note *(read this first)*
+
+Today the page an experiment tests is frozen the instant the draft is created. The page picker is
+disabled for *every* experiment that already exists, so an editor who picked the wrong page has
+exactly one remedy: delete the experiment and start over.
+
+That restriction is broader than the data actually requires, and it rests on a gap rather than a
+decision. Partial updates to an experiment silently discard a submitted page — the field is accepted
+by the request contract, then never read. Nothing rejects the change; nothing applies it either.
+
+The **real** constraint is variants. Creating a variant copies the page's layout into that variant,
+so the copy is meaningful only for the page it came from. Repoint the experiment at a different page
+and every such copy is orphaned. The control variant is deliberately exempt from that copy — it owns
+no duplicated layout, it *is* the page. So a page swap is safe in exactly one situation: the
+experiment is still a draft and the control is its only variant.
+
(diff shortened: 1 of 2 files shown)


specs/37176-draft-experiment-page-change/spec.md
```diff
@@ -0,0 +1,252 @@
+# Feature Specification: Changing the Page of a Draft Experiment
+
+**Feature Branch**: `oidacra/allow-changing-the-page-of-a-draft-experiment-th`
+
```
erickgonzalez, line 4: Missing issue number

specs/37176-draft-experiment-page-change/spec.md
```diff
@@ -0,0 +1,252 @@
...
+1. **Given** a draft experiment whose only variant is the control, **When** the page card renders,
+   **Then** the page selection control is enabled.
+2. **Given** a draft experiment with at least one non-control variant, **When** the page card
+   renders, **Then** the control is disabled and explains that variants must be deleted first.
+
+---
+
+### User Story 4 - A locked experiment keeps its existing reason (Priority: P3)
+
+An editor opening a running, scheduled, ended, or archived experiment sees the page picker disabled
```
erickgonzalez, line 160: I think we should be able to change/update an experiment which is Scheduled but haven’t started?

erickgonzalez (changes requested):

specs/37176-draft-experiment-page-change/spec.md
freddyDOTCMS: In this case we need to make sure that the Experiment is valid to be running after the changes, I mean If the user remove all the variants to e allow to change the page, then this Experiment can be in the scheduled state before it has new variants in it

specs/37176-draft-experiment-page-change/spec.md
```diff
@@ -0,0 +1,252 @@
...
+assert the read-only reason is the one shown.
+
+**Acceptance Scenarios**:
+
+1. **Given** a non-draft experiment, **When** the page card renders, **Then** the control is
+   disabled with the existing read-only reason.
+2. **Given** a non-draft experiment that *also* has non-control variants, **When** the page card
+   renders, **Then** the read-only reason is shown, not the variants reason.
+
+---
```
freddyDOTCMS, line 178: What about a draft Experiment with variants, then we remove the variants and then we change the page, is it out of the scope?

specs/37176-draft-experiment-page-change/spec.md
oidacra: Fixed — and it turned out to be the same root cause as the failing issue-link check.

The branch is now `issue-37176-draft-experiment-page-change` and the spec's `**Feature Branch**` field matches it. I used branch naming rather than a `Closes #37176` keyword on purpose: this is the spec PR (1 of 2), so a closing keyword would close the issue on merge, before the implementation exists.

One consequence I should flag: renaming the branch closed this PR, because GitHub could not re-point an open PR at the renamed ref. Continued in #37206 — sorry for the re-review.

specs/37176-draft-experiment-page-change/spec.md
oidacra: Good question, and it found a real gap — the spec asserted this boundary without ever justifying it. I've now written the reasoning into the Scope Note (#37206).

The short version: the answer is still no, but not for the reason the spec implied. It isn't about data collection. Starting an experiment validates that its page is free of overlapping experiments (`validateNoConflictsWithScheduledExperiments`, plus the running-experiment-on-page check in `ExperimentsAPIImpl.start()`), and it validates that against *the page the experiment had at that moment*. A SCHEDULED experiment has already passed that gate and is committed to starting on its own via the quartz job.

Let it change page and the clearance goes stale: the new page may already host a scheduled or running experiment with an overlapping window, and nothing re-checks before the job fires it.

That's exactly why FR-007 can say we don't re-validate schedule conflicts on a page change. Safe for a DRAFT, which has never been cleared and will be checked at start. Not safe for SCHEDULED. DRAFT is the last state in which the page is still un-cleared — that's what makes it the boundary.

Admitting SCHEDULED would mean either accepting that hole, or re-validating conflicts here — and the second contradicts FR-007 and is a bigger change than this issue.

Worth adding: in practice a scheduled experiment would rarely qualify anyway. `start()` requires `hasAtLeastOneVariant`, which is really `variants().size() > 1` (`ExperimentsAPIImpl:1577`) — the control plus a real variant — so the variants condition would refuse it even if status allowed it. The state is reachable only by deleting variants after scheduling, which `deleteVariant` permits; that inconsistency looks like its own issue, separate from this one.

oidacra: @freddyDOTCMS you're right, and I understated this in my reply above — I called it "its own issue, separate from this one" when it's actually the strongest argument for where the boundary goes.

I went and checked what happens when the quartz job fires a SCHEDULED experiment. `startScheduled()` re-validates only three things: the license, the user's page-edit permission, and that the status is still SCHEDULED. It does **not** re-check `hasAtLeastOneVariant`, it does not re-check that a goal is set, and it does not re-check page conflicts. It goes straight to `save(... RUNNING)` and `innerStart`.

So a SCHEDULED experiment is a promise that was already validated — against the state it had at scheduling time. Every rule that made it valid ran then and never runs again.

That means your scenario is worse than "needs a check": deleting a scheduled experiment's variants is permitted today, and the experiment will still fire and become a RUNNING A/B test with nothing to test. **That is reachable right now, with no page change involved at all.** It's a pre-existing defect, not something this feature introduces — but it is exactly why this feature should not add a second way to invalidate an already-validated promise.

I've written that reasoning into the Scope Note (#37206). Two things I'd like your read on:

1. Should I file the `startScheduled` / `deleteVariant` gap as its own issue? I'd rather not fold it into #37176 — it has a different blast radius (it can already produce a running experiment with no variants) and deserves its own fix and tests.
2. If we ever do want SCHEDULED page changes, the prerequisite is that `startScheduled` re-validates at fire time. Worth capturing that as the dependency rather than leaving it implicit.

specs/37176-draft-experiment-page-change/spec.md
oidacra: Not out of scope — that flow is fully supported, and it's the whole point of the message in FR-009.

The tooltip on a draft that has variants reads: *"This experiment already has variants. Delete them to change the page."* Deleting them and then changing the page is precisely the remedy that message sends the user to. It would be a bad spec if the instruction didn't work.

Mechanically it falls out of how the rule is written: eligibility is evaluated against the experiment's state **at the moment of the request**, never against its history. Once the last non-control variant is gone, the experiment is a DRAFT holding only the control, and it qualifies like any other. An experiment that *used to* have variants isn't penalised for having had them — the orphaned-layout risk left along with the copies.

That said, you and @erickgonzalez both read the spec and landed on this question, which means the spec was leaving it to inference. Fixed in #37206:

- a new edge case stating it explicitly, and
- a third acceptance scenario on User Story 3: *given a draft with a non-control variant, when the editor deletes the last one, then the page selection control becomes enabled* — so the remedy loop is closed by a test rather than by assumption.

oidacra:
Closing in favour of #37206, which carries the same work on `issue-37176-draft-experiment-page-change` — a branch name that carries the issue number, as @erickgonzalez asked.

The four review comments here are all addressed in that PR:

- **Missing issue number** — the branch now carries it, which is also what makes the issue-link check pass. A `Closes #37176` keyword would have gone green too, but it would close the issue when the spec merges, before the implementation exists.
- **SCHEDULED experiments** (@erickgonzalez, @freddyDOTCMS) — the rationale is now written into the Scope Note instead of asserted, and the broader question is filed as #37214, with #37215 to establish the real state machine first.
- **Delete the variants, then change the page** (@freddyDOTCMS) — in scope; now an explicit edge case plus a User Story 3 acceptance scenario.

The threads here stay readable, and the replies are in them. Sorry for the churn — please continue the review on #37206.

PR #37182, opened 2026-08-24
fix(block-editor): register link, emoji and youtube regardless of Allowed Blocks (#37175)

### LINKED ISSUE #37175: Block Editor renders blank on edit when Allowed Blocks is configured
## Problem Statement

Any Block Editor field with **Allowed Blocks** configured renders **blank when editing** if
the stored content contains a link. The content is intact — the API returns the full
document — but the editor aborts parsing and boots an empty document with the
"Type '/' to insert a block…" placeholder. The same content renders correctly on live.

https://github.com/user-attachments/assets/35c43403-6acc-4371-b330-ee33a9ad8a43

```
RangeError: There is no mark type link in this schema
    at t.fromJSON → markFromJSON → nodeFromJSON
```

**Root cause:** `createEditorExtensions()` gates the Link extension behind the Allowed
Blocks list (`core-web/libs/new-block-editor/src/lib/editor/extensions/editor-extensions.ts:118`),
but `link` is a **mark** and the Allowed Blocks UI only offers node options
(`core-web/libs/block-editor/src/lib/shared/utils/suggestion.utils.ts:115`) — `link` is not
in that list and cannot be selected. So **every** field with a non-empty `allowedBlocks`
drops the Link mark from the schema, and TipTap's `Node.fromJSON` throws on the first link
mark it meets, aborting the entire document. No configuration avoids it.

**Scope:** new Block Editor only (`FEATURE_FLAG_NEW_BLOCK_EDITOR`); the legacy editor
registers `Link` unconditionally. Affects every restricted field whose content contains a
link — reported independently through two tickets (#38997, #38993).

**Data-loss risk:** no error is surfaced — the field simply looks emptied. If an author
types anything and saves, the original content is overwritten and permanently lost. Until
then the stored JSON is intact; this is a load-path failure only.

Second occurrence of this failure mode: #37145 was the same crash from an unregistered
`highlight` mark. Unknown *nodes* degrade to a placeholder that round-trips; unknown
*marks* have no equivalent and abort the document.

Introduced by #35257 (`669631d735`), where `link` was swept into the same `has()` list as
the node extensions. The legacy editor never supported restricting links, so this does not
look like a deliberate product decision.

## Steps to Reproduce

2. On a content type with a **Block Editor** field, open the field settings → **Allowed
   Blocks** → select any subset, e.g. `Bulleted List`, `Numbered List`, `Code Block`,
   `Table`. Note that **Link is not among the options** — there is no way to include it.
3. In a content item, add a paragraph containing a **hyperlink** to that field, and save.
4. Reopen the content item for editing.

**Actual:** the field is empty with the placeholder; the toolbar has no link button; the
console logs the `RangeError` above alongside the full document JSON that was received.
Live renders the content correctly.

**Expected:** the stored content loads. Allowed Blocks restricts which blocks an author can
*insert*; it must never prevent existing content from *loading*.

## Acceptance Criteria

- [ ] A field with Allowed Blocks configured loads content containing links, with the links
      editable as links.
- [ ] The Link mark is registered in the schema regardless of `allowedBlocks`.
- [ ] When `link` is not allowed, the restriction is still enforced on the *authoring* paths
      (toolbar button hidden, no autolink, no link-on-paste) without touching the schema.
- [ ] Existing fields are fixed by the code change alone — no content type or content
      migration required.
- [ ] An unknown mark no longer aborts the document: the rest of the document loads and the
      affected text survives a save.
- [ ] `emoji` and `youtube` are gated the same way while also absent from the Allowed Blocks
      options — reviewed and fixed (these degrade rather than crash, being nodes).
- [ ] Regression tests: with a restricted `allowedBlocks`, the schema contains `link` and a
      document with a link mark parses; a document with an unknown mark does not abort.

## dotCMS Version

`main` (2026-08-24). Present since #35257 (2026-05-08) — every build where
`FEATURE_FLAG_NEW_BLOCK_EDITOR` is available.

## Severity

**High - Major functionality broken.** The field is unusable and appears to have lost its
content; a save on top makes that loss real. Not Critical because it is scoped to fields
with Allowed Blocks configured and there is a no-deploy workaround: clearing Allowed Blocks
on the field re-registers every extension and restores the content without touching stored
data.

## Links

- [Freshdesk #38997](https://dotcms.freshdesk.com/a/tickets/38997)
- [Freshdesk #38993](https://dotcms.freshdesk.com/a/tickets/38993)
- Related: #37145 — same crash from an unregistered `highlight` mark
- Introduced by: #35257 (`669631d735`)

adrianjm-dotCMS:
## Proposed Changes

Fixes #37175 — a Block Editor field with **Allowed Blocks** configured rendered blank on
edit whenever the stored content contained a link.

`createEditorExtensions()` gated `link`, `emoji` and `youtube` behind the field's Allowed
Blocks list. None of the three is offered as an Allowed Blocks option — that list comes
from `getEditorBlockOptions()`, which only exposes block nodes — so gating on those keys
never restricted anything. It dropped the extension on **every** field with a non-empty
`allowedBlocks`, with no configuration able to avoid it.

`link` is a **mark**, so its absence made `Mark.fromJSON` throw mid-recursion and abort
`Node.fromJSON` for the whole document. The field rendered blank even though the stored
JSON was intact and the live site rendered correctly. Authors saw what looked like deleted
content, and saving on top of it would have made that loss real. `emoji` and `youtube` are
nodes, so they degraded to `Unsupported block (…)` placeholders rather than crashing.

All three now register unconditionally, so stored content always loads. The restriction
moves to the authoring paths, where it actually belongs:

| Extension | Schema | Authoring gate |
|---|---|---|
| `link` | always registered | toolbar button (already gated) + `autolink` + `linkOnPaste` |
| `emoji` | always registered | toolbar button (already gated) + `enableEmoticons` (`:)` → 😀) |
| `youtube` | always registered | "Add asset by URL" popover, already gated by `showAssetByUrl()` |

A field that excludes links still cannot create them — it just stops destroying the
content it already has.

The invariant is documented once on `has()` rather than at each call site, so the next
gated extension gets checked against the Allowed Blocks options first.

This is the second occurrence of this failure mode; #37145 was the same crash from an
unregistered `highlight` mark.

### Deliberately out of scope

- **Unknown marks still abort the document.** Unknown *nodes* degrade via the
  `dotUnsupportedBlock` placeholder, which round-trips the original JSON; marks have no
  equivalent, so any future missing mark reproduces this class of bug. Hardening that means
  threading a `knownMarkNames` set through `preserveUnknownNodesInDocument` and deciding
  whether an unknown mark is discarded (loses formatting on the next save) or preserved via
  a `dotUnsupportedMark` placeholder. That is a schema-affecting change and deserves its own
  review — tracked in the ACs of #37175.
- **Pasting HTML that contains `<a>` into a link-restricted field.** `linkOnPaste` only
  covers pasting a raw URL over a selection; HTML paste goes through the mark's `parseHTML`.
  Previously the mark was absent from the schema so those links were silently stripped; now
  they are kept. Closing it needs a `transformPasted` handler — notably **not**
  `appendTransaction`, which would also run on `setContent` and strip the very links this PR
  restores.

## Issue

Fixes #37175

Support tickets: [Freshdesk #38997](https://dotcms.freshdesk.com/a/tickets/38997) ·
[Freshdesk #38993](https://dotcms.freshdesk.com/a/tickets/38993)

## Verification

Frontend-only; no schema, API contract or DB change, so this is rollback-safe and a clean
backport candidate.

- `pnpm nx test new-block-editor` → **101 passed** (12 suites)
- `pnpm nx lint new-block-editor` → clean
- `pnpm nx format:write` → no changes

The 6 new specs were verified to **fail without the production change** (fix stashed, suite
re-run), including the customer-facing error verbatim:
`RangeError: There is no mark type link in this schema`. Two of them assert the restriction
is still enforced on an unrestricted vs. restricted field, guarding against over-correcting
by leaving `autolink` off everywhere.

Manually verified against a local dev server with `FEATURE_FLAG_NEW_BLOCK_EDITOR` enabled:
a field restricted to lists + code block, holding content with a hyperlink and an emoji,
loads both instead of rendering blank / `Unsupported block (emoji)`.

### Suggested reviewer checks

1. Restricted field + stored link → content loads, link editable.
2. Same field → no link button in the toolbar, and typing a bare URL does not auto-link.
3. Same field → no emoji button, and `:)` does not auto-convert.
4. Unrestricted field → autolink, link-on-paste and emoticons all still work.


adrianjm-dotCMS:
Closing this one to open https://github.com/dotCMS/core/pull/37205

Issue #37184, opened 2026-08-24
Content Drive: field filter re-scans the folder candidate query up to 4x per request


ihoffmann-dot:
## Description

When a Content Drive field filter (e.g. a Text field contains-match) is fully resolvable against the search index — no Tag, no Relationship, no workflow, no free-text term — the system still repeats the expensive database candidate-scan query multiple times per request instead of once. On a ~21,000-item folder this took ~4 repeats and made the response ~6x slower than an equivalent search (606ms vs 88ms).

The routing to the index is correct per ADR-0018 (fields flagged `searchable` on the content type are index-resolved by design — this is not a mis-routing bug). The inefficiency is in the DB-first hybrid chunk loop, which keeps re-scanning in fixed-size chunks (`BROWSER_CONTENT_CHUNK_SIZE`, default 900) until enough matches survive index narrowing, regardless of match density.

`buildPureESQuery` (BrowserAPIImpl) already builds nearly everything needed for a single-pass, index-only query for this specific case, but is missing a folder-scoping clause (`conFolder`, an already-used, low-risk ES field). Adding it and routing to this path only when zero DB-routed criteria, no workflow filter, and no free-text term are present should collapse the repeated scans to one.

----

Adding some independent measurements that corroborate this issue, gathered testing against a large dataset (~718,174 contentlets):
- Filtering by a specific searchable field (Content Type: File Asset + File Name filter) took ~2 minutes to return a response, versus ~31.5 seconds for the equivalent global search filter term — roughly 3.8x slower, consistent with the field-filter re-scan behavior described here.
- Combining a secondary-language filter with the global search filter took ~12 seconds to generate a response.
- A plain global search in Content Drive took ~25 seconds to return results, versus ~24ms for the same term in the Search All portlet.

These numbers are from a larger dataset than referenced in the issue description, so the relative slowdown compounds further at scale — an additional data point for the "falls within 20% of an equivalent content-search operation" acceptance criterion.

## Acceptance Criteria

- [ ] A field filter with no DB-required criteria over a large folder resolves with at most one database candidate-scan, not four
- [ ] Response time for that case falls to within 20% of an equivalent content-search operation
- [ ] Filters combined with Tag, Relationship, workflow, or free-text terms are unaffected — same result sets, same database-first resolution, no behavior change
- [ ] Result sets for Text, Date-range, Multi-select, and Category field filters are unchanged from current behavior
- [ ] The mandatory read-permission filter continues to run, sourced from the database, on every request
[...]

Issue #37148, opened 2026-08-20
Content Drive: folder listing query scans by mod_date instead of by folder (450ms to 101ms), plus 3 related performance items


ihoffmann-dot:
## Description

Content Drive is set to replace Content Search and Site Browser. We benchmarked it against both on a large production-scale dataset (~418k contentlets, ~14k folders, ~411k identifiers) running in local Docker, measuring endpoints and UI with interleaved A/B execution, warm-up discarded, 15 iterations per case, and two roles (administrator and a permission-limited user).

**Content Drive comes out ahead where it matters most.** Against Site Browser it wins all 34 cases — on a folder with ~4,200 listable assets Site Browser takes 7.5 s while Drive takes ~110 ms, because Drive paginates in SQL while Site Browser fetches everything and slices. In the UI, opening the portlet takes 247 ms vs 1038 ms for Content Search (−76%).

**The gap is in filtered search over large folders**, where Drive is up to 6× slower than an equivalent `/api/content/_search` query (705 ms vs 117 ms). Profiling with `pg_stat_statements` and Glowroot traced it to a single dominant cost, plus three smaller ones.

### Root cause: the candidate-scan query enters by `mod_date` instead of by folder

`BrowserAPIImpl` builds this query to collect candidate inodes for a folder listing:

```sql
select cvi.working_inode as inode
from contentlet_version_info cvi, identifier id, structure struc, contentlet c
where cvi.identifier = id.id
  and struc.velocity_var_name = id.asset_subtype
  and c.inode = cvi.working_inode
  and cvi.variant_id = ?
  and struc.inode not in (?)
  and cvi.lang in (?)
  and id.parent_path = ?
  and cvi.deleted = ?
order by c.mod_date asc LIMIT ? OFFSET ?
```

The `ORDER BY` is on `contentlet.mod_date` while the selective filter is on `identifier.parent_path` — a different table. No index can satisfy both, so the planner enters through `idx_contentlet_mod_date` and hopes to find enough matches early. On a folder whose content is recent, with an ascending sort, that bet fails badly:

```
Limit (actual time=55.169..552.369 rows=400 loops=1)
  Buffers: shared hit=909424
  ->  Nested Loop  (rows=400)
        ->  Nested Loop  (rows=98349)
              ->  Index Scan using idx_contentlet_mod_date on contentlet c  (rows=104277)
              ->  Index Scan using idx_contentlet_vi_working on contentlet_version_info cvi
                    Index Searches: 104277
[...]

Issue #36814, opened 2026-07-30
Content Drive: Search & Performance at Scale


fmontes:
## What we need

Content Drive has to stay fast and return the right thing on instances with millions of assets. Today it does neither.

- **Search and pagination can take up to seven minutes** on large repositories.
- **Search returns results that don't contain the search term** — searching `IMG` for a file named `IMG_1004.jpg` surfaces unrelated files such as `og-preview.vtl`.

Both hit the everyday path, not an edge case: browsing a folder, moving to the next page, typing part of a filename you know exists.

## Why it matters

Content Drive replaced Site Browser and Content Search with one workspace. On a demo instance it behaves well. On a real enterprise repository it doesn't, and at that point the unified workspace is worse than the tools it replaced — at least Content Search came back quickly.

Seven minutes is past the point where a user assumes the page has hung. And once a result set contains entries a user can't account for, the whole result set stops being trustworthy — so they search again with different terms, which multiplies the cost of the slowness.

Asset filenames matter here specifically. They're a primary way people locate things, especially developers looking for VTLs, CSS, and theme files.

## What "done" looks like

On an instance with millions of assets:

- A user types a fragment and gets matching results in a time that feels interactive.
- The fragment is found whether it lives in the **filename, title, metadata, path, or URL** — all five.
- Every returned result visibly relates to what was typed, and it's clear which field matched.
- Paginating through a large folder is equally fast.

## Acceptance criteria

- [ ] Search matches filename, title, metadata, paths, and URLs
- [ ] Results are ordered by a defined and documented precedence between those five — a filename match should not rank equally with a metadata match
- [ ] The matched field is visible in the result row, so a path or metadata hit doesn't read as noise
- [ ] Partial and prefix filename matching works, including underscores and extensions
- [ ] A known filename is findable by typing a leading fragment of it
- [ ] Every result matches the entered term in a field search is documented to cover — no unexplainable results
- [ ] Search and pagination meet an agreed response-time target at millions-of-assets scale
- [ ] Search behaves correctly when combined with the existing filter chips
- [ ] Behaviour is verified at that scale, and protected against regression

## Out of scope

- **The filter chip UX** — owned by #35430 (Content Drive: Search). Same surface, so the two need coordinating.
- **New search capabilities** — semantic or vector search, natural-language queries, AI-assisted search. This is about making literal search work; anything smarter builds on it.
- **New metadata or DAM features** — auto-tagging, custom metadata schemas, where-used tracking.
[...]

PR #37187, opened 2026-08-24
docs(content-drive): spike spec for folder-listing query plan instability (#37183)

### LINKED ISSUE #37183: Spike: Content Drive folder-listing query has planner-driven latency on very large folders
## Research Question

Why does the Content Drive/Site Browser folder-listing candidate-scan query take 564ms on a ~21,000-child folder but stay under 20ms on folders up to ~4,000 children — and can that be fixed without regressing the common (small/medium folder) case?

## Timebox

4h

## Context

Root cause identified as PostgreSQL planner instability tied to `LIMIT`: below ~500 rows the planner walks an index on the sort column hoping to satisfy the limit early; above it, it switches to gather+sort, which is ~10x faster on large folders. The production chunk size (400) lands at the worst point of the slow plan.

A candidate rewrite (enter via folder instead of sort column) is 4.3x faster on the largest folder but 3x SLOWER on a mid-size one — rejected as a blanket fix. Raising the chunk size was also tested and rejected (regressed a mid-size folder by 42%).

Latency on the largest folder is also bimodal (some requests ~165ms, ~27% spike to ~587ms) independent of the above, cause unidentified.

## What we need to learn

- Which of three candidate directions (schema/denormalization change, plan-stabilization without rewriting the query, or a distinct large-folder code path) is viable, and at what cost/risk
- What is causing the latency bimodality on large folders

## Acceptance Criteria

- [ ] Findings and a recommended direction (or explicit "needs a follow-up prototype/design work") documented as comments on this issue
- [ ] A recommended direction is agreed with the team before any implementation work starts
- [ ] Coordination constraint with the field-filter chunking fix (#37184) is preserved: whichever direction is chosen must not break the single-scan-per-request fast path that fix depends on

## Note on process

This is a spike — per team convention, spikes don't carry a formal Spec-Kit spec/PR. Investigation and findings are tracked as comments on this issue directly.

## Parent Epic

#36814

ihoffmann-dot:
**Spec-Kit PR 1 of 2 — Spike.** Carries the spike spec only; this investigation does not prescribe a fix (see below). Needs a developer **approval** (not a merge) before /speckit-plan proceeds on whichever direction the team picks.

Resolves the spec phase of #37183.

### Proposed Changes
* `spec.md` — problem statement, investigation to date (measured facts), three candidate directions with tradeoffs and open unknowns per direction (none verified — no live instance was available to prototype), a hypothesis-ranking on the unexplained latency bimodality on large folders, and a recommendation for what a follow-up prototype should measure before committing to a design.

### Why a spike, not a feature spec

This started as a proposed fix (a folder-first query rewrite) that measured 4.3x faster on the largest folder tested — but 3x **slower** on a mid-size folder, a real regression verified with identical result sets in both directions. The root cause isn't a bad query, it's PostgreSQL planner instability tied to `LIMIT`: below ~500 rows the planner walks the sort-column index hoping to satisfy the limit early; above it, it switches to gather+sort, ~10x faster on large folders. The production chunk size (400) sits at the worst point of the unstable range. A configuration-only mitigation (raising chunk size) was also tested and rejected — it regressed a different mid-size folder by 42%.

None of the three candidate directions in the spec is picked as "the" fix. That decision needs either a prototype or a plan-phase design spike, informed by this document.

### Coordination note

Shares the underlying query with #37184 (field-filter chunk multiplier), already spec'd separately. Whatever direction is chosen here must preserve that fix's single-scan-per-request assumption.

### Checklist
- [ ] Direction chosen and a follow-up plan/prototype scoped, before /speckit-plan proceeds
- [ ] Latency-bimodality hypothesis (plan-cache generic/custom plan flip on this same query, unconfirmed) investigated with a live instance

### Additional Info
Parent epic #36814. Originally investigated as item 1 of #37148 (umbrella investigation issue covering 4 related performance items in Content Drive/Site Browser).

`/speckit-adr-context` consulted ADR-0018 (database-first search for Content Drive) — none of the three candidate directions moves folder-scoping or ordering resolution to the search index, so no conflict identified. **No new ADR proposed** at this stage; one of the candidate directions (a) is itself a possible schema change that, if pursued, likely warrants its own ADR before implementation.

This PR fixes: #37183

### DIFF
--- a/specs/37183-content-drive-folder-listing-plan-instability/spec.md
+++ b/specs/37183-content-drive-folder-listing-plan-instability/spec.md
@@ -1,0 +1,422 @@
+# Spike Specification: Content Drive Folder-Listing Query — Planner Instability at Scale
+
+**Feature Branch**: `37148-content-drive-folder-listing-plan-instability`
+
+**Created**: 2026-08-24
+
+**Status**: Draft
+
+**Type**: Spike / Investigation (not a design or implementation spec — no fix is prescribed here)
+
+**Related GitHub Issue**: [#37183](https://github.com/dotCMS/core/issues/37183). Parent epic: [#36814](https://github.com/dotCMS/core/issues/36814). Originally investigated as item 1 of [#37148](https://github.com/dotCMS/core/issues/37148) (umbrella investigation issue); items 2–4 of that investigation are tracked as separate sibling issues and are out of scope here except where they create a coordination constraint (see "Coordination Risk").
+
+**Input**: Issue #37148 description + comment 1 (which supersedes the description's framing for item 1) + comment 2 (context only) + comment 3 (context only, item 4).
+
+<!--
+  No dedicated spike template exists in .specify/templates/ (only spec-template.md for
+  features and spec-issue-template.md for defects; .github/ISSUE_TEMPLATE/spike.yaml is a
+  GitHub issue form, not a spec-kit spec template). This document reuses the
+  issue-resolution shape where it fits (Problem Statement, Scope, Regression Risk) but
+  replaces the "Root-Cause Hypothesis → Fix Scope" sections with the sections the task
+  explicitly asked for, since a spike's deliverable is a decision-ready investigation
+  record, not an acceptance-tested fix. All findings below come from reading code and from
+  the measurements already recorded in issue #37148; there is no live/profiled instance
+  available in this environment, so every candidate direction is a code-reading-informed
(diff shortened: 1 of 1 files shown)


ihoffmann-dot:
Cerrado sin merge: la convención del equipo es no crear specs formales para spikes. El contenido de esta investigación queda documentado directamente en el issue #37183 en su lugar.

PR #37216, opened 2026-08-25
feat(content-drive): Status filter (Archived, Unpublished, Locked), three regression fixes, and a Refresh role gate (#37066)

### LINKED ISSUE #37066: Content Drive: Status filter (Archived, Unpublished, Locked) on drive search and in the toolbar

zJaaal:
**Implementation PR, stacked on #37170** (the approved spec) — stack #37217. Review #37170 first; this diff is the feature only.

Resolves #37066.

> **Ready for a Docker image.** Backend + frontend are complete and the unit/Jest suites are green. The **integration tests have not been run** — see Verification below. Manual testing against a build is genuinely useful here.

### What it does

An optional `status` array on `POST /api/v1/drive/search`, plus a multiselect chip in the Content Drive toolbar. Selected statuses combine with **OR** — checking more boxes returns more content, matching the content-type and locale filters beside it.

```
[shared assets] [content type] [workflow] [status] [locale]
```

### The five things most worth reviewing

**1. One OR-ed group, with the archived baseline OUTSIDE it.**
```sql
  and cvi.deleted = false                                        -- baseline, unless ARCHIVED selected
```
Folding the baseline into the group would make `[UNPUBLISHED, LOCKED]` read `(deleted = false or …)` and match nearly every row — a filter that silently stops filtering.

**2. Lucene needs an explicit group.** `+` means REQUIRED, so `+deleted:true +live:false` is an **AND**. `buildPureESQuery` emits `+(deleted:true OR live:false)`. Caught by @nollymar in review on #37170; it follows the convention already in that same method (`+(conhost:… OR conhost:SYSTEM_HOST)`).

**3. An empty selection emits nothing at all.** Not an empty group — `and ( )` is a SQL syntax error and `+()` is invalid Lucene. This is the *default* path: every drive search that exists today sends no status, so an unfiltered request stays byte-identical.

**4. `showWorking` now covers `ARCHIVED`/`UNPUBLISHED`.** Neither state has a live version, so without this the query joins `live_inode` and returns nothing — silently, with no error. `LOCKED` doesn't need it; a locked item may well be live.

**5. `ARCHIVED` suppresses the archive-step reconciliation** the same way `showArchived` does, so it can't contradict `appendWorkflowQuery`'s per-branch `cvi.deleted`.

### A real bug the tests caught

`parseStatuses` originally built its 400 message with `String.format` and handed the result to `BadRequestException`. But `HttpStatusCodeException` runs `String.format` over the message **again** — so `status: ["50%"]` raised `UnknownFormatConversionException` and surfaced as a **500 instead of a 400**. User input was reaching a format string. The value now travels as a format *argument*, with a regression test.

The same test also showed `BadRequestException.getMessage()` returns `"HTTP 400 Bad Request"`; the useful text is in the `error-message` header, which is what a client actually sees.

### Regressions found while testing, fixed here

Three Content Drive issues surfaced during manual testing of this filter. Carried in rather than deferred, at @zJaaal's call, so the portlet ships whole.

**1. A loading dialog could not be closed.** The folder dialog's footer sat inside `@if ($formReady())`, so while the form loaded there was no Cancel button and no way out but the backdrop. The footer is now outside the guard and Cancel always renders; only the submit button waits on the form.

**2. The toolbar lost its bottom border.** Traced to `border-none!` on the toolbar, added to strip `p-toolbar`'s default border on the other three sides. Narrowed to `border-x-0! border-t-0!`, which keeps the bottom edge that separates the toolbar from the table. Scoped to Content Drive only.

**3. The context menu was one undifferentiated list.** It now reads as three named groups:

```
ACTIONS                 <- caption
Edit Content
Lock
Push Publish
Add to Bundle
────────────────
WORKFLOWS               <- caption
Assign Workflow
Save
Save / Publish
────────────────
Archive                 <- separator alone, no caption
```

A folder gets the same treatment: an **Actions** caption, and Delete held apart below a separator. There the caption is unshifted after the fact rather than pushed first — every entry on a folder is permission-gated, so that is the only point where the group is known to be non-empty.

On the caption itself: `p-contextMenu` has **no group-label class at all** — its style map is `item`, `separator`, `submenu`, `submenuIcon`, and it renders `items` as a flyout. PrimeNG's inline group label (`p-menu-submenu-label`) lives on `p-menu`. So the caption is a regular item wearing that class, made inert by `disabled` (PrimeNG's `isValidItem` skips disabled items, keeping it out of keyboard navigation) plus `pointer-events-none` (it would otherwise take a click and close the menu). Three utilities cancel what the surrounding styles impose: `p-0!` so it aligns on the item padding instead of being inset twice, `pointer-events-none`, and `text-inherit` on the content wrapper, since `.p-contextmenu-item-content` sets `color` on a descendant and would win.

Switching wholesale to `p-menu` for its native grouped model was considered and rejected: `p-menu` popup calls `absolutePosition(container, target)` and anchors to an **element**, while `p-contextMenu` positions at `event.pageX/pageY`. A right-click menu opening at the row instead of the cursor is a worse regression than a faked caption row.

The destructive split reads `hasArchiveActionlet || hasDeleteActionlet || hasDestroyActionlet` — the action's **actual sub-actionlets**, never its name, so a scheme's "Retire this blog" or "Purge" lands there too. A parameterized test names the actions nothing like Archive or Delete precisely to hold that line. A folder's Delete gets the same separator; its gate is strictly narrower than Edit Folder's, so it can never lead the menu.

Finer grouping than this is not possible from the data: the API exposes no actionlet class names, no category and no tag, `order` is a within-scheme sort index, and `icon` is admin-authored free text. Anything more would be guesswork that breaks on custom schemes.

### Frontend notes

- **The whole row is clickable, label included.** First cut bound `[ngModel]` on the listbox without `(ngModelChange)`, so only the checkbox responded. Regression test clicks the label specifically.
- **`decodeByFilterKey` needs the explicit `status` entry.** Without it a lone `status:ARCHIVED` decodes to the *string* `'ARCHIVED'` — whose `.length` is `8`, so every `?.status?.length` guard misreads it and the filter looks fine until someone selects exactly one status.
- **Placement:** after Workflow, before Locale. Content type and workflow stay adjacent because the workflow filter derives its scheme list from the content-type selection — the row's only real dependency.
- **`archived: false` pin removed** from the store request (FR-019); the endpoint already defaults it, and pinning it would contradict an Archived selection.

https://github.com/user-attachments/assets/a0c39520-aaf0-4a03-b80a-da3ed5b0670b

### Two questions reviewers keep asking

Answering both here rather than in the source — the code comments already carry their weight.

**Why a new `ContentStatus` enum, when `WorkflowState` already has these names?**

`com.dotmarketing.portlets.workflows.model.WorkflowState` is `NEW, LOCKED, UNLOCKED, PUBLISHED, UNPUBLISHED, ARCHIVED, LISTING, EDITING` — all three of ours, same spelling. It looks like a duplicate and isn't:

- **It is a different concept.** Its own javadoc: the `show_on` set deciding whether a workflow *action* renders. That is why it also carries `LISTING` and `EDITING` — view contexts, not states a contentlet can be in, with nothing to resolve against in a query. Reusing it would put `status: ["EDITING"]` in the public API contract.
- **Its parsing contradicts ours.** `WorkflowState.toSet` catches any exception and returns an **empty set**, so one bad value silently drops the whole filter and returns a *wider* result than the caller asked for. This filter deliberately returns a `400` instead — the behaviour the tests and this PR's review cycle settled on.
- **It would couple the API surface to workflow internals.** Adding a `show_on` value would silently widen `/v1/drive/search`'s accepted input. We implement three states; that enum has eight and grows for unrelated reasons.

Same words, different job.

**Why does `decodeFilters` drop every empty array, not just `status`?**

Raised by the automated review, and fair: the behaviour is general while the comment reads `status`-specific. The general rule is intended.

Every consumer reads these through `?.length`, so `undefined` and `[]` already mean the same thing — there is no behavioural difference to preserve. Storing the key would re-encode a bare `status:` into the URL for the next decode to trip over. `status` is simply what surfaced it: sanitizing `status:BOGUS` is the first decode that can legitimately produce an empty array.

### Verification

| Suite | Result |
|---|---|
| `ContentDriveHelperStatusTest` (unit) | ✅ 8/8 |
| Frontend (`portlets-content-drive`, full) | ✅ 1304/1304 |
| Lint | ✅ clean |
| `tsc --noEmit` | ✅ no new errors (35 pre-existing, none in touched files) |
| `ContentDriveStatusFilterTest` (integration, 16 cases) | ✅ **16/16 in CI** |
| `ContentDriveWorkflowArchiveStepTest` (regression guard) | ✅ green after `b09fb47` — see below |

The integration class covers each status alone, every pair, all three, the empty default, the never-shrinks property and the archive-step regression. Registered in `MainSuite3a`.

**`MainSuite 3a` has now run, and it caught one thing — a bad assertion of mine, not a code defect.** 1 failure out of 803, deterministic across all three retries: `testUnpublishedStatusWithArchiveStepExcludesArchivedContent` claimed `UNPUBLISHED` + an archive-target step must exclude archived content. False — an archive-target step makes `appendWorkflowQuery` admit `cvi.deleted = true` rows in that branch by itself, with no status involved (`testMixedFilterScopesArchivedToArchiveBranch` pins exactly that and passes), and `ContentletAPI.archive` unpublishes so the row satisfies `live_inode is null` too. Both clauses match; returning it is correct.

Replaced in `b09fb47` by two tests rather than one: `…StillReturnsArchivedContent` asserts what actually holds, and `…KeepsTheGlobalArchivedBaseline` is the real complement, using a **normal** step where nothing lifts the baseline. Strictly stronger than what it replaced — it still guards the baseline and now documents the interaction that misled me.

The fix itself was not verified locally: `com.dotcms.tika-api`'s pom in the local repo is unflattened, so `:dotcms-integration` fails dependency resolution before any test runs. CI is the verification.

`PURE_ES` parity holds for `ARCHIVED` and `LOCKED`, and **deliberately not** for `UNPUBLISHED` — see below.

### Two things I won't paper over

**The TDD gates for US2–US5 are not satisfied.** I implemented all three SQL disjuncts in one switch rather than story-by-story, because splitting them leaves states where selecting `LOCKED` parses fine and filters on nothing. Their tests were therefore written against working code — characterization, not true Red. US1's gates were honored properly (Red confirmed, then implementation).

**`UNPUBLISHED` means something slightly different under `PURE_ES`, and we accepted that.** It means *no live version exists* — a question about the content, not about one version. The index stores `live` per version, so a published item with newer unpublished edits has a working document carrying `live:false`, which the index query matches and `cvi.live_inode is null` does not. There is no index-side fix short of redefining the status per-version, which would degrade the **default** path to match a limitation of a strategy nobody runs (`PURE_ES` is opt-in and is not set in any config file in this repo). ADR-0018 already routes structural predicates to the database for exactly this reason and states `PURE_ES` forfeits that guarantee. Recorded as **FR-009a** in the spec, with FR-009 and SC-005 narrowed to match, and a predicate-comparison table in `contracts/`. `ARCHIVED` and `LOCKED` are unaffected.

**`openapi.yaml` is unchanged, deliberately.** `/v1/drive` has **zero** entries in the generated spec, unlike `/v1/folder` (8), `/v1/browser` (4) and `/v1/workflow` (45) — `ContentDriveResource` is excluded from OpenAPI generation entirely. The `@Operation` text was updated and is good source documentation, but it does not reach the yaml. Pre-existing gap; flagging rather than working around it, and worth its own ticket.

### Unrelated polish: Refresh is now role-gated in the UI (#36845)

Not part of the status-filter spec. Carried here rather than held back, so it is in the demo build.

QA on `POST /api/v1/content/_bulkrefresh` turned up that a non-admin sees Refresh in Quick Actions,
fires it, and gets a 403. That was a deliberate call at the time: `BulkRefreshHelper.canRefresh`
reads as *Power User OR Administrator*, and the browser cannot tell whether someone is a Power User,
so gating on `isAdmin` looked like it would hide the action from people entitled to it.

It does not, because the Power User half never fires. It resolves the role **key** `"CMS Power User"`,
no role ships with that key, `loadRoleByKey` answers null, and `RoleFactoryImpl` returns false for a
null role. What survives is `doesUserHaveRole(user, loadCMSAdminRole())`, which is character-for-character
`User.isAdmin()`, which is what already feeds `currentUserIsAdmin` in the store. Client and server
evaluate the same expression, so the row can predict the refusal rather than discover it.

**Disabled, not hidden** — the same treatment Push Publish gets with no environment configured. The row
keeps saying the capability exists; the tooltip says who it is for.

| | Before | After |
|---|---|---|
| Non-admin | Row enabled, fires, 403 | Row disabled, tooltip names the requirement |
| Admin | Row enabled, fires | Unchanged |

Follows the existing gate shape exactly: `requiresAdmin` on the action def, `missingAdminRole` computed
in `getQuickActions`, added to the `[disabled]` binding, the `quickActionHint` tooltip and the
`onSelectQuickAction` guard. Three new `getQuickActions` tests plus three at component level; five
existing Refresh tests now set the admin precondition, since the behaviour they cover starts after the
gate.

**Known limit, left alone:** an install that hand-created a role keyed `"CMS Power User"` would let a
non-admin holder through server-side while this row stays shut. dotCMS ships no such role. The clean fix
is to drop the dead branch from `canRefresh` so the gate is admin-only *by definition*, but that is a
change to a server-side authorization path and does not belong in a demo-eve commit. Worth its own ticket.

### Also touched

Renamed the `status:published` placeholder in existing `decodeFilters` tests to `owner:jane`. They used `status` as a made-up key to exercise unknown-key fallback, and it is now a real key. Changing their expectations to arrays instead would have forced them green while destroying what they test.

### Checklist
- [x] Tests — unit + integration + Jest/Spectator. Integration not yet executed locally (see Verification)
- [x] Translations — four `content-drive.status-filter.*` keys in `Language.properties`, plus `content-drive.action-center.requires-admin` for the Refresh gate
- [x] Security Implications Contemplated — input is a closed enum, validated and rejected with a 400, never interpolated into SQL. The format-string bug above was found and fixed. Permission filtering untouched, so the filter grants no new visibility

This PR fixes: #37066


zJaaal:
Both 🟡 findings are real. Fixed in `eabb797` — and the second one surfaced something worse than the review could have known.

**🟡 Locale — fixed.** Exactly right. `"unpublished".toUpperCase()` under `tr-TR` yields `UNPUBLİSHED` and 400s a valid value on Turkish-locale servers only. Now `toUpperCase(Locale.ROOT)`, with a regression test that flips the default locale rather than trusting the reasoning.

**🟡 SQL vs PURE_ES `UNPUBLISHED` — confirmed, your model of the index is correct.** `ESMappingAPIImpl:523` sets `live` from `contentlet.isLive()`, i.e. per VERSION. So a published item with pending working edits has a working document carrying `live:false`, which `+working:true +(live:false)` matches, while SQL's `cvi.live_inode is null` excludes it. Two different questions: *"the identifier has no live version"* versus *"this version is not the live one"*.

**The verification you suggested could not have run.** There is no `PURE_ES` parity case in `ContentDriveStatusFilterTest`. It was listed as covered in `tasks.md` T031, in this PR's body and in `quickstart.md` — and was never written. That is my error, and the more serious of the two: I have been reporting coverage that does not exist. All three claims are corrected in this commit, with T031 un-ticked and marked ⚠️ NOT DONE.

The fixture was also blind to it — no published-then-edited item, so a parity test would have compared two identical answers and passed for the wrong reason. That fixture now exists, plus a test pinning the intended semantics on the default path: a published-then-edited item HAS a live version, so `UNPUBLISHED` must exclude it.

**The divergence itself is left open deliberately.** "The identifier has no live version" is not expressible in a per-document index query — `-live:true` does not help, because the working doc genuinely carries `live:false`. So the options are a spec decision, not an implementation one:

1. Accept it and narrow FR-009/SC-005, leaning on ADR-0018 already stating `PURE_ES` forfeits consistency guarantees for every criterion
2. Redefine `UNPUBLISHED` as "this version is not live" — matches legacy `ContentletAjax` and ES, but changes what users get on the **default** path
3. Reject or degrade `UNPUBLISHED` under `PURE_ES`

Recorded in `tasks.md` rather than guessed at. @zJaaal's call.

On your non-blocking note about `LOCKED` + `live:true`: agreed, and it is documented in `data-model.md` as deliberate — `LOCKED` does not force working-version scoping because a locked item may well be live, so `live:true` + `LOCKED` is the coherent "live content that is locked" query.

<sub>Analysis and reply by Claude (Claude Code), posted from @zJaaal's account.</sub>

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-status-filter/dot-content-drive-status-filter.component.ts
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /**
+     * Selection changed in the listbox. In multiple mode it hands back the full array of selected
+     * values, so this replaces rather than toggles — which also means clicking anywhere on a row,
+     * label included, works without a second code path.
+     */
+    protected onSelectionChange(selection: string[]): void {
+        this.$selection.set(selection ?? []);
+        this.#syncStore();
```
nicobytes: Why not run this.#syncStore() reactively instead of running for onSelectionChange and onClearAll? Maybe use an effect or signalMethod to run syncStore on selection signal changes?

nicobytes (dismissed):

dotCMS/src/main/java/com/dotcms/rest/api/v1/drive/AbstractDriveRequestForm.java
```diff
@@ -356,6 +356,32 @@ public interface AbstractDriveRequestForm {
...
+     * here: {@code ARCHIVED} is the only status that admits archived content, so
+     * {@code ["UNPUBLISHED"]} means "unpublished and not archived".
+     * <p>
+     * Defaults to empty, which is skipped entirely and preserves today's behavior exactly. An
+     * unrecognized value is rejected with a 400 by {@link ContentDriveHelper}, consistent with how
+     * {@link #userSearchable()} rejects unknown keys. Declared as strings rather than the enum so
+     * that rejection stays an explicit {@code BadRequestException} rather than a Jackson
+     * deserialization error.
+     * <p>
+     * A non-empty selection also drops folders from the results — folders carry no status.
```
oidacra: This javadoc says a non-empty selection "drops folders from the results", and the `@Operation` description in `ContentDriveResource` says "also excludes folders" — but `driveSearch` deliberately does not touch `showFolders`, and its inline comment explains why that is the caller's decision. `testStatusDoesNotOverrideExplicitShowFolders` and FR-015 both confirm the code is the correct side of this. Should these two doc strings drop the sentence?

Related: the log comment further down was edited to list "status" among the branches that override `showFolders`, which looks like it belongs to the same earlier version of the behavior.

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/drive/ContentDriveStatusFilterTest.java
```diff
@@ -0,0 +1,512 @@
...
+    /**
+     * A status combines with a keyword search rather than replacing it: the result is archived
+     * items matching the text (FR-009, US1 AC3).
+     *
+     * <p>Under the default {@code HYBRID_SINGLE_CHUNKED_QUERY_ES} heuristic the SQL supplies the
+     * candidate set and the index only narrows by text, so the status clause applies with and
+     * without a keyword.</p>
+     */
+    @Test
+    public void testArchivedCombinesWithTextSearch()
```
oidacra, line 273: `admitsArchived` was added specifically so an `ARCHIVED` selection cannot contradict `appendWorkflowQuery`'s per-branch `cvi.deleted` and collapse to an empty result — but nothing in this suite combines a status with a workflow filter pinned to an archive-target step, and `ContentDriveWorkflowArchiveStepTest` was not extended either.

Worth noting the review on #36666 raised the same untested interaction on that branch and it was accepted as non-blocking then too, so this would be the second pass where it goes unpinned. Given the integration suite has not been run yet, would it be worth adding a case that seeds an archive-target step and asserts `status=["ARCHIVED"]` still returns the item?

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/dot-content-drive.store.ts
```diff
@@ -152,12 +152,27 @@ export const DotContentDriveStore = signalStore(
                             folderCursor: page.folderCursor ?? 0,
                             maxResults: paginationSignal?.limit,
                             sortBy: sort()?.field + ':' + sort()?.order,
-                            archived: false,
+                            // Sent only when non-empty: an absent key leaves the request
+                            // byte-identical to one that never mentioned status, which is what
+                            // keeps the unfiltered drive exactly as it was. The `archived: false`
+                            // pin that used to sit here is gone — the endpoint already defaults it,
+                            // and pinning it would contradict an Archived selection.
+                            status: filters()?.status?.length ? filters()?.status : undefined,
```
oidacra, line 163: Not introduced here, but this PR makes it reachable in a new way. Further down, `loadItems`'s `catchError` sets `status: ERROR` and returns `EMPTY` with no logging and no user feedback, and `DotContentDriveStatus.ERROR` is not read by any template or component outside this store — only `LOADING` is checked. Since `items` is not cleared either, a failed search leaves a stale or blank grid with the spinner simply stopping.

Until now drive filters degraded server-side rather than returning a 400 (an unknown `contentType` id is dropped via `Try.of(...).getOrNull()`), but a hand-edited `?filters=status:BOGUS` now produces a real validation 400 that lands in that handler invisibly. It also masks any 401/403/5xx from the endpoint.

This feels like a separate ticket rather than something to widen this PR with — worth opening one?

dotCMS/src/main/java/com/dotcms/browser/ContentStatus.java
```diff
@@ -0,0 +1,36 @@
...
+ * A content state the Content Drive Status filter can narrow by.
+ * <p>
+ * The three are independent boolean facts about the same {@code contentlet_version_info} row, so a
+ * single contentlet may hold several at once. Selected statuses combine with <b>OR</b> — the filter
+ * asks whether an item is in <i>any</i> selected state, not all of them — matching the Content Type
+ * and Language filters beside it in the toolbar. Adding a status therefore never shrinks the result
+ * set.
+ * <p>
+ * <b>Excluding archived content is not a member of this set.</b> It is the browse query's standing
+ * default ({@code appendExcludeArchivedQuery}), applied on every request today. The selected
```
oidacra: Two small things in this javadoc.

"applied on every request today" describes a baseline that already had two exceptions before this PR (`showArchived`, and an archive-target workflow step), and the next sentence describes the third one this PR adds — so the phrase undercuts the paragraph it opens. Would something like "applied unless something explicitly lifts it" be closer?

Separately, the `UNPUBLISHED` member below reads "`cvi.live_inode is null`, indexed as `live:false`", which presents the two as the same predicate. The `publishedThenEditedItem` fixture javadoc in `ContentDriveStatusFilterTest` explains that they are not — one is identifier-scoped, the other per-version. Since this enum is where someone will look first, is it worth spelling that distinction out here too?

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-status-filter/dot-content-drive-status-filter.component.ts
zJaaal: Went with your suggestion — and with `signalMethod` rather than a bare `effect`, per @zJaaal. Final shape is `f2fcf3f`.

First, a correction: **my objection was wrong.**

I claimed a reactive sync would reset pagination. It would not. `$selection` returns the array *by reference* from the filter bag, so an unrelated `patchFilters` produces the same reference, the signal never notifies, and the sync never runs. On top of that `patchFilters` resets pagination itself, so a second write would have changed nothing. I asserted that from reading the code without checking reference identity. Not a deadlock either — with the store's current semantics the cycle settles in at most two passes.

What *was* real is narrower, and it is fragility rather than correctness: it terminated only because `#syncStore` passed the array by reference and `removeFilter` happens to be guarded by `if (removedFilter)`. Both incidental. A later `patchFilters({ status: [...selection] })` would mint a new reference on every write and turn it into a genuine infinite loop.

So the fix is on the `linkedSignal`, which is where the load-bearing guarantee belongs:

```ts
readonly $selection = linkedSignal<string[]>(
    () => {
        const raw = this.#store.getFilterValue(STATUS_FILTER_KEY);
        return Array.isArray(raw) ? raw : [];
    },
    { equal: (a, b) => a.length === b.length && a.every((value, i) => value === b[i]) }
);
```

Comparing by **value** means a write that does not change the selection produces no notification — so the cycle cannot start regardless of how the array is constructed downstream, rather than merely converging by luck.

The sync itself is then a `signalMethod`, matching the analytics filters and the UVE palette:

```ts
readonly #syncStore = signalMethod<string[]>((selection) => { ... });

constructor() {
    this.#syncStore(this.$selection);
}
```

It takes the signal directly and runs its body outside the reactive context, so the store writes cannot become dependencies of the signal that triggers them — no manual `untracked` to remember. Handlers now only `set()`, which is the maintainability win you were after: no way to change the selection and forget to persist it.

One more thing worth flagging: **the existing spec mocks the store**, so `getFilterValue` never reflects what `patchFilters` wrote and the feedback path cannot occur there at all. When I first tried the reactive version, all 154 tests passed — and proved nothing about the actual question. There is now a second `describe` block using the real store that exercises the write-back path directly, including that an unrelated filter change does not trigger a second write.

<sub>Analysis and reply by Claude (Claude Code), posted from @zJaaal's account.</sub>

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/drive/ContentDriveStatusFilterTest.java
```diff
@@ -0,0 +1,512 @@
...
+            throws DotDataException, DotSecurityException {
+        final Set<String> inodes = driveInodes(baseRequest()
+                .status(List.of("ARCHIVED"))
+                .filters(QueryFilters.builder().text("archived").build())
+                .build());
+
+        assertTrue("The archived item matching the keyword must be returned",
+                inodes.contains(workingInode(archivedItem)));
+        assertFalse("A non-archived item must not be returned even if it matches the keyword",
```
oidacra: This assertion passes trivially. Its message reads "even if it matches the keyword", but `liveItem` is titled "live" and does not match the keyword "archived" at all — so it would be excluded by the text filter alone, whether the status clause is ANDed, applied, or dropped entirely. A regression where the status clause is skipped whenever a keyword is present would still leave this test green, which is the FR-009 "same results with and without a keyword" direction.

Most other tests in this file guard against exactly this (several open with an `assertNotNull` on the fixture before asserting). Would seeding a second archived item that does *not* match the keyword, or a live item whose title does contain "archived", make the intersection observable and let the message describe what the fixture actually sets up?

dotCMS/src/main/java/com/dotcms/browser/BrowserQuery.java
```diff
@@ -152,7 +163,12 @@ private BrowserQuery(final Builder builder) {
         this.sortBy = UtilMethods.isEmpty(builder.sortBy) ? "moddate" : builder.sortBy;
         this.offset = builder.offset;
         this.maxResults = Math.min(builder.maxResults, MAX_FETCH_PER_REQUEST);
-        this.showWorking = builder.showWorking || builder.showArchived;
+        // ARCHIVED and UNPUBLISHED rows have no live version by definition, so the working inode
+        // must be the one joined (see selectQuery) or the join can never match and the filter
+        // silently returns nothing. LOCKED does not need this — a locked item may well be live.
+        this.showWorking = builder.showWorking || builder.showArchived
```
oidacra, line 176: The rationale for forcing `showWorking` is sound for `ARCHIVED`/`UNPUBLISHED` on their own — neither has a live version, so joining `live_inode` would silently match nothing.

The case that is not covered is the mixed one: since `workingLiveInode` picks the join for the whole query rather than per disjunct, `{"live": true, "status": ["LOCKED", "ARCHIVED"]}` resolves the `LOCKED` disjunct against `working_inode` too, so a live, locked item with pending edits comes back with its working version's fields rather than the live ones the caller asked for.

Unreachable from the Content Drive UI, since `live()` defaults to false — this only affects direct API callers. Neither `live()` nor `status()` mentions the interaction. Worth a line in one of those javadocs?

dotCMS/src/main/java/com/dotcms/rest/api/v1/drive/AbstractDriveRequestForm.java
zJaaal: Right on both counts — fixed in `12af6dc`.

When folder suppression moved to the client I updated `spec.md`, `contracts/` and `data-model.md` and left the Java untouched, so the javadoc and the `@Operation` description were both still describing the behaviour we deliberately removed. On a public API description that is the worst place for it to linger: a caller sending `showFolders:true` would have expected folders gone.

Both now say the field has no side effects and that `showFolders` is honoured as sent. You are right about the log comment too — it was edited in the same earlier pass and listed `status` among the branches that override flags, which contradicted the comment sitting two lines above it. Removed, with a parenthetical saying why status is deliberately absent from that list.

<sub>Reply by Claude (Claude Code), posted from @zJaaal's account.</sub>

dotCMS/src/main/java/com/dotcms/browser/ContentStatus.java
zJaaal: Both fair — fixed in `12af6dc`.

"applied on every request today" did undercut its own paragraph. Now "applied unless something explicitly lifts it", naming all three lifters (`showArchived`, an archive-target workflow step, and `ARCHIVED`).

The `UNPUBLISHED` one is the more important catch. Presenting `cvi.live_inode is null` and `live:false` as one predicate was wrong, and this enum is exactly where someone looks first — burying the distinction in a test fixture's javadoc was the wrong place for it. The member now spells out that the SQL condition is identifier-scoped while `live` is indexed per version, so a published-but-edited item's working document carries `live:false` and would match, and states that the identifier-scoped reading is the definition.

<sub>Reply by Claude (Claude Code), posted from @zJaaal's account.</sub>

dotCMS/src/main/java/com/dotcms/browser/BrowserQuery.java
zJaaal: Good catch — documented in `12af6dc`.

You are right that the join is picked for the whole query rather than per disjunct, so `{"live": true, "status": ["LOCKED", "ARCHIVED"]}` resolves the `LOCKED` disjunct against `working_inode` too, and a live locked item with pending edits comes back carrying working-version fields.

Added to the `showWorking` comment in `BrowserQuery`, and cross-referenced from `live()` on the form so a direct API caller reading that field finds it. Left as documentation rather than a behaviour change: making the join per-disjunct is a much larger change to `selectQuery`, and as you say it is unreachable from the UI.

<sub>Reply by Claude (Claude Code), posted from @zJaaal's account.</sub>

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/drive/ContentDriveStatusFilterTest.java
zJaaal: You are right, and worse than you knew: `tasks.md` T033 claimed this case was written. It never was. That is the second false completion in this feature — T031 (the `PURE_ES` parity case) was the first, also claimed and also absent. Both corrected in `tasks.md` to record what actually happened.

Two cases added in `12af6dc`, in `ContentDriveWorkflowArchiveStepTest` rather than the status suite, since it already owns the archive-target scheme and content archived at it — duplicating that fixture just to reach it would have been the wrong trade:

- `testArchivedStatusWithArchiveStepStillReturnsArchivedContent` — `ARCHIVED` plus the archive-target step still returns the archived item. This is the one that fails if `admitsArchived` regresses: `appendWorkflowQuery` forces `deleted = false` on its live branch while the status group asserts `deleted = true`, and without the suppression they contradict into an empty result.
- `testUnpublishedStatusWithArchiveStepExcludesArchivedContent` — the complement, pinning that only `ARCHIVED` lifts the baseline whatever the workflow filter says.

Noted that #36666 let the same interaction through. It will not be a third pass — though worth being straight that these have still never executed; `MainSuite 3a` in CI is the first real run.

<sub>Reply by Claude (Claude Code), posted from @zJaaal's account.</sub>

dotcms-integration/src/test/java/com/dotcms/rest/api/v1/drive/ContentDriveStatusFilterTest.java
zJaaal: You are right, it proved nothing. `liveItem` is titled "live" and fails the keyword filter on its own, so the assertion held whether the status clause was ANDed, applied, or dropped entirely — and dropping the status clause whenever a keyword is present is precisely the FR-009 regression the test exists to catch.

Fixed in `12af6dc` by seeding `liveItemMatchingKeyword`, published and titled `archived-decoy`, so it DOES match the keyword "archived" and is not archived. The assertion now targets that item, which can only be excluded by the status clause, and the message describes what the fixture actually sets up.

Your observation about the rest of the file is the useful part — several tests do open with an `assertNotNull` guard against exactly this, and this one did not. Worth me checking the others in that class rather than assuming the one you spotted is the only one.

<sub>Reply by Claude (Claude Code), posted from @zJaaal's account.</sub>

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/dot-content-drive.store.ts
zJaaal: Agreed on the analysis, and thank you for separating "not introduced here" from "newly reachable here" — that distinction is exactly right. Until this PR the drive's filters degraded server-side (an unknown `contentType` id is dropped via `Try.of(...).getOrNull()`), so a hand-edited `?filters=status:BOGUS` producing a real 400 is a new path into that handler, and it lands silently: `status: ERROR` is set, `items` is not cleared, nothing outside the store reads `ERROR`, so the spinner just stops over a stale grid. It masks 401/403/5xx the same way.

@zJaaal has asked to keep Content Drive delivery in this PR, so I am checking with him whether this comes in here too. My hesitation is not scope but that it needs a product decision rather than a code change — what should the user actually see? A toast, an inline error state in the grid, a cleared grid with a retry? Picking one silently is how it ends up wrong.

Will follow up here once that is settled.

<sub>Reply by Claude (Claude Code), posted from @zJaaal's account.</sub>

zJaaal:
Responding to the automated review (job [32977672907](https://github.com/dotCMS/core/actions/runs/32977672907)). All four addressed in `12af6dc`; three were real, one is a non-issue and the check that ruled it out is worth recording.

**🟡 1–3, the doc drift — all correct, all fixed.** Same root cause: when folder suppression moved to the client I updated `spec.md`, `contracts/` and `data-model.md` and left the Java behind. So `AbstractDriveRequestForm.status()`, the `@Operation` description, and the `build()` log comment were each still describing behaviour the endpoint deliberately does not have. The `@Operation` one mattered most — it is public API documentation telling callers folders would disappear when they had asked for them. Same three items @oidacra raised; replied in thread there too.

**🟡 4, pagination reset on restore — non-issue, and correctly hedged.** The review flagged the right condition: *"If content-drive never restores a page offset from the URL, this is a non-issue."* It does not. The URL carries only `path`, `filters`, `isTreeExpanded`, `editContent` and `editContentLang` (`store.ts:563-566`, `shell.component.ts:593-604`) — pagination is never encoded or restored, and `initContentDrive` resets it to page 1 regardless (`store.ts:204-208`). So there is no reachable state where a restore carries a non-first page for the sync to reset.

Since then the sync moved from `effect` to `signalMethod` (`f2fcf3f`), and `$selection` compares by **value**, so a write that does not change the selection produces no notification at all — the cycle cannot start rather than merely converging. There is now a real-store `describe` block covering the write-back path, including that an unrelated filter change triggers no second write. Worth flagging that the pre-existing suite mocks the store, so this class of bug was structurally invisible to it.

**On the non-blocking items:** the `loadItems` `catchError` and the `PURE_ES` `UNPUBLISHED` divergence are both tracked — the first pending a product decision on what the user should see on error, the second recorded in `tasks.md` as a spec decision. The archive-step interaction is no longer untested: two cases added to `ContentDriveWorkflowArchiveStepTest`.

<sub>Analysis and reply by Claude (Claude Code), posted from @zJaaal's account.</sub>

core-web/libs/portlets/dot-content-drive/portlet/src/lib/store/dot-content-drive.store.ts
zJaaal: Settled — @zJaaal's call was to validate at load and drop the bogus value, rather than surface an error. Done in `3475208`.

The `status` decoder now drops values that do not name a real `ContentStatus`, so a hand-edited or stale `?filters=status:BOGUS` never reaches the endpoint. It degrades to "no status filter", which is exactly the behaviour you pointed at: the drive's other filters already degrade this way, an unknown `contentType` id being dropped via `Try.of(...).getOrNull()`.

`decodeFilters` also drops a multi-value key that decoded to nothing, so a sanitized value does not round-trip back into the URL as a bare `status:`.

To be clear about what this does and does not do: it closes the path **this PR opened** into that handler. It does not fix the handler. Your analysis stands — `loadItems`' `catchError` still sets `status: ERROR` and returns `EMPTY` with no logging, nothing outside the store reads `ERROR`, `items` is never cleared, and 401/403/5xx are still swallowed into a stale grid with a stopped spinner. That remains worth its own ticket, and the decision there is a product one (toast, inline error state, cleared grid with retry) rather than a code change.

Two side effects worth flagging since they touch shared code:
- `multiSelector` lost its `DotContentDriveDecodeFunction` annotation. That type returns `string | string[]`, which hid the array from the `status` decoder composing on top of it. Still assignable everywhere a decode function is expected.
- One more pre-existing test used `status` as a placeholder key for "some multi-value filter" with values `published,draft`. Those are not valid statuses and are now sanitized away, so the placeholder is renamed to `owner` — the third instance of that collision on this branch. The test still covers unknown-key decoding, which is what it was for.

<sub>Analysis and reply by Claude (Claude Code), posted from @zJaaal's account.</sub>

dotCMS/src/main/java/com/dotcms/browser/ContentStatus.java
zJaaal: Follow-up: the distinction is now a decision, not just a documented caveat.

Spelling it out raised the obvious next question — the spec claimed both query paths return the same set (FR-009 / SC-005), and `UNPUBLISHED` demonstrably does not. @zJaaal's call was to **accept the divergence and lean on ADR-0018** rather than chase parity.

The reasoning: `UNPUBLISHED` means *no live version exists* — a question about the content, not about one version. The index stores `live` per version, so "does any version of this identifier have `live:true`?" is not answerable from a single document. There is no index-side fix short of redefining the status per-version, which would degrade the **default** path to match a limitation of a strategy nobody runs (`PURE_ES` is opt-in and is not set in any config file in this repo).

ADR-0018 already covers this class of thing: it routes structural predicates to the database precisely because the index cannot answer them reliably, and states that `PURE_ES` forfeits that guarantee for every criterion. This is one instance of a limitation that decision accepted, not a new one.

So the identifier-scoped reading in this enum stays the definition, and the spec now says where it does not hold:

- **FR-009** narrowed to the default strategy
- **FR-009a** (new) records the exception and the ADR that justifies it
- **SC-005** scoped to match
- `contracts/drive-search-status.md` gains a table contrasting the two predicates and the question each actually answers

`ARCHIVED` and `LOCKED` are unaffected on both paths. T031 (the `PURE_ES` parity test, still outstanding) is scoped accordingly: assert set equality for those two, and assert the default path's identifier-scoped meaning for `UNPUBLISHED` rather than parity.

Spec changed after sign-off, so #37170 needs another look before this merges.

<sub>Reply by Claude (Claude Code), posted from @zJaaal's account.</sub>

oidacra (approved):

hmoreras (approved):

PR #37218, opened 2026-08-25
#37209: feat(users): accept role IDs alongside keys in role endpoint

### LINKED ISSUE #37209: [Task] Users API: accept role IDs (not only keys) in `roles` on `POST/PUT /v1/users`
## Context

`POST /v1/users` and `PUT /v1/users` take `roles` as a list of **role keys** (`UserForm.roles`, `UserForm.java:39`). Resolution is key-only:

- update: `processRoles` → `roleAPI.loadRoleByKey(roleKey)` (`UserResource.java:1398-1418`)
- create: `UserHelper.addRole(user, roleKey, …)` → `loadRoleByKey` (`UserResource.java:1024`, `UserHelper.java:55-58`)

Roles created in the Roles portlet commonly have **no key** (`roleKey` is optional on `POST /v1/roles` and on the Dojo dialog). Such roles cannot be expressed in `roles` at all.

Since #37156 the payload is the source of truth: every `roles` list (including `[]`) reconciles the user's `editUsers=true` memberships to the desired set. A keyless role can never be in the desired set, so **any user update that sends `roles` strips the user's keyless custom roles** — silent permission loss. The Angular user-edit screen (#36930 arc) therefore cannot use this endpoint for users holding custom roles; it must fall back to N role-centric calls (`POST /v1/roles/{roleid}/users/{userId}`, `DELETE /v1/roles/{roleid}/users`).

## Proposal

Let each `roles` entry be **either a role key or a role ID**. Resolve by key first (`loadRoleByKey`), then fall back to `loadRoleById`. Apply identically on create and update.

| Field | Value |
|---|---|
| Endpoints | `POST /v1/users`, `PUT /v1/users` — existing, body field `roles` |
| Change | `UserResource#processRoles` (update) and the create path via `UserHelper#addRole`: resolve key → fallback ID |
| Request body | unchanged shape — `roles: ["<roleKey or roleId>", …]` |
| Response | unchanged |
| Auth | unchanged |
| OpenAPI | update the `roles` field description on `UserForm`; regenerate `openapi.yaml` |

## Behavior

- Entry matches a role key → that role (legacy precedence preserved)
- Otherwise entry matches a role ID → that role
- Otherwise → unchanged: ignored with a debug log (do **not** turn this into 400 in this task — that is a breaking change for existing clients sending stale keys; track separately if wanted)
- `null` list / `[]` / non-empty semantics from #37156 unchanged
- `editUsers=false` roles remain non-assignable through this endpoint (unchanged)
- Key/ID ambiguity is not a practical concern (IDs are UUIDs); key-first order makes it deterministic regardless

## Backward compatibility

Additive. Key-only clients see identical behavior. The reconcile already diffs by role ID internally (`desiredRoleIds`), so ID entries slot in without changing removal logic. Net effect for keyless roles: they can now be *kept* in the desired set, so the current silent stripping stops.

## Acceptance Criteria

- [ ] `roles` entries resolve by key, then by ID, on both `POST /v1/users` and `PUT /v1/users`
- [ ] Integration tests: update with a keyless role's ID → user holds it; update with a mix of keys and IDs → reconciled set is exactly the union; keyless role held by the user is **preserved** when its ID is in the payload and **removed** when it is not (source-of-truth semantics unchanged); create with a role ID; existing #37156 tests (null / `[]` / blank entry → 400) still green
- [ ] OpenAPI description of `UserForm.roles` states "role key or role ID"; regenerated `openapi.yaml` committed
- [ ] No change to unknown-entry handling (still ignored), auth, or response shape

hassandotcms:
### Proposed Changes
* `roles` on `POST /v1/users` and `PUT /v1/users` accepts **role IDs as well as keys** — each entry resolves by key first (legacy precedence), then by ID. Roles created in the Roles portlet may have no key and could not be sent at all
* Change is confined to `UserResource` (`processRoles` + the create loop). `UserHelper#addRole` is untouched — SAML provisioning depends on it
* `@Schema` on `UserForm.roles` documents keys-or-IDs and the create/update semantics; regenerated `openapi.yaml` (additive)

### Fixes a latent nondeterminism
`RoleFactoryImpl.loadRoleByKey` consults the role cache first, and `RoleCacheImpl.get` check the ID group before the key group — so an ID entry already worked whenever the role happened to be cached (e.g. roles the user holds, warmed by the update's permission check) and was silently ignored otherwise. The explicit resolver makes the outcome independent of cache state.

### Backward compatibility
| Client | Effect |
|---|---|
| Key-only payloads (Angular users portlet, Postman, any integration) | Identical — key resolution runs first |
| `null` / `[]` / reconcile / blank-entry-400 semantics (#37156) | Unchanged |
| Unknown entries | Still ignored (turning this into 400 would break clients — out of scope) |
| Create's default role and its throw on non-assignable roles | Unchanged |
| SAML / SSO | Not affected (`UserHelper` untouched) |
| Keyless custom roles | Now expressible; the silent stripping on every `roles` payload stops |

### Checklist
- [x] 4 integration tests (TDD red→green) in `UserResourceIntegrationTest`: update by IDs (keyless + keyed-by-ID, reconcile), keyless kept-by-ID / removed-when-absent, create by ID, create default pinned — 11/11 incl. the #37156 contract tests
- [x] `UserResourceTest` 9/9
- [x] No breaking changes

Fixes #37209

This PR fixes: #37209


ihoffmann-dot (dismissed):

ihoffmann-dot:
**PR Review:**  
✅ Looks good to me. The change correctly extends the roles payload on `UserResource` create/update to accept role IDs alongside keys, the fallback in `resolveRole()` preserves existing behavior, and the existing `addRoleToUser` assignability check (`isEditUsers()`) is still enforced since the code routes through the same API method. 
Test coverage is solid. The Swagger/OpenAPI description update is consistent with the new behavior. Looks consistent with the codebase's existing patterns and ready to merge.

dario-daza (dismissed):

dario-daza (approved):

PR #37219, opened 2026-08-25
refactor(sdk-release): replace regex-based PR subject parsing with GitHub commit-to-PR association API in release workflow

### LINKED ISSUE #37200: SDK breaking-change gate: replace squash-subject regex with commits→PRs API
Part of dotCMS/private-issues#673.

`.github/workflows/cicd_6-release.yml:157` extracts PR numbers via `grep -oE '\(#[0-9]+\)$'` on commit subjects. Merge-commit subjects (`Merge pull request #N from …`) don't match, so under merge commits the gate silently passes and PRs labeled `SDK Breaking Change` ship unflagged.

Fix: resolve PRs via `GET /repos/{owner}/{repo}/commits/{sha}/pulls` instead of subject parsing. Works under both merge strategies — safe to land now.

KevinDavilaDotCMS:
To fix the new mechanism without squash-merge

This PR fixes: #37200

### DIFF
--- a/.github/workflows/cicd_6-release.yml
+++ b/.github/workflows/cicd_6-release.yml
@@ -145,16 +145,21 @@ jobs:
           # release-prepare hasn't created this release's tag yet at this point in the
           # pipeline — main is the correct proxy for "what's about to ship" (release_commit
           # defaults to main's tip the same way).
-          # Only the commit SUBJECT (first line) — the full multi-line message's body
-          # routinely has prose or footnotes ending in "(#N)" that reference an unrelated,
-          # often much older PR/issue (confirmed: ~20 real examples in this repo's last 3000
-          # commits). Scanning the full message would treat those stray references as PRs in
-          # this release's range, risking both a false block (an old referenced PR happens to
-          # carry the label) and an unwarranted floor bump if the operator reruns with
-          # bump_min_sdk_version=true to "fix" it.
-          PR_NUMBERS=$(gh api "repos/${REPO}/compare/${PREV_TAG}...main" --paginate \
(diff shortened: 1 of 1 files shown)


.github/workflows/cicd_6-release.yml
```diff
@@ -145,16 +145,21 @@ jobs:
...
+          # always passed. The API resolves off git's actual PR-merge metadata, not the
+          # subject's text shape, so it stays correct under squash, merge, or rebase.
+          COMMIT_SHAS=$(gh api "repos/${REPO}/compare/${PREV_TAG}...main" --paginate \
+            --jq '.commits[].sha' || true)
+
+          PR_NUMBERS=$(
+            for sha in ${COMMIT_SHAS}; do
+              gh api "repos/${REPO}/commits/${sha}/pulls" --jq '.[].number' 2>/dev/null || true
+            done | sort -un
+          )
```
sfreudenthaler: What happens if we hit a api rate limit 429?

We might have to watch for api rate limits and impliment backoff.  But at least we need to fail gracefully

KevinDavilaDotCMS: Added: each per-commit lookup now checks the API call's own exit status separately from an empty result. A real failure (rate limit, transient 5xx, network blip) now fails the step loudly with `::error::` naming the commit, instead of being swallowed by `2>/dev/null || true` and silently dropping that commit's PR from the count. See 9be1d6d1f0.

sfreudenthaler (approved):
love it thanks @KevinDavilaDotCMS

Issue #37174, opened 2026-08-24
Content Drive toolbar control, shared filter chips and row selection in the list view


nicobytes:
## Problem Statement

Follow-up work on the components that Content Drive and the **Asset Picker** dialog now share, after the folder-tree unification landed in #36733 / PR #36848.

> **This issue was split.** It originally carried nine findings; the folder-tree ones and the two Asset Picker ones now live in their own issues under the same parent (#36702) — see [Split-out issues](#links) below. What remains here are the three findings that sit on the **Content Drive toolbar and list-view chrome**.

Three things are tracked here:

1. **Content Drive toolbar control** (finding 5), from the [QA Failed comment by @erickgonzalez on #36733](https://github.com/dotCMS/core/issues/36733#issuecomment-5197164875) — the collapse/expand control appears to "do nothing" to the tree nodes.
2. **Filter chips do not reach the Asset Picker** (finding 6) — every concrete chip lives inside the Content Drive portlet lib and injects `DotContentDriveStore` directly, so features added to Content Drive's toolbar never show up in the picker.
3. **Row selection in the shared list view** (finding 7) — clicking a row does not select it; only a click landing exactly on the checkbox/radio does.

**Impact:** cosmetic + minor behavioral, but it compounds. The picker silently lags behind Content Drive every time a filter chip is added there, and the widest, most natural click target in a list row does not do the thing users expect.

**Note:** the item about the sites search bar in the original QA comment is struck through and is **not** part of this issue — the search bar only appears with more than 10 sites, which is working as intended.

### Findings

| # | Where | Finding | Type |
|---|---|---|---|
| 5 | Content Drive toolbar | The collapse/expand control appears to "do nothing" to the tree nodes | Unclear affordance |
| 6 | Asset Picker toolbar | Filter chips added to Content Drive (e.g. **Show Shared Assets**) never appear in the picker, even though both toolbars are built from the same `dot-chip-filter` primitive | Parity gap |
| 7 | Asset Picker / Content Drive list | Selecting a row requires clicking exactly on the checkbox/radio; clicking elsewhere on the row does not select it | UX |

### Root cause notes

**Finding 5 (toolbar control)**

[...]

PR #37220, opened 2026-08-25
Add site selector and site-scoped folder search to the AssetPicker sideb

### LINKED ISSUE #37208: AssetPicker: new sidebar UI — site selector + site-scoped folder search above the folder tree
### Description

Apply the new AssetPicker UI (design from product — link in the first comment). The right-hand column (asset search, filter chips, table, pagination) already matches the design; **the change is the left column**.

**Today** — `dot-asset-picker-sidebar.component.html` renders two things:

1. one `dot-search-input` labelled *"Search sites & folders"*, and
2. a `dot-folder-tree` whose **roots are every site the user can browse**.

The single search term is applied to *both* halves at once: it filters the sites query and, in parallel, lists matching folders flat under the browsed site's root (`with-asset-folder-tree.feature.ts` → `searchFoldersUnderSite`). Sites and folders are mixed in one control, and the browsed site is implicit — whichever root the user happened to expand.

**New design** — the left column becomes three stacked controls:

| # | Control | Purpose |
|---|---|---|
| 1 | **Site selector** (globe icon + hostname + chevron) | Choose the site being browsed. Own searchable list — **sites only**. |
| 2 | **Folder search** (`Search folders...`) | Recursive, name-contains search **inside the selected site**. Renders a flat result list in place of the tree; each row shows the folder name plus its full path (`demo.dotcms.com / activities`, `demo.dotcms.com / images / …`). |
| 3 | **Folder tree** | Rooted at a single `All` node = the selected site's root. No site roots any more. |

So the "cross sites without leaving the dialog" model — roots = sites — is **replaced**: the picker is pinned to one site at a time, chosen explicitly in control 1. That removes the search-term coupling between the sites query and the folders query, which is the source of the workaround documented at `with-asset-folder-tree.feature.ts:152-157` (re-synthesising the browsed site's root because the shared term dropped it out of `roots`).

**Reuse:** `DotSiteComponent` (`libs/ui/src/lib/components/dot-site/`) is already a lazy, virtual-scrolled `p-select` with a `Search sites...` filter template — it is the intended control for #1. `DotFolderTreeComponent` stays as-is (presentational); the changes are in `DotAssetPickerSidebarComponent` and in the store's folder-tree feature.

### Acceptance Criteria

**Site selector**
- [ ] The sidebar renders a dedicated site selector at the top, above the folder search, showing the browsed site's hostname with a globe icon and a chevron
- [ ] Its dropdown has its own search input that filters **sites only** — no folders appear in it
- [ ] The site list is lazily loaded / paged (reuses `DotSiteComponent`, not a bespoke control)
- [ ] Choosing a site loads that site's folder tree and re-scopes the asset list on the right to the new site
- [ ] On open, the preselected site is derived from the **global last-used asset path** in `localStorage` (see #36702); when that path is absent, or its site no longer exists or is no longer visible to the user, it falls back to the current site from the global site switcher

**Folder search**
- [ ] A second input (`Search folders...`) sits directly under the site selector
- [ ] It matches folder names by **contains** (not prefix), e.g. `a` matches `activities`, `images`, `thumbnails`, `application`, `containers`
- [ ] The search is **recursive** across the whole selected site — nested folders match (e.g. `thumbnails` under `images`)
- [ ] Results are scoped to the **selected site only** — no folders from other sites are returned
- [ ] While a term is active, the flat result list **replaces** the tree in the sidebar
- [ ] Each result row shows the folder name on the first line and its full path on the second (`demo.dotcms.com / activities`); long paths truncate with an ellipsis instead of wrapping or overflowing
- [ ] Selecting a result scopes the asset list to that folder and **keeps the result list visible**, with the chosen row rendered as selected, so the user can jump to another folder without retyping
- [ ] Clearing the input (`×` or empty string) returns to the folder tree
- [ ] Changing the site while a term is active clears the term and returns to the tree — a term is only meaningful against the site it was typed for
- [ ] An empty result set renders an empty-state message, not a blank panel

**Folder tree**
- [ ] The tree renders exactly one root, labelled `All`, representing the selected site's root (`//<hostname>/`)
- [ ] Selecting `All` scopes the asset list to the whole selected site
- [ ] Site nodes are no longer tree roots; the tree contains only folders of the selected site
- [ ] Lazy expansion and `Load more` paging keep working per level, unchanged
- [ ] Reopening the picker on a remembered path still expands the tree down to that folder and marks it selected

**Regression / non-goals**
- [ ] The right column (asset search, content-type / workflow / locale chips, table, pagination) is unchanged by this issue
- [ ] The silent mimetype filter, content-type restrictions, and locale preselection from #36702 keep working for both the File and the Image entry points
- [ ] The new AssetPicker is still not used in the legacy Dojo editor (see #37132)
- [ ] `DotAssetPickerSidebarComponent` and the folder-tree store feature have unit tests covering: site change, contains-match search, site-scoped results, result selection keeping the list open, clearing the term, and the single `All` root

### Priority

Medium

### Additional Context

- Part of epic #36702 (Create AssetPicker component reusing Content Drive).
- Files most affected:
  - `core-web/libs/ui/src/lib/components/dot-asset-picker/components/dot-asset-picker-sidebar/` — template + component
  - `core-web/libs/ui/src/lib/components/dot-asset-picker/store/features/with-asset-folder-tree.feature.ts` — drop the shared-term coupling between the sites query and the folders query; add `browsingSite` as an explicit, user-chosen input
  - `core-web/libs/ui/src/lib/components/dot-site/` — reused for the site selector
  - `core-web/libs/ui/src/lib/components/dot-folder-tree/` — presentational, expected to need no change
- Sibling work in flight that touches the same tree: #37174 (folder tree QA regressions in Content Drive and Site/Folder Field) — coordinate to avoid conflicting edits to `dot-folder-tree` / `site-tree.utils`.
- Content Drive is already pinned to a single site via the global switcher, so this change moves the picker **closer** to Content Drive's model, not further from it — worth checking whether the new site selector can be lifted into a shared sidebar-header component for both.

nicobytes:
- Split the combined "search sites & folders" control into three stacked pieces: a dedicated site selector, a folder-only search, and a folder tree rooted at a single site (was previously rooted at every browsable site), matching Content Drive's one-site-at-a-time model.
- Extract the flat search-result row into a shared DotFolderSearchResultsComponent so the AssetPicker and the Site/Folder field render identical results instead of duplicating the presentation.
- Update e2e helpers/fixtures and specs to cover switching sites, recursive folder search, and the new sidebar layout.

#37208


nicobytes:
### ⚠️ Merge-order conflict with #37174 that git will not flag

This PR and **#37174** both change **`host-folder-field.component.html`** — and specifically the
same branch of it: how the Site/Folder field renders results while a folder search is active.

- **This PR** replaces that branch. Search results stop being a `p-tree` and become
  `<dot-folder-search-results>`, the row extracted into `@dotcms/ui` so the AssetPicker renders the
  identical thing instead of a second copy of it (#37208).
- **#37174** fixes folder-tree QA regressions in Content Drive *and* this field.

The two changes have **different intents for the same rendering**. Whoever merges second inherits a
conflict that may resolve cleanly on paper while producing a broken or duplicated search view.

**What to check when merging second**

1. The `@else if (isSearching)` branch in `host-folder-field.component.html` — make sure the swap
   survived and did not get reverted into a tree.
2. `helpers/host-folder-field.ts` → `folderNode()`. It now matches **either** rendering. Two e2e
   specs search and then click a result; before this change the locator only knew about
   `.p-tree-node-content` and both broke.
3. `pnpm nx test edit-content --testPathPatterns=host-folder-field` and the field's four e2e specs.

**What was deliberately kept out of the collision**

`dot-folder-tree/`, `site-tree.utils.ts`, `DotBrowsingService` and Content Drive are **untouched by
this PR** — verified as an absence of diff. The shared tree was never turned into a mode-switching
component precisely so this overlap would stay confined to one file.

The sequencing conversation this warning was meant to replace was waived deliberately, so this note
exists for whoever hits it.

wezell (commented):
Should this remember state when it is used in different places?  the state key would need to understand user behavior and which control is using it but  I know that is a common ask to not have to dive deep into the file picker to get to your images folder.

oidacra (dismissed):

adrianjm-dotCMS (dismissed):

.specify/feature.json
oidacra, line 1: We need to remove this file

adrianjm-dotCMS (approved):

oidacra (approved):

PR #37222, opened 2026-08-25
feat: add FEATURE_FLAG_DOTAI_CONFIG_UI and implement legacy guard to support dotAI configuration redesign

### LINKED ISSUE #36826: dotAI: Provider Configuration UI

KevinDavilaDotCMS:
https://github.com/user-attachments/assets/8bc6b2e4-d907-430b-9599-f88e8a66fa0a

### Why DotAiConfigDetailLegacyComponent exists

This is not a new screen — it's a verbatim restoration of what DotAiConfigDetailComponent looked like before #37048 (the "Provider Config" JSON editor on the left + "Example JSON" reference on the right), pulled from the commit right before #37048 overwrote it (git show 2a161f6fd6^:.../dot-ai-config-detail.component.ts, where 2a161f6fd6 is #37048's merge commit).

#37048 redesigned DotAiConfigDetailComponent in place, so the old UI no longer exists anywhere in the tree except git history. Since this redesign now ships behind FEATURE_FLAG_DOTAI_CONFIG_UI (default false), simply falling back to the generic Apps config screen when the flag is off would have been a regression for existing customers — that's not the screen they use today for dotAI. Restoring the pre-#37048 component as a separate file (DotAiConfigDetailLegacyComponent, selector dot-ai-config-detail-legacy) preserves their exact current experience: same markup, same EXAMPLE_CONFIG, same DotAiService.getConfig/saveConfig calls against /completions/config, which are unchanged and still live.

dot-apps.routes.ts picks between the two via canMatch: flag on → DotAiConfigDetailComponent (new capability-cards UI); flag off/default → DotAiConfigDetailLegacyComponent (this one), so nothing changes for customers who haven't opted in.

This PR fixes: #36826


rjvelazco (dismissed):

nicobytes (dismissed):

dario-daza (dismissed):

KevinDavilaDotCMS:
Addressed all 5 review comments from the automated review (commit 89ae034c2d):

1. **Stale comment** — fixed the property/javadoc comment on `FEATURE_FLAG_DOTAI_CONFIG_UI` (`dotmarketing-config.properties` + `FeatureFlagName.java`); it now correctly says the off-flag fallback is `DotAiConfigDetailLegacyComponent`, not the generic Apps config screen.
2. **Missing `.scss`** — added `dot-ai-config-detail-legacy.component.scss` and wired it via `styleUrls`, per the repo's file-structure standard.
3. **Signal `$` prefix** — renamed the legacy component's signals to the repo convention (`$app`, `$configJson`, `$saving`).
4. **Blank render path** — added an explicit `$loading` state so the template now handles loading / loaded / error instead of rendering blank while `$app()` is null.
5. **Missing test coverage** — added unit tests for `getFreshFeatureFlag()` in `dot-properties.service.spec.ts`: value coercion parity with `getFeatureFlag()`, `NOT_FOUND` handling, and confirming it bypasses `featureFlagCache`.

Verified locally: `dotcms-ui` build, ESLint on all changed files, and Jest (`dot-properties` service spec + full `dot-apps` suite) — all green, no regressions.

nicobytes (dismissed):

dario-daza (dismissed):

wezell (approved):
Trying to merge again

PR #37226, opened 2026-08-25
feat(cicd): generate post-merge QA test plans for eligible merged PRs

### LINKED ISSUE #37225: Skill proposal: dot-test-plan (adds new 'test' domain prefix)
## Description

Proposal to add a first-party skill **`dot-test-plan`**, and — the actual review point here — to add
a new approved domain prefix **`test`** to `.claude/skills/skills.config.json`.

Per [CONTRIBUTING.md §6](../blob/main/.claude/skills/CONTRIBUTING.md), a proposal issue is
*required* when a skill introduces a new domain prefix, so this is that gate.

### The problem

After an eligible PR merges, the developer who will verify it has to reconstruct a lot of context by
hand: the issue and PR history, what the merged implementation actually does, which product surfaces
it touches, which regression scenarios matter, and then write and format a complete post-merge test
plan and hand it to whoever executes it.

This skill generates that plan as a **starting point** — one consolidated plan per PR covering every
issue it fixed, posted as a comment on each of them. It does not replace developer judgment,
independent execution, or final quality ownership.

It is the skill half of the post-merge test-plan automation; a GitHub workflow
(`ai_claude-post-merge-test-plan.yml`) invokes it on `pull_request.closed`.

### Proposed name and domain

| | |
|---|---|
| Skill | `dot-test-plan` |
| New domain | `test` |
| Owner | `@dotcms/falcon` |
| Status | `experimental` |

### Why a new domain, and why not an existing one

`dot-test-plan` is about test plans, which none of the seven approved domains
(`issue`, `pr`, `release`, `cicd`, `content`, `ui`, `sdk`) covers:

- **`dot-pr-…`** was the closest alternative and would need no governance change. Rejected because
  the skill also runs interactively against an issue with no PR in hand, so the prefix would
  mislead on discovery.
- **`dot-issue-…`** is taken by issue triage/creation workflows (`dot-issue-manage`); a test plan is
  not issue management.
- **`dot-cicd-…`** covers build and pipeline diagnosis (`dot-cicd-diagnose`), not test design.

`test` is a durable category — it is plausible that future skills (test-data generation, flaky-test
triage) would sit under it rather than forcing each into an ill-fitting prefix.

### Why not extend an existing skill

Checked [CATALOG.md](../blob/main/.claude/skills/CATALOG.md). Nearest matches:

- **`dot-cicd-diagnose`** — diagnoses *failing CI runs*. Opposite direction: it reads red builds,
  this writes a manual plan for a green merge.
- **`dot-issue-manage`** — creates, updates and queries issues. This skill only ever *comments*; it
  never changes issue state, labels, assignees or milestones.

No overlap worth merging. `skill-lint`'s duplicate-similarity check reports 0 warnings against the
existing 21 first-party skills.

## Acceptance Criteria

- [ ] `test` added to `approvedDomains` in `.claude/skills/skills.config.json` **in the same PR** as
      the skill, so this proposal is the review point for the new domain
- [ ] `dot-test-plan` passes `just skills-lint` (naming, `owner`/`status` frontmatter, catalog
      freshness)
- [ ] `SKILL.md` stays under the ~500-line convention, with supporting material in `references/`
- [ ] `CATALOG.md` regenerated, not hand-edited
- [ ] A maintainer 👍 on the new `test` domain before the implementation PR is reviewed

## Priority

Medium

## Additional Context

- Part of the 8-week post-merge test-plan Rock (weeks 1–2, "Polish and publish the skill").
- A first real plan generated by this skill is already posted on #36991 for merged PR #37112:
  https://github.com/dotCMS/core/issues/36991#issuecomment-5415227495
- **Scope note:** the skill deliberately does **not** audit main-branch test coverage and does not
  recommend automation to add. Every case it emits is manual. That scope reduction is intentional
  and agreed, not an oversight.

erickgonzalez:
### Proposed Changes

Adds the post-merge QA test plan automation: when a PR labeled `Area : Backend` or
`Area : Frontend` merges, generate the manual hand-check plan a developer runs against the
post-merge build, and post it to every related issue.

* **New skill `dot-test-plan`** (`.claude/skills/dot-test-plan/`) — one consolidated plan per PR
  covering every issue it fixed. Every case is `Manual` and starts `Not Run Yet`. `SKILL.md` is 373
  lines with supporting material in `references/`, per the ~500-line convention.
* **New `test` domain** in `skills.config.json` — **this is the review point of the PR**. Proposed
  and reasoned in #37225; CONTRIBUTING.md §6 requires a maintainer 👍 on a new domain prefix.
* **Discovery script** (`.github/scripts/related-issues/resolve-related-issues.js`) — resolves
  related issues from the union of branch name, the `This PR fixes` statement, and GitHub
  Development links, validating every candidate against the API.
* **Workflow** (`ai_claude-post-merge-test-plan.yml`) — `preflight → prep → claude → finalize`,
  following `issue_autodoc.yml`. Everything deterministic happens in `prep`, so the model only reads
  the diff, writes the plan, and posts it.
* **Prompt** (`.github/prompts/qa-postfix-test-plan.md`) — kept out of the workflow YAML, matching
  `gpt-auto-review.md`.

### Checklist
- [ ] Tests — *see Additional Info; no automated tests ship with this, deliberately*
- [x] Translations — n/a, no user-facing UI strings
- [x] Security Implications Contemplated — notes below

### Additional Info

**It merges dormant.** `preflight` is gated on `vars.POST_MERGE_TEST_PLAN_ENABLED == 'true'`, so
nothing fires on real merges until someone sets that variable. This is deliberate: the workflow
would otherwise start running against live team issues the instant it lands, and `workflow_dispatch`
only works from the default branch, so there is no way to rehearse it beforehand. Plan is to
dispatch it by hand against a closed PR first, then arm it.

**Why three discovery sources.** Across 21 real merged PRs, only 7 had a `This PR fixes` line at
all, and it never once contributed an issue the other sources missed — while Development links alone
found 15 of 16 in one sample, and were the *only* source for several. Two traps that shaped the
implementation:

* Development links are **not repo-scoped**. PR #37167 links `dotCMS/private-issues#671`, and
  `dotCMS/core#671` is an unrelated 2012 pull request. The script keeps `repository.nameWithOwner`
  and uses it to disqualify that number when it also appears in the branch name — otherwise a
  security plan could land on a stranger's decade-old ticket.
* The `This PR fixes` line is frequently written by `issue_comp_link-pr-to-issue.yml` on this same
  `pull_request.closed` event, names only one issue, and races us. `prep` re-fetches the PR body
  from the API rather than trusting the event payload.

**Label strings.** The gate matches `Area : Backend` / `Area : Frontend` exactly — with spaces
around the colon, which is how they actually exist in this repo.

**Testing.** No unit tests ship with the discovery script. That was a deliberate call against adding
a third near-duplicate PR validator to a repo with 69 workflows; the edge cases that justified them
are documented in the script's own comments instead. It was verified by running against **21 real
merged PRs** (including author-prefixed branches, version digits in branch names, multi-issue
branches, and the cross-repo case above). The workflow's `finalize` job errors when a plan fails to
land on an issue, so the regression is caught at runtime. **If you change that script, re-run it
against real PRs.**

Also verified: label gate against real label data plus adversarial variants; the idempotency check
(keyed on PR + merge SHA within a single comment); prior-plan retrieval and revision handling
against live data; `finalize`'s verification loop; and cross-job output wiring. Not yet exercised:
Bedrock OIDC, the reusable-workflow call, and whether the model's output conforms to the comment
skeleton — that is what the dispatch run is for.

**Security notes.**
* Issue bodies, PR descriptions, comments and prior plans are treated as **untrusted data, never
  instructions**. Text directed at the model is quoted as a flagged anomaly and ignored rather than
  acted on. Documentation is followed one hop only, capped, and only from links already present in
  the issue or PR.
* `allowedTools` is scoped to `gh pr diff`, `gh issue view`, `gh issue comment` — no repo-wide
  `grep`/`find`, and no `git`, since the executor checks out shallow on the PR ref where the
  squash-merge commit does not exist.
* Job permissions are least-privilege: `prep` and `finalize` are read-only apart from the AI job's
  `issues: write`. The workflow never changes issue state, labels, assignees or milestones.
* Prior-plan ingestion is bounded (latest plan only, 400 lines) to cap both cost and injection
  surface.

**Sample output.** A plan generated by this skill for merged PR #37112 is posted on #36991:
https://github.com/dotCMS/core/issues/36991#issuecomment-5415227495

**Scope note.** The skill deliberately does **not** audit main-branch test coverage and does not
recommend automation to add. Every case it emits is manual. That reduction is intentional and
agreed, not an oversight.

This PR fixes: #37225

### DIFF
--- a/.claude/skills/CATALOG.md
+++ b/.claude/skills/CATALOG.md
@@ -3,7 +3,7 @@
 
 Auto-generated inventory of dotCMS skills. **Check here before creating a new skill** — if something close exists, extend it or mark yours `related`, don't fork. See [CONTRIBUTING.md](CONTRIBUTING.md).
 
-_20 first-party · 6 external (symlinked)._
+_21 first-party · 6 external (symlinked)._
 
 ## First-party skills (`dot-*`)
 
@@ -14,6 +14,7 @@ _20 first-party · 6 external (symlinked)._
 | `dot-release-backport-lts` | active | @dotcms/maintenance | Backport closed GitHub issues labeled LTS Next Patch to the current LTS release branch. Finds linked PRs, applies diffs, commits, and pushes. Use when applying… | — |
 | `dot-release-rollback-check` | active | @dotcms/maintenance | Check whether a dotCMS release can be safely rolled back to a previous version by inspecting all PRs merged between the two versions for rollback-safety labels… | — |
 | `dot-sdk-analytics` | active | @dotcms/falcon | Use this skill when the user asks to install, configure, or set up @dotcms/analytics, sdk-analytics, analytics SDK, add analytics tracking, or mentions install… | — |
+| `dot-test-plan` | experimental | @dotcms/falcon | Generates the manual post-merge QA test plan for a merged dotCMS pull request. Reconstructs issue and PR context, consolidates every issue the PR fixed into ON… | — |
 | `dot-ui-angular-standards` | experimental | @dotcms/falcon | dotCMS Angular coding standards for the core-web Nx workspace. Use this skill for ANY frontend work under core-web/ — writing or editing a component, service, … | — |
 | `dot-ui-vtl-migration` | active | @dotcms/falcon | Migrates VTL (Velocity Template Language) custom field templates from the legacy DotCMS Dojo/Dijit API to the modern DotCustomFieldApi. Use this skill whenever… | — |
 | `skill-doctor` | _legacy_ | — | Use when a repo skill fails, produces errors, gives wrong instructions, or references stale information. Also use when a command from a skill returns "not foun… | — |
--- a/.claude/skills/dot-test-plan/README.md
+++ b/.claude/skills/dot-test-plan/README.md
@@ -1,0 +1,150 @@
+# dot-test-plan
+
+Generates the **manual post-merge QA test plan** for a merged dotCMS pull request: the hand-check
+list a developer executes against the build that contains the fix.
+
+Runs two ways — **interactively** in Claude Code, or **unattended** in the post-merge GitHub
+workflow, which consolidates every issue a PR fixed into one plan and posts it to each of them.
+
+---
+
+## What it produces
+
+A numbered list of `TC-###` cases, each with risk, scenario, self-contained reproduction steps, an
+observable expected result, and a result field the executor fills in. Every case is `Manual` and
+starts `Not Run Yet`.
+
+The plan body is exactly two sections — **Summary → Test Cases** — wrapped in the GitHub comment
+skeleton (marker, ownership table, four-column summary, full plan in `<details>`). Deliberately
+lean: no assumptions table, no per-issue coverage index, no out-of-scope list, and no completion
+block, since the ownership table already carries both statuses.
+
+## What it does *not* do
+
+- Write or run tests, or record their results.
+- Audit main-branch test coverage, or recommend automation to add. **Deliberately out of scope** —
+  the only code it reads is the merged diff, and only to avoid duplicating tests the PR itself added.
+- Propose `Unit`, `Integration`, or `E2E` tests. Every case in the plan is manual by definition,
+  so cases carry no test-type field at all.
+- Assign an executor, approve a plan, or infer a pass from green CI or a closed issue.
+- Modify issue state, labels, assignees, or milestones. It only comments.
+
+---
+
+## When it triggers
+
+"post-merge test plan", "QA plan for PR #X", "verify this merged fix", "regression checklist for the
+merged PR", "what should QA exercise post-merge", "test plan for issue #X", "qa-postfix plan" — or
+automatically, when a PR labeled `Area : Backend` or `Area : Frontend` merges.
+
+Force it with `/dot-test-plan`.
+
+---
+
+## Interactive vs. automated
+
+| | `interactive` | `automated` |
+|---|---|---|
+| Clarifying questions | Asked in one batch, waits | **Never asked** — the documented default is applied and made visible inside the affected case |
+| Missing input | Blocks until answered | Documented default applied and recorded |
+| Failed `gh` call / dead doc link | Reported | Recorded, run continues |
+| Output | Printed; posted only if you ask | Posted to **every** related issue |
+| Local files | May save | **None** — the runner discards them |
+
+Selected by `EXECUTION: automated`, `DOTCMS_TEST_PLAN_EXECUTION=automated`, or `GITHUB_ACTIONS=true`
+with no human turn. **When unsure it picks `automated`** — that output is always a complete plan,
+safe to hand a human. The reverse isn't true: an interactive run that stops to ask produces nothing
+at all in CI.
+
+---
+
+## Multi-issue consolidation
+
+One PR often fixes several issues. The result is **one plan**, posted verbatim to each.
+
+The issue set is the **union** of three sources, with no precedence:
+
+1. **Branch name** — consecutive leading numeric tokens, with or without an `issue-` prefix.
+   `issue-37085-bouncycastle-185` → `[37085]` (the `185` is a library version);
+   `36937-36938-roles-api` → `[36937, 36938]`.
+2. **PR description** — only references inside a `This PR fixes` statement. Bare `#123` elsewhere,
+   `Related:`, and `Fixes` / `Closes` / `Resolves` are not interpreted.
+3. **GitHub Development relationships** — every linked issue, not only those closing on merge.
+
+Every candidate is validated against the API and dropped if it 404s or is actually a PR. In
+automated runs the workflow supplies this list pre-validated and the skill trusts it.
+
+Guarantees: every issue gets ≥1 case naming it, and every case carries an `Issues:` field — that
+field is the whole of the per-issue traceability, which is why it is mandatory. No case
+cross-references another issue, since the same text is posted in several places.
+
+---
+
+## Revisions
+
+A later PR touching the same issue continues the previous plan. The skill finds the latest one by
+marker, deduplicates copies by `pr` + `merge-sha`, then preserves cases that still apply, adapts or
+replaces those the new PR affects, drops obsolete ones with a reason — and **resets every result to
+`Not Run Yet`**, because a new PR means a new execution pass.
+
+Case IDs restart at `TC-001` each revision. A new plan is always a new comment; it never overwrites
+an earlier one.
+
+---
+
+## Roles and statuses
+
+| Role | Who | Owns |
+|---|---|---|
+| **Plan Reviewer** | Normally the PR author | Accuracy, scope, missing cases, hallucinations. Approves. |
+| **Test Executor** | A **different** developer | Executes every case, records every result. |
+
+The separation is the point: implementation context from the person who made the change,
+independent verification from someone who didn't.
+
+- **Result** (per case, in the summary table only): `Not Run Yet` · `Passed` · `Failed` · `Blocked` · `Not Applicable`
+- **Plan Review Status**: `Pending Review` · `Approved`
+- **Execution Status**: `Not Started` · `In Progress` · `Completed` · `Completed With Failures` · `Blocked`
+
+Both statuses live in the `Status` column of the ownership table — the reviewer's row holds the
+Plan Review Status, the executor's row holds the Execution Status. Case results live in the
+`Result` column of the summary table and nowhere else, so the executor updates each result in one
+place. The summary table is `Test Case Number | Scenario | Result | Notes`; a `Failed`, `Blocked` or
+`Not Applicable` result puts its required explanation — or a link to the issue filed for it — in
+that row's `Notes` cell.
+
+`Failed`, `Blocked`, and `Not Applicable` each need a short explanation.
+
+### Adding your own cases
+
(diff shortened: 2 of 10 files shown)


erickgonzalez:
Thanks — the 🟠 High was a real bug and a good catch. Addressed in c6969e1.

### 🟠 High — marker SHA contract: **fixed**

Confirmed exactly as described. I wrote `"short form is fine, but be consistent"` in
`comment-format.md` *before* writing the consumers that anchor on the full `mergeCommit.oid`, and
the only worked example showed a 7-char SHA — which the skill tells the model to reproduce
verbatim. Demonstrated against the real marker from #36991:

```
marker uses 98bc757d8816d8ba8c18f86871ad100697584a03  -> idempotency=1  finalize-grep=1
marker uses 98bc757                                   -> idempotency=0  finalize-grep=0
```

Both consumers score 0 on the short form, so the two failures compound exactly as you set out: a
re-run duplicates onto every related issue, *and* the verification meant to catch it reports every
issue missing and fails the run.

Made the contract single-valued on the full SHA, since `prep` and `finalize` only ever hold the full
oid:
* rule now reads "in full — all 40 characters, never abbreviated"
* the canonical example carries a full 40-char SHA
* the pre-post checklist asserts it
* the prompt says so where it hands `{{MERGE_SHA}}` over

### 🟡 Medium — empty `merge_sha`: **fixed**

Didn't verify how often GitHub returns a null `mergeCommit` on the closed event, but the consequence
is bad enough that guarding beats measuring: the marker degenerates to `merge-sha: `, the dedupe
regex to `^merge-sha: $`, and every empty-SHA plan then matches every other — the idempotency key
collapses rather than merely missing. `Fetch PR metadata` now fails fast with an actionable
`::error::` pointing at the `workflow_dispatch` re-run path.

### 🟡 Medium — reusable-workflow contract: **not a defect, no change**

I'd verified the input names against `claude-orchestrator.yml@v3` before opening this; the auth
question you raise is the genuinely new part, and it checks out. The orchestrator declares:

```yaml
secrets:
  ANTHROPIC_API_KEY:
    description: 'Anthropic API key — required only when using the direct Anthropic API path (model_id empty)'
    required: false
```

So the two are **not** mutually exclusive — with `model_id` set the key is simply unused, and it is
`required: false`. `ai_claude-rollback-safety.yml` calls the same reusable workflow with the same
combination (`model_id` + `bedrock_role_arn`, plus the secret), so this matches the established
pattern in this repo rather than diverging from it.

The `issue_autodoc.yml` comparison isn't quite like-for-like: that one calls a different reusable
workflow (`autodoc-executor.yml@v3.5.0`) with its own inputs and its own dotCMS API-token secrets.

Keeping the secret passed through deliberately: if `vars.BEDROCK_MODEL_ID` were ever unset, the
orchestrator routes to the direct Anthropic path, which needs it.

You're right that the end-to-end call is still unproven — that's what the dispatch rehearsal is for,
before `POST_MERGE_TEST_PLAN_ENABLED` gets set.

---

Also confirmed the plan already posted on #36991 carries the full 40-char SHA, so it stays valid
under the tightened contract.

erickgonzalez:
Second round addressed in 152dd1c. The 🟠 High was real — and one of the Semgrep autofixes would have broken the workflow, details below.

### 🟠 High — partial re-run duplicates plans: **fixed**

Correct, and the contradiction you spotted was the tell: `finalize`'s rerun message promised
*"issues that already carry a plan for this merge SHA are skipped"* and that skip was never
implemented. The guard counted (`ALREADY -eq TOTAL`) instead of tracking *which* issues.

`prep` now carries the not-yet-planned subset and passes it as `POST_TO_CSV`. Proven against live
data, using #36991 (which really does carry this plan) and #37118 (which doesn't):

```
issues=[36991,37118] -> ALREADY=1/2  POST_TO=37118      partial re-run: skips the planned one
issues=[36991]       -> has_work=false                   nothing to do
issues=[37118]       -> ALREADY=0/1  POST_TO=37118      fresh
```

Old behaviour on the first row was `POST_TO=36991,37118` — a second identical plan on a live issue.

Plan **content** still covers every related issue and the `issues:` marker still names them all; only
delivery is narrowed. The prompt now separates the two explicitly and tells the model not to
"helpfully" expand back to the full list.

### 🟡 Medium — cross-repo in the body source: **fixed**

Fair, and cheap to close rather than reason about how narrow the window is. `issuesFromBody` is now
repo-scoped: qualified `owner/repo#N` refs and issue URLs count only when they point at the target
repo, and both are stripped before the bare `#N` scan so a foreign number can't leak through.
Foreign refs also feed `foreignByNumber`, so a body reference now disqualifies a matching branch
number the same way a Development link does.

```
This PR fixes: #31904                                              -> [31904]
This PR fixes: dotCMS/private-issues#671                           -> []
This PR fixes: dotCMS/core#31904                                   -> [31904]
This PR fixes: [#671](https://github.com/dotCMS/private-issues/issues/671) -> []
This PR fixes: dotCMS/private-issues#671 and #31905                -> [31905]
```

### 🔴 Semgrep — actions pinned, but **the suggested SHA is wrong**

Pinned both. Worth flagging before anyone clicks the one-click fix on the `github-script` finding:

```
Semgrep suggests: 60a0d83039c74a4aee543508d2ffb2c3793c4a9    (39 chars)
Real v7.0.1:      60a0d83039c74a4aee543508d2ffcb1c3799cdea    (40 chars)
```

It is not merely truncated — it diverges from character 29. Applying it verbatim pins to a commit
that does not exist and the workflow fails to resolve the action at runtime. Both pins here use SHAs
resolved from the GitHub API instead:

* `actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1`
* `actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0`

On the checkout pin: I did **not** use `8ade135a…`, which four workflows in this repo already use.
That SHA is actually **v4.1.0** (Sept 2023), yet it is labelled `# v4.2.2` in two files and
`# v4.2.0` in two others — all three labels disagree with the commit. I pinned to the same v4.4.0
SHA `cicd_pr_skill-lint.yml` uses, which I verified. The mislabelled pins elsewhere are pre-existing
and out of scope here, but they're worth a separate cleanup.

---

Re-verified after the changes: discovery still resolves **20 of 21** real PRs (#37167 correctly
excluded — its only link is `dotCMS/private-issues#671`), workflow YAML and all four shell blocks
parse, and skill-lint is clean.

wezell (commented):
My only question is I feel like we are saturating our runners and a lot of our slow pipeline is job and runner contention.   Would it be better if it ran on demand? on a tag?  that way whoever was in charge of doing the QA could tag the issue and a test plan would be generated for them?

wezell:
Third round addressed in 95865087. Both findings were 🟡 Medium; neither blocked, both closed anyway.

### 🟡 Medium — date-prefixed branches yield spurious candidates: **fixed**

Fair — `2024-01-15-hotfix` parsed as `[2024, 1, 15]`, and `validate()` cannot filter those, because low-numbered core issues like #15 really do exist. Took the floor suggestion rather than requiring the `issue-` prefix, which would have broken real bare forms like `36937-36938-roles-api-role-membership`:

`issuesFromBranch` now discards every token below `MIN_ISSUE_NUMBER = 10000`. core passed #10000 in 2018 and every component of a date prefix is ≤ 9999 by construction, so nothing legitimate is lost. The floor applies to the `issue-`-prefixed form too, deliberately: `issue-671-…` is exactly how #37167 pointed at private-issues#671, so an explicit-looking prefix deserves no more trust than a bare number.

```text
issue-31904-Key-Value-Field-Preserve-order -> [31904]         unchanged
issue-12345-43565-testing                  -> [12345, 43565]  unchanged
36937-36938-roles-api-role-membership      -> [36937, 36938]  unchanged
issue-37085-bouncycastle-185               -> [37085]         unchanged
issue-37085-samlbundle-26.08.21            -> [37085]         unchanged
nicobytes/36950-remove-dead-core-web-libs  -> []              unchanged
issue-xmllint-apt-hang                     -> []              unchanged
2024-01-15-hotfix                          -> []              was [2024, 1, 15]
2026-08-25-release-prep                    -> []              was [2026, 8, 25]
issue-671-cross-repo-trap                  -> []              was [671], previously only saved by Development attribution
1.2.3-something                            -> []              unchanged (dots never matched)
```

Re-ran discovery against **22 recently merged PRs** after the change (per this PR's own rule about touching that script): every result is identical to before — including #37167, which stays correctly empty and is now doubly defended — and the date cases above return `[]`.

### 🟡 Medium — pin verification: **verified, no change**

The API wasn't reachable from the review run; from here:

```text
repos/actions/github-script/git/ref/tags/v7.0.1 -> 60a0d83039c74a4aee543508d2ffcb1c3799cdea  (commit) — matches our pin
repos/actions/checkout/git/ref/tags/v4.4.0      -> 11d5960a326750d5838078e36cf38b85af677262  (commit) — matches our pin
```

Both are lightweight tags pointing straight at commits, and both resolve exactly to what this PR pins. Semgrep's suggested `60a0d83…ffb2c3793c4a9` remains the bogus one (39 chars, diverges at character 29), so nothing was changed — the pins stand as committed.

erickgonzalez:
Closing this PR. B/c @wezell infected it 😛

PR #37236, opened 2026-08-26
#37233: feat(users): return direct roles per user on GET /v1/users/filter

### LINKED ISSUE #37233: [Task] Users API: `GET /v1/users/filter` should be able to return each user's direct roles
## Context

The Angular Users portlet list (#36930 arc) shows each user's roles beyond the default access roles in a column. `GET /api/v1/users/filter` returns no role data — items are `user.toMap()` plus an optional `requestPassword` flag (`UserPaginator#addRequestPasswordAttr`).

The FE therefore fans out one `GET /api/v1/roles/users/{userId}` per row after every list load (`dot-users-list.store.ts:107-124`): 40 extra requests per page at the default `per_page=40`, each paying REST auth init. That endpoint also returns explicit **and** implicit roles (`loadRolesForUser(userId)` defaults `includeImplicitRoles=true`), so the FE filters access roles and the personal role out client-side.

Legacy reference: the Dojo Users portlet list grid had no roles column; its Roles tab listed **direct** memberships only (`UserAjax#getUserRoles` → `loadRolesForUser(userId, false)`).

## Requirements

- A single `GET /api/v1/users/filter` request can return, per user, the user's **directly assigned** roles — inherited/implicit roles excluded (legacy parity with the Roles tab, and consistent with the direct-membership semantics of the `/v1/roles/{roleid}/users*` endpoints from #37077 and #37070; inherited roles cannot be removed from a user, so listing them alongside a Remove action would mislead)
- Each role entry must carry at least `id`, `name`, and `roleKey`. `id` is required because roles created in the Roles portlet may have no key (#37209)
- The behavior is **opt-in**: callers that do not ask for roles get a byte-identical response to today. `UserPaginator` is shared; no extra payload or DB work for callers that don't need roles
- All existing `filter` parameters (`query`, `roleKey`, paging, sorting, `includeAnonymous`, `includeDefault`, `assetInode`/`permission`) keep working unchanged in combination
- No change to authorization of the endpoint
- OpenAPI documents the new option and the per-item role shape; regenerated `openapi.yaml` committed

## Acceptance Criteria

- [ ] Roles returned per user when requested; absent when not requested (existing payload unchanged)
- [ ] Only direct memberships are listed; a role held via inheritance (child of a held role, or the implicit CMS Anonymous role) is not
- [ ] Keyless roles appear with `id` and `name`, `roleKey` null/absent
- [ ] Integration tests: opt-in on/off, direct-vs-inherited, keyless role, combined with `roleKey` filter and paging
- [ ] Existing `UserResourceIntegrationTest` and Postman `UserResource` collection still green
- [ ] OpenAPI updated and regenerated

## Out of scope

- Changing `GET /v1/roles/users/{userIdOrEmail}`
- Inherited/implicit role expansion in the list (could be a later opt-in if a use case appears)

hassandotcms:
Proposed Changes

  - New includeRoles query param on GET /v1/users/filter. When true, each item gains a roles array of the user's directly assigned roles as {id, name, roleKey}; inherited roles and the personal role are excluded (legacy Roles-tab parity). Lets the Users list drop its per-row GET /v1/roles/users/{id} calls.
  - Opt-in: without the flag the response is byte-identical to today.
  - includeRoles=true requires CMS Administrator or Roles + Users portlet access (same gate as GET /v1/roles/users/{id}), otherwise 403.
  - OpenAPI updated; openapi.yaml regenerated.
  
  Checklist

  - [x] Tests — UserResourceIntegrationTest (7 new: opt-out parity, direct roles, inherited excluded, keyless role, roleKey + paging, 403 gate, root role named "User…" kept), UserPaginatorTest, UserResourceTest

This PR fixes: #37233

This PR fixes: #37233

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/rest/api/v1/user/UserResource.java
+++ b/dotCMS/src/main/java/com/dotcms/rest/api/v1/user/UserResource.java
@@ -411,13 +411,25 @@ public class UserResource implements Serializable {
 	 * @param permission       The permission type that Users may have on the previous asset.
 	 * @param roleKeys         Optional Role keys -- repeatable and/or comma-separated -- that restrict the results to
 	 *                         Users holding any of them, e.g. {@code roleKey=DOTCMS_BACK_END_USER}.
+	 * @param includeRoles     Opt-in flag. When {@code true}, every returned User carries a {@code roles} list with
+	 *                         its <b>directly assigned</b> Roles -- inherited Roles and the User's personal Role are
+	 *                         excluded -- as {@code {id, name, roleKey}} entries. Defaults to {@code false}, in which
+	 *                         case the response is identical to the one returned before this flag existed. Reading
+	 *                         role membership requires the same privilege as {@code GET /v1/roles/users/{id}}: CMS
+	 *                         Administrator, or access to both the Roles and the Users portlets; otherwise 403.
 	 *
 	 * @return A {@link Response} containing the list of dotCMS users that match the filtering criteria.
 	 */
 	@Operation(
 		operationId = "filterUsers",
 		summary = "Filter users",
-		description = "Returns a list of dotCMS users based on specified search criteria with pagination support"
+		description = "Returns a list of dotCMS users based on specified search criteria with pagination support. "
+				+ "Each item is the user's data map. When `includeRoles=true` is passed, each item additionally "
+				+ "carries a `roles` array listing the user's directly assigned roles as `{id, name, roleKey}` "
+				+ "objects (`roleKey` may be null for roles created without a key); roles held through the role "
+				+ "hierarchy and the user's personal role are not included. Requesting roles requires being a CMS "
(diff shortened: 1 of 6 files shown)


wezell:
Is there a way to get inherited roles with the API?

wezell (commented):
Is there any way or flag to include inherited roles with this api?

hassandotcms:
> Is there any way or flag to include inherited roles with this api?

currently no, in this api.

if consumer need inherited roles per user as well, it can use `GET /v1/roles/users/{id}`.

@wezell

hassandotcms:
> > Is there any way or flag to include inherited roles with this api?
> 
> currently no, in this api.
> 
> if consumer need inherited roles per user as well, it can use `GET /v1/roles/users/{id}`.
> 
> @wezell

@wezell can you please give a like 
or do you want me to add a flag for inherited roles in this api too?

wezell (approved):
Looks good even though it cannot be used to retrieve the canonical list of all roles that the user belongs to, including user roles and inherited roles.

PR #37244, opened 2026-08-26
fix(uve): lock layout canvas during save to prevent container data loss (#36197)

### LINKED ISSUE #36197: Layout: Lose `containers`/`content` because the editing canvas is never locked during save

gortiz-dotcms:
## Summary

Fixes dotCMS/core#36197 — UVE Layout tab data loss caused by the GridStack canvas never being locked while a layout save is in flight.

### Root cause

Every layout change emits `templateChange` → `updateTemplate$` → 5-second debounce → `POST /api/v1/page/{id}/layout`. The canvas remained fully interactive during the entire debounce window, the in-flight POST, and the `pageReload()` + `updateOldRows()` re-hydration cycle. Any edit made in that window was silently discarded when the server response overwrote local state. Because the backend performs a destructive diff (deletes any container absent from the incoming payload), mid-flight edits could permanently remove content.

---

### `TemplateBuilderComponent` (`template-builder.component.ts` / `.html`)

- **`@Input() disabled = false`** — new input that locks the canvas.
- **`applyGridDisabled(disabled)`** — called synchronously from `ngOnChanges` (before any render cycle):
  - `grid.disable()` / `grid.enable()` on the main grid and every subgrid — blocks drag, resize, and external drop at the GridStack event level.
  - Dispatches `keydown Escape` on `document` — closes any open PrimeNG dropdowns (container pickers, etc.) that are appended to `<body>` outside the `[inert]` subtree.
  - **Drag cancellation** — if a drag is in progress when the canvas locks, dispatches a synthetic `document mouseup` to terminate it. `suppressStoreUpdates` is set for the synchronous cancel window so that GridStack's forced commit events (and the subsequent `grid.load()` restore) do not write to the store. Placeholder widgets committed by the cancelled drag are removed via `removeWidget` before the lock completes.
  - **`preDragGrid` tracking** — records the specific GridStack instance (main grid or subgrid) that owns the dragged widget. Row drags restore the main grid; column/box drags restore only the affected subgrid via its own `save(false)` snapshot.
- **`suppressStoreUpdates` flag** — silences `moveRow`, `addRow`, `updateColumnGridStackData`, and `subGridOnDropped` calls inside GridStack event handlers during the cancel window, keeping the store consistent with pre-drag state.
- **Visual overlay** — `@if (disabled)` block rendered after the grid content with `bg-white/50 backdrop-blur-[2px]` and a centered `p-progress-spinner`. `[attr.inert]` on both the toolbar and the grid container blocks all keyboard/pointer interaction.

### `EditEmaLayoutComponent` (`edit-ema-layout.component.ts` / `.html`)

- **Component-local `#layoutSaveInFlight = signal(false)`** — replaces the `computed(() => uveStore.uveStatus() === LOADING)` approach. Decouples the canvas-disabled state from the global UVE status so that unrelated LOADING events (workflow lock toggles, page-api re-fetches via `withWorkflow.ts`) cannot unlock the canvas mid-POST or trigger a spurious spinner.
- **`nextTemplateUpdate` guard** — returns early when `#layoutSaveInFlight` is true, dropping any `templateChange` events that fire after the canvas is frozen (in-progress drags, open dropdown selections).
- **`tap` before `debounceTime`** — gates the Page Properties nav item immediately on any edit (`setUveStatus(LOADING)`) so users cannot navigate to page settings while layout changes are pending, while the actual canvas lock (`#layoutSaveInFlight`) only engages when the POST is actually sent.
- **`catchError` → `EMPTY`** — keeps the `switchMap` subscription alive on HTTP error so a failed save doesn't permanently disable autosave.

---

## Lock lifecycle

```
User edits → tap fires → setUveStatus(LOADING)   [Page Properties nav gated]
  → 5 s debounce settles
  → #layoutSaveInFlight = true                    [canvas locks + overlay appears]
  → POST /api/v1/page/{id}/layout
    (any new edit dropped by nextTemplateUpdate guard)
  → handleSuccessSaveTemplate()
  → #layoutSaveInFlight = false                   [canvas unlocks]
  → pageReload()
```

---

## Test plan

- [ ] Edit layout (move a row) → wait 5 s → "Saving…" toast appears → canvas shows overlay + spinner → cannot drag/add/resize during save + reload cycle → canvas unlocks after reload
- [ ] Start dragging a row → hold during save → row disappears on lock, no ghost row created, no row added to layout → canvas unlocks cleanly
- [ ] Start dragging toolbar "Add Box" → hold during save → box disappears on lock, no ghost box created → canvas unlocks cleanly
- [ ] Start dragging toolbar "Add Row" → hold during save → Row widget disappears, no row added to layout → canvas unlocks cleanly
- [ ] Edit layout → open "+" container dropdown in a box → wait 5 s → dropdown closes (Escape), overlay appears → canvas is frozen
- [ ] Edit layout → save fails (disconnect network) → canvas unlocks, `uveStatus` set to `ERROR`, user can retry
- [ ] `pnpm nx test portlets-edit-ema-portlet --testPathPattern=edit-ema-layout`
- [ ] `pnpm nx test template-builder --testPathPattern=template-builder.component`

---

https://github.com/user-attachments/assets/b77977ad-8b99-43ff-83f4-faa5705b2a4d

This PR fixes: #36197

### DIFF
--- a/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.spec.ts
+++ b/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.spec.ts
@@ -2,9 +2,9 @@ import { expect, describe } from '@jest/globals';
 import { SpyObject } from '@openng/spectator';
 import { Spectator, createComponentFactory, mockProvider } from '@openng/spectator/jest';
 import { MockComponent, MockProvider } from 'ng-mocks';
-import { of } from 'rxjs';
+import { Subject, of, throwError } from 'rxjs';
 
-import { provideHttpClient } from '@angular/common/http';
+import { HttpErrorResponse, provideHttpClient } from '@angular/common/http';
 import { provideHttpClientTesting } from '@angular/common/http/testing';
 import { signal } from '@angular/core';
 import { fakeAsync, tick } from '@angular/core/testing';
@@ -38,7 +38,7 @@ import {
     MockDotRouterJestService
 } from '@dotcms/utils-testing';
 
-import { EditEmaLayoutComponent } from './edit-ema-layout.component';
+import { DEBOUNCE_TIME, EditEmaLayoutComponent } from './edit-ema-layout.component';
 
 import { DotActionUrlService } from '../services/dot-action-url/dot-action-url.service';
 import { DotPageApiService } from '../services/dot-page-api/dot-page-api.service';
@@ -158,6 +158,8 @@ describe('EditEmaLayoutComponent', () => {
     });
 
     beforeEach(async () => {
+        jest.clearAllMocks();
+
         spectator = createComponent();
         component = spectator.component;
         dotRouter = spectator.inject(DotRouterService);
@@ -185,8 +187,18 @@ describe('EditEmaLayoutComponent', () => {
             expect(dotRouter.forbidRouteDeactivation).toHaveBeenCalled();
         });
 
-        it('should trigger a save after 5 secs', fakeAsync(() => {
+        it('should set uveStatus to LOADING immediately when templateChange is emitted', fakeAsync(() => {
             const setUveStatusSpy = jest.spyOn(store, 'setUveStatus');
+
+            templateBuilder.templateChange.emit();
+
+            // tap fires synchronously before debounce — no tick needed
+            expect(setUveStatusSpy).toHaveBeenCalledWith(UVE_STATUS.LOADING);
+
+            tick(5000); // flush timer to avoid pending-timer warning
+        }));
+
+        it('should trigger a save after 5 secs', fakeAsync(() => {
             const reloadSpy = jest.spyOn(store, 'pageReload');
 
             templateBuilder.templateChange.emit();
@@ -194,7 +206,6 @@ describe('EditEmaLayoutComponent', () => {
 
             expect(dotPageLayoutService.save).toHaveBeenCalled();
             expect(reloadSpy).toHaveBeenCalled();
-            expect(setUveStatusSpy).toHaveBeenCalledWith(UVE_STATUS.LOADING);
 
             expect(messageService.add).toHaveBeenNthCalledWith(1, {
                 severity: 'info',
@@ -224,17 +235,6 @@ describe('EditEmaLayoutComponent', () => {
             expect(store.isClientReady()).toBe(false);
         }));
 
-        it('should NOT set uveStatus to LOADING before the debounce fires', fakeAsync(() => {
-            const setUveStatusSpy = jest.spyOn(store, 'setUveStatus');
-
-            templateBuilder.templateChange.emit();
-            tick(1000); // Less than DEBOUNCE_TIME (5000 ms)
-
-            expect(setUveStatusSpy).not.toHaveBeenCalledWith(UVE_STATUS.LOADING);
-
-            tick(4000); // flush remaining timer to avoid pending-timer warning
-        }));
-
         it('should save right away if we request page leave before the 5 secs', () => {
             const saveTemplate = jest.spyOn(component, 'saveTemplate');
 
@@ -259,35 +259,72 @@ describe('EditEmaLayoutComponent', () => {
         });
     });
 
-    describe('LOADING guard and disabled binding', () => {
-        it('should drop templateChange events and not forbid navigation while uveStatus is LOADING', () => {
-            store.setUveStatus(UVE_STATUS.LOADING);
-            // The mock is shared across all tests — clear accumulated calls from earlier tests
-            // before asserting this specific action had no effect.
-            (dotRouter.forbidRouteDeactivation as jest.Mock).mockClear();
+    describe('Canvas lock (#layoutSaveInFlight)', () => {
+        it('should drop templateChange events and not forbid navigation while save is in-flight', fakeAsync(() => {
+            const saveSubject = new Subject();
+            (dotPageLayoutService.save as jest.Mock).mockReturnValue(saveSubject.asObservable());
 
+            // First emit starts the debounce; forbidRouteDeactivation called once
+            templateBuilder.templateChange.emit();
+            tick(DEBOUNCE_TIME); // debounce fires → POST sent → #layoutSaveInFlight = true
+
+            // Save still in-flight. Second emit should be dropped by the in-flight guard.
             templateBuilder.templateChange.emit();
 
-            expect(dotRouter.forbidRouteDeactivation).not.toHaveBeenCalled();
-        });
+            expect(dotRouter.forbidRouteDeactivation).toHaveBeenCalledTimes(1);
 
-        it('should process templateChange events and forbid navigation when uveStatus is not LOADING', () => {
+            saveSubject.complete(); // clean up
+        }));
+
+        it('should process templateChange events and forbid navigation when not saving', () => {
             templateBuilder.templateChange.emit();
 
-            expect(dotRouter.forbidRouteDeactivation).toHaveBeenCalled();
+            expect(dotRouter.forbidRouteDeactivation).toHaveBeenCalledTimes(1);
         });
 
-        it('should pass disabled=true to the template builder when uveStatus is LOADING', () => {
-            store.setUveStatus(UVE_STATUS.LOADING);
+        it('should pass disabled=true to the template builder while save is in-flight', fakeAsync(() => {
+            const saveSubject = new Subject();
+            (dotPageLayoutService.save as jest.Mock).mockReturnValue(saveSubject.asObservable());
+
+            templateBuilder.templateChange.emit();
+            tick(DEBOUNCE_TIME);
             spectator.detectChanges();
 
             expect(templateBuilder.disabled).toBe(true);
-        });
 
-        it('should pass disabled=false to the template builder when uveStatus is not LOADING', () => {
+            saveSubject.complete();
+        }));
+
+        it('should pass disabled=false to the template builder when not saving', () => {
             spectator.detectChanges();
 
             expect(templateBuilder.disabled).toBe(false);
         });
+
+        it('should unlock canvas (disabled=false) after save completes', fakeAsync(() => {
+            const saveSubject = new Subject();
+            (dotPageLayoutService.save as jest.Mock).mockReturnValue(saveSubject.asObservable());
+
+            templateBuilder.templateChange.emit();
+            tick(DEBOUNCE_TIME);
+
+            saveSubject.next(PAGE_RESPONSE);
+            saveSubject.complete();
+            spectator.detectChanges();
+
+            expect(templateBuilder.disabled).toBe(false);
+        }));
+
+        it('should unlock canvas (disabled=false) on save error', fakeAsync(() => {
+            (dotPageLayoutService.save as jest.Mock).mockReturnValue(
+                throwError(() => new HttpErrorResponse({ status: 400 }))
+            );
+
+            templateBuilder.templateChange.emit();
+            tick(DEBOUNCE_TIME);
+            spectator.detectChanges();
+
+            expect(templateBuilder.disabled).toBe(false);
+        }));
     });
 });
--- a/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.spec.ts
+++ b/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.spec.ts
@@ -326,5 +326,24 @@ describe('EditEmaLayoutComponent', () => {
 
             expect(templateBuilder.disabled).toBe(false);
         }));
+
+        it('should keep canvas locked through the pageReload window and unlock only when reload completes', fakeAsync(() => {
+            // Prevent the real pageReload from running so we can control when it "finishes"
+            const pageReloadSpy = jest.spyOn(store, 'pageReload').mockImplementation(jest.fn());
+
+            templateBuilder.templateChange.emit();
+            tick(DEBOUNCE_TIME); // POST fires and succeeds synchronously (default mock)
+            spectator.detectChanges();
+
+            // POST returned 200 but reload hasn't signalled LOADED yet — canvas must stay locked
+            expect(templateBuilder.disabled).toBe(true);
+            expect(pageReloadSpy).toHaveBeenCalled();
+
+            // Simulate the reload completing
+            store.setUveStatus(UVE_STATUS.LOADED);
+            spectator.detectChanges();
+
+            expect(templateBuilder.disabled).toBe(false);
+        }));
     });
 });
--- a/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.spec.ts
+++ b/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.spec.ts
@@ -262,6 +262,9 @@ describe('EditEmaLayoutComponent', () => {
     describe('LOADING guard and disabled binding', () => {
         it('should drop templateChange events and not forbid navigation while uveStatus is LOADING', () => {
             store.setUveStatus(UVE_STATUS.LOADING);
+            // The mock is shared across all tests — clear accumulated calls from earlier tests
+            // before asserting this specific action had no effect.
+            (dotRouter.forbidRouteDeactivation as jest.Mock).mockClear();
 
             templateBuilder.templateChange.emit();
 
--- a/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.spec.ts
+++ b/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.spec.ts
@@ -327,6 +327,23 @@ describe('EditEmaLayoutComponent', () => {
             expect(templateBuilder.disabled).toBe(false);
         }));
 
+        it('should unlock canvas when pageReload fails (uveStatus = ERROR) to avoid a permanent lock', fakeAsync(() => {
+            const pageReloadSpy = jest.spyOn(store, 'pageReload').mockImplementation(jest.fn());
+
+            templateBuilder.templateChange.emit();
+            tick(DEBOUNCE_TIME);
+            spectator.detectChanges();
+
+            expect(templateBuilder.disabled).toBe(true);
+            expect(pageReloadSpy).toHaveBeenCalled();
+
+            // Simulate re-fetch failure
+            store.setUveStatus(UVE_STATUS.ERROR);
+            spectator.detectChanges();
+
+            expect(templateBuilder.disabled).toBe(false);
+        }));
+
         it('should keep canvas locked through the pageReload window and unlock only when reload completes', fakeAsync(() => {
             // Prevent the real pageReload from running so we can control when it "finishes"
             const pageReloadSpy = jest.spyOn(store, 'pageReload').mockImplementation(jest.fn());
--- a/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.spec.ts
+++ b/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.spec.ts
@@ -224,6 +224,17 @@ describe('EditEmaLayoutComponent', () => {
             expect(store.isClientReady()).toBe(false);
         }));
 
+        it('should NOT set uveStatus to LOADING before the debounce fires', fakeAsync(() => {
+            const setUveStatusSpy = jest.spyOn(store, 'setUveStatus');
+
+            templateBuilder.templateChange.emit();
+            tick(1000); // Less than DEBOUNCE_TIME (5000 ms)
+
+            expect(setUveStatusSpy).not.toHaveBeenCalledWith(UVE_STATUS.LOADING);
+
+            tick(4000); // flush remaining timer to avoid pending-timer warning
+        }));
+
         it('should save right away if we request page leave before the 5 secs', () => {
             const saveTemplate = jest.spyOn(component, 'saveTemplate');
 
@@ -245,6 +256,35 @@ describe('EditEmaLayoutComponent', () => {
                 summary: 'Success',
                 detail: 'dot.common.message.saved'
             });
+        });
+    });
+
+    describe('LOADING guard and disabled binding', () => {
+        it('should drop templateChange events and not forbid navigation while uveStatus is LOADING', () => {
+            store.setUveStatus(UVE_STATUS.LOADING);
+
+            templateBuilder.templateChange.emit();
+
+            expect(dotRouter.forbidRouteDeactivation).not.toHaveBeenCalled();
+        });
+
+        it('should process templateChange events and forbid navigation when uveStatus is not LOADING', () => {
+            templateBuilder.templateChange.emit();
+
+            expect(dotRouter.forbidRouteDeactivation).toHaveBeenCalled();
+        });
+
+        it('should pass disabled=true to the template builder when uveStatus is LOADING', () => {
+            store.setUveStatus(UVE_STATUS.LOADING);
+            spectator.detectChanges();
+
+            expect(templateBuilder.disabled).toBe(true);
+        });
+
+        it('should pass disabled=false to the template builder when uveStatus is not LOADING', () => {
+            spectator.detectChanges();
+
+            expect(templateBuilder.disabled).toBe(false);
         });
     });
 });
--- a/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.spec.ts
+++ b/core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.spec.ts
@@ -167,6 +167,11 @@ describe('EditEmaLayoutComponent', () => {
         dotPageLayoutService = spectator.inject(DotPageLayoutService);
         messageService = spectator.inject(MessageService);
 
+        // Reset save mock to default — jest.clearAllMocks() does not reset mockReturnValue/
+        // mockImplementation overrides, so tests that call mockReturnValue(throwError(...))
+        // would contaminate subsequent tests that rely on the default of(PAGE_RESPONSE) behavior.
+        (dotPageLayoutService.save as jest.Mock).mockImplementation(() => of(PAGE_RESPONSE));
+
         store.pageLoad({
             clientHost: 'http://localhost:3000',
             language_id: '1',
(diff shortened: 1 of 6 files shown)


core-web/libs/template-builder/src/lib/components/template-builder/template-builder.component.ts
```diff
@@ -271,6 +276,42 @@ export class TemplateBuilderComponent implements OnDestroy, OnChanges, OnInit {
...
+    private applyGridDisabled(disabled: boolean): void {
+        if (disabled) {
+            this.grid.disable();
+            // Cancel any in-progress drag (internal row move or external toolbar widget drag-in).
+            // GridStack's DDDraggable registers a document keydown listener and calls cancel()
+            // on Escape, removing the drag clone from the DOM so the overlay becomes visible.
+            // PrimeNG dropdowns also close on Escape, covering the open-dropdown edge case.
+            document.dispatchEvent(
+                new KeyboardEvent('keydown', { key: 'Escape', bubbles: true, cancelable: true })
+            );
```
rjvelazco: **🔴 Critical — this Escape dispatch is a no-op, and `grid.disable()` *commits* the held drag instead of cancelling it.**

The comment describes GridStack behaviour that does not exist in the version we pin. `gridstack@8.4.0` has **no keyboard handling at all**:

```bash
grep -rn "Escape\|keydown" node_modules/gridstack/dist/*.js   # → no matches
```

`DDDraggable` (`dd-draggable.js:93-95`) registers only `mousemove` / `mouseup` on `document`. The Escape-cancel path landed in GridStack v10+.

**What actually ends the drag is the `grid.disable()` on the line above — and it commits it:**

`disable()` → `enableMove(false)` + `enableResize(false)` → `_prepareDragDropByNode` (`gridstack.js:1928-1932`) → `_removeDD(el)` → `DDDraggable.destroy()` → `if (this.dragging) this._mouseUp(this.mouseDownEvent)` (`dd-draggable.js:57-59`).

So `_mouseUp` fires `dragstop` → grid `change` → `store.moveRow(...)` → `templateChange` emit → **dropped by the new LOADING guard**. The row is now moved *locally but never sent to the server*, and `updateOldRows` then merges backend data onto those mutated rows by `y` position.

That is the exact local/server divergence class this PR exists to fix — reproduced by the fix.

**Suggested direction:** drop the Escape dispatch as a cancel mechanism. Before `grid.disable()`, check `this.draggingElement`; if a drag is in flight either defer the lock to the next `dragstop`, or snapshot the pre-drag rows (`grid.save()`) so the committed move can be reverted. At an absolute minimum, correct the comment so nobody downstream relies on a mechanism that isn't there.

**Secondary effect (also unintended):** PrimeNG `Select` does *not* close from this — `primeng-select.mjs:1283-1285` handles Escape on its own element keydown, and a synthetic event dispatched on `document` never reaches it. But `DynamicDialog` **does** (`primeng-dynamicdialog.mjs:740-747`). Realistic scenario: user drags a box (debounce starts) → within 5s opens "Add style classes" → at t=5s the synthetic Escape closes the dialog and discards their in-progress edit. The shell's `p-confirmDialog` and `dot-edit-ema-dialog` are exposed the same way.

core-web/libs/template-builder/src/lib/components/template-builder/template-builder.component.ts
```diff
@@ -271,6 +276,42 @@ export class TemplateBuilderComponent implements OnDestroy, OnChanges, OnInit {
                 isAnonymousTemplate: currentTemplate?.anonymous || this.template.anonymous // We createa a custom template for the page
             });
         }
+
+        if (changes.disabled !== undefined && this.grid) {
```
rjvelazco, line 295: **🟡 Minor — the `&& this.grid` guard silently drops the lock when `disabled` is already `true` at first change.**

`this.grid` is created asynchronously (`ngOnInit` → `defaultContainer$` → `requestAnimationFrame(() => this.setUpGridStack())`, line 265). Meanwhile `withUve.ts:50` initialises `uveStatus` to `LOADING`, so `disabled === true` on the *first* `ngOnChanges` is reachable — the guard skips, and the grid is later created **enabled** behind the overlay.

Same hole for subgrids created by the `rows.changes` handler (line 369/384) while `disabled` is true.

**Fix:** call `this.applyGridDisabled(this.disabled)` at the end of `setUpGridStack()`, and after `GridStack.addGrid(...)` in the `rows.changes` branch.

core-web/libs/template-builder/src/lib/components/template-builder/template-builder.component.ts
```diff
@@ -271,6 +276,42 @@ export class TemplateBuilderComponent implements OnDestroy, OnChanges, OnInit {
...
+            this.grid.enable();
+        }
+
+        const subGridEls = this.grid.el.querySelectorAll(
+            '.grid-stack'
+        ) as NodeListOf<GridHTMLElement>;
+
+        subGridEls.forEach((el) => {
+            el.gridstack?.[disabled ? 'disable' : 'enable']();
+        });
```
rjvelazco, line 371: **🟡 Minor — worth a comment saying why this manual loop exists, or someone will "clean it up".**

`disable(recurse = true)` already walks `n.subGrid` internally (`gridstack.js:1603, 1635-1640`). This loop is only necessary because this component creates subgrids via `GridStack.addGrid(el, ...)` rather than `subGridOpts`, so the engine may never have `node.subGrid` populated. One line of explanation prevents a future "this is redundant" removal that would silently break subgrid locking.

core-web/libs/template-builder/src/lib/components/template-builder/template-builder.component.html
```diff
@@ -116,3 +116,13 @@
...
+@if (disabled) {
+    <div
+        class="absolute inset-0 z-50 flex cursor-not-allowed items-center justify-center bg-white/50 backdrop-blur-[2px]"
+        data-testId="template-builder-disabled-overlay">
+        <p-progress-spinner
+            [style]="{ width: '3rem', height: '3rem' }"
+            strokeWidth="4"
+            aria-label="Saving layout" />
+    </div>
+}
```
rjvelazco, line 131: **🟠 Important — the overlay blocks pointers only; keyboard users can still mutate the builder behind it.**

No `inert`, no `aria-busy`, no focus containment. Tab order still reaches the toolbar (add row / add box / theme selector) and every box's delete + edit-classes control while the overlay is up.

That is the same failure mode as the drag issue: the store mutates locally, the emit is dropped by the `LOADING` guard, and `updateOldRows` merges backend data over the mutated rows → divergence.

**Fix:** add `[attr.inert]="disabled ? '' : null"` to the `p-toolbar` (line 2) and the `#templateContainerRef` wrapper (line 37), and `role="status" aria-busy="true"` on this overlay div.

Two smaller things in this block:
- **Line 126 — `aria-label="Saving layout"` is a hardcoded English string.** Every other user-facing string in this template goes through `| dm`. Also, a bare `aria-label` on the `<p-progress-spinner>` host is unlikely to be surfaced — PrimeNG exposes an `ariaLabel` input that lands on the element carrying `role="progressbar"`.
- **Line 121 — `bg-white/50` hardcodes a light-theme colour.** Worth checking against the PrimeNG theme tokens used elsewhere in the workspace.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.ts
```diff
@@ -54,6 +54,8 @@ export class EditEmaLayoutComponent implements OnInit, OnDestroy {
 
     protected readonly $layoutProperties = this.uveStore.$layoutProps;
 
+    protected readonly $isSaving = computed(() => this.uveStore.uveStatus() === UVE_STATUS.LOADING);
```
rjvelazco: **🟠 Important — `uveStatus` is a shared global flag, so unrelated features can unlock the canvas mid-POST.**

`uveStatus === LOADING` is not owned by the layout save. Reachable cross-feature writer from this very route: `withWorkflow.ts:179` patches `LOADING` on page lock/unlock (driven by the shell banner button, which renders above the router outlet), then `LOADED` at `:221` / `:231`.

**Consequence:** unlock the page while a layout POST is in flight → `LOADED` lands early → overlay disappears, grid re-enables, and the `nextTemplateUpdate` guard re-opens **while the POST and `pageReload()` are still running**. The guarantee this PR is buying evaporates in exactly the window it's meant to protect.

**Inverse, cosmetic:** any unrelated global `LOADING` (lock toggle, etc.) throws a "saving" spinner over the layout canvas.

**Fix — own the signal:**
```ts
readonly #layoutSaveInFlight = signal(false);
protected readonly $isSaving = this.#layoutSaveInFlight.asReadonly();
```
Set it in the `switchMap`, clear it in `handleSuccessSaveTemplate` / `handleErrorSaveTemplate`. That spans `pageReload()` without being unlockable by strangers, and it also resolves the Page Properties regression flagged below — one flag serving two purposes is the pattern that produced the original bug.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.ts
```diff
@@ -87,6 +89,13 @@ export class EditEmaLayoutComponent implements OnInit, OnDestroy {
...
+        // Drop edits that complete mid-flight: in-progress drags or open dropdowns
+        // can still emit templateChange after the canvas is frozen. Discarding them
+        // prevents corrupting the payload of the in-flight save.
+        if (this.uveStore.uveStatus() === UVE_STATUS.LOADING) {
+            return;
+        }
+
         this.dotRouterService.forbidRouteDeactivation();
         this.updateTemplate$.next(template);
         this.lastTemplate = template;
```
rjvelazco: **🟡 Minor — the early return also skips `this.lastTemplate = template` (line 101).**

The guard itself is a correct second barrier (and `updateOldRows` already sets `shouldEmit: false` in `template-builder.store.ts:434-439`, so this is good defence in depth).

But bailing before line 101 means the page-leave force-save at `:216` (`this.saveTemplate(this.lastTemplate)`) can persist a **stale** template. Compounds with the keyboard-bypass issue on the overlay: an edit made behind the overlay is dropped here *and* leaves `lastTemplate` pointing at the pre-edit state.

Consider assigning `lastTemplate` before the guard, or explicitly documenting that dropped edits are intentionally not persisted on leave.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.ts
```diff
@@ -134,10 +143,11 @@ export class EditEmaLayoutComponent implements OnInit, OnDestroy {
             .pipe(
                 // debounceTime should be before takeUntil to avoid calling the observable after unsubscribe.
                 // More information: https://stackoverflow.com/questions/58974320/how-is-it-possible-to-stop-a-debounced-rxjs-observable
-                tap(() => this.uveStore.setUveStatus(UVE_STATUS.LOADING)), // Prevent the user to access page properties
```
rjvelazco, line 137: **🟠 Important — removing this line silently drops the Page Properties gate.**

The comment on this line (`// Prevent the user to access page properties`) was load-bearing. `$menuItems` gates the `properties` nav item on `uveStatus === LOADING` (`dot-ema-shell.component.ts:143, 188`).

Moving the flag into the `switchMap` removes that gate **for the entire 5-second pending-save window**. The user can now open Page Properties with unsaved layout edits queued, and saving properties can trigger a page reload that races the pending debounce.

This is the same conflation the PR is (rightly) unwinding — one flag serving two purposes. Gate the properties item on a separate "pending layout save" signal, or on `dotRouterService`'s forbidden-deactivation state, which `nextTemplateUpdate` already sets.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.ts
```diff
@@ -134,10 +143,11 @@ export class EditEmaLayoutComponent implements OnInit, OnDestroy {
...
                 debounceTime(DEBOUNCE_TIME),
                 takeUntil(this.destroy$),
                 switchMap((layout: DotTemplateDesigner) => {
+                    // Lock the canvas only when the POST is actually sent, not on every edit.
+                    this.uveStore.setUveStatus(UVE_STATUS.LOADING);
```
rjvelazco: **🟠 Important — requirement 4 ("canvas unlocks on failure so the user can retry") doesn't actually hold: the save stream is dead after one error.**

There is no `catchError` inside this `switchMap`. An HTTP error propagates to the outer `.subscribe(next, error)` at line 166, which **terminates the `updateTemplate$` subscription for the component's lifetime**.

This is pre-existing, but the PR now unlocks the canvas on `ERROR` and thereby advertises a retry path that silently no-ops: every subsequent edit goes into a dead subject. Only the page-leave `saveTemplate()` path still works, so the user gets a re-enabled canvas that quietly saves nothing.

**Fix:**
```ts
.pipe(
    finalize(() => this.dotRouterService.allowRouteDeactivation()),
    catchError((e) => { this.handleErrorSaveTemplate(e); return EMPTY; })
)
```
on the inner `save()` pipe, keeping the outer subscription alive.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.spec.ts
```diff
@@ -247,4 +258,36 @@ describe('EditEmaLayoutComponent', () => {
             });
         });
     });
+
+    describe('LOADING guard and disabled binding', () => {
+        it('should drop templateChange events and not forbid navigation while uveStatus is LOADING', () => {
+            store.setUveStatus(UVE_STATUS.LOADING);
+            // The mock is shared across all tests — clear accumulated calls from earlier tests
+            // before asserting this specific action had no effect.
+            (dotRouter.forbidRouteDeactivation as jest.Mock).mockClear();
```
rjvelazco: **🟠 Important — this workaround is papering over shared mutable state between tests.**

`MockDotRouterJestService` is instantiated **once at factory-definition time** and shared across every test in this file, which is why this `mockClear()` and its explanatory comment are needed here.

`docs/frontend/TESTING_REVIEW_RULES.md` lists *"Tests share mutable state or depend on execution order"* as a **Critical 🔴** test violation.

**Fix:** add `jest.clearAllMocks()` to the file-level `beforeEach`, then delete this inline `mockClear()` and its comment.

rjvelazco (commented):
## Review — canvas lock during layout save

Thanks for the clear write-up in the description; the root-cause analysis is correct and made this reviewable. Full-file review of both components plus the UVE store features, cross-checked against the pinned `gridstack@8.4.0` and `primeng@21.1.3` sources in `node_modules`.

**Verdict: needs changes before merge.** Requirement 2 (don't lock during the debounce) is precisely implemented, but requirement 3 (cancel in-progress interactions) is unmet in a way that reproduces the original bug, and coupling the lock to `uveStatus` leaves it unlockable by unrelated store writers.

---

### What's well done

- **The diagnosis is right.** The debounce → POST → re-hydration window is exactly where the data loss lives.
- **Requirement 2 is met precisely.** Moving `setUveStatus(LOADING)` out of the pre-`debounceTime` `tap` and into the `switchMap` is the minimal correct change.
- **`ngOnChanges` ordering is correct.** `updateOldRows` (layout branch) runs *before* the grid is re-enabled, so re-hydration happens under the lock.
- **The store cooperates.** `withPageApi.ts:340-348` patches `uveStatus: LOADED` and `setPageAsset` in one synchronous tap, so `layout` and `disabled` reach `ngOnChanges` in a single CD pass — no interim "unlocked with stale layout" frame.
- **Defence in depth is sound.** `updateOldRows` already sets `shouldEmit: false` (`template-builder.store.ts:434-439`); the new `nextTemplateUpdate` guard is a correct second barrier.
- **Overlay placement is correct.** It sits outside the `@if (vm$ | async)` block, and `:host { @apply relative }` gives it the right containing block — so it covers the toolbar (the external drag-in source), not just the canvas.
- **Backward compatible.** `disabled = false` leaves the other consumer (`apps/dotcms-ui/.../dot-template-builder.component.html`) untouched.
- **Test hygiene is generally good** — `byTestId`, `spectator.setInput`, the real `UVEStore` rather than a stub, `fakeAsync`/`tick` with an explicit flush comment.

---

### Blocking (details inline)

| # | Where | Issue |
|---|-------|-------|
| 1 | `template-builder.component.ts:296-303` | 🔴 The Escape dispatch is a **no-op** — GridStack 8.4.0 has no keyboard handling (added in v10+). What actually ends the drag is `grid.disable()`, and it **commits** the move instead of cancelling it → the row moves locally, the emit is dropped by the new guard, and `updateOldRows` merges over it. Same divergence class the PR fixes. |
| 2 | `edit-ema-layout.component.ts:57` | 🟠 `uveStatus` is a shared flag — `withWorkflow.ts:179` (page lock/unlock, reachable from the shell banner on this route) can patch `LOADED` mid-POST and unlock the canvas early. |
| 3 | `edit-ema-layout.component.ts` (removed `tap`) | 🟠 The deleted line was gating the **Page Properties** nav item (`dot-ema-shell.component.ts:188`). That gate is now gone for the whole 5s pending-save window. |
| 4 | `edit-ema-layout.component.ts:148-150` | 🟠 No `catchError` in the `switchMap` — one HTTP error kills `updateTemplate$` for the component's lifetime, so requirement 4 advertises a retry that silently no-ops. |

Issues 2 and 3 both dissolve if the lock gets its own component-local signal instead of borrowing `uveStatus` — one flag serving two purposes is the pattern that produced the original bug.

### Also worth addressing

- **Keyboard bypasses the overlay** — no `inert`/`aria-busy`; tab order still reaches the toolbar and box controls.
- **Escape closes PrimeNG dialogs** — the dispatch's one real effect, and unintended: it discards an in-progress "Add style classes" edit.
- **Test interdependence** in the new layout specs (shared router mock, worked around with an inline `mockClear()`) — a Critical 🔴 per `docs/frontend/TESTING_REVIEW_RULES.md`.
- **Missing coverage** for unlock-on-ERROR and for the lock surviving `pageReload()` — the requirement most central to the fix.
- **Minor:** init race when `disabled` is already `true` before the grid exists; hardcoded `aria-label` (needs `| dm`); `bg-white/50` is light-theme-only.

### Note on `@Input()` over `input()`

`docs/frontend/ANGULAR_STANDARDS.md:51` prefers signal inputs, but this is a **justified** deviation — the JSDoc correctly explains the state must apply synchronously before render, which a post-CD `effect()` can't do, and the component's four existing inputs are all decorators. Consistent locally, no change requested. (An `@Input() set disabled(v: boolean)` setter would express the intent more directly than the `ngOnChanges` branch, if you want it.)

### Follow-up (not this PR)

`apps/dotcms-ui/.../dot-template-builder` is the other consumer of this component and has the same debounce-save shape without a lock. It won't regress from this PR, but it carries the same bug — worth a ticket.

---

**Note:** review is static — the two Jest targets in the test plan were not run.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-layout/edit-ema-layout.component.spec.ts
```diff
@@ -247,4 +258,36 @@ describe('EditEmaLayoutComponent', () => {
...
+
+            templateBuilder.templateChange.emit();
+
+            expect(dotRouter.forbidRouteDeactivation).not.toHaveBeenCalled();
+        });
+
+        it('should process templateChange events and forbid navigation when uveStatus is not LOADING', () => {
+            templateBuilder.templateChange.emit();
+
+            expect(dotRouter.forbidRouteDeactivation).toHaveBeenCalled();
```
rjvelazco: **🟠 Important — this assertion can pass on calls leaked from earlier tests, so it doesn't prove what it claims.**

Because the router mock is shared file-wide (see previous comment), a bare `toHaveBeenCalled()` here is satisfied by any earlier test that emitted a `templateChange`. The test would still be green if the guard blocked *everything*.

**Fix:** with `jest.clearAllMocks()` in `beforeEach`, this becomes a real assertion. Better still, assert the call carries this emission — `toHaveBeenCalledTimes(1)`.

core-web/libs/template-builder/src/lib/components/template-builder/template-builder.component.spec.ts
```diff
@@ -431,6 +431,93 @@ describe('TemplateBuilderComponent', () => {
...
+            };
+
+            beforeEach(() => {
+                mockGrid = {
+                    disable: jest.fn(),
+                    enable: jest.fn(),
+                    el: { querySelectorAll: jest.fn().mockReturnValue([]) }
+                };
+                // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                spectator.component.grid = mockGrid as any;
```
rjvelazco, line 469: **🟡 Minor — these tests verify a hand-rolled mock, not behaviour.**

Assigning `spectator.component.grid = mockGrid as any` overwrites internal state directly and needs two `eslint-disable @typescript-eslint/no-explicit-any` lines to compile. Every assertion in this block reduces to *"my fake's method was called"*.

Given the Critical finding above — `grid.disable()` commits the held drag rather than cancelling it — this block is the clearest possible demonstration of why the repo's testing rules discourage this: **all of these pass while the feature does the wrong thing.**

**Suggestion:** this spec already boots a real GridStack instance elsewhere in the file (unlike the layout spec, which `jest.mock`s the module). Assert against `grid.opts.disableDrag` / the `ui-draggable-disabled` class on real nodes instead of on fake method calls.

core-web/libs/template-builder/src/lib/components/template-builder/template-builder.component.spec.ts
```diff
@@ -431,6 +431,93 @@ describe('TemplateBuilderComponent', () => {
...
+                expect(mockGrid.disable).toHaveBeenCalled();
+            });
+
+            it('should dispatch a document Escape keydown event when disabled becomes true', () => {
+                const dispatchSpy = jest.spyOn(document, 'dispatchEvent');
+                spectator.setInput('disabled', true);
+                expect(dispatchSpy).toHaveBeenCalledWith(
+                    expect.objectContaining({ type: 'keydown', key: 'Escape' })
+                );
+                dispatchSpy.mockRestore();
```
rjvelazco, line 564: **🟠 Important — this test gives false confidence: it asserts the mechanism was *invoked*, not that anything was *cancelled*.**

`expect(dispatchSpy).toHaveBeenCalledWith(...)` is green as long as the line executes. It cannot detect that GridStack 8.4.0 has no Escape handler (see the Critical comment on `template-builder.component.ts:296-303`), which is the actual defect.

A test for this requirement has to assert the **outcome**: start a real drag, flip `disabled`, and assert no row was added/moved in the store — or assert on the pre/post `grid.save()` snapshot.

**Also missing from this PR's coverage:**
- **Requirement 4 (unlock on ERROR)** — no test. Make `DotPageLayoutService.save` throw and assert `templateBuilder.disabled` returns to `false`.
- **No positive counterpart** to `should NOT set uveStatus to LOADING before the debounce fires`. That test also passes if saving is broken entirely; after `tick(4000)` assert `setUveStatusSpy` **was** called with `LOADING`.
- **No test that the lock survives `pageReload()`** — the requirement most central to the fix.

rjvelazco (dismissed):

danielsolis-dotcms (dismissed):

danielsilva-dotcms (approved):

Issue #37003, opened 2026-08-11
Experiments Portlet — Screen 2: Create/Update (/experiments/new + /:id/configuration)


oidacra:
## Description

Creation is a routed screen, not a dialog: `/experiments/new` **is** the Configure screen with an empty draft. This retires the current drawer (name + description only, page implied by the UVE route context) and lets a user create, configure and start an experiment end-to-end without opening the page editor.

Design: [approved prototype](https://claude.ai/design/p/36aed5a6-ca42-4624-b187-a465a23ce909?file=Experiments.dc.html&via=share), `edit` screen + Select A Page modal + Add Variant modal.

### Routing and creation flow

- `/experiments/new` and `/experiments/:experimentId/configuration` render the same component in `new` vs `edit` mode.
- **Autosave is the creation mechanism** (design footer: "Changes are saved as you edit." — no Save button): in `new` mode, debounced autosave issues the `POST` as soon as `name` + `page` exist, then `replaceUrl` to `/experiments/:id/configuration`; every subsequent change is a debounced `PATCH` per field group. **Hard dependency on #36988's PATCH fix** — without it every autosave call creates an empty Rule.
- **Page preselection via query param**: `/experiments/new?pageId=<uuid>` or `/experiments/new?url=/index` (path resolved by HTMLPAGE search on the current site) prefills the Page card and skips the picker. Invalid param → inline message, never a blank screen. (`?page=` avoided — collides with pagination; `?url=` matches UVE.)
- Locked state when `status ∈ {RUNNING, ENDED, ARCHIVED}`: banner per the design, every field disabled, add/delete variant hidden; RUNNING copy differs from generic read-only copy.

### Layout

Fixed header (back · title falling back to "New Experiment" · status tag · subline `{pageTitle} · {pagePath}` or "No Page selected" · View Results / Stop when applicable) + scrolling card body (max-width 1040 centered) + pinned footer (hint · Back To Experiments · primary Start/Schedule). Header kebab with end / abort / cancelSchedule / pushPublish / addToBundle gated by `AllowedActionsByExperimentStatus` (parity with the current screen's menu).

### The five cards

1. **Details** — name (required) + description textarea.
[...]

PR #37256, opened 2026-08-27
feat(experiments): flagged UVE entry point and variant Edit Content round-trip (#37005)

### LINKED ISSUE #37005: Experiments Portlet — UVE integration behind FEATURE_FLAG_EXPERIMENTS_PORTLET + variant Edit Content round-trip

oidacra:
### Proposed Changes

* Adds `specs/37005-experiments-uve-integration/spec.md` — the Spec-Kit specification for #37005. Spec only; the implementation lands in this same PR.

The spec covers two halves of the work: switching the Universal Visual Editor's Experiments entry point to the new portlet behind a runtime switch, and completing the portlet's Variants card with the Edit Content round-trip into UVE.

Three premises in the issue did not survive contact with the code. Each is resolved in the spec as D1/D2/D3, and both #37005 and #37008 were corrected on GitHub to match.

**D1 — the flag named in the issue is not free.** The issue proposed reviving `FEATURE_FLAG_EXPERIMENTS`, described as "declared, zero consumers today". That holds for the frontend enum entry only. The property is the live backend kill-switch for the whole Experiments feature — `ConfigExperimentUtil.isExperimentEnabled()`, gating experiment JS injection (`ExperimentWebAPIImpl:338`) and experiment resolution during page render (`HTMLPageAssetRenderedAPIImpl:176,239`). Setting it to `false`, the issue's intended default, would stop running experiments from serving to visitors. The spec requires a new dedicated switch, `FEATURE_FLAG_EXPERIMENTS_PORTLET`, shipped explicitly `false`. The explicit default is required rather than optional: `getFeatureFlags()` and `withFlags()` both map an absent flag to enabled, so a merely-declared flag ships on.

**D2 — the flag-on destination.** The issue deferred it. The spec settles it as the site-wide list filtered to the page in hand, with defined behavior for zero, one and many experiments, and a filter the editor can clear.

**D3 — no "edited" signal.** The issue asked a variant's meta line to read "Edited in the Universal Visual Editor". No such signal exists on the variant model on either side of the wire, and the legacy per-page card shows no edit indication at all. The spec forbids inventing one. The Variants card as shipped by #37003 (PR #37064) makes no per-variant claim about content state, so FR-007 is a constraint to hold rather than a change to make: enabling Edit Content must add nothing that states or implies an edit state. The one string this work removes is the disabled button's tooltip, `experiments.configure.variants.edit-content.unavailable`, which stops being true the moment FR-001 lands.

### Checklist

- [x] Tests — written first, developer-approved and confirmed failing before each implementation commit, per Constitution Principle V. Roughly 90 tests across the four stories: the switch and deep-link builder, the round-trip, the switch-off destinations, the page filter, and the read-only rules. Where an assertion described behaviour that was already correct, it was inverted and confirmed failing rather than counted as Red. Three exclusions are recorded and signed off rather than silent: SC-002's flip time and SC-004's interaction time are human measures (the mechanisms behind both are asserted), and SC-003's participation comparison needs live visitor traffic. Full-journey E2E belongs to the dedicated E2E issue per the spec's Dependencies.
- [x] Translations — seven keys added for the page filter and its page-scoped empty state, one repurposed, none removed. The locked-page reason reuses the existing `experiment.configure.edit.page.blocked` copy, which already reads as a full sentence. `experiments.configure.variants.edit-content.unavailable` was the disabled button's "available in an upcoming release" tooltip and is now the refusal reason shown when a variant's page cannot be opened.
- [x] Security Implications Contemplated — no new surface. The visitor-facing experiments kill-switch is protected from being repurposed or disabled (FR-014, FR-015a), which is the security-relevant risk here, and `ConfigExperimentUtilTest` now asserts that as a negative rather than leaving it a promise.

### Walkthrough

Recorded end to end on a local instance with the flag on: editor on `/index` → A/B → the list narrowed to that page → New with the page prefilled → the name typed → Save Draft → the crumb back to the still-narrowed list. The six frames and the recording are attached to this PR, in the order the captions below name them.

**The list, narrowed to the page.** No page-filter chip on the toolbar: it repeated what the trail and the page-scoped empty state already say, and its three jobs — the label, the way back to the page, the clear — were covered by the page's own crumb above it.

**New, naming itself on the trail.** `Home > Experiments List` in the path, **New Experiment** as the title, and the Page field prefilled from the narrowing. Before this, nothing put Configure or Results on the trail — it ended at the list's crumb, so the shell titled both screens "Experiments List", the screen the editor had just left, and left the list out of the path instead of in it.

**Save Draft: the New screen becomes the Configure screen.** The address swaps to the experiment's own and the crumb follows it in place — one crumb, not two. The name is in the header only; the crumb says where you are, not which experiment. (The recording's instance still serves the previous copy of `experiment.container.configuration.title`; that key now reads "Configure Experiment".)

<details>
<summary>The other three frames</summary>

**1 — the editor on Home.** A/B in the right rail is the entry point; the switch selects the destination and nothing else.

**4 — the typed name reaches the header, not the crumb.**

**6 — the `Experiments List` crumb returns to the list still narrowed,** with the Configure crumb truncated away and the new draft as the first row.

</details>

### Additional Info
* **The three human measures, as signed off at T112.** None of them is a manual step any more.
  * **SC-002 — covered by tests, no timing taken.** The operator flip is asserted as a sequence rather than as a duration: `dot-ema-shell.component.spec.ts` pins the destination with the switch off (`:1558`), with it on (`:1569`), the fallback when the switch cannot be read (`:1595`), and the **off → on → off** round trip (`:1610`) — written as a sequence deliberately, since a one-way test passes on a latched value. `experiments-portlet-switch.util.spec.ts` pins the reader itself. Note the reader is `DotPropertiesService.getKey`, not `getFreshFeatureFlag`: it was changed during review because the normaliser behind `getFreshFeatureFlag` maps an **absent** key to *enabled*, which would have switched the entry point on for a build that never set the property. Both readers are uncached, so a flip still lands on the next gesture.
  * **SC-004 — waived.** No measurement. The Edit Content dialog on a variant renders no workflow action bar, so the round trip cannot be timed to completion; `GET /api/v1/workflow/contentlet/<inode>/actions` answers Save | Publish | Move, so the actions exist and the gap is in the dialog, outside this PR's diff. Recorded as waived rather than as passing.
  * **SC-003 — waived.** Works as is. `ConfigExperimentUtilTest` asserts the visitor-facing kill-switch is untouched by this work, which is the mechanism SC-003 checks; a participation comparison over live traffic is not scheduled by this issue.

* **QA cases closed here.**
  * **TC-030 — the Scheduling card showed no start date for a running experiment.** Two independent causes. The form filled once per experiment *identity*, so a transition's answer never reached it — starting an experiment with no schedule is the server dating it for real (`ExperimentsAPIImpl.startNowScheduling`), and the card locked showing the two empty pickers it was started with; it now fills once per experiment *state*, and the transition reducer moves the saved baseline with it so the refill does not read as unsaved work. Separately, the pickers' `minDate` was doing double duty — it is the floor for *choosing* a start, and PrimeNG also renders no value below it, so the field blanked for **every** RUNNING or ENDED experiment, not only one started in the session. The pickers now widen their bounds to admit the value they carry; the rule refusing a past start stays in the form's schema, so nothing became choosable. The note beside them claimed "starts automatically on <date>" for an experiment already running, and now says "started on" once the moment has passed.
  * **TC-011 — an unresolvable `?pageId=` or `?url=` on Configure.** Already handled by the Page card reading `pagePrefillError`; verified in the browser for **both** params, which render "The Page in this link could not be found. Select a Page to continue." with the Select escape rather than the card's empty state. Covered by tests at both layers: the store distinguishes not-found from a failed lookup and guards a `?pageId=` that is not shaped like an identifier, and the card asserts it renders the reason instead of looking unset.
  * **TC-031 — Clear Schedule then Save Draft kept the old dates.** Fixed in the frontend, and **no API contract change**. `PATCH /api/v1/experiments/{id}` applies only the keys its body carries — `ExperimentsResource` builds the update with `if (experimentForm.getScheduling() != null)` — so the `null` the form was sending for an emptied card meant "leave it alone". The clear was already expressible: `Scheduling` is a pair of `Optional<Instant>`s, and `ExperimentsAPIImpl.emptyScheduling` reads a range carrying neither date as no schedule at all — the state `startNowScheduling` replaces when an experiment is started without one. The form now sends that empty range instead of `null`. Measured against a running instance before and after: PATCH with `scheduling: null` answers 200 and keeps the dates; with an empty range it clears them, and Clear Schedule → Save Draft now persists `{startDate: null, endDate: null}`.

* **What shipped, by user story.** US1 the variant round-trip (Edit Content opens the variant in UVE; the chip returns to the originating experiment). US2 the switch-off guarantee (every legacy destination byte-identical, enforced by destination-level tests). US3 the switch-on entry point (site-wide list filtered to the page, the narrowing visible on the breadcrumb trail and named in the page-scoped empty state, and carried through every door out and back). US4 read-only for the control, any non-draft experiment, and a page locked by another user — with the lock reason now *stated* inline, not only on hover.

* **FR-021c is met in its visible half only, on purpose.** The spec asked for the narrowing to be *visible* and *clearable*, and the first implementation did both with a toolbar chip. The chip is gone: the trail carries the page, and the page's own crumb is the way back to the editor. Two things went with it, and neither is buried here.

  The clear affordance: an editor who arrives narrowed cannot widen to the site-wide list from inside the app. The portlet is opt-in, so there is usually no `experiments` entry in `/api/v1/menu` either — verified on a local instance, 43 items and none of them this one. And on a fresh session (a bookmark, or cleared session storage) the page's crumb is not on the trail, so with rows present the only sign of the narrowing is the repeated value in the `Page` column.

  The store half that would have provided the clear is deleted rather than left unreachable: `pageAssetFilterChanged` had no dispatcher outside its own spec once the chip went, so the narrowing now has exactly one writer — the address. The gap is stated on `$pageFilter` where the next reader will be. The next iteration moves list/new/config into a UVE drawer, where a chip on a toolbar has nowhere to sit and this affordance gets designed with the rest of it.

* **Three unrelated keys ride along.** `analytics.error.not.available`, `.subtitle` and `.retry` in `Language.properties` are consumed by `dot-analytics-error.component.ts`, which is not in this diff — a pre-existing gap where the component shipped without its copy. Fixed here because it surfaced while checking this PR's own keys, but it belongs to neither #37005 nor this PR's story.
* Spec review page: https://claude.ai/code/artifact/8deabd7c-6fb0-40a1-9241-7f5c111ec019
* Depends on #36989 (Screen 1), #37003 (Screen 2 — delivers the Variants card this work completes), #37004 (Screen 3).
* Migration #37008 retires `FEATURE_FLAG_EXPERIMENTS_PORTLET`, not `FEATURE_FLAG_EXPERIMENTS`.
* D2 confirmed as specified: with the switch on, the UVE Experiments button leads to the site-wide list filtered to the page in hand. FR-021a-c, US3 and SC-009 stand.
* 34 functional requirements, 9 success criteria, 4 user stories, 9 edge cases. No open `[NEEDS CLARIFICATION]` markers.
* **Test-tooling fix picked up while implementing this issue.** `libs/portlets/dot-experiments/portlet/jest.config.ts` set `reporters: [['jest-junit', …]]`, which *replaces* Jest's default reporter instead of adding to it — so a failing run in that project printed a coverage table and a non-zero exit and no test results at all: no suite name, no failing test, no assertion. Diagnosing a Red run meant re-running everything through `npx jest --reporters=default`. Adding `'default'` back restores the console output; the JUnit XML is unchanged and still carries all 973 tests. Two other configs share the pattern and are deliberately left alone here — `apps/dotcdn/jest.config.ts` and `libs/portlets/dot-experiments/data-access/jest.config.ts`.
* `mode` on the variant deep link sends `UVE_MODE`, not `DotPageMode`. The two carry identical strings for EDIT/PREVIEW and diverge only at LIVE (`ADMIN_MODE` vs `LIVE`), and `#getPageParams()` validates `?mode=` against `Object.values(UVE_MODE)` — so a `DotPageMode.LIVE` would have been silently rewritten to EDIT. The planning notes claimed the opposite; corrected.
* **CI note — this PR surfaces pre-existing breakage in `block-editor`.** Adding two constants and one enum entry to `libs/dotcms-models` puts every dependent project into Nx's affected set, and `block-editor` imports that lib in 31 files. Its own test suite has **37 failures** — `Unexpected "TestComponent" … is marked as standalone and can't be declared in any NgModule`, an Angular-migration breakage in its own specs. Every `block-editor` file is byte-identical to `origin/main`, so this PR did not cause them; it makes the project *affected*, which is what runs the suite. The rot went unnoticed precisely because `block-editor` is rarely in the affected set. A first local run appeared to pass because Nx served that task from cache — `--skip-nx-cache` reproduces the 37 failures reliably.

  `dotcms-js`, `dotcms-webcomponents` and `dotcms-block-editor` also enter the affected set and fail on a missing `karma` module (karma is absent from `package.json` entirely), but all three carry `tags: ["skip:test"]`, so CI's `--exclude=tag:skip:test` skips them. `block-editor` carries no such tag.

  Every project this PR actually changes is green. `block-editor` deserves its own issue; flagging rather than fixing it here.
* The variant return destination is resolved by **origin**, not by the switch. Branching on the switch alone makes FR-018 and FR-005/FR-027 mutually exclusive on a supported path: with the switch off an editor can still reach the portlet from the main navigation (FR-026) and still open a variant (FR-027), and a switch-only branch would return them to a legacy screen they never came from. The outbound leg carries an origin marker; the switch is the fallback for a deep-linked variant that has none.

Closes #37005

This PR fixes: #37005


erickgonzalez (dismissed):
Feedback:
- Missing user story to turn off the FF in case something goes wrong

freddyDOTCMS (dismissed):

oidacra:
Spec updated after approval, re-review requested.

Commit 2085fc4169 rewords FR-007 and D3 (plus the #37003 dependency line and the D3 rationale in Section D3). Reason: the spec was written against a Variants card that rendered a fixed per-variant meta line ("No content changes yet"). The implementation of #37003 (PR #37064, merged Aug 31) shipped the card as a table with no per-variant state text, so that premise no longer holds. FR-007 keeps the same requirement (the card must not assert whether a variant's content was modified) but now reads as a constraint to hold rather than a change to make, and it names the string this work actually removes: the disabled Edit Content tooltip `experiments.configure.variants.edit-content.unavailable`. FR-007a and SC-008 are untouched; no FR ids changed.

Also recorded: the open PM/PO question on D2 (switch-on destination is the site-wide list filtered to the page) was resolved as specified, so FR-021a-c, US3 and SC-009 stand.

Planning artifacts added since approval: data-model.md and contracts/ (commits 5624aedee4 and 37e8ba6534). plan.md, research.md and tasks.md are gitignored per repo convention.

@freddyDOTCMS could you re-approve the spec? @erickgonzalez FYI.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-configure/components/dot-experiments-configure-variants/dot-experiments-configure-variants.component.ts
```diff
@@ -392,7 +467,8 @@ export class DotExperimentsConfigureVariantsComponent {
                 id: 'variant-preview',
                 label: this.#dotMessageService.get('experiments.configure.variants.action.preview'),
                 icon: 'visibility',
-                // Previewing a variant is a UVE round-trip, the same one Edit Content waits on.
+                // TODO(#37005 US4): wire to onEditContent with PREVIEW once the read-only
+                // rules land. Edit Content on the row already opens the editor.
                 disabled: true
```
adrianjm-dotCMS: Orphan TODO: the PR says `Closes #37005`, so `TODO(#37005 US4)` points at an issue that is closed the moment this merges — and the kebab's Preview then stays disabled for good. The row-level Preview button does work, so it is only this menu item. Either wire it to `onEditContent(row)` with `UVE_MODE.PREVIEW`, or re-point the TODO at the follow-up issue (#37008 or a new one).

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-list/dot-experiments-list.component.ts
```diff
@@ -261,12 +266,52 @@ export class DotExperimentsListComponent {
...
+     * `null` when there is no page filter. When there is one but `pageInfoByPageId` cannot resolve
+     * it — the page was deleted after the editor left it — the empty state still says so, rather
+     * than reading as a page that merely has no experiments.
+     *
+     * The narrowing arrives in the address and there is no control here that widens it: FR-021c's
+     * "starting point, not a cage" half is not met, and is deliberately left for the UVE drawer
+     * that takes over this screen. With rows present the only sign of it is the `Page` column and
+     * the page's own crumb, which a fresh session does not have.
+     */
+    readonly $pageFilter = computed<{ path: string; resolved: boolean } | null>(() => {
```
adrianjm-dotCMS, line 288: FR-021c: I would like the missing half tracked outside the code. With `pageAssetFilterChanged` gone there is no way to widen back to the site-wide list from inside the app, and on a fresh session (bookmark, cleared session storage) the page crumb is not on the trail either — so the only sign of the narrowing is the repeated value in the Page column. The PR body and this JSDoc are honest about it, but a user-visible gap deserves an issue rather than a note the next reader has to stumble on. Could you open one and link it here?

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-toolbar/dot-uve-toolbar.component.ts
```diff
@@ -248,6 +283,44 @@ export class DotUveToolbarComponent {
         );
     }
 
+    /**
+     * Chip for the control variant previewed from the portlet, or `null` when that is not the
+     * case. Same shape the store builds for a real variant, so the action handler and the template
+     * need no special case: the id stays `variant` and the back arrow means the same thing.
+     */
+    #controlFromPortletProps(): InfoOptions | null {
+        const { queryParams } = this.#activatedRoute.snapshot;
```
adrianjm-dotCMS, line 304: Non-reactive read inside a computed: `$infoDisplayProps` (line 164) calls this, and `this.#activatedRoute.snapshot.queryParams` is not a signal — so the chip only re-derives when the store signals happen to change. It works today because arriving at a variant also moves `pageParams()`/`pageExperiment()`, but nothing enforces that pairing. `toSignal(this.#activatedRoute.queryParams)` would make the dependency explicit.

adrianjm-dotCMS (commented):
Three notes from a read of the production diff — none of them blocking on their own.

adrianjm-dotCMS (dismissed):

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-configure/components/dot-experiments-configure-variants/dot-experiments-configure-variants.component.ts
oidacra: Done — wired, in 7561d65567.

You were right that it was only this item, and its premise turned out to be already met: `row.editorMode` (`:193`) resolves to `PREVIEW` for the control and for any read-only experiment, and `onEditContent` already hands that to the link builder. So nothing was waiting on the read-only rules — they landed here.

`onEditContent(row, mode = row.editorMode)` now takes an override and the entry passes `UVE_MODE.PREVIEW` explicitly, rather than following the row. That distinction is the point of it sitting beside Edit Content: the kebab renders only on a non-control variant of an editable draft, where the row's own button says *Edit*, so this is the only way to look at such a variant without opening it for writing.

The `disabled: true` and the TODO are gone rather than re-pointed. Covered by a test that the entry navigates with `mode=PREVIEW` whatever the row would have done.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-list/dot-experiments-list.component.ts
oidacra: Agreed, and it is tracked now: **#37478** — *Experiments Portlet — the page's experiments as a UVE panel, behind FEATURE_FLAG_EXPERIMENTS_PORTLET*. It names FR-021c in the description and carries it as an acceptance criterion: "The panel offers an explicit way into the full portlet (FR-021c's 'clearable', honestly resolved)".

The JSDoc now points at that number instead of describing an unnumbered future (d4d4b6cf9c). I did not open a second issue — the gap and its fix are the same piece of work, and splitting it would fragment the tracking rather than sharpen it. Say the word if you would rather have a narrow FR-021c ticket that survives #37478 being descoped.

Worth adding the reason the gap is wider than "no chip": the portlet is opt-in, so on a stock instance there is no `experiments` entry in `/api/v1/menu` either — measured on a local instance, 43 items and none of them this one. So clearing the filter would have been a dead end even with a control for it. That is why #37478 resolves it by making the surface page-scoped by construction, with a deliberate link out, rather than by putting the chip back.

One thing has changed since you wrote this. A narrowing by **path** now has a way out: `?url=` was previously ignored altogether — a UVE-shaped address pasted into the list answered a request for one page with every experiment on the site — and it is now a real narrowing, so one that resolves to no page reads as "nothing matched" and the empty state's button clears it. The `?pageId=` case you are describing is unaffected: it shows the page-scoped empty state, which offers to create the first experiment for that page.

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-toolbar/dot-uve-toolbar.component.ts
oidacra: The mechanic is exactly as you describe, and you are right that nothing said why it is safe. I have written the invariant down rather than changed the read (69cb1bb2d1) — three reasons, and I am happy to be talked out of any of them:

**1. The marker is write-once per component instance.** It can only arrive on the navigation *into* UVE from the portlet's Configure screen, which constructs a new toolbar whose first evaluation already sees the param. Nothing in a live session adds, removes or rewrites `experimentReturn` on an instance that already exists — so the pairing you spotted is not luck, it is that the only writer is the entry navigation itself. The comment now says that, and says what would break it: any flow that sets the marker on an already-mounted editor.

**2. `toSignal` would not make it fresher where staleness is conceivable.** UVE writes its own query params through `Location.go` (`#updateLocation` in the shell), which bypasses the router — `route.queryParams` does not emit on it. So the signal would be exactly as stale as the snapshot for every param UVE moves itself; it would only track real router navigations, which are the case that already works.

**3. It costs more than it looks.** The toolbar spec's `ActivatedRoute` mock exposes only a `snapshot` getter, and about ten test sites set `routeQueryParams` *inside* the `it`, after the component exists, relying on that getter being re-read on the next `detectChanges`. Moving to `toSignal` means giving the mock a Subject and rewriting all of them — churn in the best-covered part of this diff, to enforce an invariant that currently has one writer.

If you would still rather have it enforced than documented, say so and I will do the refactor — it is mechanical, just not free.

adrianjm-dotCMS (dismissed):

adrianjm-dotCMS (approved):

PR #37257, opened 2026-08-27
docs(experiments): spec for swapping the list to the server-side contract (#37007)

### LINKED ISSUE #37007: Experiments Portlet — swap the List to the server-side contract

oidacra:
Spec-first PR for #37007, using the single-PR flow: the specification lands as the initial commit and the implementation will land in this same PR once its entry condition is met.

Closes #37007. Visual review page (summarizes; the spec decides): https://claude.ai/code/artifact/908c4144-2b25-40f9-a2d4-abdc18b1000b

### What the spec defines

Swap the Experiments portlet list from the #36989 client-side interim (download every experiment plus a bulk page lookup, then fake paging, sorting, filtering, counts and site scoping in the browser) to the server-side contract on `GET /api/v1/experiments`. Pure consumer swap: no new endpoint; every new query parameter is additive and optional, so existing callers are untouched.

- One request per view change carrying `page`, `per_page`, `orderby`, `direction`, `filter`, `status`, `goal`, `created_by`, `running_from`, `running_to`, `siteId`; the pager total comes from `pagination.totalEntries`; the interim client-side utilities are deleted.
- The URL contract stays byte-identical for every pre-existing parameter — the #37006 deep-link E2E suite must pass unchanged. The three new parameters appear only when their filter is active.
- Two stakeholder-requested filters (added 2026-08-27): by creator (`created_by`) and by running date range (`running_from`/`running_to`, inclusive overlap semantics, orthogonal to the status filter). Their two toolbar controls are the only permitted exception to the "no component changes" boundary.
- Accepted behavior changes, owned by #36823: permission narrowing (users see only experiments on pages they can edit) and free-text search no longer matching descriptions.

### Entry condition: implementation blocked

#36823 (the backend contract, expanded scope) is open with no implementation PR, and none of the contract is on `main` yet. Implementation in this PR starts only after the expanded #36823 lands and the dependency state is re-verified. Details in the spec's Scope Note and Dependencies sections.

### DIFF
--- a/specs/37007-list-server-side-swap/spec.md
+++ b/specs/37007-list-server-side-swap/spec.md
@@ -1,0 +1,369 @@
+# Feature Specification: Experiments List on the Server-Side Contract
+
+**Feature Branch**: `oidacra/experiments-portlet-list-server-side-swap`
+
+**Created**: 2026-08-26
+
+**Status**: Draft
+
+**Type**: Task (interim-to-final contract swap; no user-visible feature change intended)
+
+**Epic**: [#36763 — Experiments: A/B Testing v2](https://github.com/dotCMS/core/issues/36763)
+
+**Work item**: [dotCMS/core#37007 — Experiments Portlet — swap the List to the server-side contract](https://github.com/dotCMS/core/issues/37007)
+
+**Input**: User description: "Swap the Experiments portlet list from the client-side interim to the server-side contract from #36823" — taken from issue #37007.
+
+---
+
+## Scope Note *(read this first)*
+
+The Experiments portlet's site-wide list (#36989, merged via PR #37034) ships against
+`GET /api/v1/experiments` **as it exists today**: the endpoint returns every experiment on every
+site, and the list computes paging, sorting, text filtering, status filtering, status counts and
+site scoping **client-side**, plus a bulk page lookup that resolves each experiment's page to a
+URL and a site. That was an explicit, documented interim — the service method that fetches the
+full set is annotated as "the single swap point" for this change.
+
+This issue replaces that interim with the server-side contract defined by #36823: the list sends
+`page`, `per_page`, `orderby`, `direction`, `filter`, `status` and `siteId` to the endpoint, reads
+the total from the response body's `pagination.totalEntries`, and deletes the client-side
+utilities that faked all of this. The URL parameter contract was designed to be final from day
+one, so the swap changes **where** the parameters are applied — server instead of browser — not
+what they are called, how they default, or how they deep-link.
+
+The boundary is deliberate and is itself an acceptance criterion: only the data-access service
+and the store's load path change. **No component should need touching.** If one does, #36989's
+service boundary was drawn wrong — that is flagged in review, not patched around.
+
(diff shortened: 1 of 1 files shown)


erickgonzalez:
On-Hold until #36823 gets fixed, since probably needs to be re-written based on those changes.

freddyDOTCMS (approved):

Issue #37259, opened 2026-08-27
GET /api/v1/roles/layouts is reachable without authentication


hmoreras:
## Problem Statement

`GET /api/v1/roles/layouts` (`RoleResource#getAllLayouts`) has no authentication gate. Any unauthenticated caller receives `200 OK` and the full tool-group catalog of the instance.

It is the **only one of the 16 endpoints** in `RoleResource` without a `WebResource.InitBuilder` block. Every sibling opens with:

```java
new WebResource.InitBuilder(this.webResource)
    .requiredFrontendUser(false).rejectWhenNoUser(true)
    .requiredBackendUser(true).requiredPortlet("roles")
    .requestAndResponse(request, response).init();
```

`getAllLayouts` goes straight to `APILocator.getLayoutAPI().findAllLayouts()`.

**What is exposed:** every tool group's `id`, `name`, `description`, `tabOrder`, its full `portletIds` list, and the localized `portletTitles`. In effect, the complete map of the instance's administrative surface — including custom portlets, whose names are often customer- and business-specific.

**Impact:** information disclosure. No authentication, escalation, or data modification is possible through this endpoint — the sibling write endpoints (`POST`/`DELETE /v1/roles/layouts`) are properly gated. The value to an attacker is reconnaissance: knowing which features an instance runs, what they are called, and the real UUIDs those write endpoints consume.

**Not protected by anything else.** Three other layers were checked and none of them cover it:

| Layer | Result |
|---|---|
| Class-level annotation | `RoleResource` carries only `@Tag(name = "Roles")` and `@SuppressWarnings` |
| Jersey `RequestFilter` | Only wires session/context for `InitRequestRequired`. No `throw`, no `abortWith`, no 401 |
| `web.xml` filters | Nothing mapped under `/api/` except the GraphQL servlet |

The method also survives an anonymous call: it invokes `getLoggedInUser(request)`, which returns `null`, but `LanguageUtil.get(User, String)` falls back to the default company user (`LanguageUtil.java:209`) rather than throwing.

## Steps to Reproduce

Against any running instance, with no credentials and no cookies:

```bash
curl -s -o /dev/null -w '%{http_code}\n' http://localhost:8080/api/v1/roles/layouts
```

**Actual:** `200`, ~2.8 kB of JSON.
**Expected:** `401`, consistent with every other endpoint on this resource.

Control run on the same instance, same conditions — proving authentication is working and this endpoint simply lacks it:

```
[...]

PR #37260, opened 2026-08-28
feat(dot-roles): Tools tab in Angular, roles service consolidation, and edit-gate fixes

### LINKED ISSUE #36930: [Task] Roles and Tools: Angular implementation + backend wiring
## Goal

Ship the Angular Roles and Tools portlet end-to-end — UI, `portlet.xml` registration as an opt-in Beta portlet, and rollback story — in a single PR.

## Frontend scope

New Nx library: `core-web/libs/portlets/dot-roles/`

- `dot-roles-shell.component.ts`
- `dot-roles-page.component.ts`
- `lib.routes.ts` exported as `dotRolesRoutes`, imported at `roles` in `core-web/apps/dotcms-ui/src/app/app.routes.ts`
- SignalStore: `DotRolesStore` (role tree, selected role, active tab, per-tab pagination + search, drag state, in-flight flags)
- Data-access service: extend `core-web/libs/data-access/src/lib/dot-roles/dot-roles.service.ts` to cover `/v1/roles/*`, `/v1/users/filter` — replacing every `RoleAjax` DWR call needed by the Users tab and the roles tree that the Angular portlet renders directly. Permission endpoints and Tool Group endpoints (`/v1/permissions/*`, `/v1/roles/layouts`, `/v1/portlet/custom`, `/v1/toolgroups/*`, DWR `RoleAjax.getRolePermissions` / `saveRolePermission` / `loadRoleLayouts` / `saveRoleLayouts` / `addNewLayout` / etc.) are **not** consumed from the Angular portlet in this Epic — they stay behind the iframe wrappers described below
- Reuse models from `libs/dotcms-models`: `DotRole` (extend as needed)
- **PrimeNG only** for every UI component: `p-tree` (roles hierarchy with drag-and-drop), `p-tabView` (Users / Permissions / Tools), `p-table` (users grid), `p-dialog` (Add Role / Edit Role / confirm-destroy), `p-select` (Parent dropdown), `p-checkbox` (Can Grant flags), `p-textarea` (Description), `p-overlay-panel` or `p-autoComplete` (Grant to User popover)
- Users table follows the project's shared table styling conventions (see `dot-users` and `dot-tags` for reference); the design is a guide, not pixel-locked
- Tests: Jest + Spectator on shell / page / store / each tab, with `data-testid` selectors on every interactive element

### Screen structure (per new design)

**Layout**: two-column shell — left `Roles` panel, right `Role detail` panel. No page-level top bar beyond the shared dotCMS header (breadcrumbs `Home > Settings > Roles`, site selector, notifications, user avatar).

**Left panel — Roles**
- Header: `ROLES` label + `New` button (opens Add Role dialog)
- `Filter roles` search input (debounced client-side filter of the loaded tree)
- Hierarchical `p-tree` list:
  - Top-level roles use a **folder icon** with expand/collapse chevron
  - Child roles use a **shield icon**
  - Per-row badge: user count with person icon (e.g., `👤 2`)
  - Hover reveals inline `+` action to add a child role under that parent (opens Add Role dialog with `Parent` prefilled)
- **Drag-and-drop reparenting**: dragging a role node onto a different parent updates `Role.parent` via a v1 endpoint (endpoint identified during the spike; if the current REST surface does not expose a reparent operation, file it as a backend hardening `[Task]` linked to this issue)
- Selected node highlighted; selection drives the right-hand detail area

**Right panel — Role detail**
- Header row: role icon (folder for parents, shield for children) + role name + subtitle `N users · M tools granted` + `Edit Role` button top-right
- Three tabs (`p-tabView`):

  **1. Users** — full Angular implementation
  - `Grant to User` button top-right opens an overlay panel with:
    - `Search users` input (debounced, hits `/v1/users/filter` or the equivalent identified in the spike)
    - Scrollable list of users (avatar + name + email); click a row to grant this role to that user
  - Below: `p-table` of current members
    - Columns: **NAME** (avatar + name), **EMAIL**, **GRANTED FROM** (chip showing the role that granted access — direct assignment shows the current role's name; inherited shows the parent role's name)
    - Row hover reveals `Remove` action; multi-select supports bulk remove
    - Empty state: dashed card with helper text `This role has no users yet` and copy `Grant this role to a user using the selector above.`

  **2. Permissions** — **iframe interim**
  - Angular tab renders an `<iframe>` whose `src` points at a new lightweight JSP wrapper (see Backend wiring below) that hosts the existing Dojo/DWR permissions UI unchanged
  - No Angular re-implementation of the accordion permission matrix in this Epic; a follow-up `[Task]` is filed for the eventual Angular replacement so the iframe can be swapped by dropping in the new Angular component without touching the surrounding shell
  - Existing permission endpoints (`RoleAjax.getRolePermissions`, `saveRolePermission`, `/v1/permissions/*` where already used) stay behind the iframe
  - Iframe height / resize behavior: fills the tab body with a `postMessage` or auto-resize helper so users don't see nested scrollbars

  **3. Tools** — **iframe interim**
  - Angular tab renders an `<iframe>` whose `src` points at a new lightweight JSP wrapper hosting the existing Dojo/DWR Tool Groups management UI unchanged (available / assigned layouts, `New Layout` dialog with `dot-material-icon-picker`, save role-layouts binding)
  - No Angular re-implementation of Tool Groups in this Epic because the design is still being defined; a follow-up `[Task]` is filed for the eventual Angular replacement so the iframe can be swapped by dropping in the new component
  - Existing Tool Group endpoints (`RoleAjax.loadRoleLayouts` / `saveRoleLayouts` / `addNewLayout` / etc., `/v1/roles/layouts`, `/v1/portlet/custom`) stay behind the iframe

**Add Role dialog** (triggered by `New` in the roles panel or the inline `+` on a row)
- Title: `Add Role`
- Fields:
  - `Role *` (required, red border on validation fail)
  - `Key`
  - `Parent` (`p-select`, defaults to `None (top level)`; when opened from an inline `+`, prefilled with that parent role)
  - `Can Grant` (three `p-checkbox`: Users, Permissions, Tools — all checked by default per the design)
  - `Description` (`p-textarea`)
- Actions: `Cancel` / `Save` (Save disabled until required fields valid)
- `closable: true`, `closeOnEscape: true`

**Edit Role dialog** (triggered by `Edit Role` in the detail header)
- Same fields as Add Role, prefilled with the selected role's values
- Additional `Delete Role` destructive action bottom-left (red outline). Click triggers a confirm-destroy dialog
- `closable: true`, `closeOnEscape: true`

**Confirm-destroy dialog**
- Confirm-remove role, per dotCMS dialog standards (`closable: true`, `closeOnEscape: true`)

## Backend wiring

`dotCMS/src/main/webapp/WEB-INF/portlet.xml`:

- Add a new `roles-beta` portlet entry: `<portlet-class>com.dotcms.spring.portlet.PortletController</portlet-class>` + `<portlet-url>/roles</portlet-url>` + display name `Roles and Tools (Beta)`
- The existing Dojo `roles` portlet entry stays untouched so it remains reachable via the "Add Portlet" picker
- Not added to any default layout or `UpgradeTask` — admins opt in via "Add Portlet" in Tools
- Angular route wired at `roles` in `core-web/apps/dotcms-ui/src/app/app.routes.ts` loading `@dotcms/portlets/dot-roles/portlet`
- Reuse the existing `PortletID.ROLES` enum entry and the existing `roles` i18n title key; add a `roles-beta` i18n title key for the Beta display name
- License + role gating preserved (`requiredPortlet("roles")` on v1 endpoints, admin-only surfaces enforced)

**Permissions iframe wrapper JSP**

- Add a new minimal JSP under `dotCMS/src/main/webapp/html/portlet/ext/roleadmin/` (naming to be decided in the spike, e.g., `view_role_permissions_wrapper.jsp`) that includes the existing `view_role_permissions_inc.jsp` + `view_role_permissions_js_inc.jsp` output and nothing else
- The wrapper accepts the `roleId` via query string, initializes the existing Dojo/DWR components, and communicates readiness / size to the parent Angular tab via `postMessage`
- Kept intentionally thin so the eventual Angular replacement (tracked as a separate `[Task]`) only requires swapping the iframe `src` for a new Angular component in `dot-roles-page.component.ts`

**Tools iframe wrapper JSP**

- Add a second minimal JSP under `dotCMS/src/main/webapp/html/portlet/ext/roleadmin/` (e.g., `view_role_tools_wrapper.jsp`) that surfaces just the Tool Groups management section of the existing `view_roles_js_inc.jsp` (available / assigned grids, `New Layout` dialog, save role-layouts)
- Same `postMessage` contract as the permissions wrapper, same `roleId` query-string input, same thin-wrapper philosophy so the future Angular replacement is a drop-in

**Untouched files**

- `view_roles.jsp` remains untouched — the wrappers are new files dedicated to the iframe use case, so the current Dojo portlet at `/#/c/roles` keeps working for admins who don't opt into the Beta

**Reparent endpoint**

- If the drag-to-reparent flow requires a new v1 endpoint (spike output will confirm), open a backend `[Task]` linked to this issue and unblock it before merge

## Inputs

- Spike output from the consolidated audit issue (parity matrix + API audit + UI mapping + hardening backlog + reparent endpoint decision + iframe wrapper design for Permissions and Tools)
- Portlet conventions: `core-web/libs/portlets/CLAUDE.md`
- Structural reference: `core-web/libs/portlets/dot-users/` (same Beta pattern, same iframe-for-undesigned-surfaces approach) and `core-web/libs/portlets/dot-velocity-playground/`
- New design: https://claude.ai/design/p/20548219-a454-4b14-becf-5dd4dc0b84a1?file=Roles.dc.html&via=share (internal — auth required)

## Definition of Done

Every acceptance criterion in epic #36909 passes manually against the local build:

- Beta portlet appears in the "Add Portlet" picker as `Roles and Tools (Beta)`
- Adding the Beta portlet to a layout serves the Angular shell at `/#/c/roles`
- Roles panel renders the full hierarchy with folder icons for parents and shield icons for children, expand/collapse, and user-count badges
- `Filter roles` narrows the visible tree client-side
- `New` opens the Add Role dialog; save creates the role and refreshes the tree
- Inline `+` on a parent row opens Add Role dialog with `Parent` prefilled
- Drag-and-drop of a role onto a different parent persists the reparent via v1 endpoint
- `Edit Role` in the detail header opens the Edit dialog; save updates the role; `Delete Role` opens a confirm-destroy and removes the role
- Users tab: `Grant to User` popover searches users and grants the role on selection; the members table renders NAME / EMAIL / GRANTED FROM (with inheritance chip); remove and bulk-remove work; empty state matches the design copy
- Permissions tab: iframe loads the new JSP wrapper and renders the existing permissions UI functional (matrix, cascade toggle, apply changes, cascade-progress) — behavior parity with the current Dojo portlet
- Tools tab: iframe loads the new JSP wrapper and renders the existing Tool Groups management UI functional (available / assigned layouts, `New Layout` dialog, save role-layouts) — behavior parity with the current Dojo portlet
- License + role gating preserved on every write path
- Dojo `roles` portlet entry remains registered so admins can revert by removing the Beta portlet from their layout
- Follow-up `[Task]` filed for the Angular replacement of the Permissions tab, linked to this issue and to epic #36909
- Follow-up `[Task]` filed for the Angular replacement of the Tools tab (once design is defined), linked to this issue and to epic #36909

hmoreras:
### Proposed Changes

https://github.com/user-attachments/assets/5271f754-c40a-4fc2-b07f-e7692416fd66

Follow-up work on the Roles (Beta) Angular portlet.

**Consume the backends that landed** — `GET /v1/roles/{roleId}/users` (#37070) replaces the previous pair of member-loading calls. The old id-based `/rolehierarchyanduserroles` fallback returned `Role` objects rather than users, which is why the Users tab rendered an empty EMAIL column; that column now populates. `childCount` and `userCount` (#37071) drive leaf-vs-chevron detection on first paint — no more chevrons that expand into nothing — and unblock the per-row user-count badge the design called for.

Worth flagging: the code's comments predicted `/roles/{id}/users` would resolve inheritance server-side and collapse the whole flow into one call. It shipped as a *direct grants only* resource by design, so the ancestor-chain fan-out stays and is now documented as permanent rather than as a stopgap. For the same reason the members table keeps client-side paging — a server page of one ancestor is not a page of the merged union, so `[lazy]="true"` would be incorrect, not pending.

**Migrate the Tools tab to Angular** — replaces the Dojo iframe. Scope is deliberately narrower than the JSP it replaces: it lists tool groups and toggles which ones the role gets. Creating, editing and deleting tool groups is not here — those have no v1 endpoints (only `RoleAjax` DWR) and belong to a Tools portlet of their own. Inherited grants render checked but locked; echoing one back into the full-replace `POST` would silently promote it to a direct grant. The iframe component and its two JSPs are deleted rather than left behind.

**Consolidate the roles surface into `DotRolesService`** — three libs had grown their own implementation of the same `/api/v1/roles/**` surface, and three TypeScript models of the same backend `RoleView`. Reads, writes and models now live in one place; method names drop the redundant `role` (`createRole` → `create`). `get` and `search` keep their legacy names — the workflow assign components inject this service and those paths are untouched. The full-hierarchy walk did **not** move into the service: it is one way of composing two endpoints for one UI, so it lives as a pure, HTTP-free function in `dot-users`, and it now prunes on `childCount` — most roles below the first level are leaves, which removes the large majority of the request burst the Roles tab produced on open.

**Fix edit gates that were stricter than the backend** — the Tools tab greyed itself out for CMS Administrator while the legacy portlet allows the edit. `RoleHelper` applies a different contract per operation: update/delete checks `isSystem() || isLocked()`, user grants check `isEditUsers()` alone, and saving layouts places no restriction on the target role at all.

### Review fixes

A multi-agent review and hands-on QA surfaced defects that are fixed in this PR:

- **Two cross-role races.** `saveToolGroups` and the user grant/remove flows captured the role id before their `await` and reconciled with it afterwards, so a role switch mid-flight repainted the new role's tab with the old role's data. For tool groups this was worse than cosmetic: the tab derives its next POST payload from `toolGroups`, so one role's grants could be written onto another. Guarded at both ends, with regression tests verified by neutering each guard and confirming they go red.
- **"Could not verify" was being rendered as "not granted".** A per-ancestor `catchError` returned an empty set on failure, so a transient error showed an inherited tool group unchecked — and an admin trusting the grid would create a redundant direct grant. Failures now surface instead of being absorbed.
- **`childCount` regressions.** Local tree mutations left it stale, so a parent that just gained its first child rendered as a leaf and the new role was unreachable. And `POST /v1/roles` answers with `Role.toMap()`, which carries no `childCount`, so a freshly created role drew as a folder.
- **The tree user-count badge never refreshed** after a grant or revoke. It now syncs from the direct-grant count.
- **`dot-users-list.store.spec.ts`** still mocked a method deleted in this PR and never mocked `DotRolesService`, so the roles column ran against the real root service and was effectively untested.

### Add / Edit role dialogs

Required markers use the `dotFieldRequired` directive (which is what makes them red) instead of hand-written text, and `dot-field-validation-message` is wired in — there was no validation feedback at all. The dialog title moves from an h2 in the body into the dialog header.

Parent selection becomes a `p-treeSelect`: indentation alone gets ambiguous past two levels. It starts collapsed, hydrates a branch on expand (the backend sends two levels per request), and its filter runs the same deep search the roles tree uses, so a role in an unloaded branch is findable.

Two PrimeNG behaviours needed working around, both from one cause — the component keeps state inside the node objects it is given, and these options come from a `computed` that hands it new ones. Expansion is mutated onto `node.expanded`, so branches snapped shut the moment their children loaded. And `Tree.getRootNode()` returns its cached `filteredNodes` once the client filter has run, ignoring `value`, so server results never rendered. Both are handled explicitly, with the reasoning in the code — the second reaches one internal method, which is the most upgrade-fragile line in this PR.

### Design review

Empty states now use the shared `DotEmptyContainerComponent`; the hand-rolled dashed card they replaced exists nowhere else in the product. `shield` → `shield_person`. The `+` on a tree row is primary with a pointer cursor. Untranslated portlet titles no longer print as raw i18n keys.

The user-count badge takes a real grey. `text-color-secondary` is a PrimeFlex-era class — PrimeFlex is not installed and `tailwindcss-primeui` does not provide it either, so it compiles to nothing and elements inherit the label's near-black. Verified against the built CSS rather than assumed. **104 more occurrences survive across 35 files** and need their own pass.

### Checklist
- [x] Tests
- [x] Translations

### Additional Info

**Verification** — 36 projects linted, 35 tested, `nx format:check` clean, and the `dotcms-ui` production build green. The `dot-roles` suite is at 136 tests.

**Security note.** While tracing the tool-group endpoints, `GET /api/v1/roles/layouts` turned out to have no authentication gate — the only one of 16 endpoints in `RoleResource` without a `WebResource.InitBuilder` block. Confirmed at runtime: it returns `200` with the full tool-group catalog to an unauthenticated caller, while three sibling endpoints return `401` under identical conditions. It predates this work (introduced 2023-05-26, shipped since v23.06) and is filed separately as **#37259**. This PR does not change that behaviour, but it does put the endpoint on a more exercised path.

**Standards note.** `ANGULAR_STANDARDS.md` asks for Signal Forms on new forms, but the standard validation components are Reactive-Forms-bound — `dotFieldRequired` injects `FormGroupDirective` and `dot-field-validation-message` takes an `AbstractControl`. These dialogs stay on Reactive Forms; adopting Signal Forms would mean dropping both.

**Follow-ups not included here**
- `loadRootRoles(true)` → `false` is now possible thanks to `childCount`, and would lighten the initial payload considerably on installs with many roles. It changes lazy-load behaviour, so it deserves measuring on its own.
- `onNodeExpand` marks a node fetched before the load resolves, so a failed lazy-load never retries.
- Creating a role under a parent that is not currently loaded splices into a node that is not there — the role is created but does not appear until a reload.
- The `text-color-secondary` sweep described above.

This PR fixes: #36930

This PR fixes: #36930


AP2300 (commented):
Code review — 9 findings, ranked most-severe first. Correctness bugs first, cleanup/altitude at the end. The recurring theme is that the cross-role-race guard the PR added for `saveToolGroups` isn't lifted to a shared helper and reapplied uniformly across mutations.

Findings referenced against the current head (`d2e386b`).

---

### 1. `updateRole` / `createRole` skip the cross-role guard the PR added for saves

[`dot-roles.store.ts:837-840`](https://github.com/dotCMS/core/blob/d2e386b3da6569abe2b09b70527a0de9a69ab18a/core-web/libs/portlets/dot-roles/src/lib/dot-roles-page/store/dot-roles.store.ts#L837-L840), [`:881-891`](https://github.com/dotCMS/core/blob/d2e386b3da6569abe2b09b70527a0de9a69ab18a/core-web/libs/portlets/dot-roles/src/lib/dot-roles-page/store/dot-roles.store.ts#L881-L891)

Both write `selectedRole` / `selectedRoleId` unconditionally after the await. `saveToolGroups` and the user grant/remove flows added `selectedRoleId() === roleId` guards for exactly this reason.

*Scenario:* admin edits role A and saves, then clicks role B before the PUT resolves. `selectRole` clears state and loads B. PUT for A resolves after: `selectedRole` is overwritten with A while B is highlighted; the header, `canModifyRole` computeds, and the Users/Tools tabs all now describe A. A subsequent save on that panel PUTs against role A.

Consider extracting a `withSelectedRoleScope((roleId) => …)` helper (or the `switchMap`-keyed-by-roleId pattern in a shared feature slice) so every mutation acquires the guard by construction.

---

### 2. `collectAncestorChain` fabricates a parent-less stub → silent under-reporting of inherited grants

[`dot-roles.tree-utils.ts:163-179`](https://github.com/dotCMS/core/blob/d2e386b3da6569abe2b09b70527a0de9a69ab18a/core-web/libs/portlets/dot-roles/src/lib/dot-roles-page/store/dot-roles.tree-utils.ts#L163-L179)

When the picked role isn't in the merged tree, `findRoleInTree(...) ?? { id, name: id }` produces a stub with no `parent`; the loop exits after zero iterations and the fan-out shrinks to a single direct-grants request.

*Scenario:* a search-selected role whose branch hasn't been merged into `roles` / `searchResults`, or a race with a `roles: []` reset (post-`createRole` root reload). `loadMembers` / `loadToolGroups` fire only `GET /v1/roles/{targetId}/{users,tools}`, silently omitting every ancestor's grants. Users tab renders an incomplete member roster and the Tools tab misses inherited grants — admin trusts the UI and re-creates a redundant direct grant. No error surfaced.

If the caller doesn't know the parent, this should fail loudly (or fall back to a `getById`) rather than short-circuit.

---

### 3. `createRole` seeds `selectedRole` with the raw POST payload — edit-gate flags are absent

[`dot-roles.store.ts:797-841`](https://github.com/dotCMS/core/blob/d2e386b3da6569abe2b09b70527a0de9a69ab18a/core-web/libs/portlets/dot-roles/src/lib/dot-roles-page/store/dot-roles.store.ts#L797-L841)

The comment on L799 says `POST /v1/roles` returns `Role.toMap()`, not a `RoleView`; the comment on L833 then claims the response is "already a hydrated `RoleView`" and seeds `selectedRole` directly. The type is asserted with a cast + `childCount: 0` spread — TS doesn't catch it. `system`, `locked`, `editUsers`, `editPermissions`, `editLayouts` are absent, and downstream computeds fall back to permissive defaults (`?? true`) until the next page reload.

*Scenario:* admin creates a role whose contract would forbid user grants (`editUsers === false` per template). The just-created payload lacks the flag; the Users tab renders enabled; adds are either 403'd (bad UX) or, if the BE is lenient on fresh roles, silently promoted.

Two consistent options: (a) drop the "seed from POST" optimization and fire a follow-up `GET`; (b) have `createRole` echo a hydrated `RoleView` on the BE (worth a Java-side follow-up in either case, since PUT already does).

---

### 4. `removeUsersFromRole` runs the optimistic `patchState` before the cross-role guard

[`dot-role-users-tab.component.ts` grant/remove path → `dot-roles.store.ts` around L1006]

The guard only gates the follow-up `loadMembers`; the optimistic filter has already written into `members`, which by then may be a different role's slice.

*Scenario:* admin bulk-removes u1,u2 from role A, then clicks role B (inherits u1) before the DELETE resolves. `selectRole` flips the pane to B; the optimistic filter strips u1 from `members` — now B's slice — and the guarded reload skips because `selectedRoleId !== A`. B's member table wrongly shows u1 missing until the admin re-selects B.

Move the optimistic write behind the same guard, or key it against a captured `roleId` snapshot rather than the current `members`.

---

### 5. `$searching` never resets on stale-token drop or the <3-char short-circuit

[`dot-roles-add.component.ts:183-216`](https://github.com/dotCMS/core/blob/d2e386b3da6569abe2b09b70527a0de9a69ab18a/core-web/libs/portlets/dot-roles/src/lib/dot-roles-add/dot-roles-add.component.ts#L183-L216), [`dot-roles-edit.component.ts` `#runSearch`]

`$searching.set(true)` is set right before `await`; on `token !== #searchToken` the function returns without `$searching.set(false)`. Same shape in the edit dialog.

*Scenario:* admin types "abc" then a fourth char before the promise resolves; `#searchToken` bumps, old promise resolves into the guard-return branch, `$searching` stays `true` for the rest of the dialog session — the parent-picker looks permanently busy.

`try/finally { $searching.set(false); }` around the await would fix both.

---

### 6. Partial fan-out failure hides successful ancestor rows behind a generic ERROR

[`dot-roles.store.ts` around L397-L401 in `loadMembers` / `loadToolGroups` result mapping]

A single ancestor request failing flips `membersStatus` (and the tool-groups equivalent) to `'ERROR'` even after the other ancestors' rows are already merged into state — the template hides the partial table behind a generic error banner while `httpErrorManager` fires a toast per failed request.

*Scenario:* one of five inherited role fetches returns a transient 5xx. Four ancestors' users are in `members`; template renders "Could not load" instead of them. Admin retries by re-selecting, fan-out repeats and can hit the same intermittent failure.

Introduce a `'PARTIAL'` status (or a `verified: boolean` on the slice) so the table renders alongside an inline "some ancestors couldn't be verified" warning.

---

### 7. `updateRole` reparent branch mistakes a missing `parent` in the response for detach-to-root

[`dot-roles.store.ts:876-891`](https://github.com/dotCMS/core/blob/d2e386b3da6569abe2b09b70527a0de9a69ab18a/core-web/libs/portlets/dot-roles/src/lib/dot-roles-page/store/dot-roles.store.ts#L876-L891)

`nextParentId = updated.parent && updated.parent !== updated.id ? updated.parent : null`. If the PUT response ever omits `parent` (defensive undefined, partial view, future serializer refactor), `previousParentId === nextParentId` is false and the code splices the node out of its parent and appends it as a new root. `childCount` on the true parent isn't decremented, so the tree also shows the wrong count until reload.

Contract-wise the BE encodes root as `parent === role.id`; `parent === undefined` isn't in that contract. If it arrives, fail closed (keep the node where it was) rather than silently reparenting.

---

### 8. `_filter` private-API workaround + treeSelect state duplicated across Add and Edit

[`dot-roles-add.component.ts:194,215`](https://github.com/dotCMS/core/blob/d2e386b3da6569abe2b09b70527a0de9a69ab18a/core-web/libs/portlets/dot-roles/src/lib/dot-roles-add/dot-roles-add.component.ts#L194-L215), [`dot-roles-edit.component.ts` equivalent]

The PR description calls out this line as "the most upgrade-fragile line in this PR" — and it exists twice. The surrounding `#expandedKeys` / `onNodeExpand` / `onNodeCollapse` / `#searchToken` / `#runSearch` / `#toTreeNodes` triad is ~90% duplicated between the two dialogs.

*Cost:* the next PrimeNG bump that renames `_filter` or restructures `treeViewChild` breaks both dialogs; fixes will drift over time as one gets updates the other doesn't.

Extract a shared `dot-role-parent-picker` (or a directive that owns the workaround) so the fragile access lives behind one seam. The two dialogs then differ only in their initial value, submit call, and the descendant-exclude filter.

---

### 9. Tool-group toggle fires a full reconcile per click — debounce coalesces only the toast

[`dot-role-tools-tab.component.ts` `onToggle` → `dot-roles.store.ts:693` `saveToolGroups`], [`:218` `toolGroupsSaved$.pipe(debounceTime(...))`](https://github.com/dotCMS/core/blob/d2e386b3da6569abe2b09b70527a0de9a69ab18a/core-web/libs/portlets/dot-roles/src/lib/dot-roles-page/store/dot-roles.store.ts#L218)

The 1.5s `toolGroupsSaved$` debounce coalesces only the toast; each save still triggers a full catalog GET + N ancestor tool-group GETs. A burst of 5 checkbox clicks on a role with 3 ancestors is 5 POSTs + 5 catalog GETs + 15 ancestor GETs.

Coalesce the reconcile into the same debounce (queue the id-set of dirty groups, single reconcile after the burst). Under network jitter, out-of-order reconciles can also restamp state with a stale union — the guarded write only gates the last one to resolve, not the earlier stale ones.

---

*Out-of-scope for this PR but worth capturing:* `dot-users-list.store.ts` bulk-delete path has no aggregate toast when every deletion fails (only per-request errors). Behavior pre-dates this PR; noting here so it isn't lost.

Nice work overall — the consolidation into `DotRolesService`, the Tools-tab Angular migration, and the tree-node child-count / reparent handling read cleanly. Findings 1–4 are the ones I'd want addressed before merge; the rest can be follow-ups.

hmoreras:
@AP2300 the mention options was fixed; except 7 and 9 since are in some way false positive

AP2300 (dismissed):

AP2300 (approved):

PR #34236, opened 2026-01-07
feat(task): upgrade to Postgres 18

wezell:
ref: #34230

### Proposed Changes
* This updates all examples and compose files to use Postgres 18 (or the pgvector equiv)

### Checklist
- [x] Test Harness Passes

This PR fixes: #34230

PR #32980, opened 2025-08-12
feat(monitoring): Create LGTM Observability Docker Compose Stack for (#32979)

spbolton:
## Description

Complete LGTM observability stack implementation for dotCMS development and testing. Integrates micrometer metrics, OpenTelemetry instrumentation, and Pyroscope profiling with Grafana LGTM stack (Loki/Grafana/Tempo/Prometheus). 

## Changes
- [ ] Docker compose configuration with dotCMS, LGTM stack, PostgreSQL, and OpenSearch
- [ ] Custom Prometheus configuration extending LGTM defaults to scrape dotCMS metrics  
- [ ] 5 comprehensive Grafana dashboards for database, JVM, cache, HTTP, and logs monitoring
- [ ] OpenTelemetry and Pyroscope agent integration via init container
- [ ] Complete documentation covering architecture, configuration, and troubleshooting

## Testing  
- [ ] All services start successfully with proper health checks
- [ ] Prometheus scrapes dotCMS micrometer metrics from /dotmgt/metrics endpoint
- [ ] Custom dashboards display real-time dotCMS observability data
- [ ] OpenTelemetry traces, metrics, and logs flow to LGTM stack
- [ ] Comprehensive validation checklist and troubleshooting guide verified

## Changes

- [ ] [List the main changes made]

## Testing

- [ ] [Describe testing approach]

Closes #32979

**Issue:** Create LGTM Observability Docker Compose Stack for

This PR fixes: #32979

PR #29915, opened 2024-09-06
chore(documentation): Update Docker Compose examples to use pgvector:pg16 (#29818)

dcolina:
### Proposed Changes
* This pull request updates the Docker image version used for the database to pgvector/pgvector:pg16. The following Docker Compose manifests have been adjusted accordingly, with some improvements:

- single-node
- single-node-debug-mode
- single-node-demo-site
- with-opensearch-dashboard
- push-publish

*Pending Updates*
The following examples have not been updated in this pull request, as they are currently not functioning correctly. A separate task has been created to address these issues during the next gardening day. Please refer to Issue #29909 for more details:

- cluster-mode
- with-redis
- with-redis-session

### Additional Info
Related to #29818 (Update Docker Compose examples to use pgvector:pg16).

This PR fixes: #29818

PR #37263, opened 2026-08-28
docs(create-app): issue-resolution spec for local Docker start failure and the resulting permanent UVE 403 (#37262)

### LINKED ISSUE #37262: create-app: local Docker run never starts dotCMS, then a transient UVE 403 aborts the CLI and discards the project
## Problem Statement

Following the [Headless dotCMS Quick Start](https://www.dotcms.com/blog/headless-dotcms-quick-start-introducing-dotcms-create-app-cli) blog post, `npx @dotcms/create-app` (v1.2.5) fails end-to-end in local Docker mode. Two independent bugs, plus a resilience problem that turns the second one into total data loss.

Observed sequence (interactive: `No - Spin up dotCMS locally with Docker`, `Next.js`):

1. CLI reports "dotCMS containers started successfully" — but the `dotcms` container had already exited.
2. The health loop burns 12 attempts (`ECONNRESET` → `ECONNREFUSED` → timeouts) against a container that isn't running.
3. The user manually clicks ▶ on `dotcms-1` in Docker Desktop. By then Postgres/OpenSearch are up, so it boots.
4. Health goes green, token issued, default site resolved.
5. `POST /api/v1/apps/dotema-config-v2/{siteId}` returns **403** → `process.exit(1)`.

Result: no project, no `.env`, and the working token and site ID are never printed. Everything except one non-essential config call had succeeded.

```
✔ dotCMS containers started successfully.        ← the container had already exited
⏳ dotCMS not ready (attempt 1/60) - ECONNRESET - Retrying in 5s...
   … 11 more attempts …
✔ dotCMS is running locally at http://localhost:8082
✔ Generated API authentication token             ← discarded on exit
✔ Retrieved default site (8a7d5e23-da1e-…)       ← discarded on exit
failed to setup UVE config: status=403, code=ERR_BAD_REQUEST
✖ Failed to setup UVE configuration in Dotcms.
```

---

### Root cause 1 — the compose file, not the CLI

The CLI downloads `docker/docker-compose-examples/single-node-demo-site/docker-compose.yml` from `main` at runtime. It is the least hardened compose example in that directory:

- `dotcms` and `opensearch` have **no `restart:` policy** (only `db` has one), so when dotCMS exits it stays exited.
- `depends_on: [db, opensearch]` has **no `condition: service_healthy`** — dotCMS boots against a Postgres that isn't accepting connections yet. The `db` service already *defines* a healthcheck that nothing consumes.
- `opensearch` has no healthcheck; `dotcms` has neither a healthcheck nor its management port (8090) published.

Six sibling examples already use `condition: service_healthy`. `lgtm-observability/docker-compose.yml` is the model to copy.

Because the CLI fetches this file at runtime, **fixing it ships to every already-installed CLI without a release.**

### Root cause 2 — the 403 is a startup race, not permissions

Running the CLI's exact three-call sequence against a fully-started dotCMS returns **200** for all three calls (`POST /api/v1/authentication/api-token` → `GET /api/v1/site/defaultSite` → the apps API for the same site ID). That rules out:

- **License gating** — `AppsAPIImpl` throws `InvalidLicenseException` without an enterprise license, but `LicenseUtil.getLevel()` has returned `PLATFORM` unconditionally since #31261 (Feb 2025). Dead code on any current image.
- **Apps portlet / backend-user access** — `GET /api/v1/apps` and `GET /api/v1/apps/dotema-config-v2/{siteId}` both return 200 for a token minted the way the CLI mints one.
- **Wrong site ID** — `8a7d5e23-…` is the normal default site.

What remains: `AppsAPIImpl.userDoesNotHaveAccess()` (`AppsAPIImpl.java:104`) calls `user.isAdmin()`, which is wrapped in `Try.of(…).getOrElse(false)` (`com/liferay/portal/model/User.java:321`). **Any** exception during the role lookup silently reports "not an admin", which becomes a `DotSecurityException` and maps to 403 via `ExceptionUtil.SECURITY_EXCEPTIONS`.

Timing supports this: the CLI's readiness probe is `/api/v1/appconfiguration`, which answers as soon as the web layer is up. It went green ~60s after container start — far too fast for a demo-starter import to have finished. The CLI then immediately wrote app secrets to an instance still settling roles, permissions and caches.

**The readiness signal is wrong.** dotCMS ships a real readiness probe at `/dotmgt/readyz` (verified: responds `ready`, unauthenticated, no IP ACL) — but only on port **8090**, which the demo compose doesn't publish (`/dotmgt/readyz` on 8080 is a 404). Even `/readyz` isn't a full guarantee, though; its registered checks cover CDI, memory, threads and the servlet container — not "starter import finished". For a CLI the reliable rule is: **readiness means the call you're about to make succeeds.**

### Root cause 3 — the CLI discards recoverable state (the expensive one)

Worth fixing regardless of the above:

- **UVE setup runs before scaffolding.** `src/index.ts:370` exits; the clone + `npm install` at `:377` never run. The user gets an empty directory, not a project with one missing setting.
- **The token and site ID are never printed**, even though both were successfully obtained.
- **No retry on the UVE call** — authentication retries 3×, this doesn't retry at all.
- **Re-running is blocked by the CLI's own success.** `checkPortsAvailability()` (`src/utils/index.ts:479`) hard-fails on 8082/8443/9200/9600 — which the now-running dotCMS holds. Retrying means tearing down a healthy instance and waiting out another cold start.
- **And retrying would delete the escape hatch.** `prepareDirectory()` (`src/asks.ts:180`) offers to empty the target directory — which now contains the `docker-compose.yml` needed to run `docker compose down`.

### Additional bugs found while reading

- **`npm install` failures are reported as success.** `installDependenciesForProject()` returns a `Result`, but `src/index.ts:597` tests `if (!result)`. `Err()` returns `{ ok: false, val }` — a truthy object. The failure branch is unreachable.
- **Orphaned compose file.** `moveDockerComposeOneLevelUp()` runs at `src/index.ts:376`; if scaffolding fails it calls `process.exit(1)` internally, so `moveDockerComposeBack()` at `:378` never runs, stranding `docker-compose.yml` in the parent directory. Needs `try/finally`.
- **Multi-minute silence.** `execa('docker', ['compose','up','-d'])` swallows image-pull progress — a frozen spinner for the length of a ~1.5GB pull on a cold machine.
- **Unguarded download.** `downloadFile()` uses raw `https.get` with no timeout, no retry, no redirect handling, against an unpinned `main` URL.
- **Status mismatch.** `fetchWithRetry` accepts any 2xx; `isDotcmsRunning` then demands exactly 200.
- **Interleaved output.** `fetchWithRetry` calls `console.log` while an `ora` spinner is active — the cause of the mangled retry block above.
- **No tests.** Not one spec file in the package (E2E suite tracked in #35096).

## Steps to Reproduce

1. On a machine with Docker Desktop and no dotCMS containers running, run `npx @dotcms/create-app my-app`.
2. Answer: directory `.`, `No - Spin up dotCMS locally with Docker`, framework `Next.js`.
3. Observe the `dotcms` container start and immediately exit while `db`/`opensearch` are still initializing; the CLI still prints "containers started successfully" and enters the retry loop.
4. Manually start the `dotcms` container from Docker Desktop.
5. Once the health check passes, observe `failed to setup UVE config: status=403` and a non-zero exit with no project scaffolded.
6. Attempt to re-run the same command — it now aborts at "Required ports are already in use", because the dotCMS started in step 4 holds 8082/8443/9200/9600.

## Acceptance Criteria

**Compose (P0 — ships without a CLI release)**

- [ ] `single-node-demo-site/docker-compose.yml`: add an `opensearch` healthcheck and `restart: unless-stopped`
- [ ] Change `dotcms` `depends_on` to `condition: service_healthy` for both `db` and `opensearch`
- [ ] Add a `dotcms` healthcheck against `http://127.0.0.1:8090/dotmgt/livez` with `start_period: 180s`, plus `restart: unless-stopped`
- [ ] Publish port `8090:8090`

**CLI (P0)**

- [ ] The CLI never exits without printing recoverable state (host, token, site ID) — and writes the `.env` it already has all the values for
- [ ] UVE setup failure is non-fatal: warn, print manual setup steps, and continue to scaffolding
- [ ] Gate the UVE write on a read: poll `GET` the UVE app endpoint until 200, then `POST` with retry on 401/403/5xx

**CLI (P1)**

- [ ] Port check probes before failing — if dotCMS already answers on 8082, offer to reuse it instead of exiting
- [ ] Use `docker compose up -d --wait` and stream pull progress so the wait is visible
- [ ] Fix the truthy-`Result` check at `src/index.ts:597`; wrap the compose move in `try/finally`
- [ ] Switch readiness to `/dotmgt/readyz` on 8090 once the compose publishes it, with `/api/v1/appconfiguration` as fallback

**Backend / follow-up (P2)**

- [ ] `user.isAdmin()` should not swallow exceptions into a false "not admin" — a transient lookup failure shouldn't read as a permission denial
- [ ] Pin the dotCMS image tag alongside `CUSTOM_STARTER_URL` (`latest` + hardcoded `starter-20260630` will drift apart)
- [ ] Land the E2E suite from #35096, including a fault-injection case that kills dotCMS mid-run

## dotCMS Version

`@dotcms/create-app` 1.2.5 · `dotcms/dotcms:latest` · compose from `dotCMS/core@main` · macOS + Docker Desktop

## Severity

## Links

- Slack thread: https://dotcms.slack.com/archives/C0AGPFATCUF/p1787168520337159
- Related: #35096 (E2E test suite for `@dotcms/create-app`)
- Blog post users are following: https://www.dotcms.com/blog/headless-dotcms-quick-start-introducing-dotcms-create-app-cli

fmontes:
### Proposed Changes

Spec-only PR (**PR 1 of 2**) for [#37262](https://github.com/dotCMS/core/issues/37262), per the repo's spec-kit flow. No code changes — this PR is the approval gate for the approach. Implementation lands in the stacked PR on top of this branch.

* Adds `specs/37262-create-app-docker-uve/spec.md` — an issue-resolution spec for the `@dotcms/create-app` local-Docker failure
* Points `.specify/feature.json` at the new feature

**What's broken** (three compounding defects):

1. **dotCMS never starts.** `single-node-demo-site/docker-compose.yml` lets `dotcms` boot before Postgres accepts connections and gives it no restart policy, so it exits and stays exited. The CLI reports "containers started successfully" and burns its retry budget on a dead container. Users get past this only by pressing ▶ in Docker Desktop.
2. **A non-essential config call aborts the run.** Token and default site both resolve fine; the UVE app-config `POST` then returns a transient 403 and the CLI calls `process.exit(1)`. Because UVE setup runs *before* scaffolding, the user is left with an empty directory and the working token and site ID are discarded without ever being printed.
3. **Recovery is blocked by the CLI's own side effects.** The port pre-check hard-fails on the now-running dotCMS's ports, and the directory-clearing prompt would delete the `docker-compose.yml` needed to tear it down.

**Verified against the tree** before writing — every claim in the issue holds:

* compose: `dotcms` has `depends_on` with no `condition`, no `restart:`, no healthcheck, and does not publish 8090; `opensearch` has no healthcheck; `db`'s healthcheck exists but nothing consumes it
* `src/index.ts:370` exits before the scaffolding at `:377`
* `src/index.ts:597` tests `if (!result)` against a truthy `{ ok: false, val }`, making the npm-install failure branch unreachable
* `checkPortsAvailability()` (`src/utils/index.ts:479`) hard-fails on exactly the ports a successful run holds
* the package contains **no spec files at all** — this fix establishes the test harness

**Deliberately deferred** as explicit non-goals, so the P0 fix stays bounded:

* `user.isAdmin()` swallowing exceptions into a false "not admin" — real defect, but legacy Liferay code on a hot permission path with wide blast radius, and the P0 fix doesn't depend on it
* Pinning the image tag alongside `CUSTOM_STARTER_URL` — intersects binding **ADR-0019** (date-lockstep SDK versioning), so it deserves its own decision rather than riding along in a bug fix

### Reviewers — and two questions for you

`.github/CODEOWNERS` does not cover `docker/`, so the highest-blast-radius file in this fix gets no automatic reviewer. Reviewers below are derived from `git blame` on the exact hunks the implementation will change, weighting design ownership over line count (the two largest raw counts are a bulk restore, #27432, and a pgvector bump, #29915 — both mechanical).

| Reviewer | Why |
| --- | --- |
| @wezell | OpenSearch 1.x + SSL in this file (#27754) and the Postgres 18 upgrade (#34236); owns the `db`/`opensearch` shape we're adding healthchecks to. Most senior still-active owner of `docker/`. |
| @erickgonzalez | Most recent semantic change to the `dotcms` service (#36490); closest to the `CUSTOM_STARTER_URL` / starter-import behavior the readiness race depends on |
| @jcastro-dotcms | Not from blame on this file — added because blame's strongest signal is unavailable (below). Second-most-active contributor to `docker/` over the last 12 months. |

**Blame's two strongest signals can't be assigned.** `spbolton` holds dominant blame on *every* hunk in scope and authored `lgtm-observability` (#32980) — the exact `condition: service_healthy` pattern this change copies — and `dcolina` holds part of the `opensearch` block (#29915). Both are no longer collaborators on this repo (last commits 2026-03-30 and 2026-04-07), so GitHub rejects the review request. Flagging it rather than quietly dropping them: **the person who designed the pattern we're copying is not available to check that we're copying it faithfully.** If anyone has context on the lgtm-observability healthcheck choices, that's the gap to fill.

> ### ⚠️ Updated — the diagnosis changed
>
> Planning tested this spec's own root cause 2 and **disproved it**. The 403 is not a transient
> startup race. It is **permanent damage caused by root cause 1**, and the two "independent bugs"
> are one causal chain. Details in Root-Cause Hypothesis; summary below.

**Experiment 1 — a clean boot has no settling window.** Measured on an M5/64GB host with dotCMS constrained to 2 CPUs / 4G:

| Signal | First success |
|---|---|
| `POST api-token` | **46s** |
| `GET apps/dotema-config-v2/{site}` | **46s** |
| `/dotmgt/livez`, `/dotmgt/readyz` | 48s |
| `/api/v1/appconfiguration` (current CLI probe) | 49s |

The UVE endpoint is usable **two seconds before `readyz` goes green**. The starter import (T+20s) and ES reindex (T+44s) both complete *inside* Tomcat startup, and the connector accepts no traffic until after them. The hypothesised race cannot happen.

**Experiment 2 — the reporter's actual path reproduces it, permanently.** Killing dotCMS mid starter-import and hand-starting it (reproduction step 4):

```
T+39s  appconfiguration 200 — CLI proceeds
T+41s  api-token 200 ・ defaultSite 200 ・ UVE GET 403 ・ UVE POST 403
   … 193 consecutive attempts over ~7 minutes, zero successes …
T+440s UVE GET 403 ・ UVE POST 403
```

The reported log, line for line. Server-side:

```
DotSecurityException: User 'Admin User [dotcms.org.1][admin@dotcms.com]'
  does not have READ permissions on Site 'demo.dotcms.com'
```

The interrupted import never wrote the site's permission rows; the restart re-runs `Task00004LoadStarter`, reports a clean startup, and the permissions never appear. **The instance does not recover.**

**What changed in the spec as a result:**

* **AC-005 no longer polls until 200** — it would never terminate. Retry is restricted to `5xx`; 403 explicitly does not retry.
* **The failure guidance is rewritten.** "Configure UVE manually at this URL" is wrong advice — manual setup fails identically. The CLI must say the instance is unrecoverable and to run `docker compose down -v`.
* **The P2 backend non-goal is re-pointed.** Not `user.isAdmin()` swallowing an exception (there is nothing to swallow — the data is absent), but a larger defect: **any interrupted first boot silently bricks the instance** while reporting success. Filed separately as #37268.
* **Title and Reproducibility corrected** — the 403 is deterministic once the crash has happened, not timing-dependent.

**This raises US1's value.** Fixing the compose file doesn't just stop a crash — it removes the 403 entirely, because it removes the interrupted boot that causes it.

*Evidence limits: one host, one starter, one image, kill point fixed at 25s. Which kill-points corrupt is unmapped.*

---

**Updated after planning.** Verifying the spec's claims against the tree turned up four that didn't hold, and I've folded the corrections back in before sign-off so you're approving what will actually be built:

* **"Six siblings use `condition: service_healthy`"** — there are **three**, and **none of them gates `dotcms` on OpenSearch being healthy** (all use `service_started`). So doing both is a deliberate deviation from every precedent, not the house pattern the spec implied. Kept, with the reasoning stated, using `single-node-os-migration`'s proven probe.
* **The flapping risk I flagged can't happen.** Compose restart policies react to container *exit*, not health status. Replaced with the real exposure: a wrong probe blocks `docker compose up --wait` until timeout.
* **`8090:8090` → `127.0.0.1:8090:8090`.** `InfrastructureManagementFilter` authorizes purely by arrival port — no credential check, no IP allowlist — so a wildcard binding puts `/dotmgt/health` and `/dotmgt/metrics` on the local network. New **AC-011** asserts it's refused on the LAN address.
* **The Jest harness already exists**; only spec files are missing. The spec said this fix establishes it.

Also added **AC-012** for a landmine found while reading: `updateDockerComposeStarterUrl` rewrites the compose file with a regex and **throws on no match**, so reformatting `CUSTOM_STARTER_URL` would break `--starter` for every already-installed CLI, with no release able to reach them.

**What's left for you to settle** (the 8090 question is now answered above):

1. **Gating `dotcms` on `opensearch: service_healthy`** goes further than all four existing examples. Right call for a stack driven by an unattended CLI, or should it match the house pattern? *(My recommendation: keep the stricter gate.)*
2. **Is `start_period: 180s` enough** for a cold demo-starter import on a slow machine? Above both precedents (lgtm 120s, metrics-monitoring 20s). Too short and `--wait` blocks to timeout.

### Note on blast radius

The compose file is **fetched from `main` at runtime by every installed `@dotcms/create-app`**. A change ships instantly and unversioned to all existing CLI users — including older versions that won't know about port 8090. That cuts both ways: it's why the fix is P0 and reaches users without a release, and it's the single largest regression risk here. Worth a careful look.

### Checklist
- [ ] Tests — n/a for a spec-only PR; the spec defines the Jest specs and cold-start compose verification that PR 2 must land first (constitution Principle V: Red before Green)
- [ ] Translations — n/a
- [x] Security Implications Contemplated — publishing management port 8090 on the host is called out in the spec's Regression Risk and raised as review question 1 above

### Additional Info

* Follows the spec-kit flow in `CLAUDE.md`: `/speckit-specify-fix` → **PR 1 approved** → `/speckit-plan` → `/speckit-tasks` → `/speckit-implement` → PR 2
* ADR consultation ran as the `before_plan` hook. **ADR-0019** (accepted, binding) is recorded in the spec; ADR-0016 matched on keywords but is an unfilled template and imposes nothing. No conflicts with an accepted ADR.
* Related: #35096 (E2E suite for `@dotcms/create-app`) — owns the fault-injection coverage, out of scope here

Refs #37262

This PR fixes: #37262

This PR fixes: #37262

### DIFF
--- a/specs/37262-create-app-docker-uve/spec.md
+++ b/specs/37262-create-app-docker-uve/spec.md
@@ -200,7 +200,7 @@ the permission data is genuinely absent, so `isAdmin()` has nothing to throw abo
   from an interrupted first boot and must be recreated with `docker compose down -v`.
 - **A separate backend defect is implied**: any interrupted first boot silently bricks the
   instance, and a restart neither repairs nor reports it. That is broader than this issue and is
-  filed separately.
+  filed as **#37268**.
 
 **On the readiness signal.** Switching to `/dotmgt/readyz` on 8090 is still worth doing — it is
 the purpose-built probe and does not depend on the web app — but it is a correctness tidy-up,
@@ -294,7 +294,7 @@ runtime, so it reaches every already-installed CLI immediately):*
   silently bricks the instance**: the starter import leaves site permissions unwritten, a restart
   re-runs `Task00004LoadStarter` and reports success, and every Apps API call 403s forever with no
   warning to the user. That is out of scope here (it is a starter-import/permissions problem in
-  legacy `com.dotmarketing.*`, not a CLI one) and is filed separately. This fix removes the
+  legacy `com.dotmarketing.*`, not a CLI one) and is filed as **#37268**. This fix removes the
   *trigger* by stopping the crash.
 - **Pinning the dotCMS image tag alongside `CUSTOM_STARTER_URL`.** Real drift risk
   (`latest` + hardcoded `starter-20260630`) and it intersects binding ADR-0019, so it deserves
--- a/specs/37262-create-app-docker-uve/spec.md
+++ b/specs/37262-create-app-docker-uve/spec.md
@@ -269,13 +269,20 @@ runtime, so it reaches every already-installed CLI immediately):*
 
 - The CLI never exits without printing recoverable state (host, token, site ID), and writes the
   `.env` it already has every value for.
-- UVE setup failure is non-fatal: warn, print manual setup steps, and continue to scaffolding.
-  The warning must link the user to the official headless UVE configuration guide —
-  <https://dev.dotcms.com/docs/author/pages-and-visual-editing/universal-visual-editor/uve-headless-config>
-  — alongside the concrete values they need (host, site ID, and the app key
-  `dotema-config-v2`), so the one unset setting is self-serviceable rather than a dead end.
-- Gate the UVE write on a read — poll `GET` on the UVE app endpoint until 200, then `POST` with
-  retry on 401/403/5xx.
+- UVE setup failure is non-fatal: warn and continue to scaffolding. The message depends on why
+  it failed, because the two cases need opposite advice:
+  - **403 (terminal)** — the instance's permissions were never written by an interrupted first
+    boot. Manual UVE configuration would fail identically, so **do not** offer the manual steps.
+    Tell the user the instance is unrecoverable and to run
+    `docker compose down -v && docker compose up -d --wait`, and reference #37268.
+  - **anything else** — link the official headless UVE configuration guide,
+    <https://dev.dotcms.com/docs/author/pages-and-visual-editing/universal-visual-editor/uve-headless-config>,
+    alongside the concrete values needed (host, site ID, app key `dotema-config-v2`), so the one
+    unset setting is self-serviceable rather than a dead end.
+- Probe before the UVE write — a **single** `GET` on the UVE app endpoint; on 200, `POST` with
+  retry on `5xx` **only**. Never poll, and never retry a 403: measurement showed a 403 here is
+  terminal (193 consecutive failures over ~7 minutes), so a poll would spin forever. On 403 the
+  CLI reports the instance as unrecoverable and stops — see the terminal-403 message below.
 
 *CLI (P1):*
 
--- a/specs/37262-create-app-docker-uve/spec.md
+++ b/specs/37262-create-app-docker-uve/spec.md
@@ -413,35 +413,26 @@ so only a warm npx cache stays behind.
   empty directory by the old behavior simply re-run; the P1 port-reuse work is what makes that
   re-run possible without tearing down a healthy instance.
 
-### Required reviewers for the compose change
+### Compose design decisions
 
-**Rescoped.** These reviewers were selected by `git blame` on the shared
-`single-node-demo-site/docker-compose.yml` back when this work modified it. **It no longer does** —
-the CLI ships its own file — so the blast radius that made their review essential is gone. They are
-still the right people to sanity-check a *new* dotCMS compose stack (healthcheck shapes, ordering,
-port exposure), but this is now a review of new code in the SDK tree rather than a change to
-infrastructure they own. `.github/CODEOWNERS` covers neither path.
+**No `docker/` reviewers are required.** An earlier draft carried a `git blame`-derived reviewer
+list for `single-node-demo-site/docker-compose.yml`, chosen back when this work modified that file.
+The rescope means it does not: what ships is a new compose file inside
+`core-web/libs/sdk/create-app/assets/`, reviewed as SDK code along with the rest of the CLI change.
+There is no change to infrastructure anyone owns, so the list is dropped rather than carried as
+courtesy CCs. (`.github/CODEOWNERS` covers neither path, so PR 2 draws no automatic reviewer
+either way — worth knowing when requesting review.)
 
-| Reviewer | Why they should review |
-| --- | --- |
-| **Will Ezell** (`wezell`) | Authored the OpenSearch 1.x + SSL setup in this file (#27754) and the Postgres 18 upgrade (#34236) that touched both this file and the model stack. Owns the current `db`/`opensearch` shape we are adding healthchecks to, and is the most senior still-active owner of `docker/`. |
-| **Erick González** (`erickgonzalez`) | Most recent semantic change to the `dotcms` service (#36490, 2026-07-13) and prior starter-version work (#36362). Closest to the `CUSTOM_STARTER_URL` / starter-import behavior that the readiness race depends on. |
-| **Jose Castro** (`jcastro-dotcms`) | Not a blame match on this file. Added to cover the gap below: second-most-active contributor to `docker/` over the last 12 months. |
+One piece of guidance from that analysis survives, because it is about the code and not about who
+signs off: **the plan phase should read `lgtm-observability/docker-compose.yml` directly as the
+reference for correct `condition: service_healthy` usage**, rather than assume the pattern was
+copied faithfully. Its author is no longer a collaborator, so there is nobody to ask — the file is
+the specification. Note it binds 8090 on the wildcard (L195), which this design deliberately does
+not copy.
 
-**Unavailable — the two strongest blame signals.** `spbolton` (Steve Bolton) holds dominant blame
-on every hunk in scope *and* authored `lgtm-observability/docker-compose.yml` (#32980), the exact
-`condition: service_healthy` pattern this change copies. `dcolina` (Daniel Colina) holds 13 lines
-of the `opensearch` block via #29915. Neither is a collaborator on `dotCMS/core` any more (last
-commits 2026-03-30 and 2026-04-07 respectively), so GitHub rejects a review request for them.
-
-This leaves a real coverage gap: **nobody currently assignable designed the pattern being copied.**
-The plan phase should treat the lgtm-observability compose as the specification for correct
-usage — reading it directly rather than relying on a reviewer to catch a faithful-copy error.
-
-**All open questions are now settled.** Both review questions this section previously put to
-reviewers have been decided by the issue owner, and the decisions are recorded in Fix Scope rather
-than left for the reviewer to resolve. Kept here with their reasoning so reviewers see what was
-chosen and can object, instead of having to rediscover it:
+**All open questions are now settled.** Both questions this section previously left open have been
+decided by the issue owner, and the decisions are recorded in Fix Scope. Kept here with their
+reasoning so a reviewer can see what was chosen and object, rather than having to rediscover it:
 
 1. **Gate `dotcms` on `opensearch: service_healthy`** — **decided: keep the stricter gate**, using
    `single-node-os-migration`'s proven probe (L61–65). The framing this question originally carried
@@ -461,7 +452,7 @@ chosen and can object, instead of having to rediscover it:
    by arrival port with no credential check and no IP allowlist. Note this deviates from
    `lgtm-observability`, which binds the wildcard (L195); the deviation is deliberate.
 
-What remains for reviewers is not a decision but a **confirmation**: `/dotmgt/livez` must be
+What remains is not a decision but a **confirmation**: `/dotmgt/livez` must be
 verified on the released `dotcms/dotcms:latest` image before the healthcheck can depend on it. Both
 in-repo examples that probe it run `dotcms/dotcms-test`. This is the first step of the plan's
 verification guide, and a failure there invalidates the compose design. See Assumptions.
--- a/specs/37262-create-app-docker-uve/spec.md
+++ b/specs/37262-create-app-docker-uve/spec.md
@@ -97,8 +97,12 @@ reproduces when the CLI reaches the UVE call while the instance is still settlin
 
 - **Affected area**: Three surfaces, in priority order.
   1. **Docker compose examples** — `docker/docker-compose-examples/single-node-demo-site/docker-compose.yml`.
-     This is the least hardened example in that directory; six siblings already use
-     `condition: service_healthy`, and `lgtm-observability/docker-compose.yml` is the model.
+     This is the least hardened example in that directory. **Three** siblings already use
+     `condition: service_healthy` (`lgtm-observability`, `single-node-metrics-monitoring`,
+     `experiments`, plus `single-node-os-migration` via provision jobs) —
+     `lgtm-observability/docker-compose.yml` is the model. Note none of them gates `dotcms`
+     on OpenSearch being *healthy*; every one uses `db: service_healthy` +
+     `opensearch: service_started`. See Fix Scope for why this fix goes further.
      Verified in-repo: `dotcms` has `depends_on: [db, opensearch]` with no condition, no
      `restart:`, no healthcheck, and does not publish 8090; `opensearch` has no healthcheck and
      no `restart:`; only `db` defines a healthcheck, which nothing consumes.
@@ -187,9 +191,21 @@ runtime, so it reaches every already-installed CLI immediately):*
 
 - Add an `opensearch` healthcheck and `restart: unless-stopped`.
 - Change `dotcms` `depends_on` to `condition: service_healthy` for both `db` and `opensearch`.
+  Gating on `db` is what fixes the reported crash. Gating on `opensearch` as well is a
+  **deliberate deviation** from all four existing examples, which use `service_started` there —
+  justified because this stack is driven by an unattended CLI, so an OpenSearch that is
+  up-but-not-ready is a failure with nobody present to diagnose it. The OpenSearch probe should
+  be the one already proven in `single-node-os-migration` (`-k` for the self-signed cert,
+  `-u admin:admin`, since this stack sets `DOT_ES_AUTH_BASIC_PASSWORD: 'admin'`), not a new one.
 - Add a `dotcms` healthcheck against `http://127.0.0.1:8090/dotmgt/livez` with
   `start_period: 180s`, plus `restart: unless-stopped`.
-- Publish port `8090:8090`.
+- Publish the management port **bound to loopback**: `127.0.0.1:8090:8090` — not `8090:8090`.
+  The management port is authorized purely by the port a request arrives on
+  (`InfrastructureManagementFilter`): no credential check, no IP allowlist. A wildcard binding
+  would put `/dotmgt/health` and `/dotmgt/metrics` on the local network for every user of this
+  demo stack. Loopback gives the CLI and the container's own healthcheck everything they need
+  (both already use `127.0.0.1`) and gives the network nothing. This is stricter than the two
+  existing examples that publish 8090.
 
 *CLI (P0):*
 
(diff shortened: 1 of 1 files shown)


specs/37262-create-app-docker-uve/spec.md
```diff
@@ -0,0 +1,564 @@
...
+  server will work — `docker compose down -v` must never be suggested there, as there is no stack
+  to recreate.
+- **AC-006** *(P1)*: With a healthy dotCMS already answering on 8082, a second run offers to
+  reuse it instead of aborting with "Required ports are already in use". Reproduction step 6 no
+  longer blocks.
+- **AC-007** *(P1)*: A failed `npm install` causes the CLI to report failure — the branch at
+  `src/index.ts:597` is reachable and correct.
+- **AC-008** *(P1)*: If scaffolding fails after `moveDockerComposeOneLevelUp()`, the compose
+  file is restored to the project directory (no orphan in the parent).
+- **AC-009**: Feedback is **continuous for the entire wait**, which may be up to ten minutes
```
nicobytes: **The P0/P1 split is not separable as written, and AC-009's tier is ambiguous.**

Two related things:

1. **AC-009 carries no `(P1)` tag**, unlike AC-006/007/008 — but its content (streamed per-container `Waiting → Healthy` transitions, the ~2s ticker, visible pull progress) maps directly to the "CLI (P1)" bullet in Fix Scope (`docker compose up -d --wait` and stream pull progress). Is the missing tag deliberate or an oversight?

2. **More importantly, `--wait` looks like a P0 dependency labelled P1.** It is listed under "CLI (P1)", yet AC-001 ("brings the stack up **without manual intervention**") and AC-002 ("only printed when the containers are actually running and healthy") are P0 and have no other stated mechanism. If P1 ships separately, AC-001 and AC-002 lose the thing that satisfies them. #37264 reinforces this reading — its contract states the CLI waits until healthy "via `docker compose up -d --wait --wait-timeout 600` (D7)" as part of the core contract, with no tier qualifier.

Also worth deciding explicitly: **are AC-006, AC-007 and AC-008 part of this PR's Definition of Done, or a follow-up PR?** The spec never says. Note the plan artifacts in #37264 do not mention `P0`/`P1` anywhere, and never reference AC-006/007/008 — so the tiering has effectively dropped out of the design rather than been resolved.

Suggested resolution: promote `--wait` (and the AC-009 feedback it needs) to P0 explicitly, and state whether the remaining P1 items are in or out of this PR.

nicobytes (commented):
Two questions from a `/speckit-clarify` ambiguity pass over the spec. Both are gaps that change task decomposition or test design, so they are worth settling before `/speckit-plan` output is treated as final — I checked #37264 and neither is resolved there. Details inline on the two ACs they affect.

specs/37262-create-app-docker-uve/spec.md
```diff
@@ -0,0 +1,564 @@
...
+  and the `DOTCMS_COMPOSE_URL` override.
+- **AC-011**: Port 8090 is reachable on `127.0.0.1` and **not** on the host's LAN address —
+  the management surface is not exposed to the network.
+- **AC-012**: `npx @dotcms/create-app --starter <url>` still works against the bundled compose
+  file — the `CUSTOM_STARTER_URL` rewrite regex in `updateDockerComposeStarterUrl` must still match.
+  Enforced by a **Jest spec that loads the real bundled asset**, asserts the regex matches it and
+  asserts the rewritten line, so a reformat of that line fails CI on the PR that causes it. No
+  Docker and no cold start are required, which is what makes this cheap enough to gate every PR;
+  it is available because the rescope put the regex and the file it rewrites in the same package.
+- **AC-013**: The bundled compose file is actually present in the published package. `package.json`
```
nicobytes: **AC-013 has no verification method — and it is the one AC the spec itself calls the most likely way to break the release.**

The Verification method section defines no gate for AC-013, unlike AC-012 which names a Jest spec explicitly and justifies it as "cheap enough to gate every PR".

The risk is concrete in the current config:
- `package.json` has `files: ["*.js", "README.md"]` — a flat glob that will **not** match `assets/docker-compose.yml`.
- Both esbuild `assets` entries in `project.json` use `output: "."`, so the built output path needs deciding too.

#37264 identifies the same gotcha (D4a "Packaging gotchas", C7 "MUST be listed in both ... or it will silently not ship") but **still adds no check for it**. `verify-cold-start.sh --static` asserts T008 (`CUSTOM_STARTER_URL`), the `depends_on` condition, healthchecks, restart policies and the 8090 loopback binding — none of which is the packaging assertion. And it resolves the file as `$SCRIPT_DIR/../assets/docker-compose.yml`, i.e. from the **source tree**, so it passes identically whether or not the asset ever ships.

So the failure mode is: every static check goes green, the package publishes without the compose file, and every local-Docker run fails at the first step — exactly what AC-013 exists to prevent.

What is the intended gate? Candidates: a post-build assertion over `dist/libs/sdk/create-app`, an `npm pack --dry-run` content check in CI, or a Jest spec resolving the asset relative to the bundle entry point rather than the source dir.

zJaaal:
## Review: spec is approvable, with one missed defect and two claims to soften

I checked the spec's claims against the tree rather than just reading it. Everything cited verifies. Details below, in the order I'd act on them.

### 1. Missed defect: two unawaited async calls (`src/index.ts:376`, `:378`)

`moveDockerComposeOneLevelUp()` and `moveDockerComposeBack()` are both `async` (`src/git/index.ts:76`, `:82`, each doing `await fs.rename`), and both are called without `await`:

```ts
// required since git requires empty directory
moveDockerComposeOneLevelUp(finalDirectory);
await startScaffoldingFrontEnd({ spinner, selectedFramework, finalDirectory });
moveDockerComposeBack(finalDirectory);
```

The comment states the very requirement the missing `await` breaks: the rename may not have landed before the git clone runs against a directory that must be empty. AC-008 covers this file but only asks for `try/finally`, which does not fix a floating promise. Suggest AC-008 also require the `await`s.

Related wording slip in Cause 3: the spec says `moveDockerComposeOneLevelUp()` "calls `process.exit(1)` internally". It does not. The exits are in `startScaffoldingFrontEnd` at `:588` and `:599`. The orphaned-compose-file conclusion still holds, only the attribution is wrong.

### 2. The OpenSearch "prior art" is thinner than written

The spec leans on `single-node-os-migration` twice: for the probe (L61-65, verified verbatim) and for "transitive prior art" that `dotcms` gates on OpenSearch health. That file is the tester harness for the **unreleased** OpenSearch 3.x migration (`DOT_FEATURE_FLAG_OPEN_SEARCH_PHASE`, phases 0-3), and its gate exists because provision jobs must create search users before dotCMS connects, not as a boot-ordering practice.

The three non-migration examples that gate `dotcms` all use `db: service_healthy` + `opensearch: service_started` (lgtm L90/L92, metrics-monitoring L88/L90, experiments L139/L141), exactly as the spec states. So the stricter gate **is** a deliberate deviation from every non-migration precedent. Regression Risk already owns the new failure mode, and since the file is now CLI-owned the blast radius is contained, so this is not a design objection. I'd just drop the "not the clean break from precedent an earlier draft claimed" framing: the earlier draft was closer to right.

Footnote on the same probe: `single-node-os-migration` pins `opensearchproject/opensearch:1.3.20`, while the demo stack uses floating `:1`. The independent verification on this stack at ~15s is what actually matters, so this is minor, but "proven in `single-node-os-migration`" was proven against a different image.

### 3. Question on AC-010

AC-010 freezes `docker/docker-compose-examples/single-node-demo-site/docker-compose.yml`, which leaves everyone on CLI <=1.2.5 **and every README reader** on the broken file indefinitely. Worth deciding explicitly rather than by omission: should a one-line `db: condition: service_healthy` land there separately? The healthcheck already exists at L13 and nothing consumes it, so that single change addresses the actual root cause for existing users with none of the risk the rescope removed: no OpenSearch health gate, no 8090 publishing, no new failure mode.

### What verified clean

Every cited line and count holds: `src/index.ts:226-228` and `:369-371` are byte-identical fatal blocks; `:597`'s `if (!result)` against a truthy `Err()` is real; `src/utils/index.ts:479` and `src/asks.ts:180` are correct; the `CUSTOM_STARTER_URL` regex sits at `:490` and throws at `:496`; `fetchWithRetry` accepts any 2xx (`utils:55`) while `isDotcmsRunning` demands exactly 200 (`index.ts:507`); zero spec files present, Jest harness in place, `passWithNoTests: true` at `project.json:75`.

**AC-013 is well aimed and I'd treat it as the highest-value AC in the list.** `package.json` `files` is `["*.js", "README.md"]` and the esbuild `assets` block lists only README and package.json, so a bundled compose asset ships missing unless both are edited.

Compose precedents also check out: four files use `condition: service_healthy`, lgtm binds 8090 on the wildcard at L195, and both `/dotmgt/livez` probers run `dotcms/dotcms-test:1.0.0-SNAPSHOT` with `start_period` 120s and 20s respectively. The "confirm `/dotmgt/livez` on the released image before the healthcheck depends on it" prerequisite is correctly flagged as unverified, and I agree it gates the compose design.

One incidental find while verifying, not worth an AC: `checkPortsAvailability()` still labels 9200/9600 as "Elasticsearch HTTP" and "Elasticsearch Transport" while the stack has run OpenSearch since #27754. Cosmetic, but it is the kind of thing that makes readers doubt which search backend the demo actually uses.

---
*Reviewed by Claude Code on @zJaaal's behalf; the claim verification above was run against the working tree.*

zJaaal (dismissed):

zJaaal (approved):

PR #37273, opened 2026-08-28
37207 migrate openbrowsermodal to the new assetpicker and delete the legacy browser selector

### LINKED ISSUE #37207: Migrate openBrowserModal to the new AssetPicker and delete the legacy Browser Selector
## Description

`DotCustomFieldApi.openBrowserModal()` — the public asset/content browser exposed to **custom field VTL templates** — is the last production caller of the legacy `DotBrowserSelectorComponent`. Every other asset-selection entry point (Block Editor, WYSIWYG, File/Image) moved to `DotAssetPickerComponent` in #36944. Switching this one over is therefore what unblocks deleting the legacy picker outright.

**#37112 is the enabler.** Merged 2026-08-24 (fixes #36991), it added `showLinks` support plus independent cursor-based paging for menu links to `/api/v1/drive/search` (`BrowserAPIImpl.getPaginatedContents`, new `linkCursor` / `linkCount` / `hasMoreLinks` / `nextLinkCursor`). `showLinks` is part of `openBrowserModal`'s published contract via `ContentByFolderParams`, and the AssetPicker — which browses through the Content Drive API — could not honor it until that landed.

### Current state (verified on `main`)

| Item | Location | Note |
|---|---|---|
| `openBrowserModal()` opens the legacy picker | `core-web/libs/edit-content-bridge/src/lib/bridges/angular-form-bridge.ts:529-568` | PrimeNG `DialogService`, 90% width / 1040px max, maps result to `BrowserSelectorResult` |
| Legacy Dojo bridge is a no-op | `core-web/libs/edit-content-bridge/src/lib/bridges/dojo-form-bridge.ts:329-335` | `// TODO: Implement browser selector modal for Dojo` — the Dojo path never opens this modal at all |
| Only remaining consumer of the old picker | `angular-form-bridge.ts:533` | Confirmed by grep across `libs` + `apps` |
| Legacy picker subtree | `core-web/libs/ui/src/lib/components/dot-browser-selector/` | 13 files: component, `store/browser.store.ts`, `components/dot-dataview/`, `components/dot-sidebar/`. **No external consumers of any internal** |
| Public export | `core-web/libs/ui/src/index.ts:70` | `DotBrowserSelectorComponent` |
| Real VTL consumers | `dotCMS/.../static/content/file_browser_field_render_new.vtl:21`, `dotCMS/.../static/htmlpage_assets/redirect_custom_field_new.vtl:55` | The file-browser one passes `showFiles/showPages/showFolders`, `showDotAssets:false`, `showWorking:false`, `showArchived:false`, `sortByDesc:true` and consumes `result.url` |

### ⚠ This is not a component swap — there is a real capability gap

`openBrowserModal` is a **browser** (it can return pages, folders and menu links); `DotAssetPickerComponent` is an **asset picker**. The new picker must be extended before it can stand in:

| `openBrowserModal` capability (`ContentByFolderParams`) | AssetPicker today |
|---|---|
| `showFolders` | Hardcoded `showFolders: false` as a documented **invariant** (`store/features/with-asset-browse.feature.ts:52,138`; the paging model relies on the folder cursor never advancing) |
| `showLinks` (menu links) | No support at all — newly possible only thanks to #37112 |
| `showPages` / `mimeTypes: ['application/dotpage']` | Restricted to `DOTASSET` + `FILEASSET` via `ASSET_PICKER_ASSET_BASE_TYPES`; no `HTMLPAGE` path |
| `showWorking`, `showArchived`, `sortByDesc`, `extensions`, `showDotAssets` | No equivalents on `DotAssetPickerConfig` |
| Result carries `url` / `urlMap` | Picker returns contentlets; `BrowserSelectorResult.url` is what both VTL consumers actually read |

**Decision taken:** the new AssetPicker becomes the single picker in **every** host, including the legacy Dojo editor, and the legacy picker is deleted in this work. `dot-asset-picker.component.legacy-host.spec.ts` already proves the picker constructs in the bare `dotcms-binary-field` host (no `Router`, no app-shell providers), which is the evidence this is viable.

**Consequence for #37132:** #37132 asks for the *opposite* — restore `DotBrowserSelectorComponent` (and `DotAssetSearchDialogComponent`) as the legacy-host picker behind an `ASSET_PICKER_LAUNCHER` token. That approach is superseded by this decision. **#37132 must be reoriented or closed as part of this work** — it cannot be delivered alongside it.

### Extend the AssetPicker to cover the browse contract
- [ ] The picker can list and return **folders** when the caller asks for them, and the paging model no longer assumes the folder cursor never advances
- [ ] The picker can list and return **menu links**, consuming the `showLinks` / `linkCursor` / `hasMoreLinks` / `nextLinkCursor` contract added in #37112
- [ ] The picker can list and return **pages** (`HTMLPAGE` / `application/dotpage`) when the caller asks for them, without `ASSET_PICKER_ASSET_BASE_TYPES` leaking pages into the File/Image/media entry points
- [ ] `showWorking`, `showArchived`, `showDotAssets`, `sortByDesc` and `extensions` are honored, or each unsupported flag is explicitly documented as dropped with the reason
- [ ] The existing entry points (File, Image, video, audio) keep their current narrowing exactly — no folders, no links, no pages, same silent mimetype restriction
- [ ] Each newly supported capability is opt-in per caller, so the default remains today's asset-only behavior

### Migrate `openBrowserModal`
- [ ] `AngularFormBridge.openBrowserModal()` opens `DotAssetPickerComponent`
- [ ] `BrowserSelectorOptions`, `BrowserSelectorResult` and `BrowserSelectorController` in `core-web/libs/edit-content-bridge/src/lib/interfaces/browser-selector.interface.ts` are unchanged — this is a published contract for customer VTL templates
- [ ] Every field of `BrowserSelectorResult` is still populated: `identifier`, `inode`, `title`, `name` (falling back to `fileName`), `url` (falling back to `urlMap`), `mimeType`, `baseType`, `contentType`
- [ ] Selecting a **folder** and selecting a **menu link** both return a usable `url`, since that is the only field the shipped VTL templates read
- [ ] Cancelling — ✕, Esc, or mask click — invokes `onClose(null)` exactly once
- [ ] The returned controller's `close()` still closes the dialog programmatically
- [ ] `options.header` is respected, given the picker renders its own header rather than using `DynamicDialogConfig.header`
- [ ] Both shipped VTL templates work end to end: `file_browser_field_render_new.vtl` (browse → path lands in the `vlUri` input and the field value) and `redirect_custom_field_new.vtl`

### Delete the legacy picker
- [ ] The whole `core-web/libs/ui/src/lib/components/dot-browser-selector/` subtree is removed, including `store/browser.store.ts`, `components/dot-dataview/` and `components/dot-sidebar/`
- [ ] The `DotBrowserSelectorComponent` export is removed from `core-web/libs/ui/src/index.ts:70`
- [ ] `DotBrowsingService` and the shared `DotFolderTreeComponent` survive — they live outside this subtree and have other consumers
- [ ] `grep -rn "DotBrowserSelectorComponent"` over `core-web` returns nothing outside deleted files
- [ ] It is decided and recorded whether `DotAssetSearchDialogComponent` (`libs/ui/src/lib/components/dot-asset-search/`) is also deleted — it likewise has zero production consumers after #36944, and #37132 was the only thing asking for it back
- [ ] No dead i18n keys, SCSS, or model types are left behind by the deletion

### Resolve the #37132 conflict
- [ ] #37132 is reoriented or closed with a written rationale, before this work merges
- [ ] The Block Editor and File/Image entry points are confirmed to behave acceptably in the legacy Dojo editor with the new picker — this is the concern #37132 raised and it must be answered, not skipped
- [ ] No `ASSET_PICKER_LAUNCHER`-style host-discriminating token is introduced, since the new picker is now used in every host

### Tests
- [ ] `angular-form-bridge.spec.ts`'s `openBrowserModal` suite is updated to the new picker and still covers option pass-through, result mapping, cancellation, `close()`, and the `NgZone` wrapping
- [ ] `dot-asset-picker.component.legacy-host.spec.ts` is kept and extended, since the legacy Dojo host is now a supported host rather than an excluded one
- [ ] New coverage for folder, menu-link and page selection returning correct `url` values
- [ ] The Playwright specs from #36944 (`wysiwyg-field-asset-picker.spec.ts`, `block-editor-field-asset-picker.spec.ts`, `file-field.spec.ts`, `image-field.spec.ts`) still pass
- [ ] Frontend lint/tests green: `pnpm nx test ui`, `pnpm nx test edit-content-bridge`

## Priority

Medium

## Additional Context

- **Epic:** #36702 — Create AssetPicker component reusing Content Drive
- **Enabling PR:** #37112 — `showLinks` support in `/api/v1/drive/search` (fixes #36991)
- **Prior migration:** #36944 — moved Block Editor / WYSIWYG / File-Image to the new picker
- **Conflicting issue to reorient or close:** #37132 — New AssetPicker must not be used in the legacy Dojo editor
- **Rollback note:** `openBrowserModal` is a documented public API for customer custom-field VTL templates. Any change to `BrowserSelectorOptions` / `BrowserSelectorResult` would be a breaking API-contract change — see [Rollback-Unsafe Change Categories](docs/core/ROLLBACK_UNSAFE_CATEGORIES.md). Keeping those interfaces intact is what keeps this rollback-safe.
- **Suggested sequencing:** extend the picker (folders → links → pages) → switch `openBrowserModal` → verify both VTL templates → resolve #37132 → delete the legacy subtree.

nicobytes:
This pull request introduces support for browsing and selecting menu links (in addition to files, folders, and pages) in the asset picker modal used by custom field templates. It refactors the Angular form bridge to use the new asset picker, adds a mechanism for resolving the current site, and updates the relevant models and APIs to handle the new "link" type. The Dojo form bridge remains a stub for the browser modal, but now emits a warning for unsupported usage. The most important changes are grouped below.

**Asset Picker and Browsing Enhancements**

* Added a new `DotContentDriveLink` interface and updated `DotContentDriveItem` to include links, allowing the asset picker to display and select menu links alongside other item types. (`core-web/libs/dotcms-models/src/lib/dot-content-drive.model.ts`)
* Extended the search request and response models to support paginated retrieval of links (`showLinks`, `linkCursor`, `hasMoreLinks`, `nextLinkCursor`). (`core-web/libs/dotcms-models/src/lib/dot-content-drive.model.ts`) [[1]](diffhunk://#diff-d429093a2475fa9eb85bf95a1b82a39f77f1ecc5b96754a827524814925aed3bR235-R252) [[2]](diffhunk://#diff-d429093a2475fa9eb85bf95a1b82a39f77f1ecc5b96754a827524814925aed3bR347-R355)

**Angular Form Bridge Refactor**

* Refactored `AngularFormBridge.openBrowserModal()` to use the new `DotAssetPickerComponent`, supporting selection of pages, folders, and links, and returning a consistent handle with a promise-based API. (`core-web/libs/edit-content-bridge/src/lib/bridges/angular-form-bridge.ts`)
* Added a `resolveSite` mechanism to allow the bridge to obtain the current site context asynchronously, ensuring the picker always browses the correct site. (`core-web/libs/edit-content-bridge/src/lib/bridges/angular-form-bridge.ts`) [[1]](diffhunk://#diff-08b7f2dbb6ec1f4033d7a5801c7f63b080a8f9565b11c3f9b7c40ea512578e6fR57-R81) [[2]](diffhunk://#diff-08b7f2dbb6ec1f4033d7a5801c7f63b080a8f9565b11c3f9b7c40ea512578e6fR92-R108) [[3]](diffhunk://#diff-08b7f2dbb6ec1f4033d7a5801c7f63b080a8f9565b11c3f9b7c40ea512578e6fL111-R138) [[4]](diffhunk://#diff-08b7f2dbb6ec1f4033d7a5801c7f63b080a8f9565b11c3f9b7c40ea512578e6fR148-R151)

**Legacy Editor (Dojo) Handling**

* Updated the Dojo form bridge to emit a warning and return a well-formed, always-null handle when `openBrowserModal` is called, making unsupported usage more transparent. (`core-web/libs/edit-content-bridge/src/lib/bridges/dojo-form-bridge.ts`, `core-web/libs/edit-content-bridge/src/lib/bridges/dojo-form-bridge.spec.ts`) [[1]](diffhunk://#diff-fe3a388c08be205717a8a85cc52f63bf2b3aacac655960871dc86e0276987e09L2-R3) [[2]](diffhunk://#diff-d0fd14ebe4c48f46fc11d776a7a09ec7b511a7004d8aa03fc7731089e27c0334R379-R424)

**Feature Directory Update**

* Changed the feature directory in `.specify/feature.json` to reflect the new asset picker modal for menu links. (`.specify/feature.json`)

This PR fixes: #37207


core-web/libs/ui/src/lib/components/dot-asset-picker/store/features/with-asset-browse.feature.ts
```diff
@@ -49,7 +49,8 @@ const initialState: DotAssetPickerBrowseState = {
  *
  * Mirrors the shape of Content Drive's request builder, minus everything the picker has no use for
  * (workflow filters, user-searchable fields) and plus the two things it needs: a silent mimetype
- * restriction and `showFolders: false` as an invariant.
+ * restriction, and folders and menu links switched off unless the caller opts in through
+ * `DotAssetPickerBrowseOptions`.
```
oidacra, line 53: This comment is about `$totalRecords` at lines 74-83 (unchanged by this PR, so anchoring here at the nearest changed line).

That computed still inspects only `hasMoreContent`:

```ts
return bookmark?.hasMoreContent
    ? limit * (page + 1)
    : limit * (page - 1) + items().length;
```

Now that folders and links are independently paginated streams, is this still sufficient? Walking a browse-mode case with `limit=20` — if page 1 comes back `hasMoreContent: false` but `hasMoreFolders: true, hasMoreLinks: true` with 15 rows rendered, this returns `20*0 + 15 = 15`. The paginator then sees `first + rows >= totalRecords` and disables Next, so the remaining folders and links look unreachable even though the server still has them.

The converse (folders/links exhaust first, content continues) looks fine, since `hasMoreContent` keeps it growing.

Would something like this be right?

```ts
const hasMore = bookmark
    ? bookmark.hasMoreContent || bookmark.hasMoreFolders || bookmark.hasMoreLinks
    : true;

return hasMore ? limit * (page + 1) : limit * (page - 1) + items().length;
```

The docstring just above ("answers is there anything past what is on screen") would want updating too, since it now speaks for three streams rather than one.

core-web/libs/edit-content-bridge/src/lib/bridges/angular-form-bridge.ts
```diff
@@ -470,105 +501,220 @@ export class AngularFormBridge implements FormBridge {
...
+        );
+    }
+
+    return {
+        ...(kinds.includes('folder') ? { showFolders: true } : {}),
+        ...(wantsLinks ? { showLinks: true } : {}),
+        // Three states, expressed as the two flags the picker already understands.
+        ...(options.status ? { showWorking: options.status !== 'live' } : {}),
+        ...(options.status ? { showArchived: options.status === 'archived' } : {}),
+        ...(options.sort ? { sortByDesc: options.sort.direction === 'desc' } : {})
```
oidacra: Only `sort.direction` is read here — is `sort.field` intentionally dropped?

`DotAssetPickerBrowseOptions` has no sort-field property, so there is currently no way to pass it through even if a caller sets it. That makes the drop structural rather than an oversight at this line.

The public JSDoc on `openBrowserModal` advertises the full shape:

```
sort: { field: 'modDate', direction: 'desc' },
```

so a template author passing `sort: { field: 'title', direction: 'desc' }` would reasonably expect title ordering and instead get `modDate` from `DEFAULT_ASSET_PICKER_SORT`. Both shipped VTL templates happen to pass `modDate`, which is already the default, so this stays invisible today.

`extensions` right above carries an explicit "not yet, ships with the endpoint support" note. Would it be worth giving `sort.field` the same treatment — either wiring it through to the store's sort state, or documenting it as direction-only for now?

core-web/libs/ui/src/lib/components/dot-asset-picker/store/features/with-asset-browse.feature.ts
```diff
@@ -173,7 +197,14 @@ export function withAssetBrowse() {
...
                                 pages[page] = {
                                     contentCursor: response.nextContentCursor,
-                                    hasMoreContent: response.hasMoreContent
+                                    hasMoreContent: response.hasMoreContent,
+                                    folderCursor: response.nextFolderCursor,
+                                    hasMoreFolders: response.hasMoreFolders,
+                                    // Only a `showLinks` response carries these. Defaulting the
+                                    // cursor to the one we sent keeps it from rewinding to 0 on a
+                                    // page that did not query links.
+                                    linkCursor: response.nextLinkCursor ?? 0,
```
oidacra: The comment and the code seem to disagree here. The comment says the fallback keeps the cursor "from rewinding to 0" by defaulting to "the one we sent", but the code is `?? 0` — which is the rewind.

Preserving the last-sent cursor would be `?? bookmark?.linkCursor ?? 0`.

Does this matter in practice? Both `hasMoreLinks` and `nextLinkCursor` are optional on the response type, so nothing enforces that they arrive together. If a response ever reports `hasMoreLinks: true` without a `nextLinkCursor`, this stores `linkCursor: 0` while `hasMoreLinks` stays true — the next request then re-requests links from the start while content and folder cursors keep advancing, which would surface as duplicate link rows on a later page.

I could not confirm the backend ever does that, so this may be purely a comment fix. But the one change satisfies both readings:

```ts
linkCursor: response.hasMoreLinks
    ? (response.nextLinkCursor ?? bookmark?.linkCursor ?? 0)
    : 0,
```

(For contrast, `folderCursor` on the line above needs no fallback — `nextFolderCursor` and `hasMoreFolders` are both non-optional, so that asymmetry looks deliberate and correct.)

core-web/libs/edit-content-bridge/src/lib/bridges/angular-form-bridge.ts
```diff
@@ -470,105 +501,220 @@ export class AngularFormBridge implements FormBridge {
...
+                            finish(item ? toSelection(item) : null);
+                        });
                     });
-                } else {
-                    options.onClose(null);
-                }
+                },
+                // Nothing to browse and nothing useful to say beyond that — same as the File field,
+                // which simply does not open when no site resolves.
+                error: () => finish(null)
```
oidacra: Would it be worth logging here before resolving null?

`getCurrentSite()` is a plain `HttpClient.get()` with no `catchError` upstream, so this branch is what runs on a 5xx, a network failure, or an expired session. As written, the template author's `onClose(null)` fires and that is indistinguishable from the user pressing Cancel — with nothing in the console and no signal for whoever is debugging a "the browse button does nothing" report.

The comment says there is "nothing useful to say beyond that", which is fair for the user-facing side, but the developer-facing side currently gets nothing at all:

```ts
error: (err) => {
    console.error('DotCustomFieldApi.openBrowserModal: failed to resolve the current site', err);
    finish(null);
}
```

The `!site` branch above has the same property, though that one is at least a expected-shape outcome rather than a failure.

core-web/libs/ui/src/lib/components/dot-asset-picker/dot-asset-picker.component.ts
```diff
@@ -280,15 +292,21 @@ export class DotAssetPickerComponent implements OnInit {
...
         }
 
-        const lastSlash = asset.url?.lastIndexOf('/') ?? -1;
-        const path = lastSlash >= 0 ? asset.url.slice(0, lastSlash + 1) : this.store.path();
+        // A folder IS a location — its own path, not the folder its url sits in.
+        if (isFolder(item)) {
+            return { siteId: site.identifier, hostname: site.hostname, path: item.path };
+        }
+
+        const url = (item as DotCMSContentlet).url;
```
oidacra: Is this cast still accurate? After `isFolder(item)` returns false on the line above, the narrowed type is `DotCMSContentlet | DotContentDriveLink` — and a link genuinely reaches here, via the `!isContentlet(asset)` early-return in `confirm()` which calls this function with link rows.

Both union members declare `url: string`, so the cast also looks unnecessary — `item.url` should compile directly on the union:

```ts
const url = item.url;
```

Dropping it would mean a future browse-item variant without `.url` fails at compile time rather than silently reading `undefined` here.

core-web/libs/ui/src/lib/components/dot-asset-picker/asset-picker-config.spec.ts
```diff
@@ -191,4 +191,100 @@ describe('buildAssetPickerConfig', () => {
...
+        });
+
+        it('should carry the browse options through untouched', () => {
+            const browse = {
+                showFolders: true,
+                showLinks: true,
+                showWorking: false,
+                showArchived: false,
+                sortByDesc: true,
+                extensions: ['jpg']
```
oidacra: `extensions` is not a field on `DotAssetPickerBrowseOptions` — should it be in this fixture?

The interface defines five fields (`showFolders`, `showLinks`, `showWorking`, `showArchived`, `sortByDesc`), and this PR's own spec.md documents `extensions` as deliberately not exposed yet because the endpoint has no such parameter. The bridge-side comment makes the same point ("No `extensions` here on purpose").

The assertion still passes because excess-property checking does not apply to a variable (only to an inline object literal), and `toEqual` compares the same object against itself. So the test would keep passing even if `browse` pass-through broke for the real fields.

Worth dropping `extensions: ['jpg']` so the fixture matches the contract as shipped? Otherwise the next reader could reasonably conclude it is a supported browse option today.

core-web/libs/edit-content-bridge/src/lib/bridges/angular-form-bridge.ts
```diff
@@ -470,105 +501,220 @@ export class AngularFormBridge implements FormBridge {
...
+        return {
+            ...base,
+            baseType: row['baseType'],
+            contentType: row['contentType']
+        } as DotBrowserSelection;
+    }
+
+    return {
+        ...base,
+        name: row['name'] ?? row['fileName'],
```
oidacra: These four values are typed `unknown` (from `Record<string, unknown>`), and the trailing `as DotBrowserSelection` is what makes the object literal type-check rather than any actual narrowing.

The `base` fields above all go through `String(...)`, so they are guaranteed strings. These do not, yet `DotBrowserSelection` declares them as `string | undefined` — so a malformed row handing back a number or an object for `baseType` would flow through to consumers that read `.baseType` as a string with full confidence.

Would a small narrowing helper be worth it here?

```ts
const asString = (v: unknown): string | undefined =>
    typeof v === 'string' ? v : undefined;
```

then `name: asString(row['name']) ?? asString(row['fileName'])`, and likewise for the other three — which would also let the `as DotBrowserSelection` cast go away.

core-web/libs/ui/src/lib/components/dot-asset-picker/dot-asset-picker.component.spec.ts
```diff
@@ -121,7 +142,7 @@ const createMockStore = () => {
         folders: signal([]),
         foldersStatus: signal(ComponentStatus.LOADED),
         selectedNode: signal<{ data: unknown } | undefined>(undefined),
-        selectedAsset: signal<DotCMSContentlet | null>(null),
+        selectedAsset: signal<DotContentDriveItem | null>(null),
```
oidacra: Should this mock match the widened store contract? The real `DotAssetPickerSelectionState.selectedAsset` became `DotContentDriveBrowseItem | null` in this PR, which includes `DotContentDriveLink`; the mock is still `DotContentDriveItem | null`, which does not.

That is what forces the `as unknown as DotContentDriveItem` double-cast on `SELECTED_FOLDER` and `SELECTED_LINK` above — and a double-cast means the type-checker can no longer catch a drift between these fixtures and what the store actually accepts.

Typing the signal as `signal<DotContentDriveBrowseItem | null>(null)` should let both fixtures use a single direct cast (or none, for the link, which already matches the shape).

core-web/libs/ui/src/lib/components/dot-asset-picker/store/features/with-asset-browse.feature.ts
```diff
@@ -127,15 +138,28 @@ export function withAssetBrowse() {
...
-                        archived: false,
-                        // Invariant, not a default: the picker's list is for assets. Folders are
-                        // navigated through the sidebar tree.
-                        showFolders: false
+                        // Version state. Not a flat `false` any more: `openBrowserModal` callers can
+                        // ask for live-only or for archived content. Absent `browse`, this is
+                        // exactly what it always was.
+                        archived: browse?.showArchived ?? false,
+                        // `live: true` means published-only. The endpoint's own default is `false`
+                        // (working included), which is what every entry point but `browse` wants, so
```
oidacra: Small wording nit: should this say `openBrowserModal` rather than `browse`?

Every sibling comment in this file draws the same contrast that way — line 118 ("every entry point but `openBrowserModal` uses") and line 143 ("every entry point but `openBrowserModal`"). Here it reads "every entry point but `browse`", and `browse` is the name of the options object (`pickerConfig?.browse`) rather than an entry point, so it briefly scans as a different thing.

dario-daza (dismissed):

nicobytes:
QA: 

https://github.com/user-attachments/assets/f6444abf-4b09-473a-af7a-c866b287bf55

nicobytes:
### Re: rollback-safety finding (H-8)

The finding rests on the shipped custom-field VTL being **copied into the field definition** in the DB. For these two fields it isn't — what's stored is a path reference:

```java
// VanityUrlContentType.java:83
.values("$velutil.mergeTemplate('/static/content/file_browser_field_render.vtl')")

// PageContentType.java:171
.values("$velutil.mergeTemplate('/static/htmlpage_assets/redirect_custom_field.vtl')")
```

That path points at the **dispatcher**, which `#parse`s `_new` or `_old` at render time based on `$structures.isNewEditModeEnabled()`. The template body never lands in `field.values`.

So on rollback to N-1:

| | |
|---|---|
| `field.values` | unchanged — it is a path, not a call |
| the `.vtl` file | reverts with the deployment (it ships in the WAR, under `src/main/webapp`) |
| the Angular bundle | reverts |

The template and the bridge revert **together**: an N-1 build serves the N-1 template, which calls the N-1 shape against the N-1 bridge. There is no window where a new-shape call meets an old-shape implementation through this path.

**Residual risk**, much narrower than described: someone who manually pasted the *body* of `_new.vtl` into their own custom field while N was live. That is the same exposure as any customer-authored template, and it is bounded by the same mitigation already noted in the finding — `openBrowserModal` is only reachable from the new Edit Content, which is not the default editor. There are no consumers of this API today.

**Not taking the suggested compatibility mapping.** Reading legacy `header`/`params` keys would add permanent back-compat handling to an API that has never shipped to a consumer, to cover a case the reference indirection already handles. The redesign deliberately avoided carrying that shape forward.

The H-8 category check is worth having — the mechanism just does not apply to these two fields, because what is stored is a reference rather than a copy.

oidacra (approved):

PR #37284, opened 2026-08-28
refactor(ui): redesign the shared key/value field across its three consumers (#37191)

### LINKED ISSUE #37191: Refactor shared Key/Value field to new design across edit content, field variables, and apps portlet
### Description

Refactor the **Key/Value field** component in the new Edit Content to follow the new design, and **roll the new design out to all consumers** of this shared component.

🎨 **Design:** https://claude.ai/design/p/8d442fe1-3266-45ca-a36e-3430a1aad5cb?file=Key+Value+Field.dc.html&via=share
💬 **Context (Slack):** https://dotcms.slack.com/archives/C0893BF7GRJ/p1786378691544909

**⚠️ Shared component — reused in 3 places:**

| # | Consumer | Where |
|---|---|---|
| 1 | **New Edit Content** key/value field | Editing a contentlet |
| 2 | **Field Variables** | Editing a field in the Content Type portlet |
| 3 | **Apps portlet options** | App configuration extra options (e.g. the **Samwell** app's extra options) |

The refactored design must be applied **consistently across all three** consumers, and each must be verified to work after the refactor.

### Acceptance Criteria

**Design refactor**
- [ ] The Key/Value component is refactored to match the new **design** (linked above)
- [ ] Add/edit/remove key-value pairs works with the new design
- [ ] Validation and existing behavior (e.g. duplicate keys, empty values) are preserved

**All consumers get the new design (no regressions)**
- [ ] The **new Edit Content** key/value field uses the new design and reads/writes values correctly
- [ ] The **Field Variables** editor (Content Type portlet) uses the new design and reads/writes values correctly
- [ ] The **Apps portlet options** (e.g. Samwell app extra options) use the new design and read/write values correctly
- [ ] Existing key/value data in each consumer continues to load and save without loss

**Consistency**
- [ ] The component looks and behaves consistently across all three consumers (single shared component, one design)

### Suggested Tests

Frontend refactor of a **shared component** (no backend/endpoint change). **Unit-first** for the component itself; **E2E** carries extra weight here because the main risk is regressions across the **three consumers**. Integration/Postman are not applicable.

**Unit (Jest/Spectator)** — primary coverage of the component
- [ ] Renders the new design (key + value rows, add/remove controls)
- [ ] Adding a key/value pair emits the updated model
- [ ] Editing a key or value emits the updated model
- [ ] Removing a pair emits the updated model
- [ ] Existing validation is preserved (e.g. duplicate keys, empty key/value handling)
- [ ] Loads and renders pre-existing key/value data correctly (round-trips input → output without loss)

**E2E (Playwright)** — one smoke per consumer (regression guard is the point)
- [ ] **Edit Content**: open a contentlet with a key/value field → add/edit/remove a pair → save → values persist
- [ ] **Field Variables**: in the Content Type portlet, edit a field's variables → add/edit/remove → save → persist
- [ ] **Apps portlet**: open an app with key/value extra options (e.g. Samwell) → add/edit/remove → save → persist

**Integration / Postman**
- [ ] No tests — the refactor is presentation/behavior on a shared UI component; each consumer saves through its **existing** endpoints with no contract change.

> The three E2E smokes are justified specifically because this is a shared component with three distinct integration points; the per-interaction logic is otherwise covered by unit tests.

### Priority

Medium

### Additional Context

Part of the new Edit Contentlet work (`dotCMS: New Edit Contentlet`). Because the Key/Value component is shared by the Content Type portlet's Field Variables and the Apps portlet options, the refactor must be validated in all three places to avoid regressions. Product context is in the linked Slack thread. Scope decision: apply the new design to **all three consumers** (not edit content alone).

adrianjm-dotCMS:
Fixes #37191

### Proposed Changes

Reworks the one shared Key/Value editor in `libs/ui` to the new design and rolls it out to all three
of its consumers — the **Edit Content field**, the **Content Type Field Variables** tab, and the
**Apps custom-properties** panel — so they finally look and behave the same.

Built from stock PrimeNG 21. **No new component enters the workspace**; the net component count for
this feature is zero.

**What changed for the user**

* Values render as plain text at rest and become an input when activated — Enter commits, Escape restores.
* Row actions (drag handle, remove) are revealed on hover instead of competing for attention on every row.
* Hidden values move from a toggle switch in its own column to an **eye inside the value field**. It is a button in both states, so hiding a value is reversible — previously it was a one-way trip.
* The hidden indicator is never hover-gated: which values are secret has to be readable at a glance.
* Empty state gains an icon and reuses the existing copy, which already matched the design.
* Reordering is offered in all three consumers rather than only one.

**Three findings worth a reviewer's attention**

* **The missing table styling was a selector problem, not a CSS one.** PrimeNG themes its table with direct-child combinators (`.p-datatable-tbody > tr > td`). The row components used element selectors with `display: contents`, so the component element sat between the `tbody` and the `tr` and **no table style from the theme ever applied** — no borders, no header band. They now attach by attribute selector (`tr[dotKeyValueTableRow]`), making the host the `tr` itself.
* **`onRowReorder` must not re-apply the move.** PrimeNG's `onRowDrop` already calls `reorderArray` on the array bound to `[value]`. Recomputing from `dragIndex`/`dropIndex` applies it twice. The handler only re-publishes the array with a fresh reference so the signal notifies.
* **Hover-hidden actions use `opacity-0`, never `display:none` or an `@if`.** Those remove the control from the tab order and strand keyboard and touch users with no visible symptom for anyone on a mouse. A test asserts the mechanism, not merely that the icon is invisible.

**Removed along the way**

* `DotKeyValueUtil` — 87 lines nothing imported.
* A stale `[autoFocus]` binding in the Apps template, plus the spec stub that made it look real. It compiled only because `strictTemplates` is off in that app.
* A duplicated copy of the row template that the tests were asserting against instead of the real one.
* The `dragAndDrop` input, once reordering became universal and it had nothing left to gate.
* Both SCSS files — styling is Tailwind against PrimeNG theme tokens, so no colour is hardcoded.

### Checklist

- [x] Tests
- [x] Translations — none needed; the design's copy already existed as `keyValue.value_no_rows.label`
- [x] Security Implications Contemplated — value masking is a **UI concern only, not a security boundary**. It keeps a secret off a shoulder-surfer's screen; the underlying value is transmitted exactly as before. Stated in the spec so nobody mistakes it for protection.

### Additional Info

**Tests**: 61 in `libs/ui` (from 31), 24 in `edit-content`, 70 in `dotcms-ui`, plus Playwright smokes for each consumer. Lint and build clean.

**Not yet run**: the Playwright specs have a clean typecheck but have **not been executed**, and the manual keyboard-only pass is still pending. Flagging it rather than implying green.

**Scope**: frontend only. Order *persistence* for Field Variables and Apps was scoped out of this issue — it would mean database, REST and encrypted-storage work.

`spec.md`, `data-model.md` and `contracts/` are included as the reviewed contract; the Spec-Kit process artifacts are gitignored by design.

### Screenshots

This PR fixes: #37191


erickgonzalez (dismissed):
Spec looks good! Let's continue with the implementation.

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-table-row/dot-key-value-table-row.component.html
```diff
@@ -1,55 +1,126 @@
...
+@let readOnly = $readOnly();
+@let isEditing = $isEditing();
+@let isEditingKey = $isEditingKey();
 
-<tr class="dot-key-value-table-row" [pReorderableRow]="index">
-    @if ($dragAndDrop()) {
-        <td class="p-2 align-middle">
-            <span class="pi pi-bars text-gray-500" pReorderableRowHandle></span>
-        </td>
+<td class="kv-gutter w-[30px] min-w-[30px]">
```
oidacra: This `<td>` (added by this PR) renders unconditionally, while the delete `<td>` below at line 105 is wrapped in `@if (!readOnly)`. Combined with the unconditional `[pReorderableRow]` in the parent template and `onRowReorder()` having no read-only guard, a field passed `[readOnly]="true"` can still be drag-reordered and will emit `updatedList`.

For Edit Content this does not reach the saved payload, since Angular excludes disabled controls from `form.value`, so nothing is corrupted. But the field still accepts a mutating gesture it advertises as unavailable, and the `$readOnly` JSDoc says reordering is disabled while the `$colspan` comment says it stays.

Was reordering meant to survive read-only? If yes, the JSDoc needs correcting; if no, gating this `<td>` and adding an early return in `onRowReorder()` would settle it.

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-ng.component.ts
```diff
@@ -25,133 +50,197 @@ export interface DotKeyValue {
...
-     * The list of key-value pairs to be displayed and manipulated
-     */
     $variables = input<DotKeyValue[]>([], { alias: 'variables' });
 
     /**
-     * Controls whether drag and drop functionality is enabled for reordering items
+     * Renders the list without any way to change it.
+     *
+     * For fields whose pairs are produced elsewhere — a file asset's metadata is
+     * regenerated on every save — where adding, editing, reordering or clearing would
```
oidacra, line 64: This JSDoc lists reordering among what read-only removes, but `$colspan` at line 104 says the opposite ("reordering stays, so the drag column does too"), and the drag handle renders unconditionally with no guard in `onRowReorder()`. Both comments are added by this PR.

Which is the intended behavior? Whichever it is, one of the two comments is going to mislead the next reader.

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-ng.component.ts
```diff
@@ -25,133 +50,197 @@ export interface DotKeyValue {
...
-     * Updates the variable in the local list and emits events for parent components
+     * Adds a whole pasted block at once.
+     *
+     * The block keeps its own order and goes on top as one group, so pasting
+     * `A,B,C` reads `A,B,C` and not reversed — which is what adding them one by one
+     * through {@link saveVariable} would produce.
      *
-     * @param {DotKeyValue} variable - The updated key-value pair
-     * @param {number} index - The index of the variable to update in the array
+     * `save` is emitted per pair on purpose: Field Variables persists row by row
```
oidacra: This comment (added by this PR) says `save` is emitted per pair because "Field Variables persists row by row through that output" — but this same PR changes Field Variables to defer everything to `saveChanges()`, and its new class doc says "nothing leaves this component now until `saveChanges` runs". The old `updateFieldVariable` / `deleteFieldVariable` methods that wrote per row were deleted here.

Is there still a consumer that persists per pair? If not, would it be worth restating the reason as output-compatibility rather than per-row persistence, so the justification does not point at behavior that no longer exists?

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-ng.component.ts
```diff
@@ -25,133 +50,197 @@ export interface DotKeyValue {
...
-            return [...variables];
+    /**
+     * Empties the list, behind a confirmation.
+     *
+     * Confirmed through `ConfirmationService`, so the dialog each host already renders
+     * is the one that appears — this component brings none of its own. Its options
+     * match the unsaved-changes prompt in Edit Content: a header and no icons. The
+     * reject is a plain text button, as `dot-tags-list` does.
+     *
+     * `delete` is emitted per pair because Field Variables persists row by row through
```
oidacra: Same as the note on `saveVariables` above: this comment (added by this PR) justifies the per-pair `delete` emission with "Field Variables persists row by row through it", which this PR removed. Worth updating alongside it?

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-paste.util.ts
```diff
@@ -0,0 +1,75 @@
...
+        if (!match) {
+            continue;
+        }
+
+        const [, key, rawValue] = match;
+
+        // A key already on screen, or repeated within the paste itself. Skipped rather
+        // than overwritten — a paste should not quietly replace a value already set.
+        if (existingKeys[key] || seen.has(key)) {
+            continue;
```
oidacra, line 67: There are three silent `continue`s in this loop (added by this PR): a line failing the `ASSIGNMENT` regex at 53 and 59, and here for a key that already exists or repeats within the paste.

In `handleKeyInputPaste`, if no pair parses at all the handler skips `preventDefault()` and the browser does a normal paste, which is a good fallback. But if at least one pair parses, `preventDefault()` fires and only the parsed pairs are committed — so pasting `FOO=bar` followed by a line reading `justText` silently drops the second line.

The manual add-row path shows an inline duplicate error and keeps the text visible for correction. Should the paste path report skipped lines too, or is silent dropping the intended behavior for a bulk paste?

core-web/libs/edit-content/src/lib/fields/dot-edit-content-key-value/components/key-value-field/key-value-field.component.ts
```diff
@@ -40,26 +41,57 @@ export class DotKeyValueFieldComponent extends BaseControlValueAccessor<
     }
 
     /**
-     * Updates the field.
-     * It is used to update the field.
+     * Reports the pairs to the form as JSON **text**, assembled from the array
```
oidacra, line 52: Anchoring here because the line I mean is unchanged: the class declaration at line 24, `extends BaseControlValueAccessor<Record<string, string | null>>`, which this PR did not update.

This PR widens `parseToDotKeyValue` and `handleChangeValue` to accept `string | Record<string, string | null> | DotKeyValue[]`, and the doc comment states the control carries JSON text. But the class generic still declares the control's value as a plain object, so `$value` and `writeValue` statically promise a type the control no longer holds, and the string branch is unreachable through the typed `handleChangeValue(this.$value)` call.

Would widening the generic (or naming the union) be worth it, so a future edit gets a compiler signal that a raw string can arrive?

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-table-row/dot-key-value-table-row.component.html
```diff
@@ -1,55 +1,126 @@
...
+    <span
+        class="inline-flex cursor-grab text-surface-500 opacity-0 transition-opacity group-focus-within:opacity-100 group-hover:opacity-100 focus-visible:opacity-100 dark:text-surface-400"
+        [attr.aria-label]="'Reorder' | dm"
```
oidacra: `'Reorder' | dm` passes a literal English string to the translation pipe. The same applies to `'Delete' | dm` at line 109 here and `'add' | dm` at line 77 of the header-row template — all added by this PR.

Every other label in these files uses a bundle key (`'keyValue.hidden_header.label' | dm` sits at line 54 of the same header template). Since `DotMessageService` echoes the key when no translation exists, these three render English in every locale rather than failing loudly.

Should these get real keys added to the bundle?

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-table-row/dot-key-value-table-row.component.html
```diff
@@ -1,55 +1,126 @@
...
+            autocomplete="false"
+            data-testId="dot-key-value-key-input"
+            pInputText
+            pSize="small"
+            type="text" />
+    } @else {
+        <!-- `title` keeps the whole key reachable once the column clips it. -->
+        <span
+            (click)="startEditKey()"
+            (keydown.enter)="startEditKey()"
```
oidacra: This span is exposed as `role="button"` with `tabindex="0"` (added by this PR) but only binds `(keydown.enter)`. The value span at line 93 does the same.

Per the ARIA button pattern, Space should also activate it — and because this is not a native button, Space will scroll the page instead.

Would adding `(keydown.space)` with `preventDefault()` work here, or would making it a real `<button>` styled as text be cleaner?

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-table-row/dot-key-value-table-row.component.html
```diff
@@ -1,55 +1,126 @@
...
-            <span class="pi pi-bars text-gray-500" pReorderableRowHandle></span>
-        </td>
+<td class="kv-gutter w-[30px] min-w-[30px]">
+    <span
+        class="inline-flex cursor-grab text-surface-500 opacity-0 transition-opacity group-focus-within:opacity-100 group-hover:opacity-100 focus-visible:opacity-100 dark:text-surface-400"
+        [attr.aria-label]="'Reorder' | dm"
+        pReorderableRowHandle
+        role="button"
+        tabindex="0"
+        data-testId="dot-key-value-drag-handle">
```
oidacra: This handle gets `role="button"` and `tabindex="0"` (added by this PR), so it takes focus and announces itself as actionable, but `pReorderableRowHandle` only drives PrimeNG's pointer-based drag — there is no keydown handler, so Enter and Space do nothing.

That leaves a focus stop a screen-reader or keyboard user can reach but cannot operate, which is arguably worse than not being focusable at all.

Is a keyboard reorder (arrow keys while focused) in scope here, or should the role and tabindex come off until it is? Related: the PR notes the manual keyboard-only pass is still pending, and this is the kind of thing it would surface.

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-table-row/dot-key-value-table-row.component.spec.ts
```diff
@@ -1,286 +1,394 @@
...
+            spectator.component.editControl.setValue('   ');
+            spectator.component.commitEdit();
 
-            it('should show the drag handle when dragAndDrop is true', () => {
-                const dragHandle = spectator.query('.pi-bars');
-                expect(dragHandle).toBeTruthy();
+            expect(saveSpy).not.toHaveBeenCalled();
+        });
+
+        it('should not collide with itself when only whitespace is trimmed', () => {
```
oidacra: This test (added by this PR) sets `forbiddenkeys: { name: true }`, edits the key to `'  name  '`, and asserts `save` was not called — but I do not think it exercises the self-exclusion logic it describes.

In `commitEdit`, `unchanged = text === #textBeforeEdit` is computed at line 142 and `if (unchanged || rejected) return;` fires before the dedupe runs. The trimmed text equals `#textBeforeEdit`, so the early return happens and `#isTaken` is never reached — the assertion would pass whether or not self-exclusion works.

Following from that, `#isTaken`'s `key !== $variable().key` guard at line 175 looks unreachable through the public API, since `#textBeforeEdit` is always seeded from the row's own key.

Is that branch worth keeping, and could the test be reshaped to actually reach it?

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-table-header-row/dot-key-value-table-header-row.component.spec.ts
```diff
@@ -1,244 +1,261 @@
...
-            expect(resetSpy).toHaveBeenCalled();
+            expect(spectator.query(byTestId('dot-key-value-new-visibility-toggle'))).toBeTruthy();
+            expect(spectator.query(byTestId('hidden-switch'))).toBeFalsy();
         });
-    });
 
-    describe('Drag and Drop', () => {
-        it('should not show drag handle column when dragAndDrop is false', () => {
-            spectator.setInput('dragAndDrop', false);
+        it('should switch the value field to a password when the toggle is used', () => {
```
oidacra, line 195: This test (added by this PR) clicks the visibility toggle once and confirms the field becomes a password.

Given the PR's headline change is that hiding a value is now reversible where it used to be a one-way trip, would a second click asserting it returns to `type="text"` (and that the icon and `aria-pressed` flip back) be worth adding? The previous spec's hidden-fields block was removed, so the reversal direction currently has no coverage anywhere.

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-table-row/dot-key-value-table-row.component.spec.ts
```diff
@@ -1,286 +1,394 @@
...
 
-    beforeEach(() => {
-        spectator = createComponent({
-            props: {
-                showHiddenField: false,
-                variable: mockVariable,
-                index: 0,
-                dragAndDrop: false
-            } as unknown
+    describe('click-to-edit (FR-006, FR-007)', () => {
```
oidacra, line 94: This click-to-edit block (added by this PR) covers the key side thoroughly, including "should refuse an empty key" at line 292, but there is no value-side counterpart for clearing a value to empty and committing it.

Worth noting the production behavior: `editControl` at line 76 has no validators, and `rejected` at line 143 is computed only for `cell === 'key'`, so committing an emptied value emits `save` with `value: ''`. That looks pre-existing rather than introduced here — the previous `valueChanges` auto-save did not check validity either — but the new explicit Enter/blur commit makes it considerably easier to reach.

Is an empty value intentionally allowed? Either way, a test pinning the expected behavior would stop it drifting.

core-web/apps/dotcms-ui/src/app/portlets/shared/dot-content-types-edit/components/fields/dot-edit-field-dialog/dot-edit-field-dialog.component.spec.ts
```diff
@@ -116,6 +116,9 @@ class DotCustomFieldSettingsStubComponent {
 class DotContentTypeFieldsVariablesStubComponent {
     field = input<DotCMSContentTypeField>();
     showTable = input(false);
+    // Variables hands the dialog its own Save now, like the Settings tabs do.
+    $changeControls = output<DotDialogActions>();
```
oidacra, line 120: These stub outputs (`$changeControls` here, and `$save`) are added by this PR so the Variables tab can hand the dialog its own footer, but I do not see a test that drives an emission from either one — the only new assertion is `hideButtons` becoming false at line 369.

The Settings-tab equivalent does test this, driving `changesDialogActions` and asserting the dialog's `saveBtn` is replaced and then restored. Would an analogous test for the Variables tab be worth adding, so the subscription is actually verified rather than just declared?

core-web/libs/edit-content/src/lib/utils/key-value-order.util.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+    describe('degrading safely', () => {
+        it('should return the contentlet untouched when the ordered parse is unusable', () => {
+            const contentlet = { keyValue: { a: '1' } };
+
+            expect(attachOrderedFields(contentlet, null)).toBe(contentlet);
+            expect(attachOrderedFields(contentlet, 'not an object')).toBe(contentlet);
+            expect(orderedKeyValueText(contentlet, 'keyValue')).toBeNull();
+        });
+
+        it('should decline the recovered order when the keys do not match', () => {
```
oidacra, line 167: This block (added by this PR) is strong overall — the degrading-safely cases and the NUL-prefix collision test are the right instincts for a parser this load-bearing. Two directions look untested though.

This mismatch case covers a key present in the plain parse but missing from the recovered order. The reverse — a key in the recovered order that is absent from the plain parse — is not exercised, so if the guard compares counts rather than key sets, that direction could pass a mis-paired result instead of degrading safely.

Separately, duplicate keys in the raw text (for example an object literal repeating `"a"` twice) are not covered. `JSON.parse` dedupes to the last occurrence, but an order-preserving text parser may count both — and the NUL-prefix test suggests the implementation is count-sensitive.

Would both be worth adding?

core-web/apps/dotcms-ui/src/app/portlets/shared/dot-content-types-edit/components/fields/dot-content-type-fields-variables/dot-content-type-fields-variables.component.ts
```diff
@@ -90,82 +132,176 @@ export class DotContentTypeFieldsVariablesComponent implements OnChanges, OnDest
...
+        const operations = [
+            ...removed.map((variable) =>
+                track(this.fieldVariablesService.delete(this.field, variable), () =>
+                    succeeded.push(variable)
+                )
+            ),
+            ...written.map((variable) =>
+                track(this.fieldVariablesService.save(this.field, variable), () =>
+                    succeeded.push(variable)
+                )
```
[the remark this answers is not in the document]
adrianjm-dotCMS: Correct, and the gap was wider than described: `#storedAfter` rebuilt the snapshot from what was on screen, so even capturing the response would not have carried the `id` through.

Fixed on both ends in 7589c65 — the save branch records `{ ...variable, ...saved }`, and `landed` became a `Map` so an operation that succeeded is taken from the response instead of from the row. Covered by a test that fails against the previous code.

core-web/libs/edit-content/src/lib/utils/key-value-order.util.ts
```diff
@@ -0,0 +1,170 @@
...
+const stripPrefix = (key: string): string =>
+    key.startsWith(NUMERIC_KEY_PREFIX) ? key.slice(NUMERIC_KEY_PREFIX.length) : key;
+
+/**
+ * Where the order-preserving parse of the response is parked on the contentlet.
+ *
+ * A separate namespace, so every existing property is byte-for-byte what the normal
+ * parse produced and no consumer of `getContentById` sees a changed shape.
+ */
+export const ORDERED_FIELDS = '__dotOrderedFields' as const;
```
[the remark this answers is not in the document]
adrianjm-dotCMS: Leaving this one as it is.

The collision needs a field whose variable name is literally `__dotOrderedFields`. Velocity variable names are generated from the label and start with a letter, so only a hand-crafted one through the API could reach it. A `Symbol` would work — the object is never serialised — but it is hardening against something no supported path can produce, and the current key is greppable in a way a `Symbol` is not.

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-paste.util.ts
```diff
@@ -0,0 +1,75 @@
...
+        }
+
+        seen.add(key);
+        pairs.push({ key, value: unquote(rawValue) });
```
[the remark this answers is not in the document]
adrianjm-dotCMS: Agreed — the paste was the only way a blank value could enter.

It is now dropped rather than added, and by the same measure the rows use: `value.trim()`. `EMPTY="   "` survived `unquote` as three spaces and would otherwise have slipped through where a bare `EMPTY=` was turned away. Recorded as FR-045a.

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-table-row/dot-key-value-table-row.component.html
```diff
@@ -1,55 +1,126 @@
...
-        <td class="p-2 align-middle">
-            <span class="pi pi-bars text-gray-500" pReorderableRowHandle></span>
-        </td>
+<td class="kv-gutter w-[30px] min-w-[30px]">
+    <span
+        class="inline-flex cursor-grab text-surface-500 opacity-0 transition-opacity group-focus-within:opacity-100 group-hover:opacity-100 focus-visible:opacity-100 dark:text-surface-400"
+        [attr.aria-label]="'Reorder' | dm"
+        pReorderableRowHandle
+        role="button"
+        tabindex="0"
```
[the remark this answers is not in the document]
adrianjm-dotCMS: Right, and the conclusion is to stop advertising it.

Keyboard reordering is not in this scope, so `role` and `tabindex` are gone from the handle: PrimeNG binds `mousedown` and the HTML drag events only, and a focus stop announcing a button that cannot be operated is worse than not being focusable at all. FR-019 now says reordering is pointer-only rather than implying the opposite.

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-table-row/dot-key-value-table-row.component.html
```diff
@@ -1,55 +1,126 @@
...
-            @if (valueControl.value !== passwordPlaceholder && !variable.hidden) {
-                <p-toggleSwitch
-                    [formControl]="hiddenControl"
-                    data-testId="dot-key-value-hidden-switch" />
-            }
-        </td>
+        <!-- `title` keeps the whole value reachable once the column clips it. -->
+        <span
+            (click)="startEdit()"
+            (keydown.enter)="startEdit()"
```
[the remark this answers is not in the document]
adrianjm-dotCMS: Fixed. Both spans now handle `(keydown.space)` with `preventDefault()`, so Space activates the editor instead of scrolling the page.

specs/37191-key-value-field-redesign/data-model.md
```diff
@@ -0,0 +1,84 @@
...
+---
+
+## Entity: `DotKeyValue`
+
+Declared in `libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-ng.component.ts` and
+re-exported from `@dotcms/ui`.
+
+| Field | Type | Required | Notes |
+|---|---|---|---|
+| `key` | `string` | yes | Unique within one editor instance. Non-empty. **Immutable after creation** (FR-008) |
```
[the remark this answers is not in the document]
adrianjm-dotCMS: Correct — left over from before FR-008 was reversed. The table now says the key is editable in place, and the validation rules were rewritten to match what the row actually enforces.

specs/37191-key-value-field-redesign/spec.md
```diff
@@ -0,0 +1,506 @@
...
+- **FR-045**: Text that yields no assignment MUST paste normally into the input, so pasting a plain key
+  still behaves like a paste.
+- **FR-046**: Multi-line values are out of scope. A certificate pasted this way would be cut at its
+  first line break, so it stays a single-pair job through the normal inputs.
+
+**Clearing every pair**
+
+- **FR-047**: The editor MUST offer a control that removes every pair at once, behind a confirmation.
+- **FR-048**: The confirmation MUST use the host's existing dialog rather than one of the editor's own,
+  and MUST match the styling of the unsaved-changes prompt in Edit Content.
```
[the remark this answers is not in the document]
adrianjm-dotCMS: Correct. FR-048 now describes the keyed dialog the editor renders itself, and says why: only Edit Content has a dialog always mounted, while dotcms-ui's exists only while `DotAlertConfirmService` holds a model — so an unkeyed confirmation reached nothing and Clear All did nothing in two of the three consumers.

Kept as a marked correction rather than a silent rewrite, so the change stays auditable.

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-table-row/dot-key-value-table-row.component.html
adrianjm-dotCMS: You are right that the two comments contradicted each other, and the intended behaviour is the one in the `$readOnly` doc: read-only does not reorder. A reorder is a change like any other, and a list the consumer was told cannot be edited should not publish one.

Three things now: the row is bound `[pReorderableRowDisabled]`, the handle is not rendered, and `onRowReorder()` returns early.

One deviation from your suggestion — the `<td>` itself stays. The header renders its gutter `<th>` unconditionally, so dropping the cell would leave read-only rows a column short. Only the handle inside it goes.

FR-051 never mentioned reordering at all, which is where the ambiguity came from. It does now.

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-ng.component.ts
adrianjm-dotCMS: Answered in full on the row template, alongside the `<td>` note: read-only does not reorder, and the `$colspan` comment was the wrong one. Both comments have been rewritten so they agree.

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-ng.component.ts
adrianjm-dotCMS: Correct — that justification described behaviour this same PR removed.

Reworded to the real reason: `save` is a per-pair output, and a consumer acting on it has to see every pair a block paste added rather than just the last. That is about the output's contract, not about who persists what.

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-ng.component.ts
adrianjm-dotCMS: Same fix, yes. `delete` is emitted per pair to keep that output's contract — one event per pair removed, however they were removed.

While in there I found a third stale one you did not flag: the `confirmClearAll` doc still said the component "brings none of its own" dialog, five lines below the `CLEAR_ALL_KEY` doc explaining the dialog it brings. Fixed too.

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-paste.util.ts
adrianjm-dotCMS: Keeping the silent drop.

A block paste is a bulk gesture, and reporting every unparsed line would put the editor in the odd position of complaining about content the user pasted deliberately — a trailing comment, a blank section header. The fallback you point at is the important half: when nothing parses, the browser pastes normally and the text is visible for correction.

What did change is that blank values are now dropped too, so the rule is uniform: a line either yields a usable pair or is ignored.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-key-value/components/key-value-field/key-value-field.component.ts
adrianjm-dotCMS: Right, the class still promised the old shape.

There is now a named `DotKeyValueFieldValue` union, used by the class generic, `parseToDotKeyValue` and `handleChangeValue` — so the string branch is reachable through the types and a future edit gets the compiler signal.

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-table-row/dot-key-value-table-row.component.html
adrianjm-dotCMS: Good catch, and the failure mode is the worst kind: `DotMessageService` echoes the key, so these rendered English in every locale without ever failing.

`'Delete'` and `'add'` are now `keyValue.action.delete` and `keyValue.action.add`, both added to the bundle. `'Reorder'` is gone entirely — the handle is no longer exposed to assistive tech, so it has no name to give.

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-table-row/dot-key-value-table-row.component.html
adrianjm-dotCMS: Added `(keydown.space)` with `preventDefault()` on both spans.

Kept them as spans rather than promoting to `<button>`: the cell swaps to a real `input` on activation, and a button wrapping the full-width truncated text would need most of its own styling reset to look like the value it replaces.

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-table-row/dot-key-value-table-row.component.html
adrianjm-dotCMS: Agreed on all counts, and the role and tabindex have come off.

Keyboard reordering is out of scope here, so presenting the handle as pointer-only is the honest description — a focus stop that announces a button and does nothing is worse than not being focusable, as you say. FR-019 has been corrected: it was being read as requiring a keyboard-operable handle, which PrimeNG's pointer-driven reorder cannot honour.

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-table-row/dot-key-value-table-row.component.spec.ts
adrianjm-dotCMS: You are right — the early return fired first and the assertion would have passed either way.

The branch is now genuinely reachable, and for a better reason than the test: validation moved into the control, so it runs on every keystroke instead of only at commit. Typing another row's key flags `duplicatedKey` as you type; typing your own back clears it. That is the self-exclusion, and it is the difference between opening the editor clean and opening it already complaining about your own key.

The test drives those three states directly now, and the whitespace case became a separate test about the unchanged path.

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-table-header-row/dot-key-value-table-header-row.component.spec.ts
adrianjm-dotCMS: Added. It asserts the second click returns the field to `type="text"` and flips both `aria-pressed` and the glyph back.

You are right that the direction had no coverage anywhere once the old hidden-fields block was removed — and it is the headline change, so it should have.

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-table-row/dot-key-value-table-row.component.spec.ts
adrianjm-dotCMS: Confirmed, and it turned out to be one of three refusals that said nothing at all: an emptied value was saved, an emptied key was refused silently, and a rename onto a taken key was refused silently. In every case the input closed and the typed text vanished.

They now share one rule: an invalid edit keeps its input open and shows the entry row's message for the same refusal. `editControl` carries the same validators the entry row applies, and the row renders the same `<small>` elements. Escape still abandons.

Blank is refused everywhere as a result — empty value, empty key, and the pasted `KEY=`. And a value now commits exactly as typed: trimming decides only whether it counts as blank, so deliberately adding a trailing space is a change like any other, which it previously was not — the old code compared trimmed and emitted untrimmed, and swallowed that edit.

core-web/apps/dotcms-ui/src/app/portlets/shared/dot-content-types-edit/components/fields/dot-edit-field-dialog/dot-edit-field-dialog.component.spec.ts
adrianjm-dotCMS: Fair — they were declared and never driven.

Two tests added: one drives `$changeControls` and checks the dialog takes the Variables Save, calls it, and restores the Overview action on leaving; the other drives `$save` and checks the dialog closes.

core-web/libs/edit-content/src/lib/utils/key-value-order.util.spec.ts
adrianjm-dotCMS: Both added.

The reverse mismatch is already caught — the guard checks membership as well as count (`entries.every(({ key }) => key in source)`), so a key present only in the recovered order degrades safely — but you are right that nothing pinned it, and a future edit could have reduced that to a count check without a test noticing.

The duplicate-key case turns out to be a non-event worth recording: both parses go through `JSON.parse`, which keeps the last occurrence, so they agree on the count and the guard is not tripped.

oidacra (dismissed):

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-ng.component.html
```diff
@@ -1,33 +1,148 @@
...
+<!--
+    PrimeNG gives cells their separators but the table no outer frame, so the border
+    is ours; the last body row's separator is then suppressed, or it lands a second
+    line just inside that frame. Framing matches `dot-relationship-field`.
+-->
+<p-table
+    [value]="$variableList()"
+    #table
+    dataKey="key"
+    class="overflow-hidden rounded-lg border border-solid border-(--p-inputtext-border-color) transition-colors hover:border-(--p-inputtext-hover-border-color) [&_.kv-gutter]:pr-0! [&_.kv-gutter]:pl-[14px]! [&_table]:table-fixed [&_tbody_tr:last-child_td]:border-b-0!"
```
nicobytes: The class `[&_.kv-gutter]:pl-[14px]!` can be written as `[&_.kv-gutter]:pl-3.5!`.

```suggestion
    class="overflow-hidden rounded-lg border border-solid border-(--p-inputtext-border-color) transition-colors hover:border-(--p-inputtext-hover-border-color) [&_.kv-gutter]:pr-0! [&_.kv-gutter]:pl-3.5! [&_table]:table-fixed [&_tbody_tr:last-child_td]:border-b-0!"
```

core-web/apps/dotcms-ui-e2e/src/tests/edit-content/fields/key-value-field/clear-all.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+    createFakePayloadTextField
+} from '@utils/dot-content-types.mock';
+import { uniqueSuffix } from '@utils/utils';
+
+import { KeyValueField } from './helpers/key-value-field';
+
+const APP_KEY = 'dotVelocitySecretApp';
+
+/** PrimeNG appends the dialog to the body, so it is located globally. */
+const confirmClear = async (page, root) => {
```
nicobytes: Parameter `page` implicitly has an `any` type (same for `root`). Type them as Playwright `Page` / `Locator`, and add those types to the existing `@playwright/test` import.

```suggestion
const confirmClear = async (page: Page, root: Locator) => {
```

nicobytes: Parameter `root` implicitly has an `any` type. Type it as Playwright `Locator` (and import `type Locator` from `@playwright/test`).

```suggestion
const confirmClear = async (page: Page, root: Locator) => {
```

core-web/apps/dotcms-ui-e2e/src/tests/edit-content/fields/key-value-field/key-value-paging.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+        await page.getByTestId('title').waitFor({ state: 'visible', timeout: 20000 });
+        const field = new KeyValueField(page, VAR);
+        await field.expectVisible();
+        for (let i = 0; i < count; i++) {
+            await field.addEntry(`key-${String(i).padStart(3, '0')}`, `v${i}`);
+        }
+        return field;
+    }
+
+    const root = (page) => page.getByTestId(`field-${VAR}`);
```
nicobytes: Parameter `page` implicitly has an `any` type. Type it as Playwright `Page` and import `type Page` from `@playwright/test`.

```suggestion
    const root = (page: Page) => page.getByTestId(`field-${VAR}`);
```

core-web/apps/dotcms-ui-e2e/src/tests/edit-content/fields/key-value-field/key-value-paging.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+    });
+
+    test.afterEach(async ({ request }) => {
+        if (contentType) {
+            await deleteContentType(request, contentType.id);
+            contentType = null;
+        }
+    });
+
+    async function openWith(page, count: number) {
```
nicobytes: Parameter `page` implicitly has an `any` type. Type it as Playwright `Page` and import `type Page` from `@playwright/test`.

```suggestion
    async function openWith(page: Page, count: number) {
```

core-web/apps/dotcms-ui/src/app/portlets/dot-apps/components/dot-apps-configuration-detail/dot-apps-configuration-detail.component.spec.ts
```diff
@@ -118,8 +118,7 @@ class MockDotAppsService {
     standalone: true
 })
 class MockDotKeyValueComponent {
-    @Input() autoFocus: boolean;
-    @Input() showHiddenField: string;
+    @Input() showHiddenField: boolean;
     @Input() variables: DotKeyValue[];
     @Output() updatedList = new EventEmitter<DotKeyValue[]>();
```
nicobytes: We should use signal `input()` even in test mocks — the real `DotKeyValueComponent` already exposes these as `input()` (`$showHiddenField`, `$variables`). Same for `@Output()` → `output()` if we are aligning the stub with the public API.

```suggestion
    $showHiddenField = input<boolean>(false, { alias: 'showHiddenField' });
    $variables = input<DotKeyValue[]>([], { alias: 'variables' });
    updatedList = output<DotKeyValue[]>();
```

core-web/apps/dotcms-ui/src/app/portlets/shared/dot-content-types-edit/components/fields/dot-content-type-fields-variables/dot-content-type-fields-variables.component.spec.ts
```diff
@@ -24,11 +30,16 @@ import { DOTTestBed } from '../../../../../../test/dot-test-bed';
...
-        <dot-content-type-fields-variables [field]="value"></dot-content-type-fields-variables>
+        <dot-content-type-fields-variables
+            [field]="value"
+            [showTable]="showTable"></dot-content-type-fields-variables>
     `,
     standalone: false
 })
 class TestHostComponent {
+    /** The dialog binds this to "is the Variables tab the visible one". */
+    showTable = true;
```
nicobytes, line 41: We should drop this `TestHostComponent` + `DOTTestBed` setup and use Spectator’s `createHostFactory` / `SpectatorHost` instead. Bind `field` and `showTable` via `hostProps` and drive the spec through `spectator` (`setHostInput`, `query`, `click`, …).

```ts
import { createHostFactory, SpectatorHost } from '@ngneat/spectator/jest';

let spectator: SpectatorHost<DotContentTypeFieldsVariablesComponent>;
const createHost = createHostFactory({
  component: DotContentTypeFieldsVariablesComponent,
  imports: [DotKeyValueComponent],
  providers: [
    { provide: LoginService, useClass: LoginServiceMock },
    { provide: DotFieldVariablesService, useClass: DotFieldVariablesServiceMock },
    DotMessageDisplayService
  ]
});

spectator = createHost(
  `<dot-content-type-fields-variables [field]="field" [showTable]="showTable"></dot-content-type-fields-variables>`,
  {
    hostProps: {
      showTable: true,
      field: {
        ...dotcmsContentTypeFieldBasicMock,
        contentTypeId: 'ddf29c1e-babd-40a8-bfed-920fc9b8c77',
        id: mockFieldVariables[0].fieldId
      }
    }
  }
);
```

core-web/apps/dotcms-ui/src/app/portlets/shared/dot-content-types-edit/components/fields/dot-content-type-fields-variables/dot-content-type-fields-variables.component.ts
```diff
@@ -50,12 +60,38 @@ export class DotContentTypeFieldsVariablesComponent implements OnChanges, OnDest
     /** When `false`, hides the key-value table (used to embed without the table UI). */
     readonly $showTable = input<boolean>(true, { alias: 'showTable' });
 
+    /** Hands the dialog its footer buttons while this tab is the one on screen. */
+    readonly $changeControls = output<DotDialogActions>();
+
+    /** Raised once every pending change has been written. */
+    readonly $save = output<void>();
```
nicobytes: `$` is for signals, not outputs. These should be `changeControls` and `save` (same for template bindings `($changeControls)` / `($save)` and the sibling settings tabs that copy this API).

```suggestion
    readonly changeControls = output<DotDialogActions>();

    /** Raised once every pending change has been written. */
    readonly save = output<void>();
```

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-table-row/dot-key-value-table-row.component.ts
```diff
@@ -1,177 +1,217 @@
...
+     * Carries the same validators the entry row applies to the same field, so adding
+     * `name` and renaming a row to `name` are refused for the same reason and say so
+     * in the same words. Which validators are attached depends on the cell, and is
+     * decided in {@link #beginEditing}.
      */
-    form = this.#formBuilder.nonNullable.group({
-        value: ['', Validators.required],
-        hidden: [false]
-    });
+    editControl = this.#fb.nonNullable.control('');
```
nicobytes: `FormBuilder` is nullable by default (`string | null`). Prefer injecting `NonNullableFormBuilder` and using `this.#fb.control('')` so the control is strictly `FormControl<string>` without going through `.nonNullable`. Same for the header-row form.

```suggestion
    editControl = this.#fb.control('');
```

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-ng.component.html
adrianjm-dotCMS: Done — dropped the `[&_.kv-gutter]` selectors entirely: the class was already gone from the cells, so they matched nothing. The padding now sits on the cells as `pl-4! pr-0!`.

core-web/apps/dotcms-ui-e2e/src/tests/edit-content/fields/key-value-field/clear-all.spec.ts
adrianjm-dotCMS: Done — `confirmClear(page: Page, root: Locator)`, with both types added to the import.

adrianjm-dotCMS: Done — covered by the same change: `root: Locator`.

core-web/apps/dotcms-ui-e2e/src/tests/edit-content/fields/key-value-field/key-value-paging.spec.ts
adrianjm-dotCMS: Done — `openWith(page: Page, count: number)`, with `type { Page }` imported.

core-web/apps/dotcms-ui-e2e/src/tests/edit-content/fields/key-value-field/key-value-paging.spec.ts
adrianjm-dotCMS: Done — `const root = (page: Page) => page.getByTestId(...)`.

core-web/apps/dotcms-ui/src/app/portlets/dot-apps/components/dot-apps-configuration-detail/dot-apps-configuration-detail.component.spec.ts
adrianjm-dotCMS: Done — the stub now uses `input()` for `$showHiddenField` / `$variables` and `output()` for `updatedList`; the assertions read them as signals.

core-web/apps/dotcms-ui/src/app/portlets/shared/dot-content-types-edit/components/fields/dot-content-type-fields-variables/dot-content-type-fields-variables.component.spec.ts
adrianjm-dotCMS: Not in this PR — it is a full rewrite of the spec setup, well past the scope here, and the suite is green as it stands. Agreed on the direction, happy to take it as a follow-up.

core-web/apps/dotcms-ui/src/app/portlets/shared/dot-content-types-edit/components/fields/dot-content-type-fields-variables/dot-content-type-fields-variables.component.ts
adrianjm-dotCMS: Done for this component: `changeControls` / `save`, plus its template bindings and spec subscriptions. The sibling settings tabs still carry the `$` prefix — left out as they are outside this PR, worth a follow-up to make the convention consistent.

core-web/libs/ui/src/lib/components/dot-key-value-ng/dot-key-value-table-row/dot-key-value-table-row.component.ts
adrianjm-dotCMS: Done — both row components inject `NonNullableFormBuilder` and call `this.#fb.control("")` / `this.#fb.group({...})` directly.

nicobytes (approved):

PR #37285, opened 2026-08-28
docs(block-editor): spec for the embedded-contentlet selection guard (#36985)

### LINKED ISSUE #36985: Block Editor: embedded contentlet cannot be selected when text or other content precedes it

rjvelazco:
## Proposed Changes

Spec-Kit **PR 1 of 2** for the reopened #36985 — specification only, no implementation.

Clicking an embedded contentlet card in the new Block Editor *does* create a `NodeSelection`. A byte-wise change-guard downstream then rebuilds the whole document and destroys it, so the card shows no selection ring and the toolbar's edit / delete / reorder actions stay disabled.

### Why it looked random

The guard compares `JSON.stringify` of the incoming value against `JSON.stringify` of `editor.getJSON()`. TipTap normalizes on parse — fills schema defaults, drops undeclared attrs, reorders keys — so any stored document that does not round-trip **byte-identically** through the current schema fails the check permanently. It is a per-row data condition, not per content type or per site, which is exactly what authors were reporting.

### Three confirmed triggers, all measured

| Trigger | Cause |
|---|---|
| `chartCount` typo | The legacy editor wrote `SetDocAttrStep('chartCount', …)` until `3cd5549d35` (#26025, 2023-09-11). `stripDocStats` strips `charCount`, not `chartCount`. |
| Missing `indent` | `IndentExtension` declares `indent` (default `0`) on heading / paragraph / blockquote. Older content has no such key; the editor adds it on parse. |
| Unknown node types | The string branch skips `preserveUnknownNodesInDocument`, so a custom block is compared placeholder-against-original. |

Supplying a missing key is not enough — it has to be in the right *position*. That is the measured Case B in the spec, and it is what makes this a byte-comparison bug rather than a missing-attribute bug.

### Why only one screen

The new Angular Edit Content screen binds with `[formControlName]`, so the value arrives once through `writeValue` and the effect short-circuits forever after. The legacy JSP contentlet editor assigns `blockEditor.value`, so its effect *is* the loader and stays live — the only host where a re-push can reach `setContent` after load. The spec carries the full host-to-branch map.

## What's in this PR

| File | Contents |
|---|---|
| `specs/36985-block-editor-selection-guard/spec.md` | Problem, reproduction (server-free and local-instance), root cause, fix scope, non-goals, regression risk, AC-001…AC-008 |
| `specs/36985-block-editor-selection-guard/data-model.md` | Value shapes and the normalization `Node.fromJSON` applies |
| `specs/36985-block-editor-selection-guard/contracts/content-match.contract.md` | The comparator's input→verdict table — the source for the Red-phase tests |

`plan.md`, `research.md`, `quickstart.md` and `checklists/` are Spec-Kit working artifacts and are gitignored by design (`.gitignore:229-233`).

## Reproduction

No customer data required. The spec gives two recipes: a schema round-trip that needs no server, and an API-seeded BROKEN/CONTROL pair on a local instance. **The legacy JSP contentlet editor is required** — the content type needs `CONTENT_EDITOR2_ENABLED=false`, and the new Angular screen will not reproduce it.

A convenience seeding script is attached to the [QA note on the issue](https://github.com/dotCMS/core/issues/36985#issuecomment-5456242587) rather than committed; the spec's steps are the source of truth.

## Still open, recorded in the spec rather than hidden

- **The local CONTROL click has not been performed.** The host reasoning rests on customer-environment evidence plus the routing in `dot-contentlet-editor.service.ts:204`. If CONTROL also fails locally, "Why only one screen is affected" needs rewriting before anyone implements against it.
- **What re-pushes the value in the legacy host is unidentified.** `commitEditor` after remote `customBlocks` extensions is excluded for the local fixture. This does not block the fix — once the comparison is structural the verdict becomes "unchanged", so a re-push is harmless rather than merely suppressed.

## Checklist

- [x] Spec validated against `specs/36985-block-editor-selection-guard/checklists/requirements.md`
- [x] No `[NEEDS CLARIFICATION]` markers
- [x] No customer data — environment, hostnames and content identifiers replaced with a local fixture
- [x] Every claim measured against the real schema via `getSchema(createEditorExtensions(...))`, not inferred
- [ ] **Approval from another dev** — required before the implementation PR proceeds

## Notes for reviewers

- The spec deliberately keeps the Spec-Kit template's mandatory H2 headings, with the readable framing inside them, so `/speckit-plan` and `/speckit-tasks` keep resolving.
- **Non-goals worth confirming**: the click handler is not touched, no `chartCount` → `charCount` data migration, no port of click-to-select to the legacy editor (`contentlet-block.node.ts:23-24` has none, so the `FEATURE_FLAG_NEW_BLOCK_EDITOR=false` workaround still reproduces the original symptom — separate ticket), and no apply-once memo in the effect while the re-push is unidentified.
- **AC-005 is a behaviour change, not a regression guard.** A bare array is measured `false` today and must be `true` — expect it to fail at Red.

Spec for #36985

This PR fixes: #36985

This PR fixes: #36985


hmoreras (dismissed):

specs/36985-block-editor-selection-guard/spec.md
```diff
@@ -0,0 +1,479 @@
...
+In the new Block Editor, clicking an embedded contentlet card does nothing. No selection ring,
+and the toolbar's edit / delete / reorder buttons stay greyed out.
+
+The click handler isn't broken. It selects the card correctly. Then, a few milliseconds later,
+an unrelated piece of code rebuilds the entire document and throws the selection away.
+
+That rebuild is guarded by a check that asks "has the content actually changed?" The check
+compares two JSON strings. For content saved by today's editor the strings match, the rebuild
+is skipped, and everything works. For older content the strings don't match — even though the
+content is identical — so the rebuild fires every time.
```
rjvelazco: Why is the JSON different? Is there a change in the Schema? It's because of a change in the way we add the `charCount`, `wordCount` and `readingTime`? See: 

https://github.com/dotCMS/core/blob/31928163afd5345a59898ce4dbd1a4c1c8d698f9/core-web/libs/new-block-editor/src/lib/editor/utils/doc-stats.utils.ts#L19-L38


and

https://github.com/dotCMS/core/blob/31928163afd5345a59898ce4dbd1a4c1c8d698f9/core-web/libs/block-editor/src/lib/components/dot-block-editor/dot-block-editor.component.ts#L438-L450

We should always preserve the same schema so if something is failing because of this, we should fix the schema.

specs/36985-block-editor-selection-guard/spec.md
```diff
@@ -0,0 +1,479 @@
...
+> [QA note on the issue](https://github.com/dotCMS/core/issues/36985#issuecomment-5456242587).** It is deliberately not committed to the repo — the steps above
+> are the source of truth, and the script is a convenience.
+
+### What you should and shouldn't see
+
+**Expected Behavior**: the card takes a `NodeSelection` and keeps it. Ring visible, toolbar
+actions enabled, card can be edited, deleted or reordered.
+
+**Actual Behavior**: the selection is created and immediately discarded. The document is
+rebuilt, the cursor collapses to the end, no ring, "Edit Contentlet" stays disabled.
```
rjvelazco, line 194: Is there a script we can add to the editor/JSP so we can ensure the Block Editor is rebuilding? This can be done locally while testing when you read this comment. Post here the result of the check

specs/36985-block-editor-selection-guard/spec.md
```diff
@@ -0,0 +1,479 @@
...
+
+**Reproducibility**: always, for content matching the data condition — either trigger alone is
+enough. Never, for content last saved by the new editor. Note that dragging a text *range*
+across the card still works, which is what distinguishes this from the original #36985.
+
+### Confirming it in the browser
```
rjvelazco: Looks like this is a manual confirmation that we should do in the browser console which it's not the best practice. Is there a script we can add either in the block editor or in the JSP?

specs/36985-block-editor-selection-guard/spec.md
```diff
@@ -0,0 +1,479 @@
...
+
+- **Key order counts.** `{"textAlign":…,"level":…}` and `{"textAlign":…,"indent":…,"level":…}`
+  are different strings even when they describe the same thing.
+- **`stripDocStats` barely normalizes anything.** It touches only the root `attrs`, and only
+  three keys inside it. Everything below the root is compared raw.
+
+Meanwhile TipTap normalizes heavily when it parses: it fills in schema defaults, drops
+attributes the schema doesn't declare, and reorders keys to schema order. So the editor's side
+of the comparison is always the normalized shape, and the stored side is whatever happened to be
+written years ago.
```
rjvelazco: Do we have a list of attr/marks/node/extension that were dropped? For compatibility we should support what the old editor supported when possible

specs/36985-block-editor-selection-guard/spec.md
```diff
@@ -0,0 +1,479 @@
...
+through `writeValue` at form bind. Angular reactive forms never echo a control-value-accessor's
+own `onChange` back to it, and the `value` input is never set, so the effect at `:619`
+short-circuits on `if (!v) return` forever. `setContent` does fire once at load on broken
+content, but that's before any click and costs nothing. Nothing can call it again afterwards, so
+clicks are always safe here.
+
+**Legacy JSP contentlet editor** — assigns `blockEditor.value` on the custom element
+(`edit_field.jsp:316`), so the effect at `:619` *is* the loader and stays live on `value()` and
+`editor()`. This is the only host where a re-push can reach `setContent` after load, and
+therefore the only host where a failed check destroys a click-made selection.
```
rjvelazco: Why we allowed JSP to re-push the `value`/`setContent`?  I think content should be loaded once because after that the Block Editor has control over the content and, if I'm not wrong (we should double check), JSP or the legacy form shouldn't modify the Block editor content outside itself.

specs/36985-block-editor-selection-guard/spec.md
```diff
@@ -0,0 +1,479 @@
...
+This doesn't block the fix. Once the comparison is structural, the verdict for these documents
+becomes "unchanged", so a re-push is harmless rather than merely suppressed.
+
+---
+
+## Fix Scope & Non-Goals *(mandatory)*
+
+### What we're changing
+
+- Replace the string comparison in `editorContentMatchesParsed` with a structural ProseMirror
```
rjvelazco: How heavy can this be? Considering that the Block Editor is a Tree object and can has a lot of nesting? 

If this happens to be a heavy check, can we skip comparison and always send the new object to the JSP? If we do that, how can we skip temporal state that we don't want to push? Ex: The highlight we add on text selection or the class we add to the `dotContentlet`?

specs/36985-block-editor-selection-guard/spec.md
rjvelazco, line 1: Before going with the fix you propose, I want you to check this: Why we keep listening to the `this.value();`? Is there a way the content of the block editor changes outside the block editor? Is there any issue/ticket that back up this behavior?

rjvelazco (commented):
Analyze my comments; don't take anything for granted, be critical, and respond as concisely as possible.

specs/36985-block-editor-selection-guard/spec.md
rjvelazco: Is there a `ProseMirror`/`Tiptap` util for this?

specs/36985-block-editor-selection-guard/spec.md
```diff
@@ -0,0 +1,479 @@
...
+how the content was last written. Content saved by the current editor works. Content saved by
+an older editor, an importer, or the REST API often doesn't.
+
+**Severity / Impact**: customer-blocking on Evergreen. Every author editing pre-existing Story
+Block content with embedded contentlets is affected. The documented workaround — switching back
+to the legacy editor — does not help; see [Non-goals](#non-goals).
+
+---
+
+## Reproduction *(mandatory)*
```
rjvelazco, line 68: Another way to reproduce this issue without the script:

1. Set the `FEATURE_FLAG_NEW_BLOCK_EDITOR` to `false`
2. Create a Block Editor content with this structure:
```
# Heading 1

text testing

[Insert a `dotContent`]

Text testing
```
3. Save and Publish
4. Verify that the `dotContent` is clickable
5. Remove the `FEATURE_FLAG_NEW_BLOCK_EDITOR` or set it to `true`
6. Go back to the just created content
7. See the error when attending to click the `dotContent`

specs/36985-block-editor-selection-guard/spec.md
rjvelazco: Both — and you're right that the schema part needs fixing, but it isn't what causes *these two* triggers.

**The two triggers in the spec are stale data, not a live schema divergence:**

| Trigger | Why the stored JSON differs | Fixed in |
|---|---|---|
| `chartCount` in root `attrs` | Legacy wrote `SetDocAttrStep('chartCount', …)` — a typo | `3cd5549d35` (#26025), **2023-09-11** |
| no `indent` on heading/paragraph/blockquote | Legacy didn't declare `indent` at all before this | `5eb1a74e05` (#32235), **2025-05-27** |

I diffed the two `indent.extension.ts` files line by line — they are equivalent (`INDENT_TYPES = ['heading','paragraph','blockquote']`, `default: 0`). So today's legacy editor and the new editor agree; only content written **before those dates** is short an attr. `stripDocStats` strips `charCount`/`wordCount`/`readingTime` and cannot strip the `chartCount` typo (`doc-stats.utils.ts:4`).

**But there ARE live schema divergences.** I built both schemas with `getSchema()` and diffed nodes/marks/attrs:

| Divergence | Legacy | New |
|---|---|---|
| `listItem.textAlign` | present | **dropped** |
| `aiContent.loading` | `false` | **dropped** |
| `loader` node (`isLoading`) | registered | **dropped entirely** |
| `youtube.height` / `width` | `480` / `640` | `300` / `400` |
| `link.rel` | `null` | `'noopener noreferrer'` |
| `link.target` | `'_blank'` | `'_self'` |

Source of the `listItem` one: `dot-block-editor.component.ts:736` configures `TextAlign` with `['heading','paragraph','listItem','dotImage']`; `editor-extensions.ts:151` uses `['heading','paragraph']`.

New-only additions (harmless to parse, but they do change the emitted shape): `uploadPlaceholder` node, `table.ariaLabel`/`ariaLabelledby`/`caption`, `tableHeader.scope`, `dotAudio.title`, `dotVideo.title`, `dotImage.width`/`height`.

Full list and my recommendation in the thread on line 247.

Also worth noting: `dotcms-integration/.../TestDataUtils.java:100,128,156` still writes `"chartCount":20` in its Story Block fixtures — a ready-made local repro, and something to fix.

---
_🤖 Replied by Claude on behalf of @rjvelazco_

specs/36985-block-editor-selection-guard/spec.md
rjvelazco: This turned out to be the most important question on the PR, and it changes the fix. Two findings.

**1. Nothing in the repo pushes `value` more than once.** I checked every host:

| Host | How it sets `value` | Writes back? |
|---|---|---|
| Legacy JSP | `edit_field.jsp:316` — `blockEditor.value = content` | **No.** `valueChange` (`:307-309`) writes only to the hidden input |
| UVE sidebar | `[value]="contentlet.content"` (one-way) | **No.** `(valueChange)="value.set($event)"` goes to a *separate* signal |
| Angular Edit Content | `formControlName` → `writeValue` | **No.** Once at bind |
| Dev demo app | `main.ts` sets `demo.value` | **No** |

The comment at `editor.component.ts:615-617` justifies the guard with "two-way `[value]` + `(valueChange)` does not reset the cursor". **That consumer does not exist.** I found no issue or ticket backing it — it reads as defensive.

**2. The effect re-fires with the same `value` and the same `editor`.** I instrumented it and logged both identities. On the run that rebuilds the doc: `valueIdentity=true editorIdentity=true`. So it isn't a host re-push at all — the effect is re-running spuriously.

**What actually triggers it** (7 cases, unit-tested against a real mounted `DotCMSEditorComponent`):

| # | `.focus()` | Selection | Effect re-ran | Doc rebuilt | Selection survived |
|---|---|---|---|---|---|
| A | no | none (idle CD ×2) | no | 0 | — |
| B | yes | **Node** | **yes** | **1** | **NO** |
| C | yes | Node (CONTROL body) | yes | 0 | YES |
| D | no | Text | no | 0 | YES |
| E | yes | none | no | 0 | — |
| F | **no** | **Node** | **yes** | **1** | **NO** |
| G | yes | Text | no | 0 | YES |

**A `NodeSelection` re-runs the value effect. A `TextSelection` never does. `.focus()` is irrelevant.**

That is exactly the clue from the original report — clicking the card creates a `NodeSelection` and fails; dragging a text range across it creates a `TextSelection` and works. Row C shows the guard is the deciding factor: same `NodeSelection`, but a byte-identical body → verdict `true` → no rebuild → selection survives.

The rebuild itself, traced: `TX steps=["ReplaceStep"] docChanged=true selFrom=30 meta=["preventUpdate"]` from `editor.component.ts:626` — the effect's own `setContent`.

**Recommendation:** given that no host re-pushes, load-once is the better fix and your instinct on line 312 is right. I'd do both:
1. **Load once** — the effect stops being a live sync and becomes a one-shot loader. Removes the whole defect class rather than making the comparator smarter.
2. **Still fix the comparator** — `writeValue` and `commitEditor` share it, and a structural compare is strictly more correct than `JSON.stringify`.

I'll rewrite the spec's "One thing we still don't know" and "Why only one screen is affected" around this before `/speckit-tasks`. Flagging that this is a **material spec change after your approval**, so it needs a re-read.

---
_🤖 Replied by Claude on behalf of @rjvelazco_

specs/36985-block-editor-selection-guard/spec.md
rjvelazco: We didn't — I was wrong in the spec, and you're right about the design.

Verified in `edit_field.jsp`: the JSP sets `blockEditor.value` **exactly once** (`:316`) and never writes back. The `valueChange` listener (`:307-309`) only writes to the hidden form input:

```js
blockEditor.addEventListener('valueChange', ({ detail }) => {
    field.value = !detail ? null : JSON.stringify(detail);
});
```

So the spec's premise that "something re-pushes the value" is wrong. The real trigger is a spurious effect re-run on `NodeSelection` — measured, table in the thread on line 1.

And your design point holds: the Block Editor should own the content after load. No host in the repo pushes `value` twice (JSP, UVE sidebar, Angular Edit Content, dev demo app — all one-shot), so making the effect a one-shot loader costs us nothing and removes the defect class outright. I'm folding that into the fix scope.

---
_🤖 Replied by Claude on behalf of @rjvelazco_

specs/36985-block-editor-selection-guard/spec.md
rjvelazco: Yes — I measured it by building both schemas with `getSchema()` and diffing nodes, marks and attrs. Agree on the compatibility principle.

**Dropped in the new editor (data loss on load + save):**

| Kind | Name | Legacy default | Why |
|---|---|---|---|
| node | `loader` (`isLoading`) | `true` | not registered at all |
| attr | `listItem.textAlign` | `null` | `TextAlign` types dropped `listItem` |
| attr | `aiContent.loading` | `false` | not declared |

**Defaults changed** (a legacy doc that *omits* the attr now gets a different filled value):

| Attr | Legacy | New |
|---|---|---|
| `youtube.height` | `480` | `300` |
| `youtube.width` | `640` | `400` |
| `link.rel` | `null` | `'noopener noreferrer'` |
| `link.target` | `'_blank'` | `'_self'` |

**New-only** (parse fine, but they change the emitted shape, so they matter to any byte-comparison guard): `uploadPlaceholder` node; `table.ariaLabel`, `table.ariaLabelledby`, `table.caption`; `tableHeader.scope`; `dotAudio.title`; `dotVideo.title`; `dotImage.width`, `dotImage.height`.

Two notes on the legacy side I'd flag as intentional-drop candidates rather than bugs:
- `loader` is a *transient* upload spinner (`asset-uploader.extension.ts` inserts it). It shouldn't have been persistable in the first place, so "drop it" is arguably correct — but today a stored `loader` node silently disappears instead of degrading to a placeholder.
- `dotImage.textAlign` is **not** lost: the new `image.extension.ts:87` declares `textAlign` itself.

I'd keep this out of #36985 and open a separate compat ticket — restoring `listItem.textAlign` and `aiContent.loading` and re-aligning the four defaults is a different change with its own regression surface (`link.target` in particular is a deliberate security-ish change). Say the word and I'll file it.

One of these does bear directly on this PR though — see my reply on line 48, where `listItem.textAlign` turns your flag-toggle repro into a working one.

---
_🤖 Replied by Claude on behalf of @rjvelazco_

specs/36985-block-editor-selection-guard/spec.md
rjvelazco: Measured it. Short answer: not heavy, and `Fragment.eq` isn't where the cost is.

Benchmark against the real schema, 200 iterations each, docs with nested blockquotes / lists / 3×3 tables:

| top-level blocks | JSON size | PM nodes | current `JSON.stringify` ×2 | proposed `fromJSON` + `Fragment.eq` | `Fragment.eq` alone |
|---|---|---|---|---|---|
| 20 | 12 KB | 956 | 0.048 ms | 0.095 ms | **0.0002 ms** |
| 200 | 119 KB | 9,808 | 0.408 ms | 0.702 ms | **0.0005 ms** |
| 2,000 | 1.19 MB | 100,848 | 4.19 ms | 7.03 ms | **0.0049 ms** |

- **`Fragment.eq` is effectively free** — 5 µs on a 100k-node tree. It short-circuits on child-identity, so nesting doesn't hurt it.
- All the cost is `Node.fromJSON`. Proposed is ~1.7× current, and 7 ms on an absurd 1.2 MB document — once per value push.
- And that parse isn't really new: when the guard says "changed", `setContent` parses anyway. We only add a parse in the *unchanged* case, which is the case we're trying to make cheap.

**On "skip the comparison and always send the new object":** there's no temporal state to strip — I checked both examples you named:

- **Selection highlights** are ProseMirror **decorations**, not document content. `selection-preserve.extension.ts` uses `Decoration.node` / `Decoration.inline` inside plugin state (`:42`, `:134`).
- **The `dotContentlet` class** is a host binding on the node view: `contentlet.component.ts:19` — `'[class.is-selected]': 'selected()'`. DOM only.

Neither ever reaches `editor.getJSON()`, which only serialises node types, marks and declared attrs. So an unconditional push wouldn't leak anything.

I'd still not do it, though: unconditional `setContent` destroys the selection and undo history on **every** push — that's the current bug made permanent rather than conditional. Given that no host actually re-pushes (see line 1 thread), load-once is the cheaper and safer answer.

---
_🤖 Replied by Claude on behalf of @rjvelazco_

rjvelazco: `Fragment.eq` / `Node.eq` **are** the ProseMirror utils — `prosemirror-model`, re-exported as `@tiptap/pm/model`. There's no TipTap wrapper around them; TipTap uses them internally for the same purpose.

```ts
import { Node as PMNode } from '@tiptap/pm/model';

PMNode.fromJSON(editor.schema, incomingDoc).content.eq(editor.state.doc.content);
```

Two things worth pinning down when we implement:

- **`Fragment.eq` vs `Node.eq`.** `Node.eq` also compares the doc node's own `attrs`, which would reintroduce the `chartCount` sensitivity we're trying to remove. Today `schema.nodes.doc.spec.attrs` is `null`, so they happen to behave identically — but that's a coincidence we shouldn't rely on. Contract requirement #2 pins `Fragment.eq` for this reason.
- **`fromJSON` throws on an unknown mark** (`RangeError: There is no mark type … in this schema`) while unknown *nodes* degrade to a placeholder. So the call needs a `try/catch` returning `false` — same conservative verdict as today. That's AC-004 / AC-007.

---
_🤖 Replied by Claude on behalf of @rjvelazco_

specs/36985-block-editor-selection-guard/spec.md
rjvelazco: I tested your steps against both schemas. **As written it will not reproduce** — but it does with one addition.

**Why not:** content authored on *today's* legacy editor round-trips byte-identically through the new schema. I built the same document in both and compared:

```
LEGACY emits: [{"type":"heading","attrs":{"textAlign":null,"indent":0,"level":1},…}, …]
NEW    emits: [{"type":"heading","attrs":{"textAlign":null,"indent":0,"level":1},…}, …]
guard verdict = true          ← no rebuild, click works
```

Same keys, same order, and `stripDocStats` handles the root `charCount`/`wordCount`/`readingTime`. The legacy editor has written `charCount` since #26025 (2023-09-11) and `indent` since #32235 (2025-05-27), so a doc created today has nothing for the guard to trip on.

**Add a list and it reproduces.** Legacy `TextAlign` covers `listItem`; the new editor's doesn't — so `listItem.attrs.textAlign` is dropped on parse:

```
LEGACY: {"type":"listItem","attrs":{"textAlign":null},"content":[…]}
NEW   : {"type":"listItem","content":[…]}

guard verdict (JSON.stringify) = false   ← reproduces
guard verdict (Fragment.eq)    = true    ← the fix
```

**Revised steps:**

1. Set `FEATURE_FLAG_NEW_BLOCK_EDITOR=false`
2. Create Block Editor content with this structure — the bullet list is what matters:
   ```
   # Heading 1

   text testing

   - one
   - two

   [Insert a dotContent]

   Text testing
   ```
3. Save and Publish
4. Verify the `dotContent` is clickable
5. Set `FEATURE_FLAG_NEW_BLOCK_EDITOR=true` (or remove it)
6. Reopen the content
7. The `dotContent` is not selectable on click — but dragging a text range across it does select it

That's a genuinely better repro than the spec's `curl` recipe: no hand-crafted JSON, and it exercises a **live** schema divergence rather than a historical one. I'll promote it to the primary path in §Reproduction and keep the `curl` version as the `chartCount`/`indent` (pre-2025 content) case, since those are what the customer actually hit.

Worth noting this makes the bug **not purely historical** — any content with a list authored on the legacy editor is affected today. See the line-247 thread for the full divergence list.

---
_🤖 Replied by Claude on behalf of @rjvelazco_

specs/36985-block-editor-selection-guard/spec.md
rjvelazco: Agreed the console snippet is the wrong artifact. I couldn't run it — there's no dotCMS instance up in this session, so I have no browser result to post. But I got the same evidence headlessly, which I think is a better answer than a debug script.

I mounted a real `DotCMSEditorComponent` in Jest, loaded the BROKEN body, and did exactly what `contentlet.component.ts:selectNode` does. Traced every transaction:

```
>>> dotContent pos = 23
>>> immediately after setNodeSelection: NodeSelection = true | from = 23
>>> TX steps=["ReplaceStep"] docChanged=true selFrom=30 meta=["preventUpdate"]
        at editor.component.ts:626:25
>>> after CD: NodeSelection = false | from = 30 | doc identity unchanged = false | same editor = true
>>> docSize before=31 after=31
```

The whole chain, confirmed: the effect's own `setContent` (`:626`) dispatches a `ReplaceStep` that rebuilds a document of **identical size** (31 → 31) and moves the selection from 23 to the doc end. Same `Editor` instance, new doc, `NodeSelection` gone.

Control run with a byte-identical body: 0 rebuilds, `NodeSelection` survives.

**Recommendation:** no debug script in the JSP or the editor. Replace both console snippets in the spec with:
1. **The Jest harness above** — it asserts "click a contentlet on a BROKEN body → selection survives" directly, no browser, and it's the natural Red-phase test for `/speckit-tasks`. Already written; I'll promote it in the implementation PR.
2. **A Playwright e2e** for the real click path — which also retires the Principle V exception I'd declared in `plan.md`.

Shipping a diagnostic script would just be a worse version of test 1.

---
_🤖 Replied by Claude on behalf of @rjvelazco_

specs/36985-block-editor-selection-guard/spec.md
rjvelazco: Same answer as line 133 — the manual console check should go, replaced by the Jest harness (and a Playwright e2e for the real click), not by a script we ship in the editor or the JSP.

One thing worth keeping from this thread: the console snippet I originally circulated was also *wrong*. It read `document.querySelector('.ProseMirror').pmViewDesc.view`, and `ViewDesc` never stores `view` — I checked `prosemirror-view@1.41.8/dist/index.js:741-753`, where `dom.pmViewDesc = this` is set but `this.view` is not (the only `this.view =` assignments are in `ViewTreeUpdater:1765`, `MouseDown:3332`, `DOMObserver:4578`). So the comparison was `undefined === undefined` — a silent `true`. Another argument for making this an assertion in a test rather than something a human eyeballs in a console.

I'll strip both snippets and cite the tests instead.

---
_🤖 Replied by Claude on behalf of @rjvelazco_

specs/36985-block-editor-selection-guard/spec.md
rjvelazco: Applying your rule — a node counts as dropped only if there's no `node.name` equivalent **and** the extension was really a node — I re-tested and **my `loader` claim was wrong. Zero nodes are dropped.**

`loader` has no name equivalent in the new schema (`uploadPlaceholder` is a different name), and it is a real `Node.create({ name: 'loader' })`. But it isn't lost: it's an *unknown* node, so `preserveUnknownBlockNodes` wraps it in `dotUnsupportedBlock` on load and `restoreUnknownBlockNodes` unwraps it on emit. Measured round-trip:

```
in : {"type":"loader","attrs":{"isLoading":true}}
out: {"type":"loader","attrs":{"isLoading":true}}    ← byte-identical
```

Rendering degrades to the unsupported-block placeholder, but the data survives a save. That mechanism only exists for nodes — which is why the *attr* cases below have no equivalent safety net.

**Corrected table.** Each divergence scored against the actual guard:

| Divergence | Kind | Guard verdict today | `Fragment.eq` | Data lost on save? |
|---|---|---|---|---|
| `loader` | unknown node | **`true`** | `true` | **No** — preserved |
| `listItem.textAlign` | attr on a known node | `false` | `true` | **Yes** |
| `aiContent.loading` | attr on a known node | `false` | `true` | **Yes** |
| `youtube` missing `height`/`width` | changed default | `false` | `true` | No — value overwritten |
| `link` mark missing `rel`/`target` | changed default | `false` | `true` | No — value overwritten |

So the honest count is **2 real data-loss divergences, 0 dropped nodes**, and `loader` should come off the list entirely — it neither loses data nor trips the guard.

Two further notes:

- The two default changes only bite content authored *before* those attrs existed. Legacy `toJSON` serialises every declared attr — even when null — so a legacy-authored `youtube` node carries its own `height`/`width` and the new default never applies.
- `aiContent` also gained a `content` attr, but that one is present in both schemas, so it isn't a divergence.

Net effect on this PR: the compat ticket shrinks to `listItem.textAlign` + `aiContent.loading`, and the "unknown node types" trigger in the spec needs re-scoping — the object branch handles it correctly (`true` above). It only misbehaves on the string branch, which skips `preserveUnknownNodesInDocument`.

---
_🤖 Replied by Claude on behalf of @rjvelazco_

rjvelazco: Nice!

specs/36985-block-editor-selection-guard/spec.md
rjvelazco: Now we are on the same page here!

specs/36985-block-editor-selection-guard/spec.md
rjvelazco: After knowing that `value()` is never re-pushed, is this comparison needed at all?

specs/36985-block-editor-selection-guard/spec.md
rjvelazco: Also, we've already confirmed that no host re-pushes are happening, but the issue is still reproducible using the script mentioned in the steps to reproduce. 

A couple of things worth noting:
- It **only happens on JSP**, not in the new content editor
- The host re-push has been ruled out as the cause

Based on what we know so far, what's the current assumption on the most likely culprit?

rjvelazco: Good pushback — and you're right that the JSP-only part still needs explaining. I over-corrected earlier when I said "Why only one screen is affected" was wrong. The host asymmetry is real; it just isn't a *re-push*.

**The bug needs three conditions, all true at once:**

**1. `value()` must be non-empty — this is the JSP-only condition.**

Verified `dot-edit-content-block-editor.component.html`: the new content editor binds **only** `[formControlName]`, never `[value]`. So `value()` stays `''` and the effect hits `if (!v) return` on *every* run — `setContent` is unreachable there. The JSP assigns `blockEditor.value = content` (`edit_field.jsp:316`), so on that host the effect is the loader and stays armed. Same for the UVE side panel (`[value]="contentlet.content"`).

That's your answer: not "JSP re-pushes", but "JSP is the only host that arms `value()` at all".

**2. Something must re-run the effect after load.** Measured: a `NodeSelection` does, a `TextSelection` doesn't, `.focus()` is irrelevant.

And here's the new finding — **it is not dependency-driven.** I wrapped both reads in `untracked()`, giving the effect zero signal dependencies, and re-ran the 7 cases:

| Case | Selection | Effect runs (zero deps) |
|---|---|---|
| B, F | **Node** | **2** |
| D, E, G | Text / none | 1 |

A zero-dependency Angular effect ran twice. So the re-entry is not "a signal changed" — something is marking the effect node dirty and Angular re-flushes it.

**3. The guard must return `false`.** Permanently true for legacy content (`chartCount`, missing `indent`, `listItem.textAlign`).

---

**Most likely culprit for condition 2 — hypothesis, not yet proven:**

The contentlet card is an **Angular component node view**, created by ngx-tiptap's `AngularNodeViewRenderer` with the host component's injector. A `NodeSelection` makes ProseMirror call `selectNode()` / `deselectNode()` on that node view, which updates that Angular component — inside the host's view refresh, re-flushing the host's effects. A `TextSelection` never touches a node view, which matches D and G exactly.

**Design implication, and it's the important bit:** the fix must not assume "the effect only re-runs when `value` changes" — measured false. Load-once needs an **explicit one-shot latch**, not reliance on the dependency graph.

**Next experiment to confirm** (happy to run it with you): swap the contentlet node view for a plain DOM `addNodeView` and re-run cases B/F. If the second run disappears, the node view is confirmed as the trigger.

---
_🤖 Replied by Claude on behalf of @rjvelazco_

specs/36985-block-editor-selection-guard/spec.md
rjvelazco: Mostly no — but not entirely. It depends which of the three call sites you mean.

| Call site | Still needs the comparison? |
|---|---|
| The `value` effect (`:619-632`) | **No** — with load-once it becomes dead code |
| `writeValue` (`:747`) | **Yes** — keep it |
| `commitEditor` (`:498`) | No — one-shot `pendingValue` drain, guard is redundant |

**Why `writeValue` keeps it:** it's the `ControlValueAccessor` entry point, and Angular forms can legitimately call it more than once — `control.setValue()`, `patchValue()`, `form.reset()`. I grepped `libs/edit-content` and found no caller doing that to a Story Block control **today**, but that's an observation about the current code, not a contract. Without a guard, any future `reset()` would rebuild the document and destroy the selection and undo history — the same bug, re-entering through the front door.

**Important caveat on load-once**, from the `untracked()` experiment in the line-1 thread: the effect re-runs even with **zero** signal dependencies. So we cannot implement load-once by trusting the dependency graph — it needs an **explicit one-shot latch** (a `#valueLoaded` flag, or moving the load into `ngOnInit`/`commitEditor` and deleting the effect outright). "It won't re-run because nothing changed" is measurably false.

**So the shape I'd propose:**
1. Delete the `value` effect as a live sync; load `value` once, explicitly latched.
2. Keep **one** structural comparator (`Fragment.eq`), called from `writeValue` only.
3. Drop the guard from `commitEditor`.

That's less code than today and removes the defect class rather than making the comparator smarter — which I think is what you were driving at from the start.

---
_🤖 Replied by Claude on behalf of @rjvelazco_

specs/36985-block-editor-selection-guard/spec.md
rjvelazco: **Culprit found and confirmed in the browser.** The trigger is the Angular component node view.

Experiment: replaced the `dotContent` Angular node view with a plain-DOM one that reproduces click → `NodeSelection`, everything else untouched. Result from a real JSP editor on affected content:

```
[BE-EXP1] dotContent using PLAIN DOM node view (Angular one disabled)
[BE-EXP1] value effect run #1 -> guardMatched=false willRebuildDoc=true currentSelection=_TextSelection from=1
[BE-EXP1] value effect run #1 -> CALLING setContent()          ← load, expected

[BE-EXP1] plain-DOM node view mousedown, pos = 69
[BE-EXP1] plain-DOM selectNode()
                                                               ← NO run #2. Card is selectable.
```

The effect runs **once**, at mount. No re-run on click. The card selects.

**The chain, in ngx-tiptap 14.0.1 (`fesm2022/ngx-tiptap.mjs`):**

| Step | Code |
|---|---|
| 1 | Node view component is created with the **host editor's** injector — `elementInjector: injector` (`:240`) and `applicationRef.attachView(...)` (`:243`) |
| 2 | `handleSelectionUpdate` fires only when the selection **fully covers the node**: `from <= pos && to >= pos + nodeSize` (`:337`) |
| 3 | It calls `selectNode()` → `this.renderer.updateProps({ selected: true })` (`:403`) |
| 4 | `updateProps` → `componentRef.setInput(...)` (`:258`) |
| 5 | Angular's `setInput` ends in `markViewDirty(childComponentLView, 1)` (`_debug_node-chunk.mjs:9242`), which walks **every ancestor LView** setting the dirty flag (`:6208-6215`) — the host editor view included |
| 6 | Host view refresh → `runEffectsInView` (`:5920`) → the `value` effect runs again |
| 7 | Guard is `false` on legacy content → `setContent` → `ReplaceStep` → `NodeSelection` destroyed |

**Step 2 is why a text drag behaves differently** — a `TextSelection` that doesn't span the node never reaches `updateProps`, so nothing is dirtied and the effect never re-runs. That's the "drag works, click doesn't" asymmetry, straight out of the library.

**One layer I have NOT nailed:** why the effect *node* is marked dirty rather than just the view. `runEffectsInView` skips non-dirty effects (`:5928`), yet with both signal reads wrapped in `untracked()` — zero dependencies — the effect still re-ran on the NodeSelection cases. So the re-entry is not dependency-driven, and I can't yet name the exact flag. It doesn't change the fix, but it *does* mean **we must not implement load-once by trusting the dependency graph.**

**Corroboration from the same log:** run #1 shows `guardMatched=false` at load on real customer-shaped content — the comparator defect reproduces on a live instance, not just in unit tests.

**So the fix, now fully justified:** the `value` effect is inherently fragile, because *any* node-view `setInput` re-runs it. Don't make the comparator smarter — remove the live sync:

1. Load `value` once, explicitly latched (or move the load into `ngOnInit` / `commitEditor` and delete the effect).
2. Keep **one** structural comparator (`Fragment.eq`) in `writeValue`, where forms can legitimately re-push.
3. Drop the guard from `commitEditor`.

---
_🤖 Replied by Claude on behalf of @rjvelazco_

zJaaal (approved):

Issue #37224, opened 2026-08-25
Add GET /v1/analytics/sessions/behavior endpoint with reach-target and URL-param goal support


freddyDOTCMS:
## Description

Add a new `GET /v1/analytics/sessions/behavior` endpoint in CAEM that evaluates session behavior against experiment goals. The endpoint supports two goal types:

- **Reach-target goal**: Determines whether sessions reached a target page after visiting a reference page
- **URL-param goal**: Determines whether sessions contained a specific URL parameter during any event in the session

The endpoint accepts a behavior type, experiment context parameters (`experimentId`, `runningId`), and goal-specific required parameters. Includes service-layer dispatch with parameter validation and dedicated repositories for each goal type.

## Acceptance Criteria

- [ ] `GET /v1/analytics/sessions/behavior` accepts `behavior`, `experimentId`, `runningId`, and goal-specific required parameters
- [ ] The endpoint returns reach-target goal results when `behavior=reach-target` is requested
- [ ] The endpoint returns URL-param goal results when `behavior=url-param` is requested
- [ ] Requesting a behavior value with missing required parameters returns a 400 error
- [ ] Reach-target goal correctly counts sessions where the target page was visited after the reference page
- [ ] URL-param goal correctly counts sessions where any event in the session contained the specified URL parameter
- [ ] Existing `GET /v1/analytics/sessions` behavior is not affected
- [ ] Unit tests cover service dispatch logic and parameter validation for all behavior values
- [ ] Unit tests cover SQL logic and edge cases for reach-target queries
- [ ] Unit tests cover SQL logic and edge cases for URL-param queries
- [ ] Postman tests covering the new endpoint are added to the [CAEM Postman collection](https://github.com/dotCMS/dot-ca-event-manager/tree/main/postman)

## Additional Context

Implementation repository: [dotCMS/dot-ca-event-manager](https://github.com/dotCMS/dot-ca-event-manager)

freddyDOTCMS:
Spec PR opened in dot-ca-event-manager: https://github.com/dotCMS/dot-ca-event-manager/pull/112

This is Spec-Kit PR 1 (spec only). Implementation PR will follow once the spec is approved.

Issue #37223, opened 2026-08-25
Extend GET /v1/analytics/sessions with bounce/exit metrics and variant dimension


freddyDOTCMS:
## Description

Extend the `GET /v1/analytics/sessions` endpoint in CAEM to support bounce and exit session metrics, a variant dimension for grouping, and new optional query parameters for experiment-based filtering.

**New metrics:**
- Bounce sessions count
- Bounce rate
- Exit sessions count
- Exit rate

**New dimension:**
- Variant — enables grouping session data by experiment variant

**New query parameters:**
- `experimentId` — filter sessions by experiment
- `runningId` — filter by running experiment instance
- `referencePage` — reference page used to calculate exit rate

The new metrics query against the session-level fact table, separate from the existing engagement aggregation path. Includes a new repository component, service routing logic, and integration tests.

## Acceptance Criteria

- [ ] `GET /v1/analytics/sessions` returns bounce sessions count when the bounce sessions metric is requested
- [ ] `GET /v1/analytics/sessions` returns bounce rate when the bounce rate metric is requested
- [ ] `GET /v1/analytics/sessions` returns exit sessions count when the exit sessions metric is requested
- [ ] `GET /v1/analytics/sessions` returns exit rate when the exit rate metric is requested
- [ ] Session data can be grouped by the variant dimension
- [ ] The endpoint accepts and applies `experimentId`, `runningId`, and `referencePage` as optional filter parameters
- [ ] Existing session metrics and endpoint behavior are not regressed
- [ ] Integration tests pass for bounce rate calculations
- [ ] Integration tests pass for exit rate calculations
- [ ] Postman tests covering the new metrics and parameters are added to the [CAEM Postman collection](https://github.com/dotCMS/dot-ca-event-manager/tree/main/postman)

## Additional Context

Implementation repository: [dotCMS/dot-ca-event-manager](https://github.com/dotCMS/dot-ca-event-manager)

PR #37287, opened 2026-08-28
feat(37227): CAEM-backed experiment goal result queries

### LINKED ISSUE #37227: Add CAEM-backed experiment result query classes with legacy CubeJS fallback switch
## Description

Add new experiment goal result query implementations in dotCMS core that call the CAEM analytics API endpoints (introduced in issues #37223 and #37224) instead of CubeJS. The existing CubeJS-based classes are **not removed** — both sets coexist and a switch mechanism controls which implementation is active, allowing legacy infrastructure to remain supported during the transition.

**New implementations (alongside existing CubeJS classes):**
- HTTP client to communicate with the CAEM analytics API
- Bounce rate result query using the new sessions endpoint
- Exit rate result query using the new sessions endpoint with reference page
- Reach-target result query using the new sessions behavior endpoint
- URL-parameter result query using the new sessions behavior endpoint

**Switch mechanism:**
- A configuration switch selects between the legacy CubeJS path and the new CAEM analytics path at runtime
- No behavioral change occurs unless the switch is explicitly enabled
- The two paths query **different analytics infrastructure** (CubeJS vs. CAEM/ClickHouse), but should produce the same goal rates when both backends have ingested the same underlying events

## Acceptance Criteria

- [ ] A new HTTP client is available to call CAEM analytics API endpoints for sessions and session behavior
- [ ] New bounce rate result query returns correct goal rates using the CAEM sessions endpoint
- [ ] New exit rate result query returns correct goal rates using the CAEM sessions endpoint with reference page
- [ ] New reach-target result query returns correct goal rates using the CAEM sessions behavior endpoint
- [ ] New URL-parameter result query returns correct goal rates using the CAEM sessions behavior endpoint
- [ ] Existing CubeJS-based result query classes remain unchanged and functional
- [ ] A switch (e.g. configuration flag) selects between the legacy CubeJS path and the new CAEM analytics path
- [ ] When the switch is disabled, the system uses the existing CubeJS implementation (no regression)
- [ ] When the switch is enabled, the system uses the new CAEM-backed implementations
- [ ] Unit tests cover each new result query class
- [ ] Unit tests cover the switch dispatch logic

## Additional Context

Depends on CAEM endpoints introduced in:
- #37223 — Extend GET /v1/analytics/sessions
- #37224 — Add GET /v1/analytics/sessions/behavior

freddyDOTCMS:
> [!IMPORTANT]
> **Design Artifacts:** [View Spec & Implementation Details](https://claude.ai/code/artifact/eeddea67-f59e-49c7-83b8-a30c10f2914d)

## Summary

- Renames `CubeJSResultSet` → `AnalyticsResultSet` and `CubeJSResultSetImpl` → `AnalyticsResultSetImpl` (provider-agnostic names; 6 files)
- Introduces `ExperimentGoalResultsQuery` interface with `executeByDay` and `executeAggregate` methods — the permanent seam between the factory and any analytics backend
- Adds `CubeJSGoalResultsAdapter` (temporary shim wrapping existing CubeJS classes; deleted when CubeJS is removed)
- Adds `CaemHttpClient` (CircuitBreakerUrl + existing HMAC auth; `CaemResponse` record for testability)
- Implements four CAEM result query classes: `BounceRateCAEMResultQuery`, `ExitRateCAEMResultQuery`, `ReachPageCAEMResultQuery`, `UrlParameterCAEMResultQuery` — each calls the appropriate CAEM sessions endpoint with the correct metrics, dimensions, and goal-specific parameters
- Adds `GoalConditionUtil` for extracting `referencePage`, `targetUrl`, and `QueryParameter` from goal conditions
- Refactors `ExperimentResultsQueryFactory` to dispatch both `executeByDay` and `executeAggregate` to either the CubeJS or CAEM implementation based on `FEATURE_FLAG_CAEM_EXPERIMENT_RESULTS` (default `false`), read per call via `ConfigExperimentUtil` `AtomicBoolean` for live toggle without restart
- Updates `ExperimentsAPIImpl.getSummary()` and `getTotalSessions()` to use the new factory dispatch methods
- 37 unit tests + 5 integration tests (factory dispatch, all switch states and goal types)

Closes #37227
Depends on: #37223 (CAEM sessions endpoint), #37224 (CAEM sessions behavior endpoint)

## Test plan

- `./mvnw test -pl :dotcms-core -Dtest="ExperimentResultsQueryFactoryTest,CaemHttpClientTest,BounceRateCAEMResultQueryTest,ExitRateCAEMResultQueryTest,ReachPageCAEMResultQueryTest,UrlParameterCAEMResultQueryTest"`
- `./mvnw verify -pl :dotcms-integration -Dcoreit.test.skip=false -Dit.test=ExperimentResultsQueryFactoryIntegrationTest`


erickgonzalez (changes requested):
## Verdict
`REQUEST CHANGES` — 5 blockers.

The spec assumes a switch seam that the current interface cannot support, declares unit tests only with no justification for the other types, and leaves both the config key and the authorization of the new path undefined.

## Blockers

**B1 — [CRITICAL] Unit tests only, other types not justified**
- FR-015/016/017 and SC-004 declare unit tests exclusively. Principle V requires the layer-appropriate type(s), or an explicit statement from the developer as to why one cannot be implemented ("silence is not consent"). There is live coverage this change touches: `dotcms-integration/src/test/java/com/dotcms/experiments/business/ExperimentResultsQueryFactoryIntegrationTest.java:93` asserts exactly the query the switch replaces.
- Add FRs for integration tests (dispatch with flag on/off + `ExperimentResults` parity) and Postman against the results endpoint, or one explicit line per omitted type with its reason.

**B2a — [CRITICAL] The existing interface returns CubeJSQuery**
- Key Entities states the CAEM classes "implement this same interface", but `dotCMS/src/main/java/com/dotcms/experiments/business/result/MetricExperimentResultsQuery.java:18` declares `CubeJSQuery getCubeJSQuery(Experiment)`. A CAEM implementation cannot return a `CubeJSQuery`.
- Define the real seam (a provider-agnostic abstraction) and correct FR-001/FR-010 and Key Entities accordingly.

**B2b — [CRITICAL] The seam does not fit in the factory, and call sites do change**
- Legacy Considerations claims the `ExperimentsAPIImpl` call sites stay "unchanged", but `ExperimentsAPIImpl.java:1429-1440` return `CubeJSResultSet` via `CubeJSClient`, and the mapping to `VariantResults`/`GoalResults` that FR-013 attributes to the query classes actually happens in `ExperimentsAPIImpl.java:1324-1345`, reading `Events.*` keys.
- Rewrite FR-008/FR-013 and Legacy Considerations to place the switch where `ExperimentResults` is produced, not where the query is built.

**B3 — [HIGH] Configuration key undefined**
- FR-008 and Key Entities say "a dotCMS `Config` property" without naming the key; the PR body names `DOT_CAEM_ENABLED`. The convention in this module is a constant in `dotCMS/src/main/java/com/dotcms/featureflag/FeatureFlagName.java:10` read via `Config.getBooleanProperty` (`ConfigExperimentUtil.java:113`). Also, SC-003 (toggle without restart) conflicts with the `static Lazy` that caches the dispatch map in `ExperimentResultsQueryFactory.java:64`.
- Name the key and its default (`false`), and require in an FR that it be read per evaluation or via `SystemTableUpdatedKeyEvent` as in `ConfigExperimentUtil.java:79-85`.

**B4 — [HIGH] Authorization and credential scope not expressed**
- FR-012 says it reuses the `EventAnalyticsProxyHelper` token, but `buildAuthHeader(Host)` is package-private and site-scoped (`EventAnalyticsProxyHelper.java:275`), and **with no token the request goes out without an `Authorization` header** (lines 283-291). The CubeJS path is user+site-scoped via `resolveAnalyticsApp(user)` -> `getCurrentHost()` (`AnalyticsHelper.java:466`); no FR carries `user`/site into the CAEM path.
- Add FRs: preserve the same authorization as the CubeJS path, fail (do not call) when no token is available, and never log the token or a URL containing it.

**B5 — [HIGH] SC-007 asserts behavior that does not exist today**
- SC-007 says `CLICK_ON_ELEMENT` "continues to use the CubeJS path", but the map in `ExperimentResultsQueryFactory.java:69-74` has no entry for that type, so `getMetricCubeJSQuery` (lines 163-165) NPEs today. The criterion is unverifiable as written.
- Reword SC-007 as "no change to current behavior for unsupported types", or take it out of scope.

## Questions for the author

**Q1** — Where does the seam live: at the query-builder level, or an `ExperimentResultsProvider` that returns a complete `ExperimentResults`? Today the factory produces a query and the parsing lives in `ExperimentsAPIImpl.getResults`. My recommendation: a provider at the `getResults` level, leaving `MetricExperimentResultsQuery` and the three CubeJS classes untouched. Answering this resolves B2a and B2b, and rewrites FR-010/FR-013.

**Q2** — Which `Host` resolves the CAEM base URL and token, given that `getResults(experiment, user)` has no request context? CubeJS uses `getCurrentHost()`, which does not exist in the experiment-finalization job (`ExperimentsAPIImpl.java:1443`). Suggested: derive the Host from the experiment's page and pass it explicitly. Resolves B4.

**Q3** — Is the daily series in scope? `getSummary` uses `createWithDayGranularity` with `timeDimension("Events.day")` (`ExperimentResultsQueryFactory.java:90-92`) and the `summarize` loop consumes it. No FR mentions it: if CAEM does not return day granularity, the shape of `ExperimentResults` changes. Resolves part of B2b.

**Q4** — Does `URL_PARAMETER` use `paramName`/`paramValue` (FR-006), or the single `QUERY_PARAMETER` with a regex parameter name declared in `MetricType.java:41-45`? Today that type reuses `ReachTargetAfterExperimentPageResultQuery` (`ExperimentResultsQueryFactory.java:73`). This determines the parameters sent to CAEM.

## Recommended next steps

1. Run `/speckit-clarify` for Q1-Q4 before anything else; **do not plan yet** — Q1 and Q3 determine the abstraction and the shape of `ExperimentResults`, so a plan written first gets redone in full.
2. Rewrite FR-001, FR-008, FR-010, FR-013 and Key Entities (`MetricExperimentResultsQuery`, `ExperimentResultsQueryFactory`) per the answer to Q1 — resolves B2a and B2b.
3. Name the config key, its default, and its re-read semantics in FR-008; add the proposed constant to `FeatureFlagName` — resolves B3.
4. Add security FRs: authorization equivalent to the CubeJS path, explicit failure with no token, and a prohibition on logging the token — resolves B4.
5. Add integration and Postman FRs (or the explicit per-type justification) and reword SC-004 — resolves B1; reword SC-007 — resolves B5.
6. Convert the Edge Cases section from interrogative prose into Given/When/Then scenarios inside the story that owns each case; move the HTTP client choice and the response-envelope parsing detail to `plan.md`.
7. Reorder US5 to ship first (US1-US4 are not independently testable without the flag), and move the physical-design claims in Legacy Considerations down to `plan.md`.
8. Chain `/speckit-clarify` -> spec re-review -> `/speckit-plan` -> `/speckit-tasks` -> `/speckit-analyze`, and only then `/speckit-implement`.

erickgonzalez:
## Blocking review findings (Critical)

Two critical blockers before `/speckit-plan` can run. Both come from the same root cause: the seam this spec defines — `ExperimentGoalResultsQuery.execute(Experiment) → GoalResults` — cannot carry the data `ExperimentsAPIImpl.getResults` consumes today. They should be resolved together.

**B1a — [CRITICAL] The seam drops per-variant TotalSession**

- `getResults` builds its result from **two** distinct queries: the aggregate one and the day-granularity one ([ExperimentsAPIImpl.java:1306](https://github.com/dotCMS/core/blob/main/dotCMS/src/main/java/com/dotcms/experiments/business/ExperimentsAPIImpl.java#L1306)). The total-sessions block lives on `ExperimentResults.Builder.addTotalSession` ([ExperimentResults.java:111](https://github.com/dotCMS/core/blob/main/dotCMS/src/main/java/com/dotcms/experiments/business/result/ExperimentResults.java#L111)), **not** on `GoalResults` ([GoalResults.java:21](https://github.com/dotCMS/core/blob/main/dotCMS/src/main/java/com/dotcms/experiments/business/result/GoalResults.java#L21)). A factory whose public API returns `GoalResults` has nowhere to put it, so the spec as written produces an unimplementable plan.
- **What must change:** redefine the seam's return type in Key Entities — either the full experiment result, or an explicit `GoalResults` + `TotalSession` pair — and have FR-006a state what each call yields (aggregate vs. per-day).

**B1b — [CRITICAL] Bayesian input is not covered by any requirement**

- The Bayesian calculation is gated on `experimentResults.getSessions().getVariants().size() >= 2` ([ExperimentsAPIImpl.java:1355](https://github.com/dotCMS/core/blob/main/dotCMS/src/main/java/com/dotcms/experiments/business/ExperimentsAPIImpl.java#L1355)) and fed through `BayesianHelper.toBayesianInput`. No FR covers populating that from CAEM. With the seam as specified, the sessions block would come back empty and Bayesian results would silently switch off while the flag is enabled — a behavioral regression the spec claims cannot happen (FR-013, SC-002).
- **What must change:** add an FR requiring `TotalSession` / Bayesian parity with the CubeJS path when the switch is enabled.

**Open question that resolves both:** does the seam return `GoalResults` alone, or the full result including `TotalSession`? And does CAEM deliver the per-day rows and the aggregate in a single call, or two? Today those are two separate queries (`create` and `createWithDayGranularity`, [ExperimentResultsQueryFactory.java:106](https://github.com/dotCMS/core/blob/main/dotCMS/src/main/java/com/dotcms/experiments/business/result/ExperimentResultsQueryFactory.java#L106) and L148) — the answer changes both the HTTP client contract and the number of calls, so it is worth settling in `/speckit-clarify` before planning.

erickgonzalez (dismissed):

jcastro-dotcms (changes requested):
## Additional finding — Exit Rate / Bounce Rate optional page condition: inconsistent, and a possible CubeJS parity break

**🟠 High (parity/correctness):** `dotCMS/src/main/java/com/dotcms/experiments/business/result/BounceRateCAEMResultQuery.java:64-73` vs `dotCMS/src/main/java/com/dotcms/experiments/business/result/ExitRateCAEMResultQuery.java:68-79`

Both `EXIT_RATE` and `BOUNCE_RATE` support the same optional page-scoping condition (`MetricType.java:35-41` — both declare `.optionalParameters(Parameters.URL)`), but the two new CAEM query classes handle it differently:

- `ExitRateCAEMResultQuery.buildParams` reads the condition via `GoalConditionUtil.findConditionValue(experiment, "url")` and sends it to CAEM as `referencePage`, which **filters** sessions to that page.
- `BounceRateCAEMResultQuery.buildParams` never calls `GoalConditionUtil` at all — the identical optional condition is silently ignored. There's also no test case in `BounceRateCAEMResultQueryTest` for a configured URL condition.

That inconsistency is worth fixing on its own, but there's a bigger question underneath it: **in the current CubeJS path, this optional condition has no effect on the numbers at all, for either goal type.**

- `MetricExperimentResultsQuery.getCubeJSQuery(Experiment)` doesn't even receive the `Metric`/`Condition` — the interface only takes `Experiment`.
- `BounceRateResultQuery`/`ExitRateResultQuery` only set static CubeJS measures; no filter/dimension is derived from the condition.
- `MetricsUtil.validateGoals()` only validates that the parameter name is legal for the metric type — it never becomes a query filter.
- `ExperimentUrlPatternCalculator`'s regex path doesn't apply either, since `EXIT_RATE`/`BOUNCE_RATE` never set a `regexParameterName` (unlike `REACH_PAGE`/`URL_PARAMETER`).

So today, an Exit Rate goal's optional page condition is validated on save but never actually filters the CubeJS numbers — exit rate is always computed across all pages. `ExitRateCAEMResultQuery` changes that: with `FEATURE_FLAG_CAEM_EXPERIMENT_RESULTS` on, an experiment that has this condition set will suddenly have its exit-rate numbers filtered to just that page — a behavior change with no code change on the admin's part, and exactly the kind of backend-swap parity break this PR is meant to avoid.

**Suggested fix:** confirm with the CAEM API owner whether page-scoping for Exit Rate is an intentional improvement over the legacy behavior:
- If yes — apply the same `referencePage` extraction to `BounceRateCAEMResultQuery` for consistency, and add a spec/FR note plus a test documenting the intentional behavior change vs. CubeJS.
- If no — drop it from `ExitRateCAEMResultQuery` so both backends stay numerically equivalent for existing experiments, matching current (no-op) legacy semantics on both.

Either way this should be a deliberate, tested decision rather than an implicit divergence between two otherwise-identical sibling classes.

freddyDOTCMS:
> * 🟡 Medium: `dotCMS/.../result/CaemHttpClient.java:88-101` — `DOT_ANALYTICS_TENANT` is read and validated as required, but `buildUrl` (196‑216) only appends `project`; `tenant` is never sent or otherwise used here. Per `EventAnalyticsProxyHelper` (lines 37‑38) tenant is only used to build the admin token, which this client does **not** use (it uses the per‑site bearer token). So the mandatory `tenant` check will reject a correctly-configured CAEM setup that has no tenant. Confirm tenant is actually needed on this path; if not, drop the check.

@claude take this in account

- this one "🟡 Medium: dotCMS/.../result/CaemHttpClient.java:88-101 — DOT_ANALYTICS_TENANT is read and validated as required, but buildUrl (196‑216) only appends project; tenant is never sent or otherwise used here. Per EventAnalyticsProxyHelper (lines 37‑38) tenant is only used to build the admin token, which this client does not use (it uses the per‑site bearer token). So the mandatory tenant check will reject a correctly-configured CAEM setup that has no tenant. Confirm tenant is actually needed on this path; if not, drop the check.", no matter the tenant filter is not need it
- DOnt worry about this one "🟠 High: dotCMS/src/main/java/com/dotcms/experiments/business/result/ExitRateCAEMResultQuery.java:76 — ExitRateCAEMResultQuery sends referencePage (the experiment page URI), which scopes exit-rate sessions to that page. The legacy CubeJS ExitRateResultQuery.java:25-27 applies no page filter (root query filters only by experiment/runningId/variant), and BounceRateCAEMResultQuery sends no referencePage at all. So flipping FEATURE_FLAG_CAEM_EXPERIMENT_RESULTS on silently changes exit-rate numbers for existing experiments — the exact parity break @jcastro-dotcms flagged, still unresolved. No test asserts referencePage for exit rate, and there's no spec note documenting the intended change. This needs to be a deliberate, tested decision."
- About this one "🟡 Medium: dotCMS/src/main/java/com/dotcms/experiments/business/result/BounceRateCAEMResultQuery.java:64 — BOUNCE_RATE and EXIT_RATE both declare optionalParameters(URL), but BounceRateCAEMResultQuery.buildParams never touches GoalConditionUtil while ExitRateCAEMResultQuery does. The identical optional condition is handled inconsistently between the two sibling classes, and there's no BounceRateCAEMResultQueryTest case for a configured URL condition. Resolve alongside the High finding above so both goal types stay parity-equivalent.", Bounce_RATE does not send any page filtering and it is ok

freddyDOTCMS:
@claude This one "🟡 Medium: dotCMS/.../result/ExperimentResultsQueryFactory.java:98-118 — executeAggregate and executeByDay each read isCaemExperimentResultsEnabled() independently. ExperimentsAPIImpl.getResults() (ExperimentsAPIImpl.java:1306-1307) calls getTotalSessions() (→ executeAggregate) and then getSummary() (→ executeByDay) as two separate calls. If the flag is toggled between those two reads, one result set comes from CAEM and the other from CubeJS, and they are combined into a single ExperimentResults — mixed-backend numbers for one report. Narrow window (live toggle only), but it's a direct consequence of the per-call flag read. Consider resolving the impl once per getResults and passing it to both fetches. [Fix this →](https://claude.ai/code?q=In%20ExperimentResultsQueryFactory%2C%20resolve%20the%20ExperimentGoalResultsQuery%20impl%20once%20per%20getResults%20call%20so%20executeAggregate%20and%20executeByDay%20cannot%20straddle%20a%20feature-flag%20toggle&repo=dotCMS/core)", it is not so bad, we are not ogint o keep the 2 mode for long time

jcastro-dotcms (dismissed):

jcastro-dotcms (approved):

jcastro-dotcms (approved):

PR #37297, opened 2026-08-31
Spec: silent index delete loss (#37276)

### LINKED ISSUE #37276: Content deletes are lost silently: index removal is fire-and-forget with no durable record, unlike adds
## Parent Issue

N/A — found during a support investigation on a customer environment (details anonymized).

## Problem Statement

Content deletes are propagated to the search index as an **in-memory, post-commit, fire-and-forget** task with **no durable record of intent**. The database rows are deleted first; if anything goes wrong afterwards — a JVM restart, a rejected pool submission, or a partially failed bulk — the index document is never removed and **nothing retries it**. The result is an index document with no contentlet behind it, invisible to the application and undetectable without an external count comparison.

This is the mirror image of the add path, which writes its intent to `dist_reindex_journal` inside the transaction and is retried by `ReindexThread` until acknowledged. A durable delete mechanism already exists in the journal — and has **zero production callers**.

### Observed in the field

A customer Production environment served a search widget whose result count (taken from the index) disagreed with the number of contentlets actually loaded from the database. Comparing `esresponse.hits.hits.length` against `contentlets.length` for the same query isolated a set of index documents whose `_source.inode` resolved to nothing; `/api/content/id/{id}` returned empty for each. Staging, running the same content, showed essentially none.

Ruled out by direct check: content drift (both environments shared the same real identifiers), permissions (identical anonymous role set), multi-language, and `track_total_hits` truncation (`hits.total.relation == "eq"` throughout). No `Unable to convert Contentlet to JSON` in the logs on either node, so these were not serialization casualties dropped by `ESContentResourcePortlet`'s per-hit catch.

The affected index had been built by a **full reindex**, and a freshly built index cannot contain orphans at creation — it is populated from the database. So the documents were valid when written and the content was deleted afterwards without the index being updated. This is a **delete-propagation failure**, not a stale or partial reindex.

### This is not an argument against the commit-listener idiom

Worth stating up front, because it is the first objection this will draw: **deferring the index write to a commit listener is correct**, and this issue does not propose changing it. Indexing before the commit would risk publishing content that is then rolled back. The listener is the right pattern and should stay.

The defect is that the listener is the **only** record. If it is lost, nothing anywhere says "this index still owes a delete".

The related objection — *the database is the source of truth, so an index that disagrees is not really a data problem* — is true about where truth lives, but it does not make the divergence self-correcting. A source of truth only helps if some process re-derives the copy from it. For adds that process exists: `dist_reindex_journal` + `ReindexThread`. **For deletes there is none.** The divergence is permanent until someone runs a full reindex by hand.

And it does reach users, because the application does not read from the source of truth — it reads from the index. Search results, listings, URL maps and widget counts are all served from it. The observed symptom is exactly that:

```java
// ESContentFactoryImpl.java:1336-1337
return inodes.stream().map(inode -> conMap.get(inode)).filter(Objects::nonNull)
```

The index returns N hits, the database resolves N−k, and `filter(Objects::nonNull)` drops the k orphans silently — while `hits.total` still reports N. A widget renders "N results" above N−k rows. No error is logged and nothing surfaces; the user just sees a wrong number and a short list.

The strongest argument that "the database is enough" does not hold is that this codebase already rejected it: if a lost index write were tolerable because the database is authoritative, `dist_reindex_journal` would not exist. It was built precisely because losing an index write matters. The delete path simply was never wired to it — the machinery is implemented, tested, has both ES and OS consumers, and has zero production callers (see the table below).

So the fix is neither "make the index part of the transaction" (impossible — the search index is not transactional) nor "stop using commit listeners". It is: record the intent durably **inside** the transaction, the way the add path already does, and let the commit listener remain the fast path.

## Related ticket: 

https://dotcms.freshdesk.com/a/tickets/38711

## Steps to Reproduce

The three loss points below suggest different experiments. In all of them the assertion is the same: query the live index for the deleted identifier and confirm the document is still returned while the contentlet no longer loads.

1. **L3 — bulk rejection under load (most likely).** Shrink the ES write thread pool / queue size to saturate it, then delete content in bulk. Expected: rows gone, documents remain, one `Error reindexing` line per failed batch.
2. **L1 — restart in the commit window.** Delete content, then stop the JVM immediately after commit but before the listener pool drains (breakpoint or injected pause in `DotRunnableThread.run()`). Expected: rows gone, documents remain, **no log line at all**.
3. **L1 — pool rejection.** Fill `dotListenerSubmitter` so `submit()` is rejected while deleting. Expected: rows gone, documents remain; the rejection may or may not log.
4. **L2 — index pointers unavailable.** Make `loadProviderIndices` throw for the primary provider during a delete. Expected: rows gone, documents remain, distinctive `Could not load provider indices` warn.

## Acceptance Criteria

- [ ] A content destroy whose index removal fails transiently (bulk rejection, provider unavailable, JVM restart between commit and listener execution) is **retried** and eventually removes the document — it does not leave a permanent orphan.
- [ ] `ContentletIndexOperationsES.putToIndex` no longer treats `response.hasFailures()` as success; a partial bulk failure is surfaced to the caller.
- [ ] A `null` `ProviderIndices` for the **primary** provider during a delete no longer looks like a completed delete. (Shadow providers in dual-write phases keep warn-and-continue, consistent with how `putToIndex` already isolates OS failures.)
- [ ] The bulk failure log message distinguishes deletes from reindexes — today a delete failure logs as `Error reindexing`, which is why a log search for delete failures comes back empty.
- [ ] The stale javadoc on `IndexPolicyProvider.forSingleContent()` / `forContentDependencies()` claiming a `WAIT_FOR` default is corrected to `DEFER`.
- [ ] Integration coverage for at least the L3 path (partial bulk failure → retry → document removed).

## Technical Analysis

All line references verified against `origin/main` @ `88af0bad55`.

### The delete path as it actually executes

`ESContentletAPIImpl.destroyContentlets()` deletes the DB rows first, then asks the index API to remove the document. The removal is **not part of the transaction**:

```java
// ESContentletAPIImpl.java:3121-3133
this.contentFactory.delete(contentletsVersion);        // ← DB rows gone

final Set<String> removedFromIndex = new HashSet<>();
for (final Contentlet contentlet : contentletsVersion) {
    final String documentKey = contentlet.getIdentifier() + StringPool.UNDERLINE
            + contentlet.getLanguageId() + StringPool.UNDERLINE + contentlet.getVariantId();
    if (removedFromIndex.add(documentKey)) {
        this.indexAPI.removeContentFromIndex(contentlet);   // ← deferred
    }
    CacheLocator.getIdentifierCache().removeFromCacheByVersionable(contentlet);
}
```

Whether that removal runs inline or is deferred depends on the contentlet's index policy, and the default is `DEFER`:

```java
// ContentletIndexAPIImpl.java:3051-3071
final boolean indexIsNotDefer = IndexPolicy.DEFER != content.getIndexPolicy();
if (indexIsNotDefer) {
    this.handleRemoveIndexNotDefer(content, onlyLive, relationships);
} else {
    HibernateUtil.addCommitListener(
            content.getInode() + ReindexRunnable.Action.REMOVING,
            new RemoveReindexRunnable(content, onlyLive, relationships));   // ← default path
}
```

`Contentlet.getIndexPolicy()` (Contentlet.java:248-253) falls back to `IndexPolicyProvider.getInstance().forSingleContent()`, which reads `INDEX_POLICY_SINGLE_CONTENT` with a default of `"DEFER"` (IndexPolicyProvider.java:35). So the commit-listener branch is the **normal production path** for content deletion, not an edge configuration.

Execution order — everything after step 1 is work still owed to the index while the database is already inconsistent:

```
contentFactory.delete(versions)  → rows removed
addCommitListener(RemoveReindexRunnable)
commit → submitter.submit(DotRunnableThread)              ← L1
removeContentAndProcessDependencies() builds delete ops   ← L2
putToIndex() → bulk delete                                ← L3
```

### L1 — the removal is submitted fire-and-forget after commit

`HibernateUtil.addCommitListener` routes a `ReindexRunnable` into `asyncCommitListeners` when `ASYNC_REINDEX_COMMIT_LISTENERS` is true — and it **defaults to true** (`HibernateUtil.java:1412-1413`):

```java
// HibernateUtil.java:1052-1053
} else if (listener instanceof ReindexRunnable && asyncReindexCommitListeners()) {
    asyncCommitListeners.get().put(tag, listener);      // ← removals land here
}

// HibernateUtil.java:1144-1147
final DotSubmitter submitter = DotConcurrentFactory.getInstance().getSubmitter(LISTENER_SUBMITTER);
submitter.submit(new DotRunnableThread(listeners, true));   // return value discarded
```

The task runs on a shared pool **after** the transaction has committed and the returned handle is dropped, so nothing observes completion or failure. Three ways the removal disappears:

- The JVM stops between commit and execution — a rolling deploy on a multi-node cluster is exactly this window.
- The submitter rejects the task under saturation.
- `DotRunnableThread.run()` throws — it wraps failures in `DotStateException` (`DotRunnableThread.java:51-54`) and propagates into an executor where no one is listening.

### L2 — a provider whose indices won't load is skipped, and the delete op is never built

```java
// ContentletIndexAPIImpl.java:335-341
private ProviderIndices loadProviderIndicesQuietly(final ContentletIndexOperations ops) {
    return Try.of(() -> loadProviderIndices(ops))
            .onFailure(e -> Logger.warn(this.getClass(),
                    "Could not load provider indices for " + ops.getClass().getSimpleName()
                            + " — writes to this provider will be skipped. Cause: " + e.getMessage()))
            .getOrElse((ProviderIndices) null);
}

// caller — ContentletIndexAPIImpl.java:3135-3138
final ProviderIndices indices = loadProviderIndicesQuietly(ops);
if (indices == null) {
    continue;                       // ← delete silently not queued
}
```

`putToIndex` early-returns on an empty batch, so a skipped provider is indistinguishable from a completed delete. The same `null → continue` shape appears at `:2761`, `:2855`, `:2995`, `:3024` and `:3135`.

**Least likely of the three** — it needs an index-pointer load failure, which would probably surface other symptoms too. It does leave a distinctive log line, which makes it cheap to confirm or eliminate.

### L3 — a partial bulk failure is logged as an error and then treated as success

```java
// ContentletIndexOperationsES.java:197-218
final BulkResponse response = RestHighLevelClientProvider.getInstance()
        .getClient().bulk(bulkRequest, RequestOptions.DEFAULT);

if (response != null && response.hasFailures()) {
    Logger.error(this,
            "Error reindexing (" + response.getItems().length + ") content(s): "
                    + response.buildFailureMessage());       // ← no throw, no retry
}
```

Only a thrown exception from `client.bulk(...)` propagates. **Per-item** rejections — `EsRejectedExecutionException` from a saturated write queue, an unavailable shard, a version conflict — take the logged branch and the caller cannot distinguish a fully applied batch from one where every item was rejected. Since the DEFER path sets `RefreshPolicy.NONE` and nothing re-reads the document, the loss is invisible from the application side.

**Most likely of the three.**

> **Practical note for log searches:** that message says `"Error reindexing"` even when the failed operations were deletes. Searching production logs for delete- or remove-flavoured wording finds nothing — `grep "Error reindexing"` is the string that surfaces this.

### Why adds survive this and deletes don't

The asymmetry is the real defect, and it is visible in **the same class, on the same `IndexPolicy.DEFER` check**, resolved two different ways:

```java
// ContentletIndexAPIImpl.java:2308-2309  — ADD
if (parentContenlet.getIndexPolicy() == IndexPolicy.DEFER) {
    queueApi.addContentletsReindex(contentToIndex);   // → dist_reindex_journal row
}

// ContentletIndexAPIImpl.java:3055-3068  — DELETE
final boolean indexIsNotDefer = IndexPolicy.DEFER != content.getIndexPolicy();
if (indexIsNotDefer) {
    ...
} else {
    HibernateUtil.addCommitListener(                  // → ThreadLocal map entry
            content.getInode() + ReindexRunnable.Action.REMOVING,
            new RemoveReindexRunnable(content, onlyLive, relationships));
}
```

Both execute the index operation **outside** the transaction — unavoidable, the search index is not transactional. The difference is what is left behind if that operation never lands: the add leaves a durable row that `ReindexThread` drains and retries until acknowledged; the delete leaves an object in a thread-local map that does not survive the process.

| | Add (checkin) | Content delete (destroy) |
|---|---|---|
| DB write | transactional | transactional |
| Index intent recorded | `dist_reindex_journal` row, **in-transaction** | `ThreadLocal` map entry |
| Index execution | `ReindexThread` drains the table, after commit | commit listener on a shared pool, after commit |
| Survives restart | yes | no |
| Retried on failure | yes | no |
| Failure visibility | row persists | log line only |

The ordinary checkin entry point confirms the add side: `ESContentletAPIImpl.java:1297` calls `indexAPI.addContentToIndex(contentlet, INCLUDE_DEPENDENCIES)` — the two-arg overload — which takes the DEFER branch above and journals.

> **Caveat when reading this code:** `DotRunnableThread.java:141` calls the **`List` overload** `addContentToIndex(List<Contentlet>)` (`ContentletIndexAPIImpl.java:2337`), which goes straight to `putToIndex` with **no journal row**, and swallows `DotDataException` into `Logger.error` (`DotRunnableThread.java:145-147`). That is a legacy path for `ReindexRunnable` entries, not the normal checkin — but it means "adds are always journaled" is not universally true, and anything routed through it has the same L1/L3 exposure as deletes.

### A durable delete path already exists and is dormant

`dist_reindex_journal` already supports delete entries end to end. Everything is implemented and covered by integration tests — **nothing in production calls it**:

| Stage | Location | Status |
|---|---|---|
| Action enum | `ReindexQueueFactory.java:69-71` — `NONE, REINDEX, DELETE` | exists |
| Enqueue | `ReindexQueueFactory.addIdentifierDelete()` `:470-489` | exists |
| Public API | `ReindexQueueAPI:200, :210` · `ReindexQueueAPIImpl:178, :188` | exists |
| Read-back | `ReindexQueueFactory.java:361` — `.isDelete(dist_action == DELETE.ordinal())` | exists |
| ES consumer | `ContentletIndexAPIImpl.java:2545, :2559` | exists |
| OS consumer | `OSBulkHelper.java:192` | exists |
| Test coverage | `ContentletIndexAPIImplTest:1339` · `ReindexAPITest:322` | exists |
| **Production caller** | **— none —** | **missing** |

## Proposed Solution

**F1 — Journal content deletes instead of deferring them in memory (addresses L1, highest value).**
In `destroyContentlets()`, enqueue a `ReindexAction.DELETE` entry in the **same transaction** that deletes the rows, and let `ReindexThread` apply it. The commit-listener path can stay for the low-latency case, but the journal row is what makes the operation survive a restart or a rejected submission. Deleting the journal row only after an acknowledged bulk gives deletes the same at-least-once guarantee adds already have.

> ⚠️ **Scope F1 to destroy only.** `appendBulkRemoveRequestInternal` (`ContentletIndexAPIImpl.java:2986-3016`) builds the document id as `identifier_lang_variant` by iterating **all** languages × **all** variants — a journal entry is identifier-wide. That is correct for `destroyContentlets`, which destroys every version and language of the identifier by design, but it must **not** be reused for the unpublish/archive path, where it would remove languages that are still live.

**F2 — Make a partial bulk failure a failure (addresses L3, small change).**

```diff
  if (response != null && response.hasFailures()) {
      Logger.error(this, "Error reindexing (" + ... );
+     throw new DotIndexException(response.buildFailureMessage());
  }
```

With F1 in place, throwing is enough — the journal row survives and the entry is retried. Without F1, throwing at least surfaces the problem instead of hiding it. Split the message so deletes don't report as `"Error reindexing"`.

**F3 — Don't let a skipped provider look like a completed delete (addresses L2, small change).**
When `loadProviderIndicesQuietly` returns `null` for the **primary** provider, the operation has not been performed — escalate from `Logger.warn` + `continue` to a thrown exception. Shadow providers in dual-write phases keep warn-and-continue.

**F4 — Add a reconciliation check so drift is detectable (operational, independent of the fix).**
The index document `_id` maps one-to-one onto `contentlet_version_info`'s primary key (`identifier`, `lang`, `variant_id`), so a live-index document count compared against `SELECT count(*) FROM contentlet_version_info WHERE live_inode IS NOT NULL` yields the exact orphan count in two cheap queries. Surfacing that in the Maintenance portlet's index view turns this class of bug from a customer report into an observable number.

### Secondary site with the same shape

`ESContentletAPIImpl.java:3549-3554` runs `contentFactory.delete(...)` followed by `indexAPI.removeContentFromIndex(...)` in a loop **without** the `documentKey` dedupe used in `destroyContentlets`. Same L1/L2/L3 exposure, plus a repeated removal per version. Should be fixed alongside.

## Open Questions

1. **Does the same gap exist on the archive / unpublish path?** `removeContentFromLiveIndex()` reaches the identical `removeContentFromIndex(content, true)` code, so L1–L3 should apply there too — with the difference that an archived contentlet still has database rows, so the symptom would be a *stale live document* rather than a fully orphaned one. Note the F1 caveat above: the journal fix as designed does not transfer to this path.
2. **Is push-publish's delete/unpublish bundle handling subject to the same path?** Not traced.
3. **Why was the observed leak type-concentrated?** Simple delete-volume skew is the boring explanation and probably the right one, but if deletes for that type run through a bulk operation with a different index policy, that would be a sharper reproduction lead.

## Scope / Confidence

The three loss points and the dormant journal path are **read from the code and verified** against `origin/main` @ `88af0bad55`. Attributing the specific orphans observed in the field to any one of L1–L3 is **inference from the end state**, not from a captured failure.

A full reindex clears existing orphans and is the immediate remediation — it does not address recurrence.

fabrizzio-dotCMS:
Defect spec for #37276. Implementation follows in PR 2.

This PR fixes: #37276

This PR fixes: #37276


specs/37276-silent-index-delete-loss/spec.md
```diff
@@ -0,0 +1,223 @@
...
+- **AC-001**: For each reproduction path above, the index document for a destroyed contentlet
+  is eventually removed. The removal survives the triggering condition rather than being lost
+  to it.
+- **AC-002**: After a destroy under a forced index-write failure, a subsequent search returns
+  a total that matches the number of contentlets that actually resolve — no inflated count.
+- **AC-003**: An index write that comes back with per-item failures is surfaced to its caller
+  rather than logged and reported as success.
+- **AC-004**: A failed removal is findable in the logs by searching for removal- or
+  delete-flavoured wording, not only by knowing to search for the word "reindexing".
+- **AC-005**: Regression — content addition, publication and reindexing continue to behave as
```
nollymar, line 239: This AC is really important to track, as the `ReindexQueueFactory` assumes that every record in the dist_reindex_journal is an insert, not a deletion

nollymar (dismissed):

specs/37276-silent-index-delete-loss/spec.md
fabrizzio-dotCMS: Confirmed, and I found the precise mechanism — thanks, this is a real gap.

The write and read sides are actually fine: `addIdentifierDelete` stamps `dist_action = DELETE.ordinal()` (`ReindexQueueFactory:480-482`), `mapToReindexEntry` reads it back (`:361`), `markAsFailed` is an `UPDATE` that leaves `dist_action` untouched so the flag survives a retry, and both consumers branch on it (`ContentletIndexAPIImpl:2545/:2559`, `OSBulkHelper:192`).

The assumption is in `findContentToReindex` (`ReindexQueueFactory:288-308`):

```java
Map<String, ReindexEntry> contentToIndex = new HashMap<>();
for (ReindexEntry entry; (entry = queue.poll()) != null; ) {
    contentToIndex.put(entry.getIdentToIndex(), entry);   // keyed by identifier alone
```

The batch is keyed by identifier, but `ReindexEntry` equality includes the delete flag (`identToIndex`, `priority`, `isDelete`, `serverId` — `id`/`lastResult`/`timeEntered` are `@Value.Auxiliary`). So a REINDEX entry and a DELETE entry for the same identifier are **not** equal — the "drain duplicate items" loop right below will not collapse them — but they collide on the map key, and whichever is polled last silently overwrites the other.

That matters specifically for the fix proposed here. Today the collision is unreachable because nothing enqueues deletes. Once `destroyContentlets` does, the colliding pair becomes the *normal* case: content is saved (REINDEX entries queued), then destroyed (DELETE entry queued for the same identifier). If the REINDEX entry wins the map, the removal is skipped for that batch — the exact failure this issue is about, reintroduced through the fix.

The losing row is not lost permanently — it stays in `dist_reindex_journal` and gets re-picked once `lastIdIndexed` resets — but the ordering is arbitrary and a stale REINDEX can be applied after a DELETE, re-adding the document.

I have not amended the approved spec for this, since it lands squarely in what the spec already flags under Assumptions ("the existing journal delete machinery works as its integration tests assert — planning should verify this rather than assume it, given it has never run in production"). I will carry it into `/speckit-plan` as a required design input: the batch key has to distinguish a removal from a reindex, and the two must be ordered rather than deduplicated. Happy to amend the spec and re-request review instead if you would rather it be part of the approved contract.

fabrizzio-dotCMS: Two amendments from the pre-implementation analysis, both pushed and both needing your eyes before I build.

**1. AC-001 overclaimed the guarantee.** It said the removal "survives the triggering condition". It does not, past a point: `ReindexQueueFactory.markAsFailed:270-273` sets the priority to `ERROR.dbValue() + priority` once a row has failed `REINDEX_MAX_FAILURE_ATTEMPTS` times, and the drain query (`:329-338`) selects `priority <= 400` with `ERROR.dbValue()` being 400. An exhausted removal sits above the filter and is never picked again.

The fix still converts a permanent silent orphan into a durable, queryable record of work owed — but that is a different promise than the one the AC made. AC-001 is reworded to at-least-once **up to** `REINDEX_MAX_FAILURE_ATTEMPTS`, and a new **AC-007** requires that an exhausted removal stays discoverable: the journal row remains above `ERROR` priority with the identifier and the last failure cause, so the set of removals still owed can be read with one query. That is the reconciliation signal the spec was already asking for — after this, drift is not merely detectable by counting index docs against `contentlet_version_info`, it is enumerable from the journal with a cause attached. Retrying past exhaustion and repairing the residue are added as an explicit non-goal.

**2. Your batch-collision finding is now in the spec.** It was driving a blocking phase in the task list while tracing to nothing in the approved document, which would have put an unasked-for legacy edit in the implementation PR. It is now an in-scope item with **AC-008** (a pending removal and a pending reindex for the same identifier resolve deterministically to the newer, and the older is not applied afterwards), plus a blast-radius note: that path carries every journal entry including full reindex, so the existing deduplication of repeated identical entries must be preserved — collapsing it differently would turn a redundant-entry optimisation into a per-entry bulk operation.

The plan and data model already reflect both, so no other artifact changes. Re-requesting review.

nollymar (dismissed):

ihoffmann-dot:
Three gaps I found:

- Secondary dedupe-less site at ESContentletAPIImpl.java:3549-3554 (same L1/L2/L3 shape, mentioned in the original issue) isn't addressed as in-scope or out-of-scope here. Please make a call either way.

- F4 (reconciliation check in the Maintenance portlet) from the issue is also silently dropped, not in scope, not in non-goals. Was that intentional?

- AC-003's "enumerate callers that tolerate partial index failures" is called a prerequisite but has no AC or verification step backing it. As written it can slip through planning unnoticed, needs its own AC or explicit task.

ihoffmann-dot (dismissed):

fabrizzio-dotCMS:
All four gaps from review are closed — three in the spec (945d27e6), one in code plus spec.

@ihoffmann-dot, your three:

- **The `:3549-3554` site.** Call made, both ways: `deleteAllVersionsandBackup` **is** journalled (its version list comes from `findAllVersions(identifier)`, so an identifier-wide entry is correct). Implementing it turned up a **third** site the issue never mentioned — `ContentletAPI#delete(List, User, boolean, boolean)` — which is now an explicit non-goal, because it can delete a *subset* of an identifier's versions and an identifier-wide entry would drop languages that still exist. Both are marked in the code so neither gets "completed" by mistake later.
- **F4.** Fair hit — I dropped it silently. Now a stated non-goal: AC-007 already makes the residue enumerable with one query, which is the data that portlet view would render, so the view itself is separate work and does not gate this fix.
- **AC-003's caller enumeration.** You were right that prose in Regression Risk was not enough to survive planning. It is now **AC-009**, stated as a prerequisite for AC-003 rather than a follow-up.

The claude[bot] review also found something sharper than a spec gap, and it turned out to be real: from **Phase 2** onwards `PhaseRouter#readProvider` serves reads from OpenSearch while writes stay ES-primary / OS-shadow — so a removal lost on the OS leg orphaned a document in the index being queried. That is this defect, in the phase the migration spends the longest in, and neither the durable journal nor the partial-failure escalation covered it, because the shadow listener acked the journal entry on the ES result alone.

Fixed in #37333 by scoping the shadow treatment to *who serves reads* rather than to dual-write, with **AC-010** added here. ADR-0009's intent is preserved — Phase 1 still swallows, and there is a test pinning that so the fix cannot over-reach.

fabrizzio-dotCMS:
Closing out the two 🟡 findings from the automated review — both were real and both are fixed.

**1. Blank exception message in `ContentletIndexOperationsOS.handleBulkResponse`.** Fixed in #37333. The invariant you flagged (`errors()` implies at least one item with a cause) is almost certainly true, but I did not want to rely on it: a failure with no message is barely better than the silent return it replaced, so there is now a fallback message when no item carried a cause.

**2. Phase 2 read/shadow gap.** Confirmed, and it was the sharpest thing in either review. `PhaseRouter#readProvider:144` serves reads from OpenSearch from Phase 2 onwards while writes stay ES-primary / OS-shadow, so a removal lost on the OS leg orphaned a document in the index actually answering queries — #37276 itself, in the phase the migration spends the longest in.

You suggested it might be intentionally out of scope under ADR-0009 and that a non-goal note would close the gap. I went the other way and fixed it, because the ADR's *rationale* is "a store nobody reads from must not break a user operation" — and in Phase 2 that premise no longer holds. Scoping the shadow treatment by `isReadEnabled()` rather than by dual-write preserves the intent and corrects only the stale assumption.

Worth noting the async path mattered more than the sync one you pointed at: `createBulkProcessor` built a shadow `BulkProcessorListener` for OS in Phases 1 **and** 2, and a shadow listener never marks the reindex-queue entry failed — so the journal entry was acked on the ES result alone and the removal was never retried. The durable-delete fix would not have covered Phase 2 without this.

Fix in #37333 with `ContentletIndexAPIImplPhase2ReadDurabilityTest`, including a case pinning that Phase 1 still swallows so the change cannot over-reach. Spec updated with AC-010; the contract's per-phase table and the release note were both stale and are corrected there too.

ihoffmann-dot (dismissed):

fabrizzio-dotCMS:
@ihoffmann-dot @nollymar — re-requesting review. Your approval was dismissed by a new commit (`1fc4f23b`), and the reason is a regression I introduced.

**What broke.** `Integration Tests - MainSuite 1a` failed on this PR with `Reindex Queue is not empty after 30s`, taking down `HostBundlerHandlerTest` (4/4) and `Task240306MigrateLegacyLanguageVariablesTest` (3/5). Neither is related to this feature — both die on the same `assertEmptyQueue`.

**Why.** `journalContentDeletes`, from the implementation merged here in #37320, wrote a `dist_reindex_journal` row on **every** destroy, ignoring the contentlet's `IndexPolicy`. The add path I modelled it on is conditional (`ContentletIndexAPIImpl:2308`): it journals only under `DEFER` and indexes inline otherwise.

That asymmetry was the bug. The durable record exists to protect the `DEFER` route, where the removal goes to an in-memory commit listener that can be lost. Under `FORCE` / `WAIT_FOR` the removal runs inline and synchronously — a failure already reaches the caller and rolls the transaction back, so the row is redundant work: one extra insert per identifier on the highest-volume delete path.

Most integration tests create content with `FORCE` / `WAIT_FOR`, so the journal was normally empty and `assertEmptyQueue` passed trivially. The reindex thread does not run in that harness — `beforeBulk` fired **zero** times across the whole 3-hour run — so every row written stayed there forever.

**Verified.** `HostBundlerHandlerTest` 4/4 green (was 0/4, and drops from 337s to 42s now that `assertEmptyQueue` stops burning its 30s timeout per case). `ReindexDeleteJournalTest` 3/3 still green, so the durable row is still written in-transaction for deferred removals — the guarantee is unchanged, only its scope is correct now. 281 unit tests green. Full diagnosis in `.claude/diagnostics/run-33557280710/DIAGNOSIS.md`.

**One thing changed for the better while fixing this:** the commit originally sat on the #37333 branch, which meant this PR stayed red until that one was reviewed. It belongs here — the regression is on this branch and has nothing to do with the Phase 2 work. Moved. **#37297 and #37333 are now independent**; this one no longer has to wait for that one.

dotcms-integration/src/test/java/com/dotcms/MainSuite1b.java
```diff
@@ -26,6 +26,9 @@
         com.dotcms.content.elasticsearch.business.ESSiteSearchAPITest.class,
         com.dotcms.content.elasticsearch.business.ESMappingAPITest.class,
         com.dotcms.content.elasticsearch.business.ContentletIndexAPIImplTest.class,
+        com.dotcms.content.elasticsearch.business.ContentletDestroyIndexRemovalTest.class,
```
nollymar: tests must be added always at the end of a mainsuite

dotcms-integration/src/test/java/com/dotcms/MainSuite2b.java
```diff
@@ -272,6 +272,7 @@
         com.dotmarketing.common.reindex.ReindexThreadTest.class,
         com.dotcms.content.elasticsearch.business.ContentletIndexAPIImplMappingTimeoutIT.class,
         com.dotmarketing.common.reindex.ReindexAPITest.class,
+        com.dotmarketing.common.reindex.ReindexDeleteJournalTest.class,
```
nollymar: same here

ihoffmann-dot (dismissed):

nollymar (changes requested):
@fabrizzio-dotCMS , did you test these changes with records to insert and delete from the index ( at the same time). I think more changes are required because the select operations in the ReindexQueueFactory are not considering the dist_action field

nollymar:
## Spec-alignment review (`/speckit-converge` assessment)

I ran a convergence assessment of the implementation against `specs/37276-silent-index-delete-loss/spec.md` in this same PR. The core of the fix is faithfully implemented and well-tested — durable journalling on destroy and `deleteAllVersionsandBackup`, partial-failure escalation in both providers (AC-003), removal-flavoured log wording (AC-004), the empty-primary check (L2), the batch-key `merge` by highest `id` with dedup preserved (AC-008), exhausted-removal discoverability (AC-007), and the AC-009 caller enumeration all check out against the spec.

**One item must be addressed before merge: AC-010's Phase 2 clause is not implemented, and the contract in this PR contradicts the spec on it.**

### The gap (AC-010, Phase 2)

The spec scopes the OpenSearch shadow treatment **by who serves reads, not by dual-write**:

> "OpenSearch stays fire-and-forget while nothing reads it, and becomes durable the moment it does."

and AC-010 states:

> "In a phase where OpenSearch serves reads, a failed OpenSearch write reaches the caller and, on the journal path, marks the entry failed so it is retried."

`PhaseRouter` defines Phase 2 as *OS reads, ES+OS writes*. But the router's `putToIndex` (`ContentletIndexAPIImpl`, dual-write branch ~L2415) swallows the OS leg's failure in **every** dual-write phase, keyed only on `DualIndexBulkRequest` — it never consults who serves reads. Consequence in Phase 2: the OS bulk fails, the ES leg succeeds, the journal entry is deleted as success, and the removal is permanently lost **in the very index being queried** — the defect this PR fixes, in the phase the spec notes the migration spends the longest in.

The spec pre-scoped the ADR-0009 objection: the "Phase 1 shadow-write durability" non-goal explicitly limits fire-and-forget deference to phases where nothing reads OS.

### Three things to address

1. **Code** — in the router's dual-write branch, propagate the OS failure when OS is the read provider (an `isPhase2()`-style check), keeping the Phase 1 swallow intact. This is what confines the blast radius exactly as the spec's Regression Risk section describes ("confined to phases where OpenSearch already serves reads").
2. **Contract** — `contracts/putToIndex-failure-contract.md`'s phase table says Phase **1 and 2**: OS partial failure "logged, swallowed by the router — ADR-0009", and argues router-level escalation "would break the phase 1–2 guarantee". That codifies the opposite of AC-010. Either the contract is corrected to match the spec, or — if dropping Phase 2 durability was a conscious decision — the spec must be amended and re-approved per the sign-off rule.
3. **Test** — `ContentletIndexPartialFailurePhaseTest` forces phase 1 only (`setPhase(1)`). Add a Phase 2 case asserting the OS failure reaches the caller and, on the journal path, marks the entry failed for retry.

Everything else converged cleanly: 8/8 in-scope items otherwise satisfied, 9/10 ACs verified, non-goals (unpublish path, commit-listener retention, journal left unrewritten) respected.

*Assessment produced with Claude Code (`/speckit-converge`), reviewed and posted by @nollymarlonga.*

fabrizzio-dotCMS:
@nollymar — all three points addressed, plus the one your question uncovered. Pushed in `c1ca3c9c` and `f568f9d1`.

### 1. AC-010 / Phase 2

The fix already existed in #37333, stacked on this branch. Rather than leave this PR internally inconsistent — its own `spec.md:255` declaring AC-010 while its contract denied it — I fast-forwarded #37333 in here (GitHub has marked it merged). All three items you listed are now in this PR:

- **Code** — `ContentletIndexAPIImpl`: the shadow decision is scoped by `isReadEnabled()`, not by dual-write. The async path mattered more than the sync one you pointed at: `shadow = isDualWrite && ops == operationsOS && !isReadEnabled()` in `createBulkProcessor`, because a shadow listener never marks the journal entry failed — so the entry was acked on the ES result alone and never retried.
- **Contract** — the phase table is split: Phase 1 swallows, Phase 2 propagates. The "router-level escalation would break the phase 1–2 guarantee" claim is gone.
- **Test** — `ContentletIndexAPIImplPhase2ReadDurabilityTest`, including a case pinning that Phase 1 still swallows so the change cannot over-reach.

### 2. `dist_action`, and the concurrent insert/delete

You were right, and the concurrency question turned out to be the sharper half.

**On the literal point.** `loadUpLocalQueue` does decode `dist_action` (`mapToReindexEntry:369`). `getFailedReindexRecords:199` does not — and `ReindexEntry.isDelete()` defaults to `false`, so every exhausted removal read back as a reindex. That is AC-007 not actually holding: the Maintenance portlet and `/api/es/index/failed` rendered a parked delete as pending reindex work, and the CSV resolved a contentlet that no longer exists, leaving the row blank. Both read paths now decode the column through one place.

**On the concurrency.** The problem is the acknowledgement, not the select. `deleteReindexEntry` cleared *every* row for the identifier:

```sql
DELETE FROM dist_reindex_journal where ident_to_index = ?
```

A bulk is asynchronous, so:

| T | |
|---|---|
| T0 | the batch loads the REINDEX row for X (id=99) |
| T1 | the bulk is in flight |
| T2 | X is destroyed → the journal gets the DELETE row (id=150) |
| T3 | the T0 bulk succeeds → deletes by identifier → **takes id=150 with it** |

The removal is never applied, and no record is left saying it was owed. #37276, re-entered through the acknowledgement instead of through the write. AC-008's `merge` does not reach it: the DELETE row did not exist when the batch was assembled, so there was no collision to resolve. That window had nothing to lose before this branch, because nothing enqueued removals — the fix is what makes the row load-bearing.

The ack is now bounded by the id it applied (`ident_to_index = ? AND id <= ?`). Superseded rows still go with the entry that supersedes them, so the dedup that keeps redundant entries off the bulk path is unchanged; only rows newer than the batch survive. I dropped the MySQL branch rather than fix it in two shapes — this is PostgreSQL-only.

**Verified red, then green.** 3 of the 4 new tests in `ReindexDeleteJournalTest` fail against the previous SQL, the mid-flight case reporting an empty journal — `All rows: [] expected:<1> but was:<0>`, the ack having taken both rows. 7/7 green with the fix; `ReindexQueueFactoryBatchKeyTest` 5/5. The fourth test passes either way by design: it guards the dedup against an over-correction to a strict `id = ?` ack.

Two stale comments went with it: `findContentToReindex` claimed the losing row is "collected on a later pass" (it is not — the ack takes it, correctly, since it has been superseded), and `ReindexEntry.isDelete`'s javadoc said failed records are never deletes, which stopped being true the moment removals were journalled.

I added a clause to AC-001 and to AC-007 rather than a new AC — both cases are defects against what those criteria already promised, so I do not think this needs a spec re-approval. Say the word if you read it differently.

### 3. Suite placement

Fixed in `f568f9d1`. The entries in `MainSuite1b` and `MainSuite2b` are at the end now.

Re-requesting your review.

dotcms-integration/src/test/java/com/dotcms/MainSuite1b.java
fabrizzio-dotCMS: Done in `f568f9d1` — moved to the end of the suite.

dotcms-integration/src/test/java/com/dotcms/MainSuite2b.java
fabrizzio-dotCMS: Done in `f568f9d1` — same, moved to the end.

nollymar (dismissed):

fabrizzio-dotCMS:
One correction to my last comment, found while writing the QA note.

I said the Maintenance portlet and `/api/v1/esindex/failed` now distinguish a parked removal from a parked reindex. They did not. Decoding `dist_action` fixed the value on `ReindexEntry`, but neither consumer rendered it — the endpoint returned `identifier`, `serverId`, `failureReason`, `priority`, and the portlet CSV had `ID, Identifier To Index, Priority, Cause`.

So an exhausted removal still surfaced as a reindex with a blank title, since the endpoint resolves the contentlet by identifier and a destroyed identifier resolves to nothing. That is the row an operator is actually holding, and it pointed at content that was deliberately destroyed.

Both now carry the operation — `d2c0e641`. AC-007 needs the flag to reach the surface, not just the object.

ihoffmann-dot (approved):

PR #37306, opened 2026-08-31
feat(experiments): change the page of a draft that has variants

### LINKED ISSUE #37003: Experiments Portlet — Screen 2: Create/Update (/experiments/new + /:id/configuration)

oidacra:
### Proposed Changes

Follow-up on the Configure screen (Screen 2) of the Experiments portlet: changing the Page of a draft that already has Variants.

https://github.com/user-attachments/assets/597811a1-ecd8-4690-937e-401ea7c1a502

* **Change Page replaces Clear.** With a page in place the Page card offered an X with a tooltip; it now offers a **Change Page** button. It is disabled only once the experiment is past draft, and then it explains itself with the same locked copy every other field uses.
* **Happy path untouched.** No non-control Variant — or no experiment yet, on `/experiments/new` — goes straight to the page picker, with no confirmation in the way.
* **A confirmation when Variants would be deleted.** `ExperimentsAPIImpl.save()` refuses a page change while a non-control Variant holds a copy of the current page, so the change deletes them first. A dialog names the page and every Variant by name, states that the action cannot be undone, and puts the deletion behind the primary **Delete Variants And Change Page**, against **Keep Current Page**. Only when they are all gone does the picker open; Cancel, the X and ESC leave the experiment exactly as it was — and none of the three is available while a deletion is on the wire, since the run does not stop with the dialog.
* **The deletions run one after another** (`concatMap`). Each has the backend recompute the traffic proportion over what is left, so only the answer to the last call describes the experiment afterwards — firing them in parallel would leave which answer is current up to arrival order.
* **A rejection halfway keeps what did go through.** The failure carries the experiment the last successful deletion answered with, so the Variants card stops offering weights for rows that no longer exist. It is reported unobtrusively — a toast, not an alert over the dialog, which would bury the retry — and the dialog stays open with the failure stated inline beside its own buttons.

Removed along with the X: the `pageCleared` event, and the two tooltip keys that became unreachable once a non-draft experiment reads with the generic locked copy. `VARIANT_COLORS` moved to `shared/constants.ts`, since the dialog draws the same Variants in the same colours as the card.

### Checklist
- [x] Tests
- [x] Translations

### Additional Info

**How the pieces are split.** The confirmation is a view: it renders what it is handed and reports the go-ahead as `pageChangeConfirmed`. The wait and the failure reach it as store signals through PrimeNG's `inputValues` — signals rather than values, because `inputValues` is applied once via `setInput` when the dialog is created, so a boolean would freeze at whatever it was on open. The Page card owns both the store and the `DynamicDialogRef`, so it is what closes the dialog on `deleteVariantsSucceeded`; the dialog is created outside the card's injector and cannot reach the store at all. Both dialogs go through `DialogService`, so the handover stays one `onClose` chain.

`deleteVariantsSucceeded` / `deleteVariantsFailed` are deliberately their own pair rather than a plural `removeVariant*`: the common case deletes exactly one Variant, so the payload cannot tell the two flows apart, and they differ in error surfacing (toast vs. the standard alert) and in state. Same precedent as `abortSucceeded` vs `cancelScheduleSucceeded` in this store — one call, two events, because the consequence differs.

**Two behaviours worth exercising by hand.** The deletion is irreversible and happens *before* the picker: confirming and then cancelling the picker leaves the Variants deleted and the page unchanged. And the new page is persisted by **Save Draft**, not by picking it — `pageId` travels in the footer's PATCH, as it already did.

**Verification.** 935/935 portlet tests, `nx lint` clean, `nx format:check` clean, `nx affected -t test` green across the 4 affected projects.

### ⚠️ Open question for the reviewer: this diverges from the signed-off spec of #37176

`specs/37176-draft-experiment-page-change/spec.md` — committed on `main` and reviewed twice (`7a280d6698`, `54d378bcb8`) — specifies the opposite affordance for this exact situation:

- **FR-009**: *"When a Draft experiment has at least one non-control variant, the page selection control MUST be disabled and MUST explain: 'This experiment already has variants. Delete them to change the page.'"*
- **User Story 3** ("Variants explain why the page is locked", P2), whose acceptance test is *"assert the control is disabled and carries exactly that explanation"*.
- The spec also declares that sentence as fixed copy backed by a language key — the very key this PR deletes (`experiments.configure.page.select.has-variants.tooltip`).

This PR ships **cascade-delete-then-proceed** instead of **disable-and-explain**. It is a deliberate product decision, not an oversight, and the backend rule is untouched: `ExperimentsAPIImpl.save()` still requires DRAFT + control-is-the-only-variant (`1fbdeb837d`), which is why the variants are deleted *first* rather than the rule being relaxed.

What that means for this review:

- Nothing in `specs/37176-draft-experiment-page-change/` is updated by this PR, so **the committed spec now contradicts what ships**. Read on its own, it will tell the next person the picker is disabled.
- Per `CLAUDE.md` ("If the spec changes after sign-off, get it re-approved"), the call belongs to whoever owns that spec: either FR-008/009/010 + User Story 3 get rewritten around the new flow and re-approved, or this flow goes back to disable-and-explain.

I left the spec untouched on purpose rather than rewriting a signed-off document unilaterally. Flagging it here so it is a decision rather than a discovery.

### Review of this branch

A multi-agent review of the first commit surfaced five defects on the destructive path, all fixed in `42b120f581`:

1. **X/ESC could dismiss the confirmation mid-deletion.** Only the Cancel button was gated; PrimeNG's chrome calls `close()` directly. The run does not stop with the dialog, so a half-finished cascade could complete with nothing reporting it, and the card kept a reference it could no longer close — making the eventual success a no-op, so the picker never opened. `closable`/`closeOnEscape` are now getters over `deletingVariants`, and the success handler opens the picker whether or not a dialog is left to close.
2. **The dialog's variant list was frozen at open time** while `deleting`/`failed` were signals — for exactly the reason the list needed to be one. After a partial failure it went on naming, and offering to delete, variants that were already gone.
3. **The failure copy implied nothing had happened** ("so the Page has not changed") when some variants may be permanently deleted.
4. **The store had no re-entrancy guard.** The dialog guarded its own button, but every dispatch reached the store, and `switchMap` would cancel a run whose deletions had partly landed. `deletingVariants` now closes the door, raised by `deleteVariantsRequested` as the first call leaves — the same contract `create$`/`start$` keep.
5. **A confirmation with nothing left to delete was dropped silently**, leaving the dialog waiting on a run that never went out. It is answered as success now (reachable: the variants can go from another tab between opening the dialog and confirming it).

Not changed, and worth a second opinion: `canChangePage()` requires `variants.length > 0`, so a draft with *zero* variants reads as immutable. Unreachable today (the creation POST always makes the control), and it mirrors a server-side rule, so I left it alone rather than widening the change.

Refs #37003

This PR fixes: #37003


core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-configure/components/dot-experiments-configure-page/dot-experiments-configure-page.component.ts
```diff
@@ -151,14 +167,124 @@ export class DotExperimentsConfigurePageComponent {
...
+     * `closable` and `closeOnEscape` are getters, not booleans: PrimeNG reads both off this config
+     * on every change detection, so a getter is what lets them close for as long as a deletion is
+     * on the wire. It matters because the deletions do not stop when the dialog does — dismissing
+     * it mid-run would let a half-finished cascade finish with nothing on screen reporting it, and
+     * would leave this card holding a reference it can no longer close.
+     *
+     * Only `true` opens the picker, and only this card ever closes the dialog with it: a cancelled
+     * dialog, or one closed after a rejection, leaves the page as it is.
+     */
+    #openChangePageDialog(): void {
```
adrianjm-dotCMS, line 256: **The deletion isn't actually contingent on the page changing.**

The order is: confirm → delete every non-control variant → open the picker. If the user cancels the picker, the variants are permanently gone and the page is unchanged — they paid the whole cost for nothing, with no way back. You call this out under "Two behaviours worth exercising by hand", but it reads more like a defect than a documented trait, because the copy promises the opposite:

```
experiments.configure.page.change.warning.one=If you change the Page, 1 Variant of "{0}" will be deleted…
```

"If you change the Page" conditions the destruction on a change that may never happen.

Two ways out. The better one is to invert the flow — picker first, then confirm, then delete, then patch — so nothing irreversible happens until a new page is actually chosen. It also makes the confirmation more informative, since it can name the page being moved *to*. The cheaper one is to keep the order and change the copy to say the deletion happens now regardless of what the picker answers.

Either way I'd want the resolution written into the reworked ACs, not just into the PR description.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-configure/components/dot-experiments-configure-page/dot-experiments-configure-page.component.ts
```diff
@@ -151,14 +167,124 @@ export class DotExperimentsConfigurePageComponent {
...
+     * Here rather than inside the dialog: the dialog is created outside this card's injector and so
+     * cannot reach the store at all — listening to the event from in there would be listening to a
+     * global bus for an answer that may not even be this screen's. This card holds the store *and*
+     * the reference, so it is the one place that can tie the two together.
+     *
+     * The picker opens either way. Closing the dialog is what normally leads to it, through the
+     * `onClose` above; but the deletions outlive the dialog, so an answer that arrives with no
+     * dialog left to close still has a promise to keep.
+     */
+    #closeConfirmationWhenVariantsAreGone(): void {
```
adrianjm-dotCMS, line 312: **This opens a page picker for any `deleteVariantsSucceeded`, including runs this card never started.**

The subscription has no guard tying the answer to a run this card initiated. Today that's safe — only `pageChangeConfirmed` produces the event, and only this dialog produces that — but it's a latent trap: the moment anything else emits `deleteVariantsSucceeded`, a page picker appears out of nowhere on a screen the user wasn't interacting with.

Your own test makes the shape visible:

```ts
it('should open the picker even when the answer arrives with no confirmation left', …)
```

That test dispatches success with no dialog ever opened, and asserts the picker opens. It's testing the desired behaviour for the *dismissed-mid-run* case, but it also documents the failure mode.

A `#awaitingPageChange` flag set in `#openChangePageDialog()` and cleared in this handler closes it, keeps the dismissed-mid-run case working, and makes the intent explicit — right now the invariant lives only in the fact that there's one producer.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-configure/components/dot-experiments-configure-page/dot-experiments-configure-page.component.ts
```diff
@@ -151,14 +167,124 @@ export class DotExperimentsConfigurePageComponent {
...
+    protected onChangePage(): void {
+        if (this.store.$canChangePage()) {
+            this.openPageSelector();
+
+            return;
+        }
+
+        // Nothing to delete and still refused means the experiment is past draft, which the
+        // disabled button already covers. Reachable only if the status moved under the click.
+        if (!this.$deletableVariants().length) {
```
adrianjm-dotCMS, line 232: **A draft with zero variants now gets an enabled button that does nothing.**

`canChangePage()` requires `variants.length > 0`, `$isPageActionDisabled` is now only `$isLocked`, and this early return bails when there's nothing to delete. Net result for a draft whose variants array is empty: the button is enabled, the click is swallowed, and nothing explains why.

You flagged it as unreachable today (the creation POST always makes the control) and I agree it is. But a dead click is a worse failure than the disabled button it replaced — if the invariant ever breaks, this is silent. Either drop `variants.length > 0` from `canChangePage()` (a draft with no variants has nothing to orphan, so the page really is free) or keep the button disabled in that case. I'd lean toward the former, since it removes the special case rather than papering over it.

adrianjm-dotCMS (dismissed):

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-configure/components/dot-experiments-configure-page/dot-experiments-configure-page.component.ts
oidacra: Fixed in `ad68e8528d`, though not the way you leaned — the server disagrees with the premise.

`ExperimentsAPIImpl.hasOnlyTheControlVariant` is:

```java
return variants.size() == 1 && DEFAULT_VARIANT.name().equals(variants.first().id());
```

So a draft with **zero** variants is refused by `validatePageChange` too, not just one carrying a non-control variant. Dropping `variants.length > 0` from `canChangePage()` would have the card offer the change, send the PATCH, and take a 400 back — the clause is a faithful mirror of `size() == 1`, not a special case papering over anything.

So I took the other option you gave: the button is disabled when the refusal is one no deletion could lift.

```ts
protected readonly $isPageActionDisabled = computed<boolean>(
    () =>
        this.store.$isLocked() ||
        (!this.store.$canChangePage() && !this.$deletableVariants().length)
);
```

Existing variants still leave it enabled — they are the step the confirmation exists for. It is only the case where there is nothing to delete *and* the server would still say no that turns the button off, which is exactly the dead click you found.

No new tooltip: the state is unreachable while the creation POST always makes the control, and inventing copy for it would be a string nobody ever reads. If you would rather it explain itself even so, say the word and I will add the key.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-configure/components/dot-experiments-configure-page/dot-experiments-configure-page.component.ts
oidacra: Done as suggested, in `ad68e8528d` — `#awaitingPageChange`, set in `#openChangePageDialog()`, cleared in this handler and also when the confirmation closes with nothing, so a cancelled dialog leaves nothing outstanding.

You are right that it changes no behaviour today, and that is the point: the invariant was resting entirely on there being a single producer of the event, and nothing said so.

One consequence for the test you quoted. `should open the picker even when the answer arrives with no confirmation left` now asserts the opposite — for a run this card never started, nothing opens. The case it was protecting, dismissed-mid-run, became unreachable in `42b120f581`: the dialog withdraws `closable` and `closeOnEscape` for as long as a deletion is on the wire, so the only way to lose it mid-run is destroying the component, and `takeUntilDestroyed` takes the subscription with it.

Two more added beside it: a second answer after its own run has been served opens nothing, and a cancelled confirmation stops the card waiting on that run.

adrianjm-dotCMS (approved):

PR #37313, opened 2026-08-31
fix(block-editor): preserve unknown marks instead of blanking the field (#37175)

### LINKED ISSUE #37175: Block Editor renders blank on edit when Allowed Blocks is configured

adrianjm-dotCMS:
Closes the two acceptance criteria QA found short on #37175 — see the [QA verification comment](https://github.com/dotCMS/core/issues/37175#issuecomment-5482595119). The originally reported regression was already fixed by #37205 and stays fixed; this PR is the hardening half.

### Proposed Changes

* **AC5 — an unknown *mark* no longer aborts the document.** #37205 registered `link`, `emoji` and `youtube`, which fixed the two *known* offenders (`link`, `highlight`) but not the failure mode. Any mark the schema does not declare throws `RangeError: There is no mark type X in this schema` mid-recursion in `Node.fromJSON`. TipTap catches it, logs `[tiptap warn]: Invalid content`, and boots an **empty** document — so the field looks emptied while the stored JSON is intact, and the next save makes that loss real. Unknown *nodes* were already preserved as `dotUnsupportedBlock`; marks had no equivalent. Adds the mark-side twin `dotUnsupportedMark` (`libs/dotcms-models/src/lib/unknown-block.util.ts`), registered in both editors.
* **Preserve rather than strip**, mirroring the existing node path as QA suggested. The absence of a mark is usually *transient*: a `customBlocks` remote extension whose CDN is unreachable (`remote-extensions.loader.ts` drops failed loads and boots anyway — correctly), content written through the API or migrated from another CMS (`textStyle`/`color`/`fontFamily` are standard TipTap marks we don't register), or a version rollback. Stripping would turn a 20-minute outage into permanent loss for every author who saved during it.
* **AC3 — `linkOnPaste: false` did not close the link-on-paste path.** TipTap's Link returns its URL paste rule from `addPasteRules()` with no option guard, so pasting text *containing* a URL still created a link mark on a field where `link` isn't allowed; `linkOnPaste` only suppresses wrapping a selection. `DotLink` now overrides `addPasteRules`, mirroring how `@tiptap/extension-youtube` guards its own paste handler.
* **Ordering invariant, documented at all three sites: nodes first, then marks.** An unknown node is swallowed whole into the placeholder's `originalNode` attr, and that payload is inert data rather than part of the document tree — running the mark pass first would rewrite the very content the placeholder exists to preserve. There's a test pinning it.
* `restoreUnknownBlockNodes` now restores both halves, so each editor's single emit path picks it up with **no new call sites** on the save path.
* `libs/new-block-editor/CLAUDE.md`: new "Unknown nodes and marks (load-path invariant)" section + the mark inventory row.

### Checklist
- [x] Tests
- [ ] Translations — none needed, the placeholder renders no user-facing copy
- [x] Security Implications Contemplated — no new input surface; the preserved payload is stored/re-emitted verbatim and never evaluated. `JSON.parse` of the `data-original-mark` attribute is already try/caught with a raw-string fallback, same as the node path.

### Additional Info

**Verified in Chromium** against a local instance, with `allowedBlocks` set to the issue's exact list (`bulletList,orderedList,codeBlock,table`) and a fixture carrying an undeclared `legalCitation` mark, in **both** editors:

| | Before (`main`) | After |
|---|---|---|
| Chars rendered on open | **0** | **157** |
| Paragraphs | 1 (empty placeholder) | 3 + node placeholder for `heading` |
| Console | `RangeError: There is no mark type legalCitation in this schema` | **0 errors** |
| Type one word → Save and Publish | 995 → **203 bytes**, marks `[]` | 995 → **1151 bytes**, marks `['legalCitation','bold','link']` |
| `dotUnsupportedMark` persisted to storage | — | **no** on the JSON-parse-success path — see footnote |
| Paste text containing a URL | new `<a>` created | no new `<a>` |

Legacy editor confirmed by `dot-old-block-editor` in the DOM (feature flag forced via network interception, container config untouched). 203 bytes matches QA's reported 202 — same mechanism.

> **Footnote on "persisted to storage" — not an absolute guarantee, by design.** When the preserved payload no longer parses, `restoreUnknownMark` keeps the placeholder so `originalMarkRaw` still round-trips, which means a `dotUnsupportedMark` carrying `originalMarkRaw` *can* reach storage. That mirrors `dotUnsupportedBlock` exactly and is pinned by `keeps a placeholder whose payload is no longer valid during restore`. So a placeholder in stored JSON is not a bug; a *lost payload* would be. Thanks to @rjvelazco for measuring the case the table glossed over.

**Decision flagged for reviewers — resolved: preserve.** Both reviewers landed on preserving rather than stripping, so no change. Recorded for anyone reading later: this registers a new schema entry, `dotUnsupportedMark`. Per `libs/new-block-editor/CLAUDE.md` → *"TipTap Node Names Are Immutable"*, that name can never be changed once content exists. If that's too much of a commitment, the alternative is stripping unknown marks instead of preserving them — it also satisfies AC5 as written (*"the affected text survives a save"*), is ~20 lines instead of ~200, and the traversal and call sites are already in place.

**Out of scope, confirmed by the issue's own scoping** (*"Scope: new Block Editor only… the legacy editor registers `Link` unconditionally"*, *"The legacy editor never supported restricting links"*): the legacy editor still auto-links pasted URLs, since it uses its own `Link.extend` in `getEditorMarks()`. Verified `1 → 2` anchors there. Closing that would change behaviour the ticket explicitly excludes.

**Review feedback addressed** (second commit):
* `isJsonMark` — `isJsonContent` was validating mark payloads. It works, but a tightened `isJsonContent` would break mark restoration *silently*, with a `dotUnsupportedMark` in storage as the only symptom.
* `DotLink.addPasteRules` now gates on `autolink || linkOnPaste`. The rule being suppressed is the auto-link-on-paste rule, closer to `autolink`; `linkOnPaste: false` + `autolink: true` is coherent and the old gate would have killed the linkifying. Pinned by a new test that configures the extension directly, since the assembly path cannot produce that combination.
* Clear-formatting limitation documented in `createUnsupportedBlockMark` and `new-block-editor/CLAUDE.md`: the toolbar runs `unsetAllMarks().clearNodes()`, which strips this placeholder along with every real mark — and `clearNodes()` does the same to `dotUnsupportedBlock`. Neither renders visible formatting, so the loss is silent and irreversible. Still strictly better than the document not loading, and unlike that, user-initiated. Both notes say the fix belongs to the clear-formatting command in each editor, not to the placeholders.
* Not changed, deliberately: memoizing the known-node/known-mark `Set`s. The cost is real (the guard runs per node selection via `markViewDirty`), but `editorContentMatchesParsed` is the one function this PR and #36985 both touch, so it belongs in that rewrite.

**Also not addressed here**, both flagged by QA as separate concerns:
* QA's finding #2 — `TypeError: o.can(...).mergeCells is not a function` when `table` isn't in `allowedBlocks`. Pre-existing, untouched by #37205 and by this PR; worth its own issue.
* `libs/dotcms-models` has no `.spec.ts` of its own, so the shared util is exercised from its consumers (`new-block-editor`, `block-editor`). No E2E added to `apps/dotcms-ui-e2e`.

Tests: `nx test new-block-editor` 137/137 · `nx test block-editor --testPathPatterns=unknown-block` 5/5 · `tsc --noEmit` clean on all three libs · `nx lint new-block-editor` / `dotcms-models` pass. `nx test block-editor` full suite has 10 pre-existing failures (`this.editor.setEditable is not a function`, an incomplete spec mock) — identical on clean `main`, verified by stashing.

This PR fixes: #37175

This PR fixes: #37175

### DIFF
--- a/core-web/libs/dotcms-models/src/lib/unknown-block.util.ts
+++ b/core-web/libs/dotcms-models/src/lib/unknown-block.util.ts
@@ -34,6 +34,18 @@ function isJsonContent(value: unknown): value is JSONContent {
 }
 
 /**
+ * Mark-side counterpart of {@link isJsonContent}. A mark JSON shares the same
+ * "object carrying a non-empty string `type`" shape, so this delegates — but it is a
+ * separate name on purpose. `isJsonContent` is free to tighten around node-specific rules
+ * (requiring `content`, or checking the name against the schema's nodes), and mark
+ * restoration must not start failing when it does. That failure would be silent: the symptom
+ * is a `dotUnsupportedMark` written to storage, not an exception.
+ */
+function isJsonMark(value: unknown): value is JSONMark {
+    return isJsonContent(value);
+}
+
+/**
  * Builds the placeholder attrs stored on `dotUnsupportedBlock`.
  *
  * @param node The original unsupported TipTap node JSON.
@@ -143,7 +155,7 @@ export function renderUnknownBlockOriginalNode(
 export function renderUnknownBlockOriginalMark(
     attributes: Partial<UnknownBlockMarkAttrs>
 ): Record<string, string> {
-    if (isJsonContent(attributes.originalMark)) {
+    if (isJsonMark(attributes.originalMark)) {
         return { 'data-original-mark': JSON.stringify(attributes.originalMark) };
     }
 
@@ -219,6 +231,16 @@ export function createUnsupportedBlockNode() {
  *
  * `inclusive: false` keeps the mark from swallowing text typed at its boundary, so an
  * author extending that sentence does not silently widen a mark this editor cannot render.
+ *
+ * KNOWN LIMITATION — "Clear formatting" discards the preserved payload. That action runs
+ * `unsetAllMarks().clearNodes()` (`toolbar.component.ts`; the legacy bubble menu runs
+ * `unsetAllMarks()`), which strips this placeholder along with every real mark — and
+ * `clearNodes()` does the same to `dotUnsupportedBlock`. Since neither placeholder renders
+ * any visible formatting, the author has no way to know something was discarded, and there
+ * is no way back. It is still strictly better than the pre-fix behaviour, where the document
+ * did not load at all, and unlike that, it is user-initiated. Excluding BOTH placeholders
+ * from the clear-formatting command is worth doing, but it belongs to that command in each
+ * editor rather than here — the node half is affected the same way and predates this mark.
  */
 export function createUnsupportedBlockMark() {
     return Mark.create({
@@ -362,7 +384,7 @@ export function preserveUnknownBlockMarks<T extends JSONLikeOrUndefined>(
  * when the payload is no longer valid so the recoverable raw string still round-trips.
  */
 function restoreUnknownMark(mark: JSONMark): JSONMark {
-    if (mark?.type === UNKNOWN_BLOCK_MARK_NAME && isJsonContent(mark.attrs?.['originalMark'])) {
+    if (mark?.type === UNKNOWN_BLOCK_MARK_NAME && isJsonMark(mark.attrs?.['originalMark'])) {
         return mark.attrs['originalMark'] as JSONMark;
     }
 
--- a/core-web/libs/dotcms-models/src/lib/unknown-block.util.ts
+++ b/core-web/libs/dotcms-models/src/lib/unknown-block.util.ts
@@ -1,14 +1,24 @@
-import { JSONContent, Node } from '@tiptap/core';
+import { JSONContent, Mark, Node } from '@tiptap/core';
 
 export const UNKNOWN_BLOCK_NODE_NAME = 'dotUnsupportedBlock';
+export const UNKNOWN_BLOCK_MARK_NAME = 'dotUnsupportedMark';
 
 type JSONLike = JSONContent | JSONContent[];
 type JSONLikeOrUndefined = JSONLike | undefined;
+
+/** The JSON shape TipTap uses for an entry of `JSONContent.marks`. */
+type JSONMark = NonNullable<JSONContent['marks']>[number];
 
 export interface UnknownBlockNodeAttrs {
     originalType: string | null;
     originalNode: JSONContent | null;
     originalNodeRaw: string | null;
+}
+
+export interface UnknownBlockMarkAttrs {
+    originalType: string | null;
+    originalMark: JSONMark | null;
+    originalMarkRaw: string | null;
 }
 
 function isJsonContent(value: unknown): value is JSONContent {
@@ -43,6 +53,21 @@ export function createUnknownBlockNodeAttrs(
 }
 
 /**
+ * Builds the placeholder attrs stored on `dotUnsupportedMark` — the mark-side
+ * counterpart of {@link createUnknownBlockNodeAttrs}.
+ */
+export function createUnknownBlockMarkAttrs(
+    mark: JSONMark,
+    markType: string | null
+): UnknownBlockMarkAttrs {
+    return {
+        originalType: markType,
+        originalMark: mark,
+        originalMarkRaw: null
+    };
+}
+
+/**
  * Parses the serialized `data-original-node` HTML attribute back into JSON when
  * possible, and preserves the raw string separately when parsing fails so the
  * original content can still round-trip through the placeholder unchanged.
@@ -72,6 +97,32 @@ export function parseUnknownBlockOriginalNode(
     }
 }
 
+/** Mark-side counterpart of {@link parseUnknownBlockOriginalNode}. */
+export function parseUnknownBlockOriginalMark(
+    value: string | null
+): Pick<UnknownBlockMarkAttrs, 'originalMark' | 'originalMarkRaw'> {
+    if (!value) {
+        return {
+            originalMark: null,
+            originalMarkRaw: null
+        };
+    }
+
+    try {
+        return {
+            originalMark: JSON.parse(value),
+            originalMarkRaw: null
+        };
+    } catch (error) {
+        console.warn('[unsupported-mark] failed to parse originalMark', error);
+
+        return {
+            originalMark: null,
+            originalMarkRaw: value
+        };
+    }
+}
+
 /**
  * Re-renders the preserved original node payload back onto the placeholder DOM
  * node so unsupported blocks keep their recoverable serialized representation.
@@ -85,6 +136,19 @@ export function renderUnknownBlockOriginalNode(
 
     return typeof attributes.originalNodeRaw === 'string' && attributes.originalNodeRaw.length > 0
         ? { 'data-original-node': attributes.originalNodeRaw }
+        : {};
+}
+
+/** Mark-side counterpart of {@link renderUnknownBlockOriginalNode}. */
+export function renderUnknownBlockOriginalMark(
+    attributes: Partial<UnknownBlockMarkAttrs>
+): Record<string, string> {
+    if (isJsonContent(attributes.originalMark)) {
+        return { 'data-original-mark': JSON.stringify(attributes.originalMark) };
+    }
+
+    return typeof attributes.originalMarkRaw === 'string' && attributes.originalMarkRaw.length > 0
+        ? { 'data-original-mark': attributes.originalMarkRaw }
         : {};
 }
 
@@ -145,6 +209,60 @@ export function createUnsupportedBlockNode() {
 }
 
 /**
+ * Placeholder for a mark the current schema does not know, and the reason unknown marks
+ * no longer blank the whole field (#37175).
+ *
+ * Unlike an unknown *node*, an unknown *mark* has nothing to show: the text it decorates
+ * is ordinary text that must keep rendering and stay editable. So this mark is visually
+ * neutral — it only carries the original mark payload so it round-trips back on save
+ * instead of being dropped.
+ *
+ * `inclusive: false` keeps the mark from swallowing text typed at its boundary, so an
+ * author extending that sentence does not silently widen a mark this editor cannot render.
+ */
+export function createUnsupportedBlockMark() {
+    return Mark.create({
+        name: UNKNOWN_BLOCK_MARK_NAME,
+        inclusive: false,
+
+        addAttributes() {
+            return {
+                originalType: {
+                    default: null,
+                    parseHTML: (element) => element.getAttribute('data-original-type'),
+                    renderHTML: (attributes) =>
+                        attributes['originalType']
+                            ? { 'data-original-type': attributes['originalType'] }
+                            : {}
+                },
+                originalMark: {
+                    default: null,
+                    parseHTML: (element) =>
+                        parseUnknownBlockOriginalMark(element.getAttribute('data-original-mark'))
+                            .originalMark,
+                    renderHTML: (attributes) => renderUnknownBlockOriginalMark(attributes)
+                },
+                originalMarkRaw: {
+                    default: null,
+                    parseHTML: (element) =>
+                        parseUnknownBlockOriginalMark(element.getAttribute('data-original-mark'))
+                            .originalMarkRaw,
+                    renderHTML: () => ({})
+                }
+            };
+        },
+
+        parseHTML() {
+            return [{ tag: `span[data-mark-type="${UNKNOWN_BLOCK_MARK_NAME}"]` }];
+        },
+
+        renderHTML({ HTMLAttributes }) {
+            return ['span', { ...HTMLAttributes, 'data-mark-type': UNKNOWN_BLOCK_MARK_NAME }, 0];
+        }
+    });
+}
+
+/**
  * Replaces any node whose type is unknown to the current schema with the shared
  * unsupported-block placeholder while recursively preserving unknown descendants
  * inside otherwise-known parent nodes.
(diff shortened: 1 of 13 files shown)


core-web/libs/dotcms-models/src/lib/unknown-block.util.ts
```diff
@@ -144,6 +208,60 @@ export function createUnsupportedBlockNode() {
...
+ * neutral — it only carries the original mark payload so it round-trips back on save
+ * instead of being dropped.
+ *
+ * `inclusive: false` keeps the mark from swallowing text typed at its boundary, so an
+ * author extending that sentence does not silently widen a mark this editor cannot render.
+ */
+export function createUnsupportedBlockMark() {
+    return Mark.create({
+        name: UNKNOWN_BLOCK_MARK_NAME,
+        inclusive: false,
```
rjvelazco, line 248: `inclusive: false` is the right call and I verified it works — typing at the end of decorated text does not inherit the placeholder.

One behaviour worth a comment here, because it cuts against the premise of the whole PR: **`unsetAllMarks()` destroys the preserved payload permanently.** Measured on this branch, real editor, real schema:

```
load    : [{"type":"text","marks":[{"type":"legalCitation"}],"text":"cited"}]
selectAll().unsetAllMarks()
emit    : [{"type":"text","text":"cited"}]        ← legalCitation gone, unrecoverably
```

That is the toolbar's "clear formatting" button. An author who uses it on a paragraph they cannot even see the mark on (it renders neutral by design) silently discards a payload this PR exists to protect — and unlike the pre-PR blanking, there is no error and no way back.

I don't think it blocks: it is user-initiated, and it is still strictly better than today where the document doesn't load at all. But it is worth either a line in the docblock so the next reader knows, or excluding `dotUnsupportedMark` from `unsetAllMarks`. Your call — I'd take the comment as the minimum.

---
_🤖 Reviewed by Claude on behalf of @rjvelazco_

core-web/libs/dotcms-models/src/lib/unknown-block.util.ts
```diff
@@ -180,6 +298,77 @@ export function preserveUnknownBlockNodes<T extends JSONLikeOrUndefined>(
...
+
+    return content.map((node) => replaceUnknownMarksInNode(node, knownMarkNames)) as T;
+}
+
+/**
+ * Restores a placeholder mark back to the mark it stood in for, keeping the placeholder
+ * when the payload is no longer valid so the recoverable raw string still round-trips.
+ */
+function restoreUnknownMark(mark: JSONMark): JSONMark {
+    if (mark?.type === UNKNOWN_BLOCK_MARK_NAME && isJsonContent(mark.attrs?.['originalMark'])) {
```
rjvelazco: Nit, but the kind that rots quietly: `isJsonContent` is being used to validate a **mark** payload.

It happens to work, because it only asserts `typeof value.type === 'string'` and a mark shares that shape. But the name now claims something untrue, and the coupling is invisible — if `isJsonContent` is ever tightened (say it starts requiring `content`, or rejects objects without a node type in the schema), mark restoration breaks silently and the failure mode is a `dotUnsupportedMark` written to storage, not an exception.

Same applies at `renderUnknownBlockOriginalMark`.

An `isJsonMark` alias next to `isJsonContent` — even a one-liner delegating to it — would pin the intent and cost nothing.

---
_🤖 Reviewed by Claude on behalf of @rjvelazco_

core-web/libs/new-block-editor/src/lib/editor/extensions/link.extension.ts
```diff
@@ -21,5 +21,19 @@ export const DotLink = Link.extend({
...
+     * The base extension returns its URL paste rule unconditionally — `linkOnPaste` only
+     * governs wrapping the *selection*, so pasting text that merely contains a URL still
+     * created a link mark on a field where `link` is not an allowed block (#37175).
+     *
+     * The mark stays in the schema either way (dropping it aborts `Node.fromJSON` for the
+     * whole document); what `allowedBlocks` gates is authoring, and auto-linking pasted text
+     * is an authoring path. Compare `@tiptap/extension-youtube`, which guards its own paste
+     * handler with `addPasteHandler`.
+     */
+    addPasteRules() {
```
rjvelazco, line 43: Verified this does what the docblock says. Measured on this branch:

```
[unrestricted]        linkOnPaste=true  autolink=true  pasteRules=[RULE]
[restricted, no link] linkOnPaste=false autolink=false pasteRules=[]
```

One observation on the gate itself: the rule being suppressed is the **autolink-on-paste** rule, but it is gated on `linkOnPaste`, which the docblock correctly says governs *wrapping the selection*. Those are two different options being conflated.

It is harmless today only because `editor-extensions.ts` sets both from the same `has('link')`, so they can never disagree. The moment someone configures them independently — `linkOnPaste: false, autolink: true`, a reasonable combination meaning "don't wrap my selection but do linkify URLs I paste" — this silently disables the wrong thing.

`this.options.autolink || this.options.linkOnPaste` would survive that, or a sentence saying the two are deliberately treated as one gate here. Not blocking.

---
_🤖 Reviewed by Claude on behalf of @rjvelazco_

core-web/libs/new-block-editor/src/lib/editor/editor.component.ts
```diff
@@ -149,7 +153,13 @@ function editorContentMatchesParsed(editor: Editor, parsed: string | JSONContent
     }
     return (
         JSON.stringify(
-            stripDocStats(preserveUnknownNodesInDocument(parsed, getKnownNodeNames(editor)))
+            stripDocStats(
+                preserveUnknownNodesInDocument(
+                    parsed,
+                    getKnownNodeNames(editor),
+                    getKnownMarkNames(editor)
```
rjvelazco, line 160: **Heads-up on a collision rather than a defect in this PR** — worth knowing before either lands.

`editorContentMatchesParsed` is the exact function #36985 is about to rewrite ([spec PR #37285](https://github.com/dotCMS/core/pull/37285)). Two interactions:

**1. This PR partly invalidates one of that spec's acceptance criteria.** AC-008 there asserts "a body carrying a mark the schema can't deserialize still triggers `setContent` rather than blanking the field" — the conservative `catch` path. After this PR, on the object path, *there is no such thing*: every unknown mark becomes a placeholder and `Node.fromJSON` succeeds. That's a strict improvement, but I'll need to re-scope that AC. Good problem to have.

**2. Two fresh `Set`s per call, on a hot path.** `getKnownNodeNames` + `getKnownMarkNames` each do `Object.keys()` + `new Set()` on every invocation. The schema is immutable for the life of the editor, so this is recomputable-once work.

Normally I'd skip that as premature. I'm raising it because #36985 measured this guard running **on every node selection** — clicking an embedded contentlet re-runs the effect that calls it, via `markViewDirty`. So it's called far more often than "once per value push" suggests. Still small in absolute terms; genuinely optional, and arguably out of scope here.

No change requested. Flagging so we don't surprise each other at merge.

---
_🤖 Reviewed by Claude on behalf of @rjvelazco_

rjvelazco (dismissed):
## Approving with comments — nothing here should block

I reviewed this against a checkout of the branch rather than by reading, because I've spent the last while inside this exact load path on #36985 and wanted to check the edge cases rather than trust the diff. **Every load-bearing claim in the PR description holds up.** Four non-blocking comments inline, plus one correction to the description below.

### What I verified on the branch

Real schema built from `createEditorExtensions(...)`, real `Editor` instances, real transactions:

| Probe | Result |
|---|---|
| Issue's exact `allowedBlocks` (`bulletList,orderedList,codeBlock,table`) → `link` **and** `dotUnsupportedMark` both in `schema.marks` | ✅ |
| Document with a `link` **and** an unknown `legalCitation` mark loads under that restriction, text intact (`"a link and a citation"`) | ✅ |
| **Two different** unknown marks on the same text — both preserved, both restored, and they survive an edit elsewhere in the doc | ✅ |
| `inclusive: false` — typing at the boundary does not inherit the placeholder | ✅ |
| Applying **bold** over unknown-marked text keeps both marks | ✅ |
| Ordering invariant — an unknown mark inside a `dotUnsupportedBlock` payload is left byte-for-byte | ✅ |
| Idempotence — load → emit → re-load → emit is stable | ✅ |
| Paste rules: `[RULE]` unrestricted, `[]` when link isn't allowed | ✅ |
| Legacy emit path restores marks (`dot-block-editor.component.ts:313`, via the shared `restoreUnknownBlockNodes`) — **no new call site needed, as claimed** | ✅ |

I went in expecting to find a problem with two same-type marks collapsing, since ProseMirror mark sets are normally exclusive per type. **They don't** — `Mark.setFrom` doesn't dedupe, and I confirmed both survive a subsequent transaction. Worth recording because it's the non-obvious part of the design and it works.

I also checked the initialisation ordering in the legacy editor, since an empty `#knownEditorMarkNames` at load would turn **every** mark into a placeholder. It's populated synchronously between `new Editor(...)` and `subscribeToEditorEvents()`, and `writeValue` guards on `!this.editor`, so there's no window. Safe.

### One correction to the description

> | `dotUnsupportedMark` persisted to storage | — | **no** |

Not quite — it *can* be, by design. When the payload doesn't survive parsing, `restoreUnknownMark` keeps the placeholder so `originalMarkRaw` still round-trips. Measured:

```
in : marks:[{type:'dotUnsupportedMark', attrs:{originalMark:null, originalMarkRaw:'not json'}}]
out: marks:[{type:'dotUnsupportedMark', attrs:{originalMark:null, originalMarkRaw:'not json'}}]   ← persisted
```

That's correct behaviour and it mirrors the node path exactly — but the table reads as an absolute guarantee, and a reviewer or QA taking it literally would treat a `dotUnsupportedMark` in stored JSON as a bug. Worth a footnote.

### On the decision you flagged

> **Decision worth a reviewer's opinion:** this registers a new schema entry, `dotUnsupportedMark` … the alternative is stripping unknown marks instead of preserving them.

**Preserve.** Two reasons, and I'd hold this position under push-back:

1. **Your transience argument is the right one and it's stronger than the PR states.** A CDN blip on a `customBlocks` remote extension is not a hypothetical — `remote-extensions.loader.ts` drops failed loads and boots anyway, correctly. Under the stripping design, a 20-minute outage silently converts to permanent, unrecoverable loss for every author who saved during it, with no error at any point. That's the same shape as the bug being fixed, just slower.
2. **The immutability commitment is real but small.** You're pinning one name forever. Weigh that against a defect class that has now surfaced three times (#37145 `highlight`, #37175 `link`, and this) — every recurrence being *some* schema entry the editor didn't know about. A permanent name is a cheap price for closing the class rather than the instance.

The ~200-vs-~20 line difference is not the deciding factor: most of those 200 lines are the symmetric twin of code that already exists and is already tested, which is about the cheapest kind of 200 lines there is.

### Scope

I agree with both exclusions. The legacy auto-link behaviour is explicitly out of scope per the issue's own scoping, and QA's `mergeCells` finding is a genuinely separate pre-existing defect — worth its own issue so it doesn't get lost.

### The one thing I'd want before merge

Nothing in the code. But the `unsetAllMarks` behaviour in my first inline comment is a real, measured way to lose the payload this PR protects, and it isn't covered by a test or a comment. A docblock line would satisfy me.

---
_🤖 Reviewed by Claude on behalf of @rjvelazco_

core-web/libs/dotcms-models/src/lib/unknown-block.util.ts
adrianjm-dotCMS: Good catch, and it's worse than measured: the new editor's action is `unsetAllMarks().clearNodes()` (`toolbar.component.ts:403`), so `clearNodes()` takes `dotUnsupportedBlock` with it. The legacy bubble menu is marks-only.

Took the docblock, not the exclusion — once `clearNodes()` is in scope this is a clear-formatting-vs-both-placeholders problem, so the fix is one change in each editor's command, not a carve-out in the mark. It also has a real question I don't want to answer sideways: excluding them leaves the author no way to purge an unknown mark short of deleting the text. Documented in the docblock and in `new-block-editor/CLAUDE.md`, both saying the fix belongs to that command.

core-web/libs/dotcms-models/src/lib/unknown-block.util.ts
adrianjm-dotCMS: Fixed — `isJsonMark` delegating to `isJsonContent`, used at `renderUnknownBlockOriginalMark` and `restoreUnknownMark`.

Your reasoning is in the docblock, not just the guard: the risk isn't the name, it's that a tightened `isJsonContent` breaks mark restoration *silently*, with a `dotUnsupportedMark` in storage as the only symptom.

core-web/libs/new-block-editor/src/lib/editor/extensions/link.extension.ts
adrianjm-dotCMS: Agreed — now `autolink || linkOnPaste`. You're right that the rule being suppressed is the auto-link-on-paste rule and I gated it on the option governing selection-wrapping.

Added a test configuring the extension directly (`autolink: true, linkOnPaste: false` → 1 rule), since the assembly path can't produce that combination and nothing else would catch a regression.

core-web/libs/new-block-editor/src/lib/editor/editor.component.ts
adrianjm-dotCMS: Thanks for both.

**AC-008** — agreed, re-scope is yours. For the record: on the object path there's no undeserializable mark left to fall back from; the string path is untouched.

**The Sets** — your measurement convinces me the cost is real, but I'm leaving it, and for your first point rather than scope: `editorContentMatchesParsed` is the one function both PRs touch, so memoizing here conflicts with your rewrite. Belongs in #36985.

adrianjm-dotCMS:
Thanks @rjvelazco — the two-same-type-marks probe and the legacy init-ordering check are both things I hadn't verified, and `Mark.setFrom` not deduping is the non-obvious part of the design. Good to have it recorded.

**Your description correction is the most useful comment here.** The table read as an absolute guarantee and it isn't: the raw-string fallback keeps the placeholder on purpose so `originalMarkRaw` round-trips, mirroring the node path, and it's already pinned by `keeps a placeholder whose payload is no longer valid during restore`. Purely my wording — someone in QA reading it literally would have filed a `dotUnsupportedMark` in stored JSON as a bug. Footnoted in the description.

Decision: preserve. Your second argument is the one I'll reuse — a permanent name is cheap against a defect class that has surfaced three times (#37145 `highlight`, #37175 `link`, this).

One thing neither review covered, since half the diff lives there: the **legacy** editor was verified at runtime, not just read. `dot-old-block-editor` confirmed in the DOM (flag forced via network interception, container config untouched), restricted field, mark preserved, link and bold intact, zero console errors. That's how the legacy paste behaviour we agree is out of scope turned up.

Pushed in 6eff8fca3f: `isJsonMark`, the paste gate on either flag plus the `autolink`-only test, and the clear-formatting limitation documented. Skipped the Set memoization for the reason in that thread. 137/137.

rjvelazco (approved):

PR #37315, opened 2026-08-31
fix(evergreen-tracks): make dry-run unmissable in the admin dispatch UI

### LINKED ISSUE #37314: evergreen-tracks-admin: dry-run mode is too easy to miss in the dispatch UI
The `evergreen-tracks-admin` workflow defaults to dry-run, which is correct — but the dispatch form makes that almost invisible:

- The `apply` input is the **last** of six fields, below `repo`/`action`/`version`/`track`/`force`.
- It is an untyped string pre-filled with the word `false`, so it renders as a free-text box, not a checkbox or dropdown.
- Nothing in the run title or the Actions list distinguishes a dry-run from a real mutation. Both are titled `evergreen-tracks-admin` and both go green.

Result: on 2026-08-31, [run 33434507975](https://github.com/dotCMS/core/actions/runs/33434507975) was dispatched to taint `26.08.28-01`, went green, and was read as a completed taint. It was a dry-run — the only signal was `DRY-RUN would point ...` buried in the step log. The real taint landed 14 minutes later in [run 33435793043](https://github.com/dotCMS/core/actions/runs/33435793043). No harm done (no track pointed at that version), but the failure mode is "operator believes a bad release is blocked when it is not".

### Proposed fix

- Replace `apply: true|false` with a `mode` choice dropdown (`dry-run` | `apply`) and move it to the **top** of the form.
- Add `run-name` so the Actions list reads `DRY-RUN (no changes) — taint 26.08.28-01` vs `APPLY — taint 26.08.28-01`.
- Emit a `::notice::` on the dry-run path.
- Update RUNBOOK.md to match.

Only the admin workflow is affected; `evergreen-tracks-promote` uses a plan → gate → apply job structure and has no `apply` input.

sfreudenthaler:
Closes: #37314

Two related gaps in the `evergreen-tracks-admin` operator surface: the dispatch form hid its dry-run default, and `force` was documented as "Don't" without ever saying what it does.

## 1. Dry-run was invisible in the dispatch UI

Dry-run is the right default, but the form hid it well enough that a dry-run was mistaken for a completed taint on 2026-08-31 ([run 33434507975](https://github.com/dotCMS/core/actions/runs/33434507975) — green, titled the same as a real run, with `DRY-RUN would point ...` buried in the step log).

| Before | After |
|---|---|
| `apply` — last of six inputs, untyped string pre-filled `false` | `mode` — **first** input, `choice` dropdown: `dry-run` \| `apply` |
| Run titled `evergreen-tracks-admin` either way | `DRY-RUN (no changes) — taint 26.08.28-01` vs `APPLY — taint 26.08.28-01` |
| Dry-run silent except in the step log | `::notice::` annotation on the run summary |
| `force` untyped string `false` | `type: boolean` (renders a checkbox) |

Safe default is preserved: `dry-run` is the first option, so it stays preselected.

## 2. `force` had no what or why

The RUNBOOK said *"Don't, unless you're deliberately choosing the least-bad option"* — a prohibition with no explanation. The README never mentioned it. Verified against the code and documented in both:

- **What** — `force` lifts exactly one guard: `hold` onto a version carrying a `_tainted` marker. It is inert on `taint`, `untaint`, and `release-hold`.
- **Why** — taint and hold collide in exactly one situation: every candidate is bad and you must pick the least-bad one. You are fleeing a release actively breaking customers, and the only older release you can retreat to is itself tainted for something milder.
- **What it costs** — a held track bypasses the planner, and the planner is the only thing that reads taint markers. Promote reconciles the track tag to the hold digest every run, so the track is **actively re-pinned** to the known-bad release until someone runs `release-hold`.

**Latent trap found while verifying this**, now documented in both files and the Gotchas list: the tainted check runs *only* at hold time (`cli.py:210`). Hold onto a good version, taint it later, and the hold survives — promote keeps re-pinning the track onto a version you just declared bad. Nothing re-validates and nothing notifies. Worth a follow-up issue to make `taint` fail or warn when a `<track>_hold` marker points at the version being tainted.

The README also gains a short table of the three registry markers, which it previously never explained.

## Verification

Screenshots and run titles in the comments below, captured from a secrets-free copy of this dispatch surface in [dotCMS/core-workflow-test](https://github.com/dotCMS/core-workflow-test/actions/workflows/evergreen-tracks-admin-ui-demo.yml) plus the live authenticated Run workflow panel. That demo caught a real bug in the first `run-name` (`hold 26.08.14-01standard` — `hold` sets both `version` and `track`), fixed in 165140547f.

## Scope

Only `cicd_evergreen-tracks-admin.yml` and the two evergreen-tracks docs. `cicd_evergreen-tracks-promote.yml` has no `apply` input — it uses a `plan` -> `gate` (required reviewer) -> `apply` job structure, so the same confusion cannot arise there.

The operator runbook in the **#dot-releases** channel canvas has been updated with the same `force` guidance, since it had no evergreen-tracks section at all.

### DIFF
--- a/.github/actions/core-cicd/evergreen-tracks/README.md
+++ b/.github/actions/core-cicd/evergreen-tracks/README.md
@@ -53,6 +53,33 @@ window. A tag moved on a Tuesday has no effect until that window. Nothing in the
 watches the tags continuously (no Argo CD Image Updater / Keel, no floating-tag references,
 no `imagePullPolicy: Always` on customer pods).
 
+## State: the three marker tags
+
+There is no database. Every decision the planner makes is derived from tags in the registry:
+
+| Tag | Written by | Means |
+|---|---|---|
+| `latest` / `standard` / `trailing` | promote | the release this track currently resolves to |
+| `<version>_tainted` | `admin --action taint` | no track may **advance** onto this version |
+| `<track>_hold` | `admin --action hold` | this track is frozen; promote reconciles the track tag to this digest instead of planning |
+
+The two markers are independent guards, and a hold is the stronger one: a held track skips the
+planner entirely, and the planner is the only thing that reads taint markers.
+
+### `force`
+
+`hold` refuses to point a track at a tainted version unless `force` is set. It is the
+last-resort control for when **every candidate is bad and you must pick the least-bad one** —
+retreating from a release that is breaking customers onto an older one already tainted for
+something milder.
+
+Because a forced hold bypasses the planner, the track is not just parked on a known-bad
+release, it is **actively re-pinned there every promote run** until someone runs
+`release-hold`. The tainted check also runs *only* at hold time — tainting a version later
+does not release a hold that already points at it.
+
+Full procedure, rules, and exit path: [RUNBOOK → `force`](RUNBOOK.md#force--holding-onto-a-tainted-version).
+
 ## Operator procedures
 
 See [RUNBOOK.md](RUNBOOK.md) for tainting a release, holding a track, and holding a single
--- a/.github/actions/core-cicd/evergreen-tracks/RUNBOOK.md
+++ b/.github/actions/core-cicd/evergreen-tracks/RUNBOOK.md
@@ -85,8 +85,41 @@ skips the track and self-heals the floating tag back to the held digest if it dr
    | `track` | `standard`, `trailing`, or `latest` |
    | `version` | the version to freeze on, e.g. `26.06.22-03` |
    | `mode` | `dry-run` → then `apply` |
-2. Holding onto a **tainted** version is refused unless you also set `force` = `true`. Don't,
-   unless you're deliberately choosing the least-bad option and have said so in the ticket.
+2. Holding onto a **tainted** version is refused unless you also set `force` = `true`.
+   See [`force`](#force--holding-onto-a-tainted-version) below before you do.
+
+### `force` — holding onto a tainted version
+
+`force` lifts exactly one guard: the check that stops `hold` pointing a track at a version
+carrying a `<version>_tainted` marker. It does nothing for `taint`, `untaint`, or
+`release-hold` — on those actions the checkbox is inert.
+
+**Why it exists.** Taint and hold answer different questions. Taint says *"no track may
+advance onto this."* Hold says *"this track sits here, now."* Those only collide in one
+situation: **every candidate is bad and you have to pick the least-bad one.** A track is on a
+release that is actively breaking customers, and the only older release you can retreat to is
+itself already tainted for something milder. Retreating is still correct. `force` is how you
+say so out loud.
+
+**What it costs.** The `_tainted` marker stays, but it stops protecting *this* track. A held
+track bypasses the planner completely — the planner is the only thing that reads taint markers
+— and promote instead reconciles the floating tag back to the hold digest on every run. So the
+track is not merely parked on a known-bad release; it is **actively re-pinned there** until
+someone runs `release-hold`. Other tracks are unaffected: the marker still excludes that
+version for them.
+
+**Rules**
+
+- Never for convenience. If any untainted version is acceptable, hold onto that instead.
+- Record why in the incident ticket, naming **both** defects — the one you fled and the one
+  you knowingly accepted.
+- Treat it as temporary. Exit by `release-hold` once a good release exists — **not** by
+  untainting the version to make the warning go away.
+
+> **The guard is a one-time check, at hold time.** If you hold onto a good version and it is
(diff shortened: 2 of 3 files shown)


sfreudenthaler:
Verified against a secrets-free copy of the dispatch surface in [dotCMS/core-workflow-test](https://github.com/dotCMS/core-workflow-test/actions/workflows/evergreen-tracks-admin-ui-demo.yml) — identical `workflow_dispatch` inputs and `run-name`, but no Docker login, no secrets, and the step only echoes the command the real workflow would have run.

### Actions list, verbatim

```
success  |  APPLY — release-hold standard
success  |  APPLY — hold standard 26.08.14-01
success  |  DRY-RUN (no changes) — hold standard 26.08.14-01
success  |  APPLY — taint 26.08.28-01
success  |  DRY-RUN (no changes) — taint 26.08.28-01
```

That is the whole point of the change: on `main` today every one of those rows reads `evergreen-tracks-admin` and is green, which is exactly how [run 33434507975](https://github.com/dotCMS/core/actions/runs/33434507975) was read as a completed taint.

### The demo caught a bug

The first dispatch rendered:

```
DRY-RUN (no changes) — hold 26.08.14-01standard
```

`hold` sets **both** `version` and `track`, and the original `${{ inputs.version }}${{ inputs.track }}` concatenated them with no separator. Fixed in 165140547f by giving each present field its own leading space; all four action shapes (`taint`, `untaint`, `hold`, `release-hold`) now render correctly, as shown above.

### Also confirmed

`gh workflow run ... -f mode=dry-run` was accepted against a ref whose workflow declares `mode`, which confirms GitHub renders the dispatch form from the **selected ref** — so the new form can be previewed on this PR's branch before merge (Actions → evergreen-tracks-admin → Run workflow → branch `issue-evergreen-admin-dryrun-ui`).

I will delete the demo workflow from `core-workflow-test` once this merges — say the word if you would rather keep it.

sfreudenthaler:
## Screenshots

All real captures — no mockups. The comparison repo is [dotCMS/core-workflow-test](https://github.com/dotCMS/core-workflow-test/actions/workflows/evergreen-tracks-admin-ui-demo.yml), running a secrets-free copy of this dispatch surface (no Docker login, no secrets, step only echoes).

### Before — `dotCMS/core` today

before

Three dispatches, three identical green rows titled `evergreen-tracks-admin`. Two of them were dry-runs. Nothing on this screen tells you which.

### After — same workflow with `run-name`

after

### Dry-run now annotates the run summary

notice

### The demo caught a bug

Row #3 above reads `hold 26.08.14-01standard` — `hold` sets **both** `version` and `track`, and the original expression concatenated them with no separator. Fixed in 165140547f; row #4 is the same dispatch after the fix.

sfreudenthaler:
## The dispatch form itself

Captured from the live authenticated **Run workflow** panel.

| BEFORE — `dotCMS/core` `main` | AFTER — this change |
|---|---|
|  |  |

**Before:** `Actually mutate tags (false = dry-run)` is the **last** of six fields — a free-text box you have to scroll past five others to reach, pre-filled with the word `false`. `Allow holding onto a tainted version` is a second identical `false` text box directly above it. Two lookalike boxes at the bottom of the form, one of which is the only thing standing between a dry-run and a real registry mutation.

**After:** `Mode` is the **first** field, a real dropdown (`dry-run` shown expanded, `apply` below it), with the consequence written into the label. `force` is a checkbox, so it can no longer be confused with the mode box.

<sub>Note on the expanded dropdown: native `&lt;select&gt;` popups are OS-drawn and do not appear in browser screenshots, so the `mode` options are rendered inline as a list box. The options and their order are exactly as the workflow declares them — `dry-run` first (and therefore preselected), `apply` second.</sub>

hmoreras (dismissed):

sfreudenthaler:
@hmoreras — sorry, I dequeued this after your approval to fold in one more commit rather than send you a separate PR for the same concern. Your approval was auto-dismissed by the push; re-requested.

**What changed since you approved:** one commit, `386b54c241`, docs only plus one input-description string.

- `RUNBOOK.md` — a `force` section: what it does, why it exists, what it costs, and the exit path
- `README.md` — a table of the three registry markers, and a `force` summary
- `cicd_evergreen-tracks-admin.yml` — the `force` input description becomes `hold only — pin onto a TAINTED version (last resort, see RUNBOOK)`

No behaviour change from that commit. The workflow logic you already reviewed is untouched.

Worth a look while re-reviewing: verifying this turned up a latent trap. The tainted check runs only at hold time (`cli.py:210`), so if you hold a track onto a good version and taint that version later, the hold survives and promote keeps re-pinning the track onto it — silently. Documented here, but it probably wants a code fix in its own issue.

hmoreras (approved):

PR #37332, opened 2026-09-01
fix(permissions): File Asset Containers inherit from their folder, not the Site

### LINKED ISSUE #37331: File Asset Container permissions resolve to the Site, not the folder
## Problem Statement

A non-admin user who has been granted View on a File Asset Container's folder opens the layout
editor of a page and the Container is simply **not in the list**. Nothing was changed: not the
folder's permissions, not the user's roles, not the Container itself. Some users in the same site
still see it; others do not. It can come back on its own and disappear again later.

The reason it looks random is that the visibility depends on **which code path last recalculated
the permissions of the `container.vtl` file**, and that is decided by whichever request happens to
arrive first after dotCMS discards its cached inheritance data. Nothing in the UI tells the user
this happened.

**Who is affected:** any customer using File Asset Containers (containers defined as folders under
`/application/containers/`) together with folder-level permissions for limited back-end roles —
i.e. the supported way of letting an editor use a Container without giving them Site-wide access.
Administrators never see the problem, because admin permission checks short-circuit before any of
this code runs. That is also why it survives support triage for so long: it is invisible to the
person investigating it.

### Root cause

A File Asset Container is not a database Container. It is the `container.vtl` file inside a folder
under `/application/containers/`. dotCMS therefore loads the very same asset under **two different
Java identities**, depending on what the caller asked for:

| | Loaded as `Contentlet` (the file) | Loaded as `FileAssetContainer` |
|---|---|---|
| `getPermissionId()` | `container.vtl` identifier | the same identifier |
| `getPermissionType()` | `...model.Contentlet` | `...model.Contentlet` (overridden — `FileAssetContainer.java:103`) |
| `getParentPermissionable()` | the **container folder** | the **Site** (inherited from `Container.java:333-348`) |

The first two rows match by design. The third does not, and that is the defect.

Inheritance is cached in the `permission_reference` table, which is declared
`create unlogged table permission_reference (... unique (asset_id))`
(`dotCMS/src/main/resources/postgres.sql:538-545`). Note the **`unique (asset_id)`**: there is one
single row per asset. The two identities cannot both be represented. So whichever one rebuilds the
row last overwrites the other:

* rebuilt through the file path → `reference_id` = the container folder inode → the folder's
  permissions apply → **the editor sees the Container**;
* rebuilt through the Container path → `reference_id` = the Site id (or `SYSTEM_HOST`) → the
  folder's permissions are skipped entirely → **the editor loses the Container**.

The walk-up itself is working as designed: `PermissionBitFactoryImpl._loadParentPermissions()`
(`:1810-1826`) climbs until it finds an ancestor holding inheritable permissions of the asset's
type and adopts that ancestor's set. It is being handed the wrong starting parent.

### Why it appears and disappears on its own

The `permission_reference` row is a derived cache, and dotCMS legitimately discards it in normal
operation — after which the next request to touch that asset rebuilds it, and either identity may
get there first. Known triggers:

* saving permissions on any ancestor with **Force Cascade Permissions** checked
  (`PermissionAjax.java:308` → `ResetPermissionsJob`);
* moving a folder (`FolderAPIImpl.java:1201`);
* a push publish, a content type save, a content checkin on a type with a host field
  (`ESContentletAPIImpl.java:6006`);
* an unclean database shutdown — the table is `UNLOGGED`, so PostgreSQL truncates it on crash
  recovery (that is deliberate and safe by design, see `Task260403SetPermissionReferenceUnlogged`).

The request that then rebuilds it as a **Container** is the everyday one:
`PageRenderUtil.addPermissions()` (`:498-500`) checks WRITE and READ against the `Container` object
for every container on a page being rendered in edit mode. A limited user opening a page that uses
the Container can therefore poison the row **for everyone**, including themselves.

### Behavior change reviewers should be aware of

Once the reference correctly points at the container folder, and that folder carries its own
inheritable permissions, the folder becomes the nearest ancestor with an inheritable `Contentlet`
set — so the walk-up stops there and **Site-level grants no longer reach the Container**. A role
that could see the Container purely through a Site-level inheritable content grant, with nothing
granted on the folder, will lose it.

This is not a new rule introduced by the fix. It is the rule `container.vtl` already obeys when it
is read as a file, and it is why the two identities must agree. It is called out here because it is
a user-visible change for any environment that was relying on the broken side of the coin.

---

### Why this needs one SQL DELETE

**Everything in the setup is done in the dotCMS UI.** SQL appears as exactly two kinds of
statement, and never to inject a wrong value:

1. **One `DELETE`** that removes the cached inheritance row for the test asset. dotCMS deletes this
   row by itself many times a day (see the trigger list above), but there is **no UI action that
   deletes only that row without also changing the permissions we just set up** — which would
   destroy the very state under test. The `DELETE` restores a state dotCMS creates routinely; it
   does not create an invalid one.
2. **`SELECT`s** that show the resulting value, because the row is internal and no screen displays
   it. They are read-only.

The alternative — reproducing purely through the UI — works, but it is a **race**: after the cache
is discarded, either identity may rebuild the row first, so the defect may or may not appear on a
given attempt. That is precisely why the customer's environment healed and broke repeatedly, and
why it is not a reliable QA procedure. The production-like variant below describes it for anyone who
wants to observe that behavior; **the Deterministic Path is the one to validate against.**

### Prerequisite: make the reference write synchronous

Set this on the instance and restart **before** starting:

```
DOT_PERMISSION_REFERENCES_UPDATE_ASYNC=false
```

By default `PERMISSION_REFERENCES_UPDATE_ASYNC` is **true**
(`PermissionBitFactoryImpl.java:1736`), so dotCMS writes the `permission_reference` row on a
background thread. The verdict queries below then race that write and come back with **zero rows**,
which reads like "nothing happened" when in fact the write is simply still in flight. Every dotCMS
integration test that touches this table turns the flag off for the same reason.

### Setup (dotCMS UI only)

**Perform steps 1-8 as an administrator.** The limited user is only used from step 9 onwards. Two
notes that save time:

* Leave the **Reset all children permissions to match the above?** checkbox **unchecked**. It is not
  needed (inheritance is resolved on demand, not copied at save time) and it fires
  `ResetPermissionsJob`, which is one of the triggers this issue is about — checking it here just
  muddies the starting state.
* **Order matters from step 9 onwards.** Whichever screen the limited user opens *first* decides
  which code path rebuilds the inheritance row, and that is the whole defect. Follow the steps in
  order and do not open the Page ahead of its turn.

1. Use the existing `demo.dotcms.com` Site. Referred to below as **SITE**.
2. Create a File Asset Container: in Site Browser, create the folder
   `/application/containers/qa-test-container/` and add a `container.vtl` file to it containing:

   ```velocity
   $dotJSON.put("title", "Repro Container")
   $dotJSON.put("max_contentlets", 25)
   ```

   Confirm it appears under **Containers** in the back end.
3. Create a role named **QA Container Editor** (Settings → Roles & Tools). In the **Can Grant**
   field, tick all three checkboxes: **Users**, **Permissions** and **Tools**.
4. Give the role the portlet it needs: **Settings → Roles & Tools** → click *QA Container Editor* →
   **TOOLS** tab → tick the **Site** box.
5. Still on the *QA Container Editor* role, open the **PERMISSIONS** tab, add the **SITE**
   (`demo.dotcms.com`), and tick:
   * **View** in the Site's own row — lets the user browse the Site;
   * **Edit** in the **Pages** row — lets the user open a Page in edit mode, which is the trigger in
     step 12.

   Click **Apply Changes**.
6. On the folder `/application/containers/qa-test-container/` → **Permissions** tab:

   1. Click **Permission Individually**. This breaks inheritance from the parent so the folder
      carries its own permissions.
   2. Add the role *QA Container Editor*, and in its row tick **View**. This is View on the folder.
   3. Expand that role's row. Under *"Select what permissions QA Container Editor will have on the
      following children:"* find the **Content/Files** row and tick **View**. This is the
      inheritable grant that actually reaches `container.vtl`.
   4. Click **Apply Changes**.

   Both ticks are required and they are not the same thing: (2) lets the user see the folder,
   (3) lets the user see the assets inside it. This is the supported way to give a limited user a
   File Asset Container.

   > ⚠️ **Leave `container.vtl` inheriting — do not give the file its own permissions.**
   > The file has no permissions of its own by design: it inherits them from the folder, and that
   > inherited link is exactly what this defect corrupts.

7. Now that the role is fully configured, create the test user. Tick the **Back-end User** box so it
   can log into dotCMS. In the user's **ROLES** tab, add the *QA Container Editor* role.
   Confirm the user is **not** an administrator — an administrator can never reproduce this, because
   admin permission checks short-circuit before the code under test runs.
8. Still as an administrator, create a Template and a Page using that Template. Edit that Page's
   layout and add the `Repro Container` to it.

### Confirm the healthy state

9. As the limited user, read **the Container list only** — the list the layout editor's picker is
   built from. Use the following CURL command (update it to use your test User's credentials):

   ```bash
   curl -u 'limited.user@example.com:thepassword' \
     'http://localhost:8080/api/v1/containers/?filter=&per_page=100&system=false'
   ```

   The Container **is** in the response. ✅

   > ⚠️ **Do not open the Page yet.** Opening a Page that uses the Container is the trigger, and it
   > belongs to step 12. Doing it here rebuilds the inheritance row through the Container path
   > before the healthy state has been recorded, and step 9 fails for the very reason the test is
   > meant to demonstrate later.

Record the two ids you will need (read-only):

```sql
-- The container.vtl identifier -> VTL_ID
SELECT id AS vtl_id, parent_path, asset_name
  FROM identifier
 WHERE asset_name  = 'container.vtl'
   AND parent_path = '/application/containers/qa-test-container/';

-- The container folder inode -> FOLDER_INODE
SELECT f.inode AS folder_inode
  FROM folder f
  JOIN identifier i ON f.identifier = i.id
 WHERE i.parent_path = '/application/containers/'
   AND i.asset_name  = 'qa-test-container';
```

Verdict query — run it at every checkpoint below:

```sql
SELECT pr.asset_id,
       pr.reference_id,
       CASE WHEN pr.reference_id = '<FOLDER_INODE>'
            THEN 'CORRECT - inherits from the container folder'
            ELSE 'WRONG - inherits from the Site or System Host'
       END AS verdict
  FROM permission_reference pr
 WHERE pr.asset_id = '<VTL_ID>';
```

*No rows returned* means the cached row is absent — that is the expected state right after step 11,
not an error.

### Deterministic Path

10. Run the verdict query. Expect **CORRECT - inherits from the container folder** (the file path
    built the row when the Container list was read in step 9).
11. Clear the cached inheritance for this one asset:

    ```sql
    DELETE FROM permission_reference WHERE asset_id = '<VTL_ID>';
    ```

    Then flush the caches from the UI as the administrator: **Settings → Maintenance → CACHE →
    Flush**. Without this, dotCMS keeps answering from memory and nothing is rebuilt. Return to the
    limited user for step 12.
12. As the limited user, render the Page in edit mode with this CURL command:

    ```bash
    curl -u 'limited.user@example.com:thepassword' \
      'http://localhost:8080/api/v1/page/render/<page-uri>?mode=EDIT_MODE'
    ```

13. Run the verdict query again:

    * **Before the code fix:** `WRONG - inherits from the Site or System Host` — `reference_id` is the
    Site id or `SYSTEM_HOST`.
    * **After the code fix:** `CORRECT - inherits from the container folder` — `reference_id` is the
    container folder inode.

14. As the limited user, read the Container list again — the identical call from step 9:

    ```bash
    curl -u 'limited.user@example.com:thepassword' \
      'http://localhost:8080/api/v1/containers/?filter=&per_page=100&system=false'
    ```

    * **Before the code fix:** the Container is **gone** from the response, with no error or message. 🐞
    * **After the code fix:** the Container is still listed, exactly as in step 9.

    The same loss is visible in the UI on the **Layout tab / Template Builder** container picker,
    which calls this endpoint (`dot-containers.service.ts:95`).

    > **Do not look for it on the Page itself — it will not show there.** Opening the Page in edit
    > mode still renders the Container and leaves it usable, because
    > `PageRenderUtil.addPermissions()` (`:503-509`) computes the user's READ result and then
    > discards it whenever `SIMPLE_PAGE_CONTENT_PERMISSIONING` is true, which is the default:
    > ```java
    > contextMap.put("USE_CONTAINER_PERMISSION" + container.getIdentifier(), true);
    > ```
    > So the Page render is what *poisons* the row, yet the Page is the one screen where the damage
    > is invisible. The loss only surfaces where a user goes to **add** a Container to a layout.

### Production-like variant — non-deterministic (optional)

Do the same setup, then instead of the `DELETE`: as an admin, open **SITE → Permissions**, make any
permission change with **Force Cascade Permissions** checked, and save. This is what discards the
cached rows in a real environment. Then have the limited user browse normally. The Container will
disappear **on some attempts and not others**, depending on which request rebuilds the row first.
This is the customer-reported experience; it is not suitable as a pass/fail QA step.

### Suggested automated coverage

The defect can be pinned down deterministically in an integration test, at three levels. Each of
these should fail before the fix and pass after it:

| Scenario | What it proves |
|---|---|
| The `permission_reference` row for a File Asset Container must resolve to the container folder | The same asset resolves to two different parents depending on how it was loaded. |
| A limited user must keep seeing the Container after the reference is rebuilt through the Container path | The user-visible consequence: the Container vanishes from the picker. |
| A folder-granted role must not lose View to a Site-level grant held by another role | The customer's role split, and the case where the Site has an inheritable grant of its own — this is what produces the Site id rather than `SYSTEM_HOST`. |

Note that such a test must set `PERMISSION_REFERENCES_UPDATE_ASYNC` to `false`, for the same reason
the manual reproduction does.

---

## Proposed fix

Override `getParentPermissionable()` in `FileAssetContainer` so it resolves the **container folder**
from the Container's own Site and path, matching what the `Contentlet` view of the same file already
returns, and falling back to the current behavior when the instance carries neither (for example the
stub built while `container.vtl` is being deleted).

**Callers that assumed the parent is always a Site must be fixed with it.** Before this change every
Container's parent Permissionable was a `Host`, and two callers relied on that:

* `ContainerPaginator.hostname()` cast it with `Host.class.cast(...)` while sorting. A
  `ClassCastException` there escapes the whole `GET /api/v1/containers` request, so the Template
  Builder lists **no Containers at all** — a far louder failure than the bug being fixed. It must ask
  the Container for its own Site name (`getHostName()`), which both Container types already answer
  correctly.
* `PageAPIGraphQLTypesProvider` exposes a Container's `parentPermissionable` as a GraphQL **Site**.
  It must resolve the Site from the Container's own host id instead.

Reviewers should treat "does any caller expect a `Host` back?" as the main review question for this
change. `Container.getHostId()`/`getHostName()` also cast, but `FileAssetContainer` already overrides
both from its own Site field, so they are safe.

---

## Acceptance Criteria

- [ ] A File Asset Container and its `container.vtl` file resolve to the **same** parent
      Permissionable — the container folder — regardless of which identity triggered the load.
- [ ] The `permission_reference` row for a File Asset Container points at the container folder inode
      after the reference is rebuilt through the Container code path.
- [ ] A limited user granted View on the container folder keeps the Container in the layout editor
      across permission-cache flushes, cascaded permission saves, push publishes and restarts.
- [ ] The three scenarios above are covered by integration tests, registered in the integration test
      suite, that fail before the fix and pass after it.
- [ ] No regression for database Containers, which must keep inheriting from the Site.
- [ ] The behavior change noted above (Site-level-only grants no longer reaching a Container whose
      folder has individual permissions) is confirmed as intended and called out in the release
      notes.

---

## dotCMS Version

Reproduced on `Current Release (dotEvergreen)`. Present in current and recent LTS versions; the two
methods involved are long-standing.

## Severity

High — Major functionality broken. Affected editors cannot add or manage Containers on their pages,
there is no error message explaining why, and the condition recurs on its own after being repaired.
Mitigated by the fact that a workaround exists (see below) and that administrators are unaffected.

## Links

- [Support ticket #38795](https://dotcms.freshdesk.com/a/tickets/38795) (re-open of support ticket 38576)
- Related: #35680 — added the empty-result guard in `loadPermissions()`. That guard only prevents an
  *empty* permission list from being cached; it does not apply here, because the walk-up returns a
  perfectly non-empty set from the Site. Different symptom, same table.
- Related: #33665

### Interim workaround for affected environments

Repair the reference and then make it immune by giving `container.vtl` its own individual
permissions, so it no longer inherits from anything. Step-by-step runbook (including the queries and
the ordering constraints) was produced for support during this ticket and can be attached on request.

jcastro-dotcms:
### Proposed Changes

This PR fixes: #37331

A File Asset Container is the `container.vtl` file under `/application/containers/<name>/`. It already reports `Contentlet` as its permission type, so both views of the file share one permission identity — but its parent Permissionable was still the `Container` one (the **Site**), while the `Contentlet` view of the same file resolves the **container folder**.

Because `permission_reference` is keyed by `asset_id` alone (`unique (asset_id)`), whichever view was loaded last overwrote the row written by the other. Once the Container view won, every grant made on the Container folder was replaced by whatever the Site inherits, and roles holding View only on the folder stopped seeing the Container in the layout editor.

* **`FileAssetContainer.getParentPermissionable()`** — new override resolving the Container folder from the instance's own Site and path, so both identities agree and the reference row is stable. Instances carrying neither Site nor path (the stub built while `container.vtl` is being deleted) keep the previous behavior.
* **`ContainerPaginator.hostname()`** — was `Host.class.cast(container.getParentPermissionable())` while sorting. That cast now receives a `Folder`, and the resulting `ClassCastException` escapes the whole `GET /api/v1/containers` request, so the Template Builder lists **no Containers at all**. It now asks the Container for its own Site name via `getHostName()`, which both Container types answer correctly, and falls back to an empty sort key rather than failing the request.
* **`PageAPIGraphQLTypesProvider`** — the Container's `parentPermissionable` GraphQL field is declared as a **Site**, so it now resolves the Site from the Container's own host id instead of from its permission parent.
* **`FileAssetContainerPermissionInheritanceTest`** — new integration test, registered in `MainSuite3a`.

The main review question for this change is *"does any caller expect a `Host` back from `getParentPermissionable()`?"*. Every consumer in `dotCMS/src/main` was audited; the two above were the only ones affected. `Container.getHostId()`/`getHostName()` also cast, but `FileAssetContainer` overrides both from its own Site field.

**Behavior change worth calling out:** once the reference points at the folder and that folder carries its own inheritable permissions, the folder is the nearest ancestor with an inheritable `Contentlet` set, so Site-level grants no longer reach the Container. A role that could see a Container purely through a Site-level inheritable content grant, with nothing granted on the folder, loses it. That is the rule `container.vtl` already obeys when read as a file — which is precisely why the two identities must agree — but it is user-visible for environments relying on the previous behavior.

### Checklist
- [x] Tests
- [ ] Translations

This is a permission-resolution change, so it was treated as security-relevant throughout. It makes the effective permissions of a File Asset Container **match the folder that administrators actually configured**, rather than silently substituting the Site's. No permission is widened: the folder's own grants become authoritative, and the Site's inheritable grants stop leaking past a folder that has broken inheritance. The behavior change noted above is a narrowing, not a widening.

### Additional Info

Three integration test methods, all failing before the fix and passing after it:

| Test method | What it proves |
|---|---|
| `permissionReferenceForFileAssetContainerMustResolveToContainerFolder` | The same asset resolved to two different parents depending on how it was loaded. |
| `limitedUserMustKeepSeeingContainerAfterReferenceIsRebuiltAsContainer` | The user-visible consequence: the Container vanished from the picker. |
| `folderGrantedRoleMustNotLoseViewToSiteLevelGrantOfAnotherRole` | The reported role split, including the case where the Site holds an inheritable grant of its own — which is what produces the Site id rather than `SYSTEM_HOST`. |

The test sets `PERMISSION_REFERENCES_UPDATE_ASYNC` to `false`, since the reference write is otherwise performed on a background thread and the assertions would race it.

Manual reproduction steps for QA, verified on an unfixed build, are in #37331.

### Screenshots

N/A — the defect and the fix are both observable through `GET /api/v1/containers/` and the `permission_reference` table; there is no visual change.

https://claude.ai/code/session_012CsKDCNudZcysg4sPKHijU

---
_Generated by [Claude Code](https://claude.ai/code/session_012CsKDCNudZcysg4sPKHijU)_


jcastro-dotcms:
Valid finding — fixed in bf637b07a1.

I verified all three factual claims:

1. `Try.of(...).getOrNull()` catches `Throwable`, logs nothing.
2. The fallback is `super.getParentPermissionable()` → the Site.
3. **The wrong parent is persisted.** `PermissionBitFactoryImpl` guards the *cache* write with `if (!permissionList.isEmpty())` (`:1726`), but the `upsertPermissionReferences` at Step 4 (`:1735-1747`) is unconditional. So a transient failure writes a Site-pointing `permission_reference` row that survives until something clears it.

On the assumption you flagged for verification — a transient failure being realistic rather than only "folder missing": yes. `folderFactory.findFolderByPath` hits cache then DB, so a pool timeout mid-request is an ordinary production event.

**I went further than the suggested remedy.** Logging alone would make the failure diagnosable but still persist the wrong parent. The two cases turn out to be already separable in the API — `FolderAPIImpl.findFolderByPath` (`:138-141`):

```java
final Folder folder = folderFactory.findFolderByPath(path, host);
if (folder == null || UtilMethods.isEmpty(folder.getInode()) || permissionAPI.doesUserHavePermission(...)) {
    return folder;
}
```

A genuinely absent folder comes back as **null or inode-less** — it does not throw. It only throws on a real failure. The `Try` was collapsing those two into one silent `null`.

So the fix keeps the null / inode-less / `SYSTEM_FOLDER` path falling back to `super` (the legitimate stub case, e.g. the instance built while `container.vtl` is being deleted) and lets `DotDataException` propagate. `getParentPermissionable()` already declares it via `Permissionable`, and `_loadParentPermissions` is inside a `throws DotDataException` method, so no signature change was needed. `DotSecurityException` — unreachable for the System User — is rethrown as `DotDataException` with the folder path and Site name rather than ignored, so it can never become a silent downgrade either.

Net effect: a failed lookup now fails loudly, **nothing is written**, and the next request resolves the reference again. That matches the principle already stated three lines above the upsert in the #35680 comment — *"a failed walk-up must be retried, not persisted"* — applied to the reference row rather than the cache.

One trade-off worth naming for human reviewers: propagating turns a transient DB blip into a failed page render instead of a silent degradation to Site permissions. For a permission resolver I think loud-and-retryable beats quietly-wrong-and-durable, but it's a judgment call someone may want to take the other way.

---
_Generated by [Claude Code](https://claude.ai/code)_

jcastro-dotcms:
Valid finding — fixed in f1dfff3c90.

Confirmed: the class had only `@BeforeClass` (`:77`) and `@After` (`:91`), no `@AfterClass`. `PERMISSION_REFERENCES_UPDATE_ASYNC` was pinned to `false` and never put back, so with the class registered in `MainSuite3a` every class scheduled after it inherited synchronous permission-reference upserts.

Also confirmed both cited precedents do capture-and-restore: `ContentDriveLinksTest:368-370` / `:438-440` and `MenuLinkAPITest:113-115` / `:143-145`, each in a `try`/`finally`.

The fix captures the old value in `@BeforeClass` and restores it in a new `@AfterClass`. One deliberate deviation from the precedents: they scope the flag per test method with `try`/`finally`, whereas all three methods here need it, so it stays class-scoped. That is exactly what makes the restore mandatory rather than optional, and the comment on `prepare()` now says so — so the next person to add a method to this class does not "simplify" the `@AfterClass` away.

The property name is also lifted to a constant now that it appears three times.

---
_Generated by [Claude Code](https://claude.ai/code)_

danielsolis-dotcms (dismissed):

dotCMS/src/main/java/com/dotcms/util/pagination/ContainerPaginator.java
```diff
@@ -159,11 +160,29 @@ private PaginatedArrayList<Container> sortByTypeAndHost(
...
+     */
     private String hostname (final Container container) {
 
         try {
-            return Host.class.cast(container.getParentPermissionable()).getHostname();
-        } catch (DotDataException e) {
+            final String hostName = container.getHostName();
+            return UtilMethods.isSet(hostName) ? hostName : StringPool.BLANK;
+        } catch (final Exception e) {
+            Logger.debug(this, () -> String.format(
```
danielsolis-dotcms: Maybe we can have a Logger.warn here instead of Logger.debug? Just in case we want the message to be included by default on the logs

jcastro-dotcms: Done!

jcastro-dotcms:
Closing this PR as several commits are not verified. New PR: https://github.com/dotCMS/core/pull/37402

PR #37333, opened 2026-09-01
fix(search): make the OpenSearch write leg durable once it serves reads (#37276)

fabrizzio-dotCMS:
Follow-up to #37320 (merged). Addresses the Phase 2 gap raised in review of that PR, plus one defensive fix. Part of #37276.

## The gap

ADR-0009 says a failed write to the shadow store is logged and must not impact operations. That is correct for **Phase 1**, where nothing reads from OpenSearch.

From **Phase 2** onwards it stops being true. `PhaseRouter#readProvider` serves reads from OpenSearch (`readProvider():144` — `isReadEnabled() ? osImpl : esImpl`) while writes still fan out ES-primary / OS-shadow. So a removal lost on the OS leg left an orphaned document **in the very index being queried** — which is #37276 itself, in the phase the migration spends the longest in.

Both paths were affected, and the async one is the reason the durable-delete fix did not cover it:

- **Async (journal).** `createBulkProcessor` built a shadow `BulkProcessorListener` for OS whenever `isDualWrite`, i.e. Phases 1 **and** 2. A shadow listener never marks the reindex-queue entry failed, so the journal entry was acked on the ES result alone and the removal was never retried.
- **Sync.** `putToIndex` caught the OS exception and logged it as shadow divergence.

## The change

Scope the shadow treatment by **who serves reads**, not by dual-write:

```java
final boolean shadow = isDualWrite && ops == operationsOS && !isReadEnabled();
```

OpenSearch stays fire-and-forget while nothing reads it, and becomes durable the moment it does. ADR-0009's *intent* — don't let a store nobody reads break user operations — is preserved; only its Phase 2 assumption that a shadow is not read from is corrected.

When both legs fail, the ES exception is still the one raised. That is what callers have always seen, and demoting it would change behaviour beyond this gap.

Also guards `ContentletIndexOperationsOS.handleBulkResponse` against throwing a blank-message exception when a bulk reports `errors()` but no item carries a cause — flagged in the same review. A failure with no message is barely better than the silent return it replaced.

## Testing

`ContentletIndexAPIImplPhase2ReadDurabilityTest`, 3 cases. The first is the important one:

- **Phase 1 still swallows the OS failure.** This is the guard that stops the fix from over-reaching into the policy where ADR-0009 is correct. Without it, a later refactor could quietly make every shadow hiccup fail a user-facing save.
- Phase 2 surfaces the OS failure, and ES is still written first.
- Both legs failing raises the ES exception.

281 unit tests green. All 9 integration tests from #37320 re-run green against this change (`ReindexDeleteJournalTest`, `ContentletDestroyIndexRemovalTest`, `ContentletIndexPartialFailurePhaseTest`, `ContentletIndexProviderSkipTest`, one JVM).

## Spec

The spec on the base branch was amended in 945d27e6 to put this in scope with **AC-010**, alongside the three scope gaps raised separately in review. It had been a decision taken in code without being written down — the same complaint the reviewer had made about the other sites.

## Behaviour change worth flagging

This changes failure behaviour for **every** write in Phase 2, not only removals. An environment whose OpenSearch cluster is unhealthy will begin surfacing errors that were previously absorbed. Same "the defect becoming visible" effect as the partial-failure escalation in #37320, and confined to phases where OpenSearch already serves reads.

This PR fixes: #37276

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImpl.java
+++ b/dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ContentletIndexAPIImpl.java
@@ -2426,15 +2426,31 @@ public class ContentletIndexAPIImpl implements ContentletIndexAPI {
             } catch (final Exception e) {
                 esException = new DotRuntimeException(e.getMessage(), e);
             }
+            RuntimeException osException = null;
             try {
                 operationsOS.putToIndex(dual.osReq);
             } catch (final Exception e) {
-                Logger.warnAndDebug(this.getClass(),
-                        "OS shadow write failed in putToIndex — "
-                                + "OS index may diverge until next reindex. Cause: " + e.getMessage(), e);
+                if (isReadEnabled()) {
+                    // Phase 2: OS serves reads (PhaseRouter.readProvider), so a failure here is
+                    // not a shadow divergence — it leaves the index users actually query out of
+                    // sync with the database (#37276). Surface it once ES has been given its
+                    // chance, so the caller can retry rather than assume the write landed.
+                    osException = (e instanceof RuntimeException)
+                            ? (RuntimeException) e
+                            : new DotRuntimeException(e.getMessage(), e);
+                } else {
+                    Logger.warnAndDebug(this.getClass(),
+                            "OS shadow write failed in putToIndex — "
+                                    + "OS index may diverge until next reindex. Cause: " + e.getMessage(), e);
+                }
             }
+            // ES stays authoritative when both legs fail: its exception is the one callers have
+            // always seen, and demoting it would change behaviour beyond this fix.
             if (esException != null) {
                 throw esException;
+            }
+            if (osException != null) {
+                throw osException;
             }
         } else {
             // Single-provider phase (0 or 3): forward to the sole active provider.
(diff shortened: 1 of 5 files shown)


ihoffmann-dot:
Reviewed the diff in full (`ContentletIndexAPIImpl`, `ContentletIndexOperationsOS`, plus the new `ContentletIndexAPIImplPhase2ReadDurabilityTest`).

The fix is sound. `shadow = isDualWrite && ops == operationsOS && !isReadEnabled()` correctly scopes shadow treatment by who serves reads, not by dual-write, applied consistently on both the sync and async paths. Exception precedence (ES first, OS second, ES wins if both fail) preserves existing behavior. The blank-message guard is a good small defensive addition.

Test coverage is well-targeted too. Approving ✅

ihoffmann-dot (approved):

ihoffmann-dot (approved):

fabrizzio-dotCMS:
Folded into #37297 — the Phase 2 fix belongs in the PR that carries AC-010, and leaving it stacked meant #37297 read as internally inconsistent: its own `spec.md:255` declaring AC-010 while its contract denied it.

Same two commits, fast-forwarded, nothing lost. Closing.

PR #37344, opened 2026-09-02
fix(roles): PUT /v1/roles/{roleId} no longer returns roleChildren duplicated 4x (#37303)

### LINKED ISSUE #37303: PUT /v1/roles/{roleId} returns the role's children duplicated 4x, with childCount inflated to match
## Problem Statement

`PUT /api/v1/roles/{roleId}` returns the updated role with its `roleChildren` list **duplicated four times**, and `childCount` inflated to match. A role with one child comes back reporting four.

Only the response body is wrong. The persisted rows are correct — an immediate `GET` on the same role returns the real children — so the damage is limited to whatever a client does with the response it just received.

That is enough to break a UI. The Roles (Beta) portlet splices the PUT response into its tree instead of refetching, which is the normal way to avoid a round-trip after a save, so reparenting a role paints every child of that role four times until the admin reloads the page. Any other client that trusts the response — an integration reading `childCount` to decide whether to recurse, for instance — gets the same wrong answer.

There is a second, quieter consequence. `RoleFactoryImpl.save` walks the subtree to rewrite `DBFQN` after a reparent, and it seeds that walk from the same duplicated list:

```java
roleIdsToProcess.addAll(parentRole.getRoleChildren());
```

So every descendant is visited and re-saved once per duplicate. On a deep hierarchy that multiplies the write cost of a single reparent by four for no reason.

**Root cause.** `RoleFactoryImpl.populatChildrenForRolesHelper` (lines 859-870) reads the role's existing `roleChildren` and **appends** to it rather than replacing it:

```java
List<String> childrenList = roleMap.get(row.get("parentid")) != null?
        roleMap.get(row.get("parentid")).getRoleChildren(): null;
if (childrenList == null) { childrenList = new ArrayList<>(); }
childrenList.add(row.get("childid"));
```

A `Role` that arrives already populated therefore accumulates. Within one `save()` the same instance passes through `populatChildrenForRoles` more than once — in the `DBFQN` cascade loop and again near the end of the method — and it arrives populated in the first place because `RoleHelper.updateRole` copies from the instance `loadRoleById` returns, which is cache-resident and already carries its children.

`toRoleViews` then faithfully resolves whatever ids are in the list, and `RoleView.childCount` is computed as `getRoleChildren().size()`, so both fields carry the duplication out to the wire.

It is not cumulative across requests: a second `PUT` also returns four, not eight.

## Steps to Reproduce

Any authenticated CMS Admin, against a clean instance. No UI needed.

```bash
A=admin@dotcms.com:admin
H='Content-Type: application/json'
BASE=http://localhost:8080

mk() { curl -s -u $A -H "$H" -X POST $BASE/api/v1/roles -d "$1" | jq -r '.entity.id'; }

TARGET=$(mk '{"roleName":"TmpTarget","canEditUsers":true,"canEditPermissions":true,"canEditLayouts":true}')
PARENT=$(mk '{"roleName":"TmpParent","canEditUsers":true,"canEditPermissions":true,"canEditLayouts":true}')
mk "{\"roleName\":\"TmpChild\",\"parentRoleId\":\"$PARENT\",\"canEditUsers\":true,\"canEditPermissions\":true,\"canEditLayouts\":true}" >/dev/null

# 1 child, as expected
curl -s -u $A "$BASE/api/v1/roles/$PARENT?loadChildrenRoles=true" \
  | jq '{childCount: .entity.childCount, children: (.entity.roleChildren|map(.name))}'

# reparent -> the response reports 4
curl -s -u $A -H "$H" -X PUT "$BASE/api/v1/roles/$PARENT" \
  -d "{\"roleName\":\"TmpParent\",\"parentRoleId\":\"$TARGET\",\"canEditUsers\":true,\"canEditPermissions\":true,\"canEditLayouts\":true}" \
  | jq '{childCount: .entity.childCount, children: (.entity.roleChildren|map(.name))}'

# GET again -> back to 1, so nothing was persisted wrong
curl -s -u $A "$BASE/api/v1/roles/$PARENT?loadChildrenRoles=true" \
  | jq '{childCount: .entity.childCount, children: (.entity.roleChildren|map(.name))}'
```

Observed:

```
GET  before  ->  { "childCount": 1, "children": ["TmpChild"] }
PUT  move    ->  { "childCount": 4, "children": ["TmpChild","TmpChild","TmpChild","TmpChild"] }
GET  after   ->  { "childCount": 1, "children": ["TmpChild"] }
```

The reparent itself is applied correctly; `parent` points at `TmpTarget` afterwards.

Reproduced on a local instance built from `main`.

## Acceptance Criteria

- [ ] `PUT /api/v1/roles/{roleId}` returns each child of the updated role exactly once, and `childCount` equals the number of distinct children — matching what a subsequent `GET` on the same role returns.
- [ ] The same holds for a plain field update (no reparent) and for a reparent, including a move to root (`parentRoleId: null`).
- [ ] `populatChildrenForRolesHelper` is idempotent: calling `populatChildrenForRoles` twice on the same `Role` instance leaves the same list, not a doubled one.
- [ ] The `DBFQN` cascade in `RoleFactoryImpl.save` visits each descendant once per reparent. A subtree of N roles produces N updates, not 4N.
- [ ] Integration test covering the reproduction above: create parent + one child, `PUT` a reparent, assert the response carries one child and `childCount == 1`.
- [ ] Integration test asserting the cascade still rewrites `DBFQN` correctly for grandchildren and below after the fix — the dedup must not skip descendants.

## dotCMS Version

Reproduced on `main` (local build). The `populatChildrenForRolesHelper` accumulate-instead-of-replace behaviour is long-standing legacy code and is very unlikely to be a recent regression, so released versions are expected to be affected too. Worth confirming against the current LTS.

## Severity

## Links

NA

---

**Frontend context.** The Roles (Beta) portlet ships a client-side guard for this in dotCMS/core#37260: on reparent it keeps the `roleChildren` and `childCount` it already holds rather than taking them from the PUT response. That was needed anyway — the response only ever hydrates two levels, so lazy-loaded grandchildren are not in it — but it also means the visible symptom is already masked there. Please do not treat the portlet looking correct as evidence the endpoint is fixed; use the curl reproduction above.

hassandotcms:
Fixes #37303

## Problem

`PUT /api/v1/roles/{roleId}` returned the updated role with its `roleChildren`list duplicated four times and `childCount` inflated to match — a role with one child came back reporting four. Only the response body was wrong; the persisted rows were always correct (an immediate `GET` returned the real children). Enough to break a client: the Roles (Beta) portlet splices the PUT response into its tree, and any integration reading `childCount` to decide whether to recurse got the wrong answer. The reparent `DBFQN` cascade, seeded from the same list, also re-saved descendants redundantly.

## Root cause

`RoleFactoryImpl.populatChildrenForRolesHelper` **appended** to a role's existing `roleChildren` instead of replacing it. A `Role` instance that arrived already populated therefore accumulated. Within one `save()` the same instance passes through the population path four times (arrives populated from cache → the `DBFQN` cascade loop → `setFQNForDB`'s `getRoleById` reload → end of `save()`), producing exactly four copies. `toRoleViews` then serialized whatever ids were in the list, and `RoleView.childCount` is `getRoleChildren().size()`.

## Fix

Give `populateChildrenForRoles` **replace semantics**: reset each input role's `roleChildren` before the chunked queries, so population is idempotent. The reset is in the method (not the helper, which must keep accumulating across the 200-role chunks of a single population). This corrects the PUT response, removes the redundant cascade re-saves, and stops the shared cache-resident instance from being polluted mid-transaction. Also renames the private `populatChildrenForRoles*` method family to `populateChildrenForRoles*` (all call sites are in-file).

No REST contract change — response shape is untouched, only the values are corrected. No DB schema, ES mapping, or serialized-format change; rollback-safe.

## Testing

- **Integration (new):** PUT reparent / plain field update / move-to-root each return one child with `childCount == 1` matching a subsequent GET (`RoleResourceIntegrationTest`); factory-level idempotency, and a childless role
keeps `roleChildren == null` (`RoleAPITest`); reparent keeps children
de-duplicated at every depth of a 3-level subtree.
- **Suites:** `RoleResourceIntegrationTest`, `RoleAPITest`, `RoleResourceCountsIntegrationTest`, `RoleResourceUsersIntegrationTest` — all green (80 tests, 0 failures).
- **Manual (curl):** reparent PUT returns `childCount: 1` / `[child]` (was 4), matching GET before and after; plain update and move-to-root both `childCount: 1`.

This PR fixes: #37303


ihoffmann-dot (changes requested):
**AC-004 not directly tested.** The spec requires the DBFQN cascade to visit each descendant exactly once per reparent (today: 2x for direct children, worse deeper). The new depth test only checks that `roleChildren` lists end up correct, it doesn't count re-saves per descendant, and its own comment says so. Likely fixed as a side effect of the dedup, but nothing asserts it. Worth adding a write-count test or flagging AC-004 as indirectly covered.

hassandotcms:
> **AC-004 not directly tested.** The spec requires the DBFQN cascade to visit each descendant exactly once per reparent (today: 2x for direct children, worse deeper). The new depth test only checks that `roleChildren` lists end up correct, it doesn't count re-saves per descendant, and its own comment says so. Likely fixed as a side effect of the dedup, but nothing asserts it. Worth adding a write-count test or flagging AC-004 as indirectly covered.

updated the test, now its asserting in the correct manner

ihoffmann-dot (approved):

Issue #37063, opened 2026-08-14
Content Drive: bulk folder delete, async job endpoint and frontend wiring


zJaaal:
### Description

**Bulk** folder delete over a Content Drive multi-select, end to end: the asynchronous backend endpoint and the Content Drive wiring that drives it. One ticket, one operation, one deliverable — multi-select delete works in the product when this closes.

> **Re-split 2026-09-14.** This ticket previously carried the *frontend wiring for all three* folder operations (move, copy, bulk delete), with their backends on #37062 and #37165. The three tickets are now **one per operation, each carrying its own backend and frontend**, so each can be handed to one developer and finished without waiting on a sibling. **No design decision changed** — bulk delete is still an async job for the reasons recorded below. Copy is now #37062, move is #37165.

**Single-folder delete is already done and is not part of this ticket.** It shipped on #35161 as a **Delete Folder** context menu item over `POST /api/v1/assets/folders/_delete` (`WebAssetResource.java:314`), which takes one `assetPath` (`AbstractFolderDeletionRequestForm:24`) and needed no backend work. **That endpoint and its form are left exactly as they are.** What this ticket adds is the **multi-select** path, as a sibling bulk endpoint — the precedent workflow already sets with `.../bulk/fire` beside the single fire.

#### Why this cannot be a synchronous REST call

Delete is a **single unbounded transaction over a recursive walk**:

| Entry point | What runs inside that one transaction |
|---|---|
| `FolderAPIImpl.delete` `@WrapInTransaction` (`FolderAPIImpl.java:427`) | recursive delete of children at `:472` |

Deleting a folder holding thousands of assets holds one long transaction and times out at the proxy, and a multi-select multiplies it by the size of the selection. An earlier version of this work proposed a synchronous endpoint and designed around the expected timeout. Designing around an expected timeout is an async design with none of the machinery, so it is built as one.

Per-folder child lists are also loaded whole rather than paged, so peak memory is bounded by the widest single folder, not by the subtree. That is survivable; the transaction is the part that is not.

#### The job-queue framework already exists

No new infrastructure. `com.dotcms.jobs.business` is complete and Postgres-backed: `JobProcessor`, `@Queue`, `Cancellable`, `ProgressTracker`, retry policies, `PostgresJobQueue`, `AbandonedJobDetector`, `RealTimeJobMonitor`.

`JobQueueResource` (`/v1/jobs`) already exposes everything the caller needs after submission:

- `GET /v1/jobs/{jobId}/status`
[...]

Issue #37062, opened 2026-08-14
Content Drive: folder copy, async job endpoint and frontend wiring


zJaaal:
### Description

Folder **copy** over a Content Drive multi-select, end to end: the asynchronous backend endpoint and the Content Drive wiring that drives it. One ticket, one operation, one deliverable — copy works in the product when this closes.

> **Re-split 2026-09-14.** This ticket previously carried the *backend* for copy **and** bulk delete, with the frontend for copy, delete and move all sitting on #37063. The three tickets are now **one per operation, each carrying its own backend and frontend**, so each can be handed to one developer and finished without waiting on a sibling. **No design decision changed** — copy is still an async job for the reasons recorded below. Bulk delete is now #37063, move is #37165.

#### Why this cannot be a synchronous REST call

Copy is a **single unbounded transaction over a recursive walk**:

| Entry point | What runs inside that one transaction |
|---|---|
| `FolderAPIImpl.copy` `@WrapInTransaction` (`FolderAPIImpl.java:369,387`) | `FolderFactoryImpl.copy` recurses over file assets (`:501`), pages (`:516`), links (`:539`) and child folders (`:550`) |

Copying a folder holding thousands of assets holds one long transaction and times out at the proxy, and a multi-select multiplies it by the size of the selection. An earlier version of this work proposed a synchronous endpoint and designed around the expected timeout ("per-source transactions at least mean a timeout does not lose the folders already done"). Designing around an expected timeout is an async design with none of the machinery, so it is built as one.

Per-folder child lists are also loaded whole rather than paged, so peak memory is bounded by the widest single folder, not by the subtree. That is survivable; the transaction is the part that is not.

#### The job-queue framework already exists

No new infrastructure. `com.dotcms.jobs.business` is complete and Postgres-backed: `JobProcessor`, `@Queue`, `Cancellable`, `ProgressTracker`, retry policies, `PostgresJobQueue`, `AbandonedJobDetector`, `RealTimeJobMonitor`.

`JobQueueResource` (`/v1/jobs`) already exposes everything the caller needs after submission:

- `GET /v1/jobs/{jobId}/status`
- `POST /v1/jobs/{jobId}/cancel`
- `GET /v1/jobs/{jobId}/monitor`, an SSE `EventOutput`
- paginated `active` / `completed` / `successful` / `failed` / `canceled` / `abandoned`

[...]

PR #37358, opened 2026-09-02
feat(content-drive): bulk file upload backend

### LINKED ISSUE #37166: Content Drive: bulk file upload (multi-file selection uploads only the first file)
### Description

Content Drive accepts a multi-file selection, warns the user, and then **uploads only the first file**. There is no bulk upload on either side of the stack.

#### Current state, verified

The frontend is an explicit stub. `DotContentDriveShellComponent.resolveFilesUpload` branches on `files.length > 1` into `uploadFiles`, which shows a warning toast and then calls `uploadFile`, which uploads `files[0]` and drops the rest:

```
content-drive.multiple-files-warning=Multiple files upload is not supported yet.<br/> Only one file will be uploaded
```

(`Language.properties:7224`, under a `content-drive.work-in-progress` summary.)

Everything upstream of that already handles many files. `DotUploadDropzoneComponent` and `DotUploadButtonComponent` emit a `DotUploadFiles` payload carrying a `File[]`, drag-and-drop delivers the whole drop, and the type selector resolves one base type for the batch. The multi-file path exists right up to the request, and then discards the batch.

The backend has no bulk create either. `DotUploadFileService.uploadFileByBaseType` builds one `FormData` per file and fires `newContentletByBaseType`, so **one file equals one workflow fire request**. `WebAssetResource` (`/v1/assets`) exposes a single-asset `PUT` (`:164`) and nothing that accepts several. `workflow/.../_bulkfire` operates on existing contentlets and cannot create them from binaries. `ImportContentletsProcessor` is CSV import, a different operation.

The current call also passes `indexPolicy: 'WAIT_FOR'` (`dot-content-drive-shell.component.ts:1001`). Correct for one file, since the grid reloads immediately afterwards. Fired N times in a loop it serialises every upload behind an index refresh, so a naive "just loop the existing call" fix would be slow enough to look broken.

#### Two gaps, and they need separating

1. **Several files, one target folder.** The batch the UI already collects. This is the actual ask.
2. **Folder upload, meaning a directory tree.** `webkitdirectory` gives relative paths, so the folders have to be created before the assets land in them. Not proposed here; call it out as explicitly out of scope so it does not get assumed in.

#### Proposed shape: a job, reusing what #37062 introduces

Same reasoning as folder copy and bulk delete. A hundred files is a long operation with per-item outcomes, which is what the job queue is for, and the framework already has the multipart entry point for exactly this: `POST /v1/jobs/{queueName}/upload` consumes `MULTIPART_FORM_DATA` and is what `ImportContentletsProcessor` is fed through.

```
POST /api/v1/assets/_bulkupload   -> 202, queue "assetBulkUpload"
```

Multipart: the files, plus `targetFolder` (or site id at the root) and `baseType` (`DOTASSET` / `FILEASSET`). Answers `202` with `jobId` and `statusUrl`; status, cancel and progress come from the existing `/v1/jobs/{jobId}` endpoints.

The processor creates one contentlet per file through the same workflow fire the single-file path uses, so content type resolution, permissions and workflow stay identical. Per-file outcomes land in `getResultMetadata` with the same `successCount` / `failCount` / `results` shape #37062 defines, keyed by file name. Index policy moves off `WAIT_FOR` per file: index once at the end of the batch, or let the job's completion drive the single grid reload.

#### Partial failure is the normal case here

More so than for copy or delete. A batch of thirty files will routinely contain one that is too large, has a blocked extension, or collides with an existing asset name. The batch must not stop at the first failure, and the result has to name the files that did not make it. A single "upload failed" toast over a thirty-file drop is the outcome to avoid.

### Acceptance Criteria

**Backend**
- [ ] `POST /api/v1/assets/_bulkupload` accepts a multipart body with several files plus `targetFolder` and `baseType`, enqueues an `assetBulkUpload` job, and returns `202` with `jobId` and `statusUrl`
- [ ] The processor is `@Queue`-annotated, implements `Cancellable`, and reports progress per file through `Job#progressTracker()`
- [ ] Each file is created through the same workflow fire as the existing single-file upload, so content type resolution, permissions and workflow behavior are identical
- [ ] One file failing does not abort the batch
- [ ] `getResultMetadata` returns `successCount`, `failCount` and a per-file `results` array with a machine-readable error and a message for each failure, matching #37062's shape
- [ ] Files are not indexed with `WAIT_FOR` per item; the batch indexes once, or the job's completion drives the reload
- [ ] A user without add-children permission on the target folder is rejected at submission with `403`
- [ ] Cancelling leaves already-created contentlets in place and records in the result how far the batch got
- [ ] Per-file size and extension limits are enforced with the same rules as the single-file path, and a rejection is a per-file failure rather than a `400` for the whole batch
- [ ] `openapi.yaml` regenerated and committed with the annotation changes
- [ ] Integration tests cover: many files succeeding, mixed partial failure, a rejected extension, an oversized file, permission denied, cancellation mid-batch, and a batch large enough to exercise progress reporting

**Frontend**
- [ ] Selecting or dropping several files uploads **all** of them
- [ ] **Both upload entry points accept several files.** The hidden file input (`dot-content-drive-shell.component.html:178`) carries no `multiple` attribute, so the Upload-button path can only ever select one file and only drag-and-drop delivers a batch. `onFileChange` already reads the whole `FileList` and hands it to `resolveFilesUpload` unchanged, so this is the attribute plus its tests. The picker and the drop must behave identically
- [ ] Upload's copy follows the Action Center's shape rather than its own, so the two surfaces read the same: the in-flight label matches `content-drive.action-center.applying`, and the outcome toast matches `...toast.executed` / `-detail` / `-partial` instead of `content-drive.add-dotasset-success` and `-error`
- [ ] Content Drive stops using `content-drive.multiple-files-warning` and the `work-in-progress` summary, and the `files.length > 1` stub branch in `resolveFilesUpload` is removed. **Leave the definitions in place**: the AssetPicker still references them, along with `content-drive.file-upload-in-progress` and its `-detail`. It has the same first-file-only defect, tracked separately as #37370 and out of scope here
- [ ] Progress is shown for the batch while the job runs, via a **reusable core-web job-progress primitive owned by this ticket** (see below); it must not be Content-Drive-specific or upload-specific in its API
- [ ] The completion toast reports the server's `successCount` / `failCount`, never the number of files selected
- [ ] A partial failure names the files that failed and why, and stays on screen long enough to be read
- [ ] The grid refreshes once when the job finishes, not once per file
- [ ] Uploading stays available while the job runs, or is disabled with a reason; either way the UI does not appear frozen
- [ ] Navigating away does not lose the batch: the job keeps running and its outcome is still discoverable

#### Outcome copy unified with the Workflow Center

The Workflow Center names **what ran** and **what it ran on** (`<b>{0}</b> ran on {1} item(s).`). Lock, unlock, download and delete folder already do this. The rest do not, and the copy for the missing ones was written and never wired: `content-drive.toast.workflow-in-progress`, its `-detail`, and `content-drive.toast.workflow-executed-detail` (`Language.properties:7240-7243`) are referenced nowhere.

- [ ] Every operation's outcome names the action and the item it ran on, matching the Workflow Center's summary/detail split
- [x] Folder create's `summary: 'Success'` is a hardcoded English literal (`dot-content-drive-dialog-folder.component.ts:291`), not an i18n key. It is replaced, and the outcome names the folder
- [x] **Edit Folder**'s outcome (`content-drive.dialog.folder.message.save-success`, "Folder saved successfully") carries no detail and does not name the folder (:323). The action is the context menu's `content-drive.context-menu.edit-folder`, offered on EDIT permission, which reopens the create dialog with the folder as payload and branches to `saveFolder()` on submit
- [x] Both folder error paths pass the server's `error.message` straight through as the toast detail (:309 and :342). Replaced with resolved copy, with the raw message logged rather than shown
- [x] Lock and unlock **errors** name the item, as their success counterparts already do (`Locked {0}` versus "Something went wrong. The contentlet wasn't locked.")
- [x] The orphaned keys at `Language.properties:7240-7243` are resolved: `workflow-executed-detail` is wired up, and the two `workflow-in-progress` keys are deleted, since the toolbar indicator replaces them

#### In-progress feedback moves to the toolbar indicator

Content Drive currently reports "this is running" four different ways: the toolbar indicator (Action Center), an info toast (upload, drag-and-drop move), blanking the whole grid with skeleton rows (context-menu workflow action), and nothing at all. This ticket settles the rule and converts every intermediate state except reindex.

- [x] **The rule**: anything asynchronous that outlives the surface that triggered it reports its in-flight state on the toolbar indicator (`dot-content-drive-toolbar.component.html:107`). Not a toast, not the grid's loading state, not silence. The toast is reserved for the terminal state (success, partial, error, cancelled), and the grid's `LOADING` status means "the listing is being fetched" and nothing else
- [x] Content Drive stops using `content-drive.file-upload-in-progress` and its `-detail` (`dot-content-drive-shell.component.ts:1057`); the batch reports on the indicator
- [x] The drag-and-drop move's in-flight toasts are removed (`content-drive.move-to-folder-in-progress` and `-with-folders`, shell:1138 and 1150) and the move is routed through `withActionExecution` instead of calling `bulkFire` directly from the shell, so it reports on the same indicator. This is also what proves the keyed-runs change below, which upload alone cannot exercise
- [x] The context-menu workflow action stops calling `setStatus(DotContentDriveStatus.LOADING)` (`dot-folder-list-context-menu.component.ts:435` and :424), which today replaces the entire table with skeleton rows for an action fired on one row. It reports on the indicator instead, naming the item: "Applying **Publish** to *My Title*"
- [x] The item title is HTML-escaped before it reaches the indicator label, which is bound with `[innerHTML]`; `actionName` already is, and a content title is user-supplied
- [x] The context-menu workflow action's outcome toast names what happened, on both success and failure. `content-drive.toast.workflow-executed-detail` already exists in `Language.properties:7243` and is never passed; today the toast reads "Workflow Executed" with no action and no item, and the error path (`content-drive.toast.workflow-error`) has no detail either
- [x] The indicator shows determinate progress once the job reports it, and stays indeterminate when it does not
- [ ] ~~The indicator offers cancellation for a cancellable run~~ **Deferred to #33331**, which already owns it: *"From this task manager UI, we can also enable the user to cancel individual file uploads, or cancel the whole upload or move."* The backend still builds the capability, so this is a deferral and not a removal. What this ticket must still do is **not report a cancelled run as a success**: an unrecognised terminal state is reported as an error rather than a green outcome
- [x] `actionExecution` becomes a keyed collection of runs rather than a single slot, so an upload does not block a selection action or the reverse. The existing one-at-a-time guard is scoped per action rather than global
- [ ] The indicator has a defined rule for concurrent runs (one line per run up to two, then a collapsed count)
- [ ] Jest/Spectator specs cover: multi-file success, partial failure, a terminal state the client does not recognise reported as an error rather than a success, that a single-file upload still takes the existing path unchanged, that no operation raises an in-flight toast, that the context-menu workflow action leaves the grid rendered, and that two runs can be in flight at once without blocking each other

### Priority

Medium

### Additional Context

**Out of scope: directory upload.** Dropping a *folder* from the desktop needs `webkitdirectory` relative paths and folder creation ahead of the assets. Worth doing, materially more work, and it should not ride in on this ticket. File a follow-up when this lands.

**Out of scope: resumable or chunked single-file upload.** Large-single-file behavior is unchanged here.

**Now IN scope: per-row busy state.** *(Reversed 2026-09-03.)* Marking the individual rows an operation is acting on, so the grid shows **which** items are in flight rather than only **what** is running.

It was excluded on the reasoning that *"upload has no rows, so this ticket cannot exercise it"*. That reasoning no longer holds: this ticket also converts the **context menu** and the **drag-and-drop move** to the shared indicator, and both act on rows. The feature has two real consumers inside this PR.

It is also cheaper here than anywhere else. `dot-folder-list-view` already swaps per row inside its `#body` template, and `$lockedByOthers` is an existing keyed per-row input in the same file, already keyed by inode. The store now exposes `busyRows` from the run registry.

Additional acceptance criteria:

- [ ] While an operation runs on a row, that row shows it and stays readable and identifiable — **not** replaced by a skeleton, which loses the row the author wants to watch
- [ ] The rest of the grid stays usable; rows the operation is not touching behave normally
- [ ] Busy marks are keyed by **inode**, not identifier: the language filter is multi-select, so one identifier can occupy several rows and marking by identifier would mark siblings nothing is happening to
- [ ] The settling reload does **not** blank the table: it refetches quietly, leaving rows visible, and the busy marks clear once fresh data lands. This keeps the outcome filter-correct — an archived or unpublished row simply is not in the new result — which an optimistic in-place update could never be, since the client cannot know whether the new state still matches an active filter
- [ ] Busy marks clear on the failure path too, or rows stay busy forever
- [ ] `dot-folder-list-view` is given `dataKey="inode"`. It currently takes the `'identifier'` default, so **selecting one language version of a contentlet selects them all**. The codebase already assumes inode — the search service backfills folder inodes with a comment saying the table keys on it — the shell simply never passes it

#37322 was filed for this and is superseded; it should be closed once this lands.

**Out of scope: reindex.** `content-drive.action-center.toast.reindex-started` stays as it is. Its opt-out from the indicator is deliberate and documented, and the notification bell already records the run.

**This ticket owns the core-web job-progress primitive.** Nothing in core-web consumes `/v1/jobs` today, so the first consumer has to build it, and upload is the right owner: #32356 asks for the richest version of it (per-file progress, per-file status, cancel one file or the whole batch), so a primitive that satisfies upload satisfies the folder actions too, while the reverse is not true.

Requirements for it, so #37063 can consume it without changes:

- submit, resolve `jobId`, follow the job, expose progress, expose the terminal state and the result metadata
- follows jobs over `GET /v1/jobs/{jobId}/monitor` with native `EventSource`, falling back to `status` polling. Worth noting: `dot-content-drive-action-center.component.ts:126` records that the legacy `_bulkfire` SSE path was skipped because "native `EventSource` cannot POST a body". That objection does not apply here, since submission is a normal POST returning a `jobId` and monitoring is a separate GET
- may expose cancellation over `POST /v1/jobs/{jobId}/cancel`, but **nothing in this ticket surfaces it**; the control is #33331's. The primitive's API should not be shaped to preclude it
- a running job survives component teardown and navigation, and its outcome stays discoverable
- generic in its API: no upload-specific or Content-Drive-specific shapes

**Depends on #37062** for the job-result contract (`successCount` / `failCount` / `results`), which the folder endpoints define first.

**#37063 consumes the primitive** for folder copy and bulk delete. Whichever ticket lands first builds it; the specification lives here. Do not build it twice.

**Related:** #37062 (job-result contract), #37063 (consumes the job-progress primitive for folder copy and delete), #35436 (parent epic), #32356 (the async-upload feature this implements), #36702 (AssetPicker reusing Content Drive, second consumer of the same upload path).

---

## Amendment — 2026-09-01

Recorded during `/speckit-specify` for the backend half (spec PR #37300, reviewed by
@fabrizzio-dotCMS). These supersede the corresponding statements above; everything not listed
here still stands as originally written.

### 1. The job-result contract: dependency direction reversed

**Above:** "Depends on #37062 for the job-result contract (`successCount` / `failCount` /
`results`), which the folder endpoints define first."

**Now:** this ticket **owns** that contract; #37062 and #37063 consume it. Both are open and
unbuilt, so waiting on #37062 would block this work on a ticket nobody has started.

**And it is not a new contract.** The shape already ships: `BulkRefreshContentletsProcessor#getResultMetadata`
(#36845 / #37131) emits `total` / `processed` / `successCount` / `failedCount` / `skippedCount` /
`results[]`, with an immutable per-item record and a `SUCCESS` / `FAILED` / `SKIPPED` status. This
ticket **generalizes** that shape rather than defining a second one. Three deltas are required:

- **A machine-readable reason code.** The shipped per-item record carries a human-readable message
  only. Failures here must distinguish at least: over size limit, disallowed type, name collision,
  permission denied, staged content unavailable, and unclassified.
- **A generic item key.** The shipped record is keyed by contentlet identifier and inodes. An
  uploading file has neither — it does not exist until the run creates it — and neither does a
  folder path. A generic key is the substance of the generalization.
- **One spelling of the counters.** The shipped counter is `failedCount`; this issue says
  `failCount` above. **`failedCount` wins**, matching what already ships.

> **#37062 / #37063:** consume this shape. Please do not define a third one.

### 2. Submission entry point: a domain endpoint, not the generic job upload

**Above:** proposes reusing `POST /v1/jobs/{queueName}/upload` because it already consumes
multipart.

**Now:** a domain endpoint owning its own multipart handling. Two reasons:

- `JobParams` declares a single `@FormDataParam("file") InputStream`, so the generic entry point
  accepts exactly **one** file today. Carrying a batch through it means changing a shared type
  with other consumers.
- Submission-time validation. A missing target folder, an upload type that is neither `DOTASSET`
  nor `FILEASSET`, or a caller without add-children rights must be refused **before** a job is
  created. The generic entry point has no notion of a target folder; it would enqueue first and
  fail inside the run.

The client sends **one call** carrying the files and the batch parameters. Where the bytes wait
for the run is a server concern and stays out of the frontend contract.

### 3. Per-file size and type limits are inherited, not invented

**Above:** "Per-file size and extension limits are enforced with the same rules as the
single-file path."

**Now:** confirmed, and worth stating precisely because it was nearly specified the other way.
`ESContentletAPIImpl.validateBinary` already enforces a size ceiling (`maxFileLength`) and an
allowed-types rule (`accept`), both declared per content type on the binary field, during the
contentlet validation the creation path runs. This ticket applies **those**, not bulk-specific
ones — a batch that rejected a file the single-file upload accepts would not be equivalent.

Consequence, accepted knowingly: where an operator has configured neither, there is no ceiling.

### 4. New: a batch file-count cap

Nothing caps a selection today. **Default 50 files per batch**, configurable, enforced as a
submission-time refusal. There is deliberately **no batch-total size limit**.

### 5. New: staged content must outlive the queue wait

Not covered by the acceptance criteria above, and it is where this feature can silently lose a
user's files. The submission is answered before the files are created, so the uploaded bytes wait
somewhere in between. The product's staging mechanism expires content on a **global** timer
(`TEMP_RESOURCE_MAX_AGE_SECONDS`, default 1800s) with no per-call override, so a large batch queued
behind other work can outlive its own staged files.

Additional acceptance criteria:

- [ ] Uploaded content stays available to the run until the run reaches a terminal state, however
      long it waits in queue
- [ ] Content that cannot be retrieved is a per-file failure with its own reason — never a silent
      loss, and never reported as though the author supplied a bad file
- [ ] Staged content is reclaimed on any terminal state, including for files the run never reached
- [ ] Staged content is readable by a node other than the one that accepted the submission
- [ ] A stated position on total staged bytes per author across concurrent batches ("none,
      deliberately" is acceptable; silence is not)

### 6. Frontend: cancellation is being deferred, and needs its own ticket

The frontend acceptance criterion "the indicator offers cancellation for a cancellable run" is
being deferred to the task-manager work. The backend still builds the capability, so this is a
deferral, not a removal.

**Resolved: #33331 covers it explicitly**, so no new ticket. The epic's own description already
claims this exact capability: *"From this task manager UI, we can also enable the user to cancel
individual file uploads, or cancel the whole upload or move."* The frontend spec cites it in six
places and states the reasoning; the acceptance criterion above is now marked deferred rather than
left looking like an undelivered promise.

The one thing kept in this ticket is the half that is not about the control: a run reaching a
cancelled state must not be reported to the author as a success. That is covered by the frontend
spec's FR-024, which reports any terminal state the client does not recognise as an error.

So the backend cancel surface does have a dated consumer (#33331), and until it arrives the client
cannot mistake a cancelled run for a completed one.

---

## Amendment — 2026-09-07

Recorded after `/speckit-plan` for the backend half, and after the review of spec PR
[#37358](https://github.com/dotCMS/core/pull/37358) by @wezell. These supersede the corresponding
statements above and in the 2026-09-01 amendment; everything not listed here still stands.

The backend spec, wire contract and data model are in `specs/37166-bulk-file-upload/`.

### 1. Submission shape: one call, confirmed — and it is the shipped precedent

The 2026-09-01 amendment already said "one call". It was reversed to two steps mid-review and then
restored, so this records the outcome rather than leaving the PR history as the only account.

@wezell's objection was that a second upload endpoint duplicates the temp API. Accepted: **the
endpoint hands every part to `TempFileAPI.createTempFile` as it reads, and stores only the
resulting ids in the job parameters.** It does not stage anything of its own.

**Content import is the precedent, not a departure from it.** `ContentImportResource:102` consumes
`MULTIPART_FORM_DATA`, and `ContentImportHelper:394` calls the staging API internally. An
intermediate draft had described import as though the *client* staged; it does not.

Two findings decided it against the client-stages-first alternative:

- **The expiry clock.** `TempFileAPI.getTempFile` gates on `lastModified() + TTL`, per file. If the
  client staged, the earliest files of a large batch could expire before the author reached submit,
  and an unresolvable reference refuses the whole submission. Server-side staging starts that clock
  immediately before the job is created.
- **Bounding.** Nothing below this endpoint caps file count or request size. With the bytes already
  written at submission time, the ceilings would be a policy about what gets *processed* rather
  than a bound on what can be *written*.

**Accepted cost**, recorded so it is not rediscovered: no per-file upload progress and no retry of
a single failed file. A dropped connection means resending the batch.

### 2. Limits — supersedes §3 and §4 of the 2026-09-01 amendment

| | Then | Now |
|---|---|---|
| Files per batch | 50 | **100**, configurable (`CONTENT_BULK_UPLOAD_MAX_FILES`) |
| Batch total size | *"deliberately none"* | **1 GB**, configurable (`CONTENT_BULK_UPLOAD_MAX_TOTAL_BYTES`) |
| Per-file ceiling where the content type declares none | *"no ceiling, accepted knowingly"* | **200 MB fallback**, configurable |

**Why a batch total after all.** Without one the batch is bounded only by the file count multiplied
by a per-file ceiling that is unset by default — which is to say not bounded at all. `TEMP_RESOURCE_MAX_FILE_SIZE`
ships as `-1`, and `dotmarketing-config.properties` says so in a comment: *"authenticated users are
unlimited"*. There is also no container-level multipart limit configured. **This endpoint is the
first and only bound in the path.**

**Why 100 and not more.** The two ceilings are chosen together: at a few megabytes per file, 300
would put a realistic batch past the *total*, making the count cap decorative and the size cap the
one that surprises people. Under the one-call shape the count also governs how long a single
request stays open.

**Why the per-file fallback is a knowing exception.** Where a content type declares no
`maxFileLength`, a file over the fallback is rejected in a batch and accepted as a single upload.
That diverges from the equivalence rule this ticket otherwise holds to. It is accepted because the
alternative — no per-file bound at all — was judged worse, and it is documented wherever the limits
are.

**Three controls here are new, not one**: the file count, the per-file fallback, and the batch total.

### 3. New: two submission-time controls the endpoint does not inherit

Not covered by any acceptance criterion above.

`verifyTempResourceEnabled()` and `SecurityUtils.validateReferer()` live on `TempFileResource`, not
on the `TempFileAPI` this endpoint calls. **No global filter supplies them for API paths** —
`XSSPreventionWebInterceptor` protects a fixed list that does not include `/api/`.

- [ ] The endpoint applies a **same-origin check** on the request
- [ ] The endpoint **honours `TEMP_RESOURCE_ENABLED`** rather than writing staged content behind
      a switch an operator turned off
- [ ] Both are covered by tests

Severity, stated plainly: the classic CSRF path is already closed by `sameSiteCookies=lax` in
`context.xml`, so this is defense in depth rather than a live exploit. It is specified because the
value is operator-configurable and because **content import omits both today** — following that
precedent uncritically would reproduce the gap. Closing it there is tracked separately.

### 4. New: the ceilings are enforced while the request is read

The file count and the accumulated total are checked **as each part is staged**, aborting the read
the moment either is crossed, rather than after the whole body has arrived.

- [ ] A submission over either ceiling cannot commit more than the ceiling to disk
- [ ] Content staged before the abort is reclaimed
- [ ] Content staged before a submission that **dies while being read** — the author navigated away,
      or the connection dropped — is also reclaimed. This is a different code path from a refusal:
      nothing is raised by this side, so a reclaim scoped to the refusal path never runs. Nothing
      purges staged content on a schedule, so what leaks here leaks permanently

A declared `totalSizeBytes` may be sent for a fast refusal before reading. It is a courtesy, never
the enforcement point.

### 5. Index policy, made concrete

**Above:** "Files are not indexed with `WAIT_FOR` per item; the batch indexes once, or the job's
completion drives the reload."

**Now:** each file is created with `IndexPolicy.DEFER`, **and** the run resolves index visibility
once at the end of the batch, *before* the completion signal is emitted. Both halves are needed:

- Per-file `WAIT_FOR` does not merely block on a refresh — `indexContentListWaitFor` also flushes
  the system-wide query cache on every file, so a full batch charges every other user one flush per
  file.
- `DEFER` **alone is insufficient**: it only enqueues into `dist_reindex_journal`. A run reporting
  "finished" would hand the author files a text search cannot yet find. Per
  [ADR-0018](https://github.com/dotCMS/platform-adrs/blob/main/decisions/0018-database-first-content-drive-search-with-index-deferred-text-filtering.md)
  the plain folder listing is database-resolved, so the grid is fine either way — but free-text and
  searchable-field criteria are index-routed, and that ADR names this exact failure as a recurring
  complaint.

**The frontend depends on the ordering**: it refreshes its listing on the completion signal, so
emitting the signal first would surface as a client-side defect for a cause living entirely on the
server.

### 6. Target fields: `folderId` / `siteId`, not `targetFolder`

**Above:** "Multipart: the files, plus `targetFolder` (or site id at the root)".

**Now:** two explicit, mutually exclusive fields. The workflow API this feature fires already
separates `contentHost` (site) from `hostFolder` (folder), and Content Drive currently sends a site
id *in* `hostFolder` at the root — the overloading
[ADR-0020](https://github.com/dotCMS/platform-adrs/blob/main/decisions/0020-deprecate-folder-bypath-endpoint.md)
moves away from. Exactly one of the two must be present.

The multipart shape follows content import: repeated `files` parts plus a JSON `form` part carrying
`baseType`, `folderId` | `siteId` and the optional `totalSizeBytes`. A batch over the total size
answers `413`; the count, shape and origin refusals answer `400`.

### 7. New scope: a resumable run, and a safe resubmission

Not covered by any acceptance criterion above, and **the largest piece of hidden work in this
ticket**.

The job framework persists no mid-run per-item state: `parameters` are immutable after creation,
`progress` is a single float, and `result` is harvested once at the terminal state. Meanwhile
`AbandonedJobDetector` re-queues a stalled run **without consulting the retry policy**, so an
interrupted run is retried whether or not the processor is marked no-retry.

A re-run without resume does not duplicate data — the unique index on the lower-cased path rejects
the second create — but it makes the **report lie**: the files the first attempt created come back
as `NAME_COLLISION`, so the author is told 30 files failed when all 30 are in the folder.

- [ ] The run records completed items durably as it goes, in a new additive `job_item_result` table
- [ ] A re-queued run continues from that record instead of restarting
- [ ] Counts reflect the whole batch across all attempts, and the submitter is notified **once**
- [ ] Resubmitting after a lost or uncertain response never silently creates a second copy, and a
      duplicate resubmission is distinguishable from a batch whose files genuinely all collided

The table is named for **jobs**, not for uploads, on purpose: #37062, #37063 and #37165 all need the
same thing, and four near-identical tables would be the wrong outcome. Whether durable per-item
state belongs in the job framework itself is being proposed as an ADR in `dotCMS/platform-adrs`.

### 8. Where the batch guarantee starts — affects a frontend acceptance criterion

Every guarantee in this ticket — surviving the author leaving, resumability, cancellation, the
durable outcome — is a property of a **run**, and a run does not exist until the submission is
answered with a handle.

The frontend criterion *"Navigating away does not lose the batch"* therefore holds **after** the
`202`, not during the upload. If the author navigates away while the bytes are still being sent,
that request dies and no job is created: nothing is recorded and nobody is notified, because there
is nothing yet to record.

Agreed with @wezell on #37358 and accepted rather than engineered around. The cheap mitigation,
which the frontend half should take: dotAdmin is a single Angular application — legacy portlets are
hosted inside the same shell — so navigating between portlets is a route change and the document
survives. **An upload request owned by a root-level service rather than by the Content Drive
component therefore survives that navigation at no extra cost**, which the job-progress primitive
already requires. A `beforeunload` guard covers a genuine page unload. Only a Service Worker would
survive the document being destroyed, and dotAdmin registers none; that is judged disproportionate.

dario-daza:
Backend for Content Drive bulk file upload — closes #37166.

Today Content Drive lets an author select many files, warns them, and uploads **only the first one**. This adds an endpoint that takes the whole selection, creates and publishes every file in the background, and tells the author what happened to each one.

---

## How it works, in one pass

```
POST /api/v1/assets/_bulkupload      →  202 { entity: { jobId, statusUrl, submitted } }
                                          ↓ (background)
                                        each file created, then published
                                        per-file outcome recorded
                                          ↓
                                        author notified — push + notification bell
```

One `multipart/form-data` call carries **everything**: repeated `files` parts plus a JSON `form` part naming the target and the base type. The server stages each file as it reads, enqueues a job, and answers `202` immediately.

**Why one call and not two** (client stages, then submits references) — this is the design question most worth pushing on, so here is the reasoning up front. Content import already works exactly this way (`ContentImportResource` + `ContentImportHelper`), and two steps would have started the staged-content expiry clock on the *client*, so the earliest files of a large batch could expire before the author finished choosing. It also makes the ceilings a real bound on what reaches disk rather than a policy about what gets processed.

The honest counter-argument is that the two-step shape inherits the staging resource's request controls for free — and it does, but those turned out to be four lines to copy (`verifyStagingEnabled` and `validateReferer` in `BulkUploadResource`), not a structural property worth reshaping the endpoint for. What one call genuinely costs the client is **per-file upload progress** and **retrying a single file**: a dropped connection means resending the batch. That is recorded in the contract as C-001a rather than discovered later.

---

## Reading path

Follow it in execution order — each file is small and does one thing.

**1. Submission** · `rest/api/v1/asset/bulkupload/`

| File | What to look for |
|---|---|
| `BulkUploadResource` | The endpoint. Two controls applied **before** auth — same-origin and the staging kill switch — because neither is inherited from `TempFileAPI` and no filter covers `/api/` |
| `BulkUploadForm` | Validates in its constructor, so a malformed batch is refused before a byte is read. `folderId` XOR `siteId` |
| `BoundedMultipartReader` | Counts files and bytes **while reading** and aborts at the ceiling, reclaiming what it staged |
| `BulkUploadHelper` | Target + permission, the ceilings, and the job parameters |

**2. The run** · `jobs/business/processor/impl/BulkUploadProcessor.java`

The one large file. `process()` reads top-to-bottom: loop the files, reclaim staged content, resolve the search index once, log the terminal state. `createOne()` is where a single file is decided, created and published — `createWith()` then `publish()`, the two steps mirroring `ImportUtil#runWorkflowIfCould` and `#runWorkflowPublishIfCould`.

**3. Telling the author** · `BulkUploadCompletionListener`

Pushed system event plus a durable notification, both best-effort.

**4. Shared outcome shape** · `jobs/business/batch/`

`BatchItemResult` / `BatchItemStatus` / `BatchFailureReason` — bulk refresh reads its own records through this too (`AbstractBulkRefreshItemResult.asBatchItemResult()`), so the product ends with one batch-outcome contract rather than three. #37062 / #37063 adopt it unchanged.

---

## Eight things a reviewer would otherwise have to discover

Each of these is a defect that was found and fixed; they are the non-obvious parts of the diff.

**1. The run has no HTTP request.** `FieldHandlerStrategyFactory` resolves a temp id through `HttpServletRequestThreadLocal`, which is null on a worker thread. So the run retrieves content itself using a fingerprint captured at submission — that is why `requestFingerprint` is a job parameter.

**2. `fireContentWorkflow` with no resolved action creates nothing and returns normally.** It logs and carries on, so every file was recorded `SUCCESS` against an empty folder. The action is now always resolved explicitly, and `recordCreated()` refuses to report a success it cannot point at.

**3. `checkin()` and `publish()` are not the direct operations they look like.** Both re-enter the workflow: `validateWorkflowStateOrRunAsWorkflow` resolves `NEW` and `checkAndRunPublishAsWorkflow` resolves `PUBLISH`, each firing the action instead of doing the thing. Both fallbacks are entered *precisely* when that mapping cannot do the job — so without the `DISABLE_WORKFLOW` guard each call fires the action just rejected, and the file ends saved but not live, or lost outright. Import carries the same guard on the same two lines, commented "needed to avoid recursive call". Cost me two round trips of failing tests to find; the javadoc on both methods says why it is there.

**4. Files route to their real content type.** The type used to be hardcoded from the base type, so a PNG became the generic `dotAsset`. That is not a labelling problem: the resolved type feeds `effectiveCeiling` and `acceptedTypes`, so an instance capping its Image type at 5 MB had that rule bypassed by this endpoint alone. Now routed through the product's own `BaseTypeToContentTypeStrategyResolver`.

**5. Reasons are decided from facts, never from exception text.** The validation layer reports over-size, disallowed type, name collision and folder-filter mismatch through overlapping exception classes that differ only by a *translated* message. Size and type are pre-checked from what staging measured; the folder's own rules are pre-checked against the folder. The unique index remains the authority on a name — the pre-check only makes the message useful.

**6. One over-size file used to refuse the whole submission.** `TempFileAPI` collapses its own ceiling and a dropped connection into the same exception with the same wording. `CeilingBoundedInputStream` counts the bytes here instead, so that file fails alone and the batch continues. No behaviour change on a default install, where the ceiling ships as `-1`.

**7. A null job parameter stalls every queue.** Parameters are held in an `ImmutableMap` and the failure surfaces inside the *shared* `processJobs` loop — one bad job stops content import and bulk refresh too. `BulkUploadResourceIT` pins it.

**8. `dotcms-postman/index.js` had no `workingDir`** — so newman resolved every file-upload path against the wrong directory and sent an **empty part**, in every collection in the repo, not only this one. Fixed here because this feature is the one that noticed. Worth a look from whoever owns the Postman suite: it means existing upload assertions were passing against empty uploads.

---

## Deliberate decisions

| Decision | Why |
|---|---|
| **Files are published, not left as drafts** | A knowing divergence from the single-file endpoint, recorded as FR-006a. An author who drops thirty images expects thirty images, not thirty drafts to publish by hand — and the Content Search drop zone has fired `PUBLISH` per file for years |
| **Created, then published — two steps** | So **both** the content type's `NEW` and `PUBLISH` mappings are honoured. An earlier version fired `PUBLISH` alone: it produced published files and every test passed, but it skipped the `NEW` action entirely, so a customer whose `NEW` action notifies someone or sets a field lost that silently. Same shape as content import |
| **The administrator's mapping is never second-guessed** | An earlier version resolved the action and then discarded any that published, which made this the only upload surface in the product overriding workflow configuration. Where no single mapped action both saves and publishes, the two steps run separately — as `PublishSystemActionApiFireCommandImpl` already does |
| **`DEFER` per file, one index resolution at the end** | A per-file `WAIT_FOR` also flushes the system-wide query cache, charging every other user once per file. But `DEFER` alone would report "finished" on files search cannot find yet |
| **Resubmitting is flagged, not blocked** | The run proceeds and the outcome carries `duplicateSubmission`, so the client can say "already uploaded" instead of "everything failed" |
| **DOTASSET duplicates are accepted — and the author is told so** | A dotAsset's identifier is unique per contentlet, so the unique index can never contend for one. Refusing would make this endpoint stricter than every other path that creates one. The duplicate notification therefore splits by base type: `FILEASSET` says nothing was duplicated, `DOTASSET` says the folder now holds two copies and at `WARNING`, because that author has a decision to make |
| **No resume** | An interrupted run restarts from the first file. Whether the job framework should offer durable per-item state at all is an architectural question rather than this feature's to answer — it is the subject of a proposed ADR that #37062 / #37063 will need too. On `FILEASSET` the unique index still prevents a second copy and only the report is wrong; on `DOTASSET` it duplicates |

---

## Not in this PR

- **The frontend** — @zJaaal's half, built against `specs/37166-bulk-file-upload/contracts/bulk-upload-api.md`
- **Choosing a workflow action** — settled on 2026-09-11: the batch fires `PUBLISH` and nothing in the submission selects an action.
- **Resume** — as above

---

## Tests

**91 tests: 33 unit, 58 integration**, plus a Postman collection.

| Where | Count | Covers |
|---|---|---|
| `BulkUploadProcessorIT` | 24 | Creation, publishing and both workflow mappings, partial failure, ceilings, cancellation, collisions, routing, throughput |
| `BulkUploadNotificationIT` | 13 | Every outcome wording including both duplicate variants, addressing, the pushed payload against the contract's field list |
| `BulkUploadResourceIT` | 12 | The refusal ladder, fingerprinting, the null-parameter regression |
| `BulkUploadReasonTest` | 11 | Every reason, and that none is derived from message text |
| Others | 31 | Bounded read, reclaim, security controls, indexing, concurrency, form validation, status mapping, the shared outcome shape |

```bash
./mvnw install -pl :dotcms-core --am -DskipTests -Ddocker.skip
./mvnw verify -pl :dotcms-integration -Dcoreit.test.skip=false -Dit.test='BulkUpload*IT'
./mvnw verify -pl :dotcms-postman -Dpostman.test.skip=false -Dpostman.collections=BulkUploadResource.postman_collection
```

**Worth knowing when reading the tests**: several defects here were invisible to the suite and found by driving the endpoint by hand — the tests construct a `Job` and call `process()`, skipping the resource, the multipart parse, CDI and the queue. Where a test now covers one of those seams, its javadoc says which defect it exists for.

The two newest tests are worth a look because of *what* they assert: `test_run_firesBothTheNewAndThePublishMapping` reads the created file's **workflow history**, not its publish state. Publish state cannot catch the defect in decision 2 above — the wrong implementation also produced published files.

---

## Risk

**No schema change.** Nothing is added to or altered in the database.

No existing endpoint is modified: the single-file upload path is untouched, and bulk refresh gains only a read-through adapter with its shipped accessors intact.

Three config properties, all with defaults: `CONTENT_BULK_UPLOAD_MAX_FILES` (100), `CONTENT_BULK_UPLOAD_MAX_TOTAL_BYTES` (1 GB), `CONTENT_BULK_UPLOAD_FALLBACK_MAX_FILE_BYTES` (200 MB). The first two are readable by the client through `GET /api/v1/appconfiguration` under `bulkUpload`, so the browser can refuse an over-limit selection before uploading it.

This PR fixes: #37166


wezell (commented):
Guys, do we really need this? Why not use the temp API, which gives us a holding place for files before we convert them to permanent assets, or as a sort of fire-and-forget?  With the temp api, we could layer on hooks that do things with the files before we ingest them, things like virus scan the asset, de-dupe the asset, upload the asset to S3, whatever.   You can just keep dropping things to the tempAPI and in fact you can see this work if you batch select files and drag them onto the content search screen. I'm not quite sure why we need another upload endpoint that is probably not quite as capable. 

https://github.com/user-attachments/assets/b8eea7fa-08df-4038-b706-5d56b6b7881f

dario-daza:
> Guys, do we really need this? Why not use the temp API, which gives us a holding place for files before we convert them to permanent assets, or as a sort of fire-and-forget? With the temp api, we could layer on hooks that do things with the files before we ingest them, things like virus scan the asset, de-dupe the asset, upload the asset to S3, whatever. You can just keep dropping things to the tempAPI and in fact you can see this work if you batch select files and drag them onto the content search screen. I'm not quite sure why we need another upload endpoint that is probably not quite as capable.
> 
>  Multi-file.upload.mp4

Hi @wezell, thanks for the feedback. You're right that this should go through the temp API, we changed the shape to follow that. But we're adding some tracking that the content search flow doesn't have.

What the new endpoint does:

1. It takes the files as form data and hands them straight to the temp API underneath. We're not staging anything ourselves.
2. It creates a job and answers immediately with a handle. The author can close the tab and walk away.

That second part is the whole reason the endpoint exists, and it's the thing content search can't do today.

Content search does the same work, but it's a for loop living in a browser tab firing one PUBLISH workflow action per file. So if the author closes the tab halfway through, the remaining files simply never happen and nobody is told. And when something does fail, they get a count ("3 of 50 failed") but not which three, so the only way to find out is to compare the folder against their desktop by hand.

What we want isn't a different upload. It's the same N ordinary single creates, tracked as one job, so the batch has an identity and something is keeping track of where it got to. That's what buys us:

* authoritative counts
* per-file outcomes that name the file and say why it failed, in a code the UI can turn into real copy
* a result that's still there tomorrow if they walked away
* a run that can be cancelled, and resumed if the server restarts
* and a dotCMS notification when it finishes, the same way the bulk refresh does it

We are based in Content Import. It already takes the file as form data, stages it on temp underneath, and hands the work to a job. We're doing the same thing, just with N files instead of one CSV, and with a per-file record so we can actually say what happened to each one. We'll also put a ceiling on the batch (file count and total size) which is more than exists today.

We'd rather follow the Content Import shape (one call), but if you'd prefer the client staged on `/v1/temp` itself and sent us just the ids, we can do that too, the job is the same either way, only the upload leg changes.

https://github.com/user-attachments/assets/94c78e7c-f546-49de-a241-d7214ddf226f

zJaaal:
@dario-daza I checked the implementation against `contracts/bulk-upload-api.md` field by field from the consumer's side, ahead of building the client against it. Almost all of it lines up exactly — endpoint and multipart parts (`form` + repeated `files`), `baseType`/`folderId`/`siteId`/`totalSizeBytes`, `entity.jobId` + `entity.statusUrl`, the outcome counts including `failedCount`, `results[].key/status/reason/message`, all six `BatchFailureReason` values, and the 400/403/404/413 ladder. Thanks — that made the verification quick, and it already caught a mistake on my side: the client had guessed `DISALLOWED_TYPE` before the contract existed, and `BatchFailureReason.DISALLOWED_FILE_TYPE` is what settled it.

Two contract items I can't find, both of which the client depends on. Neither is a new ask — both are in the agreed document.

## 1. `BULK_UPLOAD_COMPLETED` is not emitted

Contract §4 specifies the pushed completion signal, `Visibility.USER`, payload carrying counts **and** `results`, and says it "Follows `BulkRefreshCompletionListener` (#37131) one-for-one".

`git grep BULK_UPLOAD_COMPLETED` over `dotCMS` and `dotcms-integration` on this branch returns nothing: no `SystemEventType` entry, and no completion listener.

I checked whether completion might already be generic and it isn't — bulk refresh needed a dedicated `BulkRefreshCompletionListener implements EventSubscriber<JobCompletedEvent>` to turn the internal job event into a client-facing system event, and there's no generic client-facing equivalent for a client to subscribe to.

Worth saying that the hard part looks done: `resolveIndexVisibility` (`BulkUploadProcessor.java:98,217`) is exactly the C-006a ordering the client depends on and cannot observe. It's the emit after it that's missing.

**Why it matters to the client**: this is the whole of C-006. Without it there's no push, and the frontend spec's Assumptions rule out compensating with a polling loop, so User Story 4 ("leaving does not lose the batch") has nothing to stand on. It's also what the single post-run refresh fires on.

## 2. `duplicateSubmission` is not on the outcome

Contract §3 lists it in the outcome and in the field table: "`true` when this run was a resubmission of a batch that had already succeeded (FR-040a). Lets the client report 'already uploaded' instead of 'everything failed'."

`BulkUploadProcessor.getResultMetadata` (`:295-300`) emits `total`, `processed`, `successCount`, `failedCount`, `skippedCount` and `results`, and `git grep duplicateSubmission` returns nothing branch-wide.

**Why it matters to the client**: it's the one thing that makes FR-040 safe to *report*. Under the collision branch a retry that actually succeeded comes back as "50 of 50 failed — file already exists", and without a flag the client has to render that as a total failure. That's what FR-040a was added for, and frontend FR-037 and C-002a both cite it.

## Not blocking me

Submission and following (`GET /api/v1/jobs/{jobId}/status`) are all there, so I'm starting on User Story 1 and 2 against what exists. It's User Story 4 and the retry reporting that wait on the two above.

---

*Written by Claude at @zJaaal's request, posted from their account.*

zJaaal:
@dario-daza One small join left, and thanks for the retry work — the fingerprint reasoning (submitter, target, ordered `(fileName, sizeBytes)` pairs using *measured* sizes so a caller cannot forge a match) is exactly right, and choosing the collision branch explicitly made the client side easy to build against.

## `duplicateSubmission` does not reach the channel the client listens on

`BulkUploadProcessor:453` puts it into the job's result metadata:

```java
metadata.put("duplicateSubmission", job.parameters().containsKey("duplicateOfJobId"));
```

But `BulkUploadCompletionListener.payloadFor` copies eight keys into the pushed payload and that is not one of them: `jobId`, `state`, `total`, `processed`, `successCount`, `failedCount`, `skippedCount`, `results`.

So the flag is only reachable by polling `/status`, and the client reports from the pushed event — which is the whole of C-006 and the reason there is no polling loop on this side.

**The consequence is the exact thing the flag was added to prevent.** A retry after a dropped connection collides on every file, so the counts read as total failure, and without the flag the client renders "50 of 50 failed — already in this folder" for a batch that had in fact already uploaded. Which is what FR-040a calls worse than offering no retry at all, because the author then deletes and re-uploads files that were already there.

One line in `payloadFor`, alongside the others.

## Why review did not catch it

Both halves are correct in isolation: you record the flag, I read the payload, and nobody owns the join. Two dotbot reviewers passed both sides clean. Worth noting because it is the shape of gap this split is going to keep producing — the contract lists `duplicateSubmission` under §3's outcome, and §4 says the completion payload carries "the §3 outcome", so the document does imply it. It just is not asserted anywhere.

Might be worth an IT that pins the pushed payload against §3's field list rather than against a hand-written set, so the next field added to the outcome cannot quietly miss the event.

## Everything else lines up

I re-checked after merging: `BatchFailureReason` still matches the client's closed set exactly, six for six, and the outcome keys match what the client reads. The per-file reasons are now rendered to the author, grouped one line per reason and named — so `NAME_COLLISION` from your reason resolver reaches them as "**report.pdf** — already in this folder" rather than a generic failure.

---

*Written by Claude at @zJaaal's request, posted from their account.*

zJaaal:
Frontend side of #37166 here. The Playwright suite ran against a live instance for the first time today (scout env, built from the stacked branches) and turned up one defect in the duplicate-resubmission path, plus a question underneath it.

## A flagged duplicate is created again anyway

Upload two files into a folder as **Asset**, then upload the identical batch again. Result: **four rows, two per file**, while the outcome says "already uploaded, nothing was uploaded twice" — and the durable notification says the same thing (`notification.bulkupload.duplicate`).

So the flag is doing its job and the message is faithful to it; what fails is the assumption underneath. `findSucceededSubmission` sets `duplicateOfJobId`, and the reasoning recorded next to it is:

> The client can then send the author to that run's outcome rather than letting this one collide against the files the first attempt created and report "every file failed"

That holds for **FILEASSET**, where `folderRefusal` does the `fileNameExists` lookup and every file in the repeat batch comes back `NAME_COLLISION`. It does not hold for **DOTASSET**: that same method documents the name check as FILEASSET-only, because a dotAsset's title is derived from its binary. Nothing collides, so the second run creates a complete second set of assets and reports `SUCCESS` — and the flag then relabels that success as "nothing was duplicated".

The path this happens on is the one the UI recommends: the upload dialog badges **Asset** as RECOMMENDED, so it is the default choice rather than an edge case.

Two ways out, and the choice is yours because it decides the copy on both sides:

1. **A recognised duplicate should not create anything.** Then the message is already right on both sides and the run needs to skip creation (or be refused with a pointer to the original job).
2. **Duplicate dotAssets are intended.** Then `notification.bulkupload.duplicate` and the client's own copy both have to stop claiming nothing was duplicated, and I will reword ours to match whatever yours says.

I have deliberately not touched the client copy, since it is only repeating what the flag asserts.

## The related question: should a repeated name create a second dotAsset at all?

Independent of the flag: uploading `report.png` into a folder that already holds one leaves two rows with that name and no refusal. Consistent with the FILEASSET-only lookup, and the spec never says a dotAsset name is unique, so this may be entirely intended. The consequence for the client is worth naming, though: `NAME_COLLISION` becomes **unreachable through the recommended upload type**, so one of the seven reasons only ever renders for File uploads.

## Fingerprint scope, for the record

Reading `submissionFingerprint`, the material is `userId | baseType | targetId | (fileName:sizeBytes per staged file, in staged order)`, matched only against a job in this queue whose state is `SUCCESS`. Which means a repeat is recognised only when the same user sends the same files, at the same sizes, with the same base type, to the same target. Same files into a **different folder** are not a duplicate, which is right.

One sharp edge in that list: it is **order-sensitive**. The same files selected in a different order fingerprint differently, so a genuine retry after a dropped connection can go unrecognised depending on how the chooser or the drop happened to order them. Sorting the parts before hashing would make it order-independent at no cost, if you think that is worth closing.

Happy to test any of this again once CI publishes a new image — the e2e now counts the rows rather than trusting the message, so the duplicate case will stay honest either way.

<sub>Written by Claude Code, posted from @zJaaal's account.</sub>

zJaaal:
One more from testing the frontend against a live instance, and it is a small ask with a disproportionate effect on how the refusal reads.

## The refusal cannot name the limit, so the author retries blindly

A submission over a ceiling is refused well, and the two ceilings stay distinguishable, so the copy can already say *which* wall was hit:

> That batch has more files than one upload allows. Nothing was uploaded, so try again with fewer files.

What it cannot say is **how many fewer**. The client has no way to learn `CONTENT_BULK_UPLOAD_MAX_FILES` or `CONTENT_BULK_UPLOAD_MAX_TOTAL_BYTES`, so "fewer" is the most precise word available. An author with 140 files then retries with 120, then 110, and every attempt uploads nothing — which is the behaviour the developer flagged: vague copy turns one refusal into several.

Naming the number turns it into one retry: *"That batch has 140 files. One upload allows 100."*

## The ask

Expose both ceilings on `/api/v1/appconfiguration`, which the admin client already fetches on startup. `ConfigurationHelper` builds that map from `Config.getIntProperty` calls, so it looks like a few lines next to the existing entries.

That buys two things, and the second matters more than the copy:

1. **The refusal names the number**, so a retry is informed rather than a guess.
2. **The client can refuse before sending the bytes.** Right now a 3 GB batch uploads in full and is then refused, because the courtesy check the contract describes (declared total against the ceiling) needs a ceiling the client can read. With the value available, that check becomes real and the author is stopped in the file chooser rather than after the wait. The frontend spec has this as a MAY and it is currently unbuilt for exactly this reason.

If exposing config is unattractive, the fallback is including the limit in the refusal body, e.g. `{"message": "...", "limit": 100, "limitKind": "FILE_COUNT"}`. That fixes the copy but not the wasted upload, so it is the weaker of the two.

Happy to do the frontend half of either as soon as the value is reachable.

<sub>Written by Claude Code, posted from @zJaaal's account.</sub>

zJaaal:
Following your FR-040b amendment — which I've now built the client half against (`5f07e86eef`) — one string on your side still carries the sentence the amendment retired.

## `notification.bulkupload.duplicate` says the FILEASSET thing to both base types

```properties
notification.bulkupload.duplicate=These files were already uploaded. Nothing was uploaded twice.
```

For `FILEASSET` that is exactly right: the unique index refuses the second writer. For `DOTASSET` it is the opposite of what just happened — FR-040b says the batch runs again and every file is created a second time, and I confirmed it on a live instance before you amended the spec: a two-file batch resubmitted left **four rows**, with that notification displayed next to them.

It is the durable half of the outcome, so it is the copy an author reads *after* walking away, which is exactly the case where they cannot see the folder and check.

## What the client now does, if it helps to mirror it

The base type travels with the job to its outcome, and the toast splits:

- `FILEASSET` → "These **N** files were already uploaded. Nothing was duplicated." (unchanged)
- `DOTASSET` → "This batch was uploaded before, and **N** file(s) have now been uploaded again. This folder holds two copies of each, so delete the ones you do not want."

The second half matters more than the wording: the author has a decision to make, and the old copy told them there was nothing to decide.

You have `job.parameters().get("baseType")` in `BulkUploadCompletionListener`, so picking between two keys there looks like the same shape as the level/message selection already in `notify`.

Worth saying plainly: this was only findable because you amended the spec rather than reinterpreting it. The old FR-042 sentence would have kept both halves of the product wrong and agreeing with each other.

<sub>Written by Claude Code, posted from @zJaaal's account.</sub>

zJaaal:
Two behaviours found while testing the frontend against a QA instance. Both are decisions in your half rather than defects in mine, and the second has a consequence past the obvious one.

## 1. Every file lands as a draft, and the code refuses to let a mapping publish it

`BulkUploadProcessor` looks up the `NEW` system action and then filters it:

```java
final Optional<WorkflowAction> draftAction = mapped
        .filter(WorkflowAction::hasSaveActionlet)
        .filter(action -> !action.hasPublishActionlet());
```

and when nothing matches, falls back to `checkin` with `DISABLE_WORKFLOW` set — deliberately, so a `NEW` mapped to Publish cannot publish anyway. The reasoning recorded there is that publishing on a mapping the author cannot see would be "against the author's intent".

**The developer's position is the opposite**: a bulk upload should land published, the way the other upload surfaces behave. An author who uploads thirty images and finds thirty drafts has to go and publish them by hand, which is worse than the mapping being invisible.

Worth deciding explicitly rather than by default, because both readings are defensible and the code currently encodes one of them quite firmly. If publishing is right, the filter and the `DISABLE_WORKFLOW` fallback are what have to change, not just the action lookup.

## 2. The content type is hardcoded, so nothing routes by media type

```java
final String variable = "FILEASSET".equals(parameters.get("baseType"))
        ? FileAssetAPI.DEFAULT_FILE_ASSET_STRUCTURE_VELOCITY_VAR_NAME
        : "dotAsset";
return APILocator.getContentTypeAPI(user).find(variable);
```

So a PNG uploaded as `DOTASSET` becomes the generic `dotAsset` type, never the instance's own **Image** type. On the QA environment that is exactly what happens: images arrive as dotAssets.

The product already resolves this properly elsewhere. `BaseTypeToContentTypeStrategyResolver` (used by `ESContentletAPIImpl`) picks the *specific* content type from the base type plus the binary's mime type — `DotAssetBaseTypeToContentTypeStrategyImpl` and `FileAssetBaseTypeToContentTypeStrategyImpl` match against each candidate type's `accept` field variable via `BaseTypeMimeTypeMatcher`, and fall back to the system default when nothing matches. That is the same matcher your own `BulkUploadReasonResolver` docblock refers to.

**The consequence worth naming**, because it is not just a label: the resolved type is what feeds your own validation.

```java
final ContentType contentType = contentTypeFor(job.parameters(), user);
...
        effectiveCeiling(contentType),
        acceptedTypes(contentType));
```

`effectiveCeiling` reads that type's `maxFileLength`, and `acceptedTypes` reads its `allowedFileTypes`. With the generic type resolved, a batch is validated against **generic dotAsset's** limits rather than the type the file would actually have become. An instance that caps its Image type at 5 MB and restricts it to `image/*` has both rules bypassed by this endpoint, while every other path that creates an image enforces them.

That is the divergence FR-006 exists to prevent — and the same objection you used to settle the dotAsset duplicates question: the batch endpoint should not be looser or stricter than every other path that creates one of these.

Happy to take whatever falls out on the client side. Neither of these needs a contract change that I can see: the outcome shape is unaffected, and my copy says "uploaded", which stays true either way.

<sub>Written by Claude Code, posted from @zJaaal's account.</sub>

dotCMS/src/main/java/com/dotcms/rest/api/v1/asset/bulkupload/BoundedMultipartReader.java
```diff
@@ -0,0 +1,179 @@
...
+            // Unbounded, which is how the staging layer ships. Nothing is wrapped, so the ordinary
+            // path is exactly what it was.
+            return staging.stage(part.fileName(), part.content());
+        }
+
+        try {
+            return staging.stage(part.fileName(),
+                    new CeilingBoundedInputStream(part.content(), part.fileName(),
+                            perFileCeiling));
+        } catch (final PerFileCeilingExceededException refused) {
```
fabrizzio-dotCMS: **The per-file ceiling catch never matches, so one over-size file still refuses the whole batch.**

`CeilingBoundedInputStream.count()` throws `PerFileCeilingExceededException`, a `RuntimeException`, from `read(byte[], int, int)`. That read happens **inside** `TempFileAPI.createTempFile`, and its `catch (Exception e)` at `TempFileAPI:194-195` rewraps everything it sees in a `DotRuntimeException`. `TempFileBatchStaging.stage` only catches `DotSecurityException`, so what reaches `stageWithinCeiling` is a `DotRuntimeException` **whose cause** is `PerFileCeilingExceededException`.

Java matches `catch` on the thrown type, not on the cause, so this clause is unreachable through the production staging layer. The `DotRuntimeException` escapes `read(...)` — `catch (IOException)` does not match it either — and the submission dies with a 500. That is exactly the behaviour design note 6 says was fixed.

Inert on a default install, where `TEMP_RESOURCE_MAX_FILE_SIZE` ships as `-1` and `perFileCeiling <= 0` short-circuits at line 136. Live the moment an operator sets a ceiling, which is the only configuration in which this code runs at all.

**Why the suite is green.** `FakeBatchStaging.stage` calls `content.readAllBytes()` directly, so the exception propagates raw and the catch matches. The double reproduces the reader's contract but not the wrap that the only real implementation applies. Worth fixing both: unwrap the cause here, and have the double wrap the way `TempFileAPI` does — otherwise the test keeps asserting a path production never takes.

dotCMS/src/main/java/com/dotcms/rest/api/v1/asset/bulkupload/BulkUploadResource.java
```diff
@@ -0,0 +1,199 @@
...
+    public Response bulkUpload(@Context final HttpServletRequest request,
+                               @Context final HttpServletResponse response,
+                               final FormDataMultiPart body)
+            throws DotDataException, DotSecurityException, IOException {
+
+        // Two controls that live on the staging layer's own REST resource and are therefore NOT
+        // inherited by calling TempFileAPI directly. No global filter supplies them either: the
+        // product's referer interceptor protects a fixed list of paths that does not include
+        // /api/. Content import omits both, which is a gap to close here rather than copy.
+        verifyStagingEnabled();
```
fabrizzio-dotCMS, line 142: **These three controls run after the entire body is already on disk, so the javadoc above them overstates what they buy.**

`body` is a JAX-RS entity parameter. Jersey's multipart `MessageBodyReader` deserializes it **before** this method is entered — that is entity binding, not something the order of statements in here can affect. `DotRestApplication:157` registers `MultiPartFeature` with no `MultiPartProperties`, so the default `bufferThreshold` of 4096 bytes applies and every part above it is spooled to `java.io.tmpdir`. I could not find a `maxPostSize` or equivalent configured anywhere in the repo.

So by the time `verifyStagingEnabled()` runs, an unauthenticated or cross-origin caller has already had the full multipart body received and written to disk. The referer check and the staging kill switch refuse the *submission*; they do not refuse the *bytes*.

This also undercuts the claim in `BoundedMultipartReader`'s class javadoc that aborting mid-body means "a submission over a ceiling cannot commit more than the ceiling to disk", and the note in `BulkUploadHelper` that "only then is the body read". Both are true of the staging step and not of the wire-level parse that precedes it. The same javadoc's research note — that nothing below caps a request and the container has no multipart limit — is what makes it matter.

`BulkUploadSecurityIT` cannot catch this: it calls `resource.bulkUpload(...)` in process, which is downstream of the parse.

Two ways out, and either is fine — what is not fine is the current wording. Take the body as `@Context HttpServletRequest` and read it yourself after the checks, so the ordering the javadoc describes is the ordering that happens; or keep the shape, correct both javadocs to say the controls are applied before *staging*, and record the unbounded receive as a residual risk that depends on a request-size limit at the reverse proxy or connector — noting that neither exists today.

dotCMS/src/main/java/com/dotcms/jobs/business/processor/impl/BulkUploadProcessor.java
```diff
@@ -0,0 +1,811 @@
...
+     */
+    private long effectiveCeiling(final ContentType contentType) {
+
+        final long declared = binaryField(contentType)
+                .flatMap(field -> field.fieldVariableValue(BinaryField.MAX_FILE_LENGTH))
+                .map(value -> ConversionUtils.toLongFromByteCountHumanDisplaySize(value, -1L))
+                .orElse(-1L);
+
+        return declared > 0 ? declared : Config.getLongProperty(
+                "CONTENT_BULK_UPLOAD_FALLBACK_MAX_FILE_BYTES", 209715200L);
```
fabrizzio-dotCMS: `Config.getLongProperty` with a constant key and a constant default, read once per file: `effectiveCeiling` is called from `createOne` at line 224, inside the per-file loop, up to `CONTENT_BULK_UPLOAD_MAX_FILES` (100) times per run. Only the per-content-type override varies; the fallback does not. Hoist it into a field via `Lazy.of(...)`, or read it once in `process()` and pass it down.

dotCMS/src/main/java/com/dotcms/rest/api/v1/asset/bulkupload/TempFileBatchStaging.java
```diff
@@ -0,0 +1,79 @@
...
+     * Deletes the staged file outright rather than leaving it to expire. Expiry is enforced only on
+     * retrieval — nothing removes the content on a schedule — so content left here is left for
+     * good, in bytes the author cannot see and no run will ever collect.
+     */
+    @Override
+    public void reclaim(final String tempFileId) {
+        try {
+            final Optional<DotTempFile> tempFile =
+                    APILocator.getTempFileAPI().getTempFile(request, tempFileId);
+            if (tempFile.isPresent() && tempFile.get().file.exists()
```
fabrizzio-dotCMS: `File.delete()` already returns `false` when the file is absent, so the `exists()` call is a pure extra round-trip — and by this class's own javadoc these files live on the shared assets volume, not node-local scratch. It runs once per staged file on every terminal state, and again from `BoundedMultipartReader.reclaimAll` on an aborted submission. Same shape at `BulkUploadProcessor:649`. Dropping `exists()` and relying on the return value alone is equivalent and halves the calls.

fabrizzio-dotCMS (commented):
## Review — bulk file upload backend

**Recommendation: request changes.** One confirmed defect that defeats a stated design goal, plus a security claim in the javadoc that the code does not support. Everything else holds up well: 91 tests, all seven ITs registered in `Junit5Suite1`, `openapi.yaml` regenerated and consistent, Immutables + Jackson for every DTO, `DotConnect` fully parameterized, and the job-framework-thread vs. worker-thread split correctly guarded with `CopyOnWriteArrayList` / `AtomicBoolean`.

Two inline comments carry the substance; two more are minor. The rest of this is what I could and could not confirm.

### Confirmed

**The per-file ceiling never fires.** Design note 6 says an over-size file now fails alone and the batch continues. It does not — see the comment on `BoundedMultipartReader:146`. The unit test passes because the test double does not model what `TempFileAPI` does to the exception, which is the part worth fixing alongside the catch.

**The controls do not run before the body reaches disk.** `BoundedMultipartReader`'s javadoc calls aborting mid-body load-bearing, and the resource's javadoc says two controls are applied before the body is read. Jersey has already deserialized and spooled the whole multipart entity by then — see the comment on `BulkUploadResource:127`. This is not introduced here; it is true of every multipart endpoint under `/api`. But this is the endpoint that accepts N files and the one that claims otherwise in writing.

### Resolved from earlier threads

`BULK_UPLOAD_COMPLETED` is emitted, content types route through `BaseTypeToContentTypeStrategyResolver`, the duplicate notification splits by base type, and `duplicateSubmission` now travels on the pushed payload (`BulkUploadCompletionListener:143`) rather than only through `/status`. All four check out against the code.

The semgrep finding on `BulkUploadHelper:258-260` is a false positive: the statement is static SQL with every runtime value bound through `addParam`.

### Before merge

`dotcms-postman/index.js` gaining `workingDir` is a repo-wide change, not a change to this feature's collection. 36 collections across 9 groups use a `formdata` `src`, and until now every one of them sent an **empty part**. That fix is right, and the consequence is that those assertions are running against real uploads for the first time.

Four of the groups that contain those collections are still pending — Postman Default, Default Split, Category Content and GraphQL — along with Integration MainSuite 1a, 1b and 3a. Green on those four is the actual gate here, not a formality.

### On the design question

For what it is worth on the temp-API thread: the shape now hands every byte to `TempFileAPI.createTempFile` and stages nothing of its own, so the hooks argument is preserved. The part that is genuinely new is the job handle, and content search cannot offer that from a browser tab.

dotCMS/src/main/java/com/dotcms/rest/api/v1/asset/bulkupload/BoundedMultipartReader.java
dario-daza: fixed, now we are validating against the `RuntimeException` and we change the `FakeBatchStaging` to follow the `TempFileAPI` flow

dotCMS/src/main/java/com/dotcms/rest/api/v1/asset/bulkupload/BulkUploadResource.java
dario-daza: Fixed

dotCMS/src/main/java/com/dotcms/jobs/business/processor/impl/BulkUploadProcessor.java
dario-daza: Fixed

dotCMS/src/main/java/com/dotcms/rest/api/v1/asset/bulkupload/TempFileBatchStaging.java
dario-daza: Applied, with one deviation worth flagging.

You're right about the round-trip. But `exists()` isn't only costing a call there
— it's also what the warning means:

- `exists() && !delete()` → "it was there and I couldn't remove it"
- `!delete()` alone → "it isn't gone", which includes "it was never there"

That warning is the only notice anyone gets that staged bytes were left behind on
a volume nothing purges. So instead of dropping `exists()`, I inverted the order:

    if (!binary.delete() && binary.exists()) { ... }

Short-circuit means `exists()` only runs when the delete actually failed. Same
saving on the happy path, warning keeps its meaning. Same shape in
`TempFileBatchStaging`.

Let me know if you want any change.

dotCMS/src/main/java/com/dotcms/rest/api/v1/asset/bulkupload/BoundedMultipartReader.java
```diff
@@ -0,0 +1,204 @@
...
+    /**
+     * Stages one part, turning a crossing of the per-file ceiling into <b>that part's</b> refusal
+     * rather than the whole submission's.
+     * <p>
+     * FR-011 requires a size rejection to be the file's own failure and to leave the batch running,
+     * and that must hold whether the ceiling is the content type's (decided later, by the run, from
+     * the measured size) or the staging layer's (decided here, because the file never finishes
+     * being written). Before this, the second case took the entire submission down with it.
+     */
+    private StagedPart stageWithinCeiling(final UploadPart part) throws IOException {
```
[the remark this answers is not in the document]
dario-daza: Not fixing this, and the reason is that the premise is wrong: it does get collected.

Worth saying where that premise most likely came from, though — our own comments. Several of them claimed nothing purges `tmp_upload` on a schedule and that anything left there leaks permanently. That was simply wrong, and it's corrected throughout now.

`BinaryCleanupJob.cleanUpTmpUploadedFiles()` deletes from `ConfigUtils.getAssetTempPath()` the same `assets/tmp_upload` that `TempFileAPI` writes to. So the partial file is debris, not a permanent leak.

It's slower than it looks, worth being precise about:

- `CLEANUP_TMP_FILES_OLDER_THAN_HOURS` defaults to 3h, but the job's default cron fires only during the midnight hour — so worst case it sits for about a day, not 3 hours.
- (`TEMP_RESOURCE_MAX_AGE_SECONDS`, the 30-minute value that looks like a cleanup, deletes nothing — it's a TTL on retrieval.)

Three things bound the impact:

1. The partial is never larger than the ceiling — we cut the stream exactly there.
2. It's only reachable when an operator sets `TEMP_RESOURCE_MAX_FILE_SIZE > 0`. It ships as `-1`, and then we don't wrap the stream at all, so the path never runs.
3. It isn't specific to this endpoint. `createTempFile` creates the file before writing and neither of its catch blocks cleans up, so *any* caller whose stream dies mid-write leaves the same partial — content import, the temp resource, the asset picker. `BinaryCleanupJob` is the product's existing answer to exactly that debris.

On the suggested fix: it is feasible — `createEmptyTempFile` is public, so we could take the id first, write ourselves, and delete in a `finally`. But that means reimplementing `createTempFile`'s write loop here, and "every byte goes through `TempFileAPI`" is a property this endpoint was deliberately built to keep. Paying that for a bounded, self-collecting, config-gated leak isn't a good trade.

If the partial-file cleanup is worth having, it belongs in `TempFileAPI`'s own catch blocks where it would fix every caller at once. Happy to raise that separately.

fabrizzio-dotCMS (approved):
Approving. All four findings from my earlier review are addressed, and I verified each fix against the code rather than the commit message.

- The per-file ceiling refusal is reachable now: `catch (RuntimeException)` + `ExceptionUtil.causedBy(...)` matches the `DotRuntimeException` that `TempFileAPI:194` actually throws. The part that makes this stick is that both `FakeBatchStaging` doubles model the rewrap — the wrap-blind double is what kept six green tests over dead code.
- `fallbackCeiling` as a `Lazy<Long>` field is safe: `JobQueueManagerAPIImpl` builds a new `@Dependent` processor per job, so "once per run" is literal, not per-JVM.
- `!delete() && exists()` keeps the warning meaning "it is still there" while dropping the stat call on the happy path.
- The javadoc corrections are the right call: the controls refuse the submission, not the bytes, and the residual risk (no request-size limit at the connector or the proxy, true of every multipart endpoint under `/api`) belongs at that layer, not at this endpoint alone.

I also checked the claim in the docs correction, since it is now load-bearing: `BinaryCleanupJob.cleanUpTmpUploadedFiles()` does sweep `ConfigUtils.getAssetTempPath()`, but `BINARY_CLEANUP_JOB_CRON_EXPRESSION` defaults to `0 */3 0 * * ?` — the midnight hour only — so "up to about a day" is accurate. `TEMP_RESOURCE_MAX_AGE_SECONDS` deleting nothing is accurate too.

On the bot comment that keeps coming back about the partial temp file on a ceiling refusal: your answer is right, and there is a second reason it cannot be done here — `createTempFile` throws before returning the id, so there is nothing to reclaim without changing `TempFileAPI`.

Not a blocker on the code, just on the clock: PR Build is still pending on the last push, and the postman groups reached by the `workingDir` fix are the first runs that upload real files. Worth a look at those before merging.

PR #37373, opened 2026-09-03
Truncate long folder names with ellipsis and tooltip in shared folder tree

### LINKED ISSUE #37363: Shared folder tree: truncate long folder names with an ellipsis and a tooltip
## Problem Statement

Split out of #37174 (finding 3). Follow-up on the shared `DotFolderTreeComponent` (`core-web/libs/ui/src/lib/components/dot-folder-tree`) after the folder-tree unification landed in #36733 / PR #36848.

Long folder names render differently in each consumer: the **Site/Folder Field** cuts the name off and lost the tooltip it used to show, while **Content Drive** wraps the name onto multiple lines. There is no shared handling, so every consumer solves it in its own CSS — or not at all.

**Decision taken during refinement:** truncate with an ellipsis and expose the full name via `pTooltip`, in both consumers, owned by the shared component.

### Root cause notes

- The shared default node template has no truncation and no `pTooltip`, so long-name handling is left entirely to each consumer's CSS.

## Steps to Reproduce

1. Create a folder with a very long name (e.g. 60+ characters).
2. View it in the **Site/Folder Field** overlay: the name is cut off with no tooltip showing the full value.
3. View the same folder in the **Content Drive** sidebar: the name wraps onto multiple lines.

## Acceptance Criteria

- [ ] Folder node labels are constrained to a single line and truncated with an ellipsis when they overflow, in both Content Drive and the Site/Folder Field
- [ ] A truncated label exposes the full folder name via `pTooltip`
- [ ] The tooltip is not shown when the label is not truncated (no redundant tooltip on short names)
- [ ] Truncation/tooltip behavior is owned by the shared component so both consumers stay aligned by default

### Regression coverage

- [ ] Unit tests in `dot-folder-tree.component.spec.ts` cover label truncation and tooltip presence/absence
- [ ] Existing folder-tree tests for all consumers still pass; no visual regression in the Browser Selector sidebar, which uses the same shared tree

## dotCMS Version

Latest from `main` (includes PR #36848, merged as `8c725747c0`).

## Severity

## Links

- Split from: https://github.com/dotCMS/core/issues/37174
- Originating issue: https://github.com/dotCMS/core/issues/36733
- QA feedback comment: https://github.com/dotCMS/core/issues/36733#issuecomment-5197164875
- Original PR: https://github.com/dotCMS/core/pull/36848
- Freshdesk ticket: NA

nicobytes:
## Problem
Long folder names looked different everywhere the shared folder tree is used: Content Drive wrapped names onto multiple lines, the Site/Folder Field overlay cut names off with no tooltip, the Browser Selector used its own tooltip heuristic, the Asset Picker truncated with no tooltip, and the Roles panel didn't handle long names at all.

## Solution
Added a new shared `dot-truncated-label` component (`@dotcms/ui`) that clips text to one line with an ellipsis and reveals the full name via a tooltip on hover or keyboard focus, only when the text is actually overflowing (not based on character count). All five tree consumers (Content Drive, Site/Folder Field, Browser Selector, Asset Picker, Roles) now use this single component instead of their own local truncation/tooltip logic, and `dot-folder-tree` forwards row focus to it for keyboard accessibility. A follow-up commit moved clipping ownership fully into `dot-truncated-label` so consumers only place the wrapper around the part of the row that should clip, rather than the tree wrapping projected templates itself.

Includes a spec document and contract describing the new ownership boundary between the shared tree and its consumers.

### Testing
- Updated/added unit specs for `dot-truncated-label`, `dot-folder-tree`, and all five consumers covering truncation, tooltip-on-overflow, tooltip-suppressed-when-fitting, and keyboard focus behavior.

This PR fixes: #37363


fmontes (dismissed):

nicobytes:
https://github.com/user-attachments/assets/17036700-b620-4ac6-9026-ed098fbe05fa

adrianjm-dotCMS (approved):

PR #37374, opened 2026-09-03
Refactor shared assets filter into reusable DotFilterBar components

### LINKED ISSUE #37174: Content Drive toolbar control, shared filter chips and row selection in the list view

nicobytes:
This pull request refactors how the "shared assets" filter is implemented and shared between the Content Drive and AssetPicker features. The filter logic and UI component are now centralized in `@dotcms/ui`, ensuring consistent behavior and removing duplicated code. The changes also introduce a store-agnostic facade for filter operations, making the filter chips reusable across different parts of the application. Additionally, the PR updates tests and documentation to reflect these architectural improvements.

**Shared Assets Filter Refactor and Reuse**

- The `dot-content-drive-shared-assets-filter` component is removed from the Content Drive and replaced with the new, shared `dot-shared-assets-filter` component from `@dotcms/ui`, allowing the same filter UI and logic to be used in both Content Drive and AssetPicker. [[1]](diffhunk://#diff-b5a304407e9714323b0cf797047d13cd7934e8546ba395706173247e947da154L1-L46) [[2]](diffhunk://#diff-c4ef79cc78c8e7bf7e701bf006b4da92a328be775e6202f447fa06dd1426ca9dL79-R79) [[3]](diffhunk://#diff-9694ac0f91d3f2a30337317456ad8484e9e3ba016ec8aa67d0755db986577db6L26-L33) [[4]](diffhunk://#diff-9694ac0f91d3f2a30337317456ad8484e9e3ba016ec8aa67d0755db986577db6L143-R146)

- The constants for the shared assets filter (`SHARED_ASSETS_FILTER_KEY`, `SHARED_ASSETS_ENABLED_VALUE`, `SHARED_ASSETS_DISABLED_VALUE`) are now re-exported from `@dotcms/ui` instead of being defined locally, ensuring a single source of truth for these values.

**Store Facade Introduction and Integration**

- A new `provideContentDriveFilterFacade` provider is introduced and registered alongside the Content Drive store. This facade abstracts filter operations so that filter chips interact with a store-agnostic interface, supporting reuse and encapsulation. [[1]](diffhunk://#diff-52568debba4600b47a2ab04ebd7bb746f891cd976cd9dae84b34378e350899fcR90) [[2]](diffhunk://#diff-52568debba4600b47a2ab04ebd7bb746f891cd976cd9dae84b34378e350899fcL192-R193) [[3]](diffhunk://#diff-52568debba4600b47a2ab04ebd7bb746f891cd976cd9dae84b34378e350899fcL3171-R3172) [[4]](diffhunk://#diff-b3aeb1ff314c2ad9572eda3ad71f3085ee3086f6420083f53825b68cec271a7cR93) [[5]](diffhunk://#diff-b3aeb1ff314c2ad9572eda3ad71f3085ee3086f6420083f53825b68cec271a7cR121-R124)

- The toolbar and shell component tests are updated to provide the filter facade, ensuring that components using the shared filter chips have access to the correct store instance. [[1]](diffhunk://#diff-26fbeab89f4d488ae2c74da028ec32e8f1ea764e884b8c7517d4971f8092ca89R27) [[2]](diffhunk://#diff-26fbeab89f4d488ae2c74da028ec32e8f1ea764e884b8c7517d4971f8092ca89R68-R79) [[3]](diffhunk://#diff-52568debba4600b47a2ab04ebd7bb746f891cd976cd9dae84b34378e350899fcL192-R193) [[4]](diffhunk://#diff-52568debba4600b47a2ab04ebd7bb746f891cd976cd9dae84b34378e350899fcL3171-R3172)

**Testing and Conformance**

- A new test suite (`content-drive-filter-facade.spec.ts`) is added to verify that the Content Drive's filter facade conforms to the shared filter chip interface and encoding expectations, ensuring consistent filter behavior and encoding/decoding logic across surfaces.

**Documentation**

- The migration guide is updated to clarify the behavior of the `status` filter, especially how the `"live"` option affects the status filter UI in the dialog.

This PR fixes: #37174


zJaaal:
## Review notes

Nice refactor. The three seams (`DOT_FILTER_FACADE` / `DOT_FIELD_FILTER_HOST` / `DOT_RELATIONSHIP_PICKER`) are the right shape, the conformance suite shared between both facades is a great idea, and I checked every moved `us.*` util body against `main`: they are faithful moves with only the parameter type widened, which takes a lot of risk off this diff. CI is green.

Findings below, roughly by severity. Numbers 1 and 2 I think are actual regressions; the rest are questions or gaps.

### 1. Content Drive's language chip lost its non-removable-by-default behavior

The old adapter passed `[removable]="$removable()"`, computed as "false when the only selection is `defaultLanguageId`":

```ts
// dot-content-drive-language-field.component.ts (deleted)
protected readonly $removable = computed(() => {
    const selected = this.$selectedLanguageIds();
    return !(selected.length === 1 && selected[0] === this.#store.defaultLanguageId());
});
```

`DotLanguageFilterChipComponent` passes nothing, and `DotLanguageFilterComponent.$removable` defaults to `true`. So on Content Drive the X now always shows, and clicking it calls `store.removeFilter('languageId')`, which re-seeds the default through `withFilterDefaults` and resets paging: a visible control that does nothing except send you back to page 1. The `removable` input's own JSDoc still documents why Content Drive needs it ("removing that selection simply re-seeds the same value, so the X would do nothing visible").

The five tests covering this (`describe('removable')` in `dot-content-drive-language-field.spec.ts`) were deleted with no replacement.

### 2. The workflow-run indicator will likely wrap to its own row

`DotFilterBarComponent` has `host: { class: 'block w-full' }`. In `dot-content-drive-toolbar.component.html` the outer container is `flex w-full flex-wrap`, so `<dot-filter-bar>` is now a full-width flex item and the `ml-auto` action-execution indicator is its sibling rather than a sibling of the chips. It should get pushed onto a new line instead of right-aligning on the filter row. The comment above it still says "Sits at the end of the filter row." Worth a screenshot with a bulk action running.

### 3. `status: 'archived'` now depends on backend behavior nothing here asserts

`archived` is dropped from the picker request entirely and `'archived'` becomes a seeded `status: ['ARCHIVED']`. That rests on `POST /drive/search` returning archived content for `status: ['ARCHIVED']` with `archived` absent (defaulting false). Content Drive already ships this pattern so I expect it holds, but every assertion in this PR is negative (`expect($request()).not.toHaveProperty('archived')`), so nothing proves the positive. Could you confirm against a running instance?

### 4. Every picker now offers an Archived filter, not just `browse` mode

`$allowedStatuses` only bounds the chip when `browse.showWorking === false`, which only `openBrowserModal({status:'live'})` sets. For a File or Image field `browse` is undefined, so all three conditions are offered and an editor can filter to Archived and reference an archived asset. The old pinned `archived: false` made that impossible. Intended? If so it is worth a line in the spec, because it is a behavior change for the four non-browse entry points that FR-014 does not obviously cover.

### 5. `showFolders` diverges between the two surfaces

Content Drive hides folders whenever `status` or `userSearchable` is set, with a long comment explaining folders carry no status. The picker's is just:

```ts
const showFolders = Boolean(browse?.showFolders) && (bookmark?.hasMoreFolders ?? true);
```

So `openBrowserModal({kinds: ['folder', ...]})` plus a Status selection lists folders next to status-filtered content. Same shared chip, two answers.

### 6. Coverage gaps in code that used to be covered

Not asking for tests everywhere, but a few of these were previously tested and are now not:

- `content-drive-relationship-picker.ts`, `content-drive-field-filter-host.ts` and `asset-picker-field-filter-host.ts` have no specs. The relationship dialog's real wiring (header, footer template, `!ref` guard, cancel to `[]`) moved out of the tested component into the untested provider; the field-filter spec now only exercises a token mock.
- The `field.listed` to "Show In List" split moved into the untested host, and its test ("should populate the Show In List fields from the content type listed fields") is gone.
- The status chip lost its whole `describe('with the real store')` block, which covered the riskiest thing in that component: the `linkedSignal` + `signalMethod` write-back not looping, and "should not reset pagination when an unrelated filter changes." It also lost "should select when the LABEL is clicked, not just the checkbox" (the `pointer-events-none` trick) and the aria-label assertion.
- `$relationshipUnavailable` has no test, though it ships a new i18n key.

### 7. `sameValue` is duplicated three times

Identical in `content-drive-filter-facade.ts`, `asset-picker-filter-facade.ts` and `filter-defaults.ts`. It is the O9 guard, i.e. the one thing the contract calls out as easy to get wrong, so it would be better as one export next to `toFilterValues` in `filter-facade.token.ts`.

### 8. Smaller things

- `dot-status-filter.component.ts`: the effect reads `$allowedOptions()` and returns before reading `$selection()`, so `$selection` is not a dependency on the null branch. Harmless as written, but the safer habit is to read both signals up front.
- `toLocalIsoString` brings `date-fns` into `@dotcms/ui`, which means into the legacy Dojo custom-element bundle. Probably fine, just flagging the weight.
- The shared chips all use `content-drive.*` i18n keys, including the bar's own `content-drive.filters.clear-all`. Reads a bit odd now that the AssetPicker renders them.
- `specs/37174-shared-picker-toolbar/data-model.md` is committed. Per our spec-kit commit policy it is spec.md always and `contracts/` when there is a cross-team contract; `data-model.md` should not land.

---

*Review by Claude Code (Opus 5), run and posted by @zJaaal.*

core-web/libs/ui/src/lib/components/dot-filter-bar/chips/dot-content-type-filter-chip/dot-content-type-filter-chip.component.ts
```diff
@@ -0,0 +1,77 @@
...
+ */
+@Component({
+    selector: 'dot-content-type-filter-chip',
+    imports: [DotContentTypeFilterComponent],
+    template: `
+        <dot-content-type-filter
+            [selectedBaseTypes]="$baseTypes()"
+            [selectedContentTypes]="$contentTypes()"
+            [allowedBaseTypes]="$allowedBaseTypes()"
+            (selectionChange)="onSelectionChange($event)" />
```
zJaaal: We should be using an html file

core-web/libs/ui/src/lib/components/dot-filter-bar/chips/dot-language-filter-chip/dot-language-filter-chip.component.ts
```diff
@@ -0,0 +1,38 @@
...
+ */
+@Component({
+    selector: 'dot-language-filter-chip',
+    imports: [DotLanguageFilterComponent],
+    template: `
+        <dot-language-filter
+            [selectedLanguageIds]="$languageIds()"
+            (selectionChange)="onSelectionChange($event)" />
+    `,
+    changeDetection: ChangeDetectionStrategy.OnPush,
```
zJaaal: We need an html file

zJaaal (dismissed):

nicobytes:
Thanks — this was a genuinely useful review. Every point is addressed below; three of them need a decision from you rather than a fix, and I've marked those.

## 1. Locale chip lost its non-removable-by-default behavior — fixed

You were right, and the diagnosis was exact: the X showed always and clicking it re-seeded the same value through `withFilterDefaults`, costing the editor their page and nothing else.

The rule now lives **once**, in the shared chip, with the surface supplying only its own datum:

```ts
// dot-language-filter-chip.component.ts
readonly $defaultLanguageId = input<number | null>(null, { alias: 'defaultLanguageId' });

protected readonly $removable = computed(() => {
    const seeded = this.$defaultLanguageId();
    const selected = this.$languageIds();
    return !(seeded !== null && selected.length === 1 && selected[0] === seeded);
});
```

Content Drive passes `defaultLanguageId()` (`?? null`, since the store leaves it `undefined` until `/api/v2/languages` answers — "nothing is re-seeded" is the right reading while that is in flight). The AssetPicker deliberately passes nothing, and that is correct rather than an omission: it re-seeds only on open and on Clear all, so removing its locale filter genuinely widens the results.

I did not put "the default for key X" into `DotFilterFacade` — it is a property of the surface, not a filter value, and the contract is already reviewed.

Tests: the five deleted cases are back in the chip's spec (plus one for the picker's case), and **the wiring half is pinned separately** in the toolbar's spec — that Content Drive actually hands the value over, and that the X appears/disappears in the DOM. You noted this affordance was already lost once when the component that computed it was deleted; with the two halves guarded separately, deleting either one leaves the other red.

## 2. Workflow-run indicator wrapping — fixed

Confirmed exactly as you described: the bar is `block w-full`, so as a flex item in a `flex-wrap` row it takes the whole line and the `ml-auto` sibling drops below. The comment above it had become false.

I did not change the bar's width (the picker uses it as a block, and breaking that would be worse). Instead the bar grew one named slot:

```html
@if (filters.$hasNonDefaultFilters()) { <p-button ... clear-all /> }
<ng-content select="[dotFilterBarTrailing]" />
```

Content Drive projects the indicator there, which restores the pre-PR DOM order exactly: chips → Clear all → indicator with `ml-auto`. The bar's spec asserts both properties that matter — the projected element is a **sibling of the chips** (or `ml-auto` has nothing to push against) and it lands **after** Clear all.

⚠️ **I verified DOM structure and order, not pixels.** Your screenshot-with-a-bulk-action-running request still stands and I have not done it.

## 3. `status: 'archived'` depending on unasserted backend behavior — resolved from code, no instance needed

The platform documents it, so this is not an open question:

> **Excluding archived content is not a member of this set.** It is the browse query's standing default (`appendExcludeArchivedQuery`), applied unless something explicitly lifts it — `BrowserQuery#showArchived`, an archive-target workflow step, or `ARCHIVED` below. […] **`ARCHIVED` is the only status that lifts it.**
> — `com/dotcms/browser/ContentStatus.java:12-18`

And `BrowserQuery.java:176-179` shows the platform additionally forcing the working join when `ARCHIVED`/`UNPUBLISHED` is selected — which is the same behavior FR-014d bounds client-side. So `status: ['ARCHIVED']` with `archived` absent does return archived content, and the positive case holds by the endpoint's own contract rather than by our assertion.

## 4. Archived reachable from every picker, not just browse — real, and fixed (FR-014f)

You were right that this was an unexamined behavior change. It is worth being precise about the blast radius, though: **neither shipped VTL template was ever exposed.** Both pass `status: "live"` → `showWorking: false` → FR-014d already narrowed their Status control to `LOCKED`. What was actually reachable were the four **field** entry points (File, Image, video, audio), where `browse` is absent.

Fixed as a second bound, from one rule:

```ts
// store/filter-defaults.ts — the one definition
export function allowedStatusesFor(config: DotAssetPickerConfig | null) {
    if (!config) return null;
    if (!config.browse) return REFERENCEABLE_STATUSES;        // [UNPUBLISHED, LOCKED]
    return config.browse.showWorking === false ? PUBLISHED_ONLY_STATUSES : null;
}
```

Unpublished stays — referencing an asset that is not published yet is ordinary editing. Archived does not, because it is not served at all. `openBrowserModal` stays unbounded, since browsing to find archived content is what it is for.

Applied in **two** places (the chip's options and the caller's seeds), because a seed arrives before any chip exists. Writing the test for that found a hole in my own reasoning: `buildAssetPickerConfig` did `...(isBrowse && browse ? { browse } : {})`, so a `mode: 'browse'` caller passing no options fell into the field bound. Both real browse callers pass `browse`, so production was fine, but the invariant I had documented was not true — it is now (`browse: browse ?? {}`, inert: every read is an optional flag and `{}` resolves the seeded sort to the same default).

🔷 **Needs your call:** this is recorded as **FR-014f** in `spec.md` (and the bound tables in `data-model.md`), explicitly as a bound the spec did not originally examine. It is restorative rather than expansive, but it *is* an edit to an approved spec — happy to take it back through re-approval if you'd rather.

## 5. `showFolders` divergence — dissolved by your own #37367

Worth flagging that this was live, not theoretical: `file_browser_field_render_new.vtl` shipped `kinds: ["file", "page", "folder"]`, so with a Status selection the list came back with archived content *plus every folder*.

Your #37367 does not adjust the picker's rule, it withdraws the capability — `showFolders: false` hardcoded, no folder cursor, `'folder'` out of `DotBrowserItemKind`, and the VTL stopped asking. So there is no divergence left to reconcile: Content Drive still hides folders under filters, and the picker has no folder answer to give. I wrote no code for this.

`main` is merged into this branch, so that change is here. One note from resolving it: the merge's `with-asset-browse.feature.ts` conflict had `includeSystemHost: true` on your side (pre-US1) — taking that half would have compiled cleanly and silently reverted the Shared Assets chip to a no-op, which is the original issue. Kept ours.

## 6. Coverage gaps — fixed (3 new specs)

- **`content-drive-relationship-picker.spec.ts`** (7 tests) — the dialog it opens and with what geometry, the filter footer (Apply enabled at zero selections), `Author.blogs → Author`, `selectionMode: 'single'`, cancel → `[]`, completion after the first close, and the `!ref` guard. You were right that I moved risk out of a tested component into an untested provider.
- **`content-drive-field-filter-host.spec.ts`** (6 tests) — including the `field.listed` split. My replacement test only proved the menu *hands over* the raw list, not that Content Drive *splits it correctly*.
- **Status chip: 10 → 19 tests.** Chip render, all three options, the aria-label, restored selection, the LABEL click (the `pointer-events-none` trick), and the write-back cycle.
- **`dot-field-filter.no-picker.spec.ts`** (5 tests) — `$relationshipUnavailable` and its i18n key, in its own file because component providers cannot be un-provided once the TestBed is instantiated.

The cycle block could not be ported literally — it used Content Drive's real store, and a shared chip cannot import a portlet's. It uses a **stateful** facade instead (signal-backed, answers with what was written), which is the whole feedback path the component owns; what the surface does afterwards is covered by each facade's conformance suite. I also reframed it to assert **termination** rather than a write count.

**That rewrite found something.** It failed at `Expected: 1, Received: 2`: the sync runs once on mount with an empty selection and calls `removeFilter` for a key that is not set. It is inert on both surfaces (Content Drive guards on `if (removedFilter)`, the picker on `if (!(filter in filters))`), so not a bug — but it is now pinned, because if either guard goes, mounting the chip resets the editor's page on every load, invisibly.

`asset-picker-field-filter-host.ts` still has no spec, now **documented as deliberate**: five arrows pointing where the file says they point. If it ever gains a decision, it earns a spec with it.

## 7. `sameValue` duplicated three times — fixed

And I nearly did it again: fixing #4 started with the bound copied into both the toolbar and the seeding. Consolidated into `allowedStatusesFor` before going further, with the reason recorded — a divergence between two copies would be *unobservable* (a seeded value the chip refuses to show, or an option the seeding would have dropped).

## 8. Smaller things

**(a) The Status effect** — went further than reading both signals up front: the `effect` is **gone**. The bound belongs to the derivation, so `$selection`'s `linkedSignal` applies it while reading the bag, and the `signalMethod` that already existed persists the narrowed value. No window in which the un-narrowed value exists, no `untracked`, and the component is down to one piece of reactive wiring.

**(b) `date-fns`** — removed from `@dotcms/ui`. It was the only import in the whole library and it was one `format()` call; `toLocalIsoString` is now local getters plus `padStart`, which is what made it wall-clock in the first place. Its three existing tests pass unchanged.

**(c) `content-drive.*` keys on shared chips** — agreed it reads oddly, and it is deliberate. Now documented on the bar itself: the keys moved verbatim, renaming them means editing `Language.properties` and every shipped translation for no user-visible gain, and the risk is orphaning a key in a language nobody notices until a customer sees the raw id. Frozen by the contract for that reason (`contracts/filter-facade.contract.md` §6, research R8).

**(d) `data-model.md` committed** — 🔷 **here I think the premise is off, and it's your call.** The repo's own rule is a judgement test, not a blanket ban:

> **`data-model.md` commit-worthiness bar:** commit it only if a future dev would need it to know the shapes without reading the code. […] If a feature's `data-model.md` just restates entities already obvious from `spec.md`, it's as ephemeral as the rest.
> — `.specify/CUSTOMIZATIONS.md:53`

And `.gitignore:227` says the same: *"spec.md (and data-model.md / contracts/ when they carry verified contracts) stay tracked."*

This one does not restate `spec.md`: it carries which shape each key takes **per surface** (`baseType` as names in the picker, as numeric keys in Content Drive), the per-surface defaults, and the before→after mapping of the request fields. A dev asking "what shape does `baseType` have on each side?" would otherwise reconstruct it by reading two facades. I removed it and put it back for that reason. Costs the same either way — say the word and it goes.

## Also from the inline threads

Both chips have `.html` files now. I also dropped `changeDetection: ChangeDetectionStrategy.OnPush` from the three components this PR *adds*, per `docs/frontend/ANGULAR_STANDARDS.md:45` ("New components: do NOT set `changeDetection`") — Angular 22 resolves `onPush: componentDefinition.changeDetection !== ChangeDetectionStrategy.Eager`, so omitting it *is* OnPush. Left it on the moved files: the line came with the move, and churning them would make your "faithful move" check harder to redo.

## Verification

`ui` 121 suites / 1922 tests · `portlets-content-drive` 30 / 1344 · `edit-content-bridge` 2 / 130 · `edit-content` 116 / 2342. Lint clean on all four; `tsc --noEmit` at the pre-existing baseline (0 ui, 0 bridge, 35 portlet).

Still open: the screenshot for #2, and your call on FR-014f re-approval and 8d.

core-web/libs/ui/src/lib/components/dot-asset-picker/store/asset-picker-filter-facade.ts
```diff
@@ -0,0 +1,80 @@
...
+import {
+    DOT_FILTER_FACADE,
+    DotFilterFacade,
+    DotFilterValue
+} from '../../dot-filter-bar/filter-facade.token';
+
+type AssetPickerStore = InstanceType<typeof DotAssetPickerStore>;
+
+/** Whether two filter values are the same selection, order included. */
+const sameValue = (a?: DotFilterValue, b?: DotFilterValue): boolean => {
```
adrianjm-dotCMS: `sameValue` is copy-pasted three times: here, in `content-drive-filter-facade.ts`, and in `asset-picker/store/filter-defaults.ts`. The O9 idempotence guard (`Object.entries(patch).some(...)` + early return) is also duplicated verbatim in both facades.

Worth exporting one `sameFilterValue()` from `filter-facade.token.ts`, next to `toFilterValues` — a divergence between the two facades' comparators is exactly the drift this seam exists to prevent.

core-web/libs/ui/src/lib/components/dot-filter-bar/chips/dot-status-filter/dot-status-filter.component.ts
```diff
@@ -50,38 +44,72 @@ import { DotContentDriveStore } from '../../../../store/dot-content-drive.store'
...
-    readonly #store = inject(DotContentDriveStore);
+export class DotStatusFilterComponent {
+    readonly #filters = inject(DOT_FILTER_FACADE);
     readonly #dotMessageService = inject(DotMessageService);
 
     protected readonly listboxPt = CHIP_FILTER_LISTBOX_PT;
     protected readonly popoverPt = CHIP_FILTER_POPOVER_PT;
-    protected readonly LISTBOX_SCROLL_HEIGHT = PANEL_SCROLL_HEIGHT;
-    protected readonly options = STATUS_FILTER_OPTIONS;
+    protected readonly LISTBOX_SCROLL_HEIGHT = '25rem';
```
adrianjm-dotCMS: Magic literal where a constant already exists: line 15 already imports from `../../../dot-chip-filter/constants`, which exports `CHIP_FILTER_SCROLL_HEIGHT = '25rem'`.

With this line there are now four copies of that value (`CHIP_FILTER_SCROLL_HEIGHT`, `FIELD_FILTER_PANEL_SCROLL_HEIGHT`, content-drive's `PANEL_SCROLL_HEIGHT`, and this literal).

adrianjm-dotCMS (dismissed):

core-web/libs/ui/src/lib/components/dot-asset-picker/store/asset-picker-filter-facade.ts
nicobytes: Agreed on both halves of the duplication. Fixed in c544353.

`filter-facade.token.ts` now exports two things next to `toFilterValues`:

- `sameFilterValue(a, b)` — the comparator, single definition. All three copies (`content-drive-filter-facade`, `asset-picker-filter-facade`, `filter-defaults`) import it.
- `isNoOpFilterPatch(patch, read)` — the O9 guard itself, since it was the other verbatim duplication. Both facades now call it instead of re-implementing the `some(...)` + early return.

Content Drive still encodes to its numeric keys **before** the check, so a name→number round-trip keeps reading as unchanged.

core-web/libs/ui/src/lib/components/dot-filter-bar/chips/dot-status-filter/dot-status-filter.component.ts
nicobytes: Fixed in c544353 — the literal now uses `CHIP_FILTER_SCROLL_HEIGHT`, which was already in scope from the line 15 import.

Took the "four copies" point too: the other two now alias the constant rather than repeat the value — `FIELD_FILTER_PANEL_SCROLL_HEIGHT = CHIP_FILTER_SCROLL_HEIGHT` and content-drive's `PANEL_SCROLL_HEIGHT = CHIP_FILTER_SCROLL_HEIGHT`. `25rem` is now defined once, in `dot-chip-filter/constants.ts`.

adrianjm-dotCMS (approved):

PR #37375, opened 2026-09-03
docs(block-editor): add issue-resolution spec for #37340

### LINKED ISSUE #37340: Block Editor splits a link into two anchors when ©, ®, or ™ appears inside the linked text
## Problem Statement

---

## Root Cause (verified)

|---|---|---|---|
| `™` | U+2122 | ✅ | `tm` |

### Scope

---

## Steps to Reproduce

**In the product:**

```
```

```json
[
]
```

**Actual rendered HTML:**

```html
<p>
</p>
```

**Expected rendered HTML:**

```html
```

```json
[
]
```

---

### Editor — root cause

### Regression coverage

---

## dotCMS Version

---

## Severity

---

## Links

rjvelazco:
## What this PR is

**PR 1 of 2** — the spec only, per [Spec-Kit Quick Start §3](../blob/main/docs/core/SPEC_KIT_QUICK_START.md). No code.

Fixes #37340 (spec; the implementation lands in PR 2).

Spec: `specs/37340-emoji-text-node/spec.md`

---

## What you're approving

**The problem.** The Block Editor's emoji extension replaces any Unicode-emoji character in text with a standalone `emoji` node. The node is created bare, so it carries none of the surrounding formatting. Two persisted consequences: a single linked phrase is torn into two `<a>` elements (WCAG 2.2 Level A failures — 1.3.1, 2.4.4, 4.1.2), and the character itself is dropped from VTL-rendered pages because the stored node holds only a shortcode, no literal character.

**The approach** — an emoji typed or pasted into paragraph text stays part of that text node. The automatic character-to-node conversion is suppressed; the `emoji` node stays **registered** so content already saved with `emoji` nodes still parses and renders.

Chosen over a per-character allowlist because it closes the whole class in one change, needs no list maintained across extension upgrades, and removes the marks problem structurally — text nodes carry marks natively, so there is no bare node left to strip them. It also matches how insertion already works: the emoji picker calls `insertContent(emoji.native)` with a raw character, so it never needed the node.

---

## Four findings that changed the scope

Each was verified against a real TipTap editor in jsdom, not inferred from the code.

| Finding | Consequence for the fix |
| --- | --- |
| The defect spans **219 text-presentation characters**, not the 3 reported. 1907 of 1949 catalogued entries convert on typing. | `✔` in a linked list item splits that link exactly as `©` does. The fix targets the class. |
| **Mark inheritance needs no schema change.** Applying a link over an existing `emoji` node already yields one `<a>`, and the mark survives a round-trip. | The split is an *ordering* defect: the conversion rebuilds the node without marks after the mark is applied. Simpler fix than #37340 assumes. |
| **Narrowing the extension's `emojis` option is not viable.** That option also feeds rendering. | Filtering it makes already-stored nodes render as literal `:copyright:`. Verified. The option must stay unfiltered. |
| **The Angular SDK ships two Block Editor renderers** that #37340 never mentions. | Gap A and Gap B live in **five** renderer implementations, not three. `grep` for an `emoji` node type across `libs/sdk/` returns zero hits. |

---

## Decisions recorded in the spec

- **Legacy data: render-only.** No data migration, no heal-on-load. Renderer coalescing satisfies AC-015 without a re-save, keeping this out of Java, the DB, and rollback-relevant territory. Accepted trade-off: old `emoji` nodes persist.
- **`:)` emoticons insert the literal character** rather than creating a node. AC-006.
- **AC-002 uses a representative character sample**, not all 1907 — stable across extension upgrades.

---

## Biggest risk, flagged for planning

`VM_global_library.vm` renders **every** Story Block field on **every** VTL-rendered page, for all customers, feature flag or not. Gap B changes how text runs are grouped there. A defect introduced in coalescing would affect far more content than this bug does. Five parallel Gap B implementations must also stay behaviourally identical, or the same content renders differently per framework — which is why the spec requires one shared fixture driving all five test suites.

---

## Two things to settle before/with approval

1. **An AC on #37340 contradicts this spec.** The issue requires pictographic emoji to keep producing `emoji` nodes; this spec deliberately stops producing them for all newly authored content. The issue AC needs amending so the record doesn't self-contradict.
2. **Triage precondition still unconfirmed** — that the customer on helpdesk ticket 39197 authored the content in the **new** Block Editor. Only the new editor registers the extension, so only it can create this split.

---

## Not in this PR

`plan.md`, `tasks.md`, and `checklists/` are process artifacts and are gitignored. `/speckit-plan` runs only after this PR is **approved** — approval is the gate, not merge.

This PR fixes: #37340

### DIFF
--- a/specs/37340-emoji-text-node/spec.md
+++ b/specs/37340-emoji-text-node/spec.md
@@ -87,7 +87,9 @@ Decisions taken during `/speckit-clarify`. Earlier decisions from the specify se
   (3) `:name:`. Every renderer also emits a warning, in the JS SDKs via `console.warn` and in
   Java via `Logger.warn` (Constitution Principle II), with the message
   `[dotCMS Block Editor]: Emoji <name> is not supported`. Warned once per distinct name per
-  render pass so a page with repeats does not flood the console or the log.
+  render scope so a page with repeats does not flood the console or the log. The scope is
+  defined per surface (one render call in the JS SDKs; one `StoryBlockRenderHelper` instance —
+  i.e. one Story Block field render — in Java), so no request-scoped state is required.
 - Q: The two linked text nodes in the reported data are not adjacent — an unmarked `emoji` node
   sits between them. Should coalescing absorb it? → A: **Option A, narrowed to `emoji` nodes
   only.** A run of `text(link) + unmarked node + text(link)` collapses into a single `<a>` only
@@ -390,6 +392,16 @@ function.
   - The generated map is a **new drift surface**: if it falls behind the extension's `emojis`
     list, renderers resolve a `name` the editor can produce to nothing. The CI equality check in
     AC-012 is what prevents that, and is why generation is required over a hand-written table.
+  - **The cross-build handoff is the riskiest unproven mechanic in this change, and the plan
+    must prove it rather than assume it.** The map is generated from `@tiptap/extension-emoji`,
+    a `core-web` npm dependency, but consumed by `StoryBlockRenderHelper` on the Java
+    classpath. The `openapi.yaml` precedent is **not** a like-for-like comparison: that artifact
+    is generated *inside* the Maven build, whereas this one originates in the frontend
+    toolchain and must cross into Java resources. The plan must answer, concretely: which build
+    generates it, where the artifact is committed, how Java loads it, and how the CI diff-check
+    runs against a JS-produced file without requiring the frontend toolchain in the Java build.
+    If that handoff proves awkward, generating the table inside the Maven build from a
+    committed source of truth is the fallback to evaluate.
   - The map adds weight to three **published** npm packages (`@dotcms/react`, `@dotcms/vue`,
     `@dotcms/angular`). It must be tree-shakeable and must not land in `@dotcms/types`.
   - **Gap B is a consumer-visible output change** in those same published packages: content a
@@ -454,7 +466,14 @@ function.
   node's `text` when it carries one and otherwise the literal `:name:` — **never** empty
   output — and emits `[dotCMS Block Editor]: Emoji <name> is not supported` as a warning
   (`console.warn` in the JS SDKs, `Logger.warn` in Java). The warning fires once per distinct
-  unresolved name per render pass, not once per occurrence.
+  unresolved name per **render scope**, not once per occurrence. The render scope is defined
+  per surface, because "render pass" has no common meaning across them:
+  - **JS SDKs** — one render call of the block-editor renderer component.
+  - **Java / VTL** — one `StoryBlockRenderHelper` instance, i.e. one Story Block **field**
+    render including its nested blocks. `GenericRenderableImpl` creates at most one helper per
+    `toHtml` call and shares it across the recursion, so this scope needs no request-scoped
+    state. A page rendering N Story Block fields may therefore warn up to N times per distinct
+    name — bounded, and preferable to threading request state through the render path.
 
 **Renderers — Gap B (link run coalescing)**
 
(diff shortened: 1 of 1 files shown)


rjvelazco:
## Spec updated — `/speckit-clarify` session (5 decisions)

_Posted by Claude (Claude Code), on behalf of @rjvelazco._

Pushed `20ad1b03c1`. Two of the five clarifications corrected **material gaps in the spec as originally written** — worth a look before approving, since both change what gets built.

### 1. AC-009 was not implementable as written 🔴

The spec said the VTL renderer should output the emoji character. There was no way to do that:

- `grep -i emoji` across `dotCMS/src/main/java/` returns **nothing**
- no shortcode table in any SDK, and the SDKs carry no tiptap dependency
- the stored node holds only `name: "copyright"` — no character, no codepoint

**Decision:** generate the `name` → character map at build time from `@tiptap/extension-emoji`'s own list, so it cannot drift. SDKs consume it via a small runtime module (**not** `@dotcms/types`, which is types-only); VTL consumes it through `StoryBlockRenderHelper`.

Good news on Legacy Impact: that helper already exists in modern `com.dotcms.*`, is already in the Velocity context as `$dotStoryBlockRenderHelper`, and is already called `.render($element)`. So this extends an established hook rather than adding a Velocity tool. The spec's "no Java changes" claim was wrong and is corrected.

### 2. Gap B could not fix the reported payload 🔴

Gap B said "coalesce **adjacent** text nodes sharing an identical `link` mark." In the customer's data they are **not adjacent** — the unmarked emoji node sits between them:

```
text(link) "dotCMS Copyright "  +  emoji(NO marks)  +  text(link) "All rights reserved"
```

Two shapes exist in stored content, both confirmed against a real editor:

| Shape | How it was authored | Emoji node marks | Today |
| --- | --- | --- | --- |
| 1 | emoji typed **into** linked text | none | 2 anchors — **the reported case** |
| 2 | link applied **over** an existing emoji node | carries `link` | already 1 anchor |

Plain adjacency merging only ever fixed Shape 2. **Decision:** a link run continues across an intervening unmarked **`emoji`** node, which is absorbed into the anchor — and only an `emoji` node, so a `hardBreak` or any future atom is never silently swallowed into a link. New AC-015 (Shape 1) and AC-016 (the negative case).

### The other three

- **VTL coalescing lives in Java**, not the Velocity macro — `StoryBlockRenderHelper` pre-groups before the macro renders. This materially lowers the spec's top blast-radius risk: the equality logic becomes JUnit-testable and the interpreted `.vm` edit stays additive.
- **Coordinated minor bump** for `@dotcms/react`, `@dotcms/vue`, `@dotcms/angular`, each with a changelog entry stating explicitly that rendered HTML changes for existing content. Not opt-in — the prior output is a WCAG 2.2 Level A failure.
- **Unresolvable emoji names** resolve map → node `text` → `:name:`, never empty, with `[dotCMS Block Editor]: Emoji <name> is not supported` warned once per distinct name. Reachable because no Story Block schema validation exists, so arbitrary JSON can arrive via REST.

### Net effect

ACs grew **17 → 22**. Scope grew by a generated map artifact plus a CI equality check, and one Java helper change in a modern package. Everything else was already in scope.

Still open from the original spec, unchanged by this session:

- [ ] Amend the AC on #37340 requiring pictographic emoji to keep producing `emoji` nodes — this spec deliberately stops that
- [ ] Confirm from helpdesk ticket 39197 that the customer authored in the **new** Block Editor

adrianjm-dotCMS (dismissed):

rjvelazco:
## Review findings addressed — `871a489fc3`

_Posted by Claude (Claude Code), on behalf of @rjvelazco._

All three Medium findings verified against the code and applied. Each was correct.

### 1. JUnit coverage omitted the absorption rules ✅

Confirmed. The bullet read "Covers AC-014, AC-017 and the no-op path" — the reference had been carried forward mechanically through two rounds of AC renumbering without revisiting what it should cover. AC-015 and AC-016 are precisely what the Java grouping exists to implement, so asserting them only at the Postman layer contradicted the Q2 rationale.

Now covers **AC-014 through AC-017**, calling out AC-015 (emoji absorbed) and AC-016 (non-emoji atom breaks the run) explicitly, plus the no-op path. AC-018 (nested marks) is deliberately left to the VTL end-to-end layer, since it depends on macro rendering rather than grouping.

### 2. "Render pass" was undefined for Java ✅

Confirmed, and the code gave a clean answer. `GenericRenderableImpl.java:191-195` creates **at most one** `StoryBlockRenderHelper` per `toHtml` call and shares it across the whole recursion:

```java
// baseTemplatePath is constant for the whole tree, so a single helper is shared across
// the recursion - only create one if the context does not already carry it.
if (null == context.get(StoryBlockRenderHelper.CONTEXT_KEY)) {
    context.put(StoryBlockRenderHelper.CONTEXT_KEY, new StoryBlockRenderHelper(baseTemplatePath, context));
}
```

So the scope is now defined per surface:

| Surface | Render scope |
| --- | --- |
| JS SDKs | one render call of the block-editor renderer component |
| Java / VTL | one `StoryBlockRenderHelper` instance — one Story Block **field** render, nested blocks included |

A page with N Story Block fields may warn up to N times per distinct name. Bounded, and it avoids threading request-scoped state through the render path.

### 3. Cross-build handoff was asserted, not sketched ✅

The sharpest of the three. You're right that the `openapi.yaml` comparison was a bad analogy — that artifact is generated *inside* the Maven build, whereas this one originates in the frontend toolchain and has to reach the Java classpath.

Now flagged explicitly as **the riskiest unproven mechanic, which the plan must prove rather than assume**, with the concrete questions named: which build generates it, where it is committed, how Java loads it, and how the CI diff-check runs against a JS-produced file without pulling the frontend toolchain into the Java build. Added a fallback to evaluate if that handoff proves awkward — generating the table inside the Maven build from a committed source of truth.

### Observations

Agreed on all three, no changes needed. The two contradictions stay open as real approval gates — in particular the #37340 AC amendment should land before this spec is approved, as the spec itself states.

adrianjm-dotCMS (approved):

PR #37381, opened 2026-09-03
fix(cicd): refuse to re-release a version that already has a GitHub Release

### LINKED ISSUE #37380: Release pipeline silently clobbers the release branch when a published version is re-dispatched
## What happened

Release `26.09.02-01` shipped successfully on 2026-09-02 ([run 33659151567](https://github.com/dotCMS/core/actions/runs/33659151567)) — tag `v26.09.02-01` @ `7fe2ff5`, GitHub release published, `dotcms/dotcms:26.09.02-01` pushed to Docker Hub.

On 2026-09-03 the **same version string** was dispatched again ([run 33780197124](https://github.com/dotCMS/core/actions/runs/33780197124)), this time against a `main` that was 54 commits ahead. The run got 24 minutes in — through Prepare and a full 15-minute artifact build — before being cancelled by hand at the Docker deploy step.

By then Prepare had already force-recreated `release-26.09.02-01` at `81563d9`, a commit 54 commits newer than the tag the branch is supposed to match. Had the run not been cancelled, it would have published Docker and Maven artifacts for `26.09.02-01` containing entirely different code than what already shipped under that version.

## Root cause

`.github/workflows/cicd_comp_release-prepare-phase.yml` has two "already exists" guards, and they don't agree:

```yaml
# line 148 — tag guard, checked against the LOCAL clone
if git rev-parse "${release_tag}" >/dev/null 2>&1; then
  echo "Tag ${release_tag} exists, removing it"
  git push origin :refs/tags/${release_tag}
fi

# line 165 — branch guard, checked against the REMOTE
remote=$(git ls-remote --heads https://github.com/dotCMS/core.git "${release_branch}" | wc -l | tr -d '[:space:]')
if [[ "${remote}" == '1' ]]; then
  echo "Release branch ${release_branch} already exists, removing it"
  git push origin :${release_branch}
fi
```

The checkout at line 103 is `fetch-depth: 1` and pulls no tags, so `git rev-parse "${release_tag}"` is **always false** — the tag guard is dead code that can never fire. The branch guard queries the remote and works correctly.

That asymmetry is the whole bug: the branch was deleted and recreated, the tag was left alone. It broke in the safe direction purely by accident. If that `rev-parse` ever *did* resolve, the workflow would silently delete a published release tag — which is worse.

Neither guard is the right behavior anyway. Deleting and recreating a published release's branch/tag is never correct; the run should refuse to start.

## Impact

- `release-26.09.02-01` misrepresented what shipped for ~2 hours (manually reset to `7fe2ff5`).
- 24 minutes of runner time and a full artifact build burned before a human caught it.
- Nothing was published — QEMU setup was cancelled, so Docker build/push and the Release / Release Notes / Changelog jobs never ran.
- Changelog generation was **not** affected: `gather-release-data` resolves its range from the releases API (`findPreviousTag` in `.github/scripts/gather-release-data/src/github.ts`), never from release branches.

## Proposed fix

Fail fast in Prepare, immediately after `Set Version Variables` and before anything is mutated: if a GitHub Release already exists for `release_tag`, error out and tell the operator to bump the counter. Delete the dead `git rev-parse` tag block.

No `force` escape hatch. Bumping the counter on a failed attempt is already the established convention — `26.08.19-01`, `-02` and `-03` all left behind empty-bodied releases before `-04` shipped, and `findPreviousTag` already skips undocumented releases so the changelog range stays correct.

## Acceptance criteria

- Dispatching a `release_version` whose `v<version>` release already exists fails within the first minute with an actionable error.
- The failure happens before the release branch is touched.
- Dispatching a fresh version is unaffected.
- The dead `git rev-parse` tag-deletion block is gone.

sfreudenthaler:
Closes: #37380

## Problem

`26.09.02-01` shipped on 09-02. On 09-03 the same version string was dispatched again against a `main` 54 commits ahead. The run burned 24 minutes — through a full 15-minute artifact build — before a human cancelled it at the Docker deploy step, by which point Prepare had already force-recreated `release-26.09.02-01` at a commit the published tag never referenced.

The prepare phase had two "already exists" guards, and they disagreed:

| Guard | Query | Fires? | Effect |
|---|---|---|---|
| Tag (line 148) | `git rev-parse` on the local clone | **Never** | dead code |
| Branch (line 165) | `git ls-remote --heads` on the remote | Yes | deletes + recreates the branch |

The checkout is `fetch-depth: 1` and pulls no tags, so `git rev-parse "${release_tag}"` can never resolve. The branch got clobbered; the tag survived only by accident. Had that `rev-parse` ever resolved, it would have done something worse — silently `git push origin :refs/tags/<published tag>`.

Neither guard is right. Re-cutting a published version is never correct, so the run should refuse to start rather than repair state.

## Change

```mermaid
flowchart LR
    D[workflow_dispatch] --> SV[Set Version Variables]
    SV --> G{"gh release view v&lt;version&gt;"}
    G -->|exists| F["❌ fail at ~1 min<br/>nothing mutated"]
    G -->|not found| CB[Create Release Branch and Tag]
    CB --> B[Build → Deploy → Release]

    style F fill:#ffdddd,stroke:#cc0000
    style G fill:#fff4cc,stroke:#d6a000
```

- New `Reject already-released version` step, placed immediately after `Set Version Variables` and **before** the branch is touched.
- Dead `git rev-parse` tag-deletion block removed.

Net: +18 / −6, one file.

## Why no `force` override

Bumping the counter on a retry is already how the team works — `26.08.19-01`, `-02` and `-03` each left behind an empty-bodied release before `-04` shipped. An escape hatch would just be a foot-gun re-armed, and `findPreviousTag` already skips note-less releases, so a burnt counter never corrupts the changelog range.

## Verification

The guard's predicate run against live repo state:

| Tag | State | Result |
|---|---|---|
| `v26.09.02-01` | published, has notes | REJECT ✅ |
| `v26.08.19-02` | failed attempt, empty body | REJECT ✅ |
| `v26.12.31-09` | does not exist | ALLOW ✅ |

`yaml.safe_load` passes on the modified workflow.

## Not in scope

`release-26.09.02-01` has already been reset out-of-band to `7fe2ff5` (the commit `v26.09.02-01` points at), so the branch matches the published tag again. `26.09.03-01` was cut separately as [run 33797883368](https://github.com/dotCMS/core/actions/runs/33797883368).

Changelog generation was never affected — `gather-release-data` resolves its range from the releases API, not from release branches.

This PR fixes: #37380

### DIFF
--- a/.github/workflows/cicd_comp_release-prepare-phase.yml
+++ b/.github/workflows/cicd_comp_release-prepare-phase.yml
@@ -141,15 +141,27 @@ jobs:
             echo "date=$(/bin/date -u "+%Y-%m")"
           } >> "$GITHUB_OUTPUT"
 
+      # A GitHub Release for this version means the pipeline already got at least
+      # this far for it, so re-dispatching would republish different artifacts under
+      # a version that is already public (see #37380). Refuse before anything is
+      # mutated. There is deliberately no force override: bumping the counter on a
+      # retry is already the convention (26.08.19-01..-04), and the changelog range
+      # tolerates it because findPreviousTag skips releases with no notes.
+      - name: Reject already-released version
+        env:
+          GH_TOKEN: ${{ secrets.CI_MACHINE_TOKEN || github.token }}
+          RELEASE_TAG: ${{ steps.set-version.outputs.release_tag }}
+          REPO: ${{ github.repository }}
+        run: |
+          if gh release view "${RELEASE_TAG}" --repo "${REPO}" --json tagName >/dev/null 2>&1; then
+            echo "::error::Release ${RELEASE_TAG} already exists. Re-releasing a published version would republish different artifacts under the same version, and force-recreates its release branch. Dispatch the next counter instead."
+            exit 1
+          fi
+          echo "No existing release for ${RELEASE_TAG} — proceeding."
+
       - name: Create Release Branch and Tag
         id: create-branch
         run: |
-          release_tag=${{ steps.set-version.outputs.release_tag }}
-          if git rev-parse "${release_tag}" >/dev/null 2>&1; then
-            echo "Tag ${release_tag} exists, removing it"
-            git push origin :refs/tags/${release_tag}
-          fi
-
           release_commit=${{ steps.set-version.outputs.release_commit }}
           # Shallow checkout (fetch-depth: 1) only guarantees HEAD of the default
           # branch. When an explicit release commit was supplied, fetch just that
--- a/.github/workflows/cicd_comp_release-prepare-phase.yml
+++ b/.github/workflows/cicd_comp_release-prepare-phase.yml
@@ -153,11 +153,29 @@ jobs:
           RELEASE_TAG: ${{ steps.set-version.outputs.release_tag }}
           REPO: ${{ github.repository }}
         run: |
-          if gh release view "${RELEASE_TAG}" --repo "${REPO}" --json tagName >/dev/null 2>&1; then
-            echo "::error::Release ${RELEASE_TAG} already exists. Re-releasing a published version would republish different artifacts under the same version, and force-recreates its release branch. Dispatch the next counter instead."
-            exit 1
-          fi
-          echo "No existing release for ${RELEASE_TAG} — proceeding."
+          # Read the HTTP status rather than an exit code: a bare non-zero from the
+          # API can mean rate-limited, unauthorized or offline just as easily as
+          # "absent", and treating those as absent would fail the gate OPEN — the
+          # branch guard below would then clobber the release branch, i.e. exactly
+          # the corruption this step exists to stop. Only a literal 404 proceeds.
+          status=$(curl -sS -o /dev/null -w '%{http_code}' \
+            -H "Accept: application/vnd.github+json" \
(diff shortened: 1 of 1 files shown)


sfreudenthaler:
Valid — fixed in 1d19ef4.

Took the HTTP status rather than the suggested stderr grep: `grep -qi 'release not found'` matches on English prose that isn't part of `gh`'s contract, so it rots silently on a CLI upgrade — and it rots *open*. A status code is unambiguous and needs no pinned-version verification.

```bash
status=$(curl -sS -o /dev/null -w '%{http_code}' ... /releases/tags/${RELEASE_TAG})
case "${status}" in
  404) proceed ;;
  200) reject  ;;
  *)   refuse — state indeterminate ;;
esac
```

Same `curl` idiom as the `Create GitHub Release` step further down, so no new dependency.

Verified against the live API:

| Case | Result |
|---|---|
| `v26.09.02-01` (published) | 200 → reject |
| `v26.08.19-02` (failed attempt, empty body) | 200 → reject |
| `v26.12.31-09` (absent) | 404 → allow |
| absent tag + bad token | **401 → refuse** (was the fail-open) |
| DNS failure | curl rc=6 → `set -e` aborts the step |

wezell (approved):
looks good

PR #37394, opened 2026-09-04
fix(content-drive): user-cache sizing + warm-up to cut per-row DB lookups (#37186)

### LINKED ISSUE #37186: Content Drive/Site Browser: undersized user cache causes repeated user_ lookups during folder listings

ihoffmann-dot:
## Summary
- Raises `cache.userdotcmscache.size` from the effectively-1000 default to 4000 (FR-002), sized against a reference ~1,828 active-user base.
- Adds a warm-up pass (`BrowserAPIImpl#warmUpUserCache`) that resolves each distinct `modUser`/`owner` on a listing page once, ahead of parallel row hydration, cutting redundant per-row `UserAPI#loadUserById` calls (User Story 1).
- Dedupes the mod-user-name lookup already done once per row (`DefaultTransformStrategy#resolveModUserName`, User Story 2).
- Adds an orphan-user fallback (`FR-004a`): a listing row whose `modUser`/lockedBy user was deleted no longer fails the whole page — it now degrades to `N/A` for that row only.
- Adds a `@VisibleForTesting` DB-lookup counter (`UserFactoryImpl`) used only by the new integration test to assert warm-up actually reduces lookups to one per distinct author.

## Test plan
- [ ] `./mvnw test -pl :dotcms-core -Dtest=UserFactoryImplTest,BrowserAPIImplTest,UserCacheConfigTest,DefaultTransformStrategyTest`
- [ ] `just test-integration-ide`
- [ ] `./mvnw verify -pl :dotcms-integration -Dcoreit.test.skip=false -Dit.test=BrowserAPITest`
- [ ] Confirm no `System.out`/`System.getProperty`/`System.getenv` introduced (checked via diff at implementation time)

Branched off the approved spec branch per this repo's Spec-Kit flow (spec.md-only in PR1, not merged to `main` yet).

This PR fixes: #37186

This PR fixes: #37186


nollymar (dismissed):

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
```diff
@@ -1127,6 +1127,58 @@ private List<Contentlet> loadContentletsParallel(Set<String> inodes, int chunkSi
...
+     * needing the same not-yet-cached id each miss the cache and query the database independently.
+     * A resolution failure for one id (e.g. an orphaned user) is logged and skipped — it must not
+     * abort the warm-up for the remaining ids, and the per-row fallback (FR-004a) still applies
+     * later during hydration for whichever row referenced it.
+     *
+     * @param contentlets the page of contentlets about to be hydrated
+     */
+    private void warmUpUserCache(final List<Contentlet> contentlets) {
+        for (final String userId : collectWarmUpUserIds(contentlets)) {
+            Try.run(() -> userAPI.loadUserById(userId))
```
fabrizzio-dotCMS, line 1182: SC-001 ("N distinct authors -> exactly N DB lookups") does not hold when one of those ids is an orphan, and I think that is worth either fixing or recording explicitly as a known residual.

`loadUserById` does not cache negative results: on a miss that finds no row it throws `NoSuchUserException` without adding anything to `UserCache`. So an orphan id stays permanently cold, and every subsequent resolution of it in the same request goes back to the database:

- once here in the warm-up,
- once per row in `DefaultTransformStrategy#addAuditProperties`,
- and for file assets and pages, once more per row inside `FileAsset#getMap` / `HTMLPageAsset#getMap` / `WebAsset#getMap`, which each call `loadUserById` themselves.

A page with one orphan author therefore costs `1 + N` lookups for that id, not 1 — the exact amplification pattern this issue set out to remove, just triggered by a different condition than the original diagnosis. The spec ruled out negative caching as *this dataset's* bottleneck (zero orphans in the test data), which is not the same as ruling it out as a mechanism.

The cheapest way to keep this honest without reopening negative caching: extend `test_getPaginatedContents_orphanModUser_doesNotFailWholeListing` to also assert the lookup count, so the number is pinned and visible rather than silently unbounded in the number of rows.

dotCMS/src/main/java/com/dotmarketing/portlets/contentlet/transform/strategy/DefaultTransformStrategy.java
```diff
@@ -413,10 +429,14 @@ private void addVersionProperties(final Contentlet contentlet, final Map<String,
...
-
-            final User user = toolBox.userAPI.loadUserById(lockedByOpt.get());
-            map.put("lockedBy", Map.of("userId", user.getUserId(),
-                    "firstName", user.getFirstName(), "lastName", user.getLastName()));
+            // issue #37186 (FR-004a): an orphaned locked-by id must degrade this one field, not
+            // fail the whole transform — same fallback pattern as modUser above and
+            // addAuditProperties, instead of an uncaught NoSuchUserException.
+            final User user = Try.of(() -> toolBox.userAPI.loadUserById(lockedByOpt.get())).getOrNull();
+            if (null != user) {
+                map.put("lockedBy", Map.of("userId", user.getUserId(),
```
fabrizzio-dotCMS, line 440: This changes the shape of the response, not just its failure mode: when the locked-by user cannot be resolved, `lockedBy` is now **absent** from the map, where before the request threw. The spec allows either ("omitted or falls back"), so this is in scope — but `locked` and `lockedOn` are still populated a few lines below, so a consumer can now see `locked: true` with no `lockedBy` at all, which was previously impossible.

Worth checking the Content Drive / Site Browser front end for a `lockedBy.firstName`-style dereference before merging. If any consumer assumes presence, a fallback entry (e.g. `userId` set to the unresolved id, names `N/A`) is safer than omission and stays closer to FR-003's "identical listing content" intent.

dotCMS/src/main/resources/dotmarketing-config.properties
```diff
@@ -515,7 +515,10 @@ cache.dotauthtokenreplaycache.size=200000
 #cache.rulescache.size=1000
 #cache.structuresbytypecache.size=1000
 #cache.tagsbyinodecache.size=1000
-#cache.userdotcmscache.size=1000
+#issue #37186 (FR-002): raised from the effectively-1000 default to guard against thrashing on
+#instances with a large active-user base (reference: 1,828 users). Line below (email region) is
+#a dead cache region (UserCacheImpl key-mismatch bug) and is deliberately left untouched.
+cache.userdotcmscache.size=4000
```
fabrizzio-dotCMS, line 522: SC-004 asks for the new default's memory footprint to be documented — approximate size of one cached `User` entry times the new entry count — so the trade-off is explicit rather than assumed. The comment above explains *why* the value changed and which sibling region to leave alone, both useful, but it does not quantify the cost.

A single line here with the rough per-entry size and the resulting worst-case for 4000 entries would close it. It also gives an operator raising this further a basis to estimate what they are signing up for.

dotcms-integration/src/test/java/com/dotcms/browser/BrowserAPITest.java
```diff
@@ -2531,4 +2534,172 @@ private static File fileNamed(final String name) throws IOException {
...
+        final BrowserQuery browserQuery = BrowserQuery.builder()
+                .showContent(true)
+                .withHostOrFolderId(folder.getIdentifier())
+                .offset(0)
+                .maxResults(20)
+                .build();
+
+        UserFactoryImpl.resetDbLookupCountForTesting();
+        final PaginatedContents firstPage = browserAPI.getPaginatedContents(browserQuery);
+        assertEquals("First (cold) request must resolve each distinct author exactly once",
```
fabrizzio-dotCMS: This asserts exact equality on `UserFactoryImpl.dbLookupCount`, which is a single JVM-global `AtomicLong`. Anything else in the JVM that resolves a user between the `reset` on the previous line and this assertion inflates the count and fails the test for reasons unrelated to the warm-up — background threads, cache-flush listeners, or any concurrently running test in the same suite.

The request under test also does more than hydrate rows: `hydrate` runs permission and workflow resolution per row, and any user lookup on those paths lands in the same counter.

Two options, either is fine: assert an upper bound (`<= expectedIds.size()`) instead of equality, which still fails loudly pre-fix where the count exceeded the distinct-author count; or keep equality but make the counter per-thread or scoped to the request so unrelated activity cannot contribute.

dotcms-integration/src/test/java/com/dotcms/browser/BrowserAPITest.java
```diff
@@ -2531,4 +2534,172 @@ private static File fileNamed(final String name) throws IOException {
...
+        final Set<String> expectedIds = new LinkedHashSet<>();
+        for (int i = 0; i < 12; i++) {
+            final User author = authors.get(i % authors.size());
+            final Contentlet contentlet = new ContentletDataGen(contentType)
+                    .setProperty("title", "warmup-" + i)
+                    .host(host)
+                    .folder(folder)
+                    .setPolicy(IndexPolicy.WAIT_FOR)
+                    .nextPersisted();
+            new DotConnect().executeUpdate("update contentlet set mod_user = ? where inode = ?",
```
fabrizzio-dotCMS, line 2591: This reassigns `mod_user` (and `identifier.owner` below) with raw SQL after `nextPersisted()`, which bypasses cache invalidation for both the contentlet and the identifier. If the listing path serves either from cache, `getModUser()`/`getOwner()` still return the system user and the warm-up set collapses to one already-warm id — the assertion below would then see 0 or 1 lookups, not 3, and the test fails for a fixture reason rather than a code reason.

The comment explains convincingly why `ContentletDataGen#user(User)` cannot be used on this persist path. Given that, the fixture would be sturdier if it flushed the affected caches after the updates (or re-read the contentlets from the DB) so the state the assertion depends on is guaranteed rather than incidental.

dotcms-integration/src/test/java/com/dotcms/browser/BrowserAPITest.java
```diff
@@ -2531,4 +2534,172 @@ private static File fileNamed(final String name) throws IOException {
...
+        // Two FKs stand in the way of a direct delete, unlike a genuinely orphaned production
+        // row (created by a path -- e.g. cross-environment content push -- that doesn't enforce
+        // this constraint against the destination's user_ table): users_cms_roles' fkusers_cms_
+        // roles2 (role assignment) and contentlet's fk_user_contentlet (the reference this test
+        // needs to keep). Postgres registers the "no referencing rows left" check trigger for a
+        // DELETE on the *parent* table (user_), not on contentlet -- disable triggers there for
+        // just this delete, scoped with try/finally so it is always restored even if the delete
+        // or an assertion below fails.
+        new DotConnect().executeUpdate("delete from users_cms_roles where user_id = ?", orphan.getUserId());
+        new DotConnect().executeUpdate("alter table user_ disable trigger all");
```
fabrizzio-dotCMS: I'd drop this DDL. Three separate problems with it:

1. **Permissions.** On Postgres, `DISABLE TRIGGER ALL` includes the internal triggers that implement foreign keys, and that requires **superuser** — table ownership only permits `DISABLE TRIGGER USER`. If the integration DB role is not superuser, this fails with a permission error that has nothing to do with what the test is verifying.
2. **Global DDL on a shared table.** While disabled, `user_` has no referential integrity for anything else touching that database, not just for this statement.
3. **Residue.** The `finally` restores the triggers but not the data: the contentlet keeps pointing at a `mod_user` that no longer exists in `user_` for the remainder of the suite. Any later test that transforms that contentlet or lists that folder inherits the orphan.

I checked whether the obvious alternative works and it does not: `contentlet.mod_user` has a real FK to `user_.userid` (`fk_user_contentlet`, `postgres.sql:1993`), so writing an invented id is rejected, and `DISABLE TRIGGER USER` on `contentlet` would not suppress an internal RI trigger either. There is no clean SQL route here.

The fix FR-004a actually delivers lives entirely in `DefaultTransformStrategy` — the `Try.of(...).getOrNull()` wrapper. That is fully covered by a unit test with a mocked `UserAPI` whose `loadUserById` throws `NoSuchUserException`, in the same style as the two `resolveModUserName` tests already added in this PR. No database, no DDL, no cross-test residue, and it asserts the same contract: the transform completes and the field falls back to `N/A`. I'd move the orphan coverage there and delete this integration test.

dotCMS/src/test/java/com/dotmarketing/business/UserCacheConfigTest.java
```diff
@@ -0,0 +1,53 @@
...
+            props.load(in);
+        }
+        return props;
+    }
+
+    /** The user cache must no longer fall back to the shared cache.default.size=1000. */
+    @Test
+    public void userCacheSize_isExplicitlySetTo4000() throws IOException {
+        final Properties props = loadDotmarketingConfigProperties();
+        assertEquals("4000", props.getProperty("cache.userdotcmscache.size"));
```
fabrizzio-dotCMS: This pins the literal string `"4000"`, which means the test fails the moment someone legitimately tunes the value — and a generous operator-facing default is exactly the kind of thing that gets tuned. The test then reports a false failure for a change the spec explicitly permits (FR-002 words it as SHOULD, with 4000 as a reference point, not a contract).

What is actually worth protecting is the property no longer being absent and no longer falling back to `cache.default.size`. Asserting that the key is present and parses to a value greater than 1000 keeps that guarantee and survives future tuning.

Same reasoning applies to `userEmailCacheSize_remainsUnsetDeadRegionNotRaised` below: the real invariant is that nobody sizes a region whose entries are never read back, and if the `UserCacheImpl` key mismatch is ever fixed, that test starts blocking a correct change. A comment in the properties file — which you have already added — carries that warning to the person who would actually make the mistake.

dotCMS/src/test/java/com/dotmarketing/business/UserFactoryImplTest.java
```diff
@@ -0,0 +1,49 @@
...
+ */
+public class UserFactoryImplTest {
+
+    @Before
+    public void resetCounter() {
+        UserFactoryImpl.resetDbLookupCountForTesting();
+    }
+
+    @Test
+    public void dbLookupCounter_startsAtZeroAfterReset() {
```
fabrizzio-dotCMS: These four tests exercise `AtomicLong` (increment adds one, reset zeroes) without ever calling `loadUserById`, which the class Javadoc is candid about. They cannot fail unless `java.util.concurrent` is broken, so they add maintenance surface without adding coverage.

The behavior worth asserting — that the counter increments exactly once per real cache miss and not at all on a hit — needs a database and belongs with the integration coverage. I'd delete this class and let the integration test be the counter's only test.

fabrizzio-dotCMS (commented):
Reviewed the implementation against `specs/37186-content-drive-user-cache-sizing/spec.md`.

The design is sound and matches the spec's revised root cause: sequential warm-up before the parallel fan-out, `modUserName` dedupe, orphan-user fallback, and the raised `cache.userdotcmscache.size`. Two things I verified rather than assumed, both fine:

- The dedupe in `resolveModUserName` is safe. Every option set in `DotTransformerBuilder` that contains `VERSION_INFO` also contains `COMMON_PROPS` (lines 165, 171, 192, 202, 217, 232), and `DotContentletTransformerImpl.defaultOptions` does too, so `addAuditProperties` always runs first and writes `MOD_USER_NAME_KEY`. The reuse branch is well-defined in every real path; the `WebAsset`/`FileAsset`/`HTMLPageAsset` `getMap()` overrides that pre-seed `"modUserName"` are always overwritten before `addVersionProperties` reads it.
- The dead-region claim in FR-002 holds: `UserCacheImpl#add` writes the email region keyed by the raw email address, while `#get` reads it keyed by `primaryGroup + key`, so nothing in that region is ever read back.

What I'd like addressed, roughly in order of impact — details inline:

1. SC-001 does not hold when a page contains an orphan author id (no negative caching), and no test covers the lookup count for that case.
2. The integration test asserts exact equality on a JVM-global static counter, and sets up its fixture with raw SQL that bypasses cache invalidation.
3. `alter table user_ disable trigger all` needs superuser on Postgres, is global DDL on a shared table, and leaves a permanent orphan row for the rest of the suite.
4. `lockedBy` changes from "throws" to "key absent" — worth confirming no consumer assumes it is present when `locked` is true.
5. SC-004 (documented memory footprint of the new default) is not satisfied.
6. Two of the new unit tests assert on things that will break under legitimate future changes without catching a real defect.

One process note: the spec's SC-001 requires the DB-lookup counter to be written and its test confirmed red *before* any implementation change (Constitution Principle V), but the first commit (980fc81) ships production code in `BrowserAPIImpl`, `UserFactoryImpl` and `DefaultTransformStrategy` with no tests. Also, the Javadoc on `collectWarmUpUserIds` and the test comments cite `plan.md`, which is not in the repo or in this PR — either add it or drop the references, since they are currently unresolvable for a reader.

fabrizzio-dotCMS (approved):
Third pass. Approving — nothing blocking left.

All six items from my first review are closed, and I verified each against the code rather than the commit messages: the DDL-based orphan test is gone and its coverage now lives in `DefaultTransformStrategyTest#addAuditProperties_orphanModUser_fallsBackToNotApplicable` with a mocked `UserAPI`; the cold-request assertion is an upper bound; the fixture flushes `ContentletCache`/`IdentifierCache` after the raw SQL updates; `lockedBy` is always present with a fallback entry when `locked` is true; the memory footprint is quantified in `dotmarketing-config.properties:521`; `UserCacheConfigTest` asserts a lower bound instead of pinning `4000`; and the SC-001 orphan amplification is recorded as an accepted residual in `warmUpUserCache`'s Javadoc. I also re-checked that the config key matches the cache region (`UserCacheImpl#primaryGroup` = `UserDotCMSCache` → `cache.userdotcmscache.size`).

Four small things left, none of them worth another round on its own — happy for them to ride along with any other change:

1. `dotCMS/src/test/java/com/dotcms/browser/BrowserAPIImplTest.java:153` still cites *"plan.md Legacy Impact carry-forward note 1"*. `plan.md` is not in the repo or this PR, so the reference is unresolvable for a reader. The equivalent rationale is already inline in `BrowserAPIImpl#collectWarmUpUserIds` — copying it here closes the last plan.md reference.

2. `dotcms-integration/src/test/java/com/dotcms/browser/BrowserAPITest.java:2629` is still a strict `assertEquals(0L, ...)` against the JVM-global `dbLookupCount`. The cold assertion was correctly relaxed to `<=`; the same shared-counter fragility applies to the warm one. Either relax it the same way or add a one-line comment on why the warm case cannot be inflated by unrelated activity where the cold case could.

3. `dotCMS/src/main/java/com/dotmarketing/portlets/contentlet/transform/strategy/DefaultTransformStrategy.java:395-407` — the old `addVersionProperties` Javadoc ("Use this method to add any additional property") is now stranded above `resolveModUserName`, so there are two stacked `/** */` blocks on that method and `addVersionProperties` is left undocumented. Move the old block back down.

4. `dotcms-integration/src/test/java/com/dotcms/browser/BrowserAPITest.java:60` — `import com.dotmarketing.util.UtilMethods` is added by this PR but unused; the only textual occurrence is inside a comment at line 1385.

One precision note on the dedupe, since my first review overstated the guarantee: `addAuditProperties` writes `MOD_USER_NAME_KEY` only `if (contentletProperties.containsKey(MOD_USER_KEY))`, and `FileAsset#getMap`, `WebAsset#getMap` and `HTMLPageAsset#getMap` pre-seed `modUserName` with `"unknown"` / the system user id on their own failure path. So the real invariant behind the reuse branch is "the map carries `modUser`", not merely "`addAuditProperties` ran first". Every option set in `DotTransformerBuilder` that requests `VERSION_INFO` also requests `COMMON_PROPS`, and listing rows always carry `modUser`, so no real path can reach the reuse branch with a pre-seeded `"unknown"` — not a defect, just the condition worth knowing if someone later calls this strategy with a hand-built option set.

CI note: unit tests, Junit5 Suite 1 and the OpenSearch Upgrade suite are green, but `MainSuite 3a` — the suite that actually runs `BrowserAPITest` — is still pending on this head, so the new integration test has not reported yet. Worth a glance before merge given item 2.

dotCMS-Machine-User (approved):

PR #37397, opened 2026-09-04
fix(content-drive): fold folder-scoped candidate resolution into a materialized CTE (#37229)

### LINKED ISSUE #37229: Content Drive: use a materialized folder-first CTE for the candidate-scan query on large folders

ihoffmann-dot:
## ⚠️ Not validated against FR-010 — read before reviewing

This spec's own plan makes running `EXPLAIN ANALYZE` against a real dataset (with workflow,
tag/relationship sub-queries, and the new tiebreaker in play) a **mandatory gate before the query
shape is trusted** — explicitly called "the single highest-risk unknown in this spec." That
validation requires a live Postgres instance with realistic data volume, which was not available
in the environment this PR was written in. **This PR was written and pushed without that
validation, as an explicit, disclosed risk** — see `specs/37229-content-drive-folder-cte/tasks.md`
task T004/T023, both left unchecked.

**Do not merge without running the FR-010 EXPLAIN ANALYZE gate locally first.** If it doesn't
hold, the query-shape tasks (see below) need rework, not just the tests.

## Summary
- Restructures `BrowserAPIImpl#selectQuery`/`buildSelectBaseQuery` to resolve folder (+ per-case `host_inode` + `fileName`) scoping via a `with candidates as materialized (...)` CTE, joined in place of the raw `identifier` table, before joining out to `contentlet_version_info`/`structure`/`contentlet` (FR-002).
- Scoped strictly to folder-scoped requests (`folder != null && !skipFolder`) — every other caller of this shared method is byte-identical to before, by design, to contain the blast radius of an unvalidated change.
- Adds a deterministic `id.id` `ORDER BY` tiebreaker (FR-001) so rows sharing the same `mod_date` sort reproducibly.

## Known gaps
- R3 names five site/host-scoping code paths; only two (explicit site, and `ignoreSiteForFolders=true`) were exercised with a test — the other three require constructing a `BrowserQuery` with `site == null`, which isn't reachable through the public `withHostOrFolderId` builder path as read in this pass. Flagged rather than faked.
- FR-009 (execution-count non-regression) and the deep-pagination cursor-stability case (FR-001) have no dedicated test yet.

## Test plan
- [ ] **FR-010 gate: run `EXPLAIN ANALYZE`** per quickstart.md against a folder confirmed to trigger today's slow plan, with the full predicate set (workflow, tags, tiebreaker)
- [ ] `just test-integration-ide`
- [ ] `./mvnw verify -pl :dotcms-integration -Dcoreit.test.skip=false -Dit.test=BrowserAPITest`
- [ ] Confirm no `System.out`/`System.getProperty`/`System.getenv` introduced (checked via diff)

Branched off the approved spec branch per this repo's Spec-Kit flow (spec.md-only in PR1, not merged to `main` yet).

This PR fixes: #37229

## Verification (2026-09-04, local) — correctness tests only, FR-010 still NOT run
- **8/8** unit tests (`BrowserAPIImplTest`) pass.
- **46/46** integration tests (`BrowserAPITest`) pass.
- Three real bugs were found and fixed, **all in this test's own permission-fixture setup, none in the CTE/tiebreaker production change**: reuse of a shared/singleton test user that polluted an unrelated pre-existing test, a missing explicit permission grant (folder-level READ doesn't propagate to new content by inheritance in this flow), and an extraneous `permissionIndividually()` call that didn't block access as intended. Confirmed via two `doesUserHavePermission` sanity checks isolating the permission setup from the candidate-scan path this PR actually touches.
- **FR-010 (the mandatory `EXPLAIN ANALYZE` gate) is still not run.** These tests confirm the CTE produces correct *results*; they say nothing about the query *plan* / latency claim, which is the actual point of this fix and requires a real dataset with a folder already confirmed to trigger today's slow plan. **Still do not merge without running it.**

## FR-010 — VALIDATED (2026-09-05)

Re-checked against `dotcms/dotcms:issue-37229-content-drive-folder-cte_SNAPSHOT` (commit `95b6031025`) on the original #37148/#37183 reference dataset (418k contentlets), using real `EXPLAIN ANALYZE` + `pg_stat_statements` attribution — not a simplified query.

**Result: the fix works where it matters.**
- `/outreach/` (21,383 children, the pathological case from #37148): plan now enters via `identifier_parent_path_asset_name_host_inode_key` instead of `idx_contentlet_mod_date`. Buffer hits: ~909k baseline → 90,351. Per-request DB-time attribution: 272ms → 114ms (~2.4x).
- Combined with a **tag filter**: still enters via the same good index (269,449 buffers); DB attribution 307ms → 272ms (marginal — the tag subquery itself is untouched by this fix, as expected).
- Combined with a **workflow filter**: still enters via the good index (24,331 buffers, already a fast case); DB attribution 40ms → 59ms (noise-level on an already-fast case).
- **No plan instability reintroduced** when the CTE is combined with workflow/tag filters — this was FR-010's actual open question, now closed.

**Trade-off found, spec updated**: on folders where the pre-fix plan was already good, the materialized CTE adds a real (not catastrophic) overhead purely from materialization. The spec's documented "+25-30ms" (SC-002) was optimistic — a second reference folder (`/uploads/news/`, 7,154 all-file children) measured +40ms (35-37ms → 75-77ms, ~2.1x), with buffer counts nearly identical before/after confirming the delta is materialization overhead, not extra data reads. **`specs/37229-content-drive-folder-cte/spec.md`'s SC-001/SC-002 were updated to +25-40ms and to record this confirmation** (pushed to the spec branch, PR #37230 — needs re-approval since the spec changed after sign-off).

**Known limitation of this re-check**: row-count parity was verified (40/40 in every case, matching pre-fix), but not row-by-row ID parity against the old code path, since the old code is no longer deployed and redeploying it for this specific diff was judged disproportionate. Recommendation from the re-check: add a large, all-file-asset folder shape (like `/uploads/news/`) to the permanent regression matrix, since it exposed a bigger delta than the original single reference case.

**Verdict**: fix confirmed to do what it claims, no correctness regression, overhead trade-off is bounded and now accurately documented. Recommend merge once PR #37230's spec update is re-approved.

This PR fixes: #37229


dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
```diff
@@ -2108,17 +2148,28 @@ static class SelectQuery {
...
+        final String baseClause = " from contentlet_version_info cvi, " + identifierSource
+                + " id, structure struc, contentlet c "
                 + " where cvi.identifier = id.id and struc.velocity_var_name = id.asset_subtype and  "
                 + " c.inode = cvi." + workingLiveInode + " and cvi.variant_id='"
                 + DEFAULT_VARIANT.name() + "' ";
 
-        final StringBuilder baseQuery = new StringBuilder(
-                "select cvi." + workingLiveInode + " as inode " + baseClause);
+        final StringBuilder baseQuery = new StringBuilder(candidatesCte)
+                .append("select cvi.").append(workingLiveInode).append(" as inode ").append(baseClause);
```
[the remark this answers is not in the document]
ihoffmann-dot: /fp Neither concatenated value is attacker-controlled. `workingLiveInode` is assigned from a fixed ternary (`browserQuery.showWorking || browserQuery.showArchived ? "working_inode" : "live_inode"`) — it can never be anything other than those two hardcoded column names, and this concatenation pattern already existed before this PR. `candidatesCte` is assembled entirely from fixed SQL text plus parameterized predicates: every append helper that contributes to it (appendFolderQuery, appendSiteQuery, appendFileNameQuery) binds values via `?` placeholders into the `parameters` list, never inlining caller input into the SQL string. No request data reaches either value as syntax.

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
```diff
@@ -2008,31 +2008,67 @@ private SelectQuery selectQuery(final BrowserQuery browserQuery) {
...
+                    appendSiteQuery(candidatesPredicates, browserQuery.site.getIdentifier(),
+                            browserQuery.forceSystemHost, parameters);
+                } else if (browserQuery.forceSystemHost) {
+                    appendSystemHostQuery(candidatesPredicates);
+                }
+            }
+            if (fileNameHandledByDb) {
+                appendFileNameQuery(candidatesPredicates, browserQuery.fileName, parameters);
+            }
+            candidatesCte = "with candidates as materialized (select * from identifier id where 1=1 "
```
fabrizzio-dotCMS: `select *` materializes every column of `identifier` for every row in the folder, and the outer query needs almost none of them.

I grepped the whole generated statement for `id.`-qualified references. Outside the CTE there are exactly two: `id.id` (the join to `cvi.identifier`, the `id.id in (...)` tag subquery, and the new ORDER BY tiebreaker) and `id.asset_subtype` (the join to `struc.velocity_var_name`). The other three columns the CTE touches — `parent_path`, `host_inode`, `asset_name` — appear only inside the CTE's own `where`, so they need to be scanned but not projected.

Everything else `identifier` carries is materialized into the work table and never read: `asset_type`, `owner`, `create_date`, `syspublish_date`, `sysexpire_date`, `full_path_lc`. On `/outreach/` that's 21,383 rows' worth of columns written to a work table to serve two of them.

This looks like the source of the trade-off the re-check found. On `/uploads/news/` you measured +40ms with buffer counts nearly identical before and after, and concluded the delta is materialization overhead rather than extra data reads — which is exactly what a narrower projection reduces. Worth re-running that same measurement with:

```sql
with candidates as materialized (
    select id.id, id.asset_subtype from identifier id where 1=1 ...
)
```

If the +40ms shrinks meaningfully, SC-002 gets better rather than being documented as a cost. Since `/uploads/news/` is all file assets and exposed the largest delta, that is the right folder to test it on — which also lines up with the re-check's own recommendation to add that shape to the permanent regression matrix.

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
```diff
@@ -2688,12 +2739,24 @@ private void appendExcludeArchivedQuery(StringBuilder sqlQuery) {
...
+        //
+        // FR-001 scopes this to folder-scoped requests only ("every folder-scoped listing
+        // request"), matching useFolderCte exactly -- every other caller's ORDER BY stays
+        // byte-identical to before (found in review: this was previously unconditional for any
+        // caller with sortBy set, silently changing tie order and pagination cursors for
+        // non-folder-scoped callers too).
         sqlQuery.append(" order by ");
         if (orderByDesc) {
-            sqlQuery.append(" c.mod_date desc");
+            sqlQuery.append(" c.mod_date desc").append(useFolderCte ? ", id.id desc" : "");
```
fabrizzio-dotCMS: `(c.mod_date, id.id)` is still not a total order, so FR-001's determinism guarantee has a hole left in it.

`appendLanguageQuery` emits `cvi.lang in (...)` over however many language ids the request carries, and `showDefaultLangItems` can add the default language on top of that. Nothing in the statement restricts the result to one row per identifier — the join is `cvi.identifier = id.id` with `cvi.lang` free — so a single identifier legitimately comes back as several rows, one per language.

For those rows `id.id` is identical. When they also share a `mod_date` (which they routinely do: a multi-language contentlet's versions are often saved together), the sort has nothing left to disambiguate them by, and the order is planner-dependent again — the precise condition the comment above says this tiebreaker converts into "a deterministic, reproducible-run-to-run guarantee". On a single-language folder the guarantee holds; on a multi-language one it does not.

Adding `cvi.lang` closes it, and it's already in scope with no new join:

```java
sqlQuery.append(" c.mod_date desc").append(useFolderCte ? ", id.id desc, cvi.lang desc" : "");
```

The existing tie test wouldn't catch this — it creates two file assets in the default language, so their `id.id` values differ and the tiebreaker resolves them. A version of it with the same identifier in two languages, both forced to one `mod_date`, would.

dotcms-integration/src/test/java/com/dotcms/browser/BrowserAPITest.java
```diff
@@ -2531,4 +2531,275 @@ private static File fileNamed(final String name) throws IOException {
...
+                .map(row -> (String) row.get("identifier"))
+                .filter(id -> id.equals(first.getIdentifier()) || id.equals(second.getIdentifier()))
+                .collect(Collectors.toList());
+        final List<String> secondRunOrder = browserAPI.getPaginatedContents(query).list.stream()
+                .map(row -> (String) row.get("identifier"))
+                .filter(id -> id.equals(first.getIdentifier()) || id.equals(second.getIdentifier()))
+                .collect(Collectors.toList());
+
+        assertEquals("Both tied rows must be present", 2, firstRunOrder.size());
+        assertEquals("Order among tied mod_date rows must be reproducible run-to-run",
```
fabrizzio-dotCMS, line 2591: This proves run-to-run reproducibility within a single page, which is real but is the weaker half of FR-001. The half that matters operationally — deep-pagination cursor stability, listed under Known gaps as having no test — is what the tiebreaker actually buys, and this fixture is already most of the way there.

The mechanism worth pinning: `getContentByChunks` pages by DB offset (`buildPaginatedDotConnect(sqlQuery, chunkSize, dbOffset)`), so with a non-total `ORDER BY` two requests for adjacent pages can order the tied rows differently and the boundary row is then either returned twice or skipped entirely. At the ~1.2% tie rate #37148 measured, that is not exotic on a 21k folder. Before this fix the ordering had nothing to disambiguate ties; after it, page boundaries are stable — that is the guarantee, and nothing currently fails if it regresses.

What would cover it, reusing this test's forced-tie approach: create enough contentlets in one folder to span several pages, force an identical `mod_date` across all of them with the same `DotConnect` update you use here, then page through with a small `maxResults`, collecting identifiers. Assert the union has no duplicates and its size equals the unpaged result count. That fails loudly on any non-total order and passes deterministically with the tiebreaker in place.

fabrizzio-dotCMS (commented):
Reviewed against `specs/37229-content-drive-folder-cte/spec.md`.

The FR-010 disclosure at the top of the description, and then the follow-up validation with real `EXPLAIN ANALYZE` and `pg_stat_statements` attribution on the 418k-contentlet reference dataset, is the right way to handle this. Recording that SC-002's "+25-30ms" was optimistic and updating the spec to +25-40ms rather than quietly letting the original number stand is worth more than the fix itself. Two things I checked and found sound:

- **Parameter binding order is correct.** The CTE text is prepended to the query while its parameters (folder, then site, then fileName) are appended to `parameters` first, so positional order still matches: CTE predicates, then workflow, then filter. Easy thing to get wrong in a change that moves predicates to the front of a statement.
- **The "byte-identical for non-folder-scoped callers" claim holds.** `shouldApplySiteFiltering` already requires `folder != null`, so the `!useFolderCte` branch is only reachable with `skipFolder=true`, and the `folder == null` case appends no site predicate either way — same as before. Scoping the `id.id` tiebreaker to `useFolderCte` in 7bcb1198 was the right correction; unconditional would have changed tie order and pagination cursors for every other caller.

Three code findings inline, plus one on mechanics:

**Base branch.** This targets `issue-37229-content-drive-folder-cte`, not `main`, while the three sibling implementation PRs (#37394, #37395, #37396) all target `main`. As it stands, merging this ships nothing to `main`, and CI is running against a base that isn't `main` — so a green build here says less than a green build on those. Worth retargeting once #37230's spec update is re-approved, or at least saying explicitly that the merge order is #37230 first and then a rebase.

dotCMS-Machine-User (dismissed):

dotCMS-Machine-User (dismissed):

dotCMS-Machine-User (approved):

PR #37403, opened 2026-09-04
37401 strict mode validate a diff scoped strict typecheck gate to stop new non strict code landing on main

### LINKED ISSUE #37401: strict mode: validate a diff-scoped strict typecheck gate to stop new non-strict code landing on main
## Research Question

Can a **diff-scoped** strict typecheck gate block new non-strict TypeScript from landing on `main`, **without** requiring the dependency libs to be strict first?

The gate would run each project's existing `tsconfig` with `--strict` forced on the command line, then **discard every diagnostic whose file is not part of the PR diff**. If that works, `main` stops accumulating non-strict code today — independently of when the workspace-wide PR (#37198) merges.

## Timebox

4h

## Acceptance Criteria

**PoC**

- [ ] A throwaway script exists that: (1) lists changed `.ts`/`.tsx` files (`git diff --name-only --diff-filter=ACMR origin/main...HEAD`), (2) maps them to their owning Nx project, (3) runs `tsc -p <project tsconfig> --noEmit --strict` per project, (4) filters diagnostics down to the changed files, (5) exits non-zero if any survive.
- [ ] The script needs **zero changes to any `tsconfig.json`** — confirmed that CLI flags override the inherited `strict: false` (see Context).

**Validation — it catches what it should**

- [ ] Run against PR #37262 (base = its merge-base on `main`, head = its merge commit). The gate **must flag** the 3 strict errors that PR introduced into `sdk-create-app`: `src/index.ts` (`TS4111`) and `src/utils/readiness.spec.ts` ×2 (`TS2345`). These are still present on `main` today, so the case is reproducible.
- [ ] Run against at least 2 more merged PRs that touched `.ts` in non-strict libs; record what it flags and whether each finding is real.

**Validation — it does not cry wolf**

- [ ] Run against at least 3 merged PRs that introduced no strict debt. The gate must pass on all of them.
- [ ] Record the false-positive rate and state an explicit **go / no-go for blocking on day 1**. A blocking gate is the target; if the measured noise makes that untenable, say so and recommend the fallback.
- [ ] Confirm that diagnostics originating in dependency sources are actually filtered out (the whole premise) — report how many were discarded per run.

**Decisions the spike must land**

- [ ] **Flag set**: full `--strict`, or start with `--strictNullChecks --noImplicitAny` only. Report the error count each produces on the sample PRs.
- [ ] **Granularity**: whole changed file, or only the changed lines/hunks. Whole-file is simpler but makes anyone touching one line of a legacy file inherit the whole file's debt; line-level is more code but far more adoptable. Recommend one, with the measured cost of each on the sample PRs.
- [ ] **Runtime cost per PR**, measured (baseline: 2.4s for one portlet, see Context).

**Edge cases to prove out**

- [ ] PR that touches no `.ts` at all → gate is a no-op, passes.
- [ ] Deleted and renamed files in the diff → no crash, no phantom failures.
- [ ] PR that touches `tsconfig.base.json` or `nx.json` → these are `sharedGlobals` in `nx.json`, so `nx affected` returns **all 56 projects**. Decide whether the gate fans out to everything or stays scoped to projects that own changed files (the latter is the intent).
- [ ] Project that has no `tsconfig.lib.json` (apps, `.tsx`/Vue projects) → resolves the right tsconfig or skips explicitly, never silently.
- [ ] Runs on a `fetch-depth: 1` shallow checkout and under `merge_group`, where `origin/main` must be fetched explicitly.

**Handoff**

- [ ] Findings + recommendation written up on this issue.
- [ ] Follow-up Task opened for the real implementation (script + CI hook + `lint-staged` hook), or this issue closed with a documented "won't work, because X".

### Why now, and why this is not the same as PR #37198

Epic #35932 is being delivered by PR #37198 — **1455 files** (1111 `.ts`: 670 production + 441 spec; 152 HTML; 71 `tsconfig`s; 48 `.tsx`). That PR is expected to merge eventually, but it needs full team QA before it can, and that is not imminent.

The problem this spike addresses is **not** getting #37198 merged. It is that while #37198 waits, **new non-strict code keeps landing on `main`**. Every sync from `main` into the strict branch imports fresh type errors that have to be fixed by hand, so the branch's diff grows and the QA target keeps moving. A recent example: PR #37262 landed 3 strict violations in `sdk-create-app`, which had to be fixed on the branch after a merge.

This gate is what stops the bleeding, and it pays off in both directions:

- If #37198 merges soon → the gate keeps `main` strict-clean from then on.
- If #37198 keeps waiting → the gate keeps its diff from growing, so QA has a stable target.

Either way it is decoupled from that PR's timeline, which is why it is worth its own spike.

### Why per-lib opt-in stalls (the mechanism, measured)

`main` has `strict: false` in `tsconfig.base.json`, with **22 of 55** project tsconfigs opting into `strict: true` locally. Opting in a 23rd stalls, because the paths in `tsconfig.base.json` point at **sources** (`./libs/x/src/index.ts`), not `dist` — so a dependency's sources become part of your program and get checked with **your** flags.

Measured on `libs/portlets/dot-locales/portlet` (`tsc --listFiles`, excluding `node_modules`):

| Files in program | Origin |
|---|---|
| 8 | the portlet itself |
| 143 | `libs/ui/src` |
| 107 | `libs/dotcms-models/src` |
| 93 | `libs/data-access/src` |
| 25 | `libs/dotcms-js/src` |
| 11 | `libs/global-store/src` |
| 8 | `libs/utils/src` |
| **387** | **total from 6 dependency libs** |

To check 8 files of your own, TS checks 387 files it did not ask about. `skipLibCheck: true` (already on) does not help, because these are `.ts` sources, not `.d.ts`.

Fan-in of the bottleneck libs, by number of files importing them:

| Consumers | Lib |
|---|---|
| 1130 | `@dotcms/dotcms-models` |
| 1006 | `@dotcms/data-access` |
| 585 | `@dotcms/ui` |
| 207 | `@dotcms/dotcms-js` |
| 91 | `@dotcms/utils` |

Until those three land, any opt-in upstream of them drowns. **The insight this spike tests: the dependency does not need to be strict — its errors just need to stop counting.** That is a filter, not a migration.

### Key finding that makes this cheap

`tsc -p <tsconfig> --strict` **overrides** the `strict: false` inherited from `tsconfig.base.json`. Verified against a synthetic project (`strict: false` in the config, two deliberate violations):

```
$ tsc -p tsconfig.json                    # → no errors
$ tsc -p tsconfig.json --strict           # → TS7006 implicit any, TS18047 possibly null
$ tsc -p tsconfig.json --strictNullChecks # → TS18047 only
```

Two consequences: **no overlay `tsconfig` and no edits to existing configs are needed**, and the flags can be adopted one at a time.

**Cost**: `tsc -p libs/portlets/dot-locales/portlet/tsconfig.lib.json --noEmit --strict` = **2.4s wall** for a 396-file program. A PR touching 1–3 projects should land in 5–10s.

### Existing infrastructure this plugs into

- **A typecheck gate already exists — but only on the strict branch.** Commit `6be735de51` added `nx affected -t typecheck --base=origin/main --exclude=tag:skip:typecheck` to `core-web/pom.xml`. It is **11 files, +65/-24 lines**, is build-only, and is **not on `main`**. Worth considering as its own small PR regardless of this spike's outcome.
- Per that target's own comment in `pom.xml`, it covers **4 projects** against lint's 47. The root `core-web` project carries `skip:typecheck` because `tsc -p tsconfig.base.json` compiles the whole workspace as one program and reports **~2570 errors** — so "one program for everything" is not a viable gate shape.
- **CI hook point**: `core-web/pom.xml`, `generate-resources` phase under the `-Pvalidate` profile, next to the existing `typecheck-test` / `lint-test` / `format-test` executions. `cicd_comp_test-phase.yml` already runs a `Fetch origin/main for Nx affected` step (the checkout is `fetch-depth: 1`), so the base ref the gate needs is already guaranteed — **no workflow changes required**.
- **Local hook point**: `core-web/lint-staged.config.mjs`, which already runs `nx affected -t lint` over staged files via husky.
- **Precedent for baselining debt already accepted in this repo**: 8 `eslint-suppressions.json` files (ESLint bulk suppressions), with `{projectRoot}/eslint-suppressions.json` declared as a `lint` input in `nx.json`. Proposing the TS equivalent is not asking for a new pattern.

### Alternatives to weigh in the write-up

| Option | Trade-off |
|---|---|
| **Diff-scoped filter** (this spike) | Cheap, no config changes, immune to dep debt. Weaker signal: an `any` coming out of a non-strict dep still passes silently. |
| **Baseline / suppressions file** | Also catches regressions in untouched files. Baseline is large and conflict-prone. Precedent exists (ESLint). |
| **Per-project opt-in** | Already 22/55 done. Blocked on the three high fan-in libs. |
| **Project references / buildable libs** | The actual cure for dep bleed-through (deps compile to `.d.ts` and stop being re-checked). Large migration; out of scope here. |

### Known blind spots of the proposed gate

- **Angular templates**: `tsc` does not check HTML. `strictTemplates` is a separate gate via the Angular compiler, and is out of scope.
- **`any` propagation**: while `dotcms-models` / `data-access` / `ui` stay non-strict, sloppy types flow into strict files without error. The gate gets stronger as those migrate; it does not depend on them to start being useful.
- **`nx affected` is not a substitute**: it scopes *projects*, not *files*. Inside an affected project you still get all 387 dependency files. It limits how much you run, not what counts.

## Links

- Epic: #35932 — strict mode across the core-web workspace (6/45 sub-issues done)
- PR #37198 — the workspace-wide strict mode PR this gate is meant to protect
- PR #37262 — the reproducible regression case: landed `TS4111` + 2× `TS2345` in `sdk-create-app`
- Commit `6be735de51` — the `nx affected -t typecheck` CI gate, currently branch-only
- `core-web/pom.xml` — `typecheck-test` / `lint-test` executions, and the `~2570 errors` note on the root project
- `core-web/lint-staged.config.mjs` — the local hook point
- `.github/workflows/cicd_comp_test-phase.yml` — the `Fetch origin/main for Nx affected` step

nicobytes:
This pull request adds a comprehensive feature specification document for a "diff-scoped strict typecheck gate" in the `core-web` workspace. The spec outlines the motivation, user scenarios, requirements, measurable outcomes, and legacy considerations for implementing a TypeScript strict-mode gate that only blocks new type errors introduced in a pull request, rather than all existing debt. The goal is to prevent new non-strict code from landing on `main` while a broader migration is still pending.

Key additions and themes:

**Specification and Problem Framing**
* Introduces a detailed problem statement explaining why a diff-scoped strict typecheck gate is needed, focusing on the challenges of partial strictness and the ongoing migration effort.
* Clearly defines the research question: whether it is feasible to block only new strict-mode violations in changed files, decoupling debt prevention from the full migration timeline.

**User Scenarios and Acceptance Criteria**
* Outlines four user stories (with priorities), including ensuring the gate catches new debt, avoids false positives, supports measured operating decisions, and produces a durable, actionable write-up. Each story includes independent tests and acceptance scenarios.
* Lists edge cases to ensure robustness, such as handling deleted files, shared configs, unmapped files, and shallow checkouts.

**Functional Requirements and Success Criteria**
* Specifies functional requirements for the evaluation harness, covering file detection

This PR fixes: #37401


nicobytes:
Thanks for the review. I dug into the `55` vs `56` finding and it turns out both numbers are correct — they count different things, so the denominator shouldn't become `56`.

**What the workspace actually has:**

| Count | Value | Command |
|---|---|---|
| Nx projects (`project.json`) | **56** | `find apps libs -name project.json -not -path '*/node_modules/*' \| wc -l` |
| Projects with a `tsconfig.json` | **55** | `find apps libs -name tsconfig.json -not -path '*/node_modules/*' \| wc -l` |
| `"strict": true` | 22 | `grep -rl '"strict"[[:space:]]*:[[:space:]]*true' apps libs --include='tsconfig*.json' \| wc -l` |
| `"strictTemplates": true` / `false` | 30 / 4 | same, on `strictTemplates` |

The gap is `libs/dotcms-scss`, which is a pure-SCSS library and ships no `tsconfig.json`.

So `22 of 56 project configs` would be inaccurate — there are only 55 TypeScript configs to opt in. The numerator (22) and the denominator both come from tsconfigs; taking the denominator from `project.json` mixes two different populations.

The `56` at `spec.md:207`, `lib/project-map.mjs:6` and `project-map.test.mjs:53` is a separate, correct figure and needs no change: it's the Nx *project* fan-out from `nx affected` when a `sharedGlobals` file is touched — a count of projects, not of TypeScript configs.

**Fix applied** — the real defect was terminological, not numeric: the same phrase ("project configs" / "projects") was carrying two different meanings. `spec.md:18` now reads:

> the shared TypeScript baseline turns strict mode **off**, and 22 of 55 TypeScript project configs opt back in locally (the workspace has 56 Nx projects; one of them ships no `tsconfig.json`).

That keeps both counts accurate and removes the clash with "all 56 projects" further down.

**The three non-blocking notes are all fixed**, and the `merge-base` one turned out to deserve more than a comment — thanks for flagging it.

- `changed-files.mjs:75-78` — you were right that the fallback reinstates the two-tree comparison the surrounding comment warns against. That is the same defect §7 records as having reported *50 findings, essentially none of them the branch's own*, so silently degrading back into it is worse than it first looks. It still falls back (that is the safe behaviour), but now warns on stderr and reports `baseResolution: 'merge-base' | 'base-tip'` so the numbers can be distrusted rather than believed.
- `run.mjs:159` / `replay.mjs:140` — left as-is, agreed it is not reachable in practice.
- `filter.mjs:39` — guarded anyway; a path with no separator can no longer produce an empty prefix that matches everything.

While in there, a review of the same area turned up two silent degradations of the same family, both verified by running them:

- An unknown `--granularity` fell through the `=== 'line'` test and behaved as **whole-file**, while the report still echoed the name it was given. §6 measures whole-file at 83 % inherited findings, and a plural typo (`--granularity lines`) was enough to trigger it.
- `check-ng` resolved an unknown flag set to `strict` where `check-ts` threw on the same input, so a `--templates` run could measure one flag set and report another.

Both now fail by name, `run.mjs` validates all three enumerated options at the CLI edge, and four regression tests cover them. 97/97 passing (93 before).

Two other things landed on the branch since your review: a `DECOMMISSION.md` recording how this gate is removed once #37198 merges, and a correction to `findings.md`, whose status header still listed US3 and US4 as pending when §6 and §7 are exactly those, complete.

oidacra (approved):

PR #37416, opened 2026-09-04
feat(sdk-cli): `dotcms agent setup` — one command to connect an IDE to dotCMS

### LINKED ISSUE #37390: Add `dotcms agent setup` — one command to connect an IDE to dotCMS
### Description

Connecting an AI coding agent to a dotCMS instance takes four manual steps today: find the admin panel, mint an API token by hand, hand-edit whichever config file your editor happens to read, then install the dotCMS skills separately. The `@dotcms/mcp-server` README is a copy-paste walkthrough, and it covers only two editors.

This adds a `dotcms` CLI whose first command collapses all four into one. The developer answers three questions and can immediately start their agent with the dotCMS MCP server connected and the dotCMS skills installed.

```bash
# Interactive — detects your editors, asks URL, auth, and which to wire up
npx dotcms agent setup

# Scripted: mint a token from credentials
npx dotcms agent setup --url https://demo.dotcms.com \
  --user admin@dotcms.com --password ***** \
  --agent cursor --agent claude-code --global

# Scripted: use a token you already have (mutually exclusive with --user/--password)
npx dotcms agent setup --url https://demo.dotcms.com --authToken ***** --agent cursor
```

Supplying the URL and one auth mode is enough — setup then runs to completion without prompting. Targets default to every detected editor, scope to the folder you ran it in.

**Targets:** Claude Code, Cursor, VS Code/Copilot, Codex, Antigravity, Devin, OpenCode.

**Spec:** `specs/37390-dotcms-agent-setup/spec.md`

### Acceptance Criteria

- [ ] `npx dotcms agent setup` takes a developer from nothing to a connected agent in ≤3 answers, with no manual file editing
- [ ] Installed editors are detected and pre-selected; any supported editor can also be chosen explicitly
- [ ] Auth is either `--user`/`--password` (mints a token) **or** `--authToken` — passing both is a usage error, not a silent preference
- [ ] The token is validated against the instance **before** any config file is written; if validation fails, nothing is written at all
- [ ] Every pre-existing MCP server and unrelated setting in a touched config file survives verbatim
- [ ] Files holding the token are owner-only; the token is never logged, shown in full, or passed as a process argument
- [ ] Project scope names every file it wrote a token into and offers to `.gitignore` them
- [ ] Skills install in the same run; failure is non-fatal and prints the exact command to re-run
- [ ] The configured server is launched and confirmed to list its tools before the run reports success
- [ ] A failure on one target still configures the others and exits non-zero
- [ ] A fully non-interactive run with all options supplied produces the same result as the interactive one
- [ ] All 7 targets supported; adding an 8th requires describing only that editor

### Priority

Medium

### Additional Context

- **Scope is `agent setup` only.** `status` and `remove` are deliberately not in this release — re-running setup replaces a stale entry, and removing one is a hand edit of a single named key. `create-app` stays where it is and the dotCLI port is not started here; the `agent` sub-command group is the seam all of these fold into later.
- Ships as the `dotcms` npm package, date-lockstep with the dotCMS release per ADR-0019. Repurposing that name replaces a dormant `0.0.21` library — confirm download volume before first publish.
- **Known pipeline gap:** the SDK release action's "already published" guard is `npm view "@dotcms/${dir}@${version}"` — scope hardcoded, name from the directory. An unscoped package publishes fine the first time but every re-run then fails the release step. Needs a fix alongside this.
- `dotCMS/agent-toolkit` is now public, so the skills half works for external developers.

fmontes:
file-1fec1824ff29861aa0bc7dba5669a1c1

**Spec-Kit PR 2 of 2** — implementation. The spec is [#37392](https://github.com/dotCMS/core/pull/37392); its commits are the shared ancestor here and disappear from this diff once it merges.

Fixes #37390

## What it solves

Connecting an AI coding agent to dotCMS took four manual steps: find the admin panel, mint an API token, hand-edit whichever config file your editor reads, install the skills separately. Seven editors, no two storing it the same way. The `@dotcms/mcp-server` README covered two of them.

```bash
npx dotcms agent setup
```

Three questions — instance, credentials, editors — and the agent is connected.

## How it works

```
resolve URL → confirm it IS dotCMS → resolve credentials → mint → VERIFY
  ── nothing has touched the filesystem yet ──
→ write configs (merge, chmod 0600) → .gitignore offer → skills → launch server → summary
```

Four properties do most of the work:

**Nothing is written until the token verifies.** A rejected token leaves no file, no directory, no skills install — so a bad credential can't produce seven configs that fail confusingly later. `--yes`/`--force` cannot disable it.

**The instance is validated before you're asked for anything.** `/api/v1/appconfiguration` is fingerprinted for a real dotCMS payload; a proxy or CDN answering 200 is rejected. Typing a password against a wrong address was wasted effort.

**Merge, never clobber.** These files already hold other MCP servers. Only the `dotcms` key is touched; an unparseable file is a named error that leaves it untouched. For Codex the writer *parses to validate and splices to write*, so hand-written comments survive.

**The run proves the agent connects.** After writing, it launches the configured server and confirms it lists tools. A valid token is not proof the server starts.

## Command surface

```bash
npx dotcms agent setup --url … --user … --password …      # mints a token
npx dotcms agent setup --url … --authToken …              # uses one you have
```

`--url` plus one auth mode are the only required inputs; supply both and it runs unprompted, terminal or not. Targets default to every detected editor, scope to the current folder. The two auth modes are mutually exclusive — passing both is a usage error, not a silent preference.

`--agent` (repeatable) · `-g/--global` · `--skip-mcp` · `--skip-skills` · `--skip-verify` · `-y` · `--force`. Exit `0` all succeeded, `1` a target or the connection check failed, `2` usage error.

Targets: Claude Code, Cursor, VS Code/Copilot, Codex, Antigravity, Devin, OpenCode.

## New library: `@dotcms/http`

`create-app` and this CLI both authenticate against a dotCMS instance. `http.ts` sets `Authorization` headers and follows redirects — the exact surface **axios was removed from this workspace over** (its Node adapter leaked `Proxy-Authorization` across a redirect, #37264). Two copies of that code is where the next such fix lands in one package and not the other. That, not DRY, is the case for extracting it.

**`core-web/libs/http/`** — internal, never published:

| | |
|---|---|
| `http.ts` | `httpGet`/`httpPost` over native `fetch`; bearer auth, `AbortController` timeout, `HttpError` carrying an HTTP status or a transport code |
| `fetch-retry.ts` | `describeRequestFailure()` — turns `ENOTFOUND` into a sentence, kept diagnostic so callers own the remedy |
| `result.ts` | `Result<T,E>` |
| `endpoints.ts` | the three `/api/v1` paths, so the two CLIs can't drift |

**Consumers:** `@dotcms/create-app` (moved off its own copies — 8 import edits, its 132 tests pass unchanged) and `dotcms`. Both bundle it in at build time, so **nothing new reaches npm**.

It sits **outside `libs/sdk/`** deliberately: the SDK release action publishes every direct child of that directory. `private: true` is intent, not the mechanism — npm only enforces `private` for workspace publishes — so `verify-package` asserts the boundary instead.

## What to review for

- **`shared/` vs `commands/agent/`** — the package is organised by command group, not technical layer, because `create-app` and the dotCLI port fold in as siblings later. An ESLint rule enforces one-way imports.
- **`targets/registry.ts`** — every per-editor difference is data. Adding an eighth editor is one object literal.
- **The ordering guarantee** in `setup.ts`, and its test. It's the thing most worth breaking on purpose to check.

## Also in here

- **CI fix**: the SDK publish guard was `npm view "@dotcms/${dir}@${version}"` — scope hardcoded, name from the directory. `dotcms` is unscoped, so the guard could never match: first publish succeeds, every re-run then fails the release. Now reads `.name`. Of eleven SDK packages, this is the only one that differs.
- `apps/mcp-server/README.md` points at the command; the manual steps stay.
- `libs/sdk/cli/**` and `libs/http/**` added to `nx.json`'s jest include — without it a project gets no `test` target at all.

## Testing

`sdk-cli` **217** · `http` **39** · `create-app` **132** — all green, lint clean, production build emits the shebang.

`nx run sdk-cli:verify-package` asserts eight packaging invariants against the artifact npm would upload, and runs as part of `nx test`. Each check corresponds to a defect that actually happened: a missing shebang, `@dotcms/http` surviving as an unresolvable import, declared-but-unused dependencies, an internal library drifting under `libs/sdk/`.

Fixes were mutation-verified rather than assumed: reverting one turns its test red. That found three holes where a test existed but could not fail.

## Known limitations

- `agent status` and `agent remove` are not in this release. Re-running setup replaces a stale entry.
- File permissions are POSIX-only; `chmod` doesn't touch Windows ACLs, and the summary says so rather than implying protection.
- A failed run can leave an orphaned token — minting precedes writing. Accepted: recovering it would mean printing it, which FR-022 forbids.
- The server reference is unpinned (`@latest`). Justified deviation from ADR-0019 — `@dotcms/mcp-server` isn't date-lockstep, so there's no matched version to pin to.

## Before merge

- [ ] **PR #37392 approved** — this is gated on the spec, not on its merge
- [ ] **Confirm `dotcms@0.0.21` download volume before first publish** (381/month; `^0.0.21` pins exactly, so only bare `npm i dotcms` changes)
- [ ] Manual `quickstart.md` run, including real editors connecting


core-web/libs/sdk/cli/src/commands/agent/setup.ts
```diff
@@ -0,0 +1,321 @@
...
+    const explicitTargets = resolveTargets(opts);
+    const scope = opts.scope ?? 'folder';
+
+    const step = opts.onProgress ?? (() => undefined);
+
+    // 2. The address first, and CHECKED first. Only once the instance is confirmed to be a
+    //    real dotCMS is anyone asked for a credential: a password typed against a wrong
+    //    address is wasted effort, and the failure would land after the work rather than
+    //    before it.
+    const url = await resolveInstanceUrl(opts, opts.promptPort);
```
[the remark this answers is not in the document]
fmontes: Fixed in c3c2a1c78e. `checkReachable`'s result is now kept and compared against `TOOL_VERSION`, surfaced via `onWarning` (live) and `warnings[]` (summary). `TOOL_VERSION` reads `package.json` at build time — the release pipeline rewrites `.version` before building, so a literal would report `0.2.0` from a package published as `26.9.x`.

core-web/libs/sdk/cli/src/commands/agent/setup.ts
```diff
@@ -0,0 +1,321 @@
...
+    step(`Checking ${url}`);
+    await checkReachable(url);
```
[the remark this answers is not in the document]
fmontes: Duplicate of the finding above — fixed in c3c2a1c78e.

core-web/libs/sdk/cli/src/commands/agent/setup.ts
```diff
@@ -0,0 +1,321 @@
...
+            token = await verifyToken(url, token);
+            break;
+        } catch (error) {
+            const rejected =
+                error instanceof TokenRejectedError || error instanceof CredentialsRejectedError;
+            if (!rejected || !opts.promptPort || attempt >= MAX_AUTH_ATTEMPTS) throw error;
+
+            opts.onAuthRetry?.((error as Error).message, attempt, MAX_AUTH_ATTEMPTS);
+            // Ask again from scratch: the url is settled, the credential is what was wrong.
+            inputs = await resolveRequiredInputs({ url, cwd: opts.cwd }, opts.promptPort);
```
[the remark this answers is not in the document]
fmontes: Fixed in c3c2a1c78e, though not with `delete process.env[...]` — mutating the environment would also affect anything else reading it. Instead extracted `promptForAuth(port)`, which consults neither options nor env, and the retry calls that. For a credential the instance just rejected, the only useful source is the human.

core-web/libs/sdk/cli/src/commands/agent/connect.ts
```diff
@@ -0,0 +1,111 @@
...
+
+        // Speak MCP: initialize, then ask for the tool list.
+        child.stdin?.write(
+            `${JSON.stringify({
+                jsonrpc: '2.0',
+                id: 1,
+                method: 'initialize',
+                params: { protocolVersion: '2024-11-05', capabilities: {}, clientInfo: { name: 'dotcms', version: '0' } }
+            })}\n`
+        );
```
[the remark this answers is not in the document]
fmontes: Valid, but deliberately not in this commit — scoped to the two findings above. Tracking as follow-up.

core-web/libs/sdk/cli/src/commands/agent/connect.ts
```diff
@@ -0,0 +1,111 @@
...
+const DEFAULT_TIMEOUT_MS = 60_000;
+
+/** Separating these matters: FR-024c requires naming the causes we can distinguish, and with no
+ *  verbose mode the message is all the developer gets. */
+function classify(stderr: string, code: number | null): ConnectFailure {
+    if (/404|E404|not found|ETARGET|ENOTFOUND|registry/i.test(stderr)) return 'fetch-failed';
+    if (/Unsupported engine|requires Node|SyntaxError|Unexpected token/i.test(stderr)) {
+        return 'runtime-unsupported';
+    }
+    return code === null ? 'exited' : 'exited';
```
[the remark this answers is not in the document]
fmontes: Valid, but deliberately not in this commit — scoped to the two findings above. Tracking as follow-up.

core-web/libs/sdk/cli/src/shared/ui.ts
```diff
@@ -0,0 +1,119 @@
...
+        lines.push(chalk.green('  ✓ server responded'));
+    } else if (input.connection === 'skipped') {
+        lines.push('  · connection check skipped');
+    } else {
+        // Deliberately avoids the word "ready" — the run did not reach that state.
+        lines.push(chalk.red(`  ✗ ${input.connectionReason ?? 'the server did not start'}`));
+        lines.push('    Configuration was written and left in place; the server did not come up.');
+    }
+
+    const allGood = input.connection === 'ok' && input.outcomes.every((o) => o.result !== 'failed');
```
nicobytes: With `outcomes: []`, `[].every()` is `true`, so `allGood` stays true and a run that **wrote nothing** prints:

```
  ✓ server responded

  Ready — open your editor and start using dotCMS.
```

(verified by running `renderSummary({ outcomes: [], connection: 'ok' })`)

This happens on two real paths: non-interactive with no editors detected (`setup.ts` → `targets = detected` = `[]`), and interactive where the developer deselects everything. The spec forbids it explicitly — Edge Cases, line 117: *"No supported editor is detected at all — setup must say so and let the developer pick targets explicitly rather than exiting silently."*

Suggestion: an explicit case for `plan.length === 0` in `setup.ts` that pushes a warning naming the selectable ids, and have `allGood` require at least one outcome.

core-web/libs/sdk/cli/src/commands/agent/setup.spec.ts
```diff
@@ -0,0 +1,1018 @@
...
+            cwd: dir,
+            skipSkills: true,
+            skipVerify: true
+        });
+
+        expect(detect).toHaveBeenCalled();
+        expect(result.outcomes.map((o) => o.targetId).sort()).toEqual(['claude-code', 'cursor']);
+    });
+
+    it('writes nothing and does not throw when no editor is detected', async () => {
```
nicobytes: This test locks in the wrong behavior: it asserts `outcomes: []` and `exitCode: 0` without requiring anything to be said. The spec asks for the opposite (Edge Cases, line 117). It should be inverted along with the fix in `ui.ts`/`setup.ts`.

core-web/libs/sdk/cli/src/commands/agent/connect.ts
```diff
@@ -0,0 +1,124 @@
...
+ * merely convenient. It is also exactly how the editor will launch it.
+ */
+export async function confirmConnection(args: {
+    url: string;
+    token: string;
+    timeoutMs?: number;
+}): Promise<ConnectResult> {
+    const timeoutMs = args.timeoutMs ?? DEFAULT_TIMEOUT_MS;
+
+    const child = childProcess.spawn('npx', ['-y', MCP_SERVER_PACKAGE], {
```
nicobytes, line 43: `spawn('npx')` without `shell: true` does not work on Windows: `npx` there is `npx.cmd`, and since the CVE-2024-27980 fix Node refuses to execute `.cmd`/`.bat` without a shell.

Windows is in scope — the design handles it explicitly (`CAN_RESTRICT`, `%APPDATA%` in the registry, `target-configs.md:24`, research R5). Result: FR-024a fails on **every** Windows run with `cause: 'exited'` / `detail: 'spawn npx ENOENT'` and `exitCode: 1`, with the configuration already written correctly. It is the last stage of the flow, so the user sees the whole run end in red.

`shell: process.platform === 'win32'` (or resolving `npx.cmd`) closes it.

core-web/libs/sdk/cli/src/commands/agent/skills.ts
```diff
@@ -0,0 +1,44 @@
...
+ * public, so nothing here needs credentials.
+ */
+export async function installSkills(args: {
+    agentIds: string[];
+    global: boolean;
+}): Promise<SkillsResult> {
+    const argv = buildSkillsArgs(args.agentIds, args.global);
+    const command = `npx ${argv.join(' ')}`;
+    try {
+        const result = childProcess.spawnSync('npx', argv, { stdio: 'inherit' });
```
nicobytes: Three things on this line:

1. **Windows:** same problem as `connect.ts:41` — `spawnSync('npx')` without a shell fails, so FR-025 never installs anything there.

2. **`result.error` is discarded:** `spawnSync` does not throw; on a spawn failure it returns `{ status: null, error: Error }`. The `catch` never runs and the message comes out as `"skills exited with code null"`, losing the real `spawn npx ENOENT`. With no verbose mode that message is all the developer gets, and "code null" is not actionable (FR-032a).

3. **`stdio: 'inherit'` underneath the spinner:** `setup.ts:310` starts `step('Installing the dotCMS skills')` and the child writes to the same terminal the spinner is repainting. It is the same class of bug you fixed for the prompts in 51266ee (`pausingPort`), but the subprocess was left uncovered and `setup.ts` has no access to `progress.pause()`. Either inject the same seam, or capture the output and print it afterwards.

core-web/libs/sdk/cli/src/shared/config-file.ts
```diff
@@ -0,0 +1,225 @@
...
+                      tree,
+                      args.containerKey,
+                      { [args.entryKey]: args.entry },
+                      unit,
+                      1
+                  );
+    }
+
+    await ensureDir(path.dirname(args.file));
+    await fs.writeFile(args.file, next, 'utf8');
```
nicobytes: There is a window where a freshly created file holding a token lives at `0644` before the `chmod` on the next line. FR-021 requires it to be restricted.

Cheap fix: `fs.writeFile(args.file, next, { encoding: 'utf8', mode: 0o600 })` — `mode` applies at creation; the `chmod` is still needed for pre-existing files. Same case in `toml-target.ts:137`.

core-web/libs/sdk/cli/src/shared/redact.ts
```diff
@@ -0,0 +1,6 @@
+/** First 6 + last 4. The only form in which a token may appear anywhere (FR-022). */
+export function redact(secret: string): string {
```
nicobytes: `redact()` has no production caller (`grep -rn redact src/` only returns this module and its spec). The doc comment claims it is *"the only form in which a token may appear anywhere (FR-022)"*, but nothing enforces it — FR-022 holds today simply because the token is never printed.

The 4 tests give apparent coverage over an invariant that is not in force. Wire it into the summary, or delete it along with its spec?

nicobytes (changes requested):
I went through the whole flow plus the tests (239 sdk-cli / 39 http green, lint green). The structural work is solid: the FR-008a ordering guarantee, parse-to-validate/splice-to-write, and the `registry.format → WRITERS` seam that makes FR-013 real.

Two blockers and four small things, all inline.

**Blockers**
1. A run that configures no editor at all prints `Ready` and exits 0 (`ui.ts`). That contradicts the spec's Edge Case (line 117).
2. `spawn('npx')` without `shell: true` fails on Windows (`connect.ts`, `skills.ts`) → FR-024a fails on every Windows run and FR-025 never installs skills.

**Minor:** inherited stdio underneath the spinner, `spawnSync().error` discarded, `writeFile` + `chmod` as two steps, and `redact.ts` with no callers.

Side note, non-blocking: `nx test sdk-cli` did not run here because `verify-package` (its `dependsOn`) runs `npm pack --dry-run` and this workspace is pnpm-only — a missing `npm` is reported as a packaging failure and blocks all 239 tests.

dotCMS-Machine-User (dismissed):

core-web/libs/sdk/cli/src/shared/ui.ts
fmontes: Fixed in a242f0a4. `allGood` now requires at least one outcome, and 43542753 further requires every outcome to be written/replaced. `setup.ts` also names the selectable ids when nothing is detected.

core-web/libs/sdk/cli/src/commands/agent/setup.spec.ts
fmontes: Inverted in a242f0a4 — it now asserts the warning names the remedy and `--agent`. You were right that it passed by asserting nothing was said.

core-web/libs/sdk/cli/src/commands/agent/connect.ts
fmontes: Fixed in a242f0a4: `shell: process.platform === 'win32'`. Tested by faking `process.platform`, since CI here is not Windows.

core-web/libs/sdk/cli/src/commands/agent/skills.ts
fmontes: Fixed 1 and 2 in a242f0a4 — `shell` on win32, and `result.error` is now reported instead of "code null". Left 3 (stdio under the spinner): `spawnSync` blocks the event loop so ora cannot repaint during the install; the cost is one overwritten line after the child exits, not a hang. Cosmetic, tracked separately.

core-web/libs/sdk/cli/src/shared/config-file.ts
fmontes: Fixed in a242f0a4 — `mode: 0o600` at creation in both writers, `chmod` kept for pre-existing files. Verified end-to-end: `-rw-------`.

core-web/libs/sdk/cli/src/shared/redact.ts
fmontes: Deleted with its spec in a242f0a4. No caller, and no error embeds a token, so FR-022 held by accident while four tests implied enforcement. Tokens are simply never displayed, which is the stronger property.

core-web/libs/sdk/cli/src/commands/agent/targets/toml-target.ts
```diff
@@ -0,0 +1,142 @@
...
+ * The line span of our own tables — `[mcp_servers.dotcms]` and any sub-table of it.
+ *
+ * Returns null when the entry is not present. The span runs from our header to the next header
+ * that is not ours, so anything the developer wrote around it is untouched.
+ */
+function findEntrySpan(
+    lines: string[],
+    containerKey: string
+): { start: number; end: number } | null {
+    const ours = new RegExp(`^\\s*\\[\\s*${containerKey}\\.${ENTRY_KEY}\\s*(\\.[^\\]]+)?\\]`);
```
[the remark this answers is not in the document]
fmontes: Fixed in a242f0a4. Confirmed the failure first: `hasTomlEntry` returned false for the quoted form, so a second table was appended — a duplicate-table error that makes the whole file unparseable. Now matches bare, "quoted", 'quoted' and spaced forms.

core-web/libs/sdk/cli/src/commands/agent/targets/toml-target.ts
```diff
@@ -0,0 +1,142 @@
...
+            next = `${original}${sep}${block}`;
+        }
+    }
+
+    await ensureDir(path.dirname(file));
+    await fs.writeFile(file, next.endsWith('\n') ? next : `${next}\n`, 'utf8');
+    // Report what chmod actually did. The flow used to assert `CAN_RESTRICT` here, which is a
+    // platform constant rather than an observation — the summary claimed a protection nobody
+    // had checked.
+    return { path: file, permissionsApplied: await restrictFile(file), replacedExisting };
```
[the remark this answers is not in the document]
fmontes: Fixed in a242f0a4 — `mode: 0o600` applied at creation in both writers.

core-web/libs/sdk/cli/src/shared/config-file.ts
```diff
@@ -0,0 +1,225 @@
...
+    const existing = raw === null ? {} : parseOrThrow(raw, args.file);
+    const container = (existing[args.containerKey] as Record<string, unknown> | undefined) ?? {};
+    const replacedExisting = Object.prototype.hasOwnProperty.call(container, args.entryKey);
+
+    // A brand-new file has no formatting to preserve, so serialize it outright. An existing
+    // one is EDITED, never rebuilt: only the bytes of our own key change (FR-016).
+    let next: string;
+    const tree = raw === null ? undefined : parseTree(raw, [], PARSE_OPTIONS);
+    if (raw === null || raw.trim() === '' || tree?.type !== 'object') {
+        next = `${JSON.stringify({ [args.containerKey]: { [args.entryKey]: args.entry } }, null, 2)}\n`;
```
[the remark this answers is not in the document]
fmontes: Fixed in a242f0a4. Reproduced it: a file containing `[1,2,3]` came back as `{"mcpServers":…}`. A non-object root is now a named MalformedConfigError and the file is left untouched.

core-web/libs/sdk/cli/src/shared/instance.ts
```diff
@@ -0,0 +1,116 @@
...
+
+/** Trailing slashes only — the rest of the address is the user's to get right. */
+export function normalizeUrl(url: string): string {
+    return url.trim().replace(/\/+$/, '');
+}
+
+export function validateUrl(url: string): void {
+    if (!/^https?:\/\//i.test(url)) throw new InvalidUrlError(url);
+    try {
+        new URL(url);
```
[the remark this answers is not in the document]
fmontes: Fixed in a242f0a4 — good catch, this was the most serious one. Verified the password reached the config files. Now refused via a dedicated `CredentialInUrlError`; the stripping happens inside the error types so no call site can leak it.

core-web/libs/sdk/cli/src/commands/agent/connect.ts
```diff
@@ -0,0 +1,124 @@
...
+                id: 1,
+                method: 'initialize',
+                params: {
+                    protocolVersion: '2024-11-05',
+                    capabilities: {},
+                    clientInfo: { name: 'dotcms', version: '0' }
+                }
+            })}\n`
+        );
+        child.stdin?.write(`${JSON.stringify({ jsonrpc: '2.0', id: 2, method: 'tools/list' })}\n`);
```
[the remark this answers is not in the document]
fmontes: Fixed in a242f0a4 — `initialize` alone, then `notifications/initialized` + `tools/list` once the server answers id 1. No existing test drove the server side of the conversation, which is why it survived; there is one now.

core-web/libs/sdk/cli/src/commands/agent/gitignore.ts
```diff
@@ -0,0 +1,89 @@
...
+
+    const already = new Set(current.split('\n').map((line) => line.trim()));
+    const toAdd = args.files
+        .map((file) => path.relative(root, file).split(path.sep).join('/'))
+        .filter((entry) => !already.has(entry));
+
+    if (toAdd.length) {
+        const prefix = current === '' || current.endsWith('\n') ? '' : '\n';
+        const block = `${prefix}\n# dotCMS agent configuration — contains an access token\n${toAdd.join('\n')}\n`;
+        await fs.writeFile(gitignorePath, current + block, 'utf8');
```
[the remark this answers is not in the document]
fmontes: Fixed in a242f0a4 — now a warning that still names the token-bearing files, rather than an exception that takes the whole summary with it.

core-web/libs/sdk/cli/src/commands/agent/setup.ts
```diff
@@ -0,0 +1,347 @@
...
+    if (explicitTargets.length) {
+        targets = explicitTargets.map(getTarget);
+    } else {
+        // Use the objects detection returned, never re-derive them from the registry by id:
+        // that silently replaces whatever the caller resolved, which is a real bug and not
+        // only a testing inconvenience.
+        const detected = await detectTargets();
+        const detectedById = new Map(detected.map((t) => [t.id as string, t]));
+        if (opts.promptPort) {
+            const chosen = await opts.promptPort.multiSelect(
```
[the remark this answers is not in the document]
fmontes: Your reading of FR-003i was correct, but resolved in the spec rather than the code (43542753): the picker is the wanted behaviour, since which editors to modify is the developer's call. FR-003i now says an unprompted run means supplying targets too; FR-010 says passing `--agent` skips the question. No-terminal behaviour unchanged.

core-web/libs/sdk/cli/src/commands/agent/setup.ts
```diff
@@ -0,0 +1,347 @@
...
+            committedByConvention: plan
+                .filter(({ target }) => target.folderConfigIsCommitted)
+                .map(({ file }) => file),
+            cwd: opts.cwd ?? process.cwd(),
+            confirmExclude: opts.yes ? async () => true : opts.confirmExclude
+        });
+    }
+
+    // 7. Skills — non-fatal by design (FR-026).
+    if (!opts.skipSkills) {
```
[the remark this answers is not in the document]
fmontes: Fixed in 43542753 — the summary now prints `· skills installation skipped (--skip-skills)`.

core-web/libs/sdk/cli/src/commands/agent/setup.ts
```diff
@@ -0,0 +1,347 @@
...
+            for (const o of outcomes) {
+                if (o.result === 'failed') continue;
+                // FR-027: never claim skills landed where the editor is not confirmed to read
+                // them. Driven by the registry rather than a hardcoded editor id.
+                const target = byId.get(o.targetId);
+                o.skillsInstalled = !skills.ok
+                    ? 'no'
+                    : target?.skillsLocationVerified
+                      ? 'yes'
+                      : 'unverified';
```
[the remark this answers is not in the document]
fmontes: Fixed in a242f0a4 — a failed install now warns with the reason and the exact re-run command, which FR-026 requires be handed back.

core-web/libs/sdk/cli/src/commands/agent/setup.ts
```diff
@@ -0,0 +1,347 @@
...
+    // 8. Prove the agent connects (FR-024a).
+    let connection: SetupResult['connection'] = 'skipped';
+    let connectionReason: string | undefined;
+    if (!opts.skipVerify && !opts.skipMcp) {
+        step(
+            'Starting the server to confirm it responds (this can take a minute on a cold npx cache)'
+        );
+        const result = await confirmConnection({ url, token: token.value });
+        connection = result.ok ? 'ok' : 'failed';
+        if (!result.ok) connectionReason = `${result.cause}: ${result.detail}`;
```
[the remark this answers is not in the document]
fmontes: Fixed in 43542753 — "Ready" now requires every outcome to be written or replaced, and says so when a target was left unchanged.

core-web/libs/sdk/create-app/src/uve/configure-uve.ts
```diff
@@ -1,5 +1,6 @@
+import { httpGet, httpPost } from '@dotcms/http';
```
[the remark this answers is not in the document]
fmontes: Correct, and resolved as you suggest: spec.md and contracts/cli-interface.md amended in 43542753 to record the extraction and why (this is the request layer axios was removed over — its Node adapter leaked `Proxy-Authorization` across a redirect). Marked as amended after approval, so #37392 needs re-approval before merge.

fmontes:
@nicobytes — the `nx test sdk-cli` blocker is fixed in a242f0a4. `verify-package` shells to `npm pack` and is a `dependsOn` of `test`, so on a pnpm-only machine a missing `npm` read as a packaging *failure* and blocked all tests. A check that cannot run is now a SKIP. Verified with node present and npm absent: exit 0.

Both blockers and all four minor points are addressed, plus the 11 Copilot findings — replies inline. Two were resolved in the spec rather than the code (the editor picker, and the `@dotcms/http` extraction), so **#37392 needs re-approval before this merges**.

Left deliberately: the skills subprocess writing under the spinner. `spawnSync` blocks the event loop, so ora cannot repaint during the install — the cost is one overwritten line after the child exits, not a hang.

265 sdk-cli · 39 http · 132 create-app · lint clean · verify-package 8/8. Every fix with observable behaviour is mutation-verified.

dotCMS-Machine-User (dismissed):

nicobytes (dismissed):

core-web/libs/sdk/cli/src/shared/errors.ts
```diff
@@ -0,0 +1,135 @@
...
+export class InvalidUrlError extends CliError {
+    constructor(value: string) {
+        // Describing the problem is not enough: with no verbose mode, the message has to
+        // carry the fix too (FR-032a).
+        // Strip ANY scheme, not just http(s): stripping only http(s) turned
+        // "ftp://demo.dotcms.com" into the suggestion "https://ftp://demo.dotcms.com".
+        // Drop any `user:password@` before the value is used at all. A rejected address may
+        // be rejected BECAUSE it carries a credential, and echoing it — even partially — would
+        // put that password on screen (FR-022a).
+        const safe = value.replace(/^([a-z][a-z0-9+.-]*:\/\/)[^/@]*@/i, '$1');
```
nicobytes: **A password containing `@` partially leaks into the message (FR-022a).**

This regex stops at the *first* `@`; the URL spec uses the *last* one as the userinfo delimiter:

```
https://admin:p@ssw0rd@demo.dotcms.com  ->  https://ssw0rd@demo.dotcms.com
```

`ssw0rd` reaches the terminal — in the one code path whose whole reason for existing is that the address was rejected *because* it carried a credential. Same regex, same bug at line 48 in `CredentialInUrlError`.

Build the safe form from the parsed URL instead: `const u = new URL(value); const safe = \`${u.protocol}//${u.host}${u.pathname}\``. At minimum make the userinfo match greedy (`[^/]*@`).

`grep -rn CredentialInUrl src` returns only the definition and the throw site — the FR-022a guarantee is currently unasserted, which is how this got through.

core-web/libs/sdk/cli/src/commands/agent/skills.ts
```diff
@@ -0,0 +1,54 @@
...
+ * public, so nothing here needs credentials.
+ */
+export async function installSkills(args: {
+    agentIds: string[];
+    global: boolean;
+}): Promise<SkillsResult> {
+    const argv = buildSkillsArgs(args.agentIds, args.global);
+    const command = `npx ${argv.join(' ')}`;
+    try {
+        const result = childProcess.spawnSync('npx', argv, {
```
nicobytes, line 40: **The comment above says "No secret is passed" — but `spawnSync` defaults to `env: process.env`.**

So `npx skills add dotCMS/agent-toolkit`, plus every `postinstall` npm runs underneath it, receives `DOTCMS_AUTH_TOKEN` and `DOTCMS_PASSWORD` whenever the developer supplied them by environment — which is what this CLI's own `--password` help text recommends.

The claim is right that no credential is needed here, so make it true:

```ts
const { DOTCMS_AUTH_TOKEN, DOTCMS_PASSWORD, ...safeEnv } = process.env;
childProcess.spawnSync('npx', argv, { stdio: 'inherit', env: safeEnv, shell: ... });
```

`connect.ts:47` spreads the full env too. That child does need `DOTCMS_URL`/`AUTH_TOKEN`, but not `DOTCMS_PASSWORD` — and it is an unpinned `@latest` package by design, so the smaller the environment it sees the better.

core-web/libs/sdk/cli/src/commands/agent/gitignore.ts
```diff
@@ -0,0 +1,89 @@
...
+    const gitignorePath = path.join(root, '.gitignore');
+    let current = '';
+    try {
+        current = await fs.readFile(gitignorePath, 'utf8');
+    } catch {
+        /* no .gitignore yet */
+    }
+
+    const already = new Set(current.split('\n').map((line) => line.trim()));
+    const toAdd = args.files
```
nicobytes: **Unescaped, unanchored patterns can silently fail to exclude the token file.**

`path.relative` output goes in raw, but `.gitignore` entries are globs. A directory whose name contains `[`, `]`, `*`, `?`, `#` or a trailing space produces a pattern that matches nothing — under `~/work/my [wip] app/`, the entry `my [wip] app/.mcp.json` reads `[wip]` as a character class.

The failure mode is the bad one: `excluded: true` is returned, `ui.ts:90` prints `✓ added to .gitignore`, and the file holding a 365-day token stays tracked. That is exactly the false assurance this module's header argues against.

Escape the glob metacharacters and anchor with a leading `/` (today `.mcp.json` matches at any depth). Cheapest real guarantee: confirm with `git check-ignore -q <file>` after writing, before reporting `excluded: true`.

core-web/libs/http/src/lib/http.ts
```diff
@@ -115,7 +113,7 @@ async function request<T>(
 
     const data = await readBody<T>(response);
```
nicobytes: **`timeoutMs` covers the headers only — a trickling body hangs the CLI forever.**

`clearTimeout(timer)` runs in the `finally` the moment `fetch` resolves, i.e. when the headers arrive. This `readBody` then awaits `response.text()` with no abort armed and no size cap, so a host that answers `200` and never finishes the body hangs the process indefinitely — the failure the timeout exists to prevent.

Reachable from a plain typo: `checkReachable` (`sdk/cli/src/shared/instance.ts:51`) is the first request against a user-supplied address and passes `acceptAnyStatus`, so whatever that host returns is read to completion.

Move the `clearTimeout` to after `readBody`, and cap the body length — an unbounded `response.text()` on an arbitrary address is also an easy OOM.

core-web/libs/sdk/cli/src/shared/config-file.ts
```diff
@@ -0,0 +1,235 @@
...
+                      1
+                  );
+    }
+
+    await ensureDir(path.dirname(args.file));
+    // `mode` applies at CREATION. Writing at the umask and restricting afterwards left a
+    // token-bearing file world-readable for the width of that gap, and permanently so if the
+    // process died in between (FR-021). chmod below still matters for a pre-existing file,
+    // whose mode `mode` does not change.
+    await fs.writeFile(args.file, next, { encoding: 'utf8', mode: FILE_MODE });
```
nicobytes: **A non-atomic write can truncate the config file this PR works hard to preserve.**

Everything upstream of this line — parse-to-validate, splice-to-write, `detectIndent`, the byte-for-byte guarantee — is undone here: `fs.writeFile` truncates first, then writes. A crash, a full disk or a `SIGINT` in between leaves the developer's `.mcp.json` empty or half-written, taking every other MCP server they had configured with it. No backup.

Write to a sibling temp file with `mode: 0o600` and `fs.rename` over the target; `rename` is atomic within a filesystem, so a reader sees either the old file or the new one.

That also closes the window lines 228-231 half-acknowledge: for a **pre-existing** file `mode` is ignored, so the token bytes land at the file's old mode (0644 is common for a hand-made `config.toml`) and stay there until line 233's `chmod`. Temp-write + rename makes the token owner-only for its entire life on disk. Same applies to `targets/toml-target.ts:145`.

core-web/libs/sdk/cli/src/shared/instance.ts
```diff
@@ -0,0 +1,127 @@
...
+export interface InstanceInfo {
+    /** Normalized base URL, no trailing slash. */
+    url: string;
+    /** Instance version from /api/v1/appconfiguration, or null when absent/unreadable. */
+    version: string | null;
+}
+
+/** Trailing slashes only — the rest of the address is the user's to get right. */
+export function normalizeUrl(url: string): string {
+    return url.trim().replace(/\/+$/, '');
```
nicobytes, line 19: **The most likely paste — the admin URL — reports "not a valid dotCMS instance".**

`normalizeUrl` strips trailing slashes only and `endpoint()` concatenates, so what a developer copies out of their browser produces:

```
https://demo.dotcms.com/dotAdmin/#/home
  -> https://demo.dotcms.com/dotAdmin/#/home/api/v1/appconfiguration
  -> fetch drops the fragment, requests /dotAdmin/, gets HTML 200
  -> NotADotCmsInstanceError: "... is not a valid dotCMS instance. Check the address."
```

The address is right; the path is the problem. With no verbose mode, that message sends them checking the host. `new URL(url).origin` fixes it outright — or keep the strictness and name the real cause.

core-web/libs/sdk/cli/src/shared/instance.ts
```diff
@@ -0,0 +1,127 @@
...
+}
+
+export function validateUrl(url: string): void {
+    if (!/^https?:\/\//i.test(url)) throw new InvalidUrlError(url);
```
nicobytes, line 47: **A non-loopback `http://` address sends the password and the minted token in the clear, silently.**

Any `http://` host is accepted. For anything but localhost the password goes out in cleartext on `POST /api/v1/authentication/api-token`, the 365-day token comes back the same way, and is then persisted into up to seven config files.

The localhost default (`prompts.ts:26`) makes plain `http` necessary, so blocking it would be wrong — but this is what the non-fatal warnings channel is for. Warn when the scheme is `http` and the host is not `localhost`/`127.0.0.1`/`[::1]`; it renders in the summary next to the ADR-0019 version warning at no extra cost.

core-web/libs/sdk/cli/project.json
```diff
@@ -0,0 +1,99 @@
...
+        },
+        "verify-package": {
+            "executor": "nx:run-commands",
+            "dependsOn": ["build"],
+            "options": {
+                "command": "bash libs/sdk/cli/scripts/verify-package.sh",
+                "cwd": "."
+            },
+            "cache": true,
+            "inputs": [
```
nicobytes, line 92: **`verify-package` restores a cached PASS without looking at the bundle it just built.**

`inputs` is an explicit override, so it replaces Nx's default file set — only `package.json`, `project.json` and the script itself. A source-only change leaves this task's hash untouched.

Reproduced on this branch:

```
$ pnpm nx verify-package sdk-cli          # Cache: 0/3 hit
$ printf '\n// probe\n' >> libs/sdk/cli/src/shared/version.ts
$ pnpm nx verify-package sdk-cli          # Cache: 2/3 hit — build re-ran, verify-package did not
```

So checks 1-4 (shebang, `bin` resolves, `@dotcms/http` inlined, no unused dependency) are asserted against whichever artifact happened to be there first. Adding `{ "dependentTasksOutputFiles": "**/*.js" }` to `inputs` re-ties the check to the artifact it verifies.

core-web/libs/sdk/cli/src/commands/agent/connect.ts
```diff
@@ -0,0 +1,144 @@
...
+    });
+
+    return new Promise<ConnectResult>((resolve) => {
+        let settled = false;
+        const finish = (result: ConnectResult) => {
+            if (settled) return;
+            settled = true;
+            clearTimeout(timer);
+            try {
+                child.kill();
```
nicobytes: **On Windows this kills the shell and leaves the MCP server running with the token in its environment.**

`shell: true` (line 46) is the right fix for the `.cmd` problem, but it means the spawned process is `cmd.exe` and `npx` → `node` are its children. `child.kill()` terminates the shell only, so on a timeout or a failed check the MCP server outlives the CLI — holding `AUTH_TOKEN`, unsupervised, with nothing telling the developer it is there.

`child.stdin` is also never ended. Consider a tree kill on win32 (`taskkill /pid <pid> /T /F`), or at least `child.stdin?.end()` so a well-behaved stdio server shuts itself down when the pipe closes.

nicobytes (commented):
Security & hardening review of the new `dotcms` CLI and `@dotcms/http`. Suite is green locally (`nx run-many -t test -p sdk-cli http` → 265 passed).

Nine findings, ordered by severity. Two are reproduced below rather than asserted (#1, #8).

Deliberate decisions I checked and am **not** flagging: unpinned `@dotcms/mcp-server@latest` (documented deviation from ADR-0019 in `constants.ts:8-14`), orphaned tokens (`spec.md:268`), and the unscoped `dotcms` npm name (already owned by dotCMS, latest 0.2.1 — so the deploy-action fix is correct).

core-web/libs/sdk/cli/src/commands/agent/setup.ts
```diff
@@ -0,0 +1,381 @@
...
+    for (const target of targets) {
+        const file = target.configPath(scope, opts.cwd);
+        if (!file || seen.has(file)) continue;
+        seen.add(file);
+        plan.push({ target, file });
+    }
+
+    // Configuring nothing is not success. Silently exiting 0 here is exactly what the spec's
+    // Edge Cases forbid: say so, and name what the developer can pick.
+    if (!plan.length && !opts.skipMcp) {
```
oidacra: The comment directly above this block says "Configuring nothing is not success. Silently exiting 0 here is exactly what the spec's Edge Cases forbid" — but as far as I can trace, the branch only pushes a warning and never reaches the exit code. `anyFailed` at line 371 reads only `outcomes` (still empty when `plan` is empty) and `connection`, and the connection check needs just url+token, so it can pass on its own. That looks like `exitCode: 0` at line 379 for a run that configured nothing — including the interactive case where someone deselects every editor.

Am I reading that right? If so, would folding it into the aggregation work?

```ts
const noEditorsConfigured = !plan.length && !opts.skipMcp;
const anyFailed = outcomes.some((o) => o.result === 'failed') || connection === 'failed' || noEditorsConfigured;
```

core-web/libs/sdk/cli/src/shared/config-file.ts
```diff
@@ -0,0 +1,235 @@
...
+    return doc;
+}
+
+/** Read and parse, or return null when the file does not exist. Malformed input is a named
+ *  error — never a silent overwrite (FR-018). */
+export async function readJsonDocument(file: string): Promise<Record<string, unknown> | null> {
+    let raw: string;
+    try {
+        raw = await fs.readFile(file, 'utf8');
+    } catch {
```
oidacra: This catch treats every read failure as "the file does not exist yet" — `EACCES`, `EISDIR`, a transient I/O error all become `raw = null`. The same shape is at line 196 in `writeMerged`, where `null` feeds the fresh-file branch and writes a document containing only the `dotcms` entry.

If a config that already holds other MCP servers is momentarily unreadable but still writable (a cloud-sync or AV lock, a permissions race), would that path replace it rather than merge into it? That is the outcome the FR-016 byte-preservation work exists to prevent, reached through "unreadable" rather than through a parse failure.

Would narrowing to `ENOENT` be the right call?

```ts
} catch (error) {
    if ((error as NodeJS.ErrnoException).code !== 'ENOENT') throw error;
    raw = null;
}
```

core-web/libs/sdk/cli/src/commands/agent/gitignore.ts
```diff
@@ -0,0 +1,89 @@
...
+    const proceed = args.confirmExclude ? await args.confirmExclude(args.files) : false;
+    if (!proceed) {
+        return { files: args.files, inRepository: true, excluded: false, warnings };
+    }
+
+    const gitignorePath = path.join(root, '.gitignore');
+    let current = '';
+    try {
+        current = await fs.readFile(gitignorePath, 'utf8');
+    } catch {
```
oidacra: Same "unreadable equals absent" shape as `config-file.ts`, but here there is no parse/validate step downstream to catch it. On a non-`ENOENT` failure `current` stays `''`, and the unconditional write at line 85 (`current + block`) would replace the developer's `.gitignore` with just the token block, dropping every existing pattern — while still returning `excluded: true` with no warning.

Is that reachable in practice? If so, something like:

```ts
} catch (error) {
    if ((error as NodeJS.ErrnoException).code !== 'ENOENT') {
        warnings.push(`Could not read ${gitignorePath} — ${(error as Error).message}. Nothing was changed.`);
        return { files: args.files, inRepository: true, excluded: false, warnings };
    }
}
```

core-web/libs/sdk/cli/src/commands/agent/targets/toml-target.ts
```diff
@@ -0,0 +1,153 @@
...
+ * Is our entry already present?
+ *
+ * Backed by `findEntrySpan`, deliberately: answering this with a TOML object-lookup while the
+ * write located the entry by line span gave two notions of "present" that agreed only by luck.
+ */
+export async function hasTomlEntry(file: string, target: AgentTarget): Promise<boolean> {
+    let raw: string;
+    try {
+        raw = await fs.readFile(file, 'utf8');
+    } catch {
```
oidacra: `hasTomlEntry` returns `false` for any read error, and `writeTomlTarget` has the same catch, leaving `original` as `''`. That reaches `if (original.trim() === '') { next = block; }` at line 131, which emits only the `dotcms` table.

For a Codex `config.toml` that is momentarily unreadable but writable, would that discard the other servers and the hand-written comments this module's splice-not-reparse approach is specifically built to preserve? Narrowing both catches to `ENOENT`, and turning any other code into a named failure, would keep the guarantee intact.

Worth noting this is the third site with this shape (`config-file.ts` lines 60 and 196, `gitignore.ts` line 73) — possibly worth one shared helper that reads a file and distinguishes "absent" from "could not read".

core-web/libs/sdk/cli/src/shared/errors.ts
fmontes: Fixed in fe8b6818. You were right and it was worse than the report: `InvalidUrlError` also *suggested* `https://ssw0rd@demo.dotcms.com` as the correction. Both now build the safe form from the parsed URL. Your point about the guarantee being unasserted was the actual root cause — there are tests on it now.

core-web/libs/sdk/cli/src/commands/agent/skills.ts
fmontes: Fixed in fe8b6818 — a shared `envWithoutSecrets()` strips both from the skills child, and the MCP server child keeps the URL and token it needs but loses `DOTCMS_PASSWORD`. The comment claimed it; now it is true.

core-web/libs/sdk/cli/src/commands/agent/gitignore.ts
fmontes: Fixed in fe8b6818 — patterns are escaped and anchored with a leading `/`, and exclusion is confirmed with `git check-ignore` before `excluded: true` is reported. Doing that surfaced a second problem: the test fixture was a bare `.git` directory, which git cannot answer for, so it is a real `git init` now.

core-web/libs/http/src/lib/http.ts
fmontes: Fixed in fe8b6818 — the timer now spans the body read, and the body is capped at 10 MB enforced as bytes arrive rather than after buffering. The regression test hung for the full 5s vitest timeout before the fix, which was a good demonstration.

core-web/libs/sdk/cli/src/shared/config-file.ts
fmontes: Fixed in fe8b6818 — sibling temp file + `rename`, with a fallback to a direct write where the directory is not writable so a case that used to work still does. That also closes the mode gap you noted: my previous `mode: 0o600` only applied to newly created files.

core-web/libs/sdk/cli/src/shared/instance.ts
fmontes: Fixed in fe8b6818, taking the second option: the error now names the path and suggests the origin. I did not rewrite the address silently — a path can be a legitimate reverse-proxy prefix, and quietly retargeting could point setup at a different instance than the one asked for.

core-web/libs/sdk/cli/src/shared/instance.ts
fmontes: Fixed in fe8b6818 — warns when the scheme is `http` and the host is not loopback, alongside the ADR-0019 warning as you suggested. Not a block, since the default address is `http://localhost:8082`.

core-web/libs/sdk/cli/project.json
fmontes: Fixed in fe8b6818 with `dependentTasksOutputFiles`. Reproduced your scenario before and after — worth noting a comment-only edit still caches, correctly, since esbuild emits an identical bundle; an artifact-changing edit now drops it to 1/3 hit where it was 2/3.

core-web/libs/sdk/cli/src/commands/agent/connect.ts
fmontes: Fixed in fe8b6818 — `taskkill /pid <pid> /T /F` on win32, and `child.stdin?.end()` first so a well-behaved stdio server can shut itself down. Both under test with the platform faked.

core-web/libs/sdk/cli/src/commands/agent/setup.ts
fmontes: You are reading it right — verified: `outcomes: 0, exitCode: 0, warnings: 1`. Fixed in fe8b6818 essentially as you wrote it. I fixed the message last round and left the exit code, which meant the run still reported success. `--skip-mcp` still exits 0, since that no-op is requested.

core-web/libs/sdk/cli/src/shared/config-file.ts
fmontes: Fixed in fe8b6818 — narrowed to `ENOENT`, anything else is a named `UnreadableConfigError` that changes nothing.

core-web/libs/sdk/cli/src/commands/agent/gitignore.ts
fmontes: Fixed in fe8b6818. Agreed this one was the worst of the three: no downstream validation, so the whole `.gitignore` would be replaced while still reporting `excluded: true`. The read now throws, and `setup.ts` already catches around this call and turns it into a warning that still names the token-bearing files.

core-web/libs/sdk/cli/src/commands/agent/targets/toml-target.ts
fmontes: Fixed in fe8b6818, and your suggestion of one shared helper is what I did — `readFileIfPresent` in `config-file.ts` now backs all four sites and distinguishes absent from unreadable. The TOML writer also moved to the same atomic write.

oidacra (approved):

nicobytes (approved):

PR #37418, opened 2026-09-04
Spec: Dojo to Angular dotAI portlet migration (#37417)

### LINKED ISSUE #37417: Dojo to Angular: dotAI Portlet
## Context

The dotAI portlet (Developer Tools → dotAI) is the last dotAI surface still on the Dojo/JSP stack. It renders in the legacy iframe route, so it never picks up the admin shell's theme, breadcrumbs or site selector; it carries no translatable text and no automated tests; and it has drifted visually from every other Developer Tools screen sitting next to it. An installation with custom brand colors gets one screen that ignores them.

This migrates it to Angular following the approved redesign. **No new functionality** — every screen maps onto behavior the product already ships, so there is no backend work beyond portlet registration.

## Current state (Dojo)

| Aspect | Value |
| --- | --- |
| Entry point | Portlet `dotai`, registered as a JSP portlet and rendered in the legacy iframe route |
| UI | One hand-written script over a global state object, bootstrapped by Dojo, plus a stylesheet that re-declares the admin color palette |
| Tabs | Four — search / chat / image, embeddings, config values |
| i18n | None — every string is hardcoded |
| Tests | None |
| Gating | Portlet access via layout; index operations require an administrator role |

## Target state (Angular)

Five tabs — **Search**, **Chat**, **Image**, **Embeddings**, **Config Values** — rendered natively inside the admin shell, inheriting its theme, navigation and site context.

- A shared retrieval-settings panel serving both Search and Chat, so a narrowing set up on one carries to the other.
- A guided dialog for creating an index, adding to one, or deleting from one.
- All user-visible text externalized to the language bundle.
- Every data area handles its loading, empty, error and loaded states distinctly.

Same rollout shape as ES Search (#34733), Query Tool (#34735) and Velocity Playground (#34737): the new screen takes over the existing `dotai` menu entry so upgrades need no manual step, and the previous screen stays reachable at an unlisted `dotai-legacy` address for rollback.

## Gaps this closes

Two defects in the current screen, fixed as part of the migration:

- The "Inner Product" distance option sends a value the server does not recognize, so it silently behaves as cosine.
- The response-length field advertises a minimum of 10 tokens, while the server's own request contract declares a minimum of 128. That contract minimum is not currently enforced server-side, so a smaller value is accepted silently and truncates the answer instead of erroring.

And one dead end the current screen has no state for: a user with portlet access but without the administrator role that index operations require gets an empty index list and an empty index picker, leaving Search and Chat silently unusable with no explanation.

## Out of scope

- Any change to indexing, retrieval, generation or storage behavior.
- App-level AI configuration editing (providers, credentials, connection tests) — that stays on the existing dotAI app configuration screens.
- Three capabilities of the current screen are deliberately dropped and will not be rebuilt: sources listed beside chat answers, the raw structured-response mode, and the recent-image-prompts list. All three remain available on the legacy screen while it exists.

## Definition of Done

The dotAI portlet running in Angular, working, with the improved design.

fmontes:
**Spec-Kit PR 1 of 2 — the spec alone.** No implementation. Review it as a contract, not as code: is this the right problem, scoped right, with criteria a reviewer could check?

Closes nothing yet — implementation lands in PR 2, which will link back here.

## What this is

Dojo → Angular migration of the dotAI portlet (#37417), following the approved redesign. The legacy four tabs become five — Search, Chat, Image, Embeddings, Config Values — with a shared retrieval-settings panel on Search and Chat and a guided index dialog on Embeddings.

**No new functionality.** Every screen maps onto behavior that already ships, so there is no backend work beyond portlet registration.

Rollout follows the ES Search (#34733) and Velocity Playground (#34737) precedent: the new screen takes over the existing `dotai` menu entry so upgrades need no manual step, and the old screen stays reachable at an unlisted `dotai-legacy` address for rollback.

## What's worth your attention

**Three capabilities are dropped on purpose.** Each is behavior a user has today, so each deserves an explicit nod rather than silent acceptance:

- **Chat sources** — only the non-streaming mode carries them, and progressive rendering was chosen instead.
- **The raw structured-response mode** — the provider's own response payload and total time become unreachable from the admin.
- **The recent-image-prompts list** — browser-local, never portable, and could not have carried over to a new storage key regardless.

All three remain on the legacy screen while it exists.

**Two defects get fixed along the way** (FR-023, FR-024): the "Inner Product" distance option sends a value the server does not recognize, so it silently behaves as cosine; and the response-length field advertises a minimum of 10 tokens against a server minimum of 128.

**One dead end gets closed** (US6, FR-049): a user with portlet access but not the administrator role index operations require currently gets an empty index list and an empty picker, leaving Search and Chat silently unusable with no explanation. The approved design has no state for this either.

**One deliberate deviation from the design** (FR-053): the portlet renders in the installation's configured theme color, not the design's specific brand color. That color is customer-configurable at runtime, so hardcoding it would leave a branded admin with one screen in someone else's palette. If the design's color is meant as a new product-wide brand, that's a separate app-wide change.

**One piece of genuinely new behavior** (FR-018): the retrieval-settings panel persists between visits. The legacy screen never stored those controls. Called out rather than smuggled in.

## Contents

`specs/37417-dotai-portlet-rebuild/spec.md` — 7 prioritized user stories, 57 functional requirements, 12 success criteria, edge cases, key entities, Out of Scope, and the dotCMS Legacy Considerations section.

Next: approval here unblocks `/speckit-plan`. It doesn't need to merge first — PR 2 branches off this branch.

### DIFF
--- a/.github/workflows/issue_comp_link-issue-to-pr.yml
+++ b/.github/workflows/issue_comp_link-issue-to-pr.yml
@@ -169,10 +169,26 @@ jobs:
                   echo "is_closing_link=false" >> "$GITHUB_OUTPUT"
 
                 else
-                  echo "No linked issues found in Development section or PR body"
-                  echo "has_linked_issues=false" >> "$GITHUB_OUTPUT"
-                  echo "is_cross_repo=false" >> "$GITHUB_OUTPUT"
-                  echo "cross_repo_owner_repo=" >> "$GITHUB_OUTPUT"
+                  # ── PR title suffix: "... (#N)" ────────────────────────────────
+                  # Spec PRs routinely keep the issue open while still carrying the
+                  # tracker number in the title. Treat the conventional trailing
+                  # "(#123)" as a same-repo non-closing link, but keep it below
+                  # every explicit body/Development signal and above branch fallback.
+                  if [[ "$PR_TITLE" =~ \(\#([0-9]+)\)[[:space:]]*$ ]]; then
+                    title_issue="${BASH_REMATCH[1]}"
+                    echo "Found same-repo issue in PR title suffix: $title_issue"
+                    echo "has_linked_issues=true" >> "$GITHUB_OUTPUT"
+                    echo "linked_issue_number=$title_issue" >> "$GITHUB_OUTPUT"
+                    echo "is_cross_repo=false" >> "$GITHUB_OUTPUT"
+                    echo "cross_repo_owner_repo=" >> "$GITHUB_OUTPUT"
+                    echo "link_method=pr_title_reference" >> "$GITHUB_OUTPUT"
+                    echo "is_closing_link=false" >> "$GITHUB_OUTPUT"
+                  else
+                    echo "No linked issues found in Development section, PR body, or PR title"
+                    echo "has_linked_issues=false" >> "$GITHUB_OUTPUT"
+                    echo "is_cross_repo=false" >> "$GITHUB_OUTPUT"
+                    echo "cross_repo_owner_repo=" >> "$GITHUB_OUTPUT"
+                  fi
                 fi
               fi
             fi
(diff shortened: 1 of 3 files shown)


fmontes:
Spec updated — one correction from the planning phase, pushed as a6583b0.

**FR-023** claimed the server rejects a response length below 128 tokens. It does not. There is no `@Valid` anywhere in `com.dotcms.ai.rest`, so `CompletionsForm`'s `@Min(128)` on `responseLengthTokens` is decorative — and the builder's own default for that field is `0`, which violates its own annotation. The legacy `min="10"` therefore produces a silently truncated answer, not an error.

The requirement is unchanged (the field enforces 128). Only the reason it matters is corrected: the field is the one place the declared limit can be honored.

FR-024 was re-verified and stands as written — `OPERATORS` accepts only `distance`/`cosine`/`innerProduct`, the legacy radio sends `product`, and `getOrDefault` silently falls back to cosine. "Inner Product" has never produced `<#>`.

fmontes:
Implementation is up in #37423, branched off this spec branch as the flow requires — so until this merges, that diff also carries the spec commit.

All 57 functional requirements are implemented and `/speckit-converge` is clean: it found 9 gaps, 7 were built and 2 consciously accepted with the reasoning recorded.

Two corrections landed on this PR while building, both from measuring the running product rather than reading the source — FR-023 (the 128-token minimum is declared but not enforced) and FR-002 (the legacy twin needs a layout, not just a URL). Worth a re-read of those two before approving.

oidacra (dismissed):

oidacra (dismissed):

oidacra (dismissed):

oidacra (approved):

PR #37434, opened 2026-09-07
docs(block-editor): revise the #37340 spec to an editor-only fix

### LINKED ISSUE #37340: Block Editor splits a link into two anchors when ©, ®, or ™ appears inside the linked text

rjvelazco:
Supersedes the spec merged in #37375. `spec.md` only — no implementation.

## Why revise

v1 was implemented on `…-37340-…-impl`: **33 files, ~2.4k lines**, and two acceptance criteria withdrawn mid-implementation. It fixed the symptom downstream rather than the cause — an `emoji` branch and link-run coalescing in five renderer implementations, a shortcode map generated in the frontend build and read from the Java classpath, and a coordinated release of three npm packages.

The cause is four lines of editor configuration.

## What v2 does

* **Closes all five paths that create an `emoji` node.** The issue names one. `:copyright:`, `:)`, shortcode paste and `parseHTML` are the others. The picker needs no change — it already inserts a literal character that `appendTransaction` converts straight back.
* **Keeps the node registered** with `emojis` unfiltered, so stored content still parses (#37145 / #37175 / the `AIContent` precedent).
* **Heals stored nodes into text on the parse path** — `emoji(marks) → text(same marks, character)`. Marks preserved exactly, nothing inferred, no merging.
* **Removes the `has('emoji')` / `isAllowed('emoji')` gates.** `emoji` is not selectable in Allowed Blocks (`getEditorBlockOptions()` offers block nodes only), so they only ever fired on fields restricting something else — silently removing the emoji picker and `:)` from them. A latent defect nobody configured.

Everything lives in `core-web/libs/new-block-editor`. No VTL, no Java, no SDK, no version bump.

## The decision worth your attention

**The heal restores the character, not the link.**

A field saved before the fix comes back with its `©` visible in every renderer — and still as two anchors. Inheriting the neighbouring `link` mark would repair the reported payload, but it would be inferring one author's intent from one payload shape. Elsewhere that shape is deliberate, and a wrong guess is silent, irreversible, and lands on content that never had this bug.

Rejoining the anchors is a normal edit: select the line, reapply the link.

One case rejoins itself — where a link was applied *over* an existing node, the node already carries the mark, so ProseMirror's own normalization joins the run. The plan owes an empirical confirmation of that rather than an assertion.

## Found while writing this

`StoryBlockUtil` counts only `"text"` nodes, and it feeds required-field validation at `ESContentletAPIImpl.java:8180`. **A required Story Block field whose only content is one emoji currently fails to save as "empty."** `getCharCount` has the same blind spot, and so does ES text extraction.

All three resolve themselves once the heal turns those nodes into `text` — with no Java change. They are live defects today, independent of links, and unreported.

## Needs amending on #37340

Three acceptance criteria are superseded, listed in the spec's Assumptions:

1. Pictographic emoji still producing `emoji` nodes.
2. The two renderer gaps being in scope.
3. "Already-split content renders as a single link without requiring a re-save" — half delivered. The character returns with no re-save; the single anchor does not.

Still unconfirmed: the triage precondition that the customer on ticket 39197 authored in the **new** Block Editor.

## Reviewing

The spec is 550 lines. There's a reader's version if it helps: https://claude.ai/code/artifact/04d7687d-f710-40b3-895b-8d89d3b68a67

Per the Spec-Kit flow, this needs **approval, not merge**, before `/speckit-plan` runs. PR 2 carries the implementation.

This PR fixes: #37340

This PR fixes: #37340

### DIFF
--- a/specs/37340-emoji-text-node/spec.md
+++ b/specs/37340-emoji-text-node/spec.md
@@ -1,8 +1,8 @@
-# Issue Resolution Specification: Block Editor emoji conversion splits marked text and drops characters
+# Issue Resolution Specification: Block Editor converts typed characters into `emoji` nodes, splitting marked text
 
 **Feature Branch**: `37340-emoji-text-node`
 
-**Created**: 2026-09-03
+**Created**: 2026-09-03 · **Revised**: 2026-09-07 (v2 — scope reduced to the editor, see [Revision History](#revision-history))
 
 **Status**: Draft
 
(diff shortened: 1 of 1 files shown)


rjvelazco:
> Reply from Claude Code, on @rjvelazco's behalf.

Both Medium findings are correct, and they're the same defect in two places. Fixed in 338baf24bd.

## What happened

Adding the heal criteria renumbered the accessibility ACs from 017/018 to 020/021. The criteria list was updated; two prose references were not:

* **`spec.md:112`** — the Principle-V clarification.
* **`spec.md:495`** — the "test types deliberately not used" table, which also credited AC-016 with a hand-run query it no longer describes.

The diagnosis about *why it matters* is the part worth acting on: because `/speckit-tasks` and `/speckit-implement` map ACs to tasks from those sections, an implementer could have skipped AC-016/017/018 — and AC-017 is one of the two criteria the Verification section names as bounding the heal's risk. A spec that routes its own risk-bounding tests to manual QA is worse than one that never named them.

Both sections now state the manual set is **AC-020 and AC-021 only**, and say explicitly that AC-013 through AC-019 are Jest tests. Audited the whole file afterwards: 21 criteria defined, zero cited-but-undefined, and the other artifacts agree.

## On the non-blocking notes

* **The three empirical checks** — good catch that a spec saying "the plan must confirm" needs the plan to actually carry them. `plan.md` now lists all three as numbered verification tasks rather than one: the ProseMirror self-join, the legacy-editor drop path, and the HTML-string fall-through including the `fallbackImage` variant.
* **The three `#37340` amendments** are still outstanding on the issue. Flagging for @rjvelazco — they're prerequisites to approval, not to merge.

Thanks for verifying the code references rather than taking them on trust. The `StoryBlockUtil` / `ESContentletAPIImpl.java:8180` finding is the one I'd most want a second pair of eyes on, since it's a live defect this spec only fixes as a side effect.

adrianjm-dotCMS:
Why rule out a narrower heal rather than only the general one?

The spec rejects inheriting the `link` mark because it would infer one author's intent from one payload shape. That holds for "inherit from a neighbour" in general — but what about the much narrower signature: a **bare** `emoji` node whose **immediate** neighbours on both sides carry an **identical** `link` mark (same `href`)?

As I read it, the editor can't produce that shape deliberately — the spec itself verifies that applying a link over an existing `emoji` node marks the node too. So the sandwich seems to be exclusive to this defect, not a shape that's meaningful elsewhere.

If that's right, the narrow rule would close #37340's "single link without a re-save" criterion instead of amending it. Was it considered and rejected on other grounds, or is the general objection carrying it?

adrianjm-dotCMS (approved):

rjvelazco:
> Reply from Claude Code, on @rjvelazco's behalf.

@adrianjm-dotCMS Adopted. You found a real error, not just a narrower option.

## The general objection wasn't carrying it alone

`research.md` R8 rejected the narrow rule on two grounds, and the second was false: *"it is exactly the case ProseMirror already handles when the marks genuinely match."* It conflates two shapes.

| Shape | After the heal | ProseMirror joins? |
|---|---|---|
| Link applied **over** a node — node carries `link` | `text(link) + text(link) + text(link)` | **Yes** |
| The sandwich — node is **bare** | `text(link) + text(no marks) + text(link)` | **No** |

Normalization joins adjacent text nodes only when marks already match. In the sandwich the middle node is bare by construction, so nothing joins. R8 credited the library with handling precisely the case the rule existed for.

## Your premise, with one honest caveat

You're right that applying a link over a node marks the node — the spec verifies it — so the sandwich isn't reachable that way. And after this fix no typed symbol becomes a node, so it can't be newly created.

It *is* reachable one way: type `Foo © Bar` (the `©` converts today), then link `Foo` and `Bar` separately to the same URL, deliberately leaving the symbol out. But that requires hitting this defect first, so it lives in exactly the legacy population the heal is repairing, and it needs an author to have linked two runs separately around a symbol they wanted excluded. Rare enough to accept — recorded in R9 rather than glossed, so nobody later reads "impossible" and builds on it.

## What also changed my read

The general rule was rejected partly as *silent and irreversible*. Neither survives at this width — a wrong merge extends the link underline visibly, and it's undone by selecting the run and removing the link. Different class of harm.

## What landed

On `issue-37340-block-editor-emoji-node-impl`, commit fff3351d37:

* **AC-015** — the rule. Bare node, both immediate siblings `text`, `link` marks equal on **every** attribute.
* **AC-016** — every case where it must **not** fire, asserted separately: differing attrs, non-`text` sibling, block boundary, sibling without a link, node carrying its own marks. That test is the actual deliverable here; it's what stops the one permitted inference widening into the general rule.
* **AC-021** — the reported payload now heals to a single `<a>`.
* **R9** — the correction, the counterexample, the harm profile.

The heal still never merges nodes itself; it only decides which marks the healed text carries, and ProseMirror does the joining.

## Net effect on the issue

#37340's *"already-split content renders as a single link without requiring a re-save"* moves from **amend** to **met**. Two ACs there still need amending, not three.

PR #37436, opened 2026-09-07
feat(content-drive): keyboard navigation, selection ranges and a shared shortcut registry

### LINKED ISSUE #32591: [FEAT] CD: Add Keybindings
### User Story

As a content author, I want to use common keybindings to navigate, select and filter content.

### Shortcuts added

| Key | What it does | Where |
|-----|--------------|-------|
| <kbd>/</kbd> | Focus the search field | Content Drive, Asset Picker |
| <kbd>Cmd</kbd>/<kbd>Ctrl</kbd> + <kbd>K</kbd> | Focus the search field (alias for <kbd>/</kbd>) | Content Drive, Asset Picker |
| <kbd>↑</kbd> <kbd>↓</kbd> | Move focus between rows (does not change the selection) | Content Drive, Asset Picker |
| <kbd>Shift</kbd> + <kbd>↑</kbd>/<kbd>↓</kbd> | Grow or shrink the selection from the anchor row | Content Drive |
| <kbd>Shift</kbd> + click | Select everything between the anchor row and the clicked one | Content Drive |
| <kbd>Esc</kbd> | Clear the selection | Content Drive |
| <kbd>Cmd</kbd>/<kbd>Ctrl</kbd> + <kbd>B</kbd> | Show or hide the folder tree | Content Drive |

### Decisions worth knowing

- **<kbd>Esc</kbd> clears the selection only.** An earlier version also wiped every active filter once
  nothing was selected. That put a destructive, hard-to-undo action behind the most-reached-for key on
  the keyboard, so it was pulled. Clearing filters stays on the visible **Clear all** button.
- **<kbd>Cmd</kbd>+<kbd>F</kbd> was rejected** even though it is the obvious "search" key. It would take
  away the browser's own find-in-page, which people rely on. <kbd>/</kbd> is the web convention for
  focusing a search box and collides with nothing.
- **<kbd>Cmd</kbd>+<kbd>K</kbd> is kept as an alias, not the primary.** It conventionally opens a command
  palette, so if we ever want one, that key is still available.
- **Every shortcut stands down while a dialog is over the listing.** Search, Escape and the tree
  toggle all decline when something is stacked above, so a keypress can never act on a listing the
  user cannot see. The Asset Picker is the exception by design: it takes the search key for as long
  as it is open, then hands it straight back.
- **Selections never span pages.** Changing page clears the selection.
- **No <kbd>Delete</kbd> binding.** Content Drive has no plain delete — removal goes through
  permission-dependent workflow actions, so the same key would do nothing on some rows and destroy
  content on others.
- **<kbd>Enter</kbd> and <kbd>Space</kbd> were left alone.** The table already uses them for row selection.

### Bugs fixed along the way

Both were already in the listing before this work and are reachable without touching a shortcut:

- Every row sat in the tab order, so tabbing through the listing took roughly twenty presses to get past.
- After sorting a column, the listing became completely unreachable by keyboard — no row could be
  focused at all.

### QA

Test in Content Drive unless a row says otherwise. Shortcuts are per-surface, so the Asset Picker cases
matter as much as the portlet ones.

**Search**
- [ ] <kbd>/</kbd> focuses the search box and does **not** open the browser's quick-find bar
- [ ] Typing `a/b` into the search box enters a literal slash rather than re-triggering the shortcut
- [ ] <kbd>Cmd</kbd>+<kbd>K</kbd> does the same thing as <kbd>/</kbd>
- [ ] A term already typed is still there after pressing the shortcut again
- [ ] Open the Asset Picker over Content Drive, press <kbd>/</kbd> — the **dialog's** asset search takes
      focus, not the one on the page behind it. Close the dialog and the shortcut goes back to the portlet
- [ ] Open one of Content Drive's **own** dialogs (Action Center, New Folder, the content-type selector,
      Upload) and press <kbd>/</kbd>, then <kbd>Cmd</kbd>+<kbd>K</kbd>. Neither should do anything, and
      focus must **not** jump to the search box hidden behind the dialog. Close it and both work again

**Moving through the listing**
- [ ] <kbd>Tab</kbd> into the listing lands on **one** row, and tabbing again leaves the listing
- [ ] <kbd>↑</kbd>/<kbd>↓</kbd> move focus without selecting anything, and the listing does not scroll away
- [ ] Focus stops at the first and last row of the page — arrows never change page
- [ ] **Sort a column, then press the arrows without clicking a row first.** This is the regression that
      used to kill keyboard access entirely
- [ ] Same check after changing page, changing page size, applying a filter, and running a search

**Selecting**
- [ ] <kbd>Shift</kbd>+<kbd>↓</kbd> grows the selection; <kbd>Shift</kbd>+<kbd>↑</kbd> shrinks it back
      toward where you started rather than growing the other way
- [ ] Pressing an arrow **without** Shift collapses the selection to the focused row
- [ ] <kbd>Shift</kbd>+click a checkbox selects everything between it and the previously selected row
- [ ] Select a range, then run a workflow action — it acts on exactly the rows shown as selected
- [ ] Change page with a selection active: the selection clears

**Escape and the tree**
- [ ] <kbd>Esc</kbd> clears the selection and **leaves the filters and the search term untouched**
- [ ] <kbd>Esc</kbd> with nothing selected does nothing — filters stay
- [ ] With a dialog or the content-editing panel open, <kbd>Esc</kbd> closes that and does not touch the
      selection or filters
- [ ] <kbd>Cmd</kbd>+<kbd>B</kbd> collapses and expands the folder tree
- [ ] With the cursor inside a rich-text field, <kbd>Cmd</kbd>+<kbd>B</kbd> bolds text and does **not**
      toggle the tree

**Should not change**
- [ ] Browser back/forward still walk the folder history
- [ ] Find-in-page (<kbd>Cmd</kbd>+<kbd>F</kbd>) still works normally
- [ ] The read-only listing in the Action Center preview takes no row focus and no shortcut changes its
      selection

### Known gap

Screen readers announce which rows are selected, but the listing is not yet a full ARIA grid (no
cell-level navigation contract). Tracked separately.

### Proposed Objective

User Experience

### Proposed Priority

Priority 3 - Average

zJaaal:
Closes #32591

Adds keyboard navigation and selection to Content Drive's listing, and a reusable shortcut registry so other surfaces inherit the behaviour instead of reinventing it.

## Two of the three acceptance criteria were pre-existing defects

Worth reading first, because it shapes the diff:

- **The listing was a full-page tab trap.** Every row was a tab stop, so Tab walked through all 20 one at a time.
- **After sorting a column, no row was reachable by keyboard at all.** Not even with Tab. The table nulls its selection anchor on sort, and its tab-stop calculation compares that anchor against a row index this component never supplied, so the comparison failed in both directions.
- **Arrow keys did nothing but scroll**, because the table's key handling lives on the row and no row was ever focused.

Reproduction for the second one: load Content Drive, sort any column, then press the arrow keys without clicking a row first.

## What ships

| Combination | Behaviour |
|---|---|
| `↑` `↓` | Move focus between rows. Never changes the selection. |
| `Shift + ↑` `Shift + ↓` | Extend the selection from the anchor. Additive. |
| `Shift + click` on a checkbox | Extend the selection to the clicked row. |
| `/` | Focus the search field. |
| `Mod + K` | Focus the search field. Alias for `/`. |
| `Escape` | Clear the selection. |
| `Mod + B` | Show or hide the folder tree. |

`Mod` is Command on macOS and Control elsewhere, from one registration.

**Two keys changed after product review**, both away from the issue's original wording:

`/` is now the primary search key, with `Mod + K` kept as an alias. `Mod + F` is the one users
actually reach for, and it is deliberately not claimed: taking it removes the browser's
find-in-page. `/` is the web convention for focusing a search box and collides with nothing.
`Mod + K` conventionally opens a *command palette*, so it stays free for one.

That makes a bare printable key a shortcut for the first time, so the registry now ignores a
one-character key pressed with no modifier while focus is in something that takes text. Without it,
typing `/` into the box the shortcut just focused would re-fire the shortcut. Modifier combinations
are unaffected and keep working while typing.

`Escape` now clears the selection and stops. It used to go on to clear every active filter, which
put a destructive, hard-to-undo action behind the most-reached-for key on the keyboard. Clearing
filters stays on the visible **Clear all** control.

**Selection follows the desktop file-manager model**, verified against Finder and Windows Explorer: a range is always anchor-to-focus inclusive, so it can shrink to one row but never to none, and it extends whatever was already selected rather than replacing it.

## The shared shortcut registry

`libs/ui/src/lib/services/dot-keyboard-shortcut/`, exported from `@dotcms/ui`. This is the first keyboard infrastructure in core-web; before it, the only mechanism was a per-component `(document:keydown.x)` host binding, which cannot arbitrate between two surfaces that want the same key.

- **Per-combination last-in-wins.** Each combination keeps its own claim stack; the newest wins and withdrawing it restores the one beneath. A dialog opening over a portlet takes the key and hands it back on close, without swallowing combinations it never claimed.
- **One document listener, bubble phase**, attached lazily and ignoring already-handled events. A component handling a key on its own element is closer to the target and never reaches arbitration, so the registry needs no knowledge of, say, rich text surfaces. **Switching this to capture phase would silently break that.**
- **`preventDefault` only for a claimed combination**, so it can never eat a browser or assistive-technology key it does not own.
- **Batch registration**: one call in, one withdrawal out, however many combinations.
- **Labels are mandatory** and `activeShortcuts()` exposes the live set, so a shortcut cannot ship undocumented.

## Why this touches `libs/edit-content`

The content side panel bound `(document:keydown.escape)` on its own host: a second document-level listener the registry cannot arbitrate with, because `preventDefault` does not stop other listeners on the same node. Once Content Drive claimed Escape, **both** would have fired — the panel closing while the listing behind it cleared its selection. The panel now claims Escape through the registry, and since it opens later it wins for exactly as long as it is up.

Five tests added there, where previously **zero** covered Escape on that panel.

## Testing

Keyboard coverage went from **zero** to covering every shortcut shipped.

| Suite | Before | After |
|---|---|---|
| `ui` | 1443 | 2004 passed, 14 skipped |
| `portlets-content-drive` | 1409 | 1370 / 1370 |
| `edit-content` | 2290 | 2344 passed, 2 skipped |

`ui`, `portlets-content-drive`, `edit-content` and `dotcms-ui-e2e` all lint clean; `nx build dotcms-ui` succeeds.

**Plus 18 Playwright tests** across two suites. They exist because of the three bugs below: a unit test can prove the handler does what you told it to, not that the browser calls it.

- `tests/content-drive/content-drive-keyboard.spec.ts` (13) covers both search keys, the roving tab stop (including after a sort, where the table clears its own anchor), arrow movement, Shift+Arrow ranges growing **and shrinking**, Shift+click ranges, Escape leaving the search term alone, and that no quick-find bar opens on `/`. Folders are seeded through the REST API and torn down per test.
- `tests/edit-content/asset-picker/asset-picker-keyboard.spec.ts` (5) covers the same shortcuts where the picker inherits them: the slash reaches the *asset* search rather than the sidebar's folder search, a slash typed into either box stays a character, and the arrows walk the listing.

**Three bugs got past a green suite** because synthetic events were unlike real ones, all found by hand in the browser. They are written up in `docs/frontend/KEYBOARD_SHORTCUTS.md` so the next person does not repeat them:

- a constructed `MouseEvent` defaults to `cancelable: false`, which silently makes `preventDefault()` a no-op;
- a `click` dispatched without its preceding `mousedown` never carries the modifier;
- holding Shift fires a keydown for Shift itself **before** the arrow, and treating that as an ordinary key reset the range base so shrinking a range could never deselect anything.

## Deliberate exclusions

Each was considered and rejected, with reasons recorded in `spec.md` so they are not relitigated:

- **Cross-page ranges or selection.** A range is bounded by the current page.
- **`Enter` and `Space`.** Both already claimed by the listing's existing row selection.
- **A `Delete` binding.** Removal goes through permission-dependent workflow actions that differ per item, so a Delete key would do nothing on some rows and destroy content on others.
- **A single tab stop in single-selection listings** (the asset selection dialog). The table short-circuits its tab-stop calculation whenever the selection is empty, which is that surface's permanent state, so the anchor this feature drives is never consulted. **That consumer's behaviour is unchanged by this PR, not degraded.** Deferred as a follow-up.
- Type-ahead, cut/copy/paste, up-one-folder, refresh, F2 rename, and an in-app shortcut overlay.

## Known gaps

- **FR-023 is half satisfied.** `docs/` is developer documentation; the author-facing shortcut list belongs on the public documentation site, outside this repo. `activeShortcuts()` exists so that page can be generated rather than hand-maintained.
- **The listing's ARIA is partial, deliberately.** Rows carry `aria-selected`, but the container keeps PrimeNG's `role="table"`, where the property is not fully meaningful. `role="grid"` brings the grid keyboard contract with it — a screen reader enters focus mode on a grid row, and Left/Right, which read cell by cell in browse mode today, would then do nothing. Half a grid reads worse than a table. `aria-multiselectable` was removed for the same reason: `role="table"` does not support it, so it was wiring that only looked connected. Recorded in the template, in `KEYBOARD_SHORTCUTS.md`, and pinned by a test.

**Follow-ups filed:** #37439 (finish the ARIA grid wiring), #37440 (keyboard path for moving items — moving is drag-and-drop only today, an accessibility gap), #37441 (single-selection tab trap).

## Notes for the reviewer

- **Conflicts with #37374**, which refactors `dot-asset-picker-toolbar` and `dot-content-drive-shell`. Six files overlap. That PR is larger and structural; this one is additive, so it should rebase onto it rather than the other way around.
- **Spec and implementation are in one PR** rather than the usual two, by the author's decision.
- **`mod+b` declines while an overlay is above the listing**, matching Escape. A dialog covers the tree, so the toggle rearranged a layout the user could not see; declining also leaves the combination to whatever is on top, which may want it (rich text reads Cmd+B as bold).
- **Extracting the keyboard concern into a `[dotFolderListViewKeyboard]` directive is agreed and deferred**, not declined — see the thread on `dot-folder-list-view.component.ts`. Moving ~250 lines in the same round as these fixes would make the diff much harder to re-read against the review already done.
- An ADR for the registry is proposed separately in `dotCMS/platform-adrs`, since it sets the pattern every future shortcut will follow.

This PR fixes: #32591


core-web/libs/edit-content/src/lib/components/dot-edit-content-side-panel/dot-edit-content-side-panel.component.ts
```diff
@@ -160,6 +168,19 @@ export class DotEditContentSidePanelComponent implements OnDestroy {
         // Give the editor a clean form-bridge slot; restore the previous one on close.
         pushFormBridge();
 
+        // ESC goes through the shared registry rather than a document listener of this component's
+        // own. Two independent document listeners cannot arbitrate: `preventDefault` does not stop
+        // the other one, so the portlet behind would clear its selection at the same moment this
+        // panel closed. The registry gives the key to the most recent claim, and the panel opens
+        // after the portlet, so it wins for exactly as long as it is open.
+        this.#destroyRef.onDestroy(
```
nicobytes: This component already implements `ngOnDestroy` (form-bridge pop + nav release). Registering the Escape claim via a second teardown path (`DestroyRef.onDestroy` in the constructor) splits cleanup across two mechanisms that run at the same time.

`register()` returns the unregister callback — please call it from the existing `ngOnDestroy` next to `popFormBridge()` / `#navController.release()`, so all teardown lives in one place. `DestroyRef` still has to stay for `takeUntilDestroyed` on the `saved$` subscription.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.ts
```diff
@@ -696,6 +708,61 @@ export class DotContentDriveShellComponent {
...
+     *
+     * The filter step deliberately calls the *same* store action the toolbar's "Clear all" control
+     * calls, rather than clearing anything itself. One rule, one code path, and a visible button that
+     * advertises the shortcut. It also inherits that action's semantics for free: filters with
+     * defaults are re-seeded rather than emptied, and the browsed folder is untouched.
+     *
+     * Declines when there is nothing to back out of, so Escape keeps whatever meaning it has
+     * elsewhere instead of being silently swallowed here.
+     */
+    #onEscape(): boolean {
```
nicobytes, line 794: ### Escape clears the selection and the filters while one of the shell's own dialogs is open

`#onEscape` has no notion of an open overlay, and the shell's four `<p-dialog closeOnEscape="true" appendTo="body">` surfaces (Action Center, folder selector, content-type selector, upload — `dot-content-drive-shell.component.html:71,161`) close through PrimeNG's **own** document keydown listener.

I read that listener in the installed PrimeNG 21.1.3 (`primeng/fesm2022/primeng-dialog.mjs:890`): it compares z-indexes and **never looks at `defaultPrevented`**. So both handlers run, and since the registry's listener attaches at shell construction — before any dialog's — the order on Escape is:

1. `#onEscape` runs first. With a selection it calls `setSelectedItems([])`; with none but an active filter it calls `clearFilters()`.
2. PrimeNG then closes the dialog anyway.

Consequences, both user-visible:

- Escape to dismiss the content-type or folder selector **also silently wipes every active filter** — the user pressed Escape to back out of a dialog and lost their filter state.
- Escape on the Action Center wipes the selection it was operating on.

This is precisely the "two independent document listeners cannot arbitrate" problem the description gives as the reason to migrate the side panel's Escape. It just wasn't applied to the shell's own dialogs.

**Suggested fix:** decline (`return false`) when an overlay is stacked above, so the key falls through to PrimeNG and the shell does nothing. The side panel already has exactly this check — `#hasOverlayAbove()` comparing `ZIndexUtils.getCurrent()` against its own container. Gating on that rather than on `$dialogVisible()`/`$uploadModalVisible()` also covers confirm popups, open select panels, and any dialog added later, instead of a list of visibility signals someone has to remember to extend. Since two components now need it, it is worth lifting into one small helper in `@dotcms/ui` rather than a second copy.

Also missing a test: `dot-content-drive-shell.component.spec.ts:381` asserts the clearing path with no awareness of dialog state, and the new `describe('Escape')` block has no "with a dialog open" case.

nicobytes (commented):
Reviewed across correctness, readability, architecture, security and performance. Where I had a hypothesis I ran it rather than asserting it — a probe test on this branch and a read of PrimeNG 21.1.3's own source.

The registry is the strongest part of the diff. Per-combination last-in-wins with an identity splice rather than a pop, lazy attach/detach, `preventDefault` only for a combination something claims, `defaultPrevented` as the nesting rule — and the arbitration cases are genuinely tested, including the one most implementations get wrong (a newer surface must not swallow a combination it never claimed). The `DotSearchInputComponent.focus()` seam is right too: the box keeps working for a host that has no registry at all.

Two things should change before merge. Both are the same shape as the defects this PR set out to fix.

- **Escape acts on the listing while the shell's own dialogs are open** — comment on `#onEscape`. I checked PrimeNG 21.1.3's dialog listener: it ignores `defaultPrevented`, so both handlers fire.
- **The roving tab stop disappears when the list shrinks under the focused row** — comment on `$activeRowIndex`. Reproduced with a probe test on this branch; the listing goes back to being keyboard-unreachable.

One structural point (the list view is now 1404 lines) and a few optional/nit items are inline.

### Notes, no action required

- The registry has no policy for editable targets, and the docs are explicit that nesting is resolved by a nested handler marking the event. That holds for surfaces handling the key in JS (ProseMirror/TipTap's `mod+b` calls `preventDefault`), but not for a binding the *browser* implements natively — `mod+b` in a plain `contenteditable` has no JS handler to mark it, so the registry would claim it. Content Drive's live claims are `escape`/`mod+k`/`mod+b`, and the classic WYSIWYG runs in an iframe whose events never reach this document, so I could not construct a real case in this app today. Worth one line in `KEYBOARD_SHORTCUTS.md` as the known edge of the rule.
- Per the repo's Spec-Kit flow `spec.md` ships and is approved in its own PR before the implementation one. The description already flags this as a deliberate call; noting it only for the record.
- 2840 lines over 24 files is past the size guidance. The natural split is spec / registry+`libs/ui` / Content Drive consumers — but the registry is only meaningfully exercised by its consumers, so I would rather review it whole than stack it. Not asking for a split.
- Nothing security-relevant: no new inputs, no user data crossing a boundary, no secrets.

**Verdict: request changes** on the two items above; the rest is optional.

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
```diff
@@ -603,6 +632,327 @@ export class DotFolderListViewComponent implements OnInit, AfterViewInit, OnDest
...
+        this.$offset();
+
+        this.#seedTabStop();
+    });
+
+    /**
+     * Index of the row that is currently the keyboard tab stop, relative to the rendered page.
+     * Kept in step with the table's anchor so a range extends from where focus actually is.
+     */
+    protected readonly $activeRowIndex = signal(0);
```
nicobytes, line 705: ### The roving tab stop is lost when the row list shrinks under the focused row

`$activeRowIndex` is only ever written by `onRowFocus` and `onCheckboxClick`. When `$items()` changes, `$seedRovingTabStop` calls `#seedTabStop()`, which re-points the table's **anchor** at `$offset()` but leaves `$activeRowIndex` where it was — so the two drift apart, which is the one thing the comments say must not happen. `#applyRovingTabStop`'s write then looks for `index === activeIndex` among the new rows, finds no match, and leaves **every** row at `tabIndex = -1`.

Verified on this branch by adding a probe to the US1 suite:

```ts
it('keeps a tab stop when the list shrinks under the focused row', () => {
    renderRows();
    pressOnRow(3, 'ArrowDown');      // focus moves, tab stop follows
    expect(tabStopCount()).toBe(1);  // passes

    spectator.setInput('items', [mockItems[0], mockItems[1]]);
    spectator.detectChanges();

    expect(tabStopCount()).toBe(1);  // Expected: 1, Received: 0
});
```

Real-world trigger: focus or click a row further down the listing, then type in the search box or change a filter so fewer results come back. Tab no longer reaches the listing at all — the same defect T006 fixes for sort, reachable a different way. T007 misses it only because it never moves focus first, so index 0 still exists in the shorter list.

**Suggested fix:** keep the pair in step where the anchor is already being re-seeded. `$seedRovingTabStop` tracks `$items()` and `$offset()` already, so `this.$activeRowIndex.set(0)` alongside `#seedTabStop()` matches what the anchor does. A `Math.min(activeIndex, rows.length - 1)` clamp in the write callback would also hold the invariant and is cheaper insurance, but resetting is the one that keeps the two in agreement. Worth landing the probe above next to T007.

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
```diff
@@ -603,6 +632,327 @@ export class DotFolderListViewComponent implements OnInit, AfterViewInit, OnDest
...
+     *     to "every row is tabbable" whenever the selection is empty, before it ever consults the
+     *     anchor;
+     *   - the table clears its anchor at the *end* of a sort, after the sort event this component
+     *     listens to has already fired, so re-seeding from `onSort` is immediately overwritten.
+     *
+     * A post-render write is not the same thing as a competing binding: it is the last write in the
+     * frame, so there is nothing to lose a race against. The anchor is still driven alongside it,
+     * because that is what ranges extend from.
+     */
+    #applyRovingTabStop(): void {
```
nicobytes, line 733: **Consider (structural):** this component is now **1404 lines**, and roughly 250 of the 355 added here are one cohesive concern the rest of the file never touches — roving tab stop, focus intent, range base, two manually managed capture listeners, range extension. That is six new pieces of private state and two DOM listeners a reader has to hold before they can follow ordinary rendering, in a file that was already past the point where adding to it should be the default.

It reads like a directive on the table host rather than more surface on the component: `[dotFolderListViewKeyboard]` owning `#focusIntent`, `#rangeBase`, `$activeRowIndex`, the capture listeners and the roving tab-stop write, emitting the extended range for the component to forward through `onSelectionChange`. The component keeps the selection contract it already owns; the keyboard bookkeeping stops competing with it for space, and it becomes testable on its own instead of through the whole table harness — the spec file is 3436 lines.

Not a blocker by itself, and the concern is genuinely well documented where it sits. But it is separable, and decomposing before piling on is the cheaper order.

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
```diff
@@ -603,6 +632,327 @@ export class DotFolderListViewComponent implements OnInit, AfterViewInit, OnDest
...
+        afterEveryRender({
+            write: () => {
+                const table = this.dataTable();
+
+                if (!table) {
+                    return;
+                }
+
+                const interactive = !this.$disabled() && !this.$readOnly();
+                const activeIndex = this.$activeRowIndex();
```
nicobytes, line 743: **Optional:** `afterEveryRender` runs this write after *every* application render, not only when something here changed — including every keystroke in a form in the side panel opened over the listing, each time doing a `querySelectorAll` plus a full row loop. The `row.tabIndex !== tabIndex` guard stops it dirtying the DOM, not running.

`afterRenderEffect({ write: ... })` (Angular 19+; we're on 22.1.0) is the reactive form of exactly this: it re-runs the write only when the signals it reads change — `$activeRowIndex`, `$disabled`, `$readOnly` — which is the intent the doc comment already describes. The cost today is small; the swap is nearly free.

core-web/libs/ui/src/lib/services/dot-keyboard-shortcut/dot-keyboard-shortcut.service.ts
```diff
@@ -0,0 +1,214 @@
...
+        };
+    }
+
+    /**
+     * The shortcut that would currently receive each claimed combination, with its label.
+     *
+     * This is the source for author-facing documentation, which is why a label is required at
+     * registration: a shortcut that is not listed here cannot be documented.
+     */
+    activeShortcuts(): { combination: string; label: string }[] {
```
nicobytes, line 302: **Nit:** `.filter((shortcut) => Boolean(shortcut.label))` cannot drop anything — `label` is required and non-optional on `DotKeyboardShortcut`.

It also quietly contradicts the reason the field is required: its doc comment says a missing label "would silently omit it" from the docs, which is the failure this filter would cause if an empty string ever did get through. Dropping the filter makes the invariant the type already states the only rule.

dotCMS/src/main/webapp/WEB-INF/messages/Language.properties
```diff
@@ -7359,6 +7359,13 @@ content-drive.chip-filter.overflow-label={0} and {1} more
 content-drive.filters.clear-all=Clear all
 content-drive.shared-assets-filter.title=Show Shared Assets
 
+## Content Drive - Keyboard shortcuts (labels are the source for the author-facing docs)
+content-drive.shortcut.search=Focus the search field
```
nicobytes, line 7377: **FYI:** nothing reads these five keys yet — `activeShortcuts()` returns the raw i18n key and no template resolves it — so they land as unreferenced entries until the docs generator the description mentions exists. Fine as forward wiring; worth knowing so nobody "cleans them up".

**Nit:** `dot.asset.picker.shortcut.search` and `edit-content.side-panel.shortcut.close` sit under a `## Content Drive - Keyboard shortcuts` header, ~840 lines from the rest of `dot.asset.picker.*`. Grouping by shortcut-ness rather than by owner makes them hard to find from either direction.

core-web/libs/ui/src/lib/services/dot-keyboard-shortcut/dot-keyboard-shortcut.service.spec.ts
```diff
@@ -0,0 +1,366 @@
...
+    });
+
+    afterEach(() => {
+        registered.forEach((unregister) => unregister());
+        registered = [];
+    });
+
+    describe('registration and dispatch', () => {
+        it('should run the handler for a claimed combination', () => {
+            const handler = jest.fn();
```
nicobytes: This file imports `jest` from `@jest/globals`, so `jest.fn()` is `Mock<UnknownFunction>` and the mock returns `unknown`. That is not assignable to `DotKeyboardShortcut['handler']` (`(event: KeyboardEvent) => boolean | void`).

Type the mocks, e.g. `jest.fn<(event: KeyboardEvent) => boolean | void>()`, or a small helper used across this spec. `jest.fn(() => false)` is already fine because the return type is inferred.

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.html
```diff
@@ -81,28 +81,35 @@
...
+                 own range shortcuts, and broke the tab-stop calculation in both directions (every row
+                 tabbable on load, none reachable after a sort).
+                 `focusin`: keeps the single tab stop on the row the user actually moved to. -->
             <tr
                 class="group transition-colors duration-200"
                 [class.cursor-grabbing]="state.isDragging()"
                 [class.text-gray-900]="state.dragOverRowId() === item.identifier"
                 [class.bg-primary-100]="state.dragOverRowId() === item.identifier"
                 [pSelectableRow]="item"
+                [pSelectableRowIndex]="rowIndex"
```
nicobytes, line 115: ### The keyboard selection is invisible to assistive technology

The rows now take focus and carry a roving tab stop, and Shift+Arrow extends a multi-row selection — but nothing in the markup says a row *is* selected.

I checked what PrimeNG contributes rather than assuming: `SelectableRow`'s host bindings are `class`, `tabindex` and `attr.data-p-selectable-row` and nothing else (`primeng/fesm2022/primeng-table.mjs:4915`), and the table renders `<table role="table">` (`primeng-table.mjs:3182`). There is no `aria-selected` on the row, no `aria-multiselectable` on the table, and no `role="grid"`/`role="row"` anywhere in this template.

So for a screen-reader user the feature this PR ships is: focus moves and the row content is re-announced, with no indication that a selection changed or how large it now is. That is the one user for whom keyboard selection matters most.

The honest complication: `aria-selected` is only meaningful on a row inside `role="grid"`, and switching the table to `grid` brings the full grid keyboard contract (cell-level navigation) with it — not a small change, and not obviously in scope here. The component already passes a `[pt]` config (`$ptConfig()`), so the role and `aria-multiselectable` are reachable without forking the template, and `[attr.aria-selected]` on the `<tr>` is a one-liner given the template already has `item`.

Either do that, or add it to **Known gaps** next to the two follow-ups already listed and file the issue. What shouldn't happen is a PR justified by keyboard accessibility shipping with the AT half unaddressed and unmentioned — the spec's FR list and this review thread are the only places it would ever be noticed.

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
```diff
@@ -603,6 +632,327 @@ export class DotFolderListViewComponent implements OnInit, AfterViewInit, OnDest
...
+     * no modifier state of its own.
+     *
+     * Set from a **capture-phase** listener on this component's host, not a template binding on the
+     * row. Measured: the table's own row keydown handler runs before a template binding on the same
+     * row, so by the time a bubble-phase handler saw the key the focus had already moved and the
+     * anchor had already been reset. Capture is the only phase that reliably runs first.
+     */
+    #focusIntent: DotRowFocusIntent = 'anchor';
+
+    #trackFocusIntent(): void {
```
nicobytes, line 851: Two Angular-specific issues in these eight lines.

**`afterNextRender` is not needed here.** `ElementRef.nativeElement` is available the moment the component is constructed — the host element exists before the instance does. Deferring listener attachment by a render cycle buys nothing, and it costs the indirection of `#trackFocusIntent` plus a `DestroyRef.onDestroy` registered from inside a render callback. Attaching both listeners directly in the constructor removes the method, the render dependency, and the question a reader has to ask about why the timing matters.

**The listeners are attached inside the Angular zone.** This app is still zone-based (`zone.js` in `apps/dotcms-ui/src/polyfills.ts`; no `provideZonelessChangeDetection`), so a listener registered from code running in the zone is patched and every `keydown` and `mousedown` in the listing schedules a change detection pass. Both handlers only write plain class fields — `#focusIntent` and `#rangeBase`, neither of them reactive — so not one of those passes is needed for their sake. And each one re-runs the `afterEveryRender` write below it (`querySelectorAll` plus a full row loop), which compounds with the `afterRenderEffect` note I left on that block.

`inject(NgZone).runOutsideAngular(() => { ... })` around the two `addEventListener` calls is the whole fix. The `(focusin)` template binding stays in the zone, which is correct — that one *does* write a signal.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-search-input/dot-content-drive-search-input.component.ts
```diff
@@ -21,11 +28,44 @@ import { DotContentDriveStore } from '../../../../store/dot-content-drive.store'
...
+    constructor() {
+        const unregister = this.#shortcuts.register({
+            combination: 'mod+k',
+            label: 'content-drive.shortcut.search',
+            handler: () => {
+                const input = this.$searchInput();
+
+                // Declines rather than swallowing the key if the box is not rendered, so the
+                // combination falls through instead of silently doing nothing.
+                if (!input) {
```
nicobytes: **This branch cannot be reached.** `<dot-search-input>` is rendered unconditionally in this component's template — no `@if`, no `@defer` — so `viewChild(DotSearchInputComponent)` always resolves after the first render, and the handler only ever runs on a real keypress, which is necessarily post-render. Same in `dot-asset-picker-toolbar.component.html:2`.

The comment above it is the part that misleads: it describes the key falling through "to the portlet underneath" in a state this component cannot be in, so a future reader will preserve a fallback that protects nothing.

`viewChild.required(DotSearchInputComponent)` states the invariant that actually holds, and the branch and its comment both disappear:

```ts
readonly #searchInput = viewChild.required(DotSearchInputComponent);
// ...
handler: () => {
    this.#searchInput().focus();

    return true;
}
```

If you'd rather keep the defensive branch — `.required` throws rather than returning undefined, which is a real trade — then the comment should say it guards a case that shouldn't happen, not describe a fall-through behaviour as though it were a designed path.

nicobytes (commented):
Second pass, this time purely through an Angular 22 lens — reactivity primitives, DI, lifecycle, zone behaviour, and the repo's own frontend standards. Separate from my earlier five-axis review; no overlap with it.

The signal usage is largely current and the doc comments on the tricky parts are better than most of this codebase. What the Angular lens surfaces is one accessibility gap that goes to the heart of the feature, and a cluster of items where the implementation reaches for a heavier Angular API than the situation needs.

**The one that matters:** the roving tab stop and range selection ship with **no ARIA selection state at all** — comment on the `<tr>`. I checked PrimeNG 21's `SelectableRow` (`primeng-table.mjs:4915`): its host bindings are `class`, `tabindex` and `attr.data-p-selectable-row`, nothing else. The table renders `role="table"` (`primeng-table.mjs:3182`). So a screen-reader user can now move focus through rows and extend a selection while hearing nothing about what is selected. For a PR whose motivation is keyboard accessibility, that needs either the ARIA wiring or an explicit entry in **Known gaps** with a filed follow-up — not silence.

The rest, in descending order: two zone/render-timing items on the capture listeners, a dead branch that `viewChild.required` would delete, and three `private` fields that the repo's TypeScript standard explicitly bans (no ESLint rule catches these, which is why review has to).

### Smaller items, not worth their own thread

- `$activeRowIndex` is a standalone `signal` while the component already keeps its UI state in `state = signalState({ isDragging, languagesMap, dragOverRowId })` (line 583). The tab stop is the same kind of state; `patchState(this.state, { activeRowIndex })` keeps it in one place instead of two mechanisms in one class.
- The class now has two teardown paths: `ngOnDestroy` (line 1068) removes the scroll listener imperatively, `DestroyRef.onDestroy` removes the new capture listeners. `DestroyRef` is the right modern choice — so per the repo's Progressive Enhancement rule the scroll listener is the one to migrate, not the new code to change. Optional, and only if you're already in the file.
- `aria-keyshortcuts` has zero occurrences in `libs/` today. On the search field and the tree-toggle button it is the standard way to make a shortcut discoverable to assistive tech, and it is a cheap partial answer to the FR-023 gap the description already admits. Optional, and arguably its own PR.

Nothing here blocks on its own except the ARIA question, which I'd rather see answered as "recorded gap" than left implicit.

core-web/libs/ui/src/lib/components/dot-search-input/dot-search-input.component.ts
```diff
@@ -61,6 +70,21 @@ export class DotSearchInputComponent {
     /** Emits the trimmed term once the debounce window closes. */
     readonly search = output<string>();
 
+    /** The text field itself, so a host can hand it focus. */
+    private readonly $input = viewChild<ElementRef<HTMLInputElement>>('input');
```
nicobytes, line 77: This violates `docs/frontend/TYPESCRIPT_STANDARDS.md:87` — *"Use the `#` prefix for private class fields. Do not use the `private` keyword for instance properties."* The field is read only in TypeScript, never in the template, so `readonly #input = viewChild(...)` works with no other change.

Same pattern in two more places added by this PR:
- `dot-content-drive-search-input.component.ts:33` — `private readonly $searchInput`
- `dot-asset-picker-toolbar.component.ts:53` — `private readonly $searchInput`

Both of those sit directly beneath a `readonly #shortcuts = inject(...)` the same PR adds, so the two conventions are now adjacent lines in the same class.

Worth flagging rather than waving through because no ESLint rule enforces it — I checked. The standard is only kept by review, and per the repo's style-guide-is-authoritative rule this is the kind of thing that should not need a second conversation.

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
```diff
@@ -603,6 +632,327 @@ export class DotFolderListViewComponent implements OnInit, AfterViewInit, OnDest
...
+     * was `undefined === undefined`, making every row a tab stop (a full-page tab trap), and after a
+     * sort it became `null === undefined`, making *no* row reachable. Both were measured.
+     *
+     * The anchor is cleared by the table on sort and on the value/pagination reset, so re-seeding is
+     * keyed off the rendered rows rather than done once.
+     *
+     * A listing that cannot be interacted with gets no tab stop at all: `null` matches no row index,
+     * which is exactly the state the table itself uses to mean "nothing anchored".
+     */
+    protected readonly $seedRovingTabStop = effect(() => {
```
nicobytes: **Nit, but the file argues my case for me:** `$seedRovingTabStop` holds an `EffectRef`, not a signal. Per `docs/frontend/ANGULAR_STANDARDS.md:36` the `$` prefix means "this is a signal", so a `$`-prefixed `EffectRef` reads as one to anyone scanning the class.

More to the point, `#applyRovingTabStop()` and `#trackFocusIntent()` — 40 lines below — are deliberately *called* rather than assigned, with a comment explaining exactly why a field nothing reads is the wrong shape. That reasoning applies identically here. Two effects in this file are `$`-prefixed fields nothing reads, two render hooks are calls; same problem, two answers.

`$cleanSelectedItems` above is pre-existing so this follows local precedent — I'm not asking for a sweep. But the new one could just as easily follow the pattern the PR itself introduces three declarations later, and then the precedent starts pointing the right way.

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.html
```diff
@@ -81,28 +81,35 @@
...
                 [pSelectableRowDisabled]="$disabled() || $readOnly()"
                 data-testId="item-row"
+                (focusin)="onRowFocus($event)"
```
nicobytes: **Optional:** `focusin` bubbles, and `onRowFocus` already resolves the row itself with `closest(SELECTABLE_ROW_SELECTOR)` — it never uses `event.currentTarget`. So the binding does not have to be per row: one `(focusin)` on the `<p-table>` element (or the component host) would catch every row's focus with a single listener instead of one per rendered row, and the handler needs no change at all.

Small either way at 20 rows, and the per-row binding is arguably more local to read. Mentioning it because the handler is already written in the bubbling style, so the cheaper wiring is free.

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
```diff
@@ -58,6 +62,29 @@ import { SYSTEM_HOST_ID } from '../dot-folder-tree/constants';
  * Canonical position of the "type" column. Extra columns follow it, in the header and in the body
  * alike — read from the constant rather than hardcoded so the two cannot drift.
  */
+/**
+ * Marks a row the table considers selectable. Set by PrimeNG's selectable-row directive, and the only
+ * reliable way to enumerate rendered rows in the order the keyboard walks them.
+ */
+const SELECTABLE_ROW_SELECTOR = '[data-p-selectable-row="true"]';
```
nicobytes, line 76: **FYI, forward-looking:** this attribute is set *statically* by PrimeNG — `'attr.data-p-selectable-row': 'true'` in `SelectableRow`'s host bindings (`primeng-table.mjs:4915`), with no reference to `pSelectableRowDisabled`. So the selector matches disabled rows as well as enabled ones.

Harmless today, because `[pSelectableRowDisabled]="$disabled() || $readOnly()"` is all-or-nothing and `#applyRovingTabStop` covers that case with its own `interactive` check. But the new spec's T010 comment explicitly anticipates per-row unselectable state ("a per-row state will make this observable row-by-row"), and on that day `#rowElements()` would hand a disabled row a `tabIndex` of 0 and include it in the arrow walk order, with `interactive` unable to help because it is a whole-listing flag.

Filtering on `:not([data-p-selectable-row-disabled])` won't work either — PrimeNG doesn't emit such an attribute. Whoever adds per-row disabling will need to reflect it themselves. Worth a line in the constant's doc comment so that lands on the right desk.

dotCMS-Machine-User (dismissed):

core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.ts
zJaaal: Fixed in 3f68784. `#onEscape` now declines while anything is stacked above the listing.

I confirmed your reading of PrimeNG 21.1.3 independently: `bindDocumentEscapeListener` listens on `ownerDocument`, gates only on z-index, and never consults `defaultPrevented`. Both handlers fire, and ours first because the registry attaches at shell construction.

Went with the z-index check as you suggested rather than `$dialogVisible()`, and lifted it into `hasOverlayAbove()` in `@dotcms/ui` so the side panel shares it. One caveat worth knowing, documented on the helper: the stack is a module-level singleton and an overlay torn down without closing leaves its entry behind. This suite reads **1102** with nothing visible once an earlier test has opened a dialog, so tests must mock `getCurrent` rather than trust the ambient value. Not an issue in the app, where overlays revert on close.

Three tests added to the `Escape` block.

— written by Claude, run by @zJaaal

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
zJaaal: Fixed in 3f68784, and thank you for the probe — it reproduced immediately.

`$activeRowIndex` now resets alongside `#seedTabStop()` in the seeding effect, so the pair cannot drift, plus a `Math.min` clamp in the write as a second guarantee. Your test is in as `should keep a tab stop when the list shrinks under the focused row`, next to T007.

— written by Claude, run by @zJaaal

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.html
zJaaal: Fixed in 3f68784. `[attr.aria-selected]` on the row, driven by a `$selectedKeys` set so the template does not run an O(n) scan per row, and `aria-multiselectable` on the table through the existing `$ptConfig()`. Two tests.

Took the partial measure rather than `role="grid"`, for the reason you named: the grid role brings the cell-level keyboard contract with it, which is a much larger change. `aria-selected` outside a grid is imperfect, but it is the difference between hearing nothing about the selection and hearing it change. The full grid semantics stay in Out of Scope with a follow-up to file.

You were right that shipping this silent was the wrong half to leave.

— written by Claude, run by @zJaaal

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
zJaaal: Both fixed in 3f68784. Listeners now attach directly in the constructor — you are right that `afterNextRender` bought nothing, the host element exists before the instance — and both `addEventListener` calls are wrapped in `inject(NgZone).runOutsideAngular`. The `(focusin)` binding stays in the zone, as you said, because it does write a signal.

— written by Claude, run by @zJaaal

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
zJaaal: Tried this and reverted it, with the measurement recorded in the doc comment.

`afterRenderEffect` only re-runs when the signals it reads change, and the case this write exists for changes none of them: a column sort leaves `$activeRowIndex`, `$disabled` and `$readOnly` untouched while the table nulls its own anchor and rewrites every row to `-1`. The write has to reassert unconditionally or the listing goes back to being keyboard-unreachable after a sort. The swap turns that test red on its own.

Kept `afterEveryRender` and added a comment saying why, so the next person does not repeat the swap.

— written by Claude, run by @zJaaal

core-web/libs/ui/src/lib/components/dot-search-input/dot-search-input.component.ts
zJaaal: This one I had to revert, and the standard is the thing that is wrong.

Angular's compiler rejects a signal query on an ES-private field outright:

```
FatalDiagnosticError: Code: 1053, Message: Cannot use "viewChild" on a class member that is
declared as ES private. Update the class field to be either: public, public readonly, protected, private
```

I applied it to all three, which took **31 suites** down, and reverted. Each now carries a comment naming the constraint so nobody retries it.

`TYPESCRIPT_STANDARDS.md:87` is unachievable for `viewChild`, `contentChild` and friends as written. Worth an exception clause there — happy to open that separately rather than fold a standards edit into this PR.

The related `$seedRovingTabStop` naming point was valid and is fixed: it is now a `#seedRovingTabStop()` method called from the constructor, matching the pattern the PR introduces for the other two.

— written by Claude, run by @zJaaal

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-search-input/dot-content-drive-search-input.component.ts
zJaaal: Fixed in 3f68784. `viewChild.required` in both places, and the branch and its misleading comment are gone.

You were right that the comment was the worse half — it described a fall-through the component cannot be in, so it would have been preserved as though it protected something.

— written by Claude, run by @zJaaal

core-web/libs/ui/src/lib/services/dot-keyboard-shortcut/dot-keyboard-shortcut.service.ts
zJaaal: Fixed in 3f68784, filter removed.

Your second point was the sharper one: it contradicted the very reason the field is required. While in there I also corrected the method's contract — `activeShortcuts()` is a runtime snapshot reporting the current winner of each claim stack, not a static manifest, so a combination shadowed by an open dialog reports the dialog's label. Anything generating docs from it has to read it with no dialogs open. That was wrong in the doc comment, the frontend docs and the ADR.

— written by Claude, run by @zJaaal

core-web/libs/edit-content/src/lib/components/dot-edit-content-side-panel/dot-edit-content-side-panel.component.ts
zJaaal: Fixed in 3f68784. The unregister callback is stored and called from `ngOnDestroy` alongside `popFormBridge()` and `#navController.release()`, so all teardown this component owns is in one place. `DestroyRef` stays for `takeUntilDestroyed` on `saved$`, as you noted.

— written by Claude, run by @zJaaal

dotCMS/src/main/webapp/WEB-INF/messages/Language.properties
zJaaal: Both addressed in 3f68784.

Regrouped by owner: `dot.asset.picker.shortcut.search` now sits with the rest of `dot.asset.picker.*`, and the side-panel key moved next to the existing `edit.content.side-panel.*` group — which also made me rename it from `edit-content.` to `edit.content.` to match the prefix already in use there. Only the three `content-drive.*` keys remain under the Content Drive header.

On the "nothing reads these yet" point: correct, and it stays true. `activeShortcuts()` returns the raw key and no template resolves it. The comment on the header says they are the source for the author-facing docs so nobody prunes them.

— written by Claude, run by @zJaaal

core-web/libs/ui/src/lib/services/dot-keyboard-shortcut/dot-keyboard-shortcut.service.spec.ts
zJaaal: Fixed in 3f68784. Added a `handlerMock()` helper typed as `jest.fn<(event: KeyboardEvent) => boolean | void>()` and routed every mock in the file through it.

— written by Claude, run by @zJaaal

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
zJaaal: Added to the constant's doc comment in 3f68784, close to your wording — that PrimeNG sets the attribute statically with no reference to `pSelectableRowDisabled`, that it is harmless only while disabling is all-or-nothing, and that a `:not(...)` filter will not work because PrimeNG emits no disabled attribute, so whoever adds per-row disabling has to reflect it themselves.

Exactly the kind of thing that would otherwise be rediscovered by whoever hits it. Thanks.

— written by Claude, run by @zJaaal

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.html
zJaaal: Taken, in 3f68784. One `(focusin)` on the `<p-table>` instead of one per row; the handler needed no change since it already resolves the row from the event target.

— written by Claude, run by @zJaaal

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
zJaaal: Agreed on the substance, and not doing it in this PR.

The file is 1404 lines and the keyboard concern is genuinely separable — six pieces of private state and two capture listeners that ordinary rendering never touches. A `[dotFolderListViewKeyboard]` directive owning `#focusIntent`, `#rangeBase`, `$activeRowIndex`, the capture listeners and the roving tab-stop write, emitting the range for the component to forward, is the right shape.

My reason for deferring is re-reviewability rather than disagreement: moving ~250 lines into a new directive in the same round as these fixes would make the diff much harder to re-read against the review you have already done, and the correctness changes are the part that needs your eyes. Extracting on top of a landed, tested baseline is also safer than extracting while the behaviour is still moving.

Filing it as a follow-up and referencing this thread. Say the word if you would rather it block.

— written by Claude, run by @zJaaal

core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.ts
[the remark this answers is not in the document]
zJaaal: Fixed in 3f68784.

Your reading was right on both branches: with a dialog open `#onEscape` hit the `selectedItems().length` branch and cleared the selection the dialog was operating on. To answer the open question in your comment — PrimeNG does **not** honour `defaultPrevented`; `bindDocumentEscapeListener` gates only on z-index. So it was the second of your two cases: both fired, the dialog closed *and* the selection cleared.

The shell now declines while anything is stacked above, checked through the shared z-index helper rather than a list of visibility signals, and the missing "Escape with a dialog open" tests are in.

— written by Claude, run by @zJaaal

core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.ts
```diff
@@ -696,6 +709,74 @@ export class DotContentDriveShellComponent {
...
+            this.#shortcuts.register([
+                {
+                    combination: 'escape',
+                    label: 'content-drive.shortcut.escape',
+                    handler: () => this.#onEscape()
+                },
+                {
+                    combination: 'mod+b',
+                    label: 'content-drive.shortcut.toggle-tree',
+                    handler: () => {
```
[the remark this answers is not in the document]
zJaaal: Correct, and fixed in 3cdca3af44.

`mod+b` now goes through `#onToggleTree()`, which opens with the same `hasOverlayAbove()` guard `#onEscape` has. Two reasons rather than one: a dialog covers the tree, so the toggle rearranged a layout the user could not see and they met it changed once the dialog closed; and declining hands the combination to whatever is on top, which may want it — a rich text surface inside a dialog reads Cmd+B as bold.

Three tests in `dot-content-drive-shell.component.spec.ts`: it does nothing while an overlay is above, it resumes once the overlay closes, and it leaves `defaultPrevented` false when it declines (that last one is assertable here in a way it is not for Escape, where PrimeNG's own handler prevents the default itself). Portlet suite 155 → 158, green.

Documented in `docs/frontend/KEYBOARD_SHORTCUTS.md` — the overlay section now covers the non-Escape case explicitly, since the reasoning is not Escape-specific.

— written by Claude, run by @zJaaal

dotCMS-Machine-User (dismissed):

core-web/libs/ui/src/lib/components/dot-asset-picker/components/dot-asset-picker-toolbar/dot-asset-picker-toolbar.component.ts
```diff
@@ -1,11 +1,20 @@
...
+    viewChild
+} from '@angular/core';
 
 import {
     DotCMSBaseTypesContentTypes,
     LOAD_MORE_NODE_TYPE,
     TreeNodeContentData
 } from '@dotcms/dotcms-models';
 
+import { DotKeyboardShortcutService } from '../../../../services/dot-keyboard-shortcut/dot-keyboard-shortcut.service';
```
fmontes, line 18:
```suggestion
import { DotKeyboardShortcutService } from '@services/dot-keyboard-shortcut/dot-keyboard-shortcut.service';
```

fmontes (dismissed):

core-web/libs/ui/src/lib/components/dot-asset-picker/components/dot-asset-picker-toolbar/dot-asset-picker-toolbar.component.ts
zJaaal: Checked this before applying it, and I think the alias would break the build.

`@services/*` maps to `apps/dotcms-ui/src/app/api/services/*` in `tsconfig.base.json`, but `DotKeyboardShortcutService` lives in `libs/ui/src/lib/services/`. So the import wouldn't resolve — and if it ever did, it would be a lib → app dependency, which `@nx/enforce-module-boundaries` (set to `error`) rejects. There are also zero `@services/` imports anywhere under `libs/ui` today; it's an app-only alias.

Within a lib the relative path is the right form: importing `@dotcms/ui` from inside `libs/ui` would be a circular self-import, and `allowCircularSelfDependency` is `false`.

Happy to shorten it another way if the four `../` levels are the real objection, but I'd rather not swap it for something that doesn't resolve. Leaving as-is unless you disagree.

— written by Claude, run by @zJaaal

dotCMS-Machine-User (dismissed):

oidacra (dismissed):

core-web/libs/ui/src/lib/services/dot-keyboard-shortcut/dot-keyboard-shortcut.service.ts
```diff
@@ -0,0 +1,314 @@
...
+    if (event.shiftKey) {
+        modifiers.push('shift');
+    }
+
+    if (event.altKey) {
+        modifiers.push('alt');
+    }
+
+    return [...modifiers, event.key.toLowerCase()].join('+');
+}
```
oidacra, line 43: On a QWERTZ/AZERTY/ES layout `/` is produced with a modifier (Shift+7 on German). `combinationOf()` folds `shift`/`alt` into the combination string unconditionally, with no exemption for a printable character that requires that modifier — so the keypress resolves to `shift+/`, misses the bare `/` claim in the stack, and no handler runs. `event.key` is already `/` at that point; it's only the folded-in modifier that breaks the lookup.

The net effect is that the search shortcut does nothing for those users and `Mod+K` is the only way in — which is worth flagging because the issue chose `/` specifically on the grounds that it "collides with nothing".

Should a single printable-character combination match on `event.key` alone and ignore Shift/Alt in the lookup? Mousetrap and similar libraries compensate for exactly this case.

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
```diff
@@ -83,6 +117,8 @@ const TYPE_COLUMN_ORDER =
 })
 export class DotFolderListViewComponent implements OnInit, AfterViewInit, OnDestroy {
     private readonly renderer = inject(Renderer2);
+    readonly #hostElement = inject(ElementRef);
```
oidacra: `inject(ElementRef)` infers `ElementRef<any>`, which is what forces the `as HTMLElement` cast at line 858. Since this is the component's own injected field — unlike PrimeNG's `Table.el`, where the cast really is unavoidable and the comment says so — would `inject(ElementRef<HTMLElement>)` work here and let the cast go away? That form is already used in several places under `core-web`.

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.spec.ts
```diff
@@ -2979,4 +2979,504 @@ describe('DotFolderListViewComponent', () => {
...
+
+        /**
+         * Drives a range the way the Shift+Arrow handler is planned to: point the table's anchor at
+         * the starting row, then ask it to extend to the target row. Deliberately exercises the
+         * library call the handler will make, rather than a handler that does not exist yet.
+         */
+        const extendRange = (anchorIndex: number, toIndex: number) => {
+            const table = getTable();
+            table.anchorRowIndex = anchorIndex;
+            table.selectRange(new KeyboardEvent('keydown', { shiftKey: true }), toIndex);
```
oidacra: This block drives PrimeNG's `table.selectRange()`, but the shipped range logic in `#extendSelectionTo` (987-1015) never calls it — it computes the range, dedupes against `#rangeBase`, and emits through `onSelectionChange`. So these assertions would still pass if the kept/range dedup regressed.

The JSDoc above reads like this was a deliberate probe of the library from an earlier iteration, before the handler existed — is that right? If the intent is to cover the shipped path, driving Shift+Arrow through the existing `arrow()` helper would exercise it instead.

core-web/apps/dotcms-ui-e2e/src/tests/content-drive/content-drive-keyboard.spec.ts
```diff
@@ -0,0 +1,287 @@
...
+
+/** The one folder each seeding test creates. Derived from the suffix so teardown can find it. */
+const seededFolder = (testSuffix: string) => `cd-keys-${testSuffix}`;
+
+test.describe('Content Drive Keyboard', () => {
+    // Keyed off the per-test suffix rather than a describe-level variable, so this stays correct
+    // under `fullyParallel`. The two tests that seed nothing simply ask for a folder that was never
+    // created, which `deleteFolders` is already best-effort about.
+    test.afterEach(async ({ apiHelpers, testSuffix }) => {
+        const site = await apiHelpers.getDefaultSite();
```
oidacra, line 34: This `afterEach` calls `deleteFolders` with no status check and no try/catch, and `deleteFolders` itself (`core-web/apps/dotcms-ui-e2e/src/requests/folders.ts:22-35`) awaits both deletes blind — unlike `createFolders`, which asserts `status() === 200`. A 500 or 403 on delete is therefore indistinguishable from "already gone", and the helper's own comment warns the folders accumulate across runs until tests start failing. There'd be no CI signal when it happens.

The missing status check is pre-existing shared code, but this suite is what newly exercises it on every test. Assert the status in `deleteFolders` the way `createFolders` does, or would a follow-up issue be the better home for it?

core-web/libs/ui/src/lib/services/dot-keyboard-shortcut/dot-keyboard-shortcut.service.spec.ts
```diff
@@ -0,0 +1,588 @@
...
+            press({ key: 'k' });
+
+            expect(handler).not.toHaveBeenCalled();
+        });
+
+        it('should distinguish a shift-modified combination from the plain one', () => {
+            const plain = handlerMock();
+            const shifted = handlerMock();
+            register({ combination: 'mod+k', label: 'search', handler: plain });
+            register({ combination: 'shift+mod+k', label: 'other', handler: shifted });
```
oidacra, line 99: `normalize()` documents that `shift+mod+k` and `mod+shift+k` resolve to the same claim, and it's what keys the claim-stack map — but only the `shift+mod` ordering appears in this spec; `mod+shift` isn't exercised anywhere.

If normalisation regressed so the two orderings produced different keys, the claims would silently split into separate stacks and last-in-wins arbitration would stop working, with nothing here to catch it. Since this registry is shared infrastructure other surfaces will register against, would you add a test that registers both orderings for the same key and asserts they arbitrate as one claim?

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-search-input/dot-content-drive-search-input.component.spec.ts
```diff
@@ -79,4 +81,131 @@ describe('DotContentDriveSearchInputComponent', () => {
...
+
+        // The modifier form carries no typing guard, so it reaches the handler from anywhere inside
+        // an open dialog. It has to stand down too.
+        it('should decline the alias while an overlay is above the listing', () => {
+            jest.spyOn(ZIndexUtils, 'getCurrent').mockReturnValue(1101);
+            spectator.detectChanges();
+
+            pressModK();
+
+            expect(document.activeElement).not.toBe(input());
```
oidacra, line 176: The `/` version of this test (line 158) asserts both that focus didn't move and that `event.defaultPrevented` is false. The alias version calls `pressModK()` without capturing the returned event, so it only checks focus — even though the comment just above says the modifier form "has to stand down too".

Capture the event and assert `defaultPrevented` here as well, so both halves of the standoff are pinned?

core-web/libs/ui/src/lib/services/dot-keyboard-shortcut/dot-keyboard-shortcut.service.ts
```diff
@@ -0,0 +1,314 @@
...
+        const stack = this.#claims.get(combinationOf(event));
+
+        if (!stack?.length) {
+            return;
+        }
+
+        // Newest first. A handler returning `false` declines and the key falls through; if every
+        // claimant declines, the browser default stands untouched.
+        for (let index = stack.length - 1; index >= 0; index--) {
+            if (stack[index].handler(event) !== false) {
```
oidacra: `handler(event)` is called here without a try/catch. Since this is the single root service arbitrating every shortcut in the application, a consumer handler that throws would skip the `preventDefault()` below (letting the browser default through), deny the claims stacked beneath it their turn, and log nothing — the user would see a dead key with no diagnostic.

Is letting it propagate the deliberate choice, on the grounds that a throwing handler is a programming bug that should surface loudly? If not, a try/catch that `console.error`s the shortcut's label and continues to the next claimant would contain it.

oidacra:
## Review notes that don't attach to a single line

Three things from a multi-agent pass over this PR. The code itself came out well — the registry's documented contract held up under adversarial tracing on every point checked (per-combination last-in-wins with correct out-of-order withdrawal, listener attach/detach lifecycle, `preventDefault` scoped to claimed combinations only, atomic batch registration), and a dedicated pass over the added comments found no inaccuracies across eight load-bearing architectural claims. The items below are the ones that need a decision rather than a fix.

### AC27 contradicts itself inside this PR's own spec

The issue's QA checklist says "Pressing an arrow without Shift collapses the selection to the focused row". The shipped code does not do that: `#onKeydownCapture` sets the intent to `anchor` and clears `#rangeBase`, then `onRowFocus` moves only the tab stop and the anchor. The only two selection-emitting paths, `#extendSelectionTo` and `onCheckboxClick`'s Shift branch, are both Shift-gated, so no collapse path exists.

That is almost certainly the right behaviour — but the acceptance criteria disagree with each other, and so does this PR's own spec:

- The issue's shortcuts table says arrows "do not change the selection".
- `specs/32591-content-drive-keybindings/spec.md` FR-002 and FR-008 say focus and anchor only, no selection change.
- The same `spec.md`, User Story 2 Acceptance Scenario 4, says "the selection collapses to the newly focused row".
- The PR body, `docs/frontend/KEYBOARD_SHORTCUTS.md`, the unit tests (`should not extend when the arrow is pressed without shift`, `should not change the selection when moving focus`) and the e2e tests all side with FR-002.

So the implementation is internally consistent and only AS-4 plus the QA checkbox are out of step. Since `spec.md` ships in this PR, could FR-002 and AS-4 be reconciled here — dropping or amending AS-4 and the QA line — rather than leaving a normative requirement contradicting an acceptance scenario in the same document? If the desktop-file-manager collapse semantic is actually what product wants, that's a small addition to the `anchor` branch instead.

### Three acceptance criteria are covered by mechanism but not by test

Not blockers, just so they're recorded:

- CD's own dialogs declining `/` and `Mod+K` (Action Center, New Folder, content-type selector, Upload). Covered generically via a mocked `ZIndexUtils`, but no test opens any of the four named dialogs, and none of the 18 Playwright tests cover it. This is the AC most likely to break unnoticed, since each dialog stacks differently.
- "Arrows move focus and the listing does not scroll away." The "no selection change" half is well covered; the "does not scroll away" half is asserted nowhere and only claimed in comments.
- "Cursor in a rich-text field: `Mod+B` bolds text and does not toggle the tree." The bubble-phase plus `defaultPrevented` mechanism is real and I traced it through to TipTap's ProseMirror keymap, which does call `preventDefault()` on a successful Bold before the event reaches the registry's document listener. But every test simulates that with a synthetic already-prevented event; nothing drives the actual editor.

### The note about #37374 is now stale

The reviewer notes say this PR conflicts with #37374 and "should rebase onto it rather than the other way around". #37374 merged on 2026-09-07, and its merge commit `8f74a97` is already an ancestor of this branch, so the overlap is resolved and no rebase onto it is outstanding. Worth striking from the description so a reviewer doesn't go looking for a conflict that isn't there.

dotCMS-Machine-User (dismissed):

core-web/libs/ui/src/lib/services/dot-keyboard-shortcut/dot-keyboard-shortcut.service.ts
zJaaal: Fixed in `5160848019`, and thank you — this was the most consequential finding on the PR. `/` did nothing at all on QWERTZ, Spanish or AZERTY, and the reasoning it was chosen on ("collides with nothing") only ever held for the layout it was picked and tested on.

Took your suggestion. `lookupKeysFor` now tries the exact combination first and then, for a single-character key, the character alone. Command and Control are never dropped, so `Mod + /` stays distinct from `/`; trying the exact combination first keeps `shift+/` expressible for anything that wants to distinguish it.

AltGr is recorded as a known gap in the doc comment: on Windows it reports as Ctrl+Alt, indistinguishable from a real one, so characters typed that way keep their modifiers. Nothing shipped is produced that way.

Six unit tests, including the three guards that must keep *not* matching (a command modifier does not fall through, an explicit `shift+/` still wins, a named key like Escape still needs its modifier), and one e2e driven through CDP — `keyboard.press` maps through a US layout and cannot express "the `/` character with Shift held", while CDP still sends a real browser input event.

`docs/frontend/KEYBOARD_SHORTCUTS.md` gained a **Keyboard layouts** section, since the lesson generalises past this one key.

— written by Claude, run by @zJaaal

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
zJaaal: Done in `555cbe8217` — `inject<ElementRef<HTMLElement>>(ElementRef)`, and the cast at the old line 858 is gone. You were right that it only existed because the injection inferred `any`.

Left PrimeNG's `Table.el` casts alone for the reason you gave: that type is not ours to fix.

— written by Claude, run by @zJaaal

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.spec.ts
zJaaal: You read it exactly right, fixed in `555cbe8217`.

Your guess about the history is correct too, and the block's own comment gave it away: *"Deliberately exercises the library call the handler will make, rather than a handler that does not exist yet."* True when written, stale the moment `#extendSelectionTo` landed — and since the shipped path never calls `selectRange`, those assertions would have passed with the range dedup wholly regressed.

`extendRange` now does what you suggested: focus the anchor row, send the Shift keydown a browser sends before the arrow, then step with Shift held. So the block fails when *our* range logic breaks rather than when PrimeNG's does.

— written by Claude, run by @zJaaal

core-web/apps/dotcms-ui-e2e/src/tests/content-drive/content-drive-keyboard.spec.ts
zJaaal: Agreed on the substance, and taking the follow-up issue option you offered.

To restate it so it survives the move: `deleteFolders` awaits both deletes without checking status, unlike `createFolders` which asserts `status() === 200`, so a 500 or 403 is indistinguishable from "already gone" and the folders accumulate with no CI signal until tests start failing.

@zJaaal is picking this up on a separate ticket rather than widening this PR into shared e2e request helpers.

— written by Claude, run by @zJaaal

core-web/libs/ui/src/lib/services/dot-keyboard-shortcut/dot-keyboard-shortcut.service.spec.ts
zJaaal: Added in `555cbe8217`. Three tests: both modifier orderings resolving to one claim, withdrawal of the newer restoring the shadowed one, and surrounding whitespace/case.

Your reasoning is what sold it — the split-stack failure is silent, and this is shared infrastructure other surfaces will register against, so the invariant should not rest on one spelling happening to be the one somebody wrote a test for.

— written by Claude, run by @zJaaal

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-search-input/dot-content-drive-search-input.component.spec.ts
zJaaal: Fixed in `555cbe8217`. My gap: the comment above it claimed both halves were pinned while the test only checked one.

— written by Claude, run by @zJaaal

core-web/libs/ui/src/lib/services/dot-keyboard-shortcut/dot-keyboard-shortcut.service.ts
zJaaal: Not deliberate, and your instinct was right. Fixed in `555cbe8217`.

A throw now counts as a decline, so the next claimant gets the key, and it is reported with the shortcut's label and combination via `console.error` rather than swallowed. Four tests, including that a later press still works — that was the part that worried me most about leaving it as it was.

Your framing of the failure is what settled it: skipping `preventDefault`, denying every claim beneath it, and logging nothing adds up to a dead key with no diagnostic anywhere, which is a bad trade for a root service arbitrating every shortcut in the application.

— written by Claude, run by @zJaaal

core-web/libs/ui/src/lib/services/dot-keyboard-shortcut/dot-keyboard-shortcut.service.ts
```diff
@@ -0,0 +1,342 @@
...
+ * A single character therefore also matches on the character alone, with the exact combination
+ * tried first so `shift+/` stays expressible.
+ *
+ * Command and Control are never layout mechanics and are never dropped: `Mod + /` stays distinct.
+ * AltGr is the known gap — it reports as Ctrl+Alt on Windows, indistinguishable from a real one, so
+ * characters typed that way keep their modifiers. Nothing shipped today is produced that way.
+ */
+function lookupKeysFor(event: KeyboardEvent): string[] {
+    const exact = combinationOf(event);
+    const character = event.key.toLowerCase();
```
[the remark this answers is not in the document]
zJaaal: Real bug, fixed in `f3b9c23824`, but from the other side.

The two rules had drifted apart: `lookupKeysFor` treats Alt as something a layout can require to produce a character, while `isTyping` still exempted Alt. So an Alt-typed character in a text field was not counted as typing *and* matched the bare claim, and got stolen out of the field.

Applying the suggested fix would close it by dropping Alt from the layout rule, which reintroduces exactly the class of bug the layout fix was for, just for Alt-requiring layouts instead of Shift-requiring ones. It also contradicts a test written deliberately for that case.

`isTyping` no longer exempts Alt instead, so both rules ask the same question: did this produce a character, with no Command or Control held. Two tests, and the existing Alt layout case still passes, which is what shows the two sides no longer cancel out.

— written by Claude, run by @zJaaal

oidacra (approved):

Issue #35930, opened 2026-06-04
Migrate Angular 21 → Angular 22 (TypeScript 6) across the core-web Nx workspace


nicobytes:
## Description

Upgrade the `core-web` Nx workspace from **Angular 21 to Angular 22**, including the bump to **TypeScript 6**. The workspace is currently on Angular 21, so this work is unblocked.

Scope is the **entire workspace**: the `dotcms-ui` app, all internal libraries, and the published `@dotcms/*` SDK packages.

Angular migrations are driven through the standard Nx flow:

```bash
# 1. Confirm the Nx CLI version
nvm use
nx --version

# 2. Let Nx update package.json + migrations.json to the latest compatible versions
nx migrate latest

# 3. Install the updated dependencies
pnpm install

# 4. Run the generated migrations
nx migrate --run-migrations
```

After running the schematics, validate that the Angular automated migrations applied cleanly:
https://angular.dev/reference/migrations

### TypeScript 6 note

Angular 22 moves to TypeScript 6. To smooth the transition over removed/renamed APIs, the temporary `ignoreDeprecations` flag may be required in the relevant `tsconfig`:

```json
{ "compilerOptions": { "ignoreDeprecations": "6.0" } }
```

This should be treated as a transitional measure and removed once the deprecated usages are migrated.

### Reference material

| Resource | Link |
|---|---|
| Nx changelog | https://nx.dev/changelog |
| Angular v22 announcement | https://blog.angular.dev/announcing-angular-v22-c52bb83a4664 |
| Angular CHANGELOG | https://github.com/angular/angular/blob/main/CHANGELOG.md |
| Angular migrations reference | https://angular.dev/reference/migrations |

## Acceptance Criteria

**Preconditions / setup**
- [ ] Confirm the workspace is on Angular 21 before starting (`nx report`)
- [ ] Nx is migrated to the latest version compatible with Angular 22 via `nx migrate latest`

**Migration (behavior unchanged)**
- [ ] All Angular packages in `package.json` (app, internal libs, and published `@dotcms/*` SDKs) are on Angular 22.x
- [ ] TypeScript is upgraded to 6.x and the project compiles (`tsc` / Nx type-check passes)
- [ ] `yarn nx migrate --run-migrations` completes with all schematics applied, no pending entries left in `migrations.json`
- [ ] Angular automated migrations report success per https://angular.dev/reference/migrations (no skipped/failed migrations)

**Verification — no regressions**
[...]

PR #37437, opened 2026-09-07
docs(edit-content): spec for workflow "Allow Comments" dialog in new Edit Mode (#36883)

### LINKED ISSUE #36883: Workflow "Allow Comments" dialog not showing up in new Edit Mode
## Problem Statement

In the new (Angular) Edit Content UI, executing a workflow action that has **Allow Comments** enabled does not show the comment dialog on its first execution. The contentlet is saved/transitioned correctly, but the author is given no opportunity to enter the comment — the dialog only appears if the same action is triggered a **second** time. In the legacy (classic) Edit Content UI, the dialog appears correctly on the first execution.

This was discovered while manually QA-ing #35930 (Angular 21→22 migration). It has been confirmed to also reproduce on current `main`, independent of that migration — this is a pre-existing defect in the new edit content workflow, not a regression introduced by the Angular version bump.

Browser/OS: not specified by reporter.

## Steps to Reproduce

1. Create a test Content Type with a single text field for the title. Do **not** enable the new edit mode.
2. Go to the System Workflow and, for the `Publish` action, check the **Allow Comments** box.
3. Create a Contentlet of that test type and click `Publish`. Enter any value in the comment box and click `Send`. The dialog appears as expected and everything works.
4. Go back to the test Content Type definition and enable the new (Angular) edit mode.
5. Create another Contentlet of that type and click `Publish`.
6. **Actual:** the Contentlet is published, but the comment dialog never shows up. The dialog only appears after clicking `Publish` a second time.
7. **Expected:** the comment dialog should appear on the first click, exactly as it does in the legacy edit mode.

## Acceptance Criteria

- [ ] When a workflow action with **Allow Comments** enabled is executed from a Content Type using the new (Angular) edit mode, the comment dialog appears on the first execution — matching current behavior in the legacy edit mode.
- [ ] The workflow action does not complete (contentlet is not saved/transitioned) until the user submits or cancels the comment dialog, matching legacy edit mode gating behavior.
- [ ] Triggering the action a second time (to work around the missing dialog) does not create a duplicate workflow execution, extra version, or duplicate history/audit entries beyond a single correct execution.
- [ ] Verified against at least one other workflow action with **Allow Comments** enabled, to confirm the fix isn't scoped to the System Workflow's default `Publish` action only.
- [ ] Regression test added (e2e or component) covering: new edit mode + an action with **Allow Comments** enabled → dialog appears on first click.
- [ ] Legacy edit mode behavior is unchanged.

## dotCMS Version

Latest from `main` branch / local dev environment. Confirmed to reproduce independent of the Angular 21→22 migration in progress under #35930 (i.e., also present on `main` before that migration).

## Severity

## Links

NA

adrianjm-dotCMS:

Resolves the spec phase of #36883.

### Proposed Changes
* `spec.md` — defect-framed issue-resolution spec: problem/impact, reproduction, scope of investigation, root-cause hypothesis, bounded fix scope with explicit non-goals, regression-risk/blast-radius analysis, and 8 acceptance criteria with a per-layer verification plan. No open clarifications remaining.

### Summary

In the new (Angular) Edit Content UI, firing a workflow action with **Allow Comments** enabled skips the input dialog on its first execution. The action runs to completion — the contentlet is saved and transitioned — but the author never gets to enter the comment. The dialog only appears if the same action is fired a second time. Legacy edit mode is correct.

The comment is the audit/hand-off note recorded in the workflow task history, so this silently degrades the audit trail with no error surfaced.

### Root-cause hypothesis

Two REST shapes describe a workflow action, and only one carries `actionInputs[]`:

| Endpoint | Returns | Has `actionInputs[]`? |
|---|---|---|
| `GET /api/v1/workflow/contentlet/{inode}/actions` (`getByInode`) | `WorkflowActionView` | ✅ |
| `GET /api/v1/workflow/initialactions/contenttype/{id}` (`getDefaultActions`) | `WorkflowDefaultActionView` | ❌ |
| `GET /api/v1/workflow/defaultactions/contenttype/{id}` (`getWorkFlowActions`) | `WorkflowDefaultActionView` | ❌ |

`WorkflowDefaultActionView` is a plain `{scheme, action, firstStep}` wrapper over the raw `WorkflowAction`; `createActionInputViews` is only ever called from `convertToWorkflowActionView` (`WorkflowResource.java:892`).

For **unsaved** content the edit-content store seeds its action list from those input-less defaults, so `fireWorkflowAction` hits `if (!actionInputs.length)` (`dot-edit-content-form.component.ts:551`) and fires straight through with no wizard. There are **three** such seeding points, all in `libs/edit-content`:

| # | Location | When |
|---|---|---|
| 1 | `content.feature.ts:211-218` (`initializeNewContent`) | New content, **single scheme** — **the reported repro's path** |
| 2 | `workflow.feature.ts:192-197` (`setSelectedWorkflow`) | New content, **multi-scheme** — author picks in the sidebar |
| 3 | `locales.feature.ts:268-273` | New translation of an untranslated locale | After that first fire the store re-fetches via `getByInode(..., EDITING)` (`workflow.feature.ts:255`), which *does* carry `actionInputs` — hence the dialog on the second click. Existing content opened directly in the new editor is fine, because `initializeExistingContent` already loads via `getByInode`.

Legacy edit mode is unaffected because it reads the raw `commentable` flag, not `actionInputs`.

### ⚠️ Scope wider than the issue describes

The issue describes one reproduction path; there are **three** (table above), all sharing one input-less source and all closed by one fix. Path 3 (new translation) is not mentioned in the issue at all.

**Push Publish and Assign to** go through the same `actionInputs` check on unsaved content. Assign-to is now AC-006. Push Publish is deliberately **not** an AC — it is reachable as an initial action only if an administrator places it in the workflow's first step, and that was not verified.

### ✅ Corrections after review (@nicobytes, Q1–Q5)

Two of the five questions found real errors in the spec. Both were verified against the code before correcting; details in the new `## Clarifications` section.

- **Q1 — the root cause named the wrong path.** The spec pointed at `setSelectedWorkflow`, which the reported repro *never reaches*: with a single scheme, `initializeNewContent` auto-selects it and seeds the actions there. `setSelectedWorkflow` has exactly one caller — the sidebar's `(onSelectWorkflow)`. Corrected, and the **primary store test moves to `content.feature.spec.ts`**.
- **Q2 — the post-Reset reproduction claim was wrong.** It was reasoned, not observed. On every fire including Reset, `currentContentActions` is patched from `getByInode` (inputs present); re-picking a scheme on content that now has an inode takes `setSelectedWorkflow`'s `else` branch, which does not overwrite the list. **AC-006 removed**, no replacement — there is no defect on that path.
- **Q3** — Push Publish dropped as an AC (see above); AC-005 split so a failure points at one cause.
- **Q4** — AC-002 now defines what "unchanged" means for content that does not exist yet: nothing persisted, form values survive, a second fire still works, and no saving indicator while the dialog is open.
- **Q5** — frontend-only is the intended default, not an open fork. AC-008 → **AC-009, marked contingent**; integration/Postman marked *not expected* per ADR-0013.

**ACs renumbered 001–009** — note that any approval given before this commit predates the changes.

**Also withdrawn**: a non-goal claiming other consumers (UVE toolbar, Content Drive context menu, bulk actions) might carry the same latent defect. The premise was false — `getDefaultActions` / `getWorkFlowActions` have exactly three call sites, all listed above. No follow-up issue needed.

### Design decision deliberately left to `/speckit-plan` — but likely frontend-only

The spec does **not** pick between the candidate fixes (backend: have the default/initial-action endpoints emit `actionInputs[]`; frontend: derive the inputs from the raw action flags). That's a design call, not a spec gap.

What the spec *does* now record is a verified fact that bears on it: **every flag `createActionInputViews` reads is already serialized** on the raw `WorkflowAction` embedded in `WorkflowDefaultActionView` — `commentable`, `assignable`, `hasPushPublishActionlet`, `hasMoveActionletActionlet` / `hasMoveActionletHasPathActionlet` (all four `@JsonProperty`), plus `nextAssign` and `roleHierarchyForAssign`. None is `@JsonIgnore`. What's missing from the payload is only the *derived* `actionInputs[]` array.

So a **frontend-only fix is viable with no REST contract change at all** — no `@Schema`, no `openapi.yaml` regeneration, no integration/Postman coverage, and no risk to the endpoints' other consumers (content-drive context menu, create-content dialog, legacy edit-content bootstrap, customer integrations). The honest trade-off against it: it duplicates in TypeScript a derivation rule that today lives only in Java.

The spec keeps the guardrail either way: **if** the plan does pick the backend route, the response change must be additive — no existing property renamed, retyped, or dropped.

### Checklist
- [ ] Tests — land in PR 2, written and confirmed failing first (Constitution Principle V): Jest/Spectator specs in `edit-content` (`dot-edit-content-form.component.spec.ts`, `content.feature.spec.ts`, `workflow.feature.spec.ts`, `locales.feature.spec.ts`) plus the derivation helper's own unit spec in `data-access`, plus backend integration/Postman coverage **only if** the endpoints change, plus the Playwright e2e the issue asks for. If e2e proves impractical for this flow, the plan must say so explicitly — silence is not consent.
- [ ] Rollback-safety label — this spec's analysis reads the change as rollback-**safe** (an additive JSON field). dotCMS/core auto-applies rollback labels, so this shouldn't need hand-setting.

### Additional Info

Originally surfaced while manually QA-ing #35930 (Angular 21→22 migration), but confirmed to reproduce on `main` independently of it — a pre-existing defect, not a migration regression.

`/speckit-adr-context` has **not** run yet; it fires automatically as the `before_plan` hook on `/speckit-plan`. No ADR proposed at this stage.

This PR fixes: #36883

This PR fixes: #36883

### DIFF
--- a/specs/36883-workflow-comments-dialog-new-edit/spec.md
+++ b/specs/36883-workflow-comments-dialog-new-edit/spec.md
@@ -123,10 +123,20 @@ already loads its actions via `getByInode`.
 The legacy editor does not consult `actionInputs`; it reads the raw `commentable` / `assignable`
 flags, which is why it is unaffected.
 
+**Verified fact bearing on the fix**: every flag `createActionInputViews` reads is *already*
+serialized on the raw `WorkflowAction` embedded in `WorkflowDefaultActionView` — `commentable`
+(`isCommentable()`), `assignable` (`isAssignable()`), `hasPushPublishActionlet`,
+`hasMoveActionletActionlet` / `hasMoveActionletHasPathActionlet` (all four `@JsonProperty`), plus
+`nextAssign` and `roleHierarchyForAssign`. None is `@JsonIgnore`
+(`com.dotmarketing.portlets.workflows.model.WorkflowAction`). The default/initial-action payload
+therefore already carries everything needed to determine an action's inputs — what is missing is
+only the derived `actionInputs[]` array itself.
+
 The plan phase will confirm this and choose between the candidate fixes (backend: have the
 default/initial-action endpoints emit `actionInputs`; frontend: derive the inputs from the raw
-action flags, or fetch an input-bearing action list for unsaved content). The trade-offs —
-notably that changing an existing endpoint's response shape is an API-contract change — belong
+action flags, or fetch an input-bearing action list for unsaved content). Given the fact above, a
+frontend-only fix is viable **without any REST contract change**; the trade-off against it is that
+it duplicates in TypeScript a derivation rule that today lives only in Java. Weighing that belongs
 in the plan, not here.
 
 ## Fix Scope & Non-Goals *(mandatory)*
@@ -140,8 +150,10 @@ in the plan, not here.
   legacy gating.
 - The same correctness applies after a **Reset** action, where the author re-selects a scheme and
   the action list is re-seeded from the same source.
-- Whatever contract change this requires between `WorkflowResource`'s default/initial action
-  endpoints and the edit-content store, kept minimal and additive.
+- Whatever contract change this requires, **if any**, between `WorkflowResource`'s
+  default/initial action endpoints and the edit-content store, kept minimal and additive. A
+  frontend-only fix requiring no contract change at all is in scope and, per the Root-Cause
+  Hypothesis, appears achievable.
 - Regression coverage (see Acceptance & Verification).
 
 **Explicitly out of scope / non-goals**:
@@ -175,7 +187,9 @@ in the plan, not here.
   - Push Publish actions on unsaved content go through the same `actionInputs` check
     (`containsPushPublish`), so they are affected by the same bug and by the same fix — the
     "no publish environments configured" notification path must keep working.
-- **Backward compatibility**: The response of a published REST endpoint may change. Any change
+- **Backward compatibility**: Nothing here *requires* a REST change — a frontend-only fix leaves
+  every contract untouched, which is the lower-risk path. Should the plan pick the backend route,
+  the response of a published REST endpoint changes, and then: any change
   must be **additive only** (`actionInputs[]` appears where it previously did not); no existing
   property may be renamed, retyped, or dropped. `@Schema` annotations must match the actual
   return type and `openapi.yaml` must be regenerated and committed alongside. An additive JSON
--- a/specs/36883-workflow-comments-dialog-new-edit/spec.md
+++ b/specs/36883-workflow-comments-dialog-new-edit/spec.md
@@ -1,0 +1,243 @@
+# Issue Resolution Specification: Workflow "Allow Comments" dialog not showing up in new Edit Mode
+
+**Feature Branch**: `36883-workflow-comments-dialog-new-edit`
+
+**Created**: 2026-09-07
+
+**Status**: Draft
+
+**Type**: Issue / Bug Resolution
+
+**Related GitHub Issue**: [#36883](https://github.com/dotCMS/core/issues/36883)
+
+**Input**: User description: "https://github.com/dotCMS/core/issues/36883 — In the new (Angular) Edit Content UI, executing a workflow action that has **Allow Comments** enabled does not show the comment dialog on its first execution. The contentlet is saved/transitioned correctly, but the author is given no opportunity to enter the comment — the dialog only appears if the same action is triggered a second time."
+
+## Problem Statement *(mandatory)*
+
+In the new (Angular) Edit Content UI, firing a workflow action that has **Allow Comments**
+(and/or **Assign to**) enabled skips the input dialog on its first execution. The action runs to
+completion — the contentlet is saved and transitioned — but the author never gets the chance to
+enter the comment that the workflow was explicitly configured to collect. The dialog only shows
+up if the *same* action is fired a second time, which by then means running the action twice.
+
+In the legacy (classic) Edit Content UI the dialog appears correctly on the first execution, so
+the two editors disagree on behavior for the same workflow configuration.
+
+The comment is not merely cosmetic: it is the audit/hand-off note recorded in the workflow task
+history. Losing it silently degrades the workflow audit trail without any error being surfaced.
+
+**Severity / Impact**: Medium. Affects every author using a content type that has the **new edit
+mode** enabled, on any workflow action configured with **Allow Comments** (or **Assign to** /
+**Move** without a preset path — everything that produces an action input). It happens on the
+*first* execution of the action for content that has not yet been saved through that editor
+session, which in practice is the most common case: creating new content. The action itself
+succeeds, so the failure is silent — no error, no warning, just a missing dialog and a missing
+comment.
+
+## Reproduction *(mandatory)*
+
+**Environment**: Latest `main` / local dev environment. Reproduced independently of the Angular
+21→22 migration tracked in #35930 (i.e. it is a pre-existing defect, not a migration regression).
+Browser/OS not specified by the reporter; not believed to be browser-dependent. Requires a
+content type with the **new edit mode** enabled and a workflow action with **Allow Comments**
+checked.
+
+**Steps to Reproduce**:
+
+1. Create a test Content Type with a single text field for the title. Leave the new edit mode
+   **disabled** for now.
+2. Go to **System Workflow** and, for the `Publish` action, check the **Allow Comments** box.
+3. Create a Contentlet of that test type and click `Publish`. Enter any value in the comment box
+   and click `Send`. The dialog appears as expected and everything works (this is the legacy
+   edit mode baseline).
+4. Go back to the test Content Type definition and **enable the new (Angular) edit mode**.
+5. Create another Contentlet of that type and click `Publish`.
+6. Observe: the contentlet is published, but no comment dialog was shown.
+7. Click `Publish` again on the now-saved contentlet. The comment dialog now appears.
+
+**Expected Behavior**: The comment dialog appears on the **first** click of `Publish`, exactly as
+it does in the legacy edit mode. The workflow action does not run until the author submits or
+cancels the dialog.
+
+**Actual Behavior**: The workflow action fires immediately with no dialog and no comment. The
+dialog appears only on a subsequent execution of the same action, once the contentlet has been
+saved and its allowed actions have been re-fetched.
+
+**Reproducibility**: Always, for content that has not yet been saved in the current editor
+session (new content). It also reproduces for existing content after a **Reset** action, because
+resetting clears the selected scheme and the author must re-pick a workflow — which re-populates
+the action list from the same input-less source. Once the content has been saved through the new
+editor, the actions are re-fetched from the per-inode endpoint and the dialog behaves correctly.
+
+## Scope of Investigation *(mandatory)*
+
+- **Affected area**: Workflows + the new (Angular) Edit Content UI. Specifically the workflow
+  action list that the new editor uses to decide whether an action needs to collect input before
+  firing, and the REST endpoints that feed it.
+- **Suspected surface**: Primarily **frontend** — `core-web/libs/edit-content` (the edit-content
+  signal store's workflow/content features and `DotEditContentFormComponent`) plus
+  `core-web/libs/data-access` (`DotWorkflowsActionsService`, `DotWorkflowEventHandlerService`).
+  The root cause originates in a **modern backend** view contract
+  (`com.dotcms.rest.api.v1.workflow.*` — `WorkflowResource`, `WorkflowDefaultActionView`,
+  `WorkflowActionView`), so the fix may land on either side or both. No `com.dotmarketing.*`
+  legacy code is expected to be touched; the legacy edit mode reaches this data through a
+  different path and must be left alone.
+- **Related known decisions**: The new Edit Content UI is expected to reach behavioral parity
+  with the legacy editor for workflow execution. The plan phase formally consults
+  `dotCMS/platform-adrs`; of particular interest is anything covering the workflow REST view
+  contract, since one candidate fix changes what an existing endpoint returns.
+
+## Root-Cause Hypothesis
+
+The new editor decides whether to open the input wizard purely from the action's `actionInputs[]`
+array, and for unsaved content that array is never populated — so it takes the "no inputs, fire
+directly" branch.
+
+Two different REST shapes describe a workflow action:
+
+- `WorkflowActionView` (`convertToWorkflowActionView` in `WorkflowResource`) — carries
+  `actionInputs[]`, derived by `createActionInputViews` from `isAssignable()`, `isCommentable()`,
+  `hasPushPublishActionlet()` and the move actionlet.
+- `WorkflowDefaultActionView` — a plain `{ scheme, action, firstStep }` wrapper around the raw
+  `WorkflowAction`. It has **no** `actionInputs[]` at all.
+
+The endpoints the new editor uses for content that has no inode yet —
+`GET /api/v1/workflow/initialactions/contenttype/{id}` (`getDefaultActions`) and
+`GET /api/v1/workflow/defaultactions/contenttype/{id}` (`getWorkFlowActions`) — both return
+`WorkflowDefaultActionView`. The per-inode endpoint
+`GET /api/v1/workflow/contentlet/{inode}/actions` (`getByInode`) returns `WorkflowActionView`.
+
+In `core-web/libs/edit-content/src/lib/store/features/workflow/workflow.feature.ts`,
+`setSelectedWorkflow` seeds `currentContentActions` from `schemes[currentSchemeId].actions` —
+i.e. from the *default actions* payload, with no `actionInputs`. Then in
+`core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form.component.ts`,
+`fireWorkflowAction` does `const { actionInputs = [] } = workflow;` and, on
+`if (!actionInputs.length)`, fires the action immediately instead of calling `openWizard`.
+
+After that first execution, `fireWorkflowAction` in the store re-fetches with
+`workflowActionService.getByInode(inode, DotRenderMode.EDITING)`, which *does* include
+`actionInputs`. That is why the second click shows the dialog — and it is also why existing
+content opened directly in the new editor behaves correctly, since `initializeExistingContent`
+already loads its actions via `getByInode`.
+
+The legacy editor does not consult `actionInputs`; it reads the raw `commentable` / `assignable`
+flags, which is why it is unaffected.
+
+The plan phase will confirm this and choose between the candidate fixes (backend: have the
+default/initial-action endpoints emit `actionInputs`; frontend: derive the inputs from the raw
+action flags, or fetch an input-bearing action list for unsaved content). The trade-offs —
+notably that changing an existing endpoint's response shape is an API-contract change — belong
+in the plan, not here.
+
+## Fix Scope & Non-Goals *(mandatory)*
+
+**In scope**:
+
+- The new (Angular) Edit Content UI opens the workflow input dialog on the **first** execution of
+  any action that declares an input (comment, assign, push publish, move-without-path), including
+  for content that has not been saved yet.
+- The workflow action does not execute until the author submits or cancels that dialog, matching
+  legacy gating.
+- The same correctness applies after a **Reset** action, where the author re-selects a scheme and
+  the action list is re-seeded from the same source.
+- Whatever contract change this requires between `WorkflowResource`'s default/initial action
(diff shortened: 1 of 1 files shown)


rjvelazco (dismissed):

specs/36883-workflow-comments-dialog-new-edit/spec.md
```diff
@@ -0,0 +1,257 @@
...
+  `WorkflowAction`. It has **no** `actionInputs[]` at all.
+
+The endpoints the new editor uses for content that has no inode yet —
+`GET /api/v1/workflow/initialactions/contenttype/{id}` (`getDefaultActions`) and
+`GET /api/v1/workflow/defaultactions/contenttype/{id}` (`getWorkFlowActions`) — both return
+`WorkflowDefaultActionView`. The per-inode endpoint
+`GET /api/v1/workflow/contentlet/{inode}/actions` (`getByInode`) returns `WorkflowActionView`.
+
+In `core-web/libs/edit-content/src/lib/store/features/workflow/workflow.feature.ts`,
+`setSelectedWorkflow` seeds `currentContentActions` from `schemes[currentSchemeId].actions` —
```
nicobytes: **Q1 — `setSelectedWorkflow` is not the path this spec's own reproduction takes. Which seeding points are in scope?**

The Root-Cause Hypothesis names exactly one place where `currentContentActions` is seeded from the input-less default-actions payload. There are **three**, and the one named is the *least* likely to be hit by the repro steps:

| # | Location | When it runs |
|---|---|---|
| 1 | `content.feature.ts:215` (`initializeNewContent`) | New content, **single scheme** → `defaultSchemeId` auto-selected, seeded from `getDefaultActions` |
| 2 | `workflow.feature.ts:195` (`setSelectedWorkflow`) | New content, **multi-scheme** → author manually picks a scheme in the sidebar |
| 3 | `locales.feature.ts:271` | **New translation / copy-to-locale** flow, also seeded from `getDefaultActions` |

`setSelectedWorkflow` has exactly one caller — `(onSelectWorkflow)` in `dot-edit-content-sidebar.component.html:172`. The spec's repro uses a content type on **System Workflow only**, so `schemeIds.length === 1`, the scheme is auto-selected in `initializeNewContent`, and **`setSelectedWorkflow` is never invoked**. A fix scoped to `setSelectedWorkflow` alone would not fix the reported bug.

Questions:
1. Should the Root-Cause Hypothesis be amended to name path **1** as the primary one (with 2 as the multi-scheme variant)?
2. Is path **3** (new translation) in scope? It has the same latent defect and is currently in neither the in-scope list nor the non-goals — right now it falls through the cracks.

This also affects the Verification Method: the spec commits to a `workflow.feature.spec.ts` test for `setSelectedWorkflow`, but the repro's actual path lives in `content.feature.spec.ts`. If all three paths are in scope, a **single shared derivation helper** (in `workflows.utils.ts` alongside `parseCurrentActions`, or in `data-access`) looks like the natural fix location — it would cover all three without three separate patches.

specs/36883-workflow-comments-dialog-new-edit/spec.md
```diff
@@ -0,0 +1,257 @@
...
+**Expected Behavior**: The comment dialog appears on the **first** click of `Publish`, exactly as
+it does in the legacy edit mode. The workflow action does not run until the author submits or
+cancels the dialog.
+
+**Actual Behavior**: The workflow action fires immediately with no dialog and no comment. The
+dialog appears only on a subsequent execution of the same action, once the contentlet has been
+saved and its allowed actions have been re-fetched.
+
+**Reproducibility**: Always, for content that has not yet been saved in the current editor
+session (new content). It also reproduces for existing content after a **Reset** action, because
```
nicobytes: **Q2 — Does the Reset case actually reproduce, and if so by what mechanism?**

The stated rationale is that Reset "re-populates the action list from the same input-less source." Reading the code, that doesn't appear to be what happens:

- `workflow.feature.ts:~255` — on **every** fire, including Reset, `currentContentActions` is patched from `workflowActionService.getByInode(inode, EDITING)`, which **does** carry `actionInputs[]`.
- The Reset branch then sets `currentSchemeId: null` (when there is more than one scheme) but leaves `currentContentActions` holding those input-bearing actions.
- When the author re-picks a scheme, `setSelectedWorkflow` runs — but the contentlet now **has an inode**, so `isNew` is `false` and it takes the `else` branch, which updates only `currentSchemeId` and `currentStep` and **does not overwrite `currentContentActions`**.

So after a Reset the action list should still carry its inputs, and AC-006 may already pass on `main`.

Which is it?
- **(a)** AC-006 was verified failing — then the mechanism differs from the one stated and the spec should say what it actually is (there may be a *separate* defect here: after re-picking a scheme, `currentContentActions` still holds the actions fetched for the pre-reset state, which could be stale or wrong for the newly picked scheme — related, but not the missing-`actionInputs` bug).
- **(b)** AC-006 was reasoned about rather than reproduced — then it's still worth keeping as a regression guard, but the Reproduction section shouldn't assert it as observed behavior.

Worth settling before the plan, because (a) and (b) imply different fixes and different test cases.

specs/36883-workflow-comments-dialog-new-edit/spec.md
```diff
@@ -0,0 +1,257 @@
...
+  comment dialog on the **first** click — matching the legacy edit mode.
+- **AC-002**: The workflow action does not complete (the contentlet is not saved or transitioned)
+  until the author submits or cancels the dialog. Cancelling leaves the content unchanged and
+  untransitioned.
+- **AC-003**: A single execution produces exactly one workflow execution — one version and one
+  workflow history/audit entry. The comment the author typed is present on that history entry.
+- **AC-004**: Verified against at least one **non-`Publish`** workflow action with **Allow
+  Comments** enabled (and ideally on a non-System workflow scheme), confirming the fix is not
+  scoped to the System Workflow's default `Publish` action.
+- **AC-005**: Actions with **no** inputs still fire directly on the first click, with no dialog
```
nicobytes: **Q3 — What is the expected first-click behavior for a Push Publish action on *unsaved* content?**

AC-005 asserts Push Publish actions "present their inputs on the first click on unsaved content", and the Regression Risk section notes the `checkPublishEnvironments()` notification path must keep working. But `fireWorkflowAction` routes Push Publish to a distinct branch (`dot-edit-content-form.component.ts:549–576`): `containsPushPublish()` → `checkPublishEnvironments()` → `openWizard()` only if environments exist.

Two things to pin down:

1. Is Push Publish even **reachable** as an initial action on unsaved content? Push Publish typically requires a live/published version. If it isn't reachable pre-save, AC-005's Push Publish clause is untestable as written and should be scoped to saved content.
2. If it is reachable — when there are **no** publish environments configured, the expected outcome is the notification and **no** action fired, correct? Today the `subscribe` has no `else`: `hasEnvironments === false` silently does nothing beyond whatever `checkPublishEnvironments` itself notifies. Should the fix make that outcome explicit, or is it out of scope?

Separately (and echoing the earlier bot review): AC-005 bundles three independent behaviors — no-input actions firing directly, Assign-to presenting inputs, and Push Publish. Splitting them into AC-005/006/007 would make a failure point at one cause.

specs/36883-workflow-comments-dialog-new-edit/spec.md
```diff
@@ -0,0 +1,257 @@
...
+- **Data considerations**: None. No DB schema, no Elasticsearch mapping, no stored state. No
+  repair of existing data is needed — comments that were silently dropped by this bug are not
+  recoverable and are out of scope.
+
+## Acceptance & Verification *(mandatory)*
+
+- **AC-001**: With a content type using the **new** edit mode and a workflow action that has
+  **Allow Comments** enabled, creating a new contentlet and clicking that action shows the
+  comment dialog on the **first** click — matching the legacy edit mode.
+- **AC-002**: The workflow action does not complete (the contentlet is not saved or transitioned)
```
nicobytes: **Q4 — For *new* content, what does "leaves the content unchanged and untransitioned" mean concretely?**

AC-002 is unambiguous for existing content. For **new** content — the spec's own primary case — there is no contentlet yet: the workflow fire *is* the first save. "Unchanged" has no referent.

What should be true after the author cancels the dialog?

- **(a)** Nothing is persisted, the editor stays on the "new content" route, and the **form keeps every value the author typed** (they can correct something and fire again).
- **(b)** Nothing is persisted and the editor resets to a clean form.
- **(c)** Something else.

I'd expect (a), and it matters for the test: the component spec should assert not just that `$store.fireWorkflowAction` was *not* called, but also that the form state survived and a **second** fire still works (no stuck `ComponentStatus.SAVING`, no cleared `formValues`). Worth stating explicitly so the acceptance test is written for the right thing.

Related: the store patches `state: ComponentStatus.SAVING` and pushes a "processing" toast at the *start* of `fireWorkflowAction`. Since the fix moves the dialog **before** that call, no spinner should appear while the dialog is open — worth confirming that's the intent.

specs/36883-workflow-comments-dialog-new-edit/spec.md
```diff
@@ -0,0 +1,257 @@
...
+**Verified fact bearing on the fix**: every flag `createActionInputViews` reads is *already*
+serialized on the raw `WorkflowAction` embedded in `WorkflowDefaultActionView` — `commentable`
+(`isCommentable()`), `assignable` (`isAssignable()`), `hasPushPublishActionlet`,
+`hasMoveActionletActionlet` / `hasMoveActionletHasPathActionlet` (all four `@JsonProperty`), plus
+`nextAssign` and `roleHierarchyForAssign`. None is `@JsonIgnore`
+(`com.dotmarketing.portlets.workflows.model.WorkflowAction`). The default/initial-action payload
+therefore already carries everything needed to determine an action's inputs — what is missing is
+only the derived `actionInputs[]` array itself.
+
+The plan phase will confirm this and choose between the candidate fixes (backend: have the
```
nicobytes, line 158: **Q5 — Is the backend route genuinely still open, or is frontend-only the intended default?**

The spec defers the backend-vs-frontend choice to the plan, which is reasonable — but it leaves a meaningful chunk of the spec **conditional**: AC-008, the backend-integration run, and the Postman run are all gated on "only if the endpoints change". That collides with Constitution Principle V: the TDD gate requires tests written, approved, and confirmed failing **before** implementation, and right now we can't know which tests those are.

Which is it?

- **(a)** Frontend-only is the intended default given the "verified fact"; the backend route is a documented fallback. Then AC-008 and the integration/Postman lines can be marked *contingent, not expected*, and the Red gate is unambiguously the two Jest specs.
- **(b)** Genuinely open. Then the plan must resolve it **before** any test is written, and `/speckit-plan` should be run with that as its first decision.

Worth noting for (a): the commit on this branch (`02028ee`) is titled *"record that a frontend-only fix needs no REST change"*, which reads as (a) having already been decided — but the spec body still presents it as an open fork. Aligning the two would remove the ambiguity.

If (a) holds, one hand-off note for the plan: the TS derivation duplicates `createActionInputViews` (`WorkflowResource.java:902`). Putting it in **one** shared helper — which Q1 argues for anyway, since three call sites need it — plus a comment pointing back at the Java method, keeps the drift risk manageable.

nicobytes (commented):
## Spec review — clarification questions (`/speckit-clarify`)

Ran a structured ambiguity scan over `spec.md` before `/speckit-plan`. The root-cause analysis is strong and the "verified fact" about the already-serialized `WorkflowAction` flags holds up. Five points remain that would **materially change the implementation or the test design**, so I'd rather resolve them here than discover them mid-plan.

Q1 and Q2 are the important ones — both suggest the Root-Cause Hypothesis names an incomplete (Q1) or possibly incorrect (Q2) code path. Q3–Q5 are scope and acceptance-criteria precision.

| # | Area | Question | Blocking for plan? |
|---|---|---|---|
| Q1 | Root cause / scope | `setSelectedWorkflow` is not the path the spec's own repro takes — there are three input-less seeding points. Which are in scope? | **Yes** |
| Q2 | AC-006 / Reset | Does the Reset case actually reproduce? The code suggests the action list survives from `getByInode`. | **Yes** |
| Q3 | AC-005 / Push Publish | Expected first-click behavior for Push Publish on *unsaved* content? | No |
| Q4 | AC-002 / Cancel | What does "leaves the content unchanged" mean for content that does not exist yet? | No |
| Q5 | AC-008 / test strategy | Is the backend route genuinely still open, or is frontend-only the intended default? | No |

Each question is left inline on the line it refers to. No spec edits made — happy to fold the answers into a `## Clarifications` section once they're settled.

cc @adrianjm-dotCMS @rjvelazco

adrianjm-dotCMS:
Thanks @nicobytes — **Q1 and Q2 found real errors**, both corrected in `d61a0be`. Verified each against the code before answering; evidence in the new `## Clarifications` section.

| # | Verdict | Outcome |
|---|---|---|
| Q1 | ✅ Wrong path named | Root cause rewritten, primary test moved |
| Q2 | ✅ Reasoned, not observed | Claim + AC-006 removed |
| Q3 | 🟡 Asserted unverified | Push Publish dropped as AC; AC-005 split |
| Q4 | ✅ Undefined for new content | AC-002 now specifies it |
| Q5 | ✅ Already resolved in the plan | AC-009 marked contingent |

ACs renumbered 001–009. Also withdrew a non-goal whose premise was false — those endpoints have exactly three call sites, all in `libs/edit-content`.

Your shared-helper suggestion matches what the plan had already landed on independently.

specs/36883-workflow-comments-dialog-new-edit/spec.md
adrianjm-dotCMS: Confirmed — a fix scoped to `setSelectedWorkflow` wouldn't have fixed the bug. `initializeNewContent` (`content.feature.ts:211-218`) seeds the actions when `schemeIds.length === 1`; `setSelectedWorkflow`'s only caller is the sidebar's `(onSelectWorkflow)`.

Both answers yes: hypothesis now names all three paths with path 1 primary, and path 3 is in scope. Primary store test moved to `content.feature.spec.ts`.

On the helper — agreed, and that's the plan: `deriveActionInputs()` in `libs/data-access`, applied in `getDefaultActions()` / `getWorkFlowActions()`. At the service boundary all three call sites are fixed transitively, so `libs/edit-content` needs no production change.

specs/36883-workflow-comments-dialog-new-edit/spec.md
adrianjm-dotCMS: **(b)** — reasoned, not reproduced, and wrong. Your reading is right: `currentContentActions` ← `getByInode` on every fire (`workflow.feature.ts:287`), and post-Reset `isNew` is false, so `setSelectedWorkflow` takes the `else` branch and doesn't overwrite it.

AC-006 removed with no replacement; Reproducibility now carries an explicit "does not reproduce after Reset" note.

Your separate point — the retained list may be stale for a newly picked scheme — is real but a distinct defect. I haven't reproduced it, so I'd rather not add a second unverified claim here. Happy to file it separately.

specs/36883-workflow-comments-dialog-new-edit/spec.md
adrianjm-dotCMS: Partly valid, and the part that lands is that I asserted something unverified.

Reachable in principle, not by default: `WorkflowHelper.java:1985-1988` builds an empty contentlet and calls `findAvailableActionsEditing`, so Push Publish is returned only if an admin puts it in the first step. **No longer an AC.**

The missing `else` is real but pre-existing, so out of scope — recorded so it isn't mistaken for something the fix introduced.

Bundling: agreed. AC-005 is now just no-input actions, Assign-to is AC-006, Push Publish is gone.

specs/36883-workflow-comments-dialog-new-edit/spec.md
adrianjm-dotCMS: Agreed, **(a)**. AC-002 now states it for new content: nothing persisted, form keeps typed values, second fire still works.

Your spinner point is correct and folded in — `SAVING` is set inside `$store.fireWorkflowAction`, which the fix leaves after the dialog, so no indicator while it's open.

specs/36883-workflow-comments-dialog-new-edit/spec.md
adrianjm-dotCMS: **(a)**, and you're right that the commit title and the spec body disagreed.

The plan settled it; the decisive input was **ADR-0013** — a `core-web/**`-only PR skips integration/Postman in the merge queue, so those would never be the gate. AC-008 → **AC-009, contingent**; integration/Postman marked not-expected.

Yes to the hand-off note — one `deriveActionInputs()` with a JSDoc naming `WorkflowResource#createActionInputViews` and the sync obligation. The Java/TS duplication is the accepted trade-off in the plan's Complexity Tracking.

rjvelazco (approved):

PR #37442, opened 2026-09-07
fix(block-editor): stop creating emoji nodes and heal the ones already stored

### LINKED ISSUE #37340: Block Editor splits a link into two anchors when ©, ®, or ™ appears inside the linked text

rjvelazco:
Fixes #37340. PR 2 of 2 — the spec merged in #37434.

Typing `©`, `®` or `™` inside linked text split the link into two anchors and lost the character on published pages. The cause is four lines of editor configuration; v1 chased it downstream into five renderers, Java and three npm packages.

**13 files, over half of them tests. No VTL, no Java, no SDK, no version bump.**

## What was actually wrong

The emoji extension replaced any character Unicode classifies as an emoji with a **bare** `emoji` node — 1907 of the 1949 it catalogues, including 219 that render as ordinary typography (`✔`, `⚠`, arrows, `▪`). A bare inline atom inside a marked run ends that run, so a linked phrase became `text(link) + emoji + text(link)`, persisted into the stored JSON.

Worse, the node stores a TipTap **shortcode**, never a character, and the table that resolves one ships inside an npm dependency of the editor. Nothing downstream can perform that lookup.

## Four stories

| | Delivers |
|---|---|
| **US1** | All **five** node-creation paths closed. The issue names one |
| **US2** | Stored nodes heal to text on load |
| **US3** | The link sandwich rejoins the reported payload into one `<a>`, no re-save |
| **US4** | Emoji authoring works on every field, restricted or not |

The picker, `:copyright:` and `:)` all keep working — only their output changes.

## Found while doing it

**A live Java defect this fixes for free.** `StoryBlockUtil.isTextContentEmpty` counts only `"text"` nodes and feeds required-field validation at `ESContentletAPIImpl.java:8180` — so **a required Story Block field whose only content is one emoji currently fails to save as "empty."** `getCharCount` and ES text extraction have the same blind spot. Healing the node to text fixes all three with no Java change.

**A wrong assumption that survived three drafts.** The spec claimed ProseMirror joins adjacent identical-mark text nodes on load. It does not — `Fragment.fromJSON` constructs directly and never reaches `Fragment.fromArray`. It looked true because the editor *renders* one `<a>` regardless. Unmerged, the heal would have stored three `text(link)` nodes and rendered **three** anchors where the defect produces two. The merge is now ours. Evidence in `research.md` R10.

**The emoticon rule was eating a keystroke.** TipTap marks the text-input event handled, so the space that triggered `:) ` never reached the document. `hi :) there` now keeps its spacing.

**A gate that could only misfire.** `emoji` is not selectable in Allowed Blocks (`getEditorBlockOptions()` offers block nodes only, #37175), so `isAllowed('emoji')` was true *only* on unrestricted fields — restricting any block silently removed the emoji button and `:)`. Nobody configured that.

## Three things worth your attention in review

1. **`emoji-heal.utils.ts` — the sandwich rule is narrower than AC-015 requires.** It matches on full mark-set equality, not just the `link` marks. It can only fire less often and it keeps the merge coherent, but it is a deviation from the written criterion.
2. **US4's Red was retroactive.** The gate was removed before the tests existed; Red was confirmed by reverting the two toolbar files and re-running. The set discriminates correctly, but the TDD order was not followed. Noted in the commit and `tasks.md` rather than glossed.
3. **One pre-existing #37175 assertion is deliberately inverted** — it required `enableEmoticons: false` on restricted fields.

## Tests

259 passing, up from a 158 baseline. Jest only: no Java, DB, REST, renderer or build artifact is touched, and accepted [ADR-0013](https://github.com/dotCMS/platform-adrs/blob/main/decisions/0013-skip-integration-and-postman-tests-for-frontend-only-changes-in-merge-queue-to-increase-flow.md) skips integration/Postman/Karate for `core-web/**` in the merge queue.

The 9 negative cases in `emoji-heal.utils.spec.ts` are the real deliverable of US3 — they are what stops the one permitted inference widening.

## Still outstanding

- Manual accessibility pass (NVDA / VoiceOver) → post-merge QA plan
- Two superseded ACs on #37340 need amending. The third — "single link without a re-save" — is now **met**, not amended
- Ticket 39197 triage precondition still unconfirmed
- An ADR on editor-side normalization of deprecated Story Block nodes is *proposed*, not a blocker

### Video

https://github.com/user-attachments/assets/850f5424-65a0-49aa-b281-fa0b4c7c2e06

This PR fixes: #37340


rjvelazco:
> Added by Claude Code, on @rjvelazco's behalf.

## Added: the `:` emoji autocomplete (`a90b308d7a`)

Scope addition after manual testing, folded in deliberately rather than split out. **Spec amended to v2.2 in the same commit** — a spec saying "deliberately not done" beside code that does it is worse than either alone.

**Why it belongs here:** suppressing the emoji node left `:copyright:` working but undiscoverable. The shortcode rule was already in scope (AC-006); this is its affordance.

## Small because almost nothing is new

**Reused verbatim from upstream:** the `:` trigger, `EmojiSuggestionPluginKey` so the session can't collide with the slash menu's, and the `allow` guard.

**Reused from this lib:** `SlashMenuService` for the dropdown, positioning and keyboard nav — `open()` already takes its items as an argument, so **no new component**.

**Replaced:** upstream's `command`, which inserted an emoji **node**. Leaving it would have reopened the defect through the menu itself.

**New:** `items` only. Upstream ships no default and never has — which is what the old `items: () => []` actually was. A blank filled with nothing, not a working menu switched off.

## ⚠️ Two shared-component changes

These affect `/` rows too, so they need a look:

* **`slash-menu.component.html`** — the label span gains `truncate`. Long shortcodes are wider than the fixed `w-72` dropdown and were giving the whole list a horizontal scrollbar. Invisible for block rows, whose labels are two words. The `description` span was already truncated; the label had been missed.
* **`slash-menu.types.ts`** — `BlockItem` gains `iconKind?: 'material' | 'glyph'`, defaulting to `material`. Glyph rows render bare: no bordered box, no `material-symbols-outlined`, which sets a font-family and variation settings built for Material's own ligature names.

Explicit flag rather than sniffing the string for an emoji. This is shared UI; the next reader should see intent, not a regex.

## Two ranking defects, both caught by fixing weak tests

My first three specs here were poor — one was tautological. Strengthening them exposed real bugs:

* **`:rocket:` returned `:astronaut:` first.** A tag prefix ranked equal to a shortcode prefix, and astronaut tags "rocket". Now four tiers: exact name/shortcode → name/shortcode prefix → tag prefix → substring.
* **`smi` labelled rows `:grinning_face_with_closed_eyes:`.** Rows used `shortcodes[0]`; `smile` ships as `["grinning_face_with_closed_eyes", "smile"]`. Now labelled by `name`, which is also what `attrs.name` carries.

## The assertion that matters

**AC-025** drives the **real** `command` through the live plugin via a recording menu-service stub, asserting a character lands and no `emoji` node appears. Simulating the insert — which is what my first draft did — would have tested the assertion rather than the code, and this command is precisely the piece that replaced upstream's node-creating one.

## A gotcha for anyone writing more of these

`@tiptap/suggestion`'s plugin view declares `update` as **async**, so its render callbacks land in a microtask after the dispatch. A synchronous assertion sees the plugin state updated but the menu not yet opened, which reads as a false failure. Hence `flushSuggestion()`.

## Also in this push

`b2f086c326` — the heal now **emits** when it rewrites something, so a plain Save persists the repair. Previously it only stuck if the author also typed: `loadContent` uses `emitUpdate: false`, so a healed document sat in the editor while the form control held the unhealed string, and the editor renders both identically. Gated on `healEmojiNodes` having actually changed the document, so a field with no `emoji` node stays pristine and is never dirtied by being opened.

**270 tests, 19 suites.** AC-001 through AC-028.

rjvelazco:
@claude review this code

adrianjm-dotCMS:
**QA note — legacy content is marked dirty on open**

`emitHealedValue` (`editor.component.ts`) calls `onChange` whenever the heal actually rewrote something, so a contentlet whose Story Block holds stored `emoji` nodes becomes **dirty the moment it is opened**, with no author edit.

This is intentional and guarded (`healed !== preserved`) — without it the repaired document sits in the editor while the form control still holds the unhealed string, and a plain Save throws the repair away. Flagging it only so QA isn't surprised by it:

- Opening and then leaving a contentlet with legacy emoji content will trigger the unsaved-changes prompt even though nothing was typed.
- Save / workflow actions will light up on open for that content.
- The case worth explicitly verifying is the **negative** one: content with no stored `emoji` node must stay pristine on open. A regression there would mark every Story Block field dirty just for being opened.

Not a blocker — just something to have in the post-merge QA plan.

adrianjm-dotCMS (approved):

Issue #34347, opened 2026-01-19
[TASK] Show workflow dialogs for comments and user assignment in New Edit Contentlet mode


erickgonzalez:
### Description

In the New Edit Contentlet mode, when a workflow action has Allow Comments and/or Allow User Assign enabled, the UI does not currently display any dialog to collect this information from the user.

This behavior differs from the Old Edit Contentlet mode, which correctly prompts the user with the appropriate dialogs when these workflow options are enabled.

The New Edit Contentlet mode should support the same workflow interaction patterns to ensure feature parity and avoid loss of functionality during workflow execution.

### Acceptance Criteria

- When a workflow action has Allow Comments enabled, the New Edit Contentlet mode displays a dialog containing a textarea for user comments before applying the workflow action.
- When a workflow action has Allow User Assign enabled, the New Edit Contentlet mode displays a dialog containing a dropdown with users configured in the workflow.
- When both options are enabled, the dialog supports capturing both comment text and user assignment.
- The dialog behavior and UX are consistent with the Old Edit Contentlet mode.
- Workflow actions cannot be completed without providing required inputs when these options are enabled.

### Priority

Medium

### Additional Context

- Use the Old Edit Contentlet mode implementation as the reference for expected behavior and UX.
- This task is required to maintain functional parity between old and new editing experiences and prevent workflow regressions.

erickgonzalez:
Passed QA. Dialogs are showing correctly on the new edit mode.

https://github.com/user-attachments/assets/bd3ecfaf-23c1-4f40-a4a7-8b9b4e5d1f77

PR #37445, opened 2026-09-07
fix(edit-content): show the workflow comment dialog on first execution (#36883)

### LINKED ISSUE #36883: Workflow "Allow Comments" dialog not showing up in new Edit Mode

adrianjm-dotCMS:
**Spec-Kit PR 2 of 2.** Implementation for the spec approved in #37437.

https://github.com/user-attachments/assets/dfca8068-7da4-430e-9adc-75d1dbce3585

## What was broken

Firing a workflow action with **Allow Comments** enabled skipped the comment dialog on its first execution in the new (Angular) Edit Content UI. The contentlet was saved and transitioned, but the author never got to enter the comment — the dialog only appeared on a *second* execution. Legacy edit mode was correct.

The comment is the audit note recorded in the workflow task history, so this silently degraded the audit trail with no error surfaced.

## Root cause

Two REST shapes describe a workflow action, and only one carries `actionInputs[]`:

| Endpoint | Returns | `actionInputs[]`? |
|---|---|---|
| `contentlet/{inode}/actions` | `WorkflowActionView` | ✅ |
| `initialactions/contenttype/{id}` | `WorkflowDefaultActionView` | ❌ |
| `defaultactions/contenttype/{id}` | `WorkflowDefaultActionView` | ❌ |

The editor gates its input wizard on `actionInputs.length`, so for content with **no inode** the array was absent and it took the "no inputs, fire directly" branch. After the first fire the store re-fetches per-inode actions, which *do* carry the array — hence the dialog on the second click.

Verified against a live instance rather than inferred:

```
GET /api/v1/workflow/initialactions/contenttype/webPageContent
  Publish → commentable: true,  (no actionInputs key)

GET /api/v1/workflow/contentlet/{inode}/actions?renderMode=EDITING
  Publish → commentable: true,  actionInputs: [{body:{}, id:'commentable'}]
```

## The fix — frontend only, no REST change

Every flag the derivation needs is already serialized on the raw `WorkflowAction` (none `@JsonIgnore`), so no backend change is required.

* **`dot-workflows-actions.utils.ts`** (new) — `deriveActionInputs()`, a TypeScript mirror of `WorkflowResource#createActionInputViews`, plus `withDerivedActionInputs()` which leaves a server-provided array untouched.
* **`dot-workflows-actions.service.ts`** — applied in `getDefaultActions()` and `getWorkFlowActions()` only. `getByInode()` / `getByWorkflows()` / `getBulkActions()` are deliberately left alone (commented in place): their endpoints build the array server-side, and deriving over them would hide a future backend regression instead of surfacing it.

**`libs/edit-content` has no production-code change** beyond one clarifying comment — all three affected call sites are fixed transitively by the service-boundary fix.

### Why not the backend

Emitting `WorkflowActionView` from those two endpoints is arguably the cleaner contract, but it changes a published REST response on endpoints reachable by customer integrations, and per [ADR-0013](https://github.com/dotCMS/platform-adrs/blob/main/decisions/0013-skip-integration-and-postman-tests-for-frontend-only-changes-in-merge-queue-to-increase-flow.md) it would flip this PR onto the slow CI path — for an identical behavioral outcome. Recorded as accepted tech debt in the plan's Complexity Tracking: the derivation rule now exists in both Java and TypeScript, and each side carries a comment naming the other.

## Scope wider than the issue described

The issue describes one reproduction path. There are **three**, all seeded from the same input-less payload, all closed by this one change:

1. New content, **single** scheme — `initializeNewContent` (the reported repro)
2. New content, **multi-scheme** — `setSelectedWorkflow`
3. **New translation** of an untranslated locale — `locales.feature`

A fourth path claimed by an early draft of the spec — after a **Reset** action — was disproved during review and removed.

## Intentional behavior changes on unsaved content

`deriveActionInputs` mirrors `createActionInputViews` in full, so it emits `pushPublish` too. On content with no inode that routes the action through `checkPublishEnvironments()` instead of firing directly — so with **no publish environment configured**, a Push Publish action that previously "saved" now shows the add-an-environment error and does not run.

Deliberate: the previous save only looked like a success. It fired with no `whereToSend`, and `PushPublishActionlet` logs `"There are no valid environments to sent to"` and returns (`PushPublishActionlet.java:145-149`) — content saved, push publish silently did nothing. Deriving it also keeps new content consistent with saved content, which already gets `pushPublish` from `getByInode`.

Narrow reach: needs an admin to put a Push Publish sub-action in the workflow's first step; none of the 7 content types checked on a demo instance have one.

**`moveable` is the second one.** A Move action with no preset path also fired directly on unsaved content before; it now folds into a `commentAndAssign` step and asks the author for a destination. Same reasoning — parity with saved content, which already gets `moveable` from `getByInode` — and no environment gate, so it cannot block a save. Worth exercising in QA alongside Push Publish.

**`assignable` is the third.** Same mechanism: an action with Assign-to on unsaved content now presents the role picker on the first click instead of firing straight through. Same reasoning again — parity with saved content, which already gets `assignable` from `getByInode` — and no gate that can block a save. Already covered by a pre-existing test (`dot-edit-content-form.component.spec.ts:691`).

## Tests

25 tests across the two touched specs plus one new spec file.

| | |
|---|---|
| `deriveActionInputs` unit spec (new) | 13 — one per rule, combinations, ordering, the move-with-path exclusion, purity, idempotence |
| Service spec | 5 — derivation on both endpoints, `[]` never `undefined`, preserve server-provided, and **not** re-derived on `getByInode` |
| Component spec | 4 new — dialog gating (AC-002): nothing fires while the dialog is open, fires only on submit carrying the comment, cancel fires nothing, form values survive a cancel |
| Store spec | 1 — `actionInputs` survives the store's reshape on the reported repro's path |

**Red confirmed before implementing** (Constitution Principle V): `Expected: [{"body":{},"id":"commentable"}] / Received: undefined`.

### A note on what could and could not go Red

Only the utils and service specs can actually fail for this bug. The component's branch logic was never wrong — `actionInputs` arrives as an argument — and it was already covered. The store specs mock `DotWorkflowsActionsService` entirely, so they never see the real derivation. Rather than write always-green tests and call them regression guards, the component tests were pointed at the genuinely-uncovered AC-002 gating behavior, and two of the three planned store tests were dropped as redundant copies of the same assertion. Signed off by the developer at the task gate.

## Verification

- `pnpm nx test data-access` — 86 suites, 870 passed
- `pnpm nx test edit-content` — 116 suites, 2346 passed
- `pnpm nx run-many -t lint -p data-access,edit-content` — 0 errors (4 warnings, all pre-existing)
- `pnpm nx format:check` — clean
- `npx tsc -p tsconfig.spec.json --noEmit` — 0 errors in the touched files (pre-existing baseline unchanged)
- Manual QA in the running app: dialog on first click, gating and cancel behavior, single version and history entry, no-input actions unaffected, and legacy edit mode unchanged (AC-008)
- No file outside `core-web/`; `openapi.yaml` untouched

**Not run, deliberately**: backend integration and Postman. No Java changes, and per ADR-0013 a `core-web/**`-only PR skips both in the merge queue — confirmed in practice, neither was triggered.

## Checklist
- [x] Manual QA, including the legacy-mode check (AC-008)
- [ ] E2E — not added. `edit-content/workflows/workflow-actions-dialog.spec.ts` (the regression spec for #34347) covers this dialog but always starts from an existing inode, so it never exercises the path this bug lived on. The new-content variant is blocked on fixture behavior: an action created on the workflow's first step does not surface in `GET /initialactions/contenttype/{id}`. Worth its own investigation.

## Additional Info

Resolves #36883. Spec: #37437.

This PR fixes: #36883

This PR fixes: #36883


core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.service.spec.ts
```diff
@@ -112,6 +112,120 @@ describe('DotWorkflowsActionsService', () => {
...
     });
 
+    describe('actionInputs derivation (#36883)', () => {
+        // The default/initial-action endpoints return `WorkflowDefaultActionView`, which wraps the
+        // raw `WorkflowAction` and carries NO `actionInputs` key — unlike the per-inode endpoint,
+        // which returns `WorkflowActionView` with the array already built server-side. The editor
+        // gates its input wizard on `actionInputs.length`, so without derivation a commentable
+        // action fires with no dialog. Fixtures below deliberately omit the key, the way the real
+        // payload does.
+        const rawCommentableAction = {
```
nicobytes: Hand-rolling a `DotCMSWorkflowAction` here goes against the frontend testing standard: domain fixtures belong in `@dotcms/utils-testing` as `createFake*` helpers (same pattern as `createFakeContentlet` / `createFakeSite`), not as inline objects in the spec.

There is already `mockWorkflowsActions` in `dot-workflows-actions.mock.ts`, but it always includes `actionInputs`, so it cannot stand in for this suite — these tests need the key **omitted**, matching `WorkflowDefaultActionView`. Reusing that constant would hide the bug this describe is covering.

Please add something like `createFakeWorkflowAction(overrides?: Partial<DotCMSWorkflowAction>)` in `@dotcms/utils-testing` and use it here (and in `dot-workflows-actions.utils.spec.ts`, which currently duplicates the same shape as `buildRawAction`). Keep `actionInputs` out of the defaults so callers can still represent the raw REST payload this PR is fixing.

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form.component.spec.ts
```diff
@@ -712,6 +713,150 @@ describe('DotFormComponent', () => {
...
+                    spy.mock.calls[call][0] as DotFireActionOptions<{
+                        [key: string]: string | object | string[];
+                    }>;
+
+                // `DotWizardService` is a factory-level mockProvider, so its `open` jest.fn is
+                // shared by every test in this file. The tests below swap in never-emitting
+                // Subjects, and `jest.clearAllMocks()` clears call data but NOT implementations —
+                // so without this restore the last Subject would leak into any later test that
+                // expects `open()` to emit, breaking it by declaration order alone.
+                afterEach(() => {
```
nicobytes: **Required — this `afterEach` restores one mock out of four.**

The comment diagnoses the problem exactly right: `jest.clearAllMocks()` (line 190) clears call data but not implementations, and these are factory-level `mockProvider`s shared by the whole file. But the restore only covers `DotWizardService.open`. Three more overrides introduced in this block are never restored:

- `processWorkflowPayload` — `mockImplementation` (line 763)
- `containsPushPublish` → `true` (lines 832, 848)
- `checkPublishEnvironments` → `of(false)` / `of(true)` (lines 833, 849)

`DotWorkflowEventHandlerService` is a factory-level `mockProvider` too, so from line 832 onward `containsPushPublish` returns `true` for the remaining ~1000 lines of this spec. It is latent today only because nothing below `describe('Preview Button')` fires a workflow action — the next test added there inherits it silently, and will fail for a reason that has nothing to do with its own subject.

Cheaper and order-independent than compensating per-mock: set the default implementations in the file's root `beforeEach`, or add `restoreMocks: true` to the Jest config (alternatively `jest.restoreAllMocks()` in the root `afterEach`), and drop this `afterEach` entirely.

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.utils.ts
```diff
@@ -0,0 +1,63 @@
...
+ * endpoints that are supposed to build it themselves.
+ *
+ * @param action the workflow action as it came off the wire
+ * @returns a new action; the argument is never mutated
+ */
+export const withDerivedActionInputs = (action: DotCMSWorkflowAction): DotCMSWorkflowAction => {
+    // No optional chaining on `action` itself: `deriveActionInputs` dereferences it anyway, so a
+    // `?.` here would only move the TypeError one line down while reading as if it were handled.
+    // A missing action is a malformed payload — let it fail loudly rather than half-guard it.
+    if (action.actionInputs?.length) {
```
nicobytes: **Required — the gate and its doc comment disagree.**

The doc above says *"An array the server already sent is left untouched"*, but `action.actionInputs?.length` is truthiness on the length: a server-provided **empty** array is re-derived. The service spec already documents the gap in its own comment (*"short-circuits on any NON-EMPTY array"*), which is a good sign that the code, not the comment, is the odd one out.

It is harmless today, because both endpoints omit the key entirely. But the stated purpose of this guard is to *surface* a backend regression rather than mask it, and the one case it fails to guard is exactly the regression shape you would most want to see: an endpoint that starts emitting `WorkflowActionView` with a legitimately empty `actionInputs`.

```ts
if (Array.isArray(action.actionInputs)) {
    return action;
}
```

That makes the code match the comment, keeps the "never `undefined`" guarantee intact, and leaves every current test green — the `getByInode` assertion still holds, since that fixture omits the key rather than sending `[]`.

nicobytes (commented):
## Review: correctness, quality, simplification, Angular standards

Reviewed locally against `110d64cd46` (branch HEAD matches the PR head), with the Java side verified in the source rather than from the description.

**Verdict: approve-with-comments — nothing here blocks.** The diagnosis is right, the fix is at the correct layer (one service boundary instead of the three call sites), and the "why not the backend" argument holds up. The three Required items are test hygiene and code-vs-doc consistency, not correctness of the fix.

### Verified independently, so treat these as settled

- **The Java mirror is faithful.** `deriveActionInputs` matches `WorkflowResource#createActionInputViews` (`WorkflowResource.java:902-928`) line for line, including rule order and the `moveable`-with-path exclusion. The flags it reads are all really serialized: `isAssignable()` / `isCommentable()` carry no `@JsonIgnore`, and the `has*Actionlet()` getters carry explicit `@JsonProperty` (Jackson would not auto-detect a `hasX()` getter otherwise).
- **The blast radius is contained.** Only three call sites consume `getDefaultActions` / `getWorkFlowActions`, all inside `libs/edit-content`. No other portlet changes behavior. `getByWorkflows`, left underived, has one consumer — `dot-workflows-actions-selector-field.service` — which selects actions rather than firing them, so the omission is genuinely inert there.
- **No silent no-op path.** I checked whether a derived array could make `setWizardInput()` return `null`, which would have `openWizard` call `dotWizardService.open(null)` and hang with no dialog and no error. It cannot: every id the derivation emits maps to a wizard step. Worth noting the component's `openWizard` still lacks the `if (wizardInput)` guard that `DotWorkflowEventHandlerService.openWizard` has, so this is a near miss rather than a designed-out one.
- **Tests pass.** `pnpm nx test data-access --testPathPatterns=dot-workflows-actions` → 2 suites, 25/25 green.

### One cross-cutting note

The diff carries roughly 90 lines of comment for 120 lines of code, and three of the findings below (2, 3, 6) are places where a comment and the thing it describes have *already* diverged before merge. The comments are unusually thoughtful, which is why this is worth saying out loud rather than only case by case: a comment that restates a nearby invariant, or describes code in another file, is a second source of truth that nothing tests. The ones that carry real intent — the Java mirror pointer, the `#36883` framing, why `getByInode` is left alone — are the ones worth keeping.

Findings are labelled **Required** / **Consider** / **Nit** inline; 3 Required, 4 Consider, 6 Nit. I have left the formal approval to you rather than stamping it myself.

---

specs/36883-workflow-comments-dialog-new-edit/contracts/derive-action-inputs.md
```diff
@@ -0,0 +1,67 @@
...
+| `getByWorkflows()` | `POST /api/v1/workflow/schemes/actions/NEW` | ❌ no — out of scope; not consumed by edit-content |
+| `getBulkActions()` | `POST /api/v1/workflow/contentlet/actions/bulk` | ❌ no — different view type, out of scope |
+
+Deliberately **not** applied everywhere: blanket application would mask a future backend regression
+in the endpoints that are supposed to populate the field themselves.
+
+## Sync obligation
+
+If `createActionInputViews` gains an input type in Java, this helper and its spec must be updated
+in the same PR. Each side carries a comment naming the other. This duplication is the accepted
```
nicobytes: **Required — the sync obligation only exists on one side.**

"Each side carries a comment naming the other" is not true as of this PR: there is no Java change in the diff (0 `.java` files), and `WorkflowResource.java` contains no reference to `dot-workflows-actions.utils.ts` or to `deriveActionInputs`. So a Java dev adding a fifth input type to `createActionInputViews` has nothing pointing them at the TypeScript mirror — which is the single failure mode this whole section exists to prevent.

Either option is fine:

1. Add the one-line comment above `createActionInputViews` naming the TS mirror. Note this puts a `.java` file in the diff — worth confirming against ADR-0013 whether a comment-only Java change still keeps the PR on the fast CI path.
2. Drop the claim from this doc and file a follow-up issue, so the drift risk has a ticket instead of a sentence.

For the record, I verified the mirror itself is faithful line-for-line against `WorkflowResource.java:902-928`, including rule order and the `moveable` path exclusion.

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.utils.ts
```diff
@@ -0,0 +1,63 @@
...
+        actionInputs.push({ id: 'commentable', body: {} });
+    }
+
+    if (action.hasPushPublishActionlet) {
+        actionInputs.push({ id: 'pushPublish', body: {} });
+    }
+
+    // Has a move actionlet but the path is empty — a preset path means there is nothing to ask
+    // the author for.
+    if (action.hasMoveActionletActionlet && !action.hasMoveActionletHasPathActionlet) {
```
nicobytes, line 41: **Consider — `moveable` is a second behavior change, and it is not documented.**

The description calls out `pushPublish` as *the* intentional behavior change, but the derivation emits `moveable` too, and that is a second one. A first-step action with a Move actionlet and no preset path previously fired directly on unsaved content; it now flows through `mergeCommentAndAssign` into a `commentAndAssign` step and opens the wizard asking the author for a path.

That is the right outcome — it is parity with saved content, which already gets `moveable` from `getByInode`. I also checked the failure mode it could have had: `setWizardInput()` returning `null` would have made the action a silent no-op, since `openWizard` in the component passes the result straight to `dotWizardService.open()` without the `if (wizardInput)` guard that `DotWorkflowEventHandlerService.openWizard` has. It does not happen — `isValidActionInput` includes `MOVEABLE`, so a `moveable`-only action still yields one step — but that is a narrow miss worth knowing about.

Two asks: mention it in the description next to the Push Publish note so QA exercises it, and add a test. The closest existing one asserts the derivation, not that a `moveable`-only action reaches `openWizard`.

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.service.ts
```diff
@@ -78,7 +103,8 @@ export class DotWorkflowsActionsService {
             >(`${this.BASE_URL}/initialactions/contenttype/${contentTypeId}`)
             .pipe(
                 map((x) => x?.entity),
-                map((res) => res || [])
+                map((res) => res || []),
+                map(deriveInputsOnEach)
```
nicobytes: **Consider — three `map`s where one says the same thing.**

```ts
.pipe(map((res) => deriveInputsOnEach(res?.entity ?? [])));
```

Same behavior, one operator instead of three, two fewer intermediate emissions per call, and `?? []` reads more directly than `map((res) => res || [])`. Applies identically to `getWorkFlowActions` (line 152).

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.service.ts
```diff
@@ -14,6 +14,31 @@ import {
...
+ * Left alone, for two different reasons:
+ * - `getByInode` and `getBulkActions` — their endpoints build `actionInputs` server-side
+ *   (`WorkflowActionView`), so deriving over them would hide a backend regression rather than
+ *   surface it.
+ * - `getByWorkflows` — its endpoint (`POST /schemes/actions/{systemAction}`) returns raw
+ *   `List<WorkflowAction>` (`WorkflowHelper#findActions`) and *does* have the same omission,
+ *   despite its `@Schema` advertising `ResponseEntityWorkflowActionsView`. It is out of scope
+ *   here only because nothing in edit-content consumes it; deriving there needs its own tests.
+ */
+const deriveInputsOnEach = (
```
nicobytes: **Consider — this comment documents other functions, from the wrong place.**

Fifteen lines of comment above a six-line helper, and most of it describes code 60–100 lines away: why `getByInode`, `getBulkActions` and `getByWorkflows` are *not* touched. That is the shape that rots — apply derivation to `getByWorkflows` later and this comment becomes false, with nothing next to `getByWorkflows` pointing at it. The per-endpoint table in `contracts/derive-action-inputs.md` also already records the same rationale, so this is a second copy of it.

Suggestion: keep two lines here (what the helper does, plus `#36883`) and let the contract doc own the per-endpoint reasoning. And since the helper is pure and uses no `this`, moving it into `dot-workflows-actions.utils.ts` next to `withDerivedActionInputs` would put it where the rest of that rationale already lives, keeping this file to HTTP plumbing.

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.utils.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+
+            expect(deriveActionInputs(action).map(({ id }) => id)).toEqual([
+                'assignable',
+                'commentable'
+            ]);
+        });
+    });
+
+    describe('guarantees', () => {
+        it('should always return an array, never undefined', () => {
```
nicobytes: **Consider — `withDerivedActionInputs` has no direct spec, and three of these four tests are tautological.**

This file imports only `deriveActionInputs`. The function that actually implements the decision — preserve vs derive, `[]` vs `undefined`, non-mutation of the incoming action — is exercised only indirectly, through the HTTP service spec. Three direct tests here would pin that contract where it lives, and would make the `?.length` vs `Array.isArray` choice an explicit, visible one.

The budget for them is already in this block:

- `should always return an array, never undefined` (113) is what `should derive nothing when no flag is set` (35) already proves with `toEqual([])`.
- `should tolerate the optional actionlet flags being absent` (136): `buildRawAction()` never sets `hasPushPublishActionlet` or `hasMoveActionletActionlet`, so both `delete`s are no-ops — the test is a copy of `should derive commentable`.
- `should be idempotent when re-derived from an already-derived action` (128) passes `actionInputs` into a function that never reads it, so it asserts `commentable` a third time. Idempotence is a property of `withDerivedActionInputs`, not of this one — worth moving there, where it actually bites.

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.utils.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+ * the raw `WorkflowAction` flags, and **no `actionInputs` key at all**.
+ *
+ * Verified against a live instance — `GET /api/v1/workflow/initialactions/contenttype/webPageContent`
+ * returns `commentable: true` with no `actionInputs` property, while
+ * `GET /api/v1/workflow/contentlet/{inode}/actions` returns the same action with
+ * `actionInputs: [{ body: {}, id: 'commentable' }]`.
+ */
+const buildRawAction = (
+    flags: Partial<DotCMSWorkflowAction> = {}
+): Omit<DotCMSWorkflowAction, 'actionInputs'> => ({
```
nicobytes: **Nit** — returning `Omit<DotCMSWorkflowAction, 'actionInputs'>` forces `as DotCMSWorkflowAction` at all 12 call sites below. Declare the return type as `DotCMSWorkflowAction` (still omitting the key in the literal, with the cast done once here) and every call site reads clean.

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.utils.ts
```diff
@@ -0,0 +1,63 @@
...
+ * sent is left untouched: re-deriving it would silently paper over a backend regression in the
+ * endpoints that are supposed to build it themselves.
+ *
+ * @param action the workflow action as it came off the wire
+ * @returns a new action; the argument is never mutated
+ */
+export const withDerivedActionInputs = (action: DotCMSWorkflowAction): DotCMSWorkflowAction => {
+    // No optional chaining on `action` itself: `deriveActionInputs` dereferences it anyway, so a
+    // `?.` here would only move the TypeError one line down while reading as if it were handled.
+    // A missing action is a malformed payload — let it fail loudly rather than half-guard it.
```
nicobytes: **Nit** — three lines explaining why an optional chain *isn't* there. The `deriveActionInputs` body one line down already makes the requirement obvious, and a comment about absent code is pure carrying cost: it has to be re-read by everyone and re-verified by anyone who touches the guard.

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form.component.ts
```diff
@@ -544,6 +544,9 @@ export class DotEditContentFormComponent implements OnInit {
             identifier
         };
 
+        // `DotWorkflowsActionsService` now guarantees `actionInputs` is populated on every
+        // action it returns, including for content with no inode (#36883) — the `= []` is a
+        // belt-and-braces default, not the contract.
```
nicobytes: **Nit** — three lines to say a default is belt-and-braces. Two things make it weaker than it looks: `actionInputs` is a required (non-optional) field on `DotCMSWorkflowAction`, so `= []` was never the contract to begin with; and "the service now guarantees" reads as stale the day a fourth caller appears that does not go through `DotWorkflowsActionsService`. One line, or none.

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.service.spec.ts
```diff
@@ -112,6 +112,120 @@ describe('DotWorkflowsActionsService', () => {
             });
     });
 
+    describe('actionInputs derivation (#36883)', () => {
+        // The default/initial-action endpoints return `WorkflowDefaultActionView`, which wraps the
```
nicobytes, line 117: **Nit** — `rawCommentableAction` is the same 13-field literal as `buildRawAction` in `dot-workflows-actions.utils.spec.ts`, in the same folder. Export the builder from one place — or from `@dotcms/utils-testing`, where the other workflow mocks already live — so the two specs cannot drift apart on what "the raw payload" looks like.

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.service.spec.ts
```diff
@@ -112,6 +112,120 @@ describe('DotWorkflowsActionsService', () => {
...
+            icon: 'workflowIcon',
+            id: 'publish-action-id',
+            name: 'Publish',
+            nextAssign: 'role-id',
+            nextStep: 'step-id',
+            nextStepCurrentStep: false,
+            order: 0,
+            roleHierarchyForAssign: false,
+            schemeId: 'scheme-id',
+            showOn: ['NEW', 'EDITING']
```
nicobytes: **Nit** — `(action: object)` accepts anything and checks no keys. `Partial<DotCMSWorkflowAction>` types it properly and would catch a typo in a fixture at compile time instead of in a confusing assertion failure.

specs/36883-workflow-comments-dialog-new-edit/contracts/derive-action-inputs.md
```diff
@@ -0,0 +1,67 @@
+# Internal Contract: `deriveActionInputs()`
+
+**Plan**: [plan.md](./plan.md) · **Data model**: [data-model.md](./data-model.md)
```
nicobytes: **Nit** — `[plan.md](./plan.md)` is broken twice over: `plan.md` is not committed, and from `contracts/` this relative path resolves to `contracts/plan.md` rather than the sibling directory. Same line in `data-model.md:3`. Either drop the links or commit the plan.

adrianjm-dotCMS:
All 14 applied in `b8e0d8c0f7` — 3 Required, 4 Consider, 7 Nits.

I diverged from the suggestion on two, both explained in their threads: **Required 3** is resolved by withdrawing the claim rather than adding the Java comment, for the ADR-0013 reason you raised yourself; and **Required 1** is resolved in the root `beforeEach` rather than per-mock, because `restoreMocks` doesn't apply to `mockProvider`'s `jest.fn()`s.

Your cross-cutting note landed. Three of your findings were comments that had already drifted before merge, and one of those I had lengthened last round precisely to fix a different inaccuracy. Four comments deleted; the ones left carry real intent — the Java mirror pointer, the `#36883` framing, and why `getByInode` is left alone.

Verification: `data-access` 873 tests, component spec 98, `tsc` clean in the touched files (the 7 pre-existing baseline errors unchanged), format clean. Leaving the threads unresolved for you to close.

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.service.spec.ts
adrianjm-dotCMS: Done. `createFakeWorkflowAction(overrides?)` in `@dotcms/utils-testing`, following the `createFakeContentlet` pattern, with `actionInputs` kept out of the defaults as you asked. Used in both specs, so `buildRawAction` and the duplicated literal are gone. `b8e0d8c0f7`

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form.component.spec.ts
adrianjm-dotCMS: Confirmed, and worse than it looks: this is the same bug I fixed last round, reintroduced by the new tests.

Went with your first option rather than compensating per mock: the defaults for all four (`open`, `containsPushPublish`, `checkPublishEnvironments`, `processWorkflowPayload`) are now set in the root `beforeEach`, and the partial `afterEach` is gone. Order-independent, and it doesn't depend on the next block remembering.

Ruled out `restoreMocks: true` — `mockProvider` creates `jest.fn()`s, not spies, so `restoreAllMocks` wouldn't touch their implementations.

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.utils.ts
adrianjm-dotCMS: Right. `Array.isArray(action.actionInputs)`, exactly as proposed. Added an explicit test for the server-sent `[]` — the one case the guard claimed to cover and didn't. Also updated the guarantees table in `contracts/`.

specs/36883-workflow-comments-dialog-new-edit/contracts/derive-action-inputs.md
adrianjm-dotCMS: Confirmed: 0 `.java` files in the diff, the claim was false.

Took your option 2, for the reason you flagged yourself: a comment-only Java change puts a `.java` file in the diff and, per ADR-0013, moves the PR off the fast CI path for no behavioral gain. The doc now states the actual situation — the pointer exists only TS→Java — and why it isn't added here. The follow-up issue is still owed; I'll open it if you agree.

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.utils.ts
adrianjm-dotCMS: Both asks done: a test that a moveable-only action reaches `openWizard`, and the PR section went from "Push Publish" to "Intentional behavior changes", with a paragraph on `moveable` noting it has no environment gate and therefore cannot block a save.

Thanks for the `setWizardInput()` → `null` detail — I wasn't aware of that asymmetry between the component's `openWizard` and the one in `DotWorkflowEventHandlerService`.

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.service.ts
adrianjm-dotCMS: Applied as suggested, both methods.

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.service.ts
adrianjm-dotCMS: Done, both parts: trimmed to what the helper does plus `#36883`, and `deriveInputsOnEach` moved to `dot-workflows-actions.utils.ts`. The service is back to HTTP plumbing.

Worth saying where it came from: I lengthened that comment last round because another reviewer caught it being **wrong** about `getByWorkflows`. I fixed the accuracy by duplicating the table that already lived in `contracts/`. You were both right.

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.utils.spec.ts
adrianjm-dotCMS: Verified in the code and all three were exactly as you describe: `buildRawAction` never sets those flags (so both `delete`s are no-ops), and `deriveActionInputs` never reads `actionInputs` anywhere in its body. Three tests that could not fail.

Removed. In their place, `describe('withDerivedActionInputs')` with six real tests — derive / preserve / preserve empty / never undefined / no mutation / **idempotence**, which as you noted is a property of that function, not the other one — plus one for `deriveInputsOnEach`. Net: 25 → 28 tests.

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.utils.spec.ts
adrianjm-dotCMS: Solved via the fixture change: `createFakeWorkflowAction` returns `DotCMSWorkflowAction` with the cast done once inside. No casts at the call sites.

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.utils.ts
adrianjm-dotCMS: Removed.

core-web/libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form.component.ts
adrianjm-dotCMS: Removed entirely rather than shortened. Your point that `actionInputs` is non-optional settles it: the default was never the contract, so the comment was defending something that didn't exist.

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.service.spec.ts
adrianjm-dotCMS: Solved by `createFakeWorkflowAction` in `@dotcms/utils-testing`. One definition of "the raw payload", shared by both specs.

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.service.spec.ts
adrianjm-dotCMS: Now `Partial<DotCMSWorkflowAction>`.

specs/36883-workflow-comments-dialog-new-edit/contracts/derive-action-inputs.md
adrianjm-dotCMS: Verified broken twice over, as you said: `plan.md` is gitignored (`.gitignore:229`) and from `contracts/` the path resolved to `contracts/plan.md`. Replaced with links to `spec.md` and `data-model.md`, which are tracked.

nicobytes (approved):

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.utils.ts
```diff
@@ -0,0 +1,81 @@
...
+ * endpoints that are supposed to build it themselves.
+ *
+ * @param action the workflow action as it came off the wire
+ * @returns a new action; the argument is never mutated
+ */
+export const withDerivedActionInputs = (action: DotCMSWorkflowAction): DotCMSWorkflowAction => {
+    // Presence, not length: a server-sent empty array is a legitimate "no inputs" answer and must
+    // survive, otherwise the one regression this guard exists to surface — an endpoint starting to
+    // emit WorkflowActionView with an empty actionInputs — would be silently re-derived over.
+    if (Array.isArray(action.actionInputs)) {
```
oidacra, line 62: `DotCMSWorkflowAction.actionInputs` is declared non-optional in `dot-workflow-action.model.ts:26`, so from TypeScript's point of view this `Array.isArray` check can never be false — yet the function exists precisely because `initialactions` and `defaultactions` do omit the key.

Scope: the guard here is added by this PR; the model declaration is pre-existing.

That leaves the model type describing a shape those two endpoints never send, and the mismatch is invisible outside this file: anything writing `action.actionInputs.length` without a guard compiles clean and throws at runtime on a default-actions payload.

Would it be worth making the type carry that distinction rather than leaving it to this one runtime check? A separate `WorkflowDefaultActionView` type for the un-derived response would express which endpoints omit the field; narrowing this parameter, or marking the field optional on the model, would each work too — the last one surfaces it to every consumer rather than only here.

Or is the intent that the derivation makes the omission un-observable everywhere downstream, so a non-optional field is the accurate description of what consumers now actually receive?

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.utils.ts
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * Keep this in sync with the Java method if a new input type is ever added.
+ *
+ * @param action the workflow action to inspect
+ * @returns the derived inputs, in the same order the Java rule emits them; `[]` when none apply
+ */
+export const deriveActionInputs = (action: DotCMSWorkflowAction): DotCMSWorkflowInput[] => {
+    const actionInputs: DotCMSWorkflowInput[] = [];
+
+    if (action.assignable) {
```
oidacra, line 27: Following on from the `moveable` thread — `assignable` looks like it falls in the same category, and I don't think it has come up yet.

Scope: this line is added by this PR; the `actionInputs.length` gate it feeds is pre-existing.

Because the gate is on array length, every derived input flips a previously-direct-fire action into the wizard path on unsaved content, not just `pushPublish`. For `assignable` that means `openWizard` now renders the role picker in `DotCommentAndAssignFormComponent` on the first click for new content, where it previously fired straight through.

The same reasoning you gave for `moveable` applies — it is parity with saved content, which already gets `assignable` from `getByInode` — so this reads as the right outcome rather than a problem. Should it get the same explicit call-out in the description that `pushPublish` got, and would one test for an assignable-only action on no-inode content be worth adding alongside the commentable ones?

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.utils.ts
adrianjm-dotCMS: Confirmed — `dot-workflow-action.model.ts:26` declares it non-optional, so from the type system's point of view this guard is unreachable, while at runtime it is the line doing the work.

Your last reading is the intent: after the derivation, "always present" describes what consumers actually receive. But it isn't fully true, and the exception is worth naming — **`getByWorkflows` is not derived**. Its pipe is still `map((x) => x?.entity)`, and `POST /schemes/actions/{systemAction}` returns raw `List<WorkflowAction>` (`WorkflowHelper#findActions`), so that path hands back actions with no `actionInputs` while the type says otherwise. Inert today because its only consumer, `dot-workflows-actions-selector-field.service`, selects actions rather than firing them — but the type does lie there.

Keeping the guard as-is: the type system believing it unreachable is evidence the type is wrong, not that the check is redundant.

On fixing the type itself — I'd rather not do it here. Marking the field optional surfaces it to every consumer, which is the honest option but forces guards across code this PR doesn't touch; a separate `WorkflowDefaultActionView` type is cleaner but is a model refactor inside a defect fix. Happy to open a follow-up issue with the two options and the `getByWorkflows` detail, unless you'd rather it land here.

core-web/libs/data-access/src/lib/dot-workflows-actions/dot-workflows-actions.utils.ts
adrianjm-dotCMS: Right on the documentation — `assignable` belongs in that call-out and I missed it. Adding it to the "Intentional behavior changes" section alongside `pushPublish` and `moveable`, with the same reasoning: parity with saved content, which already gets it from `getByInode`, and no gate that can block a save.

On the test: it already exists, and predates this PR — `dot-edit-content-form.component.spec.ts:691`, `'should open wizard when action has assignable input'`, with `actionInputs: [{ id: 'assignable', body: {} }]`. Same shape as the `moveable` one I added, which was needed because `moveable` had no equivalent. Adding another for `assignable` would just duplicate line 691 — let me know if you'd still prefer an explicit no-inode variant.

PR #37452, opened 2026-09-08
feat(spec): add issue resolution specification for SDK packaging version issues

### LINKED ISSUE #36891: SDK packaging: malformed published version strings and floating/orphaned dist-tags in SDK sources and example apps
## Problem Statement

| Source | Value |
|---|---|

Observed consequences:

| File | Field | Value |
|---|---|---|

|---|---|---|

|---|---|---|

```bash
```

### Customer impact

## Steps to Reproduce

```bash
mkdir t && cd t
cat > package.json <<'JSON'
JSON
```

## Acceptance Criteria

**Defect B — floating tags**

**Verification**

## dotCMS Version

## Severity

## Links

KevinDavilaDotCMS:
This commit introduces a new specification document addressing defects in the SDK packaging mechanism, specifically focusing on malformed version strings and incorrect dist-tags. The spec outlines the problem, reproduction steps, expected vs. actual behavior, and the impact on customers, aiming to guide the resolution process for these critical issues.

### Proposed Changes
* change 1
* change 2

### Checklist
- [ ] Tests
- [ ] Translations

### Additional Info

### Screenshots

This PR fixes: #36891


specs/37399-sdk-packaging-version-fix/spec.md
```diff
@@ -0,0 +1,137 @@
...
...ependently resolved by yarn/pnpm, which is what lets those two package managers install a second, mismatched copy of `@dotcms/client` nested under the consumer's own pinned copy — the same failure mode `react-dom` avoids by declaring `react` as a peer instead of a dependency. `main`'s example apps (`nextjs`, `vuejs`) point to `"next"`, an internal dev/QA pre-release tag never meant for customers — wrong regardless of any version-skew question, since it isn't even a released version. And critically, both active LTS branches' own dedicated `examples/nextjs` copies point to a floating tag (`"latest"`) instead of a version pinned to that LTS server, letting the installed SDK run far ahead of it and breaking the scaffolded app — these LTS-branch copies exist specifically to demonstrate against that LTS server, so floating there defeats their purpose. (`main`'s own four examples deliberately tracking `"latest"` going forward — as the correct target for dotCMS Evergreen customers — is an accepted product decision, not a defect; see Fix Scope & Non-Goals.) On two package managers (yarn, pnpm) the SDK library issue above additionally silently overrides a customer's own explicit version pin.
```
rjvelazco: If we're going to remove the regular `dependencies`, we should update the docs for the **Angular**, **React**, and **Vue** SDKs.

Currently we state that running `npm i @dotcms/react` installs all the required dependencies — see:
https://github.com/dotCMS/core/blob/5023d9fa4824c58c1abd206710e9a7b4f633015d/core-web/libs/sdk/react/README.md?plain=1#L101-L109

However, that's no longer necessarily true for all packages or npm versions. We should make it explicit that to use the UVE fully, the following packages need to be installed manually:

- `@dotcms/client`
- `@dotcms/types`
- `@dotcms/uve`

rjvelazco (dismissed):

KevinDavilaDotCMS:
@rjvelazco Thanks for the review — your feedback on the README docs is incorporated, both in the spec and in the implementation:

- **Spec**: added as **AC-009** and reflected in Defect B1/B1b's Fix Scope — the spec now explicitly calls for correcting the `react`/`angular`/`vue` READMEs' "Installation" sections, which claimed installing the package "will automatically install the required dependencies."
- **Implementation** (PR #37477): `react/README.md`, `angular/README.md`, and `vue/README.md` now list `@dotcms/client`, `@dotcms/uve`, and `@dotcms/types` as required manual installs, and call out that yarn classic (1.x) and npm below v7 need this explicitly (they don't auto-install peer dependencies, unlike npm 7+/pnpm) — which is exactly the underlying npm-version-dependent behavior your comment flagged.

This PR is now re-opened for approval since it went through additional changes after your review (LTS root-cause fix, version-pipeline normalization, etc.) — appreciate you taking another look when you get a chance.

rjvelazco (approved):

Issue #37255, opened 2026-08-27
dotAI Security Hardening: RAG permissions, output safety, abuse controls


fmontes:
## Summary

Align the dotAI layer with the platform guarantees the rest of dotCMS already applies: per-user READ permission on retrieval, escaped output at the template boundary, and rate limiting at the REST boundary.

Fixes go at the boundaries, not the call sites, so current and future consumers inherit them.

## Background

`com.dotcms.ai` was built as a feature layer on top of dotCMS and does not route through core enforcement. Retrieval filters on host / content-type / index / language only, viewtools return model output and error detail verbatim, and the REST resources have no rate control. Embedding a content type is an admin decision to make it available *to the AI for permitted users*, and the retrieval path should reflect that.

## Scope

| # | Change | Boundary |
|---|--------|----------|
| #37151 | Per-user READ filter on retrieval; `summarize` uses the filtered list for both prompt and display | `EmbeddingsAPIImpl.getEmbeddingResults` |
| #37153 | Escape/sanitize by default; raw output behind an explicit opt-in accessor | `com.dotcms.ai.viewtool.*` |
| #37154 | Generic error payload to caller, full detail to `Logger` | `CompletionsTool.handleException` |
| #37152 | Labeled, delimiter-escaped data block + system-prompt guard; query stays in `user` role | `CompletionsAPIImpl.buildRequestJson` |
| #37155 | Per-user and per-IP rate limit + concurrency cap, config-overridable, `429` | `CompletionsResource`, `TextResource`, `ImageResource`, `SearchResource` |

## Out of scope

- `AppConfig.debugLogger` prompt/response logging. Explicit default-off flag, working as intended.
- Eliminating prompt injection. #37152 is mitigation; the control for untrusted input stays operational.
- Per-user result-set caching. `EMBEDDING_CACHE` keys on hashed query text to vector, so it is not a per-user leak vector; #37151 adds a comment so that does not regress.
- Agent framework and MCP v2 surfaces. Separate review.
- Provider-side spend caps and rate-limit policy numbers. Customer-owned.

## Epic acceptance

- [ ] Unauthorized user: 0 hits from `/api/v1/ai/search`, "no matching content" from `/api/v1/ai/completions`. Authorized user gets the content.
- [ ] No dotAI code path returns a stack trace to a caller or template.
- [ ] Default viewtool accessor escapes `<script>` / `onerror=` payloads; only the explicit opt-in returns raw.
[...]

Issue #37153, opened 2026-08-20
dotAI: escape viewtool output by default


fmontes:
Output-escaping defaults for the dotAI Velocity viewtools. Scope, acceptance criteria, and implementation notes are tracked internally in `dotCMS/private-issues#675`.

Epic: #37255

fmontes:
Part of epic #37255.

PR #37455, opened 2026-09-08
docs(ai): spec for #37154 stop returning stack traces from AI viewtool exceptions

### LINKED ISSUE #37154: dotAI: Stop returning Java stack traces from AI viewtool exceptions
**Note:** Part of the dotAI security review. Small hardening.

## Description

`CompletionsTool.handleException` returns the **full Java stack trace** to the caller/template, unconditionally:

```java
return Map.of("error", e.getMessage(), "stackTrace", out.toString());
```

There is no dev/prod gate — the trace is always built and returned whenever the viewtool throws. If a template renders that object, internal implementation details (class names, paths, stack frames) are exposed to end users (information disclosure).

Scope note: this issue covers **only** the stack-trace return. The separate debug logging of full prompt/response bodies (`AppConfig.debugLogger`) is gated behind an explicit, default-off flag (`DEBUG_LOGGING` / `AI_DEBUG_LOGGING`) and is considered working-as-intended troubleshooting behavior — out of scope here.

### Affected code
- `com.dotcms.ai.viewtool.CompletionsTool#handleException(Exception)`.

## Acceptance Criteria

- [ ] The viewtool returns a generic, safe error message to the caller/template (no stack trace, no internal paths/class names).
- [ ] Full exception detail is logged server-side via `Logger` for diagnostics.
- [ ] Any other dotAI viewtool exception handlers that surface stack traces are updated the same way.
- [ ] Unit test: a thrown exception yields a generic error payload with no stack trace, and the detail is logged.

## Priority

Low

## Additional Context

Trivial change (~half a day). The debug-logging behavior is intentionally left as-is.

hassandotcms:
Spec only (PR 1 of 2) for #37154. Part of epic #37255 (dotAI Security Hardening).

When a dotAI viewtool call fails, the returned result carries the full Java stack trace and the raw exception message. Any template that prints the result exposes class names, file names and line numbers to site visitors. There is no environment or debug gate.

Proposed fix: one shared exception handler for the `$ai` viewtool that returns a fixed, generic `error` message with no exception-derived text, and logs the full exception server-side via `Logger`. Covers the completions and search sub-tools and `generateImage`. `generateText` rethrows today; the plan verifies whether the rendering layer prints that exception and brings it under the handler only if it does.

Success-path behaviour is unchanged. No API, schema or data changes. Rollback-safe.

Please review `specs/37154-ai-viewtool-stack-trace/spec.md`. On approval, planning and implementation follow in PR 2.

This PR fixes: #37154

### DIFF
--- a/specs/37154-ai-viewtool-stack-trace/spec.md
+++ b/specs/37154-ai-viewtool-stack-trace/spec.md
@@ -1,0 +1,250 @@
+# Issue Resolution Specification: dotAI: Stop returning Java stack traces from AI viewtool exceptions
+
+**Feature Branch**: `37154-ai-viewtool-stack-trace`
+
+**Created**: 2026-09-08
+
+**Status**: Draft
+
+**Type**: Issue / Bug Resolution
+
+**Related GitHub Issue**: [dotCMS/core#37154](https://github.com/dotCMS/core/issues/37154) — part of epic [dotCMS/core#37255](https://github.com/dotCMS/core/issues/37255) (dotAI Security Hardening)
+
+**Input**: User description: "https://github.com/dotCMS/core/issues/37154"
+
+<!--
+  This is the dotCMS ISSUE-RESOLUTION spec (used by /speckit-specify-fix). Unlike the
+  feature spec, it is framed around a defect: what is wrong, how to reproduce it, and how
+  we will know it is fixed. It still flows into /speckit-plan, where the Legacy Impact and
+  ADR Alignment gates apply. Keep this technology-light — root-cause and fix details are
+  refined in the plan.
+-->
+
+## Problem Statement *(mandatory)*
+
+The dotAI Velocity viewtool (`$ai`, registered in `toolbox.xml`) exposes sub-tools for
+completions and semantic search. When a call to one of these sub-tools fails, the tool does not
+throw. It returns a result object that carries the failure, and that object includes the
+**full Java stack trace** of the exception alongside the exception message.
+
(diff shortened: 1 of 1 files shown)


ihoffmann-dot (dismissed):
No critical or blocking findings. Verified against `CompletionsTool`, `SearchTool`, and `AIViewTool`; the leak behavior described matches the actual code.

Two notes to sanity-check before approval:

- **AC-005 (`generateText`)** stays uncovered if planning confirms no real leak there. Intentional tradeoff, but worth epic-owner sign-off since it's the only one of the four surfaces excluded by default.
- **Merge-order dependency on #37153** (touches the same three classes). Spec's reasoning holds, just flagging the collision risk.

No ADR in `platform-adrs` overlaps this area.

ihoffmann-dot (approved):
Approved :white_check_mark:

Clean fix. The viewtools no longer put any exception detail (stack trace, message, class name) in what they return to templates.

- All failure paths in `CompletionsTool`, `SearchTool` and `AIViewTool.generateImage` go through `AIViewToolErrorHandler`: templates get one generic `error` message, and the full exception is logged on the server.
- `SearchTool` now parses its arguments inside the handled block, and `related(ContentMap, ...)` is wrapped too, so bad input can't throw straight into Velocity.
- Existing templates keep working: `JSONObject` implements `Map`, so `$result.error` still resolves.
- Good test coverage: a unit test for the handler, integration tests for every changed path using sentinel-gated WireMock stubs, and the live-render check confirming `generateText` doesn't leak anything into the page.

PR #37457, opened 2026-09-08
chore(users): #37298 follow-ups — roles-by-id, inline roles, Angular cleanups, design polish

### LINKED ISSUE #37298: chore(users): follow-ups from #36990 / #37082 / #37083 (roles-by-id, list N+1, Angular-rule cleanups)
## Overview

Consolidated follow-up work surfaced during code review of the four Users portlet PRs (profile #36990, roles #37082, permissions #37083, api-tokens #37084). Two backend counterparts have already landed / are landing (#37218 merged, #37236 open); the remaining items are pure frontend.

Verified against `origin/main` on 2026-08-31 — every referenced symbol / file / line is still present.

Parent epic: #34736.

### 1. Roles tab — send role IDs on save (backend counterpart: #37218, MERGED)

`PUT /v1/users` now accepts role IDs alongside `roleKey`s (each entry resolves by key first, then by ID). Byte-identical for existing key-only payloads.

- **File:** \`libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-roles-tab/dot-users-roles-tab.component.ts:156\` — \`grantIdentifier(role)\` returns \`role.roleKey || role.id\`. With the new endpoint, drop the fallback and emit \`role.id\` directly.
- **File:** \`libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.ts:597\` — \`mergeRoleKeysForSave\` filters out keyless roles from the outbound payload. The filter is no longer needed once the shuttle emits IDs.
- **Keep the personal-role filter** (\`roleKey === userId\`) — that's a different guard against a backend \"cannot alter users on this role\" 400 on \`editUsers=false\` roles.
- The change is confined to the roles tab and the shell's merge helper; profile-tab's Access-toggle save path is unaffected (it echoes back the fetched role list unchanged).

### 2. Users list — drop the per-row N+1 role fetch (backend counterpart: #37236, OPEN — apply once merged)

Today the list store fires \`GET /v1/roles/users/{userId}\` per row (1 + N per page load), re-running on every sort change and debounced keystroke. #37236 adds \`includeRoles=true\` to \`GET /v1/users/filter\`; each row gains \`roles: [{id, name, roleKey}]\` for the user's directly assigned roles (inherited excluded — legacy Roles-tab parity).

- **File:** \`libs/portlets/dot-users/src/lib/dot-users-list/store/dot-users-list.store.ts:109\` — \`usersService.getUserRoles(user.userId)\`. Replace the \`forkJoin(roleFetches)\` chain with a single call to \`getUsersPaginated\` carrying \`includeRoles: true\`.
- Chip derivation reads from \`row.roles\` directly instead of the parallel results map.
- The store still owns the network layer (rule 2 in the Angular project rules memory).

### 3. `dot-users-replacement-picker` — remove `OnPush` + extract HTTP into a component-scoped store

Both items live on the same file — batch them into one commit.

- **File:** \`libs/portlets/dot-users/src/lib/components/dot-users-replacement-picker/dot-users-replacement-picker.component.ts\`
- **Line 4 + 37:** drop the \`ChangeDetectionStrategy\` import and the \`changeDetection: ChangeDetectionStrategy.OnPush\` declaration — \`OnPush\` is default in Angular 22.
- **Line 82–107 (constructor):** the in-component \`Subject\` + \`switchMap(usersService.getUsersPaginated(...))\` pattern violates \"HTTP calls belong in a Store\". Extract into a component-scoped \`SignalStore\` (\`providers: [DotUsersReplacementPickerStore]\`) with methods for \`search(query)\` and state for \`suggestions / isLoading / hasError\`.
- The picker's public API (inputs / outputs) stays the same — consumers don't need to change.

### 4. `dot-users-profile-tab` — remove `OnPush`

- **File:** \`libs/portlets/dot-users/src/lib/dot-users-create/tabs/dot-users-profile-tab/dot-users-profile-tab.component.ts:2 + 68\`
- Drop the \`ChangeDetectionStrategy\` import and \`changeDetection: ChangeDetectionStrategy.OnPush\` declaration.

### 5. Shell — SCSS → Tailwind (drop `::ng-deep`)

- **File:** \`libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.scss\`
- Eliminate the \`:host ::ng-deep { p-tabs, p-tabpanels, p-tabpanel { display: flex; ... } }\` block. Layout should land on the host class + the tabpanel wrappers (utilities land on the right element instead of fighting the PrimeNG preset via \`::ng-deep\`).
- If the file empties out, delete it and remove the \`styleUrl\` from the shell component.

### 6. `dot-role-permissions-iframe` (dot-roles portlet) — hygiene pass

- **File:** \`libs/portlets/dot-roles/src/lib/dot-roles-page/components/dot-role-permissions-iframe/dot-role-permissions-iframe.component.ts:34–40\`
- Drop \`standalone: true\` (default) and \`changeDetection: ChangeDetectionStrategy.OnPush\` (default).
- Update the stale JSDoc reference on line 30 — \`DotPermissionsIframeDialogComponent\` in \`@dotcms/ui\` no longer exists; it was consolidated into \`DotJspIframeDialogComponent\` by Jalinson Diaz (\`c105b14837\`).
- **Optional but recommended:** the file still hand-rolls the same-origin guard via a UUID regex; the more permissive \`isSameOriginRelativeUrl\` from \`@dotcms/utils\` is already used by \`dot-users-permissions-tab\` and \`dot-jsp-iframe-dialog\`. Consider composing the two guards (UUID check + shared predicate) so the shared predicate stays the single source of truth for \"is this URL safe to bypass\".

## Explicitly out of scope

- **Signal Forms migration** — dropped. PrimeNG v21 doesn't implement the required control contracts, and no plan to migrate until there's a broader repo-wide need.
- **Extracting a shared user picker component** — de-scoped. The two consumers use visually different pickers (\`p-autoComplete\` vs \`p-listbox\` in \`p-popover\`); shared UI extraction doesn't fit. The remaining overlap is search HTTP, which item 3 covers.
- **Extracting a shared permissions-iframe component** — de-scoped. Decision on main is to keep both consumers portlet-local (see JSDoc on \`dot-users-permissions-tab.component.ts:44–49\`); only the same-origin guard is shared, and that already lives in \`@dotcms/utils\`.

## Sequencing

- **Item 1** is unblocked — #37218 is merged. Do first.
- **Item 2** is unblocked once #37236 merges. Track that PR.
- **Items 3, 4, 5, 6** are independent and can land in any order. Reasonable to bundle 3+4+5 (shell-adjacent) in one PR and 6 (dot-roles) in a separate one to keep review scopes clean.

## References

- Backend PRs: #37218 (merged), #37236 (open)
- Original review threads: #36990, #37082, #37083, #37084
- Project rules memory: internal \`feedback_angular_project_rules.md\`

AP2300:
Closes #37298.

Wraps up the six follow-up items from the split Users-portlet migration
(#36990 / #37082 / #37083 / #37084) plus the design review pass.

## Follow-up items (from #37298)

1. **Roles tab saves by role ID, not `roleKey`** — since #37218 the users
   endpoint accepts both; sending IDs unconditionally round-trips
   keyless custom roles cleanly.
2. **Inline roles per list row (kills the N+1)** — list load now sends
   `includeRoles=true` per #37236 contract; the Roles column reads the
   inline `roles` array on each row instead of fanning out one
   `getUserRoles` call per user. 403 retry drops `includeRoles=false`
   and leaves the column empty rather than failing the whole load.
3. **Angular-rule cleanups** across the portlet + the roles iframe —
   modern control flow, signal inputs, default change detection.
4-6. Related cleanups tracked in #37298.

## Design polish pass

- **API tokens tab**: standard portlet-table typography + per-row
  skeleton loading; empty/error states via `<dot-empty-container>`;
  ghost/circular Material Symbols copy button in the reveal dialog;
  primary "Accept" close button; create-mode hint mirrors the
  permissions tab (icon + copy on the tab body).
- **Roles tab**: granted panel switched to `<p-tree>` so both panels
  render identical PrimeNG chrome (checkboxes, hover, selection);
  empty granted state uses the shared `<dot-empty-container>` with a
  `shield_person` material symbol.
- **Create/Edit dialog**: dropped the internal avatar/name/status
  header block — the user's full name now rides in the dialog title
  bar in edit mode. Removed dead computeds, imports, tests, and
  orphaned i18n keys that only powered the removed block.
- **Users list**: search input width matches content-drive; toolbar
  height matches spec; selection ↔ New button animation restored
  without inflating row height; secondary text swept to
  `text-slate-500`.
- Language.properties swept clean of the orphaned keys.

## Known gap (separate ticket)

**Users list sort headers are decorative.** The frontend sends
`orderBy`/`direction` correctly, but `UserPaginator.getItems` ignores
both arguments and `UserFactoryLiferayImpl.getUsersByName` hard-codes
`ORDER BY first_name || ' ' || last_name ASC`. Backend fix is being
tracked separately; frontend already sends the right values so flipping
the backend on will just work.

## Test plan

- [x] `pnpm nx test portlets-dot-users-portlet` — 177 passing
- [x] `pnpm nx lint portlets-dot-users-portlet` — clean
- [x] Open the dialog in create mode — no avatar/name header; title
      says "Create User"
- [x] Open the dialog in edit mode — no avatar/name header; title
      shows the user's full name
- [x] Roles tab — granted panel matches the available side visually;
      empty granted state shows the shield_person empty container
- [x] API tokens tab — table typography matches other portlets;
      reveal dialog copy icon works (icon flips to check briefly);
      create-mode shows the key-icon hint on the tab body
- [x] Users list — search width matches content-drive; toolbar height
      matches spec

This PR fixes: #37298


core-web/libs/portlets/dot-users/src/lib/components/dot-users-replacement-picker/store/dot-users-replacement-picker.store.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import { inject } from '@angular/core';
+
+import { catchError, switchMap, tap } from 'rxjs/operators';
+
+import { DotUserListItem, DotUsersService } from '../../../services/dot-users.service';
+
+export interface DotUsersReplacementPickerState {
+    suggestions: DotUserListItem[];
+    isLoading: boolean;
```
KevinDavilaDotCMS: dont use isLoading or similar, use "state" and that represent "loading", "error", "idle", etc etc

Some helpful doc: https://dev.to/tehaisperlis/the-loading-anti-pattern-7jj

AP2300: Good catch — fixed in 64d57cf2ba. Swapped the picker store's `isLoading` + `hasError` booleans for a single `status: ComponentStatus` field (`IDLE` / `LOADING` / `LOADED` / `ERROR`) — the same shape the rest of the dotCMS SignalStores use, and the exact anti-pattern that article calls out. The component's `$isLoading` / `$hasError` are now computed derivations off `status()` so the template and the 177 tests keep passing unchanged.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.ts
```diff
@@ -568,46 +570,68 @@ export class DotUsersCreateComponent {
...
+        const base = this.currentRoleIds() ?? this.#store.roles().map((role) => role.id);
         const accessKeys = new Set<string>(Object.values(ACCESS_ROLE_KEYS));
-        const nonAccess = base.filter((key) => !accessKeys.has(key));
+        const accessRoleIds = new Set(
+            this.#store
+                .roles()
+                .filter((role) => accessKeys.has(role.roleKey))
+                .map((role) => role.id)
+        );
+        const nonAccess = base.filter((id) => !accessRoleIds.has(id));
```
adrianjm-dotCMS, line 659: **Access roles granted from the Roles tab bypass the Access toggles.**

`accessRoleIds` is built only from the roles the user **already had when the dialog opened** (`store.roles()` is the initial fetch). The old key-based strip (`accessKeys.has(key)`) applied to every entry in the base list.

Scenario: user is not a CMS Admin. The admin grants "CMS Administrator" from the Roles-tab shuttle — that puts the role **id** into `base`. That id isn't in `accessRoleIds`, so it's never stripped, and the Access toggle is still OFF so nothing re-adds it either. The save grants CMS Administrator while the Access tab says otherwise. With the toggle ON you also send both the id and the roleKey for the same role (the `Set` can't dedupe id vs key).

The three access roles have constant keys (`ACCESS_ROLE_KEYS`), so the toggle should stay the single source of truth: strip base entries that resolve to those roles even when they weren't in the initial fetch — or don't offer the three access roles in the shuttle at all.

core-web/libs/portlets/dot-users/src/lib/dot-users-list/dot-users-list.component.html
```diff
@@ -149,13 +263,39 @@
                             {{ user.lastLoginDate | date: 'MM/dd/yyyy' }}
                         }
                     </td>
+                    <td (click)="$event.stopPropagation()" data-testid="users-cell-actions">
+                        @if ($isPushPublishEnabled()) {
+                            <p-menu
+                                #rowMenu
+                                [model]="getRowMenuItems(user)"
```
adrianjm-dotCMS: This feature is outside the stated scope

Row kebab, bulk actions, `DotAddToBundleComponent`, `DotPushPublishDialogService`, two new resolvers on `lib.routes.ts`, four new i18n keys — ~150 lines of new end-user functionality. None of it appears in #37298 (six items, none push-publish related), in the PR description, or in the test plan. Could you document it in the description or split it out?

Two details inside it:
- Inconsistent gating: the bulk toolbar still shows "Add to Bundle" on non-EE (the `@else` above), but this kebab is hidden entirely when `$isPushPublishEnabled()` is false — so without a license you can bundle via selection but not from the row. Intentional? And should Add to Bundle be available at all without EE?
- `[model]="getRowMenuItems(user)"` rebuilds the `MenuItem[]` on every change-detection cycle, per row. A computed or a cached map per user avoids that.

core-web/libs/portlets/dot-users/src/lib/services/dot-users.service.ts
```diff
@@ -180,6 +203,17 @@ export interface DotUsersPaginatedParams {
...
     roleKey?: string;
+    /**
+     * Ask the backend to inline each row's directly assigned roles as
+     * `roles: DotUserListRole[]` (#37236). Additive — omitting the flag
+     * leaves the response byte-identical to previous versions.
+     *
+     * `includeRoles=true` requires CMS Administrator or Roles + Users
+     * portlet access. Callers without that gate get a 403 on the whole
+     * request; the list store handles that by retrying without the
+     * flag and falling back to the per-row `getUserRoles` fetch.
```
adrianjm-dotCMS: **Stale comment on the public contract.**

This and the `roles?` docblock above (line ~51) both describe a fallback the store no longer performs: `fetchUsersPage` retries once without `includeRoles` and leaves the Roles cells empty — which is the right behavior, and exactly what the store's own comment says. These point the next reader at an N+1 that was just deleted.

Same kind of drift in `dot-users-list.component.ts`: the `USER_ASSET_PREFIX` docblock describes a `users_` form (empty suffix) meaning "the current selection", resolved against the session by the backend. Nothing in the code produces that.

core-web/libs/portlets/dot-users/src/lib/dot-users-list/dot-users-list.component.html
AP2300: Thanks for calling this out — the scope call was intentional but you're right that it's undocumented on the PR. Here's the context, and both of your in-scope points are now fixed on `5ea04fae06`.

**Why the push-publish / bundle / kebab work is here**

Design surfaced these as part of the same review pass that produced the polish sweep at the top of this PR — the four-tab Users dialog opened up questions about the *list* view's action affordances (there was no per-row entry point at all in the legacy portlet, and no bulk actions on selection). The design team asked for:

- Row kebab with the two publishing commands
- Bulk toolbar: Push Publish + Add to Bundle + Delete on selection
- Standard hover-reveal + selection-visibility pattern used by the modernized portlets

These are additive follow-ups to #37298's original six items — I'll add a note on the parent issue right after this reply so the addition is on the record and the next reviewer isn't surprised.

**Your two inline points are fixed:**

1. **Gating inconsistency** — you were right that non-EE could bundle via bulk but not from a row. Cross-checked the reference portlet (`dot-plugins-list.component.ts:135`): `Add to Bundle` is available on every instance (bundles are plain archives, no license needed); only `Push Publish` is enterprise-gated. Row kebab now always renders; `menuItemsFor()` conditionally *adds* the Push Publish entry only when EE + at least one PP environment are present. Row ↔ bulk gating matches.

2. **`[model]` rebuild per CD tick** — good catch, that was a real allocation on every cycle. Now cached per-user in `#rowMenuItemsCache: Map<userId, MenuItem[]>`, invalidated by an `effect(() => { store.users(); … })` so a list reload never hands PrimeNG a stale command closure. Added a reference-stability assertion in the spec to lock it in.

Diff: `5ea04fae06`. 178 tests pass.

core-web/libs/portlets/dot-users/src/lib/dot-users-create/dot-users-create.component.ts
AP2300: Good catch — real bug, fixed on `e7c4533955`. Went with the second half of your suggestion ("don't offer the three access roles in the shuttle at all") because the first half — "strip base entries that resolve to those roles even when they weren't in the initial fetch" — needs a full role catalog to resolve arbitrary ids back to keys, and the shell doesn't have one (only the fetched-per-user roles). Filtering at the source is the only fix that actually closes the hole.

**How it's wired:**

- `DotUsersRolesTabComponent` gains an `excludedRoleKeys: input<readonly string[]>([])` input.
  - Matching leaves are dropped from Available.
  - `isGrantableLeaf` returns false for them, so a parent's bulk-select won't try to grant them either — no ambiguity when Root has one CMS-Admin leaf and one custom leaf.
- The shell hands the tab `ACCESS_ROLE_KEYS_LIST = ['CMS Administrator', 'DOTCMS_BACK_END_USER', 'DOTCMS_FRONT_END_USER']` and also strips those ids from the shuttle seed (`initialGrantedRoleIds` / `currentRoleIds`), so a pre-existing admin doesn't see a stuck row that the tab would then filter out of Available.
- Kept the shell's initial-fetch-based strip in `mergeRoleIdsForSave` as belt-and-suspenders for the "shuttle never touched" flow — comment reworded to say why both halves are needed.

**Tests added:**

- `should hand the three access-role keys to the shuttle as 'excludedRoleKeys'` — pins the module-level list matches the ACCESS_ROLE_KEYS constant.
- `should send exactly one identifier per access role even if the shuttle also carries the id` — covers your dedup point: when initial fetch had `role-back` and the Backend toggle is ON, the payload has exactly `['DOTCMS_BACK_END_USER']`, not `['role-back', 'DOTCMS_BACK_END_USER']`.
- `should hand the three access-role keys to the shuttle` + `should drop initial-fetch access-role IDs from the shuttle seed` — cover the seed shape.
- Roles-tab spec: `excludedRoleKeys drops matching leaves from the tree entirely` — verifies both Available and `canSelectRole` respect the input.

182 tests pass, prod build clean.

core-web/libs/portlets/dot-users/src/lib/services/dot-users.service.ts
AP2300: Both docblocks fixed on `e7c4533955`. You were right on both counts:

- **`roles?` field + `includeRoles` param docs**: rewrote to describe what the store actually does today — a 403 downgrade to `includeRoles=false` that leaves Roles cells empty for the caller who can't see them, no per-row `getUserRoles` fan-out. The old N+1 language pointed at deleted code.
- **`USER_ASSET_PREFIX` docblock**: dropped the "empty suffix = current selection" story. Grepped the FE — nothing produces the `users_` (empty) form; the only consumer is `${USER_ASSET_PREFIX}${user.userId}` in `openRow*` / `openBulk*`. If the backend's session-scoped selection ever needed a FE entry point we'd wire it explicitly.

adrianjm-dotCMS (approved):

Issue #37462, opened 2026-09-08
Hardening: a single unserializable field aborts indexing of the whole contentlet


fabrizzio-dotCMS:
> **Hardening, not a live bug.** There is no known path on trunk that reaches this today —
> #37272 removed the one trigger that did. What remains is the shape of the error handling, which
> costs the whole document whenever any future per-field failure occurs. Filed as
> `Type : Refactoring` for that reason; if a live trigger turns up it should be re-labelled a
> defect.

## Problem Statement

`ESMappingAPIImpl.loadFields` wraps the entire per-field loop body in one catch that logs at WARN and **rethrows**:

```java
// ESMappingAPIImpl.java:1143-1147
} catch (Exception e) {
    Logger.warn(ESMappingAPIImpl.class, "Error indexing field: " + field.getFieldName()
            + " of contentlet: " + contentlet.getInode(), e);
    throw new DotDataException(e.getMessage(), e);
}
```

Rethrowing aborts `toMap`, so **one field failing removes the entire contentlet from the index** — every other field of that contentlet included. The content is in the database and invisible to search, with a single WARN as the only record.

The failure is also lost from **both engines at once, in every migration phase**: the document is rendered once and shared across providers (`ContentletIndexAPIImpl.mapContentletForProcessor` — *"Compute mapping once; reuse across all providers for the same contentlet"*), so the throw happens before the fan-out. It is not an Elasticsearch-only or OpenSearch-only defect.

The `Date` branch a few lines above already shows the intended shape — it catches, degrades to `toString()`, and keeps going.

**Severity**: Medium. No known trigger on trunk after #37272, but the structure means any future per-field failure costs the whole document rather than the field.

### Relationship to #37272

#37272 fixed one specific trigger: a `TextField` on a numeric column handed a `String` to `DecimalFormat.format()`. That path no longer reaches this catch — `loadNumericField` converts, or omits the field and logs a WARN.

This issue is the **surrounding structure**, which #37272 deliberately left alone (see its spec's non-goals). Other paths still reach it: a corrupt `Date`, `ESUtils.sha256` throwing on a unique field, unreadable binary metadata, a Category with a dangling reference.

## Steps to Reproduce

No supported path reproduces this on trunk today — that is why it is filed as a structural defect rather than an active bug. It is reachable by any field whose serialization throws something the branch does not anticipate.

Historical reproduction, from #37272 before the fix:

[...]

Issue #37432, opened 2026-09-07
Phase Sweep MainSuite 1b fails at OS Phase 3: #37276 index tests are not phase-aware and blank the indicies pointer table


fabrizzio-dotCMS:
## Problem Statement

The weekly **Scheduled OpenSearch Phase Sweep** fails in the `Integration Tests - MainSuite 1b [OS Phase 3]` shard. The two integration test classes added by PR #37320 (for issue #37276) are not phase-aware: they hardcode `setPhase(0)` and `setPhase(1)`, which is false in a job whose environment was bootstrapped at **Phase 3**.

At Phase 3, `pointES` is gated off, so the `indicies` table carries **no Elasticsearch pointers at all**. Flipping `FEATURE_FLAG_OPEN_SEARCH_PHASE` mid-JVM does not retroactively provision ES indices, so the ES leg resolves `null` index names.

Three distinct symptoms, all deterministic across the four Failsafe retries:

| Test | Phase forced | Failure |
|---|---|---|
| `ContentletIndexPartialFailurePhaseTest.test_dualWritePhase_shadowFailureDoesNotReachCaller` | 1 | `NullPointerException: index must not be null` (line 101) |
| `ContentletIndexProviderSkipTest.test_shadowProviderSkip_stillCompletesTheRemoval` | 1 | `DotRuntimeException: ... the primary provider (ContentletIndexOperationsES) resolved no active index ... The removal was NOT performed.` (line 119) |
| `ContentletIndexPartialFailurePhaseTest.test_ordinaryWrites_areUnaffectedByTheEscalation` | ambient (3) | Passes run 1, then `NullPointerException: index must not be null` on runs 2–4 (line 71) |

The NPE stack is identical in every case:

```
java.lang.NullPointerException: index must not be null
  at org.elasticsearch.client.core.CountRequest.indices(CountRequest.java:92)
  at ContentFactoryIndexOperationsES.getCountRequest(ContentFactoryIndexOperationsES.java:454)
  at ContentFactoryIndexOperationsES.indexCount(ContentFactoryIndexOperationsES.java:446)
  at ESContentFactoryImpl.indexCount(ESContentFactoryImpl.java:1607)
```

The production code is doing exactly what #37276 changed it to do — raising instead of silently swallowing an unresolvable index. **The test premise is wrong for a Phase 3 environment, not the code.**

### The second, worse half: a test can blank the cluster's index pointer record

[...]

PR #37463, opened 2026-09-08
fix(index): a Text field on a numeric column no longer makes the contentlet unindexable (#37272)

### LINKED ISSUE #37272: A TextField stored in a numeric column makes the whole contentlet unindexable

fabrizzio-dotCMS:
Closes #37272. **PR 2 of 2** in the Spec-Kit flow — the implementation. The spec is PR #37393 (approved by @dario-daza); until it merges this PR's diff also shows the spec commit, which is the shared ancestor rather than a duplicate.

## The defect

`ESMappingAPIImpl.loadFields` picks its serialization branch from the **storage column** (`field_contentlet`), not the declared field type. dotCMS allows a `TextField` to be backed by an `integerN`/`floatN` column, and its own built-in types do it — `htmlpageasset.sortOrder` is an `ImmutableTextField` with `DataTypes.INTEGER`. So the value in hand may be a `String`, and handing it to `DecimalFormat.format()` threw:

```
java.lang.IllegalArgumentException: Cannot format given Object as a Number
    at ESMappingAPIImpl.loadFields(ESMappingAPIImpl.java:1118)
    at ESMappingAPIImpl.toMap(ESMappingAPIImpl.java:465)
```

Because the per-field catch rethrows, **the whole contentlet was lost from the index** — every other field included. And from both engines at once, in every migration phase: the document is rendered once before the provider fan-out (`ContentletIndexAPIImpl.mapContentletForProcessor` — *"Compute mapping once; reuse across all providers"*).

## The fix

`loadNumericField` now owns that branch:

| Value in hand | Emits |
|---|---|
| `Number` | the **identical pre-existing path** — same object, same formatter call |
| `String` that converts (`"54"`, `"54.3"`) | same entries a natively-stored number produces, padding included |
| anything else | **both keys omitted**, one WARN naming field and content type |

Conversion is `NumberUtil.toLongOrEmpty` / `toFloatOrEmpty` — new, and they **report failure instead of substituting a default**, because the caller must be able to omit rather than fabricate.

Routing a `Number` straight through is deliberate: it makes "no correctly-stored document moves" true by construction rather than dependent on the converter being equivalent.

## Why both keys must go together

This is the part worth reviewing closely — it is not stylistic.

`toMap` derives `<field>_dotraw` from `<field>_text`, and **falls back to the bare key when `_text` is absent, unpadded**. So writing the numeric key alone produces a malformed `_dotraw`. And `_dotraw` is a `keyword` that *every* sort resolves through — `addBuilderSort` appends the suffix to whatever field name the caller passes. Fixed-width zero padding is therefore the only reason ordering by a numeric field is numeric rather than alphabetical:

- raw text: `"N/A"` sorts after every padded digit string (`'N'` = 0x4E > `'9'` = 0x39)
- unpadded number: `54` sorts before `9`

Either way a listing silently reorders, with no error anywhere. The chain runs from `loadFields` through `toMap`'s post-processing to `addBuilderSort` and is invisible from any single file, so it is now written down in `docs/backend/INDEX_FIELD_EMISSION.md` — it had already been broken once, in the spec's own first revision.

## Omit, not zero

The numeric key is skipped rather than defaulted to `0`: a `0` is indistinguishable from a genuine `0` in range queries, sorts and aggregations, with only a WARN to reveal the difference. A false in-range match is a correctness bug; an unset field is an honest absence. Recorded as decision C-3 in the spec, with the rejected alternative.

## Tests

**26/26 unit** over the converters — signs, surrounding whitespace, already-a-`Number`, blanks, overflow, fractional-in-integer (rejected, not truncated), and the non-finite `NaN`/`Infinity` that `Float.parseFloat` accepts but JSON cannot carry.

**15/15 integration** over the emission contract via `toMap`:

- both columns, integer and float, with convertible and unconvertible values
- the `_dotraw` invariant — asserted as **absence** of the key, since `assertNotEquals(rawText, …)` would pass against the unpadded-number bug
- emitted runtime class matches the native path (`Integer 54` is not `equals` to `Long 54`)
- one WARN naming field and content type, and **not** the value, which is customer content
- nothing logged at all on the happy path
- the neighbouring branches of the edited `else if` chain — boolean, date, general text
- unique field: the SHA-256 entry is skipped, not computed over nothing
- **end to end through a real index**: convertible values sort numerically asc and desc, and an unconvertible document does not answer `[0 TO 10]`

Registered in `MainSuite1b` and deliberately phase-agnostic — it exercises document construction only, never calls `setPhase(...)` and never assumes an index name resolves, so the weekly OpenSearch Phase Sweep runs it without reproducing #37432.

Run:
```
./mvnw test -pl :dotcms-core -Dtest=NumberUtilTest -Dmaven.build.cache.enabled=false
./mvnw install -pl :dotcms-core --am -DskipTests
./mvnw verify -pl :dotcms-integration -Dcoreit.test.skip=false -Dit.test=ESMappingAPINumericFieldTest
```

## Deliberately out of scope

**The surrounding per-field catch still rethrows** — any *other* unserializable field still costs the whole document. Now tracked in #37462. Left out on purpose: it changes how indexing fails in the hottest path in indexing, converting a loud failure (document absent) into a quiet one (document present, one field short), and forcing an unexpected serialization failure has no test seam. That belongs in its own cycle with its own review.

Also out, per the spec's non-goals: the index mapping generator (rollback-unsafe, forces a reindex), repairing existing mis-typed stored values, content-type-save validation (it already permits this modelling, and forbidding it would break dotCMS's own built-in types at bootstrap), and `ImportStarterUtil` — the one ingestion path that can create this data, named as a known gap.

## Rollback safety

Not rollback-unsafe. No mapping, schema, API-contract or serialized-state change. For existing good data the emitted document is unchanged, so old and new pods emit identical documents during a rolling deploy. For the currently-failing contentlets the change is additive — documents that were absent start appearing; a rollback returns them to absent, which is where they are today. No data repair needed: a normal reindex recovers them.

## Review focus

The both-keys-together pairing and the omit-vs-`0` call. Everything else follows from them.


ihoffmann-dot:
Approved ✅ 

The fix correctly isolates the `DecimalFormat`-on-String failure into `loadNumericField`, converting best-effort via the new `NumberUtil.toLongOrEmpty`/`toFloatOrEmpty` and omitting both the numeric key and its `_dotraw` companion together when conversion fails.

Consistent with the existing branch structure in `ESMappingAPIImpl` and doesn't touch the index mapping generator, so no rollback risk.

Test coverage is thorough: unit + integration tests covered. 
Ready to merge.

ihoffmann-dot (approved):

PR #37468, opened 2026-09-08
DO NOT MERGE: #34154: refactor(json) batch with windowFixed instead of a shared accumulator

fabrizzio-dotCMS:
## Why this exists

A proof of concept for the Lunch & Learn (#34154): what a stream gatherer actually buys on real
dotCMS code, rather than on a toy example. **DO NOT MERGE** — the change is small and green, but it
touches an upgrade task and belongs behind its own review if we ever want it.

## The shape it replaces

`PopulateContentletAsJSONUtil` batched its inserts and updates through a mutable
`Collection<Params>` that was declared in the outer method and passed *down* into the per-record
method. That method appended to it and flushed when it filled; the caller then had to remember a
second flush for the remainder:

```java
private void processInsertRecord(String inode, String json,
        Collection<Params> paramsInsert, MutableInt totalInsertAffected) {

    paramsInsert.add(new Params(inode, json));

    if (paramsInsert.size() >= MAX_BATCH_SIZE) {     // the cut
        this.doInsertBatch(paramsInsert, totalInsertAffected);
    }
}
```

```java
for (var jsonData : jsonDataArray) {
    this.processInsertRecord(jsonData._1(), jsonData._2(), paramsInsert, totalInsertAffected);
}
if (!paramsInsert.isEmpty()) {                       // the leftover, 130 lines away
    this.doInsertBatch(paramsInsert, totalInsertAffected);
}
```

Four places, and duplicated again for updates: the buffer, the cut, the leftover, and the
`clear()` hidden in the executor's `finally`. Keeping them in step was a convention.

## What it becomes

```java
static void inBatchesOf(final List<Params> params, final Consumer<List<Params>> executor) {

    params.stream()
            .gather(Gatherers.windowFixed(MAX_BATCH_SIZE))
            .forEach(executor);
}
```

`windowFixed` emits the short final window on its own, so **the cut and the leftover are the same
thing**. The accumulator is gone, `processInsertRecord` is gone, and `processUpdateRecord` becomes
the pure mapper `toUpdateParams`.

**Net −8 lines** (+41/−49).

## The part worth noticing

`doInsertBatch` and `doUpdateBatch` used to end in `finally { params.clear(); }` — because the
caller reused one buffer. A window is **unmodifiable**, so that line would now throw. It had to go,
and its removal is the honest summary of the change: those methods never owned the buffer, they
only appeared to. One test pins that:

```java
assertThrows(UnsupportedOperationException.class, () -> batches.get(0).clear());
```

## Tests

Six unit tests, no database, `PopulateContentletAsJSONUtilBatchingTest` — pinning the behaviour the
old code got right by convention:

| | |
|---|---|
| partial final batch is emitted | 450 records → 200 / 200 / 50 |
| exact multiple emits no empty tail | 400 → 200 / 200 |
| fewer than one batch still emits | 7 → one batch of 7 |
| empty input never calls the executor | 0 → no batches |
| every record delivered once, in order | 601 records, flattened, compared |
| the window rejects mutation | `UnsupportedOperationException` |

**6/6 green**; `test-compile -pl :dotcms-core --am` passes.

## Not in scope

- The other 12 `Lists.partition` sites. `OSIndexAPIImpl:855` materialises a list purely to cut it
  up and is the next obvious one, but it has no leftover to forget, so it makes a weaker point.
- `ContentTypeDestroyAPIImpl:380` calls `Lists.partition(..., 1)` — partitioning into batches of
  one. That does not want a gatherer, it wants the line deleted.
- No behaviour change intended: same batch size, same SQL, same order.

Refs #34154

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/util/content/json/PopulateContentletAsJSONUtil.java
+++ b/dotCMS/src/main/java/com/dotcms/util/content/json/PopulateContentletAsJSONUtil.java
@@ -36,7 +36,9 @@ import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
+import java.util.stream.Gatherers;
 
 /**
  * Utility class to populate the contentlet_as_json column in the contentlet table.
@@ -276,7 +278,6 @@ public class PopulateContentletAsJSONUtil {
                                  final MutableInt totalRecords
     ) throws SQLException, DotDataException {
 
-        final Collection<Params> paramsInsert = new ArrayList<>();
         final MutableInt totalInsertAffected = new MutableInt(0);
 
         var foundData = false;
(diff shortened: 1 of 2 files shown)


fabrizzio-dotCMS:
Superseded by the BrowserAPIImpl version, which shows both halves of the gatherer API (windowFixed + mapConcurrent) on one class and surfaces a real cost: mapConcurrent has no timeout.

PR #37474, opened 2026-09-09
fix(network): handle IPv6 addresses in IPUtils checks

swicken:
### What

Adds IPv6 handling to the address classification and CIDR matching in `IPUtils`.

`IPUtils` evaluated addresses with an IPv4-only matcher against an IPv4-only default
list, so IPv6 addresses were not classified correctly. This makes the classification
family-aware and lets `REMOTE_CALL_SUBNET_BLACKLIST` accept IPv6 CIDRs.

### Changes

- `isIpPrivateSubnet` resolves with `InetAddress.getAllByName` and evaluates every
  resolved address in both families, rather than only the first. That is also the
  resolver used for outbound connections, so classification and connection agree.
- Classification uses the JDK predicates for unspecified, loopback, link-local,
  site-local and multicast, plus an explicit `fc00::/7` match, which the JDK does
  not expose.
- `isIpInCIDR` gains IPv6 prefix matching. Parsing is literal-only, so it triggers
  no DNS lookups. The IPv4 path and both short circuits are unchanged, and
  cross-family comparisons do not match.

### Configuration compatibility

`REMOTE_CALL_SUBNET_BLACKLIST` keeps its existing meaning. When the property is set it
defines the policy on its own and the built-in checks are skipped, so a deployment that
narrows the list to permit specific internal ranges continues to work unchanged. The
built-in checks apply only when the property is unset.

A deployment that sets an IPv4-only list therefore gets no IPv6 coverage. That is the
cost of honouring the override and is unchanged from current behaviour, and it now logs
a warning naming the ranges to add.

### Reviewer note

`MonitorHelper.isAccessGranted` uses `isIpInCIDR` as an allowlist. An IPv6 CIDR in
its ACL previously failed to parse and therefore never matched; it now matches as
configured, which is the documented intent of the property.

### Testing

`IPUtilsTest` covers IPv6 loopback, unique-local, link-local, IPv4-mapped and
unspecified addresses, IPv6 CIDR matching, and cross-family non-matching. Verified
failing before the change and passing after. `MonitorHelperTest` run alongside for
regression. Also exercised against a running instance to confirm public
destinations, including public IPv6, remain reachable.

Closes dotCMS/private-issues#677

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/util/network/IPUtils.java
+++ b/dotCMS/src/main/java/com/dotcms/util/network/IPUtils.java
@@ -2,6 +2,7 @@ package com.dotcms.util.network;
 
 import java.net.Inet6Address;
 import java.net.InetAddress;
+import java.util.Arrays;
 import java.util.Objects;
 import java.util.concurrent.atomic.AtomicBoolean;
 import com.dotcms.repackage.org.apache.commons.net.util.SubnetUtils;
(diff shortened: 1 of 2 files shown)


wezell (approved):
Love this.  Needed it for a long time.

PR #37483, opened 2026-09-09
fix(browser): bind the MIME type filter value in the browser query

swicken:
### What

Binds the MIME type filter value in the browser query builder instead of placing it into
the statement text, and validates it where the query is built.

`appendMIMETypeQuery` used `String.format` to put each caller-supplied MIME type into the
statement. Every sibling filter in the same builder already binds its values through the
parameter list; this method was the only one that did not.

### Changes

- The JSONPath expression is bound as a parameter,
  `jsonb_path_exists(c.contentlet_as_json, ?::jsonpath)`. Verified against PostgreSQL that
  the bound form returns identical rows to the previous form, for both plain and wildcard
  filters.
- `BrowserQuery.Builder#showMimeTypes` restricts values to MIME type characters and throws
  `IllegalArgumentException` otherwise, which maps to a 400. Validating in the builder
  covers all five call sites (`BrowserResource`, `ContentDriveHelper`, `BrowserAjax` twice
  and `BrowserAPI`) rather than relying on each one to validate its own input. The rejected
  value is not echoed into the message or the log.
- `appendMIMETypeQuery` is now static and takes the parameter list, matching the sibling
  filters and making it unit-testable without an `APILocator` bootstrap.

### Behaviour preserved

Regex metacharacters in the value are deliberately not escaped, so the emitted predicate
matches exactly what it matched before. Escaping would change the behaviour of the wildcard
forms callers send, such as `image/*`.

### Known constraint, worth a second opinion

A MIME type carrying parameters, for example `text/plain; charset=utf-8`, is now rejected.
No stored `contentType` in the file metadata has that shape on the instances I checked, but
this is a narrowing and I would like a reviewer to confirm it against real data.

### Testing

New `BrowserAPIMimeTypeQueryTest` covers parameter binding, the `application/dotpage`
structure-type branch, multiple values binding in order, the accepted character set
including wildcard and punctuated types, and rejection of values carrying characters that
are meaningful in the surrounding SQL and JSONPath. `BrowserAPIImplTest` and
`BrowserAPIWorkflowQueryTest` run alongside for regression, 20 tests total.

Verified on a running instance that `image` and `image/avif` still return the expected
asset and `application/pdf` returns none.

### Unrelated pre-existing issue found while testing

`mimeTypes: ["image/*"]` returns nothing end to end, even though the SQL predicate matches.
`filterReturnList` post-filters with a literal `mimeType.contains(mType)`, and
`"image/avif".contains("image/*")` is false. That predates this change and is untouched by
it, so it is not fixed here.

Closes dotCMS/private-issues#676

This PR fixes: #676


dotCMS/src/main/java/com/dotcms/browser/BrowserQuery.java
```diff
@@ -315,6 +316,31 @@ public static final class Builder {
...
+         * MIME types, the partial and wildcard forms the file browser sends, and the parameter
+         * forms that appear in stored metadata such as {@code text/plain; charset=iso-8859-1}.
+         *
+         * <p>Covers the RFC 6838 token characters plus {@code ;}, {@code =} and space for
+         * parameters. Deliberately excludes {@code "} and {@code \}, which are the only
+         * characters that could terminate or escape the quoted regex literal the value is placed
+         * inside, and the grouping characters {@code ( ) [ ] { } | ?}, which would allow a caller
+         * to build a pattern with catastrophic backtracking.</p>
+         */
+        private static final Pattern MIME_TYPE_PATTERN =
```
wezell, line 329: Not sure this is enough protection and not sure we can sanitize SQL via just a regex.  The fix is to loop over the mimeTypes add the parameterized ? conditions when building the SQL and then send the loop over the mimetypes again when building the prepared statement.

wezell (changes requested):

dotCMS/src/main/java/com/dotcms/browser/BrowserQuery.java
swicken: That is what the code does now, so I think the concern may be reading the validation rather
than the SQL builder. `appendMIMETypeQuery` loops the MIME types, appends a `?` condition per
value, and adds the bound value in the same loop:

```java
predicates.add("jsonb_path_exists(c.contentlet_as_json, ?::jsonpath)");
parameters.add(String.format(MIME_TYPE_JSONPATH, mimeType));
```

The caller value never reaches the statement text. I verified against Postgres that the
published probe payload executes `pg_sleep(3)` in 3.11s when concatenated, and returns a
jsonpath syntax error in 0.12s when bound.

You are right on the principle though, and there is a residual I should be straight about.
The value is still interpolated into the JSONPath expression that gets bound, and the regex
is what keeps it inside that expression's quoted literal. So regex is not protecting the SQL,
but it is protecting the JSONPath. `like_regex` requires a literal pattern, so the value
cannot be bound as a jsonpath variable.

It can be removed entirely by moving the match to SQL, which makes the jsonpath a constant and
the value a true bind parameter with no interpolation anywhere:

```sql
EXISTS (SELECT 1 FROM jsonb_path_query(c.contentlet_as_json,
          '$.fields.**.metadata.contentType') ct
        WHERE ct #>> '{}' LIKE ? ESCAPE '\')
```

I tested this and it returns the same rows for `image`, `image/*`, `text/plain` and the
parameterised `text/plain; charset=iso-8859-1`, and the hostile payload is inert as data. It
also costs nothing in index terms, since there is no GIN index on `contentlet_as_json`; the
only index on that column is the template expression index, so both forms scan either way.

Two things change and are worth a decision. `*` has to be translated to `%` and any literal
`%` or `_` escaped, and LIKE is case sensitive where the current regex is too, so that part
stays equivalent.

Happy to switch it to that form if you prefer it, which would make the character validation
pure input hygiene rather than a security control.

---

*Posted by Claude on Scott's behalf.*

wezell (approved):
Thanks for the explanation.  As long as we are passing this to the SQL in a parameterized way

PR #37485, opened 2026-09-09
feat(experiments): the page's experiments as a panel beside the canvas (#37478)

### LINKED ISSUE #37478: Experiments Portlet — the page's experiments as a UVE panel, behind FEATURE_FLAG_EXPERIMENTS_PORTLET
## Description

With `FEATURE_FLAG_EXPERIMENTS_PORTLET` on, the Universal Visual Editor's Experiments entry point currently **ejects the editor**: the `science` nav item navigates to `/experiments?pageId=…&language_id=…` (`dot-ema-shell.component.ts:222-247`), the iframe is torn down, and the editor gets back to the page through the breadcrumb or the back arrow. That is what #37005 shipped, and it is already better than the legacy per-page screens it replaces — but it still costs the page.

This issue keeps the editor on the page: the experiments for the page in hand render **in a panel beside the canvas**, behind the same flag. No new switch, no second rollout to coordinate.

It also settles what #37005 left half-done. FR-021c asks for the page filter to be "visible and clearable". The filter is applied and honoured, but there is no in-UI affordance to clear it, and clearing it would be a dead end anyway: the portlet is opt-in (declared in `portlet.xml`, no `UpgradeTask` adds it to a layout), so on a stock instance `GET /api/v1/menu` carries no `experiments` entry — verified, 43 items, `hasExperiments: false`. An editor arriving from UVE has no path to a site-wide list. In a page-scoped panel that requirement changes shape: the panel **is** the page's scope by construction, and "clearable" becomes an explicit link out to the full portlet.

## What exists today (prior art in UVE)

Four patterns, all live in this template, with their real costs:

| Pattern | Where | Mount cost | Chunk cost |
| --- | --- | --- | --- |
| CSS-collapsed, always mounted | left palette `edit-ema-editor.component.html:21`; right edit panel `:222` (widths `clamp(280px,15vw,320px)` and `clamp(360px,20vw,500px)`, `.scss:25,95`) | mounted on **every** UVE EDIT load, open or not | static import, sits in the editor chunk |
| `@if`-gated overlay drawer | `dot-block-editor-sidebar` (`p-drawer`), `:307-312` | mounts/destroys per open | static import |
| `@defer (when …)` + inner `@if` | Edit Content side panel, `:321-333` | mounts/destroys per open | **separate chunk, fetched on first open only** |
| Routed child of the shell | today's legacy experiments, `lib.routes.ts:93-113` | full route activation; `router-outlet` swap replaces the editor | lazy `loadChildren` |

The right edit panel also shows the tab variant: the wrapper is always mounted, but each tab's contents are `@if ($editorPanelActiveTab() === n)`-gated, so contents mount only while their tab is selected.

## Recommendation

**`@defer (when $experimentsPanel())` + inner `@if`, in a drawer** — the Edit Content side panel pattern, for four reasons:

1. The experiments code never enters the UVE editor chunk. The editor chunk is on the hot path for every page edit; experiments concern a minority of pages.
2. Zero cost when the panel is never opened, which is the common case.
3. Mount/destroy per open makes the panel's open/close its store's `onInit`/`onDestroy` — no stale view state, no live effects or `location` subscriptions while closed.
4. It is the newest pattern in this exact template and already carries its rationale in a comment there.

The nav item becomes an **action**, not a destination: `href` drops, and it routes through `handleItemAction` (`dot-ema-shell.component.ts:458`) like `page-tools` and `properties` already do. That also removes the `$activeHref()` highlight for it and makes the `MenuGuardService` exemption on `/experiments` (`app.routes.ts:180-198`) unnecessary for this path — the panel needs no route at all.

**Two of the three screens, in two phases.** The screens are very different shapes:

- **Phase 1 - the list.** A compact panel list (see item 2 below); rows lead into Configure.
- **Phase 2 - Configure, in the panel.** The largest screen in the lib (492K of source vs the list's 120K) and a multi-card form: Details, Goal, Variants, Scheduling, Traffic. It needs a width decision of its own before it is built - the cards are laid out for a full-width column and the Variants card is a table. Two candidate answers: a wider panel for this screen only (the Edit Content side panel already varies its own width, 80% to 100%, `dot-edit-content-side-panel.component.html`), or a single-column reflow of the cards. **Neither is chosen here.** One issue with two phases rather than two issues, because the panel's plumbing - the store's URL independence, the flag read, the resolvers - is shared, and phase 2 changes none of it.
- **Not Results.** It pulls `chart.js` + `primeng/chart` (`dot-experiments-reports-chart.component.ts:1,5`), the heaviest dependency in the portlet, for charts that need width. It stays full-screen in the portlet, opened from Configure as it is today.

Phase 1 has to ship usable on its own: a panel that can only list, with Configure still opening full-screen, is a smaller improvement than the whole thing but not a broken one.

## Why not the alternatives

- **CSS-collapsed always-mounted (palette/edit-panel pattern).** Puts the experiments chunk in the editor bundle *and* mounts the store on every UVE load — which, with the store's current `onInit`, means a health check, a full experiments fetch and a bulk page lookup fired behind a closed panel on every page edit. Non-starter.
- **A third tab in the existing right edit panel.** Mount cost is fine (tab contents are `@if`-gated), but the chunk cost is not unless the tab body is itself `@defer`ed, and the panel is 360-500px wide. It is also selection-scoped — it answers "the contentlet I clicked" — while an experiments list is page-scoped. Wrong drawer.
- **A routed child of the shell (what the legacy screens do).** `router-outlet` swaps the editor **out**, which is the exact cost this issue exists to remove. Route resolvers would run and UVE's URL would have to change, tearing down the iframe.

## What has to change (the real work)

This is not a template change. Five of these are load-bearing.

**1. The list store owns the URL; inside UVE it cannot.** `dot-experiments-list.store.ts` hydrates from `route.snapshot.queryParams` (`:598`), mirrors **every** view-state change back with `Location.go(createUrlTree(…, {queryParamsHandling:'merge'}))` (`:583,624`), and re-hydrates on popstate (`:610`). UVE writes its own params through `#updateLocation` (`dot-ema-shell.component.ts:663`) with `createUrlTree([], {queryParams})` — **no merge**, so it replaces the whole query string. Two `Location.go` writers on one address: the panel's `filter=`/`statuses=`/`page=` land in the UVE URL, and UVE's next view change silently drops them. The panel needs a mode where the store does not own the address — view state in memory, seeded from inputs (`pageId`, `languageId`) instead of query params. This is the largest piece of work here and it is a store change, not a component one.

**2. The list table does not fit a sidebar.** `LIST_TABLE_STYLE` is `min-width: 81rem` (1296px, `constants.ts:110`) across seven sortable columns. It cannot render at 360-500px, and an overlay wide enough for it would cover the page — a dialog with a different animation, not a sidebar. The panel needs a **compact row rendering**: name plus status, the `page` column dropped (the panel *is* the page), goal/schedule/modDate collapsed into a subline or dropped. The store is reused; the table is not.

**3. Fetch the page's experiments, not all of them.** The list calls `getAllUnfiltered()` — every experiment on every site — then resolves each distinct `pageId` through a bulk `htmlpageasset` lookup and narrows client-side (`dot-experiments-list.store.ts:96,112,144`). A page-scoped panel should call the `getAll(pageId)` that already exists (`dot-experiments.service.ts:66`) and skip the page lookup entirely, because UVE already holds `pageAsset()`. Three requests plus a full-set payload become one scoped request. Coordinate with #37007 (server-side list contract), which owns the same swap point.

**4. Reading the flag must not fail open.** The natural move — adding `FEATURE_FLAG_EXPERIMENTS_PORTLET` to `UVE_FEATURE_FLAGS` (`shared/consts.ts:114`) so it rides the batch `getFeatureFlags` request UVE already makes on init — **inverts the default**: `withFlags` maps `FEATURE_FLAG_NOT_FOUND` to `true` (`with-flags.feature.ts:42`). #37005 refused exactly that for this flag, which sits next to the visitor-facing kill switch. Either keep the fail-closed `readExperimentsPortletSwitch` (`experiments-portlet-switch.util.ts`) — the shell already makes that request for the nav item, so the panel adds none — or give `withFlags` a fail-closed variant. Do not join the batch silently.

**5. Resolvers the panel will not have — narrower than first thought.** The list route provides and resolves `DotPushPublishEnvironmentsResolver`, but nothing consumes it: `dot-experiments-list.component.ts` never reads route data, Push Publish opens through `DotPushPublishDialogService` with only an `assetIdentifier`, and `dot-add-to-bundle` fetches its own bundles. So Push Publish and Add to Bundle need no route plumbing in the panel (spec D9). What does still hold is the analytics health gate: the list defers its first fetch until `/api/v1/experiments/health` reports OK (`checkHealth()`), one extra request per open unless it is cached at UVE init.

**6. Breadcrumbs belong to UVE.** The three screens push crumbs onto `GlobalStore` through `putCrumbOnTrail` (`dot-experiments-breadcrumb.util.ts`). In a panel the trail is UVE's page's; the list must push nothing while it renders as a panel.

**7. `#37008` gets simpler, not harder.** With the panel as the entry point, the migration also deletes the `experiments` routed child (`edit-ema/lib.routes.ts:93-113`) and the legacy branch of the nav item, and no full-page eject survives anywhere.

## Performance budget

Measurable, and the reason for the pattern choice:

- The UVE editor chunk must not grow (build stats before/after).
- A UVE page load where the panel is never opened must fire **zero** experiments requests — no list, no health check, no page lookup.
- Opening the panel must fire **one** experiments request for the page (plus the health check, if it is not cached at UVE init).
- Closing the panel must leave no live effect or `location` subscription behind.

## Must not break

- Flag **off**: the legacy per-page screens and the `experiments/{pageId}` nav destination, untouched (`dot-ema-shell.component.ts:245`).
- The variant Edit Content round-trip from #37005: `Edit Content` on a variant opens UVE for that variant and returns to `/experiments/:id/configuration` for the originating experiment (`dot-uve-toolbar.component.ts:370-418`).
- The full portlet at `/experiments`, reachable and unfiltered from the main navigation.
- UVE's own URL: the panel writes nothing to it.

## Acceptance Criteria

- [ ] With the flag on, the UVE Experiments nav item opens a panel beside the canvas; the page stays rendered and the iframe is not reloaded.
- [ ] With the flag off, behavior is unchanged from today (specs + E2E regression).
- [ ] The panel lists only the current page's experiments, correct for zero (empty state with a way to create one), one, and many.
- [ ] The panel offers an explicit way into the full portlet (FR-021c's "clearable", honestly resolved).
- [ ] A row leads to Configure: in the panel once phase 2 lands, full-screen in the portlet before that.
- [ ] Configure in the panel edits, autosaves and transitions exactly as the full-screen screen does - same store, same form, same rules; only the layout differs.
- [ ] Results opens full-screen from Configure, and returns to the page the editor came from.
- [ ] Nothing the panel does writes to UVE's URL, and nothing UVE does resets the panel's view state.
- [ ] The panel pushes no breadcrumb of its own.
- [ ] Push Publish and Add to Bundle work from a panel row, or are deliberately absent with a stated reason.
- [ ] A UVE load that never opens the panel fires no experiments request; the editor chunk does not grow.
- [ ] The flag still fails closed on a missing key and on a failed read.

## Priority

Medium. It improves an entry point that already works; nothing is broken without it.

## Additional Context

- Depends on: #37005 (the flag, the filtered destination and the round-trip all land there).
- Related: #37007 (server-side list contract — same swap point as item 3), #37008 (migration; item 7).
- Spec to seed: `specs/37005-experiments-uve-integration/spec.md` D2 records why the destination is the page-filtered list, which is the premise this panel inherits.
- Out of scope: Results inside the panel; any change to how UVE edits, saves, locks or renders content; any change to how experiments are served to visitors; removing the flag or the legacy screens (#37008).

oidacra:
Closes #37478

With `FEATURE_FLAG_EXPERIMENTS_PORTLET` on, the Universal Visual Editor's Experiments entry point currently ejects the editor: the `science` nav item navigates to the portlet, the iframe is torn down, and the editor gets back to the page through the breadcrumb. This keeps the editor on the page — the experiments for the page in hand render in a panel beside the canvas, behind the same flag. No new switch, no second rollout.

It also closes what #37005 (flagged UVE entry point) left half-done. Its FR-021c asked for the page filter to be "visible and clearable"; the filter is applied but has no affordance to clear it, and clearing it would be a dead end — the portlet is opt-in, so on a stock instance the admin menu carries no Experiments entry to return to. In a page-scoped panel there is no filter to clear: the panel is the page's scope by construction, and "clearable" becomes an explicit link out to the full portlet.

### Proposed Changes

**The flow, end to end**

* The Experiments nav item becomes an **action, not a destination** — no navigation, no route, no address change. A panel opens beside the canvas and the page stays rendered.
* **A page with no experiments opens straight on creation.** An empty list is not a resting state — on an untested page there is exactly one useful next step.
* **A page with experiments opens on that page's list**, and the editor picks one to configure.
* **The page is fixed** in creation and configuration. No page picker, no editable page field, no change-page flow — the editor is standing on the page, so a control offering to move the experiment elsewhere contradicts the surface it renders on.
* **Preview / Edit Content on a variant** leaves for the page as that variant, with a banner naming the variant and the experiment and saying whether it is editable. **One control returns and reopens the panel on the same experiment's configuration** — not the list, not a restarted sequence.
* **Results stay full-screen**, opened from configuration as they are today.

**Two phases, one PR**

| Phase | Scope |
| --- | --- |
| 1 | The list and the creation entry point. A row's Configure still opens full-screen. |
| 2 | Configuration in the panel, including the variant round-trip. |

Phase 1 ships usable on its own, and carries the shared plumbing below — phase 2 changes none of it.

**The load-bearing work**

* **The list store stops owning the URL in panel mode.** It hydrates from query params, mirrors every view-state change back with `Location.go(…{queryParamsHandling:'merge'})` and re-hydrates on popstate. UVE writes its own params with no merge, replacing the whole query string. Two writers on one address: the panel's state gets silently dropped by UVE's next view change. In the panel the view state lives in memory, seeded from `pageId`/`languageId`. This is the largest piece of work here, and it is a store change.
* **One page-scoped request.** The list calls `getAllUnfiltered()` — every experiment on every site — then resolves each distinct `pageId` through a bulk `htmlpageasset` lookup and narrows client-side. The panel calls the existing `getAll(pageId)` and skips the lookup, because UVE already holds `pageAsset()`. Three requests plus a full-set payload become one. Same swap point as #37007 (server-side list contract).
* **The flag read stays fail-closed.** Adding it to `UVE_FEATURE_FLAGS` would invert the default — `withFlags` maps `FEATURE_FLAG_NOT_FOUND` to `true`. The dedicated `readExperimentsPortletSwitch` stays, and the shell already makes that request for the nav item, so the panel adds none.
* **One implementation of each screen, two presentations.** The portlet's list and configuration components are adapted to render as both portlet and sidebar — no parallel copy. `LIST_TABLE_STYLE`'s `min-width: 81rem` (1296px across seven columns) cannot render at panel width, so the layout is what becomes mode-dependent; the state and the rules are shared. The page column drops in panel mode.
* **Lazy, without hurting fluidity.** The experiments code never enters the UVE editor chunk. A UVE load that never opens the panel fires zero experiments requests. The panel responds to the gesture immediately in a loading state; the page never stutters or reloads while it loads.
* **Loading, error, empty and read-only are their own requirements.** Because "nothing to show" now moves the editor to creation rather than resting, none of the others may be mistaken for it — a failed load is never an empty page, and a misconfigured analytics app says so.
* **No breadcrumb.** Inside UVE the trail belongs to the page.

### Must not break

* **Flag off**: the legacy per-page screens and the `experiments/{pageId}` destination, untouched — and the flag still fails closed on a missing key and on a failed read.
* The variant Edit Content round-trip from #37005 (flagged UVE entry point).
* The full portlet at `/experiments`, reachable and unfiltered from the main navigation, keeping its table, its columns and its address-backed view state.
* UVE's own URL: the panel writes nothing to it.

### Additional Info

Stacked on `issue-37005-experiments-uve-integration` (#37256), which is still open — that branch is the base, not `main`.

The specification is the first commit here (`specs/37478-uve-experiments-panel/`): 47 functional requirements, 16 success criteria, 7 user stories, 13 resolved decisions. Implementation follows in this same PR.

**One correction to the issue.** #37478 item 5 states that the list route's `DotPushPublishEnvironmentsResolver` is read by the row menu, and that a panel mounted outside the router would break Push Publish silently. It is not read — `dot-experiments-list.component.ts` never touches route data, Push Publish opens through `DotPushPublishDialogService` with only an `assetIdentifier`, and `dot-add-to-bundle` fetches its own bundles. That resolver is a prefetch nothing consumes (contrast `dot-locales-list.component.ts:75`, which does read it). Both actions therefore work from a panel row with no route plumbing. Recorded as decision D9; the issue has been corrected. The health-gate half of item 5 does hold and is covered.

**One decision deliberately left open (O1).** The panel's width for phase 2. Configuration is a multi-card form laid out for a full-width column and its Variants card is a table, so it does not fit phase 1's width. Two candidates — a wider panel for that screen only (the Edit Content side panel already varies its own width, 80% to 100%), or a single-column reflow of the cards. It blocks nothing in phase 1.

Related: #37007 (server-side list contract, same swap point), #37008 (migration — this makes it simpler: the `experiments` routed child and the legacy branch of the nav item both go, and no full-page eject survives). Epic: #36763 (Experiments: A/B Testing v2).

### Checklist
- [ ] Tests
- [ ] Translations

This PR fixes: #37478

### DIFF
--- a/core-web/libs/portlets/dot-experiments/data-access/src/lib/dot-experiments-panel.store.spec.ts
+++ b/core-web/libs/portlets/dot-experiments/data-access/src/lib/dot-experiments-panel.store.spec.ts
@@ -1,0 +1,327 @@
+import { createServiceFactory, SpectatorService } from '@openng/spectator/vitest';
+
+import { signal, WritableSignal } from '@angular/core';
+
+import { DotExperimentsPanelStore } from './dot-experiments-panel.store';
+
+const PAGE_A = 'page-a';
+const PAGE_B = 'page-b';
+const EXPERIMENT_A = 'exp-a';
+const EXPERIMENT_B = 'exp-b';
+
+const LANGUAGE_EN = 1;
+const LANGUAGE_ES = 2;
+
+/**
+ * The panel's view state and mode signal (#37478).
+ *
+ * Everything here is asserted against
+ * `specs/37478-uve-experiments-panel/contracts/panel-store.contract.md`. Two of these tests
+ * exist because the behaviour they pin is invisible when it breaks:
+ *
+ * - `close()` and `suspendForVariant()` are two different closes. Collapsing them into one with
+ *   a boolean is the "simplification" that silently turns the variant round trip into a reset,
+ *   and FR-023 is what it would break.
+ * - A language change must transition nothing. Refetching or resetting on it would discard the
+ *   editor's view state for a set of experiments that is provably identical, because
+ *   `DotExperiment` carries `pageId` and no language at all.
+ */
+describe('DotExperimentsPanelStore', () => {
+    let spectator: SpectatorService<DotExperimentsPanelStore>;
+    let store: DotExperimentsPanelStore;
+
+    let pageId: WritableSignal<string | null>;
+    let languageId: WritableSignal<number | null>;
+
+    const createService = createServiceFactory({
+        service: DotExperimentsPanelStore
+    });
+
+    /**
+     * Creates the store and hands it the editor's live context, as the shell does.
+     *
+     * Signals rather than values on purpose: the contract requires the page and language be read
+     * live, so a captured value cannot go stale against the canvas (D10).
+     */
+    const initStore = (page: string | null = PAGE_A, language: number | null = LANGUAGE_EN) => {
+        spectator = createService();
+        store = spectator.service;
+
+        pageId = signal(page);
+        languageId = signal(language);
+
+        store.setContext({ pageId, languageId });
+        spectator.flushEffects();
+    };
+
+    describe('initial state', () => {
+        it('starts closed, on the list, with no experiment', () => {
+            initStore();
+
+            expect(store.isOpen()).toBe(false);
+            expect(store.view()).toBe('list');
+            expect(store.experimentId()).toBeNull();
+        });
+
+        it('exposes the page and language it was given', () => {
+            initStore(PAGE_A, LANGUAGE_ES);
+
+            expect(store.pageId()).toBe(PAGE_A);
+            expect(store.languageId()).toBe(LANGUAGE_ES);
+        });
+    });
+
+    describe('open()', () => {
+        it('opens on the list', () => {
+            initStore();
+
+            store.open();
+
+            expect(store.isOpen()).toBe(true);
+            expect(store.view()).toBe('list');
+            expect(store.experimentId()).toBeNull();
+        });
+
+        it('opens on the list even when the panel was last showing a configuration', () => {
+            initStore();
+
+            store.open();
+            store.showConfigure(EXPERIMENT_A);
+            store.close();
+            store.open();
+
+            expect(store.view()).toBe('list');
+            expect(store.experimentId()).toBeNull();
+        });
+    });
+
+    describe('navigating between views', () => {
+        it('shows the creation screen', () => {
+            initStore();
+            store.open();
+
+            store.showCreate();
+
+            expect(store.view()).toBe('create');
+            expect(store.experimentId()).toBeNull();
+        });
+
+        it('shows a configuration, holding the experiment it is for', () => {
+            initStore();
+            store.open();
+
+            store.showConfigure(EXPERIMENT_A);
+
+            expect(store.view()).toBe('configure');
+            expect(store.experimentId()).toBe(EXPERIMENT_A);
+        });
+
+        it('shows results, holding the experiment they are for', () => {
+            initStore();
+            store.open();
+
+            store.showResults(EXPERIMENT_A);
+
+            expect(store.view()).toBe('results');
+            expect(store.experimentId()).toBe(EXPERIMENT_A);
+        });
+
+        it('returns to the list and forgets the experiment', () => {
+            initStore();
+            store.open();
+            store.showConfigure(EXPERIMENT_A);
+
+            store.backToList();
+
+            expect(store.view()).toBe('list');
+            expect(store.experimentId()).toBeNull();
+        });
+    });
+
+    describe('openResults() — the toolbar badge', () => {
+        it('opens the panel straight onto that experiment’s results', () => {
+            initStore();
+
+            store.openResults(EXPERIMENT_A);
+
+            expect(store.isOpen()).toBe(true);
+            expect(store.view()).toBe('results');
+            expect(store.experimentId()).toBe(EXPERIMENT_A);
+        });
+
+        it('replaces what the panel was showing instead of stacking a second surface', () => {
+            initStore();
+            store.open();
+            store.showConfigure(EXPERIMENT_A);
+
+            store.openResults(EXPERIMENT_B);
+
+            expect(store.isOpen()).toBe(true);
+            expect(store.view()).toBe('results');
+            expect(store.experimentId()).toBe(EXPERIMENT_B);
+        });
+    });
+
+    describe('close()', () => {
+        it('closes and discards the view state', () => {
+            initStore();
+            store.open();
+            store.showConfigure(EXPERIMENT_A);
+
+            store.close();
+
+            expect(store.isOpen()).toBe(false);
+            expect(store.view()).toBe('list');
+            expect(store.experimentId()).toBeNull();
+        });
+
+        it('leaves nothing for a later open to show as current', () => {
+            initStore();
+            store.open();
+            store.showResults(EXPERIMENT_A);
+            store.close();
+
+            store.open();
+
+            expect(store.view()).toBe('list');
+            expect(store.experimentId()).toBeNull();
+        });
+    });
+
+    describe('the variant round trip', () => {
+        it('suspendForVariant() closes the panel but keeps where the editor was', () => {
+            initStore();
+            store.open();
+            store.showConfigure(EXPERIMENT_A);
+
+            store.suspendForVariant();
+
+            expect(store.isOpen()).toBe(false);
+            expect(store.view()).toBe('configure');
+            expect(store.experimentId()).toBe(EXPERIMENT_A);
+        });
+
+        it('resumeFromVariant() reopens on the same experiment’s configuration', () => {
+            initStore();
+            store.open();
+            store.showConfigure(EXPERIMENT_A);
+            store.suspendForVariant();
+
+            store.resumeFromVariant();
+
+            expect(store.isOpen()).toBe(true);
+            expect(store.view()).toBe('configure');
+            expect(store.experimentId()).toBe(EXPERIMENT_A);
+        });
+
+        it('does nothing when the panel was never suspended', () => {
+            initStore();
+
+            store.resumeFromVariant();
+
+            expect(store.isOpen()).toBe(false);
+            expect(store.view()).toBe('list');
+            expect(store.experimentId()).toBeNull();
+        });
+
+        it('does nothing when the panel was closed rather than suspended', () => {
+            initStore();
+            store.open();
+            store.showConfigure(EXPERIMENT_A);
+            store.close();
+
+            store.resumeFromVariant();
+
+            expect(store.isOpen()).toBe(false);
+            expect(store.view()).toBe('list');
+            expect(store.experimentId()).toBeNull();
+        });
+
+        it('a close after a suspend discards the suspended state', () => {
+            initStore();
+            store.open();
+            store.showConfigure(EXPERIMENT_A);
+            store.suspendForVariant();
+
+            store.close();
+            store.resumeFromVariant();
+
+            expect(store.isOpen()).toBe(false);
+            expect(store.view()).toBe('list');
+            expect(store.experimentId()).toBeNull();
+        });
+    });
+
+    describe('the page on the canvas changes', () => {
+        it('re-scopes to the new page, back on the list', () => {
+            initStore();
+            store.open();
+            store.showConfigure(EXPERIMENT_A);
+
+            pageId.set(PAGE_B);
+            spectator.flushEffects();
+
+            expect(store.pageId()).toBe(PAGE_B);
+            expect(store.view()).toBe('list');
+            expect(store.experimentId()).toBeNull();
+        });
+
+        it('leaves the panel open while it re-scopes', () => {
+            initStore();
+            store.open();
+
+            pageId.set(PAGE_B);
+            spectator.flushEffects();
+
+            expect(store.isOpen()).toBe(true);
+        });
+
+        it('does not reopen a panel the editor had closed', () => {
+            initStore();
+
+            pageId.set(PAGE_B);
+            spectator.flushEffects();
+
+            expect(store.isOpen()).toBe(false);
+        });
+
+        it('discards a suspended round trip that belonged to the old page', () => {
+            initStore();
+            store.open();
+            store.showConfigure(EXPERIMENT_A);
+            store.suspendForVariant();
+
+            pageId.set(PAGE_B);
+            spectator.flushEffects();
+            store.resumeFromVariant();
+
+            expect(store.isOpen()).toBe(false);
+            expect(store.experimentId()).toBeNull();
+        });
+    });
+
+    describe('the language on the canvas changes', () => {
+        it('changes nothing about what the panel is showing', () => {
+            initStore();
+            store.open();
+            store.showConfigure(EXPERIMENT_A);
+
+            languageId.set(LANGUAGE_ES);
+            spectator.flushEffects();
+
+            expect(store.isOpen()).toBe(true);
+            expect(store.view()).toBe('configure');
+            expect(store.experimentId()).toBe(EXPERIMENT_A);
+        });
+
+        it('updates the return context, read live rather than captured at open', () => {
+            initStore(PAGE_A, LANGUAGE_EN);
+            store.open();
+
+            languageId.set(LANGUAGE_ES);
+            spectator.flushEffects();
+
+            expect(store.languageId()).toBe(LANGUAGE_ES);
+        });
+    });
+});
--- a/core-web/libs/portlets/edit-ema/portlet/src/lib/dot-ema-shell/dot-ema-shell.component.ts
+++ b/core-web/libs/portlets/edit-ema/portlet/src/lib/dot-ema-shell/dot-ema-shell.component.ts
@@ -37,6 +37,7 @@ import {
     DotPageToolsSeoComponent,
     PageScannerToolType
 } from '@dotcms/portlets/dot-ema/ui';
+import { DotExperimentsPanelStore } from '@dotcms/portlets/dot-experiments/data-access';
 import { GlobalStore } from '@dotcms/store';
 import { DotCMSPage, UVE_MODE } from '@dotcms/types';
 import { DotInfoPageComponent, DotMessagePipe, DotNotLicenseComponent, InfoPage } from '@dotcms/ui';
@@ -123,7 +124,16 @@ function hasOpenContentForEdit(component: unknown): component is RouteWithOpenCo
         MessageModule,
         DotMessagePipe
     ],
-    providers: [ConfirmationService]
+    /**
+     * `DotExperimentsPanelStore` is provided **here**, not by the panel it belongs to (#37478).
+     *
+     * The panel component is mounted and destroyed by this template's `@if` on the store's
+     * `isOpen`, so a store owned by the panel would die with it — and the variant round trip
+     * needs the state to survive a period during which the panel is closed. This shell is the
+     * smallest scope that outlives that trip: leaving for a variant only changes the editor's
+     * query params, so the shell is never re-created.
+     */
+    providers: [ConfirmationService, DotExperimentsPanelStore]
 })
 export class DotEmaShellComponent implements OnInit, OnDestroy {
     @ViewChild('dialog') dialog!: DotEmaDialogComponent;
@@ -144,6 +154,18 @@ export class DotEmaShellComponent implements OnInit, OnDestroy {
     readonly #siteService = inject(SiteService);
     readonly #location = inject(Location);
     readonly #globalStore = inject(GlobalStore);
+    protected readonly experimentsPanel = inject(DotExperimentsPanelStore);
+
+    /**
+     * The page in hand, as the Experiments panel's scope (#37478).
+     *
+     * A signal rather than a value: the panel re-scopes when the editor navigates to another
+     * page without closing it, and it must never be able to describe a page other than the one
+     * on the canvas (FR-034).
+     */
+    protected readonly $experimentsPanelPageId = computed<string | null>(
+        () => this.uveStore.pageAsset()?.page?.identifier ?? null
+    );
     readonly #dotMessageService = inject(DotMessageService);
     protected readonly $lockOptions = this.uveStore.$lockOptions;
     protected readonly $workflowLockIsLoading = this.uveStore.workflowLockIsLoading;
@@ -347,6 +369,15 @@ export class DotEmaShellComponent implements OnInit, OnDestroy {
 
     constructor() {
         this.$updateBreadcrumb(this.$breadcrumbPage);
+
+        // Signals, not values: the panel reads the page and the language at the moment it uses
+        // them, so neither goes stale against the canvas. The language is return context only —
+        // it never narrows the panel, because an experiment belongs to a page and not to one of
+        // its language versions (#37478, D10).
+        this.experimentsPanel.setContext({
+            pageId: this.$experimentsPanelPageId,
+            languageId: this.uveStore.pageLanguageId
+        });
     }
 
     ngOnInit(): void {
--- a/core-web/libs/portlets/edit-ema/portlet/src/lib/dot-ema-shell/dot-ema-shell.component.ts
+++ b/core-web/libs/portlets/edit-ema/portlet/src/lib/dot-ema-shell/dot-ema-shell.component.ts
@@ -11,7 +11,9 @@ import {
     OnDestroy,
     OnInit,
     signal,
-    ViewChild
+    untracked,
+    ViewChild,
+    ViewContainerRef
 } from '@angular/core';
 import { takeUntilDestroyed, toSignal } from '@angular/core/rxjs-interop';
 import { ActivatedRoute, Params, Router, RouterModule } from '@angular/router';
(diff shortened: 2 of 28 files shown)


specs/37478-uve-experiments-panel/spec.md
```diff
@@ -0,0 +1,890 @@
...
+not delivered the entry point.
+
+**Independent Test**: on a page with no experiments, open the panel, confirm it opens on creation
+rather than on a list, confirm the page cannot be changed, create the experiment, and confirm it
+belongs to that page.
+
+**Acceptance Scenarios**:
+
+1. **Given** a page with no experiments, **When** the editor opens Experiments, **Then** the panel
+   opens on creating an experiment for that page — not on an empty list.
```
fmontes: no, still open in list, user hit the create button when they want.

specs/37478-uve-experiments-panel/spec.md
```diff
@@ -0,0 +1,890 @@
...
+becomes an explicit, deliberate way out to the whole portlet, taken only when the editor asks for
+it, and taken without costing them the page (D6).
+
+### The flow, end to end
+
+The flag chooses between the behavior that exists today and this one. With it on:
+
+1. The editor activates Experiments on a page. **No navigation happens** — a panel opens beside
+   the canvas, and the page stays rendered.
+2. **If the page has no experiments, the panel opens straight on creation.** There is no empty
```
fmontes: No, open on listing, the user have to click "New Experiment"

specs/37478-uve-experiments-panel/spec.md
```diff
@@ -0,0 +1,890 @@
...
+   list to look at and then act on: the only useful next step is to create one, so that is what
+   the panel offers (D11).
+3. **If the page has experiments, the panel lists them** — that page's, and no other page's — and
+   the editor picks the one to configure.
+4. **In creation and in configuration, the page is fixed.** The experiment belongs to the page on
+   the canvas. There is no page picker and no way to change the page (D12).
```
fmontes: yes, we need to add a form field hint explaining why can't be changed.

specs/37478-uve-experiments-panel/spec.md
```diff
@@ -0,0 +1,890 @@
...
+
+## User Scenarios & Testing *(mandatory)*
+
+### User Story 1 - See this page's experiments without losing the page (Priority: P1)
+
+An editor is working on a page in the visual editor. They want to know whether this page has
+experiments, and what state they are in. They open Experiments and the answer appears beside the
+page they are editing. The page is still there — still rendered, still scrolled where they left
+it — and closing the panel puts them back exactly where they were, with nothing to navigate back
+through.
```
fmontes, line 181: Dont we also have a badge on the toolbar when the page have an experiment running?

specs/37478-uve-experiments-panel/spec.md
```diff
@@ -0,0 +1,890 @@
...
+
+**Why this priority**: this is the whole point of the work. Everything else in this spec either
+supports it, protects it, or extends it. It is also the smallest thing worth shipping.
+
+**Independent Test**: with the flag on, open a page with experiments in the editor, open
+Experiments, confirm the page is still rendered and has not reloaded, confirm the listed
+experiments are that page's, close the panel, confirm the page is untouched.
```
fmontes, line 188: The flag is for development... after QA approved of the whole the flag is out of the picture, no keeping both behaviour.

fmontes (dismissed):

specs/37478-uve-experiments-panel/spec.md
oidacra: Agreed, reversed. D11 now says the panel always opens on the list, and the empty state carries the New Experiment action; creating is the editor gesture, not the panel taking them there. Rewrote FR-006, added FR-006a (the empty state must carry the create action, so it is not a dead end), FR-007, User Story 2, and SC-003 (two gestures now, not one).

This also removes one of the two places the spec diverged from your design canvas, which had the empty state all along.

One knock-on worth naming: because the empty state is now a resting state of the panel, loading, error and analytics-misconfigured must each be clearly distinguishable from it, or they read as "this page has no experiments" and invite the editor to create a duplicate. That is FR-026 to FR-029.

specs/37478-uve-experiments-panel/spec.md
oidacra: Same change as the thread on line 51 - reversed. The panel always opens on the list; the empty state carries New Experiment and the editor clicks it when they choose to.

specs/37478-uve-experiments-panel/spec.md
oidacra: Added as FR-014a: the page field must carry a hint explaining why it cannot be changed, not just refuse the interaction. The spec now states the reason a reviewer or QA can check against - refusing without explaining reads as a defect, and the editor should be able to tell "fixed by design" from "broken".

The flow summary in the Scope Note carries the same point so it is not only buried in the requirement.

specs/37478-uve-experiments-panel/spec.md
oidacra: Yes, and this question found a real error in the spec - thank you.

DotEmaRunningExperimentComponent is already in the toolbar, and it is fed by the page load itself: withPageApi requests getById(pageParams.experimentId ?? pageAsset.runningExperimentId) alongside the page and the languages. Two consequences:

1. FR-036 was wrong. It said a UVE load that never opens the panel must fire "zero" experiments requests, which forbids behaviour that already ships. Corrected to "no request beyond the one the editor already makes today for the badge", in FR-036 and SC-004, and the panel must reuse what the editor already holds rather than duplicate it.

2. The badge is a second entry point into experiments, and it deep-links to the legacy per-page results route - so with the flag on it would eject the editor exactly the way the nav item used to, and #37008 deletes that route underneath it. Leaving it would have shipped one door that keeps the page and one that throws it away, and the badge is the door the editor did not choose to open.

So it is now in scope (D14): with the flag on the badge opens the panel on that experiment results. That pulled Results into the panel with it (D2 reversed, phase 3 added) - a badge opening the panel on a screen the panel cannot render would be worse than the eject it replaces. Both are recorded as reversals of earlier decisions, with the reasoning, and the canvas already modelled results in the drawer.

New: FR-025a to FR-025g, User Story 6, SC-003a (after phase 3, zero routes into experiments eject). FR-025f keeps the charting dependency out of the panel cost until results are actually opened.

specs/37478-uve-experiments-panel/spec.md
oidacra: Reframed. D15 now states the flag is a development and QA gate, not a long-lived product switch, and that it is removed together with the legacy screens by #37008 - the issue that already removes /old.

User Story 3 is no longer "nothing changes for anyone who has not opted in" with an operator in it; it is about building and verifying this without disturbing what ships. Section K is retitled to "Flag off - the development gate". The technical requirement is unchanged and deliberately so: flag off must still mean exactly today behaviour and the read must still fail closed, because a dev gate that fails open exposes unfinished work by accident.

Added FR-048 as the guard against the outcome you are pointing at: no capability may be offered on one side of the flag and withheld on the other beyond the transition itself, so nothing accrues to the flag that would make it un-removable.

One thing I cannot fix from this spec, flagged in D15 rather than silently diverged from: this contradicts #37005 approved spec, whose SC-002 promises "an operator under a minute to move between entry points without a deployment or restart" - an operator-facing switch on this same flag. Both specs cannot be right. This one records the reviewed intent; #37005 SC-002 needs revisiting, and that is a change to that spec. Raising it with the coordinator.

erickgonzalez (commented):
Just a nitpick: 

**D15 contradicts an already-approved spec**

D15 says so itself: [#37005](https://github.com/dotCMS/core/issues/37005) SC-002 promises an operator switch on this same flag, and FR-047/FR-048 rest on the opposite reframing. Two approved specs cannot both be right, and FR-048 is a hard requirement built on that reading.
Resolve before planning: amend #37005's SC-002 and get it re-approved, or drop FR-048 from this spec.

oidacra:
@erickgonzalez on D15 and #37005 — you were right that the spec said this, and the spec was wrong to say it. I went back to #37005's text and there is no contradiction to resolve, so neither of the two resolutions is needed. D15 is corrected instead.

**#37005 already hands the flag's removal to #37008.** Two places in its own text:

- Scope Note: "This does **not** retire the old screens or remove the switch — a later migration issue (#37008) does that."
- Relationship to #37008: "#37008 ... retires **`FEATURE_FLAG_EXPERIMENTS_PORTLET`**, the switch this work introduces, together with its shipped default and its wiring."

Its US3 is also P2 and scoped to "someone who deliberately opts in". So on the flag's lifetime the two specs agree, and D15 was asserting a disagreement that is not in the text.

**SC-002 and FR-048 constrain different things.** SC-002 measures how fast a flip takes effect ("under one minute, without a deployment or restart") — it says nothing about how long the flag lives or what may be built behind it. FR-048 constrains what may be built on either side. This spec keeps #37005's own uncached, fail-closed reader unchanged (FR-005): `readExperimentsPortletSwitch` reads the raw key rather than the memoizing `getFeatureFlag`, precisely so that a flip lands without a restart. SC-002 therefore goes on being satisfied here, unmodified.

FR-048 also conflicts with nothing in #37005 — its FR-014, "the switch MUST NOT change the behavior of any capability other than the ... entry point", pushes the same way.

**What changed in the spec:** D15's last paragraph no longer claims a contradiction, and instead records this check with the citations above. The note under US3 that said the reframing "needs reconciling with #37005's own spec" is corrected too. FR-047 and FR-048 stand as they were, and #37005 needs no amendment.

On "resolve before planning" — the plan ran after this verification, and nothing in it turned on the outcome: the reader is #37005's, unchanged, and flag-off remains today's behavior exactly. The only planning artifact that moved was the note recording the check.

core-web/libs/portlets/edit-ema/portlet/src/lib/dot-ema-shell/dot-ema-shell.component.ts
```diff
@@ -144,6 +161,60 @@ export class DotEmaShellComponent implements OnInit, OnDestroy {
...
+     *
+     * The split is the point: three screens, four stores and chart.js stay out of what the editor
+     * loads until the panel is first opened (FR-037, SC-006).
+     *
+     * Once loaded the class is kept, and `@if` in the template mounts and destroys the panel from
+     * there — so closing still leaves nothing of it running (FR-039) without re-fetching the chunk
+     * on the next open. Nothing here has to survive the `await`: if the editor dismissed the panel
+     * while it was in flight, the template simply never renders it.
+     */
+    readonly $experimentsPanelLoader = effect(() => {
```
adrianjm-dotCMS, line 242: This dynamic `import()` has no error handling. If the chunk fails to load (stale deploy referencing a hash the CDN no longer has, a transient network failure, or a script error inside the experiments bundle), the promise rejects silently: `$experimentsPanelComponent` stays `null` forever, the `@if` in the template never renders anything, and the editor is left with a nav item that silently does nothing — no toast, no log, no monitoring signal, nothing to debug from.

Suggest wrapping the `await import(...)` in a try/catch that logs the error, surfaces a toast to the user, and calls `this.experimentsPanel.close()` so the panel isn't left stuck `isOpen: true` with nothing to show for it.

dotCMS/src/main/webapp/WEB-INF/messages/Language.properties
```diff
@@ -6213,6 +6213,21 @@ experiments.results.empty.not-started.description=This Experiment has not starte
...
 experiments.results.empty.unavailable.description=The report could not be loaded, so there are no numbers to show here.
 
+# The Experiments panel inside the Universal Visual Editor (#37478)
+experiments.panel.title=Experiments
+experiments.panel.action.close=Close the Experiments panel
+experiments.panel.action.expand=Expand the Experiments panel
+experiments.panel.action.collapse=Collapse the Experiments panel
+experiments.panel.shortcut.close=Close the Experiments panel
+experiments.panel.empty.description=Nothing is being tested on this Page yet.
+experiments.panel.action.view-all=View all Experiments
```
adrianjm-dotCMS: These two keys (`experiments.panel.action.view-all` / `.view-all.tooltip`) are added here but never referenced by any component in this PR — I grepped the full diff and found no `view-all`/`viewAll` usage outside this file and the spec docs.

Per the linked issue's acceptance criteria, the panel needs an explicit, honest way out to the full `/experiments` portlet (this is FR-021c's 'clearable' requirement, and the spec's own D6 decision: a link opening the portlet unfiltered in a new tab so the page stays open here). Right now the copy is prepared but the actual button/link isn't wired up in `dot-experiments-panel.component.html`/`.ts`.

Can you add the missing action, or drop these keys if it's intentionally deferred to a follow-up?

core-web/libs/portlets/dot-experiments/data-access/src/lib/dot-experiments-panel.store.spec.ts
```diff
@@ -0,0 +1,440 @@
...
+            spectator.flushEffects();
+            pageId.set(PAGE_A);
+            spectator.flushEffects();
+
+            expect(store.view()).toBe('configure');
+            expect(store.experimentId()).toBe('exp-9');
+        });
+    });
+
+    describe('returnFromVariant()', () => {
```
adrianjm-dotCMS: None of the 4 tests in this `describe('returnFromVariant()', ...)` block call `initStore()` — they read directly from the outer-scope `store`, which is left in whatever state the previous `describe('re-scoping on the editor page')` block's last test happened to leave it in (`view: 'configure'`, `experimentId: 'exp-9'`, `suspendedForVariant: false`).

For example, `'should return from a panel that was never suspended'` asserts `expect(store.suspendedForVariant()).toBe(false)` as if it were a verified precondition — but that's only true because of leftover state from a sibling block, not anything this test itself sets up. It also doesn't match the block's own doc comment above, which claims these tests start 'from the worst starting points on purpose... a panel that was never suspended' — no such starting point is actually constructed.

This makes the whole block order-dependent: reordering tests, filtering, `.only`, or inserting a new `describe` above this one could silently break these tests, or make them silently pass for the wrong reason. Can you add `initStore()` (with explicit setup per test, e.g. plain `initStore()` alone for the 'never suspended' case) at the top of each `it` here, the way every other block in this file already does?

adrianjm-dotCMS (dismissed):

adrianjm-dotCMS (dismissed):

adrianjm-dotCMS (approved):

PR #37489, opened 2026-09-10
fix(content-drive): fail request instead of silently returning partial results on ES sub-query failure (#37488)

### LINKED ISSUE #37488: [BUG] Content Drive: failed/timed-out ES sub-queries silently return incomplete pages as HTTP 200
# [BUG] Content Drive: failed/timed-out ES sub-queries silently return incomplete pages as HTTP 200

Related to #37184. Follow-up from #37395 (single-pass field filter resolution) — flagged in review but scoped out since it's pre-existing behavior, not introduced by that PR.

## Problem

In `dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java`:

- `processSingleESQuery` catches every exception thrown by the ES search call, logs it, and returns an **empty set** instead of propagating the failure.
- `processMultipleESQueries` (used when the inode set is split into several sub-queries) wraps each sub-query future with `.orTimeout(60, SECONDS).exceptionally(...)`, which also returns an empty set per failed/timed-out sub-query. The outer `allFutures.get(120, SECONDS)` call additionally catches `InterruptedException`/`ExecutionException`/`TimeoutException`, logs them, and falls through without failing the request.

Net effect: a slow or failed ES sub-query silently drops its share of matches from the response. The caller still gets back **HTTP 200** with a short/incomplete page — there is nothing in the response indicating that results are incomplete.

## Why it matters now

This risk is not new, but it was narrow: historically there were only 1-2 ES sub-queries per Content Drive request, so the odds that any one of them failed or timed out in a given request were low. A related change widens the field-filter fan-out from 2 sub-queries to potentially dozens per request. With more sub-queries per invocation, the chance that at least one drops out per request rises accordingly. The symptom reported by users would be something like "my filter sometimes misses content," with no error or indication anywhere in the response.

## Proposed fix direction

Change the swallow-and-continue behavior in `processSingleESQuery` / `processMultipleESQueries` so a sub-query failure propagates as a request-level failure (a visible error response) instead of silently substituting an empty result and returning a short page.

Care needed:
- Verify blast radius — check whether `hydrateContentletsInParallel` / `findContentletsInParallel` or any other caller shares code paths with these two methods and might currently depend on the lenient (swallow-and-continue) behavior.
- Callers that don't hit this failure path should see no behavior change.

## Acceptance criteria

- [ ] A sub-query failure or timeout causes the overall request to fail with a clear error, instead of silently returning a short HTTP 200 page.
- [ ] Existing callers that never hit an ES sub-query failure see no behavior change.
- [ ] Test coverage: a test that forces one sub-query to throw or time out (via a mocked/injected failure) and asserts the overall request surfaces an error rather than a partial success.

ihoffmann-dot:
## Summary
- `processSingleESQuery` / `processMultipleESQueries` in `BrowserAPIImpl` used to swallow ES sub-query exceptions/timeouts and quietly return an empty result, so a failed or slow sub-query silently dropped its share of matches and the caller got HTTP 200 with a short/incomplete page.
- Now a sub-query failure propagates as a `DotDataException`/`DotRuntimeException`, so the request surfaces a visible error instead of a silently incomplete page.

Fixes #37488. Follow-up from #37395 (single-pass field filter resolution) review — flagged as pre-existing behavior made more likely to matter once that PR widens the ES fan-out per request.

## Caller audit
Traced every caller of `processSingleESQuery`/`processMultipleESQueries`: the only path is `processESDirectly` → `getChunkFiltered` → `getContentByChunks` → `doHybridSingleChunkedQueryES` → `doElasticSearchTextFiltering` → `getContentUnderParentFromDB`, which already declares/wraps `DotDataException` and rethrows as `DotRuntimeException`. No caller depended on the old empty-result-on-failure fallback.

## Test plan
- [x] `./mvnw compile -pl :dotcms-core` — clean
- [ ] Integration test forcing a sub-query failure/timeout and asserting the request surfaces an error (flagged as needing DB/ES-level mocking beyond this unit test file's current scaffolding — see PR discussion)

### DIFF
--- a/dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
+++ b/dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
@@ -770,8 +770,12 @@ public class BrowserAPIImpl implements BrowserAPI {
      * @param browserQuery The {@link BrowserQuery} containing search criteria (filter, fileName)
      * @param inodes       The set of inodes to filter through Elasticsearch text search
      * @return A filtered set of inodes that match the text search criteria
+     * @throws DotDataException if a sub-query fails or times out. A failure here is deliberately
+     *                          NOT swallowed into an empty result: silently dropping a failed
+     *                          sub-query's share of matches would return an incomplete page as a
+     *                          successful (HTTP 200) response, which is worse than a visible error.
      */
-    Set<String> processESDirectly(BrowserQuery browserQuery, Set<String> inodes) {
+    Set<String> processESDirectly(BrowserQuery browserQuery, Set<String> inodes) throws DotDataException {
         if (inodes == null || inodes.isEmpty()) {
             return new LinkedHashSet<>();
         }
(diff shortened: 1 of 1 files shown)


ihoffmann-dot:
Blocked by #37695. That issue is a concrete case of this bug: oversized inode batches get rejected by the index server (`max_query_string_length` 32000), and the swallow here returns 200 with no contentlets. Merging this first would turn that into a hard error on large sites. Also note that #37554 documented the swallow as deliberate (FR-029, `ContentDriveFieldFilterTest#testMalformedDateBoundIsSafe`), so fail-fast needs to be settled with that in mind. Will revisit once #37695 lands.

Issue #37491, opened 2026-09-10
Harden per-site authorization and site resolution on the /api/v1/ai/* endpoints


fmontes:
### Description

Review of #37431 surfaced gaps that are shared by the already-shipped `/api/v1/ai/*` endpoints. None of these is remote-anonymous-exploitable — `InitBuilder.anonAccess` defaults to `AnonymousAccess.NONE` and `WebResource.checkAnonymousPermissions` (`WebResource.java:369-385`) 401s anonymous callers — but together they mean any authenticated user can spend an arbitrary site's LLM budget, on an arbitrary model, with no per-caller site check.

**1. No per-caller site permission check when `siteId` is omitted.** Every endpoint falls back to `getCurrentHostNoThrow(request)` (or `getHost(request)` — `HostWebAPIImpl.java:49-57`), which calls `getCurrentHost(request, null)` and resolves the site **as system user** (`HostWebAPIImpl.java:78-79`). The caller's permissions are never consulted. Only the explicit-`siteId` path checks anything, via `HostAPI.find(siteId, user, false)` in `AiHostResolver.java:75-80`.

**2. Silent cross-site config fallback.** `HostAPIImpl.resolveHostName` falls back to `findDefaultHost` when the server name matches no site or alias (`HostAPIImpl.java:140-144`), and `ConfigService.config` then falls back to SYSTEM_HOST when the resolved site has no dotAI secrets (`ConfigService.java:49-53`). A client pointed at `localhost`, an unaliased hostname, or sitting behind a proxy that rewrites `Host` silently spends the **default** (or system) site's credentials and gets a 200.

**3. Inconsistent model passthrough.** `CompletionsResource.resolveForm` forces non-admins onto the site's configured model (`CompletionsResource.java:366-369`). `TextResource.generateRequest` passes `form.model` straight through with no such check (`TextResource.java:105,116`). Same subsystem, opposite policy — the second lets any authenticated user name an arbitrary model.

**4. No cost accounting.** No AI path carries a `@RequestCost` price, so the instance-wide `RequestCostFilter` backstop (`web.xml:175-178`, off by default) does not even register an LLM call — which parks a request thread and spends real money.

**5. Unauthenticated static probes.** `ImageResource.java:46`, `EmbeddingsResource.java:66` and `SearchResource.java:54` expose `GET /test` returning a constant `{"type": …}` map with no auth. Harmless, but undocumented dead weight.

[...]

Issue #37433, opened 2026-09-07
Build and publish @dotcms/ai-sdk-provider (Vercel AI SDK provider for dotCMS)


fmontes:
### Description

Build and publish `@dotcms/ai-sdk-provider` — a first-party Vercel AI SDK provider so any AI SDK application can use a dotCMS instance as its model provider:

```ts
import { createDotCMS } from '@dotcms/ai-sdk-provider';

const dotcms = createDotCMS({ baseURL, apiKey });   // apiKey = a dotCMS API token
model: dotcms('claude-sonnet-4-6')
```

Depends on **#37431**, which adds the `/api/inference/v1` endpoints this provider talks to.

**Why a package rather than `@ai-sdk/openai-compatible` in each app.** `@ai-sdk/openai-compatible` would work, but it still ships `specificationVersion: 'v2'`, so it runs in compatibility mode under `ai@6` ([vercel/ai#12615](https://github.com/vercel/ai/issues/12615)). A native `v3` implementation avoids compat mode, and — more importantly — gives dotCMS one place to own the details every consumer would otherwise re-derive: token auth, per-site base URL resolution, mapping dotAI's error envelope onto AI SDK error types, and surfacing model fallback chains. One package, semver'd, instead of copy-pasted config in every app.

**Where it lives.** New Nx library at `core-web/libs/sdk/ai-sdk-provider`, mirroring `core-web/libs/sdk/ai` (`build`, `publish`, `nx-release-publish`, `lint`, `test` targets). Note this is the **model** layer and is orthogonal to `@dotcms/ai`, which is the tool/execution layer — they compose, and neither replaces the other.

**Spec target.** `LanguageModelV3` against `@ai-sdk/provider@3.0.11` — verified from the installed types: required members are `specificationVersion: 'v3'`, `provider`, `modelId`, `doGenerate(LanguageModelV3CallOptions)` and `doStream(LanguageModelV3CallOptions)`. The package also exports `ProviderV3` / `EmbeddingModelV3` / `ImageModelV3` for the non-chat surfaces. (The AI SDK docs page currently describes a "V4" spec; the shipped types are `v3` — follow the types.)

[...]

Issue #37431, opened 2026-09-07
Add OpenAI-compatible inference endpoints at /api/inference/v1


fmontes:
### Problem

dotAI's endpoints can't be driven by any standard AI client. This surfaced trying to point an agent (Vercel AI SDK) at dotCMS as its model provider:

- **No tool calling anywhere in the stack.** `LangChain4jAIClient` builds `ChatRequest.builder().messages(messages).build()` with no tool specifications, and `AiKeys` has no `tools` / `tool_calls` key. Any agent loop stops after one turn.
- **No `messages[]`.** `CompletionsForm` takes a single `prompt` string capped at 4096 chars, and `/rawPrompt` wraps it in one user message — no caller-supplied system prompt, no multi-turn, no assistant/tool turns.
- **Streaming isn't parseable by standard clients.** `toSseChunk` emits `{"choices":[{"delta":{"content":…},"index":0}]}` with no `id`, `object`, `created`, `model`, `finish_reason`, and no usage chunk.
- **`/api/v1/ai/completions` is a RAG endpoint, not a chat endpoint.** It runs an embeddings search, stuffs the results into the prompt, and returns a `{dotCMSResults, openAiResponse}` envelope.

The consequence: anyone wanting to use dotCMS as their model gateway has to write a bespoke client. Nobody does — they bypass dotCMS and put provider keys directly in their own app, which throws away the per-site credential governance dotAI already provides.

### Goal

Expose the multi-provider gateway dotAI **already is** behind the wire format the ecosystem already speaks, so that `baseURL` plus a dotCMS API token is the entire integration.

We already have the expensive half: seven providers via LangChain4j (OpenAI, Azure, Bedrock, Vertex, Gemini, Anthropic, OpenRouter), per-site credentials in App secrets, model fallback chains, and connection testing. What's missing is a standard door on the front of it.

**Why the OpenAI chat-completions format is the right standard:** it is the de-facto wire protocol — OpenRouter, LiteLLM, Vercel AI Gateway, Ollama, vLLM, LM Studio, Groq and Together all serve it. Adopting it means dotCMS works out of the box with the Vercel AI SDK, LangChain, LlamaIndex, Cursor, n8n and the official OpenAI SDKs, with no dotCMS-specific client to write or maintain.

### Endpoints

New family at `/api/inference/v1`. `baseURL = https://<host>/api/inference/v1`, dotCMS API token as the `apiKey`.

| Endpoint | Description |
|---|---|
[...]

PR #37493, opened 2026-09-10
docs(ai): spec for OpenAI-compatible inference endpoints at /api/inference/v1

fmontes:
**Spec-Kit PR 1 of 2.** This PR carries `spec.md` **alone** — no implementation. Per the repo flow it needs another dev's **approval**, not a merge: PR 2 branches off this branch and carries the code.

Refs #37431

### Proposed Changes

* Adds `specs/37431-openai-compatible-inference/spec.md` — the reviewed contract for exposing dotAI behind the OpenAI chat-completions wire format at `/api/inference/v1`, so that a base URL plus a dotCMS API token is the entire integration.
* 5 prioritized user stories (P1 chat + tool calling · P2 streaming · P2 per-site governance · P3 model discovery · P3 embeddings + images), 35 functional requirements, 22 acceptance scenarios, 10 success criteria, plus the dotCMS-required Legacy Considerations section.

### What reviewers should actually weigh in on

The issue is unusually thorough and its code claims all check out (verified: `toSseChunk` at `LangChain4jAIClient.java:282` emits only `delta.content` + `index`; both `ChatRequest.builder()` call sites pass messages and nothing else; `AiKeys` has no tool keys; `CompletionsForm.java:32` caps the prompt at 4096; `TextResource:110` builds the system message from the site's `rolePrompt` so a caller cannot supply one; `EmbeddingsResource.embed` is corpus indexing and backend-user only; langchain4j 1.15.1 is pinned and every tool-calling API needed is present in the jar). So the spec mostly formalizes the issue. **Two places where it deliberately diverges** are the substance of this review:

**1. Both silent site fallbacks are KEPT, and attribution replaces strictness (`FR-020`, `FR-021`).** The issue asked this family to remove two fallbacks. Neither is removed:

| Fallback | Verdict | Why |
|---|---|---|
| Site without AI secrets → SYSTEM_HOST config (`ConfigService.java:49-53`) | **Kept** | Ordinary dotCMS Apps inheritance — `ConfigService.config`'s own Javadoc states the intent. Removing it breaks every installation that configures dotAI once at system level. |
| Unmatched host name → default site (`HostAPIImpl.resolveHostName` → `findDefaultHost`) | **Kept** | This family is server-side only by design, and server-side callers routinely present a host name that is not a site alias: internal DNS, container service names, `localhost`. Strict resolution fired hardest on the primary supported deployment shape, and contradicted `FR-017` and `SC-001`, which promise a base URL alone is sufficient. |

The concern behind the original request is real, but it is **attribution, not resolution**: the objection in the issue is that a caller would "*quietly* spend the default site's credentials and receive a 200." So `FR-020` now requires the fallback to be logged with the unmatched host name and **every response to name the site whose configuration served it**, and `SC-005` is the measurable form. That is strictly more coverage than an error on the unmatched case, because it also surfaces the alias-collision case — a host matching the *wrong* site — which strict resolution never caught. A per-site spend quota is the full answer and stays out of scope.

⚠️ **Two consequences a reviewer must accept explicitly**, both recorded in the spec:
- This **reverses #37431's "no silent site fallback" acceptance criterion.**
- It **conflicts with the second acceptance criterion of #37491**, which commits to removing both fallbacks on `/api/v1/ai/*`. That criterion must be amended to match — otherwise the one shared resolver of `FR-025` needs a per-family strictness flag, which is exactly the drift `FR-025` exists to prevent.

**2. A reserved model alias is added (`FR-024`).** The issue's AC requires an unconfigured model name to return a `NoSuchModelError`-shaped 404, while its Demo Expectation requires swapping a site's provider vendor and re-running *the same agent unchanged*. Both cannot hold: an agent hardcoding `model: 'gpt-4o'` breaks the moment the site swaps to Anthropic. The spec keeps strict validation for concrete names and adds an alias meaning "this site's primary model and its fallback chain". Exact spelling left to `/speckit-plan`.

### Checklist

- [ ] Tests — n/a for PR 1 (spec only, no code). TDD gates are enforced in PR 2 per constitution Principle V: tests written → dev-approved → confirmed failing (Red) before any implementation.
- [ ] Translations — n/a, no UI in scope.
- [x] Security Implications Contemplated — much of the spec's weight. Anonymous callers rejected; any authenticated user accepted (matching `/api/v1/ai/*`, since serving AI on behalf of a visitor is supported); site READ enforced on explicit overrides; **model allowlisting is the load-bearing control**, since site READ is necessarily granted to frontend roles on any publicly delivered site; the legacy `host_id`/`host` de-facto site override is ignored here; no CORS, because the credential is a long-lived token with its owner's full authority. **Accepted residual risk, now load-bearing:** with standard resolution kept and no per-caller site check on the default path, an authenticated user can cause a site's — or the default site's — AI budget to be spent. Attribution makes that visible; only the out-of-scope **per-site spend quota** prevents it. The instance-wide `LeakyTokenBucket` cannot express one (off by default, prices local resource time, not provider spend).

### Additional Info

* Spec-folder commit policy: `spec.md` is the durable reviewed contract and ships; `checklists/requirements.md` is a process artifact and is gitignored.
* Rollback: no code in this PR. The spec flags that PR 2 adds a public API contract and should be labeled accordingly.
* Follow-ups kept out of scope on purpose: #37491 (harden the existing `/api/v1/ai/*` endpoints and adopt the shared resolver there — and amend its AC 2 per above) and the per-site AI spend quota, which this spec's resolution decision makes the more important of the two.
* Client-side counterpart, blocked by this: #37433 (`@dotcms/ai-sdk-provider`).
* Next step after approval: `/speckit-plan`, which auto-fires the mandatory `before_plan` ADR-consultation hook against `dotCMS/platform-adrs`.


specs/37431-openai-compatible-inference/spec.md
```diff
@@ -0,0 +1,215 @@
...
+- **FR-019**: The system MUST enforce read access on the target site whenever an explicit override is supplied, and MUST refuse the request as forbidden when the caller lacks it.
+- **FR-020**: Site resolution MUST follow standard dotCMS host resolution, including the fallback to the default site when the host name matches no site or alias — this family introduces no resolution semantics of its own, because an endpoint family that resolves differently from the rest of the product is its own source of surprise, and because the server-side callers this family exists for routinely present a host name that is not a site alias (internal DNS, container service names, `localhost`). The risk the fallback carries is that nobody can tell which site paid, so the system MUST answer that directly: a fallback resolution MUST be logged with the unmatched host name, and every response MUST state which site's configuration served it.
```
nicobytes, line 147: **How does a response "state which site's configuration served it"?**

FR-020 and SC-005 require every response to name the serving site, and SC-005 explicitly rules out "reading server logs". SC-008 requires payloads to deserialize into a standard client library's own result types **with no adapter**. A non-standard top-level body field satisfies FR-020 but risks SC-008 with strict deserializers; a response header satisfies SC-008 but is arguably not part of "the response". The spec does not say which.

Three sub-questions need an answer too:

- Does it apply to **streamed** responses — every chunk, or only the first and/or last?
- Does it apply to **error** responses?
- Is the value the site identifier, the hostname, or both?

A response header looks like the only mechanism that keeps SC-008 unconditionally true, behaves identically for streaming and for errors, and is readable by the proxy or log pipeline that would actually do the spend reconciliation SC-005 describes.

specs/37431-openai-compatible-inference/spec.md
```diff
@@ -0,0 +1,215 @@
...
+- **FR-007**: The system MUST emit streaming responses as standard chat-completion chunk events, each carrying the response identifier, object type, creation time, and model, with a finish reason on the final content event and the conventional terminal marker closing the stream.
+- **FR-008**: The system MUST stream tool calls incrementally, emitting each tool call's index, identifier, name, and argument fragments so a standard client can reassemble the arguments as they arrive.
+- **FR-009**: The system MUST report token usage for streamed exchanges.
+- **FR-010**: The system MUST expose a model-listing operation returning the models the resolved site has configured, including every entry of its fallback chains, in the standard model-list shape.
+- **FR-011**: The system MUST expose an embeddings operation returning vectors in the standard embeddings shape.
+- **FR-012**: The system MUST expose an image-generation operation returning results in the standard image-generation shape.
+- **FR-013**: The system MUST ignore request fields it does not support rather than rejecting the request, so a standard client's default payload succeeds.
```
nicobytes, line 137: **Which generation parameters are honored, and which are silently ignored?**

FR-013 says unsupported request fields MUST be ignored rather than rejected. The spec enumerates `messages`, `tools`, `tool_choice`, `response_format` and streaming as supported, and never says what happens to `temperature`, `max_tokens`, `top_p`, `stop`, `n`, `seed`, `presence_penalty` or `parallel_tool_calls`.

This matters beyond compatibility politeness. Silently ignoring `max_tokens` is a cost and correctness problem: the caller asked for a bounded response, got an unbounded one, and the site paid for it. Silently ignoring `n` returns one choice where the caller's code expects several, and the failure surfaces far from its cause.

FR-013 probably needs narrowing so that it covers *client-default* fields rather than fields that change output semantics, with the spec naming which parameters are passed through to the provider and which are rejected outright.

specs/37431-openai-compatible-inference/spec.md
```diff
@@ -0,0 +1,215 @@
...
+- **FR-021**: A resolved site with no AI configuration of its own MUST continue to inherit the system-level configuration, as every other dotAI caller does — this is standard dotCMS Apps inheritance, and many installations configure dotAI once at system level rather than per site. Only when neither the resolved site nor the system level has any AI configuration MUST a completion request be refused with a standard-shaped error.
+- **FR-022**: The model-listing operation MUST return an empty list when neither the resolved site nor the system level has any AI configuration, rather than any other site's models.
+- **FR-023**: The system MUST resolve a requested model against the models the resolved site has configured. A non-administrator naming a model the site has not configured MUST be refused with a standard-shaped "no such model" error; administrators MAY select any model the site has configured.
```
nicobytes, line 150: **What may an administrator ask for that a non-administrator may not?**

As written: "A non-administrator naming a model the site has not configured MUST be refused […]; administrators MAY select any model the site **has configured**." Both roles end up with the same set, so the second clause grants administrators nothing. Story 3 scenario 7 and SC-007 also constrain only non-administrators, which leaves the administrator case genuinely unspecified rather than deliberately equal.

Worth noting that FR-026 exists precisely because two dotAI endpoints already drifted to opposite model-passthrough policies. A role-conditional passthrough would re-introduce the branch FR-026 is meant to eliminate, and it makes SC-007 untestable as a blanket invariant — so pinning every caller to the site's configured models and dropping the administrator clause may be the cleaner reading.

specs/37431-openai-compatible-inference/spec.md
```diff
@@ -0,0 +1,215 @@
...
+- **FR-004**: The system MUST accept tool declarations and a tool-choice preference on a chat-completion request, and MUST return the model's tool calls in the standard tool-call shape with a finish reason indicating a tool call.
+- **FR-005**: The system MUST accept tool results as subsequent messages, correlated to the tool call that requested them, so a multi-step agent loop completes.
+- **FR-006**: The system MUST accept a response-format preference on a chat-completion request.
+- **FR-007**: The system MUST emit streaming responses as standard chat-completion chunk events, each carrying the response identifier, object type, creation time, and model, with a finish reason on the final content event and the conventional terminal marker closing the stream.
+- **FR-008**: The system MUST stream tool calls incrementally, emitting each tool call's index, identifier, name, and argument fragments so a standard client can reassemble the arguments as they arrive.
+- **FR-009**: The system MUST report token usage for streamed exchanges.
```
nicobytes, line 133: **Is streamed token usage always emitted, or gated behind the standard opt-in?**

In the format being adopted, usage on a stream is opt-in via `stream_options`, and it arrives as a chunk whose `choices` array is empty. Some stream readers assume every chunk carries a non-empty `choices`. Emitting usage unconditionally could therefore break the very clients SC-003 and SC-008 promise to support.

Since the stated acceptance bar in Assumptions is what standard clients actually require, gating on the standard opt-in flag seems right — in which case FR-009 should be reworded from "MUST report token usage for streamed exchanges" to "MUST report token usage for streamed exchanges when requested via the standard option".

nicobytes (commented):
Read through the spec — the reasoning is strong and the tradeoffs, especially around keeping standard site resolution rather than inventing per-family semantics, are argued well enough that I don't want to relitigate them.

What follows is a set of questions where the spec currently admits two readings that would produce materially different implementations and different acceptance tests. The first five are the ones I think block planning; the rest are smaller but still worth pinning down before this is signed off. Each one has a suggested answer, so none of them should need a long discussion — they mostly need a decision recorded in the spec.

specs/37431-openai-compatible-inference/spec.md
```diff
@@ -0,0 +1,215 @@
...
+
+#### Per-site isolation
+
+- **FR-027**: Per-site provider selection, credential resolution, and model fallback MUST reuse the existing site-configuration path, behaving exactly as the existing dotAI endpoints do. This family defines no exceptions to it.
+- **FR-028**: Two sites with different provider configurations under interleaved requests MUST each be served by their own provider instance, and rotating a site's credentials MUST invalidate that site's cached instances without affecting other sites.
+- **FR-029**: Provider clients MUST NOT be constructed anywhere outside the existing single client component, because the instance cache is keyed on the site's configuration rather than on the request, and constructing clients elsewhere to obtain tool support would defeat both the cross-site isolation and the credential-rotation invalidation that keying provides.
+
+#### Operational
+
+- **FR-030**: The system MUST NOT emit cross-origin headers on this family, and the endpoints MUST be documented as server-side only, because the credential is a long-lived token with the full authority of its owner and cross-origin support would invite placing it in browser JavaScript.
```
nicobytes, line 163: **Are conversation contents logged or retained anywhere, and for how long?**

The spec is silent on this, and the Operational section is where it would belong. Prompts and tool arguments routinely carry customer data, and the standing repo rule is that sensitive data is never logged. Meanwhile FR-020 does mandate logging on the fallback path, and SC-005's attribution story implies some durable record of who spent what — so "nothing is written down" is not the implicit default a reader can safely assume.

Right now an implementer has no instruction at all. Even a one-line requirement stating that only metadata is recorded (site, model, token counts, status, duration) and that request and response bodies never are, would settle it.

specs/37431-openai-compatible-inference/spec.md
```diff
@@ -0,0 +1,215 @@
...
+- **FR-031**: The system MUST translate upstream provider rate-limit and server errors into standard-shaped errors flagged as retryable, without leaking the provider's raw error envelope, so a standard client's back-off logic engages.
```
nicobytes, line 164: **How is "retryable" actually conveyed?**

The standard error shape has no retryable field, and a standard client's back-off keys off the HTTP status code, not the body. So "standard-shaped errors flagged as retryable" reads as though it requires a non-standard field — which would put this requirement in tension with SC-008, while SC-010 is really describing status-code behavior.

Suggest stating it plainly: retryability is conveyed by the HTTP status code (429 and 5xx), and the body stays byte-conformant to the standard error shape.

specs/37431-openai-compatible-inference/spec.md
```diff
@@ -0,0 +1,215 @@
...
+- **FR-006**: The system MUST accept a response-format preference on a chat-completion request.
+- **FR-007**: The system MUST emit streaming responses as standard chat-completion chunk events, each carrying the response identifier, object type, creation time, and model, with a finish reason on the final content event and the conventional terminal marker closing the stream.
+- **FR-008**: The system MUST stream tool calls incrementally, emitting each tool call's index, identifier, name, and argument fragments so a standard client can reassemble the arguments as they arrive.
+- **FR-009**: The system MUST report token usage for streamed exchanges.
+- **FR-010**: The system MUST expose a model-listing operation returning the models the resolved site has configured, including every entry of its fallback chains, in the standard model-list shape.
+- **FR-011**: The system MUST expose an embeddings operation returning vectors in the standard embeddings shape.
+- **FR-012**: The system MUST expose an image-generation operation returning results in the standard image-generation shape.
```
nicobytes, line 136: **How is a generated image delivered?**

The standard image-generation shape allows either a base64 payload or a hosted URL, and the two have very different consequences here. If it is a URL: where is the image stored, is that URL authenticated, and how long does it live? A long-lived unauthenticated URL to content generated from a possibly sensitive prompt is worth deciding on purpose rather than inheriting from whatever the existing dotAI image operation does.

specs/37431-openai-compatible-inference/spec.md
```diff
@@ -0,0 +1,215 @@
...
+- **FR-005**: The system MUST accept tool results as subsequent messages, correlated to the tool call that requested them, so a multi-step agent loop completes.
+- **FR-006**: The system MUST accept a response-format preference on a chat-completion request.
+- **FR-007**: The system MUST emit streaming responses as standard chat-completion chunk events, each carrying the response identifier, object type, creation time, and model, with a finish reason on the final content event and the conventional terminal marker closing the stream.
+- **FR-008**: The system MUST stream tool calls incrementally, emitting each tool call's index, identifier, name, and argument fragments so a standard client can reassemble the arguments as they arrive.
+- **FR-009**: The system MUST report token usage for streamed exchanges.
+- **FR-010**: The system MUST expose a model-listing operation returning the models the resolved site has configured, including every entry of its fallback chains, in the standard model-list shape.
+- **FR-011**: The system MUST expose an embeddings operation returning vectors in the standard embeddings shape.
```
nicobytes, line 135: **Which model list does the embeddings operation validate against?**

FR-023 pins a requested model to the models the resolved site has configured, but a site's embeddings model is typically configured separately from its chat models. Does FR-023 apply to this operation, and if so against which list? Also worth stating whether the reserved alias of FR-024 is meaningful here.

specs/37431-openai-compatible-inference/spec.md
```diff
@@ -0,0 +1,215 @@
...
+- Neither the resolved site nor the system level has any AI configuration and a completion is requested: an explicit standard-shaped error, never a borrowed configuration from an unrelated site.
+- The host name matches no site or alias — a container service name, internal DNS, or `localhost`: the default site serves the request, as it would anywhere else in dotCMS; the fallback is logged and the response names the serving site.
+- The caller is anonymous or presents no token: refused as unauthorized.
+- A conversation contains a tool-result turn with no preceding tool call, or malformed tool declarations: refused with a standard-shaped validation error naming the offending field.
+- The request is made from browser JavaScript: no cross-origin headers are emitted, so the browser blocks it — these endpoints are server-side only by design, because the token can do everything its owner can.
+- Many concurrent streams: each stream occupies a request thread for the life of the completion, which is the dominant capacity risk on this family and must be understood as such rather than treated as request-count load.
```
nicobytes, line 116: **No capacity or timeout requirement anywhere in the spec.**

This edge case correctly names concurrent streams as the dominant capacity risk on this family — but nothing in Requirements or Success Criteria follows up on it. There is no cap on concurrent streams (per site or per instance), no completion timeout, and no idle timeout. This is the one area of the spec with no measurable criterion at all, and it is the area the spec itself flags as the risk.

At minimum: a maximum concurrent-stream count and a hard completion timeout, both configurable, with stated defaults.

specs/37431-openai-compatible-inference/spec.md
```diff
@@ -0,0 +1,215 @@
...
+- A caller sends a request field this family does not support: unsupported fields are ignored rather than failing the request, so a client's defaults do not break the call.
+
+## Requirements *(mandatory)*
+
+### Functional Requirements
+
+#### Endpoint family and wire format
+
+- **FR-001**: The system MUST expose a new inference endpoint family whose path carries no vendor name and whose trailing version segment denotes the wire protocol version, distinct from dotCMS's own resource versioning, so a future protocol lands beside it without disturbing dotCMS versioning.
```
nicobytes, line 125: **The path is described but never stated.**

FR-001 gives the shape (no vendor name, trailing segment denoting wire protocol version) without naming the actual path. The issue title says `/api/inference/v1`. Since this is a public API contract and the spec itself classifies the change as rollback-sensitive, the literal path belongs in the requirement rather than being left to the plan phase.

specs/37431-openai-compatible-inference/spec.md
```diff
@@ -0,0 +1,215 @@
...
+- **FR-009**: The system MUST report token usage for streamed exchanges.
+- **FR-010**: The system MUST expose a model-listing operation returning the models the resolved site has configured, including every entry of its fallback chains, in the standard model-list shape.
+- **FR-011**: The system MUST expose an embeddings operation returning vectors in the standard embeddings shape.
+- **FR-012**: The system MUST expose an image-generation operation returning results in the standard image-generation shape.
+- **FR-013**: The system MUST ignore request fields it does not support rather than rejecting the request, so a standard client's default payload succeeds.
+- **FR-014**: The legacy non-chat completion operation is explicitly out of scope and MUST NOT be added; it is deprecated in the format being adopted.
+
+#### Authentication, authorization, and site resolution
+
+- **FR-015**: The system MUST authenticate callers by dotCMS API token presented as a bearer credential, and MUST refuse anonymous callers as unauthorized.
```
nicobytes, line 142: **Are any credentials other than a bearer API token accepted?**

FR-015 requires bearer API token authentication and refuses anonymous callers, but says nothing about session cookies or basic auth, both of which the surrounding dotCMS filters generally accept. Worth stating explicitly whether they are also accepted on this family or refused — the answer interacts with FR-030's deliberate absence of CORS headers.

specs/37431-openai-compatible-inference/spec.md
```diff
@@ -0,0 +1,215 @@
...
...ve, but two shared pieces are modified in place: the site-resolution helper grows into the single authorization-carrying entry point, and the provider client gains tool support and standards-conformant stream assembly. Site resolution itself reaches into the older site/host resolution surface, whose two silent fallbacks — an unmatched host name resolving to the default site, and a site without its own AI configuration inheriting the system-level one — are both **kept**, so this family resolves exactly as the rest of the product does. The source issue asked for both to be removed here; that is declined deliberately, because per-family resolution semantics are their own source of surprise and because the concern behind the request is attribution rather than resolution. FR-020 answers attribution by naming the serving site on every response; a per-site spend quota answers it fully and is out of scope. **This reverses #37431's "no silent site fallback" criterion, and conflicts with the second acceptance criterion of #37491, which must be amended to match — otherwise the one shared resolver of FR-025 would need a per-family strictness flag, which is the drift FR-025 exists to prevent.**
```
nicobytes, line 198: Nit: this paragraph refers twice to "the one shared resolver of FR-025". The shared resolver is **FR-026**; FR-025 is the ignore-legacy-parameters rule.

specs/37431-openai-compatible-inference/spec.md
```diff
@@ -0,0 +1,215 @@
...
+- **Known related decisions**: One existing dotAI endpoint already pins non-administrators to the site's configured model — this family carries that precedent forward, rather than repeating the unchecked model passthrough of the sibling endpoint that lacks it. Reading a site's AI configuration as the system user is the standard dotCMS Apps design and keeps the secret server-side, so the question here is authorization to *use* a site's credentials, not secret exposure. Related work: [#37491](https://github.com/dotCMS/core/issues/37491) hardens the existing dotAI endpoints and adopts this shared component there; [#37433](https://github.com/dotCMS/core/issues/37433) is the client-side counterpart and is blocked by this. The plan phase will formally consult `dotCMS/platform-adrs`.
+
+## Assumptions
+
+- The adopted wire format is the OpenAI chat-completions format. This is a scope decision, not an implementation detail: it is the de-facto protocol served by OpenRouter, LiteLLM, Vercel AI Gateway, Ollama, vLLM, LM Studio, Groq, and Together, so adopting it is what makes dotCMS work out of the box with the major AI frameworks and official SDKs.
```
nicobytes, line 204: **Was the Responses API considered as the adopted format, and if so why chat completions?**

Raising this because the Assumptions justify the choice by listing who implements the format, which is the right argument, but the spec never records that the alternative was weighed. OpenAI's current guidance is that Chat Completions is not deprecated and has no sunset date, while "Responses is recommended for all new projects" — so a reader coming to this spec cold will ask the question, and right now there is nothing here to answer them with.

Arguments I can see on each side, for whatever they are worth to whoever owns this decision:

**For Responses.** It is the better-designed protocol. FR-008 is arguably evidence of that — the requirement to emit each tool call's *index* and *argument fragments* so the client can *reassemble* them exists only because chat-completions streaming is delta-and-index based. Responses carries an explicit item id and needs no reassembly. It also preserves reasoning items across turns, which chat completions cannot, so multi-turn agent loops over reasoning models lose reasoning context here.

**For chat completions.** The selection criterion for this feature is arguably not design quality but what clients emit when pointed at a custom base URL, since dotCMS is the server side of this. There are also four concrete costs to a Responses-first implementation that the current scope does not carry:

- The pinned provider abstraction (LangChain4j 1.15.1) is message-based and maps to chat completions one-to-one; Responses is item-based, so its semantics would have to be synthesised rather than mapped. That contradicts the Assumption that no dependency upgrade is needed and the work stays in the provider client.
- Of the seven configured providers, only `open-ai-official` speaks Responses upstream. For the other six dotCMS would not be proxying the protocol, it would be inventing it on top of messages — a translation layer with no upstream equivalent to track.
- Responses stores conversations by default. Supporting that means per-site conversation persistence, retention and deletion, none of which is in this spec.
- Its native tools (`web_search`, `file_search`, `code_interpreter`) are OpenAI-hosted and cannot be offered across seven providers, so the implementation would necessarily be partial in a way clients cannot detect.

**A precedent worth looking at.** Vercel AI Gateway is close to what this spec describes — a multi-provider gateway reached through a custom base URL — and it exposes both formats as separate documented surfaces rather than picking one:

- https://vercel.com/docs/ai-gateway/sdks-and-apis/openai-chat-completions
- https://vercel.com/docs/ai-gateway/sdks-and-apis/responses

If that shape is also the intent here, then the question is sequencing rather than exclusion, and it would be worth saying so explicitly in Assumptions.

**One point that may be worth settling now regardless of which way this goes.** If a `/responses` surface is plausible later, the internal representation probably should not be modelled on chat-completions semantics with Responses bolted on afterwards — by then the item ids, reasoning items and typed event sequence it needs have already been discarded. Modelling the internal representation on the richer item semantics and treating chat completions as the lossy outbound adapter costs little today and makes a later `/responses` a serialization layer rather than a rewrite. That seems like a plan-phase instruction worth capturing here while the decision is still cheap.

Related: FR-001 says the version segment exists "so a future protocol lands beside it", which reads as though Responses would arrive as a v2. Worth confirming, since in the format being adopted Responses lives at `/v1/responses` — the same protocol version as `/v1/chat/completions` — so the natural landing spot would be `/api/inference/v1/responses` alongside, not a new version.

Not arguing for a change of direction — the reasoning for chat completions looks sound to me. Just that a spec this careful about recording why it declined #37431's criteria should probably record this one too, so the question does not get re-litigated in every review.

specs/37431-openai-compatible-inference/spec.md
fmontes: Response header, not a body field — your reasoning exactly: it keeps SC-008 unconditional and behaves identically for streams and errors. FR-020 now specifies the header, carrying the site identifier (hostname optional), on successes and errors alike. SC-005 updated to match. `ed09eaa`

specs/37431-openai-compatible-inference/spec.md
fmontes: FR-013 split along the line you drew: incidental fields ignored, semantic fields honored or rejected. `temperature`, `max_tokens`, `top_p` and `stop` now pass through explicitly; a semantic field we can't honor (`n`) is rejected with a validation error naming it rather than dropped.

specs/37431-openai-compatible-inference/spec.md
fmontes: Dropped, for the reason you give — the role branch is what FR-026 exists to eliminate. FR-023 now pins every caller regardless of role, and SC-007 is widened to match.

specs/37431-openai-compatible-inference/spec.md
fmontes: Gated. FR-009 now requires usage only when the client asks via the standard streaming option, and forbids emitting it otherwise. US2 scenario 3 covers both directions.

specs/37431-openai-compatible-inference/spec.md
fmontes: New FR-036: metadata only — site, model, token counts, status, duration — never request or response bodies, and nothing retained after the response. Family is stateless.

specs/37431-openai-compatible-inference/spec.md
fmontes: Stated plainly as you suggested. FR-031 and SC-010 now convey retryability by HTTP status (429 / 5xx); the body stays conformant to the standard error shape. The Error entity says retryability is not a field of it.

specs/37431-openai-compatible-inference/spec.md
fmontes: Base64 inline, no separately-addressable URL — which avoids having to decide storage, auth and lifetime for prompt-derived content at all. FR-012, with a note that any later URL form must be authenticated and time-limited.

specs/37431-openai-compatible-inference/spec.md
fmontes: FR-011 now applies FR-023 against the site's configured **embeddings** models, and states that the FR-024 alias resolves to the site's embeddings model on this operation.

specs/37431-openai-compatible-inference/spec.md
fmontes: New FR-037 plus SC-011: a maximum concurrent-stream count and a hard completion timeout, both configurable with stated defaults. Agreed it was the one flagged risk with no criterion attached.

specs/37431-openai-compatible-inference/spec.md
fmontes: FR-001 now states `/api/inference/v1` literally.

specs/37431-openai-compatible-inference/spec.md
fmontes: FR-015 now refuses session cookies and basic auth on this family — as you imply, admitting ambient browser credentials would make the absence of CORS headers a formality rather than a control.

specs/37431-openai-compatible-inference/spec.md
fmontes: Correct, and my error — introduced when FR-024 was inserted and everything after shifted. Fixed to FR-026 in both places.

specs/37431-openai-compatible-inference/spec.md
fmontes: Recorded in Assumptions, with your four costs and the both-formats precedent (Vercel, Portkey, LiteLLM — each shipped chat completions first). Framed as sequencing, not exclusion.

Two things changed because of this comment:

- **FR-001** corrected. You're right that Responses is a sibling at `/v1/responses`, not a v2 — so `/api/inference/v2` is now reserved for an incompatible revision of this family, and the sibling path is named explicitly.
- **FR-038** added, which is your last point and the one that changed my thinking. The internal representation must not be a direct binding of the chat-completions JSON, and tool-call identity is first-class rather than reconstructed from a streaming index — so a later `/responses` is a serializer rather than a re-plumb. Scoped deliberately to the internal shape: it adds no Responses-format behavior to this scope, since six of the seven providers are message-based upstream and I didn't want to build a Responses engine for an endpoint we aren't shipping.

Also took your framing on feasibility — for the other six providers dotCMS wouldn't be proxying the protocol, it'd be inventing it on top of messages. That's in the recorded rationale.

`ed09eaa`

nicobytes (approved):
Approving. Every point from the earlier review is resolved, and several are answered more thoroughly than what was asked.

Worth calling out what went beyond the review: SC-004 was reworked once the reserved alias was dropped, so the provider-swap criterion no longer rests on a mechanism that no longer exists; SC-009 moved from "byte-identical" to a contract-level invariant, which is the honest form given non-deterministic provider output; and FR-039 and FR-018 close two gaps nobody had raised — a stream that fails after delivering content, and a query parameter colliding with the legacy de-facto site override.

The decisions I found best argued are FR-023 refusing a role-conditional model passthrough on the grounds that it would reintroduce the branching FR-026 exists to eliminate, and FR-024 accepting that a provider swap does require callers to re-read the model list, rather than keeping a reserved alias that would have to be collision-proof against every provider's own naming. Both trade a convenience for an invariant that can actually be tested.

Five minor things, none blocking and none needing to be settled before planning:

1. **FR-010 does not carry the ordering guarantee FR-024 depends on.** FR-024's discovery story is that a caller reads the list and takes the first entry. That guarantee is stated in Key Entities ("The first entry is the site's primary model") but not in FR-010, which is the normative requirement defining the operation. The standard model-list shape has no ordering semantics, so this is a documented dotCMS convention rather than a protocol feature — worth saying so, since SC-001 claims zero dotCMS-specific client knowledge.

2. **LiteLLM is characterised in a way that undersells the argument.** The Responses assumption groups it with Vercel and Portkey as exposing both formats as separate surfaces. LiteLLM's `/responses` is native only for OpenAI and Azure; Anthropic, Bedrock, Vertex and Gemini are all bridged to chat completions. That is a stronger data point for the deferral than the one currently made — even the most translation-heavy gateway does not implement Responses natively across providers.

3. **Cloudflare AI Gateway is missing from that list and is the closest precedent.** Its unified multi-provider endpoint is chat completions, and `/ai/v1/responses` is scoped to models that speak it natively rather than synthesised across providers — which is exactly the shape FR-001 describes.

4. **FR-038's rationale claims more than the field practice supports.** The requirement itself is sound on its own terms: not binding the chat-completions JSON directly and making tool-call identity first-class is good domain modelling regardless of what comes later. But the justification that this makes a future `/responses` "a serializer rather than a rewrite" runs ahead of the evidence — the gateways above all bridge Responses down to chat completions rather than serializing both from a common richer model. Suggest keeping the requirement and softening the claim.

5. **Two small consistency items.** FR-007 states the terminal done marker unconditionally while FR-039 defines the exception, so FR-007 should cross-reference it — read alone it leads to the wrong implementation. And FR-037 requires bounds "all with stated defaults" without stating them or saying where they live.

PR #37494, opened 2026-09-10
fix(publishing-queue): scope Pending tab delete to its bundle and stop the dojo parse abort (#36861)

### LINKED ISSUE #36861: Publishing Queue → Pending tab "Delete" is a silent no-op when an asset is queued in two bundles (duplicate dijit widget id aborts dojo parse)
### Problem Statement

In Publishing Queue → Pending, selecting a bundle and clicking Delete does nothing — no error, no dialog, no change to the list, and nothing in the server logs. Because there is no other UI path to cancel a scheduled push publish, an author who schedules a bundle by mistake has no way to cancel it and it will publish on its due date.

**Root cause**

The asset-level checkbox id is built from the asset identifier and operation only, omitting bundle_id:

dotCMS/src/main/webapp/html/portlet/ext/contentlet/publishing/view_publish_queue_list.jsp:302

`id="queue_to_delete_<%=asset.get(PublishQueueElementTransformer.ASSET_KEY)%>$<%=asset.get(PublishQueueElementTransformer.OPERATION_KEY)%>"`

Compare the bundle-level checkbox 65 lines above, which does include it (line 237):

`id="bundle_to_delete_<%=bundle.get("bundle_id")%>"`

The Pending tab renders up to 10 bundles per page (limit default, line 56), each showing its first 20 assets (line 275). So when the same asset with the same operation is queued in two different bundles on the same page — entirely normal when shared content such as a site stripe or nav tile is included in multiple scheduled campaigns — the second dijit.form.CheckBox tries to register an id already in the registry, and dojo/parser aborts the remainder of the parse:

```
dojo/parser::parse() error Error: Tried to register widget with
id==queue_to_delete_c6026087b8b4fea6b96cad8456b1cf50$1 but that id is already registered
    at Object.add (build.js?b=26.07.06-3:1:4395)
    at Object.create (build.js?b=26.07.06-3:1:54094)
    ...
Error parsing in _ContentSetter#Setter_queueContent_6 Error: Tried to register widget with
id==queue_to_delete_c6026087b8b4fea6b96cad8456b1cf50$1 but that id is already registered
```

Why the abort makes Delete silently do nothing

Every checkbox after the collision point stays a plain <input> and is never upgraded to a widget. deleteQueue() and deleteBundle() (lines 142–172) then do:

`if (dijit.getEnclosingWidget(node).checked && !dijit.getEnclosingWidget(node).disabled) {`

For a non-widget <input>, dijit.getEnclosingWidget() walks up the DOM and returns the nearest enclosing widget — the queueContent ContentPane. A ContentPane has no .checked, so the expression is undefined → falsy → the node is skipped. It does not even throw, because .disabled is also undefined and !undefined === true.

Result: ids stays "" in both functions, so neither &delete= nor &deleteBundle= is ever appended to the URL. refreshQueueList(url) re-fetches the list with only layout/offset/limit. The pane reloads, no delete is ever requested, the server sees nothing and logs nothing.

This is deterministic and reproduces on every page load, including a hard refresh — the duplicate collides within a single parse pass, so there is no "click it again" or "reload and retry" recovery.

**Visual signature (useful for triage)**

Checkboxes render as large styled dijit widgets down to the first occurrence of the duplicated asset, then switch to small plain native checkboxes from the second occurrence onward. That visible break is the exact parse-abort point. Any bundle below it cannot be deleted.

Image

**Workaround**

The option below is safe in the sense that no content is deleted — publishing_queue holds queue entries (pointers to assets), not the assets themselves. Pages and contentlets are untouched.

1. Non-destructive — restores the Delete button (preferred)

Render one bundle per page so two copies of the same asset can never co-render. In the browser console on the Publishing Queue page:

`refreshQueueList("offset=0&limit=1");`
No deleteBundle parameter, so nothing is deleted. With the collision gone, the parse completes, checkboxes upgrade correctly, and the normal Delete button works — so the end user can complete the action themselves through the UI. This is the workaround to hand to customers.

### Steps to Reproduce

- As a CMS Admin, pick any single contentlet (a shared widget or content item).
- Add to Bundle → Bundle A. Push publish Bundle A with a future publish date.
- Add to Bundle → Bundle B, using the same contentlet as step 2. Push publish Bundle B with a different future publish date.
- Go to Publishing Queue → Pending. Both bundles are listed (default page size is 10, so both render together).
- Open the browser devtools Console.

Observed

On tab load, console shows Tried to register widget with id==queue_to_delete_<assetId>$1 but that id is already registered, thrown from dojo/parser::parse() and _ContentSetter#Setter_queueContent_N.
Checkboxes below the second occurrence of the duplicated asset render as unstyled native checkboxes.
Ticking any bundle checkbox below that point and clicking Delete does nothing. The Network tab shows a request to view_publish_queue_list.jsp without a deleteBundle= parameter. No server-side log entry. The bundle remains queued and will publish on its due date.

Expected

No dojo parser errors on load, all checkboxes functional, and Delete removes the selected bundle from the queue.

### Acceptance Criteria

- Asset checkbox id at view_publish_queue_list.jsp:302 is globally unique — include bundle_id, mirroring line 237. The value attribute needs no change, since deleteQueue() derives the asset id via nodeValue.split("$")[0].
-  With the same asset queued in two or more bundles on one page, the Pending tab loads with no dojo/parser errors and all checkboxes upgrade to dijit widgets.
-  Delete removes the selected bundle(s) and asset(s) from the queue in that scenario.
-  checkAllBundle() (line 123) correctly checks/disables child asset rows. It currently uses the descendant selector dojo.query(".b" + x + " input"), but the b<bundleId> class is on the <input> itself (line 299), so the query always returns an empty list. Should be dojo.query("input.b" + x).
-  deleteQueue() / deleteBundle() null-guard dijit.getEnclosingWidget() and surface a failure to the user rather than silently no-op'ing when no ids are collected.
-  view_publish_queue_list.jsp:223 no longer NPEs: permissionMap.get(bundle.get("bundle_id")).equals(Boolean.TRUE) returns null for a bundle with zero queue elements, because permissionMap is only populated inside the asset loop (lines 202–216).
-  Regression test covering one asset queued in two bundles rendered on the same Pending page

### dotCMS Version

Evergreen build 26.07.06-3

### Severity

### Links

https://helpdesk.dotcms.com/a/tickets/38574

danielsolis-dotcms:
**Spec-Kit PR 2 of 2.** Implementation for the spec approved in #37430.

---

## ⚠️ Spec corrected since #37430 — re-approval needed

@danielsilva-dotcms — you approved and merged the spec in #37430. It was corrected in **three** places during implementation. In all three the **code proved the spec wrong**, not the other way round. Each edit is marked inline in `spec.md` (strikethrough + a dated "revised" note) so the change is reviewable rather than silent.

| | What the spec said | What is true |
|---|---|---|
| **1** | Fix `checkAllBundle()` to use `dojo.query("input.b" + x)` | That change **breaks** the cascade. `dijit.form.CheckBox` moves the source node's classes onto its wrapper `<div>` and nests the real `<input>` inside, so `.b<id> input` is correct. Implemented, went red under test, reverted. The selector now carries a comment so it isn't "fixed" again. |
| **2** | AC-008: the NPE affects "a bundle with **zero queue elements**" | That state cannot reach the code path — the bundle list comes from `select distinct(bundle_id) from publishing_queue`, so every bundle shown is derived from queue rows. The reachable trigger is a **NULL `bundle_id`**. |
| **3** | Residual "zero queue elements" wording in Fix Scope and Assumptions | Same as (2); also described the remedy as "null-guard the lookup", whereas the block was restructured so an absent entry *means* "not permitted". |

Please re-approve the spec alongside this code.

---

## The bug

Delete did nothing — no error, no dialog, no server log — whenever one asset was queued in two bundles on the same page. Shared content in two scheduled campaigns is enough to trigger it, and the Pending tab is the only UI path to cancel a scheduled push publish.

The asset checkbox id was built from asset + operation only, so a duplicated pair collided in the dijit registry. `dojo/parser` threw and abandoned the rest of the parse. `deleteQueue()` resolves nodes via `dijit.getEnclosingWidget()`, which walks **up** and returns the enclosing `ContentPane` — no `.checked` (undefined, falsy), no `.disabled` (undefined, so `!undefined` is true). Selections were skipped without throwing.

## Proposed Changes

* **Unique checkbox id** — include `bundle_id`, so the parse completes.
* **Bundle-scoped delete** — new additive `PublisherAPI.deleteElementFromPublishQueueTableAndAuditStatus(String, String)`. Deleting an asset from one bundle no longer clears it from every bundle holding it. The old overload is retained and `@Deprecated(forRemoval = true)` per the **ADR-0020** precedent. A null/blank bundle throws rather than silently widening to `WHERE asset = ?`.
* **No more silent failure** — the delete functions reject non-checkbox widgets and raise a localized `alert()` instead of reloading the pane as if it had worked.
* **Batched permissions** — one `PermissionAPI.filterCollection` call replaces the per-bundle `doesUserHavePermission` loop (**ADR-0020**), and each bundle's queue elements are resolved once instead of twice. An unresolvable bundle is absent from the permitted set, which removes the NPE structurally.

### `PublishQueuePermissionFilter` — a class the plan did not foresee

The permission logic lived in a JSP scriptlet, where AC-010 ("the batched filter yields the same visible bundle list as the per-item loop") could not be asserted by any test. Extracting it made the gate real. It preserves two behaviours deliberately, both of which silently change who sees what if got wrong: `respectFrontendRoles = true` (matching the 3-arg overload the portlet called — `false` looks safer and tightens permissions), and **only the first queue element decides** (the loop it replaces `break`s after one).

## Checklist

- [x] **Tests** — 5 new integration tests (`Tests run: 13`) and a 7-case Playwright suite. **Every one was confirmed failing for the intended reason before its fix.** Twice that discipline caught a test that could never fail; see below.
- [x] **Translations** — `publisher_delete_nothing_selected` added to `Language.properties` via `LanguageUtil`, English-only per existing precedent.
- [x] **Security Implications Contemplated** — the permission batch is the risk surface: a wrong filter changes bundle visibility silently in either direction. Gated by a parity test comparing batched vs per-item results for an admin and a limited contributor. No schema, ES mapping or REST contract change; not rollback-unsafe.

## Two findings that corrected the issue's own diagnosis

Both were caught only by insisting a test go red first:

1. **`checkAllBundle()` was never broken by its selector** (see table above). It failed only as a *consequence* of the parse abort — no widgets, no wrapper, nothing to match. Fixing the duplicate id resolved AC-006 with no further change.
2. **The first "checkbox upgraded" assertion could not fail.** It selected `input[type=checkbox][id^=...]` as "un-upgraded", but dijit keeps the original `<input>` and puts the widget id on it — so it matched upgraded and un-upgraded alike. It now asks `dijit.byId()`. Even corrected, it passed on unfixed code with two bundles, because the collision was the last thing on the page; the fixture seeds **three** so something renders below it.

## Unrelated drive-by fix — `contentlets.ts`

`createContentlet` returned the workflow-fire response's `entity` unwrapped. That endpoint returns `{ results: [{ "<id>": {...} }], summary }`, so **`contentlet.identifier` was `undefined` for every caller** — meaning `deleteContentlets(request, [contentlet.identifier])` in `afterEach` hooks silently deleted nothing and **e2e contentlets have been leaking**.

This is unrelated to #36861 and is called out rather than left to ride along. It could not simply be dropped: the new fixtures depend on a real identifier, so reverting it breaks this suite. The unwrapping mirrors what `fixtures/relationship.fixture.ts` already does.

**Suggested follow-up issue:** audit and clean already-leaked e2e content types and contentlets on shared environments.

## Additional Info

* **Docs** — two ways an integration run reports success having executed **zero** tests are now documented (`CLAUDE.md` Critical Rules + `docs/testing/INTEGRATION_TESTS.md`): the Maven build cache short-circuiting failsafe, and `-Dit.test=Class#method` selecting nothing. `AGENTS.md` gains the Dojo-portlet rules this work needed.
* **Verified** — integration and e2e both green from a rebuilt image; re-verified after rebasing onto current `main`. `/speckit-converge` and `/speckit-docs-converge` both report **converged**.
* **Line numbers in the issue are stale** (302/237/223/123 from build 26.07.06-3); current `main` equivalents are 313/248/234/134.

## Screenshots

n/a — the fix is the *absence* of a console error. The issue carries the pre-fix visual signature: styled dijit checkboxes above the collision point, unstyled native ones below.

Fixes #36861

This PR fixes: #36861


core-web/apps/dotcms-ui-e2e/src/tests/publishing-queue/pending-duplicate-asset.spec.ts
```diff
@@ -0,0 +1,202 @@
...
+        //
+        // THREE, not two, and that matters. The parser registers the first checkbox, collides on
+        // the second and aborts the rest of the pass. With only two bundles the collision is the
+        // last thing on the page, so nothing is left un-upgraded and the damage is invisible to
+        // a DOM assertion. The third bundle is what renders *below* the collision and never gets
+        // upgraded — that is the user-visible breakage this suite has to catch.
+        seededBundleIds = await pushPublishIntoSeparateBundles(request, contentlet.identifier, 3);
+    });
+
+    test.afterEach(async ({ request }) => {
```
danielsolis-dotcms, line 46: **Seeded bundles are never cleaned up**

`afterEach` deletes the contentlet and the content type, but nothing deletes the bundles or the `publishing_queue` rows that `pushPublishIntoSeparateBundles` created — `seededBundleIds` is only ever read inside tests, and there is no delete-bundle helper anywhere in `src/requests/`.

`beforeEach` seeds 3 bundles for each of the 7 tests, so every full run leaks 21 scheduled push-publish bundles. Two concrete consequences:

1. `bundles.ts` states *"Nothing is ever actually shipped to a remote server: the bundle is scheduled days ahead and the test deletes its content long before the job would run."* The **content** is deleted, but the bundle and its queue row survive — so once `futureDate(1)` elapses, `PublisherQueueJob` picks it up and attempts a real push to the resolved environment.
2. The Pending tab pages 10 bundles at a time ordered by `publish_date` (`ETBUNDLES`, `PublisherAPIImpl.java:607`, `getQueueBundleIds` with limit 10). Leaked bundles from earlier runs carry *earlier* publish dates and therefore sort ahead of the current run's. After roughly four runs on a shared environment, this run's `bundleA` is no longer on page 1 — `waitForReady()` still passes (bundles render), but `assetIdsForBundle(bundleA)` returns `[]` and `expect(...).toContain(contentlet.identifier)` fails in *"deleting an asset row removes it from that bundle only"*.

This is the same leak class the PR description calls out and fixes for contentlets. A `deleteBundle` request helper plus an `afterEach` loop over `seededBundleIds` would close it.

dotCMS/src/main/java/com/dotcms/publisher/business/PublishQueuePermissionFilter.java
```diff
@@ -0,0 +1,111 @@
...
+                    .computeIfAbsent(permissionId, key -> new HashSet<>())
+                    .add(entry.getKey());
+        }
+
+        if (proxiesByPermissionId.isEmpty()) {
+            return permitted;
+        }
+
+        final PermissionAPI permissionAPI = APILocator.getPermissionAPI();
+        final Collection<PermissionableProxy> allowed = permissionAPI.filterCollection(
```
danielsolis-dotcms: **The batched filter drops the folder `PUBLISH` → `EDIT` substitution**

`filterCollection(Collection, int, User, boolean)` and `doesUserHavePermission` are not equivalent for folder-typed permissionables, so this is a third preserved-behaviour case alongside the two already documented in the class javadoc.

`PermissionBitAPIImpl.doesUserHavePermissionInternal` (~line 372) does:

```java
final int expecterPermissionType = (PermissionableType.FOLDERS.getCanonicalName()
        .equals(permissionable.getPermissionType()) || ...)
        && permissionType == PERMISSION_PUBLISH
        ? PERMISSION_EDIT
        : permissionType;
```

with the comment *"Folders do not have PUBLISH, use EDIT instead"*. The batched path goes straight to `PermissionBitFactoryImpl.getPermittedIds`, which tests the raw `PERMISSION_PUBLISH` bit and never inspects the permissionable's type.

This is reachable: `PermissionableProxy.setType("folder")` maps to `PermissionableType.FOLDERS`, and `PusheableAsset.FOLDER("folder")` means folder-typed queue elements are real.

**Scenario:** a contributor role holds View + Edit (no Publish) on folder `F`. A bundle whose *first* queue element is `F` was visible on the Pending tab before this change and is silently hidden after it — the user can no longer see or cancel their own scheduled push, and the Pending tab is the only UI path to do so.

The parity test can't catch this because it seeds only contentlets (see the separate comment on `PublisherAPIImplTest.java`).

dotCMS/src/main/java/com/dotcms/publisher/business/PublishQueuePermissionFilter.java
```diff
@@ -0,0 +1,111 @@
...
+ * </ul>
+ * Extracted from the JSP so this rule is unit-testable; the portlet holds no permission logic of
+ * its own.
+ *
+ * @see <a href="https://github.com/dotCMS/core/issues/36861">Issue #36861</a>
+ */
+public final class PublishQueuePermissionFilter {
+
+    /** Matches the 3-arg {@code doesUserHavePermission} overload the portlet used to call. */
+    private static final boolean RESPECT_FRONTEND_ROLES = true;
```
danielsolis-dotcms: **`respectFrontendRoles` does not mean the same thing in the batched API**

The comment says this matches the 3-arg `doesUserHavePermission` overload, and the *value* does — but the two APIs interpret the flag differently, so passing `true` is not behaviour-preserving.

`PermissionBitAPIImpl.filterUserRoles` (line 2047) seeds the role set from `loadRolesForUser(user)` and, when `respectFrontendRoles` is **false**, only *removes* the anonymous and logged-in-site roles. It never *adds* a role the user does not hold.

The batched `filterCollection` (lines ~1440-1444) instead does, when `respectFrontendRoles` is true:

```java
roleIds.add(anonRole.get().getId());
roleIds.add(APILocator.getRoleAPI().loadLoggedinSiteRole().getId());
```

unconditionally, for every caller.

For the **anonymous** role this stays at parity, because `doesUserHavePermissionInternal` has a matching early-return inside its perms loop (and `isContentlet` is false for a `PermissionableProxy`, so the live-contentlet guard doesn't apply). For the **Logged In Site User** role there is no such grant in the old path — only the dangling `//if logged in site user has permission` comment.

**Scenario:** bundle X's head asset grants `PUBLISH` to *Logged In Site User*. A backend-only contributor who does not hold that role could not see bundle X before; now they can see it and delete its queue entries.

Either pin the behaviour with a test that actually grants a frontend-role permission, or pass `false` and document that as the deliberate choice.

dotcms-integration/src/test/java/com/dotcms/publisher/business/PublisherAPIImplTest.java
```diff
@@ -151,6 +152,218 @@ public void test_deleteElementsFromPublishQueueTable_TwoAssets() throws DotPubli
...
+                    .addAssets(CollectionsUtils.list(contentlet))
+                    .nextPersisted();
+
+            final List<PublishQueueElement> elements =
+                    PublisherAPIImpl.getInstance().getQueueElementsByBundleId(bundle.getId());
+            Assert.assertFalse("seeded bundle should have a queue element", elements.isEmpty());
+            firstElementByBundle.put(bundle.getId(), elements.get(0));
+        }
+
+        final List<User> users = list(
```
danielsolis-dotcms: **This parity test cannot fail — neither user produces a partial result**

AC-010 is the gate for the batched permission change, and the PR description leans on it (*"Gated by a parity test comparing batched vs per-item results for an admin and a limited contributor"*). But both users short-circuit:

- **`APILocator.systemUser()`** — both paths bypass. `doesUserHavePermissionInternal` returns `true` for the system user; `filterCollection` returns `new ArrayList<>(permissionables)`. So `expected == actual ==` all 3 bundles, whatever the filter body does.
- **`TestUserUtils.getJoeContributorUser(host)`** — returns the *pre-existing* `joe@dotcms.com` whenever one exists (`TestUserUtils.java:224-227`), in which case the `getOrCreateContributorRole(host)` grant for this test's freshly created `SiteDataGen` host is never applied. Joe therefore holds no permission on any of the 3 seeded contentlets, and `expected == actual ==` the empty set.

Replace the whole filter body with `return new HashSet<>(firstElementByBundle.keySet())` and the contributor case still passes; replace it with `return new HashSet<>()` and the system-user case still passes. This is the same *"assertion that could not fail"* pattern the PR description flags twice elsewhere.

The two divergences I raised on `PublishQueuePermissionFilter` (folder `PUBLISH`→`EDIT`, and the injected Logged-In-Site role) both live in exactly the gap this test leaves open. A user with a real, *partial* grant — permitted on one seeded bundle's asset and not another — is what would make this assertion load-bearing.

dotCMS/src/main/java/com/dotcms/publisher/business/PublisherAPIImpl.java
```diff
@@ -862,6 +869,38 @@ public void deleteElementFromPublishQueueTableAndAuditStatus(final String identi
...
+			dc.addParam(identifier);
+			dc.addParam(bundleId);
+			dc.loadResult();
+		} catch (final Exception e) {
+			Logger.error(PublisherUtil.class, e.getMessage(), e);
+			throw new DotPublisherException("Unable to delete element " + identifier
+					+ " from bundle " + bundleId + " :" + e.getMessage(), e);
+		}
+
+		final List<PublishQueueElement> remaining = Try.of(() -> getQueueElementsByBundleId(bundleId))
```
danielsolis-dotcms: **A swallowed lookup failure deletes a live bundle's audit status**

`Try.of(...).getOrElse(Collections::emptyList)` cannot distinguish *"the bundle is now empty"* from *"the lookup failed"*, and both produce the same empty list — which then takes the `deletePublishAuditStatus` branch.

`getQueueElementsByBundleId` throws `DotPublisherException` on any DB or mapping error (lines 769-772). If that happens after the row delete has already succeeded — connection reset, statement timeout on a large bundle — `remaining` becomes `emptyList()`, and the `PublishAuditStatus` is wiped for a bundle that still has queued assets. The bundle then publishes with no audit trail, the Status/History tab shows nothing for it, and the underlying error is never logged or surfaced.

This copies the pattern from the deprecated single-arg overload, where the `Try` at least had a purpose (`bundleId` came from `queueElements.get(0)` and could legitimately be null). Here `bundleId` is already validated non-blank at line 875, so the only thing the `Try` now swallows is a genuine failure.

Letting it propagate would be both simpler and safer — the method already declares `DotPublisherException`, and `@WrapInTransaction` means the delete rolls back with it.

dotCMS/src/main/webapp/html/portlet/ext/contentlet/publishing/view_publish_queue_list.jsp
```diff
@@ -150,15 +162,32 @@
...
 		   dojo.forEach(nodes, function(node) {
-			   if(dijit.getEnclosingWidget(node).checked && !dijit.getEnclosingWidget(node).disabled){
-				   var nodeValue = dijit.getEnclosingWidget(node).value;
-				   ids+=","+nodeValue.split("$")[0];
+			   var box = queueCheckBoxFor(node);
+			   if(box && box.checked && !box.disabled){
+				   // value is <asset>$<operation>$<bundleId>. Send asset AND bundle so the server
+				   // deletes only this bundle's entry - see issue #36861.
+				   var parts = box.value.split("$");
+				   ids+=","+parts[0]+"$"+parts[2];
```
danielsolis-dotcms: **The `$`-delimited wire format mis-parses assets containing `$`, silently**

`$` is used unescaped on both sides — `box.value.split("$")` here, `entry.split("\\$")` at line 89 — and neither end validates the segment count. Client-side this takes `parts[0]` and `parts[2]` positionally; server-side it takes `parts[0]` and `parts[1]`.

`addAssetsToQueue` stores the raw identifier in `publishing_queue.asset`, and for OSGi bundles that identifier is a jar file name (`identifier.contains(".jar")` → `PusheableAsset.OSGI`, `PublisherAPIImpl.java:207`), which may legally contain `$`.

**Scenario:** asset `my$plugin.jar` in bundle `B` renders as `my$plugin.jar$1$B`. The client sends `parts[0] + "$" + parts[2]` = `my$1`. The server parses `parts[0]="my"`, `parts[1]="plugin.jar"` — `UtilMethods.isSet` passes, so it runs `DELETE FROM publishing_queue WHERE asset='my' AND bundle_id='plugin.jar'`, matches zero rows, and the pane reloads looking exactly like a successful delete. That is the same silent no-op this PR exists to remove.

The guard at line 89 only rejects *fewer* than 2 segments, and `continue`s without recording anything, so even a genuinely malformed entry produces no user-visible feedback — while the new `alert()` path covers only the empty-selection case.

Checking `parts.length === 3` client-side and requiring exactly 2 segments server-side (with an error path rather than a bare `continue`) would close it. Encoding the pair as two request parameters, or a delimiter that cannot appear in an identifier, would close it properly.

danielsolis-dotcms (commented):
Code review of the publishing-queue delete change. The core fix holds up — the bundle-qualified widget id, the bundle-scoped `DELETE ... WHERE asset = ? AND bundle_id = ?`, the null-bundle guard, and the `queueCheckBoxFor` widget-type check all do what they claim, and the NPE is removed structurally rather than patched. The three spec corrections also check out against the code: the `.b<id> input` selector is correct as written, the zero-queue-elements state genuinely cannot reach that path (`ETBUNDLES` derives the bundle list from queue rows), and a NULL `bundle_id` is the reachable trigger since `bundle_id = ?` never matches NULL.

Six findings inline, clustered around the batched permission filter and the new e2e suite. The most consequential are the uncleaned e2e bundles (they leak *and* eventually fire a real push) and the fact that the AC-010 parity test cannot fail as written — which is what lets the two permission-semantics divergences through.

core-web/apps/dotcms-ui-e2e/src/tests/publishing-queue/pending-duplicate-asset.spec.ts
danielsolis-dotcms: Fixed in `4d9e8669ed`.

Confirmed — there was no delete-bundle helper anywhere in `src/requests/`, and a 7-test run leaked 21 scheduled bundles. Both consequences you named were real.

Added `deleteBundle()` (`DELETE /api/v1/publishing/{bundleId}` → `deleteBundleAndDependencies`) and an `afterEach` loop over `seededBundleIds`, ordered **before** the contentlet delete.

Verified directly rather than by inspection: a full 7-test run now leaves **0 bundles and 0 `publishing_queue` rows**. This was invisible locally because I had been clearing bundles by hand with SQL after every run — on CI they would simply have accumulated.

dotCMS/src/main/java/com/dotcms/publisher/business/PublishQueuePermissionFilter.java
danielsolis-dotcms: Fixed in `4d9e8669ed`.

Confirmed at `PermissionBitAPIImpl:196-201` and `:373` — the scalar path substitutes `EDIT` for folders and the batched path does not, since `filterCollection` takes one permission type for the whole collection and never inspects the permissionable.

The collection is now partitioned: folder-typed proxies are asked for `PERMISSION_EDIT`, everything else for `PERMISSION_PUBLISH`. Still batched — two queries at most, not one per bundle — so the ADR-0020 intent is preserved.

You were right that this is a third preserved-behaviour case the class javadoc claimed to cover and didn't.

dotCMS/src/main/java/com/dotcms/publisher/business/PublishQueuePermissionFilter.java
danielsolis-dotcms: Fixed in `4d9e8669ed`.

Confirmed at `PermissionBitAPIImpl:2047-2063`: `filterUserRoles` seeds from `loadRolesForUser` and only **removes** the anonymous and Logged-In-Site roles when the flag is `false` — it never adds one the user doesn't hold. The batched path **adds** both when the flag is `true`.

The resolution is tidier than expected: passing **`false`** to the batched API reproduces the scalar `true` behaviour exactly, because the batch's only use of the flag is to add roles. The value flips; the semantics are preserved. The reasoning is now recorded on the constant itself.

My original research note asserted `true` was the parity-preserving choice. That was wrong — it preserved the *value*, not the *semantics*, which is exactly the distinction you drew.

dotcms-integration/src/test/java/com/dotcms/publisher/business/PublisherAPIImplTest.java
danielsolis-dotcms: Fixed in `4d9e8669ed`.

Both short-circuits confirmed, including `TestUserUtils:224-227` returning the shared pre-existing `joe@dotcms.com` and skipping the role grant for this test's fresh host.

The test now creates **its own** role and user, and grants `PUBLISH` on **one of three** seeded assets — saving an individual permission also breaks inheritance, so nothing ambient can widen it. Beyond the parity loop it asserts the limited user sees a **strict, non-empty subset**: exactly 1 bundle, and specifically the granted one.

Proved load-bearing rather than assumed: stubbing the filter to `return new HashSet<>(firstElementByBundle.keySet())` makes it fail (`Tests run: 2, Failures: 2`), then reverted. Given this was the third assertion-that-could-not-fail in this change, a green run on its own wasn't worth much.

Your observation that this vacuous gate is what let the two permission divergences through is exactly right — both were found by review, not by the test that existed to catch them.

dotCMS/src/main/java/com/dotcms/publisher/business/PublisherAPIImpl.java
danielsolis-dotcms: Fixed in `4d9e8669ed`.

Agreed on all points. `bundleId` is validated non-blank at line 877, so the `Try` — inherited from the deprecated overload where it had a genuine purpose — was only ever swallowing real failures and turning them into "the bundle is empty".

Removed. The exception now propagates, the method already declares `DotPublisherException`, and `@WrapInTransaction` rolls the delete back with it. A comment records why it is deliberately *not* wrapped, so it doesn't get reinstated.

dotCMS/src/main/webapp/html/portlet/ext/contentlet/publishing/view_publish_queue_list.jsp
danielsolis-dotcms: Fixed in `4d9e8669ed`.

Confirmed at `PublisherAPIImpl:207` — `identifier.contains(".jar")` means an OSGI asset identifier is a jar file name, which may legally contain `$`. Your worked example is right, and the failure mode is the same silent no-op this PR exists to remove.

Both sides now parse from the **right** (`lastIndexOf('$')` server-side; last-two-segments client-side, rejoining the remainder). Bundle ids and operations never contain `$` while assets may, so right-to-left is unambiguous and `my$plugin.jar$1$B` round-trips correctly.

I kept the `$` pair encoding rather than redesigning the wire format — right-to-left parsing is sufficient and correct, and a smaller change to a legacy JSP. A malformed entry now raises a localized message on both sides instead of a bare `continue`.

gortiz-dotcms (approved):

PR #37500, opened 2026-09-10
fix(index): Phase 2 ES read fallback now fires for content search (#37413)

### LINKED ISSUE #37413: Phase 2 ES read fallback never fires: content read path bypasses PhaseRouter, OpenSearch outage returns empty results

fabrizzio-dotCMS:
Fixes #37413. Spec approved in #37438 — this PR branches off that spec branch, so it carries the spec commit too.

## What was broken

The Phase 2 Elasticsearch read fallback was implemented and correct. The content read path never reached it.

`ESContentFactoryImpl` selected a provider with a bare ternary and called it directly from its five read call sites, so `PhaseRouter` — which holds the fallback — was unreachable from the busiest read path in the product: `/api/content/_search`, `ContentletAPI` search and count, Velocity `$dotcontent.pull`, URL maps, Site Search, scroll consumers and the admin content browser.

With OpenSearch down in Phase 2 and Elasticsearch healthy and dual-written, a content search returned a well-formed `200` with **zero results** for content types whose count was already cached, and a `500` for the rest. The empty result is the dangerous variant: a caller cannot tell it apart from "this content type has no content", so a live site rendered as missing content with nothing for 5xx monitoring to catch.

This blocked more than the outage case. Cloud and Support have been telling customers "Phase 2 is safe, OpenSearch failures fall back to Elasticsearch" — quoting `OPENSEARCH_MIGRATION.md`, which states it in four places. For content search that was false.

## Two failure classes, not one

Routing alone fixes an **unreachable** OpenSearch: a connection failure reaches the provider's generic handler and is rethrown as a runtime exception the router catches.

A node that is **reachable but answering with an error** is different, and routing does not fix it. The provider absorbs the failure into a legitimate-looking success, reports no failure, and the router has nothing to catch. Four such paths, all now conditional on Phase 2:

| Path | Absorbed into |
|---|---|
| `searchHits` index resolution | empty `SearchHits` — a `catch (Exception)` around `inferIndexToHit`, plus a null-index guard |
| `cachedIndexSearch` | the `ERROR_HIT` sentinel |
| `cachedIndexCount` | `-1` |
| `indexSearchScroll` | an empty list, via its own separate handler |

**Two of those four were not in the issue's analysis** — the index-resolution guard and the scroll handler. The first is the most consequential: it sits on the exact path the issue reports and converts *any* failure to determine which index to query into a clean empty result.

It also explains the reported asymmetry more precisely than the spec does. The spec attributes the empty `200` to the legacy `ContentUtils` `catch (Throwable)`. That is true but stops one layer short — there is an earlier swallow in the provider, and the *same* index-resolution failure propagates on the count and scroll paths (so the router falls back and the caller gets correct data) while the search path converted it to empty on the spot.

## Why the Phase 2 scoping is what makes this safe

AC-006 asked for an enumeration of every caller that depends on receiving an empty result rather than an exception, warning that if the list were large the provider change should be dropped.

Scoping the raise to Phase 2 makes the enumeration unnecessary: **in Phase 2 the router catches the new exception immediately above the provider and turns it into a successful Elasticsearch read, so no caller ever observes a new exception type.** Phases 0 and 1 do not read from OpenSearch at all. Phase 3 has no Elasticsearch to fall back to, so raising there would expose every such caller with nothing gained.

The one genuine behaviour change: **both engines failing on the same read now surfaces an error instead of a silent empty result.** That is the improvement being asked for.

`ContentFactoryIndexOperationsES` is deliberately untouched.

## The log line

The fallback is the migration design's early-warning signal, so it now names the operation and the root cause. Before, it could only report the wrapper's message — enough to know something fell back, not what stopped working.

```
OS read failed in Phase 2 [indexCount] — falling back to ES. OS index may be stale or
unavailable. Cause: An error occurred when executing the Lucene Query
/ root cause: ConnectException: Connection refused
```

The raised messages carry the index name and OpenSearch's own reason, **not** the query body: in Phase 2 the router logs the message at `ERROR`, and a Lucene query can carry end-user search terms (Constitution Principle III). Residual, unchanged: the provider's pre-existing `WARN` blocks still log the full query, in every phase, and the Elasticsearch provider does the same. Changing that would alter Phase 0/1 logging, which the spec declares out of scope.

## Cache poisoning

On the two cached paths the raise happens **before** the cache write. The provider caches its sentinel for `parse_exception` and `search_phase_execution_exception`; such an entry would be replayed to every later identical query as a successful empty result, outliving the outage and defeating the fallback even after OpenSearch recovers. A test asserts the ordering rather than trusting it.

## Testing

- **Unit, 14 new** (`ContentFactoryIndexOperationsPhaseRoutingTest`, in no suite): all five read operations × all four phases, on the router against the real provider interface; `ERROR` level and one event per read; both-engines-fail surfaces after exactly one attempt each; a healthy Phase 2 never touches Elasticsearch.
- **Integration, 13 new** (both registered in `MainSuite1b`): `ESContentFactoryImplPhase2FallbackTest` proves the wiring by injecting an OpenSearch provider that fails every read; `ESContentFactoryImplMissingOsIndexTest` covers the reachable-but-erroring class with a mocked client raising `index_not_found_exception`.
- Regression: 54 unit tests green across the router, both providers and the Phase 2 read-durability tests.

Phases 0, 1 and 3 are verified unchanged **by test**, not by inspection — these call sites carry essentially all content search, so a behaviour change here is felt site-wide in every phase.

Two notes on how the tests got there, because both are traps worth knowing:

**The unit tests do not go red for this bug.** 13 of the 14 pass against unmodified code, because the router was never broken — the defect was that nobody called it. Only the integration test can catch that, and it is why the wiring proof is not optional.

**The first version of the missing-index test passed for the wrong reason.** It relied on the container simply having no OpenSearch counterpart indices, which looked like the customer scenario but produced dotCMS's own index-resolution failure instead — thrown before the client is ever called, so it re-exercised the routing fix, and `index_not_found_exception` never appeared once in the run. Its precondition only checked that OpenSearch did not return the right answer, not that it failed for the right reason. The class now mocks the client and asserts the precondition explicitly.

## Reviewer decisions I need

1. **AC-006's enumeration** was not produced; the reasoning above is offered in its place. Accept it or ask for the list.
2. **AC-007's cache-hit clause** says a caller must never receive "a silent stale hit presented as current". A cache hit during an outage still does exactly that, and this PR treats it as correct rather than a defect — the cached value is a real result from a real earlier query, and the provider cannot distinguish an outage from a quiet period. The error *sentinel* half is fixed and tested.

## Two signed-off test gaps

- The **scroll swallow** has no behavioural test: the scroll resolves its client through CDI rather than the injected provider, so a mocked client cannot reach it. The fix covers it; only the proof is missing.
- That **`ContentFactoryIndexOperationsES` is unchanged** is evidenced by the diff, not by a test: its client is a static singleton with no injection seam, so no `index_not_found` can be forced through it. The Phase 0/1 tests do show Elasticsearch-served reads behaving identically.

## Docs

`OPENSEARCH_MIGRATION.md` now records what must hold for the fallback to fire, and — more useful operationally — what still does **not** fall back: Phase 3, a failure mid-scroll, an OpenSearch answering successfully with stale data, a repeated identical query served from cache, and the legacy layer that turns any read failure into an empty `200` in every phase including pure-Elasticsearch installs.

**TC-039 in the test plan is the QA case that should have caught this, and it was structurally guaranteed to pass.** Step 3 told the tester to run "the same search again" — an identical request body is served from the query cache and returns the pre-outage result, indistinguishable from a working fallback. It now requires varying the query, adds the never-queried-content-type variant that used to fail as a `500`, and warns about checking the returned total rather than just the absence of an error.

## Not in scope

- The legacy swallow in `ContentUtils` and `ContentHelper` that turns any read failure into an empty `200` in every phase — pre-existing, affects pure-Elasticsearch installs, and is what made this outage silent rather than loud. Being filed separately.
- No fallback in Phase 3, by design.
- No `IndexAPI<F>` generic parameterization.
- No write-path changes, no new config property or feature flag — the fallback is documented, unconditional Phase 2 behaviour.

## Still to do before merge

Manual validation on the Phase 2 migration environment, varying `offset` on every call, plus confirming Phase 3 still propagates there. Not yet done.

### DIFF
--- a/docs/backend/OPENSEARCH_MIGRATION.md
+++ b/docs/backend/OPENSEARCH_MIGRATION.md
@@ -108,6 +108,46 @@ In Phase 2 OS serves reads but ES is still active. If OS throws an exception on 
 `PhaseRouter` catches it, logs at **ERROR** level, and retries against ES automatically.
 
 - The caller receives a correct result from ES
+
+The fallback log line names the operation and the root cause, so an outage is identifiable from
+logs alone:
+
+```
+OS read failed in Phase 2 [indexCount] — falling back to ES. OS index may be stale or
+unavailable. Cause: An error occurred when executing the Lucene Query
+/ root cause: ConnectException: Connection refused
+```
+
+**Content search reached this fallback only from #37413 onward.** Before that fix,
+`ESContentFactoryImpl` selected a provider with a bare ternary and called it directly from its
+five read call sites, so the router — and therefore the fallback — was unreachable from the
+busiest read path in the product. An OpenSearch outage in Phase 2 returned a well-formed `200`
+with zero results for content types whose count was already cached, and a `500` for the rest.
+The empty result is the dangerous variant: a caller cannot tell it apart from "this content type
+has no content". If you are reading this against a build that predates that fix, the guarantee
+above did not hold for `/api/content/_search`, `ContentletAPI` search and count, Velocity
+`$dotcontent.pull`, URL maps, Site Search or the admin content browser.
+
+Two conditions are required for the fallback to fire, and both took work:
+
+1. **The read must go through `PhaseRouter`.** Provider selection in `ESContentFactoryImpl` now
+   has exactly one mechanism; a second one is what caused #37413.
+2. **The OpenSearch provider must actually raise.** In Phase 2, `ContentFactoryIndexOperationsOS`
+   raises instead of absorbing a failure into a legitimate-looking success — the `ERROR_HIT`
+   sentinel, `-1` for a count, an empty scroll list, and the index-resolution guard in
+   `searchHits` that returned an empty result for *any* exception. A provider that reports
+   success gives the router nothing to catch. Outside Phase 2 the absorbing behaviour is
+   unchanged, and `ContentFactoryIndexOperationsES` is not touched at all.
+
+**What still does not fall back**, by design or by limitation:
+
+| Situation | Behaviour | Why |
+|---|---|---|
+| Phase 3 read failure | Propagates | ES is decommissioned; a silent fallback would report data that is genuinely absent |
+| Failure mid-scroll | Propagates | An OS scroll id is meaningless to ES, so a half-drained scroll cannot be resumed on the other engine. `createScrollQuery` is routed for provider selection only |
+| OS answers successfully with stale or incomplete data | No fallback | Indistinguishable from a query that legitimately matches nothing; falling back on every empty result would double the load of most searches |
+| A repeated identical query during an outage | Served from the query cache | The cached value is a real earlier result, so serving it is correct — but it makes an outage invisible when reproducing by hand. Vary `offset` per call. The error *sentinel* is never cached on the Phase 2 path, so a poisoned entry cannot outlive the outage |
+| Any phase, read failure reaching the legacy layer | Empty `200` | `ContentUtils`' `catch (Throwable)` and `ContentHelper`'s `resultsSize = 0` overwrite predate the migration and affect pure-ES installs too. Tracked separately; this is why the outage was silent rather than loud |
 - The ERROR log makes the OS failure visible for operators
 - In Phase 3 there is no fallback — ES is decommissioned and OS failures propagate normally
 
--- a/docs/backend/OPENSEARCH_MIGRATION_TEST_PLAN.md
+++ b/docs/backend/OPENSEARCH_MIGRATION_TEST_PLAN.md
@@ -834,12 +834,26 @@ When the migration starts **successfully** (no shutdown) you instead see an `INF
 - **Steps:**
   1. Confirm content is searchable in Phase 2.
   2. Break OpenSearch reads (stop OS or remove the OS working index).
-  3. Run the same search again.
+  3. Search again — but **change the query on every call** (see the warning below).
+  4. Also search a content type you have **not** queried since the server started, so its count is
+     not cached either. That variant used to fail differently: a `500` instead of an empty `200`.
 - **Expected Result:**
-  - The search still returns the correct result (served from ES).
-  - The log shows an ERROR similar to
-    `OS read failed in Phase 2 — falling back to ES. OS index may be stale or unavailable. Cause: …`
+  - The search still returns the correct result (served from ES), for every content type that has
+    live content. Not zero results, and not a `500`.
+  - The log shows an ERROR naming the operation and the root cause, similar to
+    `OS read failed in Phase 2 [indexCount] — falling back to ES. OS index may be stale or
+    unavailable. Cause: … / root cause: ConnectException: Connection refused`
 - **Type:** Manual
+
+> ⚠️ **This case gave a false PASS before #37413 and can do so again.** Repeating an *identical*
+> search returns the pre-outage result from the query cache, which is indistinguishable from a
+> working fallback. Vary the query — a different `offset` on each call is enough — or you are
+> testing the cache, not the fallback. A count is cached without `offset` in its key, so also
+> exercise a content type never queried since startup.
+>
+> A second false-pass route: if a *count* is what fails, some paths propagate the error while the
+> *search* path used to convert it to an empty result on the spot. Check the returned total, not
+> just the absence of an error.
 
 ## TC-040 — Phase 3 does NOT auto-rollback (negative case)
 
--- a/dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ESContentFactoryImpl.java
+++ b/dotCMS/src/main/java/com/dotcms/content/elasticsearch/business/ESContentFactoryImpl.java
@@ -1,8 +1,6 @@
 package com.dotcms.content.elasticsearch.business;
 
 import static com.dotcms.content.elasticsearch.business.ESContentletAPIImpl.MAX_LIMIT;
-import static com.dotcms.content.index.IndexConfigHelper.isMigrationComplete;
-import static com.dotcms.content.index.IndexConfigHelper.isReadEnabled;
 import static com.dotcms.variant.VariantAPI.DEFAULT_VARIANT;
 import static com.dotmarketing.portlets.contentlet.model.Contentlet.AUTO_ASSIGN_WORKFLOW;
 import static com.dotmarketing.portlets.contentlet.model.Contentlet.TITLE_IMAGE_KEY;
@@ -17,6 +15,7 @@ import com.dotcms.business.WrapInTransaction;
 import com.dotcms.content.business.json.ContentletJsonAPI;
 import com.dotcms.content.business.json.ContentletJsonHelper;
 import com.dotcms.content.index.ContentFactoryIndexOperations;
+import com.dotcms.content.index.PhaseRouter;
 import com.dotcms.content.index.IndexContentletScroll;
 import com.dotcms.content.index.domain.SearchHit;
 import com.dotcms.content.index.domain.SearchHits;
@@ -232,8 +231,19 @@ public class ESContentFactoryImpl implements ContentletFactory {
 
     private final ContentletCache contentletCache;
 	private final LanguageAPI languageAPI;
-    private final ContentFactoryIndexOperations indexOperationsES;
-    private final ContentFactoryIndexOperations indexOperationsOS;
+    /**
+     * Phase-aware router for every index read this factory performs.
+     *
+     * <p>These read call sites are the funnel for essentially all content search in the product:
+     * {@code /api/content/_search}, {@code ContentletAPI} search and count, Velocity
+     * {@code $dotcontent.pull}, URL maps, Site Search, scroll consumers and the admin content
+     * browser. They used to pick a provider with a bare ternary and call it directly, which left
+     * the router's Phase 2 fallback to Elasticsearch unreachable from the busiest read path in
+     * the product — with OpenSearch down, a search returned a well-formed empty result rather
+     * than the data Elasticsearch was holding the whole time (issue #37413). The router is now
+     * the only way a provider is selected here; do not reintroduce a second mechanism.</p>
+     */
+    private final PhaseRouter<ContentFactoryIndexOperations> indexRouter;
 
     private static final ObjectMapper mapper = DotObjectMapperProvider.getInstance()
             .getDefaultObjectMapper();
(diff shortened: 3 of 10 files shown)


ihoffmann-dot:
## ✅ Approving

Solid analysis, matches what the diff actually does.

- Root cause is correctly identified and fixed at both layers.
- **AC-006** and **AC-007** aren't satisfied literally as written, but both deviations are argued, not glossed over. 
- Test coverage is strong: 14 unit tests against the router contract, 13 integration tests proving the actual wiring, both registered in `MainSuite1b`.
- All scope decisions, residual risks and out-of-scope items are documented in the PR and in `OPENSEARCH_MIGRATION.md`, nothing here is a silent assumption.

ihoffmann-dot (approved):

dario-daza (approved):

ihoffmann-dot (dismissed):

dario-daza (dismissed):

dario-daza (approved):

ihoffmann-dot (approved):

Issue #37307, opened 2026-08-31
Experiments listing: user and time filters, Created By column, and Content Drive-aligned pagination


erickgonzalez:
## Description

Feedback from the Global Sprint demo on the new Experiments portlet. Three additions to the listing screen:

1. **User filter** — filter experiments by who created them.
2. **Schedule filter** — filter experiments by scheduled start date.
3. **"Created By" column** — surface the creator in the listing table.
4. **Pagination aligned with Content Drive** — the listing paginates, but not the way the rest of the product does.

**Design prototype:** https://claude.ai/design/p/36aed5a6-ca42-4624-b187-a465a23ce909?via=share&file=Experiments.dc.html

**Slack threads:**
- https://dotcms.slack.com/archives/C09LGH4HNR5/p1788194416605279
- https://dotcms.slack.com/archives/C09BJJ28THR/p1788192537515369

---

### Current state

The listing already has a search input plus two chip filters (Status, Goal) built on `DotExperimentListFilterComponent`, which wraps the shared content-drive `DotChipFilterComponent` from `@dotcms/ui`. The new filters follow that same pattern — the component is deliberately generic (its own docblock notes it "knows nothing about statuses or goals"), so adding a filter is mostly a matter of supplying an option list.

All filtering happens **client-side** in `DotExperimentsListStore` as a chain of computed signals:

```
siteScopedExperiments → searchedExperiments → statusFilteredExperiments
  → goalFilteredExperiments → filteredExperiments → (paginated)
```

`GET /v1/experiments` accepts only `pageId`, `name`, and `status` and returns the full list, so **no backend filtering work is required** — the two new filters extend this client-side chain.

---

### 1. User filter

- Chip filter, multi-select, sitting alongside the existing Status and Goal chips.
- Options come from `GET /api/v1/users/filter`, the endpoint Modernization already uses. It supports everything needed: `filter` (search across id, first/last name, email, full name), `page`, `perPage`, `orderBy`, `direction`.
- **Infinite scroll** inside the popover — page in more users as the list is scrolled, rather than loading everything up front.
- **Search** inside the popover, passed through as the `filter` query param (server-side search, debounced).
- Lists **all users in the system**, not just those who have created an experiment.
- Matching is on user ID: the selected users' IDs are compared against each experiment's `createdBy` (which holds the user ID).
[...]

PR #37510, opened 2026-09-11
feat(experiments): expose the experiment creator's display name in the Experiments API (#37304)

### LINKED ISSUE #37304: Expose the experiment creator's username in the Experiments API
## Description

The new Experiments portlet needs to display **who created each experiment** in its listing. Today the API returns the creator only as an opaque user ID, so the UI has no name to render and would have to resolve every ID itself.

`AbstractExperiment` already serializes a `createdBy` property, but its value is the **user ID** — the same value that backs `getOwner()` for permission checks:

```java
// dotCMS/src/main/java/com/dotcms/experiments/model/AbstractExperiment.java
@JsonProperty("createdBy")
String createdBy();

@Value.Derived
@JsonIgnore
default String getOwner() {
    return createdBy();
}
```

Current payload:

```json
{
  "id": "0e8b8b1e-...",
  "name": "Homepage CTA test",
  "createdBy": "dotcms.org.1"
}
```

Desired payload:

```json
{
  "id": "0e8b8b1e-...",
  "name": "Homepage CTA test",
  "createdBy": "dotcms.org.1",
  "createdByUserName": "Admin User"
}
```

**Approach:** add a new `createdByUserName` field resolved from the `User` behind `createdBy`. `createdBy` keeps its current meaning and value — this is an **additive, non-breaking** change, and existing consumers (including permission logic) are unaffected.

Because the field is added at the Experiment serialization layer, it applies to every response that returns an Experiment, not just the list endpoint.

## Acceptance Criteria

**Happy path**

- [ ] `GET /v1/experiments` returns a `createdByUserName` string for each experiment in the list, containing the creator's full name (first + last name).
- [ ] `GET /v1/experiments/{id}` returns `createdByUserName` for the requested experiment.
- [ ] Every other endpoint whose response carries an Experiment also includes `createdByUserName` — create, update, archive, delete, `_start`, `_end`, `scheduled/{id}/_cancel`, and the variant operations (`/variants`, `/variants/{name}`, `/variants/{name}/_promote`).
- [ ] `createdBy` keeps its existing value (the user ID) and its existing JSON key — no consumer of the current contract breaks.
- [ ] `getOwner()` / permission behavior is unchanged: it still resolves from `createdBy`, not from the new field.

**Sad path / edge cases**

- [ ] When the user ID in `createdBy` cannot be resolved to a `User` (deleted user, orphaned reference), `createdByUserName` falls back to the raw user ID so the field is never null or empty.
- [ ] When the experiment was created by the system user, `createdByUserName` returns a value (resolved name, or the ID per the fallback rule) rather than failing the request.
- [ ] A failure to look up the user never fails the experiment request — the endpoint still returns 200 with the experiment payload and the fallback value.
- [ ] Listing N experiments does not perform N uncached user lookups that measurably degrade list response time (resolve per distinct user ID, or rely on the user cache).

**Contract / docs**

- [ ] `createdByUserName` is documented in the Swagger `@Schema` for the experiment view, and the annotation matches the actual return type.
- [ ] `openapi.yaml` is regenerated (`./mvnw compile -pl :dotcms-core -DskipTests`) and committed alongside the Java changes.

**Tests**

- [ ] Integration test: an experiment created by a known user returns that user's full name in `createdByUserName` from both the list and the single-GET endpoints.
- [ ] Integration test: an experiment whose `createdBy` points at a non-existent user returns the raw ID as `createdByUserName` and still responds 200.
- [ ] Existing experiments tests still pass — in particular any asserting on the shape/value of `createdBy`.

## Priority

Medium

## Additional Context

**Decisions made during refinement:**

| Question | Decision |
|---|---|
| Field shape | New sibling field `createdByUserName`; `createdBy` unchanged (non-breaking) |
| Scope | Every response that returns an Experiment, not just the list endpoint |
| `lastModifiedBy` | **Out of scope** — creator only. Can be added later using the same mechanism if the portlet needs "last edited by" |
| Unresolvable user | Fall back to the raw user ID so the portlet column is never blank |

**Relevant files:**

- `dotCMS/src/main/java/com/dotcms/experiments/model/AbstractExperiment.java` — where `createdBy` / `lastModifiedBy` are declared
- `dotCMS/src/main/java/com/dotcms/rest/api/v1/experiments/ExperimentsResource.java` — all experiment endpoints
- `dotCMS/src/main/java/com/dotcms/rest/api/v1/experiments/ResponseEntityExperimentView.java` and `ResponseEntitySingleExperimentView.java` — response wrappers

**Consumer:** the new Experiments portlet listing

**Note:** `lastModifiedBy` has the same ID-not-name problem. It is deliberately excluded here — file a follow-up if the portlet later surfaces a "last modified by" column.

oidacra:
Adds `createdByUserName` beside the existing `createdBy`, so every response that carries an Experiment reports who created it by name instead of only by an opaque user id. The consumer is #37307 (the Experiments portlet listing's Created By column), which is blocked on this field.

CleanShot 2026-09-16 at 4 44 25 PM@2x

Closes #37304. Visual review page (summarizes; the spec decides): https://claude.ai/code/artifact/fdc2cd0a-bf41-4a1f-9b77-3c3ae9e47e85

### What changes

The field is declared **once** on the model, so all **14** endpoints that return an Experiment carry it without being edited — `ExperimentsResource` and both response wrappers are not in the change set. `createdBy` keeps its key, its value and its role as the permission owner behind `getOwner()`, and `lastModifiedBy` is deliberately untouched.

**The value is never null, absent or empty:**

| Creator | `createdByUserName` |
|---|---|
| Resolves, has a name | the full name |
| Is the system user | `System`, short-circuited before any lookup |
| Deleted, orphaned, blank-named, or the lookup fails | `unknown` |

Those two labels are not invented here — they are what `BrowserAPIImpl.ownerName` already publishes for the Content Drive folder view, which answers this exact question for a different listing. Two listings in the same product labelling the same orphaned owner differently is a worse outcome than either label on its own. (Content Drive reaches the user through `UserLocalManagerUtil` and so bypasses `UserCache`; this resolver deliberately goes through `APILocator.getUserAPI()`, because the cost argument below rests on that cache.)

### Why a plain `default` method and not an Immutables attribute

This is the decision the whole change rests on, and it was settled by reading the **generated** `Experiment` class rather than by argument:

- **`@Value.Derived` is assigned in the constructor.** It would resolve a user on every Experiment ever built — the database transformer behind every list row, the `withTargetingConditions` rebuild inside `find()`, the running-experiments cache fill on the **page-render** path, and the push-publish dependency walk. None of those serialize the object.
- **`@Value.Lazy` memoizes per instance.** The running-experiments list cache holds long-lived shared entries, so a memoized name would outlive a rename.

A plain `default` method is not an Immutables attribute at all: no field, no builder entry, no `equals`/`hashCode`/`toString` participation, no `build()` cost. It is evaluated only when something serializes the Experiment.

`@JsonProperty(access = READ_ONLY)` is load-bearing rather than decorative. The generated `Experiment.Json` delegate binds settable attributes only, carries no `@JsonIgnoreProperties(ignoreUnknown = true)`, and the REST mapper leaves `FAIL_ON_UNKNOWN_PROPERTIES` at Jackson's default — so a serialize-only field would otherwise make a round-tripped payload unreadable. A test proves it holds when Jackson binds into the delegate; that result is what retired the `@JsonAppend` fallback the plan had pre-approved.

### Tests

13 unit tests, written and confirmed failing before any implementation existed.

Two of them guard the invariants above, and their value was **verified by mutation, not asserted**: temporarily annotating the accessor with `@Value.Derived` fails exactly those two and leaves the other eleven green. Eleven tests that all serialize cannot see that regression.

- `buildingAnExperimentWithoutSerializing_neverResolvesTheCreator` — builds, rebuilds, reads the owner, compares and stringifies an Experiment with zero interaction on the user layer, then asserts serialization is what triggers the single lookup.
- `renamingTheCreator_isReflectedOnTheNextSerialization` — serializes the same instance twice across a rename and expects the new name.

Integration coverage in `ExperimentsResourceIntegrationTest` (already registered in `MainSuite1a`) and the Postman contract assertion are still to come.

### Corrections to the issue body

The issue's endpoint list is close but not exact; the spec carries a 14-endpoint table that supersedes it. None of this changes the approach.

- `DELETE /v1/experiments/{experimentId}` is listed among the endpoints carrying an Experiment; it returns the confirmation string `"Experiment deleted"`.
- Two endpoints that *do* return an Experiment are missing: `DELETE /v1/experiments/{experimentId}/goals/primary` and `DELETE /v1/experiments/{experimentId}/targetingConditions/{id}`.
- The update endpoint is `PATCH`, not `PUT`.
- The issue specifies the **raw user id** as the fallback. Superseded by the Content Drive alignment above.

### Checklist

- [x] Tests — 13 unit tests; integration and Postman still pending
- [x] Translations — not applicable, backend only
- [x] Security Implications Contemplated — `createdBy` and `getOwner()` are untouched, so permission behaviour is unchanged. The field exposes a display name only to callers already authorized to read the experiment, and nothing thrown by the lookup escapes into the response.

This PR fixes: #37304

### DIFF
--- a/specs/37304-experiment-created-by-username/spec.md
+++ b/specs/37304-experiment-created-by-username/spec.md
@@ -1,0 +1,349 @@
+# Feature Specification: Creator Name on the Experiments API
+
+**Feature Branch**: `issue-37304-experiment-created-by-username`
+
+**Created**: 2026-09-10
+
+**Status**: Draft
+
+**Type**: Task (additive, non-breaking API contract change)
+
+**Epic**: [#36763 — Experiments: A/B Testing v2](https://github.com/dotCMS/core/issues/36763)
+
+**Work item**: [dotCMS/core#37304 — Expose the experiment creator's username in the Experiments API](https://github.com/dotCMS/core/issues/37304)
+
+**Input**: User description: "Expose the experiment creator's username in the Experiments API" — taken from issue #37304.
+
+---
+
+## Scope Note *(read this first)*
+
+Every Experiment already records who created it, but only as an opaque user ID. The payload says
+`"createdBy": "dotcms.org.1"` and nothing else. That ID is not a name: it is the same value the
+permission layer uses as the experiment's owner, and it means nothing to the person reading a
+listing.
+
+The new Experiments portlet wants a **Created By** column. With today's contract the portlet has two
+bad options: render the raw ID, or issue a second round of requests to translate every ID it sees
+into a name. Both push work onto the client that the server can do once, cheaply, from data it has
+already loaded.
+
+This work adds one field beside the existing one: `createdByUserName`, carrying the creator's full
+name. It is **additive**. `createdBy` keeps its key, keeps its value and keeps its meaning, and the
+owner/permission behaviour that reads it is untouched. Nothing that consumes the current payload has
+to change.
+
+Because the field belongs to the Experiment itself and not to one endpoint's response shape, it
+arrives on *every* response that carries an Experiment — the listing, the single fetch, and each of
+the lifecycle and variant operations that return the updated experiment.
+
+Two things this work deliberately does **not** do. It does not touch `lastModifiedBy`, which has the
+exact same ID-not-name problem and is left for a follow-up. And it does not store the name: the
+value is resolved when the experiment is read, so an experiment created by someone who later changes
+their name does not keep serving the old one.
+
+---
+
+## User Scenarios & Testing *(mandatory)*
+
+### User Story 1 - See who created each experiment in the listing (Priority: P1)
+
+A marketer opens the Experiments portlet and scans the list of experiments for the site. Each row
+tells them who created that experiment by name — "Admin User", not "dotcms.org.1" — so they can find
+their own work, or find the colleague to ask about someone else's.
+
+**Why this priority**: This is the whole point of the change and the reason the consumer issue
+(#37307, listing Created By column) is blocked. The listing is where an ID is least useful, because
+the reader is comparing many rows at once.
+
+**Independent Test**: Create experiments owned by two different known users, request the experiment
+list, and confirm each entry reports the corresponding user's full name.
+
+**Acceptance Scenarios**:
+
+1. **Given** an experiment created by a user whose full name is "Admin User", **When** the experiment
+   list is requested, **Then** that experiment's entry reports `createdBy` unchanged as the user ID
+   **and** reports the creator's name as "Admin User".
+2. **Given** a list containing experiments created by several different users, **When** the list is
+   requested, **Then** each entry reports the name of its own creator.
+3. **Given** a list containing many experiments created by the *same* user, **When** the list is
+   requested, **Then** every entry reports that user's name and the response is served without a
+   perceptible slowdown compared with the same list before this change.
+
+---
+
+### User Story 2 - The name travels with the experiment everywhere (Priority: P2)
+
+A client that has just created, started, ended, archived or otherwise modified an experiment gets the
+updated experiment back and can render the creator's name from that response alone, without a
+follow-up fetch and without special-casing which operation it just performed.
+
+**Why this priority**: Without it, a portlet that renders a row from a create/update response and a
+row from the listing has to handle two different shapes for the same object. It is also what makes
+the change cheap: one place to resolve the name, not fourteen.
+
+**Independent Test**: Exercise each endpoint whose response carries an Experiment and confirm the
+creator's name is present in each one.
+
+**Acceptance Scenarios**:
+
+1. **Given** an existing experiment, **When** it is fetched by ID, **Then** the response carries the
+   creator's name.
+2. **Given** an existing draft experiment, **When** it is started, ended, cancelled, archived,
+   updated, or has a variant added, removed, renamed or promoted, **Then** each of those responses
+   carries the creator's name for that experiment.
+3. **Given** an experiment created through the create endpoint, **When** the create response is read,
+   **Then** it already carries the creator's name — no second request is needed.
+
+---
+
+### User Story 3 - The column is never blank (Priority: P3)
+
+An administrator looks at a listing that includes experiments created by a user who has since been
+deleted, and by the system itself. Every row still shows something identifying in the Created By
+column, and no row is missing or empty. The listing loads normally.
+
+**Why this priority**: A rare data condition must not produce a broken-looking column or, worse, a
+failed request that hides every other experiment in the list.
+
+**Independent Test**: Point an experiment's `createdBy` at a user ID that no longer resolves, request
+both the list and the single fetch, and confirm the response succeeds and reports the raw ID.
+
+**Acceptance Scenarios**:
+
+1. **Given** an experiment whose creator ID does not resolve to any user, **When** the experiment is
+   requested, **Then** the request succeeds and the creator name field carries the raw creator ID.
+2. **Given** an experiment created by the system user, **When** the experiment is requested, **Then**
+   the request succeeds and the creator name field carries a non-empty value.
+3. **Given** a list where one experiment has an unresolvable creator and the others do not, **When**
+   the list is requested, **Then** the request succeeds, the resolvable entries report real names and
+   only the affected entry falls back to the raw ID.
+
+---
+
+### Edge Cases
+
+- **Creator no longer exists** (deleted user, orphaned reference): the field falls back to the raw
+  creator ID. It is never null, never absent and never an empty string.
+- **Creator resolves but has no usable name** (first, middle and last name all blank): treated the
+  same as unresolvable — fall back to the raw creator ID, so the column still identifies something.
+- **Creator is the system user**: resolves normally; if the system user has no usable name, the
+  fallback rule applies. The request never fails because of it.
+- **User lookup fails for an infrastructure reason** (database error, cache error): the experiment is
+  still returned successfully with the fallback value. A failure to decorate must never turn a
+  successful experiment read into a failed request.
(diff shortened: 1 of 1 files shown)


dotCMS/src/test/java/com/dotcms/experiments/model/ExperimentCreatorNameResolverTest.java
```diff
@@ -0,0 +1,190 @@
+package com.dotcms.experiments.model;
```
freddyDOTCMS: You must register this class in the MainSuite

dotCMS/src/test/java/com/dotcms/experiments/model/ExperimentCreatorNameTest.java
```diff
@@ -0,0 +1,338 @@
+package com.dotcms.experiments.model;
+
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
```
freddyDOTCMS: You must register this class in the MainSuite

dotCMS/src/main/java/com/dotcms/experiments/model/AbstractExperiment.java
```diff
@@ -90,6 +91,41 @@ default Instant modDate() {
...
+     * deserialize as an unknown property.
+     */
+    @JsonProperty(value = "createdByUserName", access = JsonProperty.Access.READ_ONLY)
+    @Schema(description = "Display name of the user who created the experiment. Reports \"System\" "
+            + "for the system user and \"unknown\" when the user cannot be resolved or has no name "
+            + "set, so the value is never empty.",
+            example = "Admin User")
+    default String createdByUserName() {
+        return ExperimentCreatorNameResolver.INSTANCE.resolve(createdBy());
+    }
```
freddyDOTCMS: This class is more like a Data model, so it should be a POJO, I think this set should go out of this class, maybe we should move this logic to the Rest Layer, so we affect just the responde of the endpoint

freddyDOTCMS: Also we are going to hit the UserAPI anytime this method is called

dotCMS/src/main/java/com/dotcms/publishing/BundlerUtil.java
```diff
@@ -326,10 +328,31 @@ private static ObjectMapper getObjectMapper() {
...
+     *
+     * <p>The field is a display value derived from {@code createdBy}, which IS bundled, so nothing
+     * is lost: the receiver resolves the name itself when it serves the experiment over REST.
+     * Ignoring it here also makes a current receiver tolerant of a bundle that happens to carry it.
+     */
+    private abstract static class ExperimentBundleMixIn {
+
+        @JsonIgnore
+        abstract String createdByUserName();
+    }
```
freddyDOTCMS: If we move the logic to the Rest layer we don't need this change here

freddyDOTCMS (dismissed):

dotCMS/src/main/java/com/dotcms/experiments/model/AbstractExperiment.java
oidacra: Done in aec284cc13 — you were right, and the second point turned out to be the load-bearing one.

Because the accessor was on the model, every serialization of an Experiment resolved a user, and serialization here is not only REST. `ExperimentBundler` and `ExperimentHandler` serialize the whole Experiment for push publish, and `ExportStarterUtil` does the same for starter exports. Bundles are read back through a mapper that rejects unknown properties, and `BundlePublisher` runs every handler in one transaction — so a receiver on a build without the field would have rolled back the entire bundle, not just the experiment.

The derivation now lives in `ExperimentView` and `AbstractExperiment` is a plain data object again. The view uses `@JsonUnwrapped`, so the wire shape does not change: the Experiment's own fields stay top level with `createdByUserName` beside them. The wrapping happens inside `ResponseEntitySingleExperimentView` / `ResponseEntityExperimentView`, so none of the 14 endpoints changed — `ExperimentsResource` is not in the diff. In `openapi.yaml` the `Experiment` schema is back to its original shape and a new `ExperimentView` schema carries the flattened fields plus the new one.

dotCMS/src/main/java/com/dotcms/publishing/BundlerUtil.java
oidacra: Correct, and reverted in aec284cc13. The mix-in existed only to keep the derived field out of bundles; with the derivation in the REST view there is nothing to exclude, so `BundlerUtil` is untouched by this PR again.

dotCMS/src/test/java/com/dotcms/experiments/model/ExperimentCreatorNameTest.java
oidacra: This file no longer exists — it was removed in aec284cc13, as part of moving the derivation to the REST layer that you asked for in the `AbstractExperiment` thread.

Its coverage moved to `ExperimentViewTest` (`dotCMS/src/test/java/com/dotcms/rest/api/v1/experiments/`), which exercises the same behaviour where it now happens: the view serialising the name beside the Experiment's own fields, the fallback rules, and a bare Experiment carrying no derived field and costing no user lookup. The cases that only made sense while the field lived on the model went with it.

dotCMS/src/test/java/com/dotcms/experiments/model/ExperimentCreatorNameResolverTest.java
oidacra: This is an Unit Test not an integration.

freddyDOTCMS (approved):

PR #37514, opened 2026-09-11
docs(experiments): spec for the listing Created By and schedule filters, Created By column and pagination alignment (#37307)

### LINKED ISSUE #37307: Experiments listing: user and time filters, Created By column, and Content Drive-aligned pagination

oidacra:
This PR carries the spec for #37307. No implementation yet.

## What the work delivers

Four additions to the Experiments listing screen, collected from Global Sprint demo feedback. Each is independently shippable.

1. **Created By filter** — a chip that narrows the list to experiments created by one or more chosen people. Options come from the whole user directory rather than only from creators already present in the list, paged in as the option list is scrolled, searched server-side, and shown without counts. Matching is on the creator's user id, which the payload already carries.
2. **Schedule time filter** — a single-select chip over the scheduled start date: any schedule, or the last 1, 3, 6 or 12 months. The comparison is a lower bound only, so experiments scheduled to start in the future match every window, and experiments with no schedule are excluded unless "Any schedule" is chosen.
3. **Created By column** — the creator's name in the table. Blocked by #37304.
4. **Pagination alignment** — closes the gaps that remain between this listing's paginator and Content Drive's.

Filtering stays client-side over the already-loaded set, and both new filters live in the URL alongside the existing ones.

## Corrections to the issue's description of the code

The issue describes the current code in seven places that no longer match the branch. Each one changes what the work is, so they are recorded in the spec's "State of the code as found" section and every requirement is written against the code rather than against the issue.

- The default page size is already 20 and the offered sizes are already 20/40/60. Two of the five listed pagination gaps are already closed.
- `lazyLoadOnInit` is already effectively set: the table is unconditionally lazy and the framework default for that flag is on. The defect the issue describes does not exist.
- Content Drive's paginator-visibility rule cannot be copied literally. It shows the paginator whenever the table is lazy, and this table is lazy, so adopting the expression verbatim gives the opposite of the issue's own acceptance criterion. The spec adopts the intent instead, which is available here because this listing knows its true total.
- `createdBy` is serialized by the backend but is not declared on the client's experiment type, so the filter needs a model change and not only store work.
- A shared lazy multiselect with server-side search and infinite scroll already exists and is publicly exported, so the popover is a reuse rather than the new sibling component the issue anticipated.
- The users endpoint's page parameter is 1-based, so "resets to page 0" is wording only.

One further finding, from the design review rather than the issue: the single-user lookup endpoint requires administrator rights, or access to both the Roles and Users areas, and refuses everyone else. Resolving a selected creator's name through it would have produced a filter that labels correctly when a person is picked and breaks on reload, for non-administrators only. The spec routes that resolution through the directory search instead, which any back-end user can reach, and records the constraint because the single-user lookup is the call an implementer reaches for first.

## Design prototype

Six places where the prototype and the issue disagree are recorded as assumptions with the issue's refinement decisions winning: per-option counts, the chip's count badge, a selectable "Any schedule" row, the window label wording, an initials avatar in the cell, and the "Experiment" column header with no Modified column.

## Dependency

The Created By column requires #37304, which adds the creator's display name to the experiment payload. The two filters are not blocked by it and can proceed in parallel.

## Contents

64 functional requirements, 16 success criteria, 4 user stories, 30 acceptance scenarios, 10 edge cases and 18 assumptions. No open clarifications remain.

Closes #37307. Visual review page (summarizes; the spec decides): https://claude.ai/code/artifact/2d8d855f-39cd-4ea1-b1d1-046f3433feab

### DIFF
--- a/specs/37307-experiments-list-filters-pagination/spec.md
+++ b/specs/37307-experiments-list-filters-pagination/spec.md
@@ -125,6 +125,12 @@ observing which rows survive, with the other three parts absent.
    **Then** no further requests are made.
 10. **Given** an experiment whose creator id resolves to no user, **When** any creator is ticked,
     **Then** that experiment is simply not matched, and nothing on the screen breaks.
+11. **Given** a link that already carries a creator selection, **When** it is opened cold by an
+    editor who is not an administrator, **Then** the chip shows a brief loading state and settles
+    on the selected people's names, and the table is narrowed to their experiments throughout.
+12. **Given** the same link, **When** the directory cannot resolve one of the selected ids,
+    **Then** that entry falls back to showing its id, the other names still resolve, and the table
+    stays narrowed to exactly the same experiments.
 
 ---
 
@@ -271,13 +277,34 @@ fewer rows than one page.
   are loaded and as the search term changes — including when the selected person is not present in
   the page currently displayed.
 - **FR-009a**: When a creator selection arrives from the address rather than from a click — a
-  reload, or a shared link — the selected ids MUST be resolved to their display names so the chip
-  labels people rather than opaque ids. The resolution MUST be a lookup of the selected ids, not a
-  scan of the loaded experiments: a person can be selected while having no experiment in view,
-  which is precisely the case that produces the empty state, and that is the case an
-  experiment-derived label would fail to name. A failed resolution MUST leave the filter working —
-  the filter matches on ids and does not depend on the labels — and MUST fall back to showing the
-  id rather than blanking the chip.
+  reload, or a shared link — the selected ids MUST be resolved to their display names, so the chip
+  labels people rather than opaque ids. Selections made by clicking already carry their label from
+  the option list and MUST NOT be re-resolved.
+- **FR-009b**: The resolution MUST go through the same directory search the option list uses,
+  querying one selected id at a time. It MUST NOT use the single-user lookup endpoint: that one
+  requires administrator rights, or access to both the Roles and Users areas, and refuses everyone
+  else — which would produce a filter that works when a person is picked and breaks when the page
+  is reloaded, for non-administrators only. The directory search is available to any back-end user,
+  which is the same population that can reach this screen. This constraint MUST be stated in the
+  implementation, because the single-user lookup is the call an implementer reaches for first.
+- **FR-009c**: The directory search matches an id as a **substring**, across ids, names and email
+  addresses, so a query for one id can return several people — an id that is a prefix of a longer
+  one, most obviously. The resolution MUST select the result whose id matches exactly, and MUST NOT
+  take the first result.
+- **FR-009d**: Resolution MUST NOT read the names off the loaded experiments, even though every
+  experiment will carry its creator's name once #37304 lands. One path is specified deliberately in
+  place of a fast path plus a fallback: it behaves identically before and after #37304 — which
+  matters because the filter is not blocked by #37304 and can ship first — and it removes a
+  fallback branch that would almost never run and therefore would almost never be exercised. The
+  accepted cost is one request per selected person on each hydration that carries a selection, even
+  where the name was already in memory.
+- **FR-009e**: While resolution is in flight the chip MUST show an explicit loading state, rather
+  than rendering ids that are about to be replaced. No chip in this toolbar has such a state today,
+  so this extends the shared chip — the same kind of change FR-013 and FR-014 already require of
+  shared components, and it MUST NOT alter how existing consumers of that chip render.
+- **FR-009f**: A failed resolution MUST leave the filter working — it matches on ids and does not
+  depend on labels — and MUST fall back to showing the id rather than blanking the chip or dropping
+  the selection.
 - **FR-010**: The filter's options MUST NOT show result counts. This is a deliberate divergence
   from the Status and Goal chips: a count cannot be computed for a person whose page of the
   directory has not been fetched, and a count shown for some options and not others is worse than
@@ -305,9 +332,14 @@ fewer rows than one page.
 - **FR-016**: The option list MUST be created only once its popover is actually visible, so that a
   virtualised list measures a real viewport rather than a collapsed one and renders its first page.
 - **FR-016a**: The chip's collapsed label MUST follow the behaviour the shared chip already gives
-  the Status and Goal chips — the unfiltered label while nothing is selected, the selected names
-  while there are few, and an overflow summary beyond that. A count badge MUST NOT be introduced
-  for this chip alone.
+  the Status and Goal chips, with no exception made for this one — the selected names while there
+  are few, and an overflow summary beyond that. A count badge MUST NOT be introduced for this chip
+  alone. This is consistency chosen over legibility with eyes open: two long personal names will
+  not fit the chip and will truncate, where two status names do. Three chips in one toolbar
+  behaving alike is worth more than the second name being readable in that case.
+- **FR-016b**: While nothing is selected the chip MUST read "All", reusing the label the Status and
+  Goal chips already use rather than introducing a chip-specific one. The design prototype reads
+  "All Users"; the shared label is preferred so the three chips agree.
 
 ### Schedule time filter
 
@@ -492,6 +524,10 @@ fewer rows than one page.
   template or component.
 - **SC-015**: Once #37304 has shipped, every row shows a creator name; against a backend without
   #37304, every row shows the placeholder and nothing renders as "undefined".
+- **SC-016**: Opening a shared link that already carries a creator selection shows people's names
+  in the chip, not ids — and does so for an editor who is **not** an administrator and has no
+  access to the Roles or Users areas, which is the population most likely to be handed such a link.
+  The same holds against a backend that does not yet have #37304.
 
 ---
 
@@ -563,11 +599,16 @@ fewer rows than one page.
    that pins them, so a later edit cannot silently reintroduce the divergence.
 10. **This is a frontend-only change.** No Java, no SQL, no endpoint, no `openapi.yaml`. The one
     API-shaped dependency, the creator name, belongs to #37304.
-11. **Selected creators are resolved to names by looking them up, not by reading the loaded
-    experiments** (FR-009a). Chosen over the free alternative because the experiment-derived label
-    only works for creators who have an experiment in view and only after #37304 — and the case it
-    fails is exactly the one that produces the empty state, where the chip is the only thing on
-    screen still naming what was filtered.
+11. **Selected creators are resolved through the directory on every hydration, never from the
+    loaded experiments** (FR-009b, FR-009d). The free alternative was available and was weighed:
+    the listing loads every experiment in one request and keeps the unfiltered set in memory, so
+    once #37304 lands each one carries its creator's id and name together, and a label could be
+    read from that set at no cost for any creator who has an experiment loaded. It was rejected for
+    two reasons that outweigh the saving — it would behave differently before and after #37304,
+    during a window when the filter can ship and #37304 has not; and covering the creators it
+    cannot name would need a fallback branch that runs so rarely it would not be meaningfully
+    tested. One path that always works was preferred over a fast path plus a rarely-exercised one.
+    The cost is a request per selected person on each hydration that carries a selection.
 12. **The shared user search is created in the shared data-access library and the two existing
     portlet-local copies are left alone** (FR-015, FR-015a). Chosen over migrating them, to keep a
     frontend Experiments change from touching two unrelated portlets and their test suites. The
--- a/specs/37307-experiments-list-filters-pagination/spec.md
+++ b/specs/37307-experiments-list-filters-pagination/spec.md
@@ -218,8 +218,8 @@ fewer rows than one page.
 ### Edge Cases
 
 - **A creator is selected, then the page is reloaded.** The selection lives in the address as a
-  user id. The chip has to label it, and a name for it may not be available — see
-  [NEEDS CLARIFICATION 1].
+  user id, and nothing has fetched the directory yet, so the chip holds an id and no name. The
+  selected ids are resolved to names on hydration — see FR-009a.
 - **A creator is selected and the site is switched.** A site switch already restarts paging and
   drops the page narrowing. Whether it should also drop a creator selection: it should not — the
   same person can have experiments on both sites, and status, goal and search all survive a switch
@@ -270,6 +270,14 @@ fewer rows than one page.
 - **FR-009**: People already selected MUST stay selected, and MUST stay labelled, as further pages
   are loaded and as the search term changes — including when the selected person is not present in
   the page currently displayed.
+- **FR-009a**: When a creator selection arrives from the address rather than from a click — a
+  reload, or a shared link — the selected ids MUST be resolved to their display names so the chip
+  labels people rather than opaque ids. The resolution MUST be a lookup of the selected ids, not a
+  scan of the loaded experiments: a person can be selected while having no experiment in view,
+  which is precisely the case that produces the empty state, and that is the case an
+  experiment-derived label would fail to name. A failed resolution MUST leave the filter working —
+  the filter matches on ids and does not depend on the labels — and MUST fall back to showing the
+  id rather than blanking the chip.
 - **FR-010**: The filter's options MUST NOT show result counts. This is a deliberate divergence
   from the Status and Goal chips: a count cannot be computed for a person whose page of the
   directory has not been fetched, and a count shown for some options and not others is worse than
@@ -286,11 +294,20 @@ fewer rows than one page.
 - **FR-014**: The option list's own copy — its "nothing matched" and "failed to load" text — MUST
   read as being about people. The shared component's current copy is worded for content-type
   fields and is wrong here, so it MUST be parameterised or replaced rather than inherited.
-- **FR-015**: The capability that searches the user directory MUST exist in exactly one place in
-  the client. An equivalent already exists but is not reachable from this portlet; the
-  implementation MUST make one reachable rather than add a second copy.
+- **FR-015**: The capability that searches the user directory MUST live in the shared data-access
+  library, and the Experiments portlet MUST consume it from there. It MUST NOT be reached by
+  importing another portlet's library: two portlets already hold an equivalent, and neither
+  publishes it, which is a boundary rather than an oversight.
+- **FR-015a**: Consolidating the two existing portlet-local user searches onto that shared
+  capability is explicitly OUT OF SCOPE. Both keep working untouched. This knowingly leaves three
+  implementations of the same directory call in the tree until a follow-up consolidates them; the
+  follow-up MUST be filed rather than left implied.
 - **FR-016**: The option list MUST be created only once its popover is actually visible, so that a
   virtualised list measures a real viewport rather than a collapsed one and renders its first page.
+- **FR-016a**: The chip's collapsed label MUST follow the behaviour the shared chip already gives
+  the Status and Goal chips — the unfiltered label while nothing is selected, the selected names
+  while there are few, and an overflow summary beyond that. A count badge MUST NOT be introduced
+  for this chip alone.
 
 ### Schedule time filter
 
@@ -326,18 +343,32 @@ fewer rows than one page.
 
 - **FR-028**: The table MUST have a Created By column showing the creator's name as delivered by
   the experiment payload.
-- **FR-029**: The column MUST render the creator's display name, not the creator's id. The client's
-  experiment type MUST be extended to declare that field when #37304 lands.
+- **FR-029**: The column MUST bind to the creator's display-name field and render whatever it
+  carries — including the raw user id that #37304 deliberately falls back to for an unresolvable
+  creator, which is how that field is guaranteed never to be null or empty. "Not the id" forbids
+  binding the column to the creator-id field instead; it does not ask the screen to detect and
+  suppress #37304's fallback. The client's experiment type MUST be extended to declare the field
+  when #37304 lands.
 - **FR-030**: A name too long for its column MUST be truncated with the full value available on
   hover, consistent with the Name and Page columns.
 - **FR-031**: When the payload carries no creator name — a backend older than #37304 — the cell
   MUST show the listing's existing placeholder rather than an empty cell or a literal "undefined".
-- **FR-032**: The column's position MUST follow the design prototype linked from the issue. See
-  [NEEDS CLARIFICATION 2].
+- **FR-032**: The column MUST sit between Variants and Schedule, per the design prototype, giving
+  the header order: Name, Page, Goal, Variants, Created By, Schedule, Status, Modified, actions —
+  nine columns. It is an **addition**: the Modified column stays, even though the prototype does
+  not draw one.
 - **FR-033**: The table's skeleton row MUST draw one placeholder cell per column after the column
-  is added.
+  is added — nine rather than eight.
 - **FR-034**: The table's documented minimum width MUST be recomputed to include the new column, so
-  the elastic Name column keeps the floor its source comment specifies.
+  the elastic Name column keeps the floor its source comment specifies. The comment carries the
+  arithmetic, not just the total, so the sum itself MUST be rewritten rather than the number
+  silently bumped. The new column's own width is an implementation choice, bounded by the two
+  columns it sits between.
+- **FR-034a**: The Created By column MUST be display-only. It MUST NOT be sortable, MUST NOT add a
+  sort field, a sort URL value or a comparator. Every column beside it is sortable, so this has to
+  be said: the screen's sort fields are a closed set whose values double as the table's sort key,
+  the address's sort parameter and the comparator key, so adding one would reach into the address
+  contract this feature otherwise promises to leave alone. Sorting by creator is not asked for.
 
 ### Pagination alignment
 
@@ -370,12 +401,24 @@ fewer rows than one page.
 
 ### Cross-cutting: address, clearing, and the existing narrowing chain
 
-- **FR-045**: Both new filters MUST be represented in the address, and MUST rehydrate from it on
-  reload and on browser back and forward, exactly as the search term, Status and Goal selections do.
+- **FR-045**: **Every filter of this list is part of the address.** The listing's view state —
+  each narrowing, the sort and the paging — lives in the route and nowhere else, so a filtered
+  list can be linked, reloaded and stepped through with browser back and forward without losing
+  what it was showing. This is a standing rule, stated here as a principle so that filters added
+  after these two inherit it rather than re-deciding it. Applied to this work: both new filters
+  MUST be represented in the address and MUST rehydrate from it on reload and on back and forward,
+  exactly as the search term, Status and Goal selections already do.
+- **FR-045a**: The rule covers filters **of the list**. It MUST NOT be over-applied to transient
+  interface state — in particular, the search box inside the Created By popover narrows the option
+  list rather than the data, and MUST NOT be written to the address. Routing it would also rewrite
+  the address on every settled keystroke for as long as the popover is open.
 - **FR-046**: Both MUST join the existing single-writer/single-reader address contract rather than
   writing to the address independently.
 - **FR-047**: A filter in its default state MUST be absent from the address, so a pristine listing
-  still carries no query string.
+  still carries no query string. This satisfies FR-045 rather than bending it: an absent parameter
+  means its default, so the view state stays fully derivable from the route. The stricter reading
+  — always write every parameter — is deliberately not adopted, because it would change all ten
+  existing parameters and the deep links #37005 built on them.
 - **FR-048**: Values in the address that are not recognised MUST be dropped rather than applied —
   an unknown window value, and a malformed creator entry — matching the rule the Status and Goal
   parameters already follow. An unrecognised creator id is not necessarily invalid, so it MUST be
@@ -472,7 +515,7 @@ fewer rows than one page.
     bookmarked under the old 10/25/50 options is not broken.
 - **Known related decisions**:
   - **#37304** — adds the creator's display name to every experiment response. Hard dependency for
-    User Story 3 and for FR-028 through FR-032. Currently open.
+    User Story 3 and for FR-028 through FR-034a. Currently open.
   - **#37007** — swaps the listing to server-side paging and filtering; parked on **#36823**.
     FR-052 states what it must preserve. Its existence is the reason this work does not attempt
     server-side filtering.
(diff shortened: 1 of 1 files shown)


core-web/libs/portlets/dot-experiments/portlet/src/lib/store/dot-experiments-filter-facade.ts
```diff
@@ -0,0 +1,171 @@
...
+        // ignored rather than dispatched blindly.
+        removeFilter: (key: string): void => {
+            if (readers[key]) {
+                patchFilters({ [key]: [] });
+            }
+        },
+
+        clearFilters: (): void => {
+            dispatcher.dispatch(dotExperimentsListPageEvents.filtersCleared());
+        },
```
nicobytes, line 140: **Important:** the bar's **Clear all** drops a *working* page narrowing in the portlet.

`clearFilters` dispatches `filtersCleared`, and that reducer nulls `selectedPageId` / `selectedPageUrl` / `languageId` (and the search term). Repro: open the portlet from UVE with `?pageId=<id>` (list correctly narrowed to that page) → pick any Status chip → click **Clear all** → the list now shows every experiment on the site and `pageId` leaves the URL.

That contradicts the doc on `$hasNonDefaultFilters` right below ("the page narrowing is not the user's to clear from here") and the rule in `dot-experiments-list-page.events.ts` that no control widens a page filter that is *working*. `filtersCleared` was designed for the empty-state button, where the narrowing matched nothing; the bar's button is visible in the normal state too. (The panel is not affected because its fetch is server-scoped by page.)

Suggested fix: give the bar its own event that resets only what the chips own, and keep `filtersCleared` for the dead-end button:

```ts
chipFiltersCleared: type<void>(),
// reducer
on(dotExperimentsListPageEvents.chipFiltersCleared, () => ({
    selectedStatuses: DEFAULT_EXPERIMENTS_LIST_STATUSES,
    selectedGoals: DEFAULT_EXPERIMENTS_LIST_GOALS,
    selectedCreators: DEFAULT_EXPERIMENTS_LIST_CREATORS,
    scheduleFrom: null,
    scheduleTo: null,
    page: DEFAULT_EXPERIMENTS_LIST_PAGE
})),
```

Also worth deciding explicitly whether Clear all should wipe the search term: `$hasNonDefaultFilters` ignores it (so search alone never shows the button), but clicking it currently clears it. Please add a store/component test for "Clear all with an active `pageId` keeps the page narrowing".

core-web/libs/ui/src/lib/components/dot-user-filter/dot-user-filter.component.ts
```diff
@@ -0,0 +1,205 @@
...
+
+            untracked(() => this.#namesById.update((names) => ({ ...names, ...resolved })));
+        });
+    }
+
+    /** Remembers the labels the list just showed, so a fresh pick never needs resolving. */
+    onSelectionChange(options: DotLazyMultiselectOption[]): void {
+        this.#namesById.update((names) => ({
+            ...names,
+            ...Object.fromEntries(options.map(({ value, label }) => [value, label]))
```
nicobytes: **Important:** confirming the finding from the automated review — still present at `ff3310c`.

`DotLazyMultiselectComponent.onChange` emits `label: this.#labelByValue.get(value) ?? value`, and that map starts empty every time the popover opens (the list lives inside `@if (popover.overlayVisible)`). So for a creator restored from the URL and resolved by `#resolved`, toggling a second person re-emits the first one with `label === id`, and this spread overwrites the good name. `#unresolvedIds` treats any truthy entry as resolved, so it never recovers until reload — the chip shows the raw id, which is exactly what the new `loading` input on `dot-chip-filter` exists to avoid.

Suggested fix — never let an id-as-label replace a known name:

```ts
onSelectionChange(options: DotLazyMultiselectOption[]): void {
    const labelled = options.filter(({ value, label }) => label !== value);
    this.#namesById.update((names) => ({
        ...names,
        ...Object.fromEntries(labelled.map(({ value, label }) => [value, label]))
    }));
    this.selectionChange.emit(options.map(({ value }) => value));
}
```

A spec that seeds `selected` with a resolved id, then emits a selection where that id comes back with `label === value`, would lock it in.

core-web/libs/portlets/dot-experiments/portlet/src/lib/util/dot-experiments-list-store.util.ts
```diff
@@ -87,6 +97,107 @@ export function parseViewState(reader: QueryParamReader): DotExperimentsListView
  * a missing leading slash, and a trailing one. Case is left alone; dotCMS paths are not
  * case-insensitive, and lowercasing here would claim a match the backend would not make.
  */
+/** Inclusive instant bounds of a schedule period. `Infinity` on a side the period leaves open. */
```
nicobytes, line 92: The new schedule helpers were inserted **between** `normalizePagePath`'s JSDoc (lines 92–99) and the function itself. The result: the page-path doc now sits orphaned on top of `ScheduleBounds` (two consecutive `/** */` blocks, the first one wrong for the symbol), and `normalizePagePath` further down has no doc at all. IDE hovers show the wrong text for both.

Move the `ScheduleBounds` … `normalizeScheduleBound` block above line 92 (or below `normalizePagePath`) so each docblock stays attached to its function.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-list/dot-experiments-list.component.html
```diff
@@ -195,12 +232,20 @@
...
+                                        ? (row.goalLabelKey | dm)
+                                        : EMPTY_CELL_PLACEHOLDER
                                 }}
                             </td>
                             <td class="text-center!" data-testid="experiment-variants">
                                 {{ row.variants }}
                             </td>
+                            <td
+                                class="max-w-40 truncate"
+                                [title]="row.createdByName"
```
nicobytes: **Nit:** `[title]` is a DOM property binding, and `HTMLElement.title` is a non-nullable `DOMString`, so when `createdByName` is `null` (backend without #37304 — the exact case the `—` placeholder is for) the cell gets a hover tooltip reading `null`.

```html
[attr.title]="row.createdByName"
```

`attr.` removes the attribute when the value is `null`.

core-web/libs/ui/src/lib/components/dot-user-filter/dot-user-filter.component.ts
```diff
@@ -0,0 +1,205 @@
...
+     * Every label here is a **key**, not translated text, so that the component is complete on its
+     * own and a consumer only names what it wants to differ. Defaulting to the plain noun means a
+     * screen with nothing special to say can drop the input entirely.
+     */
+    readonly $titleKey = input<string>('users.filter.title', { alias: 'titleKey' });
+
+    /** Message key for what the chip reads while nothing is selected. */
+    readonly $emptyLabelKey = input<string>('all', { alias: 'emptyLabelKey' });
+
+    /** Applied user ids. Owned by the consumer, usually backed by its address. */
```
nicobytes: **Nit:** `$selected` has two stacked JSDoc blocks; only the second one is picked up by tooling. Drop this one-liner (the longer block below already says it).

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-list/dot-experiments-list.component.ts
```diff
@@ -325,23 +328,33 @@ export class DotExperimentsListComponent {
               };
     });
 
+    /** The search box and the three chips: everything the toolbar itself can narrow by. */
```
nicobytes: **Nit:** "the three chips" — there are four now (status, goal, schedule, created by). Same stale count at line 527 of this file and in the `filtersCleared` doc in `dot-experiments-list-page.events.ts:56`. Suggest "the chips" so the comment doesn't rot on the next one.

nicobytes (commented):
Reviewed across correctness, readability, architecture, security and performance, against `ff3310c`.

Overall this is solid work. The narrowing chain is set up so the counts are taken before the new filters apply, the URL round-trip validates every parameter (strict `yyyy-MM-dd` with an overflow check, creators kept as open-set ids), and the shared-component changes are opt-in with defaults that keep today's behaviour. Name resolution is bounded (concurrency 3, 5 pages per id), and moving the `debounceTime` → `debounce(() => timer(...))` fixes a real bug with signal inputs.

**Required before merge**
1. **Clear all drops a working `pageId` narrowing** in the portlet (inline on `dot-experiments-filter-facade.ts`). `filtersCleared` was written for the dead-end empty state and is now also wired to a button that shows in the normal state.
2. **A resolved creator name gets overwritten by the raw id** after a second toggle (inline on `dot-user-filter.component.ts`). The automated review flagged this and it is still there.
3. **Orphaned JSDoc** in `dot-experiments-list-store.util.ts`: the `normalizePagePath` doc now sits on `ScheduleBounds`.

**Nits**: `[title]` renders a `null` tooltip, a duplicated JSDoc on `$selected`, and stale "three chips" comments.

**FYI (no action needed if already agreed):**
- Security: the Created By chip exposes the whole user directory, names plus the emails in the payload (also used as the display-name fallback), to any back-end user who can open Experiments. `contracts/user-directory-search.md` records this as accepted. Please make sure that sign-off comes from someone beyond the PR author, since it widens where the directory is visible. `roleKey` is the narrowing tool if the answer changes.
- Size: about 4.7k lines across four features the description itself calls independently shippable. It is reviewable only because the spec is so thorough. Next time, one PR per addition (column / pagination / schedule / created-by) would review and revert much more cheaply.

Tests look thorough for the happy paths. The two Required behaviours above have no coverage yet, so each fix should bring a regression test.

core-web/libs/data-access/src/lib/dot-users/dot-user-search.service.spec.ts
```diff
@@ -0,0 +1,274 @@
...
+    i18nMessagesMap: {},
+    pagination: { currentPage: 1, perPage: 20, totalEntries }
+});
+
+describe('DotUserSearchService', () => {
+    let service: DotUserSearchService;
+    let httpMock: HttpTestingController;
+
+    beforeEach(() => {
+        TestBed.configureTestingModule({
```
nicobytes: This HTTP service spec should use Spectator's `createHttpFactory` / `SpectatorHttp` instead of `TestBed` + `HttpTestingController`. That's the required harness for HTTP tests (`docs/frontend/TESTING_FRONTEND.md`), and the rest of `data-access` already does it (e.g. `dot-workflows-actions.service.spec.ts`).

```ts
import { createHttpFactory, HttpMethod, SpectatorHttp } from '@openng/spectator/vitest';

describe('DotUserSearchService', () => {
  let spectator: SpectatorHttp<DotUserSearchService>;
  const createHttp = createHttpFactory(DotUserSearchService);

  beforeEach(() => (spectator = createHttp()));

  it('should request the directory with query, page and per_page', () => {
    spectator.service.searchPage({ filter: 'jane', page: 2, perPage: 20 }).subscribe();

    const req = spectator.expectOne('/api/v1/users/filter', HttpMethod.GET);
    expect(req.request.params.get('query')).toBe('jane');
    expect(req.request.params.get('page')).toBe('2');
    expect(req.request.params.get('per_page')).toBe('20');
  });
});
```

`dotUserDisplayName` can stay as a plain `describe` — it does not hit HTTP.

core-web/libs/data-access/src/lib/dot-users/dot-user-search.service.ts
```diff
@@ -0,0 +1,234 @@
...
+ * screens. A caller on a screen reachable by a lower-privileged editor is widening *where* the
+ * directory is exposed, not whether it is — and should say out loud that it accepts that. The
+ * narrowing tool if it does not: the endpoint takes a `roleKey`, so candidates can be limited to
+ * holders of a given role (e.g. `DOTCMS_BACK_END_USER`).
+ *
+ * Anonymous and default users are excluded unless asked for: `includeanonymous` and
+ * `includedefault` both default to false and this service does not send them, so the list is real
+ * people rather than system accounts.
+ */
+@Injectable({ providedIn: 'root' })
```
nicobytes: **Nit:** Angular 22's `@Service()` is the shorthand for `@Injectable({ providedIn: 'root' })` (and what `ng generate service` emits now). This class already uses `inject()`, so it is eligible:

```ts
import { inject, Service } from '@angular/core';

@Service()
export class DotUserSearchService {
    readonly #http = inject(HttpClient);
```

No behaviour change — just the current decorator for a root singleton.

adrianjm-dotCMS (approved):

core-web/libs/ui/src/lib/components/dot-chip-filter/dot-chip-filter.component.html
```diff
@@ -10,7 +10,12 @@
          surrounding indentation as leading and trailing spaces inside the span, which is how
          "Locale: English" once became "Locale : English". -->
     <span class="font-semibold" data-testid="chip-title">{{ titleLabel() }}</span>
-    @if (title() && active() && !isToggle()) {
+    @if (loading() && !isToggle()) {
+        <!-- Ahead of the value branch deliberately: while the labels are still being resolved
+             there is nothing truthful to render after the title, and the raw values would be an
+             identifier the user never chose. -->
+        <i class="pi pi-spin pi-spinner ml-1 text-sm leading-none" data-testid="chip-loading"></i>
```
nicobytes: **Nit:** this loading icon is new markup, so it should be Material Symbols rather than PrimeIcons (`ANGULAR_STANDARDS.md` — PrimeIcons is legacy-only; do not mass-migrate existing ones). The house spinner is `progress_activity` + `animate-spin`:

```html
<i
    class="material-symbols-outlined ml-1 animate-spin text-sm leading-none"
    aria-hidden="true"
    data-testid="chip-loading">
    progress_activity
</i>
```

Leave the existing `pi-times` / `pi-chevron-down` on this chip alone — those were already here.

core-web/libs/ui/src/lib/components/dot-user-filter/dot-user-filter.component.ts
```diff
@@ -0,0 +1,205 @@
...
+    constructor() {
+        /**
+         * Folds each resolution into the accumulator.
+         *
+         * The resource decides *when* to fetch and what to cancel; this only records what came
+         * back. Writing to a signal read by the same resource's params would loop, hence the
+         * `untracked` — the guard is that a name just recorded leaves `#unresolvedIds` smaller,
+         * never larger, so the sequence terminates.
+         */
+        effect(() => {
```
nicobytes: **Nit:** this `effect` + `untracked` is exactly what NgRx `signalMethod` does — it tracks the input signal and runs the callback untracked, so the write to `#namesById` cannot loop back into `#resolved`.

```ts
readonly #foldResolved = signalMethod<Record<string, string> | undefined>((resolved) => {
    if (!resolved) {
        return;
    }

    this.#namesById.update((names) => ({ ...names, ...resolved }));
});

constructor() {
    this.#foldResolved(this.#resolved.value);
}
```

Same accumulator, no manual `untracked`. Optional — the current form is correct and the comment already explains the loop.

core-web/libs/portlets/dot-experiments/portlet/src/lib/store/dot-experiments-filter-facade.ts
oidacra: Confirmed, and the repro is exact. Fixed in c1f3f64.

The bar now dispatches its own `chipFiltersCleared`, and `filtersCleared` stays what it was written for — the no-results button, the only place a page narrowing that matched nothing can be widened. Both fold one shared `CLEARED_TOOLBAR_NARROWINGS` patch and the empty state's reducer adds the page fields to it, so a filter added later cannot land in one and not the other.

On the search term: it goes with the chips. `$hasNonDefaultFilters` still ignores it, so a term alone never reveals the button, but once the button is on screen a "Clear all" that leaves the search box full does not mean what it says. Written down on the event rather than left to be re-derived.

Three store tests, plus the component one now asserting the bar dispatches the chip-scoped event and *not* `filtersCleared`. The one that matters is "should keep a page narrowing and its return context": it hydrates `selectedPageId`, `selectedPageUrl` and `languageId`, sets a status chip, clears, and asserts all three survive. Red before the fix.

FR-051a said what reveals the button but not what it does — that gap is why the code and the doc could disagree. It now says both.

core-web/libs/ui/src/lib/components/dot-user-filter/dot-user-filter.component.ts
oidacra: Confirmed, including the part that makes it stick. Fixed in c1f3f64.

`onSelectionChange` now folds in only `label !== value`. The list labels what it has loaded and is rebuilt on every popover open, so its fallback to the id is exactly the case that must not be written back — and writing it back is permanent, because an id counts as a resolved name and `#unresolvedIds` never asks again.

The test is the one you described: `resolveNames` returns `{ u1: 'Jane Doe' }`, the chip renders it, then a selection arrives with `u1` labelled `u1` and `u2` labelled properly, and the chip must still read Jane Doe. Red before the fix with `expected ': u1, Name u2' to contain 'Jane Doe'` — the id had replaced the name.

core-web/libs/portlets/dot-experiments/portlet/src/lib/util/dot-experiments-list-store.util.ts
oidacra: Confirmed. Fixed in c1f3f64 — the page-path block now sits on `normalizePagePath` where it was written, and `ScheduleBounds` keeps its own one-liner.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-list/dot-experiments-list.component.html
oidacra: Confirmed, and it is the exact case the placeholder covers — the cell reads the em dash while the tooltip says "null". `[attr.title]` in c1f3f64.

core-web/libs/ui/src/lib/components/dot-user-filter/dot-user-filter.component.ts
oidacra: Confirmed. The one-liner is gone in c1f3f64; the longer block already said all of it.

core-web/libs/portlets/dot-experiments/portlet/src/lib/dot-experiments-list/dot-experiments-list.component.ts
oidacra: Confirmed — four now. All three read "the chips" in c1f3f64, which is the version that does not rot on the next one.

core-web/libs/portlets/dot-experiments/portlet/src/lib/store/dot-experiments-filter-facade.ts
oidacra: Correcting what I said above: I kept **both** page fields, and `?url=` should not have been one of them. Fixed in 659ab8f.

What decides it is which of the two the application writes. `pageId` is the scope this screen hands out on all four ways out of the list (`pageFilterParams`) and reads back on the way in (`listReturnParams`) — that one stays, exactly as you asked. `?url=` has no writer anywhere in the app: UVE's experiments item carries no `href` with the portlet switch on, and routes to `experiments/{id}` as a path segment with it off. It only ever arrives typed or pasted, which makes it a filter the user applied rather than a scope handed down.

Left in force it was also unreachable. `?url=` is echoed back into the address on every change, and the empty state's button is the only other way out — and that appears only once the path matches nothing. So a path still matching rows could not be cleared by any control on the screen.

Test first and red: a hydrated `/blog/post` must be null after the chip-scoped clear while `pageId` and `language_id` survive. Walked both ways in the browser too — `?url=/index&status=DRAFT` clears to a bare address; `?pageId=…&language_id=1&status=DRAFT` keeps the page and its language.

FR-051a and the URL-parameters contract now carry the asymmetry and why.

core-web/libs/data-access/src/lib/dot-users/dot-user-search.service.spec.ts
oidacra: Right, and thirty-four other specs in `data-access` already do it. Moved onto `SpectatorHttp` in 74f291d.

One departure from your snippet, because it does not work here: `spectator.expectOne(url, method)` compares the whole `urlWithParams`, and every call this service makes carries a query string — which is precisely what the assertions are about rather than how the request is found. So the harness is Spectator's and the matcher is a predicate through `spectator.controller`. `dotUserDisplayName` stayed a plain `describe`, as you said.

core-web/libs/data-access/src/lib/dot-users/dot-user-search.service.ts
oidacra: Checked it before taking it: `@Service` is real public API in 22.1.0 (`declare const Service: ServiceDecorator`), and it was the first use in this repo — there are zero others and no doc asks for it. Adopted in 74f291d, but not as a root singleton.

Reviewing it raised the better question: should this be in root at all? It holds no cache, no subscription and no state — it is a shaped call — so `providedIn: 'root'` was putting it in every injector that never asks for it. It now carries `@Service({ autoProvided: false })` and `DotUserFilterComponent` provides it, so its lifetime matches the control that wants it. One consumer today, so nothing else moves.

Two consequences in the spec, both real rather than bookkeeping: the mock moves to `componentProviders`, because a TestBed provider is shadowed by the component's own and every test would have hit the real HTTP path; and the injection reads from the component injector, so the tests arm the instance the component actually resolved.

core-web/libs/ui/src/lib/components/dot-chip-filter/dot-chip-filter.component.html
oidacra: Correct, and it is the one of the four that a standard actually settles: `ANGULAR_STANDARDS.md` reserves PrimeIcons for markup that already exists, and this icon is new. `progress_activity` + `animate-spin` in 74f291d, with `pi-times` and `pi-chevron-down` left alone as you said.

Worth noting what this turned up: the `loading` input had **no test at all**. It has three now — the spinner replaces the values, the icon is the house one, and the values come back when loading ends. The icon test was red against the PrimeIcons markup. Confirmed in the browser too, catching it mid-resolve: `progress_activity` with `material-symbols-outlined animate-spin`.

core-web/libs/ui/src/lib/components/dot-user-filter/dot-user-filter.component.ts
oidacra: Taken, in 74f291d. It is the same guarantee stated once instead of assembled by hand, and `signalMethod` is already used in analytics and edit-ema, so it is the idiom here rather than a new dependency.

core-web/libs/ui/src/lib/components/dot-user-filter/dot-user-filter.component.ts
oidacra: Following up, because this nit turned out to be worth more than a nit and two later commits come from it.

First, I reintroduced it myself. Adding the provider note to `DotUserSearchService` left exactly this shape on its class, and the block that would have been dropped is the one recording that `/v1/users/filter` applies no permission filtering and hands every back-end user the whole directory, email addresses included. Folded in e34b29a.

Then I swept the pattern across everything this branch touches. Five more, all predating the branch — checked against the merge base, not assumed — and four worse than cosmetic, because the orphaned block describes a *different* symbol, so the hover is wrong rather than missing:

- `DotExperimentsListViewState` had lost its doc to `DotExperimentsListPanelScope`. It was also stale — it listed "filter, status selection, paging and sort" while the interface now also carries goal, creators, schedule and the page narrowing. Reattached and rewritten without the enumeration, which is what went stale; same reason as the chip count you caught.
- `ConfigureFormModel` had lost its doc to `ConfigureFormValidity`.
- `PAGE_PREFILL_ERROR_KEY` had lost its one-liner to `PAGE_LOOKUP_LIMIT`.
- The Configure store spec had two blocks saying the same thing about one `describe`.

The fifth is deleted rather than moved: "Empty-state copy. Resolved once …, declared after the injections because field initialisers run in declaration order" describes a field that does not exist. It was born orphaned in 0b394d6056, and the copy it means now lives in the `$emptyConfiguration` computed, of which none of that sentence is true. Reattaching it would have propagated a false claim.

All in a27ba0c6, comments only. One match is left and is deliberate: the module-level block at the top of `dot-experiments-list-store.util.ts`, which is separated by a blank line and attached to nothing.

adrianjm-dotCMS (approved):

nicobytes (approved):

PR #37482, opened 2026-09-09
docs(edit-content): spec for #37192 relationship field picker and related-list fixes

adrianjm-dotCMS:
**PR 1 of 2 — spec only.** No code. Per [Spec-Kit](https://github.com/dotCMS/core/blob/main/docs/core/SPEC_KIT_QUICK_START.md#3-two-prs-the-spec-is-approved-before-you-build), this carries `spec.md` alone and needs **approval** (not merge) before planning starts. Implementation lands in PR 2, branched off this one.

Resolves the specification for #37192.

### Proposed Changes

* Adds `specs/37192-relationship-field-assetpicker/spec.md` — 36 functional requirements, 43 acceptance scenarios across 6 user stories. One file, no code.

### What the spec says

It splits #37192 into **two independent bodies of work** that happen to share a field:

**A · The "Add Relationships" picker** (US 1–3)
* Moves onto the shared search surface: full-width search bar + filter chips (site/folder, Locale). **No folder-tree sidebar** — the folder scope is a chip, per the design artboard.
* Searches through `api/v1/drive/search`, replacing the relationship-specific search path.
* It is a **selection editor**, not an add-only dialog: opens with already-related items checked, unchecking removes, confirming leaves the relationship equal to the selection.
* The confirm action is **never disabled** — confirming with an empty selection unrelates everything.
* **Keeps its own paging.**

**B · The related-content list in the form** (US 4–6)
* Loses its 6-row paging; renders 40 rows plus a **"Load more"** control, following the Key/Value field precedent (#37191). Withholding is a rendering limit only — the value always holds every item.
* Hover-only drag handle, right-aligned Status column, single-column form max width (#36615).

### Decisions worth a reviewer's eye

Three scope questions had no safe default and were settled before writing:

| Question | Settled as |
|---|---|
| How much of the AssetPicker shell is adopted? | Shared search + chips + table + paging. **No sidebar** — settled by the design. |
| How is multi-select satisfied? | It already exists: `dot-folder-list-view` defaults to `selectionMode: 'multiple'` and Content Drive already holds a set. Configuration, not a new capability — the AssetPicker's single-slot selection is untouched. |
| What bounds the form's list without paging? | 40 + "Load more" (#37191 precedent). |

**One deliberate addition beyond the issue's original ACs**: the confirm action is never disabled. The issue said nothing about it; it is the only reading consistent with "unchecking a row removes it". **The issue's acceptance criteria were updated to match** before opening this PR.

[...]

Issue #36615, opened 2026-07-17
New Edit Content: apply field padding (18px) and form max-width (720px/1000px) guidelines


erickgonzalez:
### Description

The new Edit Content component must follow these **padding** and **viewport width** guidelines.

**Padding (gap between fields):**
- **Single-column** layout: **18px** vertical spacing between fields.
- **4-column grid**: **18px** gap between fields both **row** and **column**.

**Max form/viewport width:**
- **One column**: max width **720px** — a lone field fills the full form width.
- **Multiple columns**: max width **1000px**.

### Acceptance Criteria

**Padding**
- [ ] In a single-column layout, the vertical spacing between consecutive fields is **18px**
- [ ] In the 4-column grid, the **row** gap between fields is **18px**
- [ ] In the 4-column grid, the **column** gap between fields is **18px**

**Max width**
- [ ] A single-column form is constrained to a max width of **720px**
- [ ] In a single-column layout, a lone field **fills the full form width** (up to 720px)
- [ ] A multi-column form is constrained to a max width of **1000px**

**Consistency**
- [ ] The padding and max-width rules apply consistently across content types with different field/column configurations

### Priority

Medium

### Additional Context

Part of the new Edit Contentlet work (`dotCMS: New Edit Contentlet`). These are exact layout values (18px gaps; 720px single-column / 1000px multi-column max width) to standardize the form layout in the new Edit Content component.

melissarojas-dotcms:
Passed QA

PR #37515, opened 2026-09-11
feat(edit-content): refactor the relationship field onto the shared search surface (#37192)

### LINKED ISSUE #37192: Refactor relationship field to new design, reusing AssetPicker filters/search (Edit Contentlet)
### Description

Refactor the **Relationship field** in the new Edit Content to follow the new design, **reusing as much as possible from the AssetPicker component** built for Content Drive (#36702) — filters, search, and the same backend endpoint.

🎨 **Design:** https://claude.ai/design/p/202fc776-9326-4926-b4c6-5996a8be9ea1?via=share&file=Asset+Picker.dc.html
💬 **Context (Slack):** https://dotcms.slack.com/archives/C0893BF7GRJ/p1787232866639669

**Reuse goals:**
- Reuse the **AssetPicker** shell/interaction patterns (#36702) where applicable.
- Reuse the **filters** and **search** UI.
- Use the **same endpoint** the AssetPicker/Content Drive uses (`api/v1/drive/search`) for searching related content.

> This is the larger design refactor of the relationship field. It is **separate from #36155** (smaller relationship-field refinements: chip colors, Locales column, hint, empty state, showFields) — cross-referenced, not superseded.

### Acceptance Criteria

**Design refactor**
- [ ] The relationship field is refactored to match the new **design** (linked above)
- [ ] Relating and unrelating content works with the new design
- [ ] The refactor preserves relationship-field behavior (cardinality, ordering, permissions)
- [ ] The picker is a **selection editor**: it opens with already-related items **checked**, checking adds and **unchecking removes**; confirming leaves the relationship equal to the selection
- [ ] The confirm action is **never disabled** — confirming with an empty selection unrelates every item

**Reuse AssetPicker & Content Drive patterns (#36702)**
- [ ] Filters are reused from the AssetPicker component (not re-implemented) where applicable
- [ ] The search reuses the AssetPicker/Content Drive search, using the **`api/v1/drive/search`** endpoint
- [ ] Shared UI/interaction patterns from the AssetPicker are reused rather than duplicated
- [ ] The picker presents a full-width search bar plus **filter chips** (site/folder, Locale) — **no folder-tree sidebar**; the folder scope is a chip

**Consistency & regressions**
- [ ] Existing related content loads and displays correctly after the refactor
- [ ] Saving a contentlet persists relationship changes correctly (no data loss)
- [ ] The **picker keeps its own paging** (page, prev/next, page size). Only the related-content list *in the form* loses its paging — the two lists are different lists

**Relationship component fixes (not picker-related)**

Additional fixes to the relationship **component shown in the new Edit Content** (the list of related content), independent of the picker:

- [ ] The **drag handle** is shown only **on hover** of a related-content row (hidden otherwise)
- [ ] The **Status** column is **right-aligned**
- [ ] The component respects the form **max width when there is only one column** (verify/fix against the single-column max width — see #36615)
- [ ] **Pagination is removed** from the related-content list — no paging control. The list renders the first **40** rows and appends a **"Load more"** control that reveals 40 more, following the Key/Value field precedent (#37191). Withholding is a **rendering** limit only: every related item stays in the field's value, so a drag never loses a withheld row
- [ ] With pagination removed, drag-reordering works across **all** related items (not just a single page)

### Suggested Tests

Frontend refactor + component fixes (search reuses the existing `api/v1/drive/search` endpoint — no backend change). **Unit-first**; E2E covers the relate/unrelate + drag-reorder flows that are hard to assert fully in unit. Integration/Postman are not applicable.

- [ ] The relationship field renders the new design (relate control, related-content rows)
- [ ] Relating content adds it to the model; unrelating removes it
- [ ] The search calls the **`api/v1/drive/search`** endpoint with the expected params (reused, not re-implemented)
- [ ] Filters reuse the AssetPicker filter component (assert the shared component is used)
- [ ] The **drag handle** is hidden by default and shown on row **hover**
- [ ] The **Status** column is **right-aligned** (assert the alignment class/style)
- [ ] With a single-column form, the component honors the single-column **max width** (see #36615)
- [ ] **No pagination** control is rendered in the related-content list; ≤40 items render fully with no *Load more*, >40 render 40 rows plus *Load more*, and the withheld items remain in the emitted value

**E2E (Playwright)** — flows hard to unit-cover
- [ ] Open a contentlet with a relationship field → search + relate content → the item appears → save → persists
- [ ] Unrelate an item → save → it is removed
- [ ] **Drag-reorder** across the full related set (with pagination removed) → new order is saved and reloads correctly

**Integration / Postman**
- [ ] No tests — search reuses the existing `api/v1/drive/search` endpoint; relationship persistence uses existing content endpoints. No API contract change.

### Spec

Spec-driven work ([Spec-Kit](https://github.com/dotCMS/core/blob/main/docs/core/SPEC_KIT_QUICK_START.md)): `specs/37192-relationship-field-assetpicker/spec.md`, on branch `issue-37192-relationship-field-assetpicker`. PR 1 carries the spec alone and must be approved before implementation starts.

**Not in scope:** the **"New content"** path (creating related content in place, side panel or dialog per the feature flag) is unchanged — only the "Existing content" path is refactored.

### Priority

Medium

### Additional Context

Part of the new Edit Contentlet work (`dotCMS: New Edit Contentlet`), Relationship field. Reuses the AssetPicker component (#36702) and the shared `api/v1/drive/search` endpoint for search/filters. Cross-references #36155 (smaller relationship-field refinements — kept separate per scope decision). Product context is in the linked Slack thread.

This issue also bundles fixes to the relationship **component** (the related-content list) that are independent of the picker: hover-only drag handle, right-aligned Status column, single-column max-width handling (see #36615), and **removing pagination** so drag-reordering works across the full related set.

adrianjm-dotCMS:
**PR 2 of 2 — implementation.** Spec: `specs/37192-relationship-field-assetpicker/spec.md`, approved in #37482.

Closes #37192.

## What this does

**The "Add Relationships" picker.** The bespoke dialog is replaced by the shared search surface: `dot-filter-bar` chips over a picker-owned `DOT_FILTER_FACADE`, `dot-folder-list-view` for the results, the shared site/folder browser behind a filter chip, and `api/v1/drive/search` pinned to the relationship's target content type. Full-width search bar with the chip row beneath it; the folder scope is a chip, not a sidebar.

**The related-content list in the form.** Paging removed; 40 rows plus **Load more**, following the Key/Value field (#37191). Hover-only drag handle, right-aligned Status, single-column form max width.

## Acceptance criteria

**Design refactor**
- [x] Refactored to the new design
- [x] Relating and unrelating work with the new design
- [x] Cardinality, ordering and permissions preserved — including the guard that keeps a child already claimed by another parent from being picked
- [x] The picker is a **selection editor**: already-related items open checked, unchecking removes, confirming leaves the relationship equal to the selection
- [x] Confirm is **never disabled**; confirming empty unrelates everything

**Reuse (#36702)**
- [x] Filters reused, not re-implemented
- [x] Search goes through `api/v1/drive/search`
- [x] Shared interaction patterns reused rather than duplicated
- [x] Full-width search plus filter chips, no folder-tree sidebar

**Consistency & regressions**
- [x] Existing related content loads and displays correctly
- [x] Saving persists relationship changes with no data loss
- [x] The picker keeps its own paging; only the list in the form loses paging

**Relationship component fixes**
- [x] Drag handle on hover only
- [x] Status right-aligned
- [x] Single-column form max width (#36615)
- [x] Paging removed from the list — 40 rows plus Load more, withholding is rendering-only
- [x] Drag-reorder works across all related items

**Spec amendment:** FR-014 (a review of the current selection before confirming) is **withdrawn** — the picker carries no "Show Selected" control. The spec records it and the consequence; it needs re-approval, and #37482 still shows the original text.

## Second consumer

`DotSelectExistingContentComponent` was not only the relationship field's — Content Drive's field-filter chip opens it through `DOT_RELATIONSHIP_PICKER`. That provider is re-pointed here and its contract spec passes with only presentation assertions updated.

## Tests

| | |
|---|---|
| E2E (`relationship-field`) | 50 passed |
| `edit-content` | 2279 |
| `ui` | 2031 |
| `portlets-content-drive` | 1366 |

Relate, unrelate and drag-reorder each verified against a running instance through `GET /api/v1/content/{id}?depth=1`, not only in the rendered list.

No Integration or Postman: no endpoint, request or response shape changes.

## Two calls for the reviewer

**Six `libs/ui` edits.** The plan said to consume that library and not modify it. All additive with defaults, `ui` green:
* `dot-chip-filter` — optional `icon`; a titleless chip renders its value as its label
* `dot-folder-list-view` — `unselectable` rows, which is what makes the cardinality guard visible
* `dot-asset-picker` — full-screen behaviour extracted to a shared module, now used by both

**Columns are Content Drive's, not the design's.** Shown: Name, Status, Locale, Type, Edited By, Last Edited. The artboard asks for Title, URL Title, Posting Date, Locale, Status, Last Modified.

This PR fixes: #37192


core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/constrained-identifiers.service.ts
```diff
@@ -0,0 +1,105 @@
...
+
+                            if (childId) {
+                                constrainedIds.add(childId);
+                            }
+                        }
+                    }
+
+                    return constrainedIds;
+                }),
+                catchError(() => of(new Set<string>()))
```
oidacra: This `catchError` returns an empty set on any failure of the `/api/content/_search` lookup — a 500, a network drop, a 403, a timeout all collapse to "nothing is constrained". Since `constrainedIdentifiers` then never becomes non-empty, no row is marked unselectable, `applyConstrained` prunes nothing, and the guard stays off across all three arrival paths (search results, pre-checked existing items, every later page) for the life of the dialog.

The effect is that an editor can pick a child already claimed by another parent under ONE_TO_ONE/ONE_TO_MANY, and saving silently reparents it — no toast, no banner, no log at any layer.

Was failing open deliberate here? Given the guard exists precisely to stop a silent reparent, would failing closed be better — or at least surfacing the degraded state so the editor knows the check could not run?

core-web/libs/edit-content/src/lib/fields/dot-edit-content-host-folder-field/components/host-folder-field/host-folder-field.component.html
```diff
@@ -5,61 +5,81 @@
...
-            data-testid="host-folder-trigger-label">
-            @if (store.showTriggerLoading()) {
-                <span data-testid="host-folder-trigger-skeleton">
-                    <p-skeleton width="60%" height="1rem" />
+         The wrapper carries the click: projection moves nodes, it does not hand them this
+         component's handlers, so a projected trigger with no wrapper is a button that does nothing.
+         `inline-flex` rather than `contents` because `toggleOverlay` measures the trigger to size
+         the overlay, and an element with `display: contents` has no width to measure. -->
+    @if (!$showDefaultTrigger()) {
+        <span class="inline-flex" (click)="toggleOverlay($event)">
```
oidacra: This wrapper gives the projected trigger a native `(click)`, which covers the mouse. Does it leave the keyboard path broken for the new site chip?

Tracing it: `dot-chip-filter`'s host is `role="button"` with `[attr.tabindex]`, but it is not a native `<button>`. Its `(keydown.enter)`/`(keydown.space)` bindings call `onHostKeydown`, which emits the Angular `clicked` output — and an Angular output does not dispatch a bubbling DOM click. This span listens for the native `click` only, and `add-relationships-site-chip.component.html` binds no `(clicked)`.

So the chip takes focus and announces itself as a button, but Enter and Space appear to do nothing, while a mouse click opens the browser. The component's default `<button>` trigger is unaffected — this is specific to the new `showDefaultTrigger=false` projection path.

Would binding `(clicked)="toggleOverlay($event)"` on the projected chip be the fix you'd prefer?

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/add-relationships.store.ts
```diff
@@ -0,0 +1,677 @@
...
+                                        })
+                                })
+                            )
+                        );
+                    })
+                )
+            );
+
+            /**
+             * Fetches the identifiers already claimed by another parent.
```
oidacra: There are two JSDoc blocks stacked on `applyConstrained` here. This first one ("Fetches the identifiers already claimed by another parent… Guarded, not unconditional…") reads like a description of `loadConstrained`, the method that actually performs the guarded fetch a few lines below — while `applyConstrained` prunes the selection synchronously and unconditionally. The second block immediately under it describes `applyConstrained` correctly.

On a guard this correctness-sensitive, is the first block a leftover that should move to `loadConstrained`, or be dropped?

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/add-relationships.component.ts
```diff
@@ -0,0 +1,225 @@
...
+        DotFilterBarComponent,
+        DotFolderListViewComponent,
+        DotLanguageFilterChipComponent,
+        DotSearchInputComponent,
+        DotMessagePipe,
+        AddRelationshipsSiteChipComponent,
+        AddRelationshipsFooterComponent
+    ],
+    templateUrl: './add-relationships.component.html',
+    changeDetection: ChangeDetectionStrategy.OnPush,
```
oidacra: ANGULAR_STANDARDS.md (line 45) asks new components not to set `changeDetection` at all, since OnPush is the framework default as of v22. This is set explicitly here and in two other new components in this PR — `components/footer/footer.component.ts:31` and `components/site-chip/add-relationships-site-chip.component.ts:39`.

Worth dropping the property and the now-unused `ChangeDetectionStrategy` import from all three?

core-web/apps/dotcms-ui-e2e/src/tests/edit-content/fields/relationship-field/helpers/add-relationships-dialog.ts
```diff
@@ -0,0 +1,272 @@
...
+ *   paginator, which is part of the table rather than the dialog.
+ *
+ * All locators are scoped to `this.dialog` so they cannot match the form behind the mask.
+ */
+export class AddRelationshipsDialog {
+    readonly dialog: Locator;
+    readonly table: Locator;
+    readonly rows: Locator;
+
+    constructor(private page: Page) {
```
oidacra: TYPESCRIPT_STANDARDS.md asks for the `#` prefix on private class fields and specifically says not to use the `private` keyword for instance properties. Since this file is new in this PR, would `readonly #page: Page` be preferable to `private page: Page` here?

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/add-relationships.store.spec.ts
```diff
@@ -0,0 +1,724 @@
...
+    contentCount: list.length,
+    folderCount: 0,
+    hasMoreContent: false,
+    hasMoreFolders: false,
+    nextContentCursor: 0,
+    nextFolderCursor: 0,
+    ...overrides
+});
+
+describe('AddRelationshipsStore (US2 — selection)', () => {
```
oidacra, line 37: The store's `load()` has an explicit error branch that sets `status: ComponentStatus.ERROR` and `errorMessage: 'dot.relationship.add.dialog.search.failed'`, and `add-relationships.component.html` renders that message in place of the results table under `data-testid="add-relationships-error"`. I could not find a test in this suite that makes the search throw and asserts either.

That path is the one thing standing between a failed search and an empty table that reads as "no matches", so it seems worth pinning. Would you add a case with `searchMock.mockReturnValue(throwError(() => new Error('boom')))` asserting `status()` and `errorMessage()`?

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/add-relationships.store.spec.ts
```diff
@@ -0,0 +1,724 @@
...
+    const lastRequest = () => searchMock.mock.calls.at(-1)?.[0];
+
+    beforeEach(() => {
+        searchMock.mockReset().mockReturnValue(of(response([item(1), item(2)])));
+        constrainedMock.mockReset().mockReturnValue(of(new Set<string>()));
+        spectator = createService();
+        store = spectator.service;
+    });
+
+    describe('seeding the selection (FR-009)', () => {
```
oidacra, line 76: Two branches in the store look unexercised by this suite:

- `AddRelationshipsInput.selectedInodes` drives real seeding logic in `load()` — it seeds `pendingInodes` and resolves them into `selection` as matching rows page in — but no test passes `selectedInodes` to `initialize()`; only the identifier-based `selected` array is covered.
- `setPage(number, limit?)` has a `changedSize` branch that forces page 1 and discards the cursor bookmarks; every call here passes only `number`.

Since the pre-selection path is what Content Drive relies on, is it worth covering both?

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/store/relationship-field.store.spec.ts
```diff
@@ -873,6 +642,121 @@ describe('RelationshipFieldStore', () => {
             });
         });
     });
+
+    describe('Load more (US4)', () => {
```
oidacra, line 646: This block pins "revealed rows stay revealed" for `deleteItem` and `reorderData`, but I don't see the equivalent for `setData`.

`setData` is the method the picker's close path calls, and it no longer resets any paging-like state — the old implementation reset `pagination`, the new one leaves `visibleCount` untouched. That is the same contract the two tests below protect, and it is the path most likely to produce a "the table snapped back to 40 rows" report.

Would a `loadMore()` then `setData(...)` case asserting `visibleCount()` be worth adding, whichever way the intended contract goes?

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/store/relationship-field.store.spec.ts
```diff
@@ -873,6 +642,121 @@ describe('RelationshipFieldStore', () => {
...
+        });
+
+        it('does not lose withheld items when a visible row is removed', () => {
+            store.setData(many(95));
+            store.deleteItem('inode0');
+
+            expect(store.data()).toHaveLength(94);
+        });
+
+        it('does not reorder withheld items as a side effect of a drag', () => {
```
oidacra, line 761: This test proves the negative case: reordering within the visible slice does not disturb the withheld items' relative order. The acceptance criterion is the positive one — "drag-reordering works across all related items (not just a single page)".

I could not find a test, unit or E2E, that moves an item from beyond `visibleCount` into a visible position; the E2E reorder spec still uses a 3-item fixture. The implementation looks correct — the rendered rows are always a prefix of `data`, so `dragIndex`/`dropIndex` need no translation — which is exactly why a test would lock it in cheaply.

Would you add a `reorderData` case moving an originally-withheld item to index 0?

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/components/footer/footer.component.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * US2 — the confirm action is never disabled (FR-013).
+ *
+ * This is implemented by *removing* a binding rather than adding a mechanism: the dialog this
+ * replaces carried `[disabled]="totalItems === 0"`, which made "uncheck the last row" the single
+ * case where unchecking did not unrelate. Content Drive's own footer already had it right, with the
+ * comment "clearing is a valid filter state".
+ */
+describe('AddRelationshipsFooterComponent (US2)', () => {
```
oidacra, line 19: Two gaps in this area:

- `confirmLabel` can be overridden through `DynamicDialogConfig` — the exact mechanism Content Drive's field-filter chip uses to render "Apply" — but no test here exercises the override. This suite also mocks an `AddRelationshipsStore.$selectedCount` that the component does not appear to read.
- Separately, in `add-relationships.component.spec.ts` the `sort` output wiring (`onSort` → `store.setSort`) has no test, while `paginate` gets two.

Worth covering the override at least, given a second consumer depends on it?

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/add-relationships-filter-facade.ts
```diff
@@ -0,0 +1,90 @@
...
+            // O9: a patch that changes nothing must not reset paging.
+            if (isNoOpFilterPatch(patch, (key) => store.getFilterValue(key))) {
+                return;
+            }
+
+            store.patchFilters(patch);
+            store.load();
+        },
+
+        // Same reload as `patchFilters`: clearing a chip is a filter change like any other, and a
```
oidacra: This two-line comment appears twice, back to back, above `removeFilter` — it looks like a copy-paste artifact. Drop one?

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/add-relationships.component.html
```diff
@@ -0,0 +1,87 @@
...
+            type="button"
+            [text]="true"
+            [rounded]="true"
+            [attr.aria-pressed]="$isFullscreen()"
+            [attr.aria-label]="$fullscreenLabel() | dm"
+            [pTooltip]="$fullscreenLabel() | dm"
+            data-testid="add-relationships-fullscreen"
+            (onClick)="toggleFullscreen()">
+            <ng-template #icon>
+                <i class="material-symbols-outlined text-[1.125rem]!">{{ $fullscreenIcon() }}</i>
```
oidacra: STYLING_STANDARDS favours theme scale utilities over arbitrary values. Would `text-lg` work here instead of `text-[1.125rem]!`?

And if the `!` is genuinely load-bearing against a PrimeNG rule, a short note saying so would help the next reader — the surrounding code is otherwise well annotated about its own reasoning.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/add-relationships.store.ts
```diff
@@ -0,0 +1,677 @@
...
+         * **Not `contentCount`.** The drive API is cursor-based and never returns a grand total —
+         * `contentCount` is the size of the page it just returned. Handing that to the paginator
+         * makes every full page look like the last one, so it computes a single page and disables
+         * "next". The AssetPicker documents this trap and solves it the same way; this store had
+         * walked straight into it.
+         *
+         * So: while the bookmark for the page on screen reports more, claim one page beyond to keep
+         * the arrow live; once it does not, the page is the last and the exact total is knowable.
+         */
+        totalItems: computed(() => {
```
oidacra: `totalItems` is the only computed in this store without the `$` prefix that every sibling uses (`$selectedItems`, `$constrainedList`, `$offset`, and so on). Intentional, or worth renaming to `$totalItems` for consistency? The template reads `store.totalItems()`, so it would be a small change.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/add-relationships.store.ts
```diff
@@ -0,0 +1,677 @@
...
+                            // cell in the related-content table. The dialog this replaced resolved them the
+                            // same way, for the same reason.
+                            forkJoin([
+                                contentDriveService.search(buildRequest()),
+                                // A languages failure must not take the results down with it: the
+                                // Locale column degrades to blank, which is exactly what it did
+                                // before this enrichment existed. `forkJoin` fails on its first
+                                // error, so the recovery has to sit on this arm rather than around
+                                // the pair.
+                                languagesService
```
oidacra: The graceful degrade here reads right — a languages failure should not take the search results down with it, and the comment says so. The one thing it costs is observability: the error is swallowed with no logging at all, so a recurring `DotLanguagesService` outage would surface only as a silently blank Locale column.

Worth keeping the fallback but logging the swallowed error?

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/add-relationships.store.ts
```diff
@@ -0,0 +1,677 @@
...
+            const load = rxMethod<void>(
+                pipe(
+                    tap(() =>
+                        patchState(store, { status: ComponentStatus.LOADING, errorMessage: null })
+                    ),
+                    switchMap(() => {
+                        // Nothing to ask for: the scope is not a browsable site (Shared Assets). An
+                        // empty result is the honest answer — the previous behaviour was a 500 shown
+                        // to the editor as "The search could not be completed".
+                        if (!store.assetPath()) {
```
oidacra, line 341: This short-circuit returns `status: LOADED` with no items and no error, so an unresolvable scope renders identically to a genuine zero-result search. The comment argues an empty result is the honest answer, which is fair — but the editor gets no hint that the scope itself is unusable rather than merely unmatched.

Is it worth distinguishing the two states — say a separate computed driving a short note next to the site chip — or is the site chip considered signal enough?

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/add-relationships.store.ts
```diff
@@ -0,0 +1,677 @@
...
+    rows: DotContentDriveItem[],
+    languages: DotLanguage[]
+): DotContentDriveItem[] => {
+    const byId = new Map(languages.map((language) => [language.id, language]));
+
+    return rows.map((row) => {
+        // `DotContentDriveItem` is a union with folders, which carry no language at all. The dialog
+        // asks for `showFolders: false`, so none should arrive — but the type admits them, and
+        // silently stamping `language: undefined` onto one would be a lie rather than a no-op.
+        const languageId = (row as DotCMSContentlet).languageId;
```
oidacra: These `as DotCMSContentlet` / `as DotContentDriveItem` casts (also at lines 331 and 582) rest on the API being asked for `showFolders: false` rather than on a runtime guard, so the union's folder rows are excluded by request shape plus a comment rather than by the type system. Would a small type guard be worth it here, given the row shape comes back over HTTP?

Minor and separate: `getFilterValue` (line 464) is the one public store method without an explicit return type.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/add-relationships.store.ts
```diff
@@ -0,0 +1,677 @@
...
+                /**
+                 * Adds or removes one item.
+                 *
+                 * A constrained item is refused outright rather than silently ignored downstream: it is
+                 * claimed by another parent, and letting it into the selection would produce broken
+                 * content on save.
+                 *
+                 * In `single` mode the selection is a slot, so a second pick replaces the first.
+                 */
+                toggleSelection(item: DotCMSContentlet): void {
```
oidacra, line 634: `selection` is a `Map`, so ordering on confirm follows insertion order. One consequence worth confirming: an item that is unchecked and then re-checked is re-inserted at the end rather than restored to its original position, so its place in the relationship order changes — whereas an item left untouched keeps its seeded position, and `Map.set` on an existing key updates in place without moving it.

That reads as consistent with "insertion order", and no acceptance criterion contradicts it. Is that the intended behaviour for a toggled-off-and-on-again row?

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/constrained-identifiers.service.ts
adrianjm-dotCMS: Fixed in c01c69a7 / f8fe0ca7. Failing open was deliberate but the silence was not — failing closed would make a timeout render every row unselectable. The service no longer swallows; the store flags it and the dialog says the check could not run, so the editor knows before confirming. Covered by a new `constrained-identifiers.service.spec.ts` (the store spec mocks the service, so it could not see this at all).

core-web/libs/edit-content/src/lib/fields/dot-edit-content-host-folder-field/components/host-folder-field/host-folder-field.component.html
adrianjm-dotCMS: Good catch, and the fix you proposed has a trap: `clicked` also fires on a native click, so binding it would toggle twice with the mouse. Fixed in f8fe0ca7 by handling `keydown.enter`/`keydown.space` on the wrapper instead — keydown bubbles, so it covers any projected trigger. It calls `preventDefault` first, otherwise a projected native `<button>` would get the browser's synthesised click on top and open-then-close in one keystroke.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/add-relationships.store.ts
adrianjm-dotCMS: Leftover, fixed in e914ad01. My insertion of `applyConstrained` landed between `loadConstrained`'s JSDoc and its function. Moved back.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/add-relationships.component.ts
adrianjm-dotCMS: Fixed in e914ad01 — dropped from all three, along with the now-unused import.

core-web/apps/dotcms-ui-e2e/src/tests/edit-content/fields/relationship-field/helpers/add-relationships-dialog.ts
adrianjm-dotCMS: Fixed in e914ad01 — `readonly #page`.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/add-relationships.store.spec.ts
adrianjm-dotCMS: Added in 834b6f13, asserting both `status()` and `errorMessage()`.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/add-relationships.store.spec.ts
adrianjm-dotCMS: Both added in 834b6f13. `selectedInodes` gets two cases — resolved as its row pages in, and still pending when no page contains it — which matters because that is the path the token's signature forces on Content Drive. `setPage` gets the size-change branch asserting page 1 and empty bookmarks.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/store/relationship-field.store.spec.ts
adrianjm-dotCMS: Added in e914ad01. `setData` after `loadMore` keeps `visibleCount` at 80 — same contract as the other two, and you are right that it is the path most likely to produce that report.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/store/relationship-field.store.spec.ts
adrianjm-dotCMS: Added in e914ad01: `loadMore` twice, then move item 90 to index 0 and assert it leads both `data()` and `$visibleItems()`. You are right that it works because the rendered rows are a prefix of `data` — which is exactly why it was worth pinning.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/components/footer/footer.component.spec.ts
adrianjm-dotCMS: Both covered in 834b6f13. The `confirmLabel` override now has a case asserting "Apply" through `DynamicDialogConfig`, plus the fallback. On the mock you flagged: you were right that nothing read `$selectedCount` — it turned out the whole `AddRelationshipsStore` injection in the footer was dead, so it is gone. `sort` has two cases now.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/add-relationships-filter-facade.ts
adrianjm-dotCMS: Copy-paste artefact of mine, dropped in e914ad01.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/add-relationships.component.html
adrianjm-dotCMS: Kept it, with the note you asked for (834b6f13). `text-[1.125rem]!` is the literal class the Asset Picker's own header uses for the same control, and these two headers do the same job in the same product — `text-lg` is the same value, but changing one alone separates them. The `!` is load-bearing: PrimeNG styles the button's icon slot by font-size and wins otherwise. Happy to change both together if you would rather they use the scale.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/add-relationships.store.ts
adrianjm-dotCMS: Renamed to `$totalItems` in e914ad01.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/add-relationships.store.ts
adrianjm-dotCMS: Agreed — fixed in 834b6f13. Still degrades rather than taking the results down, but now logs through `LoggerService` so a recurring outage is traceable instead of surfacing only as a blank column.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/add-relationships.store.ts
adrianjm-dotCMS: Fixed in 834b6f13. The two states are now distinct: an unbrowsable scope sets a flag and the dialog shows a note naming the scope as the problem, because no search term would help. The site chip alone was not signal enough — the table underneath said "try another search term".

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/add-relationships.store.ts
adrianjm-dotCMS: Not addressed yet — you are right on both. The casts do rest on `showFolders: false` plus a comment, which is a request-shape argument about data that arrives over HTTP, and a narrowing guard would make it hold regardless of what the endpoint returns. `getFilterValue` is the odd one out on return types. Will fix both.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/add-relationships.store.ts
adrianjm-dotCMS: Verified, and your reading is exact: `toggleSelection` does `delete` then `set`, so a re-checked row is re-inserted at the end while an untouched one keeps its seeded place. That was not a decision — insertion order was chosen for the accumulated set and this consequence was never considered.

It is a real surprise: the order is drag-reorderable and persisted, so unchecking and re-checking silently moves an item to the bottom of the relationship. I would rather restore the original position for a row that was already there when the dialog opened, and keep insertion order for genuinely new picks. Not in this PR yet — flagging so it is a decision rather than an accident.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/add-relationships.component.html
adrianjm-dotCMS: Correction — I argued the wrong way on this one. Changed to `text-lg!`: it is the same 1.125rem, so nothing renders differently, and it is the scale utility STYLING_STANDARDS asks for. The `!` stays because PrimeNG styles the button's icon slot by font-size and wins otherwise, and the comment now says that instead of defending the arbitrary value.

That leaves the Asset Picker's header on `text-[1.125rem]!` for the same control. Same value, so nothing breaks — but it should follow, and I would rather that be a separate change against that component than a drive-by here.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/add-relationships.store.ts
adrianjm-dotCMS: Fixed in 5e0f1bce. You were right that it was not intended — insertion order was chosen for the accumulated set and this followed from it unnoticed. The store now remembers the order the field opened with, so a row that was already related returns to its place; a pick made inside the dialog still lands at the end. Three cases cover it, including the one where a row is picked and unpicked within the dialog and has no seeded place to go back to.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-relationship-field/components/add-relationships/store/add-relationships.store.ts
adrianjm-dotCMS: Both fixed in 5e0f1bce. The casts are gone: narrowing now discriminates on the tag the folder type carries, so it holds whatever the endpoint returns rather than resting on `showFolders: false` plus a comment. `getFilterValue` has its return type.

oidacra (approved):

PR #37518, opened 2026-09-11
spec(content-drive): Title / All Fields search scope + literal-text search terms (#37479, #37532)

### LINKED ISSUE #37479: [TASK] Content Drive: add a Title / All Content scope selector to the search box
## Description

The Content Drive search box has one behavior: every term runs a **global, all-content** search. There is no way to say "I only care about the title". Authors who know the name of what they are looking for get back everything whose body, blocks or metadata happens to contain the word, and the row they wanted is buried.

Add a scope selector to the right of the search input, inside the same rounded box, with two options:

| Option | Behavior |
|---|---|
| **Title** | Match against the contentlet title (and folder/file names) only |
| **All Content** | Current behavior: title plus every indexed field |

The control is a small dropdown anchored to the search box, showing the active scope as its label and a check next to the selected option:

```
┌──────────────────────────────────────────────┬───────────┐
│  🔍  Search by title                         │  Title  ▲ │
└──────────────────────────────────────────────┴───────────┘
                                        ┌──────────────────┐
                                        │  Title        ✓  │
                                        │  All Content     │
                                        └──────────────────┘
```

The placeholder follows the scope ("Search by title" vs "Search all content"), so the box says what it will do before the user types.

This is scoped to Content Drive search (`POST /api/v1/drive/search`). It is not a rewrite of the search strategy, it is a switch that selects between the existing all-content query and a narrower title-only one.

## Why this also matters for performance

The all-content query is the expensive one, and today every keystroke pays for it. `GlobalSearchAttributeStrategy` builds a mandatory gate of:

```
+(catchall:<value>*^10 OR title_dotraw:*<value>*^2)
```

Two costs sit in that single line:

- **`catchall`** aggregates every field of the document (body, Story Block content, metadata). A common word matches a large slice of the index, so the term is cheap to look up and enormously expensive in what it returns.
- **`title_dotraw:*<value>*`** is a leading wildcard, which forces a scan across every distinct raw title in the index rather than a prefix seek.

In the drive path the ES hit set is not the end of the work. `BrowserAPIImpl` feeds those matches through DB hydration and per-chunk permission filtering (`BROWSER_CONTENT_CHUNK_SIZE`, default 900) before a page can be returned, so a broad match multiplies DB round trips and permission checks, not just ES time. A title-only query that drops the `catchall` clause and gates on a title prefix (`+title:<value>*`, keeping `title_dotraw` as a boost or fallback rather than as the gate) shrinks the candidate set at the source, and everything downstream gets cheaper with it.

So the scope selector is worth building twice over: it is the result quality users are asking for, and it gives them a fast path that avoids the most expensive clause in the query.

## Where to point

**Frontend**

| File | What changes |
|---|---|
| `core-web/libs/ui/src/lib/components/dot-search-input/dot-search-input.component.ts` / `.html` | Shared presentational box. Add the scope dropdown as an **optional** input (plus a `scopeChange` output) so the AssetPicker toolbar, which reuses this component, is untouched unless it opts in. |
| `core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-search-input/dot-content-drive-search-input.component.ts` | Store adapter. Opts the dropdown in, binds the current scope, forwards changes. |
| `.../store/dot-content-drive.store.ts` — `setGlobalSearch` (~line 228) | Carry the scope alongside the `title` filter; resetting pagination the same way. |
| `.../store/dot-content-drive.store.ts` — `$request` computed (~line 119) | Send the scope in the request payload next to `filters.text`. |
| `.../lib/utils/functions.ts` | Filter encode/decode, so the scope survives a URL restore and Back/Forward like every other filter. |
| `core-web/libs/dotcms-models/src/lib/dot-content-drive.model.ts` | `DotContentDriveSearchRequest` / filters model. |

**Backend**

| File | What changes |
|---|---|
| `dotCMS/src/main/java/com/dotcms/rest/api/v1/drive/AbstractQueryFilters.java` | Natural home for the new field, next to `text()` and `filterFolders()`. Default it to the current behavior so existing callers are unaffected. |
| `.../v1/drive/ContentDriveHelper.java` (~lines 179–183) | Where `filters.text()` is wired into `BrowserQuery.withFilter()` and `useElasticsearchFiltering(true)` is flipped on. The scope rides along here. |
| `dotCMS/src/main/java/com/dotcms/browser/BrowserQuery.java` | Carry the scope through to the API. |
| `dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java` — `buildBaseESQuery` (~line 1193) | The branch point: today it unconditionally calls `GlobalSearchAttributeStrategy`. Title mode selects the narrower query instead. Note `buildPureESQuery` (~line 663) still holds an older `title:'*x*'^5 OR catchall:*x*^3 OR fileName:*x*^2` shape — worth confirming whether the drive reaches it. |
| `.../v1/content/search/strategies/GlobalSearchAttributeStrategy.java` | The current all-content query. Leave it as the All Content path; add a title-scoped sibling rather than branching inside it. |

Related history: #36688 replaced the old broad `catchall:*kw*` leading wildcard with the current strategy for exactly these reasons, and #36814 tracks search performance at scale.

## Acceptance Criteria

- [ ] The Content Drive search box shows a scope dropdown with **Title** and **All Content**, the active scope as its label, and a check mark on the selected option
- [ ] Selecting a scope re-runs the current search immediately, with pagination reset to page 1
- [ ] The placeholder reflects the active scope
- [ ] **Title** mode returns only rows whose title (or folder/file name) matches; a term that appears only in a body or Story Block does not match
- [ ] **All Content** mode returns exactly what the search returns today (no regression)
- [ ] The Title-mode query contains no `catchall` clause and no leading-wildcard term as its mandatory gate
- [ ] The scope is encoded in the URL filters and survives reload and browser Back/Forward
- [ ] The backend field defaults to the current all-content behavior, so a request that omits it is unchanged
- [ ] The AssetPicker's search box is unaffected unless it explicitly opts in
- [ ] Postman coverage for `POST /api/v1/drive/search` in both scopes, including the omitted-field default
- [ ] Jest coverage for the dropdown, the store wiring and the URL round-trip
- [ ] A before/after latency comparison on a large dataset is recorded on the issue

## Open decisions

1. **Default scope.** The mock shows Title selected with a "Search by title" placeholder. Defaulting to Title gives everyone the fast path, but silently changes behavior for existing users. Defaulting to All Content is the no-regression choice. Needs a call before implementation.
2. **File names in Title mode.** For file assets, the file name is usually what the user means by "title". Should Title mode also cover `fileName` / `metadata.name`, or stay strictly on `title`?
3. **Stickiness.** Is the scope per-search (resets on reload beyond the URL), or remembered per user?
4. **Sorting.** Results are currently sorted by score descending whenever a term is present. Confirm that holds in Title mode.

## Additional Context

Behavior surfaced from the attached mock. Frontend and backend both change; the frontend cannot narrow the query on its own because the Lucene string is built server-side in `BrowserAPIImpl`.

KevinDavilaDotCMS:
**Spec-Kit PR 1 of 2.** Carries `spec.md` alone. Review it as a spec, not as code: is this the right problem, scoped right, with criteria a reviewer can tell "done" from? Approval here — not merge — unblocks `/speckit-plan`.

Refs #37479

## What it specifies

A scope dropdown beside the Content Drive search input, with two options — **Title** and **All Content** — so an author who knows the *name* of what they want can say so, instead of getting back every document whose body or Story Block happens to mention the word. It is also the cheap query path: the all-content gate is `+(catchall:<value>*^10 OR title_dotraw:*<value>*^2)`, an all-fields aggregate plus a leading wildcard, and in the drive a broad match multiplies DB hydration and permission filtering downstream, not just index time.

## The four open decisions, settled

The issue marks these as needing a call before implementation. Settled by the issue owner on 2026-09-11; **approving this PR is the record of that sign-off.**

| # | Decision | Settled as | Why |
|---|---|---|---|
| 1 | Default scope | **All Content** | No-regression. Doing nothing gives you exactly today's results; the fast path is opt-in. |
| 2 | Title-mode fields | **Contentlet `title` only** | Keeps the promise the label makes and the query on one field. File assets stay findable by name in practice — dotCMS keeps `title` in step with the file name. |
| 3 | Stickiness | **URL only, per search** | Behaves like every other Content Drive filter. No new preference storage. |
| 4 | Sorting | **Unchanged** | Void as asked — see correction 1 below. |

## Premises checked against `main` — four corrections

These are the part most worth a second pair of eyes, because three of them shrink the work and the fourth adds a rule the issue's file list has no place for.

1. **Nothing sorts by score today.** Decision 4 asks to confirm score-desc holds in Title mode. There is no score sort: `DEFAULT_SORT` is `modDate:DESC` (`shared/constants.ts:51`), the store sends it unconditionally (`dot-content-drive.store.ts:157`), and the server default is `SORT_BY = "modDate"` (`AbstractDriveRequestForm.java:81`). The only trace is a **stale comment** at `dot-content-drive.store.ts:481` sitting above code that does nothing of the kind. → Sorting out of scope; comment corrected in passing.

2. **Folders and links are already matched on name only, in both scopes.** They never reach Elasticsearch — loaded from the DB and narrowed in Java by substring (`BrowserAPIImpl.java:3026` for folders, `:2908-2913` for links). → The selector governs the **contentlet** clause alone; folder/link matching must not change.

3. **`buildPureESQuery` is unreachable for the drive under shipped configuration.** `doPureESQuery` runs only when `BROWSE_API_HEURISTIC_TYPE=PURE_ES`; the default is `HYBRID_SINGLE_CHUNKED_QUERY_ES` (`BrowserAPIImpl.java:701-709`). → Explicitly out of scope rather than changed unverified.

4. **A scope stored as a filter would offer "Clear all" on an unfiltered drive.** The issue routes the scope through filter state — which is what carries it into the URL — but stops there. That state also feeds the chip bar, and `hasNonDefaultFilters` (`utils/functions.ts:334-355`) counts **every** key but `sharedAssets` and `languageId` as a non-default filter, which is exactly what shows "Clear all" (`dot-filter-bar.component.html:7`). Picking **All Content** — the default — would light it up on a drive with nothing filtered. → **FR-021**: the scope counts as filter state only while it differs from the default, mirroring how the search term already deletes its own key when it goes empty.

## Scope boundaries

- **In**: the Content Drive search box, its filter/URL state, the drive's search request contract, and the title-scoped query.
- **Out**: `buildPureESQuery`; the all-content strategy shared with the Search portlet and the Relationships dialog, which keeps serving them unmodified — Title is a sibling path, not a branch inside it; the Asset Picker, which reuses the same search box and must stay untouched unless it opts in; sorting; any DB, index-mapping or content-model change.

## UI

Deliberately thin. The issue's ASCII diagram is the only design input — no mock image exists — so the spec fixes which components are on screen (input, scope dropdown beside it, placeholder that follows the scope) and stops there. No dimensions, spacing or layout rules.

## What a reviewer should push back on

- Decision 1 if you think Title-by-default is worth the behavior change for existing users.
- Decision 2 if excluding `fileName` / `metadata.name` will bite in practice — the spec names the gap (a file whose title was edited away from its name) and calls widening it a follow-up rather than a silent change.
- SC-003 carries no fixed latency threshold. No baseline exists yet; producing the dataset and the before/after comparison is part of the work and gets recorded on the issue.

This PR fixes: #37479

### DIFF
--- a/specs/37479-content-drive-search-scope/spec.md
+++ b/specs/37479-content-drive-search-scope/spec.md
@@ -1,0 +1,391 @@
+# Feature Specification: Content Drive — Title / All Content scope selector for the search box
+
+**Feature Branch**: `issue-37479-content-drive-search-scope`
+
+**Created**: 2026-09-11
+
+**Status**: Draft
+
+**Type**: New Feature
+
+**Related GitHub Issue**: [#37479](https://github.com/dotCMS/core/issues/37479) — related history: [#36688](https://github.com/dotCMS/core/issues/36688) (replaced the broad `catchall:*kw*` wildcard with the current strategy), [#36814](https://github.com/dotCMS/core/issues/36814) (search performance at scale)
+
+**Input**: User description: "Content Drive: add a Title / All Content scope selector to the search box. Default scope = All Content (no regression); Title mode matches strictly the contentlet title plus folder names, not fileName/metadata.name; scope persists only in the URL filters; sorting unchanged."
+
+---
+
+## Premise Corrections
+
+Four things the issue takes for granted do not survive verification against `main`. Three of them
+narrow the work; the fourth adds a rule the issue's file list has no place for.
+
+### 1. Results are **not** sorted by score when a term is present
+
+Open decision 4 asks to *"confirm that score-descending sorting holds in Title mode."* There is no
+score sorting to hold. The drive sends whatever sort the grid holds, and its default is
+modification date:
+
+- `core-web/.../dot-content-drive/portlet/src/lib/shared/constants.ts:51-54` — `DEFAULT_SORT = { field: 'modDate', order: DESC }`
+- `.../store/dot-content-drive.store.ts:157` — `sortBy: sort()?.field + ':' + sort()?.order`, unconditionally
+- `AbstractDriveRequestForm.java:81` — the server default is `SORT_BY = "modDate"` too
+
+The only trace of score sorting is a **stale comment** at `dot-content-drive.store.ts:481`
+("Since we are using scored search for the title we need to sort by score desc") sitting above code
+that does nothing of the kind. Nothing in `BrowserAPIImpl` or `ContentDriveHelper` substitutes a
+score sort when a filter is set.
+
+**Consequence for this feature**: open decision 4 is void. Sorting is out of scope entirely — both
+scopes keep sending the grid's sort, and neither introduces a scope-dependent sort rule. The stale
+comment should be corrected while the file is open (progressive enhancement), not acted upon.
+
+### 2. Folders and links are already matched on name only, in both scopes
+
+The issue frames "match against the contentlet title **(and folder/file names)** only" as new Title
+behavior. Folders and links never reach Elasticsearch at all — they are loaded from the database and
+narrowed in Java by a case-insensitive substring on their own name:
+
+- `BrowserAPIImpl.java:3026` — `folders.removeIf(f -> !f.getName().toLowerCase().contains(browserQuery.filter.toLowerCase()))`, gated on `filterFolderNames`
+- `BrowserAPIImpl.java:2908-2913` — the equivalent for links, on `Link::getTitle`, deliberately **not** gated on that flag
+
+**Consequence for this feature**: folder and link matching is scope-independent and must not change.
+The scope selector governs only the Elasticsearch clause used for **contentlets**. An acceptance
+criterion that reads "Title mode returns only rows whose title or folder name matches" is already
+half-true today and stays true in All Content mode.
+
+### 3. `buildPureESQuery` is unreachable for Content Drive under shipped configuration
+
+The issue flags the older `title:'*x*'^5 OR catchall:*x*^3 OR fileName:*x*^2` shape in
+`buildPureESQuery` (~line 610) as *"worth confirming whether the drive reaches it"*. It does not:
+`doPureESQuery` is selected only when `BROWSE_API_HEURISTIC_TYPE` is set to `PURE_ES`, and the
+shipped default is `HYBRID_SINGLE_CHUNKED_QUERY_ES` (`BrowserAPIImpl.java:701-709`), which routes to
+`buildBaseESQuery`.
+
+**Consequence for this feature**: `buildPureESQuery` is **out of scope**. Changing it would mean
+changing a code path no shipped configuration exercises, and doing so unverified is worse than
+leaving it consistent with its own heuristic.
+
+### 4. Storing the scope as a filter would offer "Clear all" on an unfiltered drive
+
+The issue's file list routes the scope through the drive's filter state — which is what carries it
+into the address — but stops there. That state also feeds the filter chip bar, and
+`hasNonDefaultFilters` (`utils/functions.ts:334-355`) treats **every** key other than `sharedAssets`
+and `languageId` as a non-default filter. That signal is exactly what shows the bar's "Clear all"
+(`dot-filter-bar.component.html:7`).
+
+So a scope written into the filters on every selection would light up "Clear all" the moment someone
+picked **All Content** — the default — on a drive where nothing is filtered at all.
+
+**Consequence for this feature**: FR-021. The scope counts as filter state only while it differs
+from the default, mirroring how the search term already deletes its own key when it goes empty
+(`dot-content-drive.store.ts:228-234`).
+
+---
+
+---
+
+## Decisions (settled 2026-09-11)
+
+The issue leaves four decisions open and marks them as needing a call before implementation. The
+issue owner settled them on 2026-09-11. **Approval of this spec (PR 1) is the record of that
+sign-off**; if any of them is reversed, the spec must be re-approved before `/speckit-plan` runs.
+
+| # | Decision | Settled as | Why |
+|---|---|---|---|
+| 1 | Default scope | **All Content** | The no-regression choice. Every user who does nothing keeps exactly the results they get today, and the fast path is opt-in. Defaulting to Title would silently change what an existing saved workflow returns. |
+| 2 | Field coverage in Title mode | **Contentlet `title` only** — not `fileName`, not `metadata.name` | Keeps the promise the label makes, and keeps the query to a single field so the cost argument holds. File assets are still reachable by name in practice (see [Assumptions](#assumptions)). |
+| 3 | Stickiness | **URL only, per search** | The scope behaves like every other Content Drive filter: it survives reload, Back and Forward, and a shared link, and resets to the default on a clean entry. No new per-user preference storage. |
+| 4 | Sorting in Title mode | **Unchanged** | Void as asked — see [Premise Correction 1](#1-results-are-not-sorted-by-score-when-a-term-is-present). Both scopes send the grid's current sort. |
+
+---
+
+## Problem Statement
+
+The Content Drive search box has exactly one behavior: every term runs a global, all-content search.
+An author who knows the **name** of what they are looking for has no way to say so. The term is
+matched against every indexed field of every document — body copy, Story Block content, metadata —
+so a common word returns a large slice of the drive and the one row the author wanted is buried
+among documents that merely mention it.
+
+The same breadth is also the expensive part of the request. The mandatory gate of the all-content
+query is `+(catchall:<value>*^10 OR title_dotraw:*<value>*^2)`
+(`GlobalSearchAttributeStrategy.java:38-40`): `catchall` aggregates every field of the document, so
+a common term is cheap to look up and enormous in what it returns, and `title_dotraw:*<value>*` is a
+leading wildcard, which forces a scan over every distinct raw title rather than a prefix seek.
+
+In the drive, the matched set is not the end of the work. Matches are fed through database hydration
+and per-chunk permission filtering (`BROWSER_CONTENT_CHUNK_SIZE`, default 900) before a page can be
+returned, so a broad match multiplies database round trips and permission checks — not just index
+time. Narrowing the candidate set at the source makes everything downstream cheaper with it.
+
+So the scope selector earns its place twice: it is the result quality authors are asking for, and it
+gives them a fast path that avoids the most expensive clause in the query.
+
+## UI Surface
+
+The ASCII diagram in the issue is the only mock — no image or design file accompanies it. What it
+establishes, and all this spec fixes, is which components are on screen:
+
+- The **search input** — the Content Drive search box as it exists today.
+- A **scope dropdown beside it**, offering **Title** and **All Content**. Its label is the active
+  scope; opening it marks the active option with a check.
+- The **placeholder** of the input, which follows the active scope.
+
+Nothing else about the control's appearance is fixed here. How the two sit together is an
+implementation decision.
+
+## User Scenarios & Testing *(mandatory)*
+
+### User Story 1 - Find a known item by its name (Priority: P1)
+
+An author knows the name of the item they want — a page called "Pricing", an image called "hero" —
+and types it into the Content Drive search box. Today they get back everything whose body or blocks
+happen to contain the word. They open the scope control next to the box, choose **Title**, and the
+list narrows to rows whose name actually matches. The placeholder changes to say *Search by title*,
+so the box states what it will do before they type again.
+
+**Why this priority**: This is the whole feature. It delivers the result quality the issue was
+raised for and the cheap query path on its own, with nothing else built. Stories 2 and 3 protect it;
+neither creates value without it.
+
+**Independent Test**: Fully testable by selecting **Title** with a term present and confirming that
+a document which contains the term only in its body or Story Block — and not in its name —
+disappears from the list, while the row whose name matches stays. Delivers the narrowing on its own.
+
+**Acceptance Scenarios**:
+
+1. **Given** the Content Drive is open with no search term, **When** the author looks at the search
+   box, **Then** a scope control is visible next to the input, reading **All Content**, and the
+   placeholder describes an all-content search.
+2. **Given** a term is present in **All Content** scope, **When** the author opens the scope control
+   and selects **Title**, **Then** the search re-runs immediately with the same term, results are
+   restricted to name matches, pagination returns to page 1, and the control reads **Title** with a
+   check mark beside that option.
+3. **Given** a document whose title does **not** contain the term but whose body or Story Block
+   does, **When** the scope is **Title**, **Then** that document is absent from the results.
+4. **Given** that same document, **When** the scope is **All Content**, **Then** it is present —
+   the all-content results are identical to what the drive returns today for that term.
+5. **Given** scope **Title** and a term that matches a folder's name, **When** the search runs,
+   **Then** the folder is listed, exactly as it is in **All Content** scope.
+6. **Given** the scope control is open, **When** the author selects the scope that is already
+   active, **Then** nothing is re-fetched and the current page is preserved.
+
+---
+
+### User Story 2 - The scope travels with the view (Priority: P2)
+
+An author narrows to **Title**, finds the row, opens it, and comes back with the browser Back
+button — the drive returns to the Title-scoped results, not to an all-content list. They copy the
+address and send it to a colleague, who opens the same narrowed view.
+
+**Why this priority**: Without it the scope silently resets on every reload and navigation, and an
+author who has narrowed their search loses that narrowing without being told. It is a correctness
+guarantee over Story 1 rather than a capability of its own, so it ranks below it.
+
+**Independent Test**: Fully testable by selecting **Title**, reloading the page, and confirming the
+control still reads **Title** and the results are still narrowed — then navigating away and back.
+
+**Acceptance Scenarios**:
+
+1. **Given** scope **Title** with a term, **When** the page is reloaded, **Then** the control reads
+   **Title**, the term is preserved, and the results are the Title-scoped results.
+2. **Given** the author switched from **All Content** to **Title**, **When** they press browser
+   Back, **Then** the view returns to the **All Content** results for that term.
+3. **Given** a Content Drive address carrying scope **Title**, **When** a different user opens it,
+   **Then** they see the same narrowed view, subject to their own permissions.
+4. **Given** the author enters Content Drive with no scope in the address, **When** the drive loads,
+   **Then** the scope is **All Content**.
+5. **Given** an address carrying an unrecognized scope value, **When** the drive loads, **Then** the
+   scope falls back to **All Content** and the drive loads normally, with no error surfaced.
+6. **Given** the author clears all filters, **When** the drive reloads its results, **Then** the
+   scope returns to **All Content** along with the other filter defaults.
+
+---
+
+### User Story 3 - Everything that is not Content Drive is untouched (Priority: P3)
+
+A developer using the Asset Picker, and an integration calling the Content Drive search endpoint,
+see no change at all. The Asset Picker's search box keeps the single all-content behavior it has
+today, with no scope control on screen. A request that does not mention the scope behaves exactly as
+it does now.
+
+**Why this priority**: It is a constraint on Stories 1 and 2 rather than a journey of its own, and
+it is verified by absence. It still has to be stated and tested, because the search box is shared
+and the endpoint is public.
+
+**Independent Test**: Fully testable by opening the Asset Picker and confirming no scope control
+appears and search behaves as before, and by replaying a stored Content Drive search request with no
+scope field and comparing the results to the current ones.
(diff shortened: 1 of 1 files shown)


specs/37479-content-drive-search-scope/spec.md
```diff
@@ -0,0 +1,391 @@
...
+
+**Created**: 2026-09-11
+
+**Status**: Draft
+
+**Type**: New Feature
+
+**Related GitHub Issue**: [#37479](https://github.com/dotCMS/core/issues/37479) — related history: [#36688](https://github.com/dotCMS/core/issues/36688) (replaced the broad `catchall:*kw*` wildcard with the current strategy), [#36814](https://github.com/dotCMS/core/issues/36814) (search performance at scale)
+
+**Input**: User description: "Content Drive: add a Title / All Content scope selector to the search box. Default scope = All Content (no regression); Title mode matches strictly the contentlet title plus folder names, not fileName/metadata.name; scope persists only in the URL filters; sorting unchanged."
```
zJaaal: Is not all content, is all metadata, 2 different things, the all content im covering it on the PR i have open https://github.com/dotCMS/core/pull/37487

specs/37479-content-drive-search-scope/spec.md
```diff
@@ -0,0 +1,391 @@
...
+4. **Given** the author enters Content Drive with no scope in the address, **When** the drive loads,
+   **Then** the scope is **All Content**.
+5. **Given** an address carrying an unrecognized scope value, **When** the drive loads, **Then** the
+   scope falls back to **All Content** and the drive loads normally, with no error surfaced.
+6. **Given** the author clears all filters, **When** the drive reloads its results, **Then** the
+   scope returns to **All Content** along with the other filter defaults.
+
+---
+
+### User Story 3 - Everything that is not Content Drive is untouched (Priority: P3)
```
zJaaal, line 364: Check if the asset picker is using the same endpoint, if they are, we need to support it

zJaaal (changes requested):

specs/37479-content-drive-search-scope/spec.md
zJaaal: With metadata I mean fields, we might need a tooltip explaining this, because is a new feature

specs/37479-content-drive-search-scope/spec.md
zJaaal: Answering my own question, from the blast-radius sweep I did for #37487: **yes, same endpoint.**

The Asset Picker builds a `DotContentDriveSearchRequest` in `with-asset-browse.feature.ts` and calls `DotContentDriveService.search()`, the same `POST /drive/search` Content Drive uses. Those two are the only callers of that service, so the picker is the one surface that inherits whatever the server does with a request that never mentions the search scope.

The good news is your spec already carries the defence. The requirement that a request omitting the scope behaves exactly as today, plus the scenario pinning an omitted request against an explicit all-content one, covers the Asset Picker without touching a line of Asset Picker code, as long as the default is all-content and stays that way.

Two things I would still add. Name the Asset Picker explicitly in that requirement rather than leaving it as "other callers", so a later change to the default has to confront it instead of only re-running Content Drive's tests. And if the implementation ends up touching the shared query building underneath the endpoint rather than only the search-text branch, the blast radius widens past those two callers to the assets API, the old file browser, the legacy admin browser and a Velocity viewtool, which all reach the same listing by a different door. Staying inside the search-text branch keeps it at two.

*Reply written by Claude on @zJaaal's behalf.*

zJaaal:
One naming ask, because our two specs are about to collide in prose and then in the same request object.

This one introduces a scope that says **which fields a search reads**. #37487 introduces one that says **where in the site you are browsing**: All, the site root, or System Host. Bare "scope" now names either, and a reader landing on either spec cannot tell which is meant. Worse, both end up as fields on the same Content Drive search request, so the ambiguity outlives the specs.

I have already renamed mine to spell **browse scope** everywhere, with a note in Key Entities pointing at yours so the two cannot be conflated. Could you do the matching half here: say **search scope** in the prose rather than plain "scope", and name the field **`searchScope`** on the wire?

Then the pair reads unambiguously wherever they meet: `searchScope` says how a search reads what is there, `browseScope` says where "there" is. Neither name is load-bearing on its own, but together they stop a future reader guessing.

Unrelated to the naming, and only because it touches the same sentence I was reading: your spec's "All Content" label is the one I flagged above as really meaning all *fields*. If that wording changes, the field values change with it, so it is worth settling before the name goes on the wire.

*Comment written by Claude on @zJaaal's behalf.*

zJaaal:
One more on the contract, following the naming note above: I think the search scope belongs **inside `filters`**, next to `text`, rather than as another top-level field.

`AbstractQueryFilters` today is `{ text, filterFolders }`, and `filterFolders`'s own Javadoc says "when text is provided". Both members exist to qualify the text search, which is exactly what the search scope does: it says which fields `text` reads, and it means nothing without `text`. Putting it beside `text` makes them travel together, and a request carrying a scope with no text becomes visibly nonsense instead of a validation rule someone has to remember later. So `filters.searchScope`.

To be clear that this is not a general "scopes go in filters" rule: the **browse** scope in #37487 stays at the top level, for two reasons. It qualifies `assetPath` rather than `text`. And the filter bar's "Clear all" resets the filters object, so a browse scope living there would let clearing your filters navigate you out of System Host. A filter chip narrows what you see where you are; it must not be able to move you somewhere else.

Related, since it came up while looking at this: we also considered letting the path alone carry the browse scope, with no field at all, mirroring how the Content Drive URL now encodes it in one value. It does not survive contact with this endpoint, and for a reason that matters to your spec too: `assetPath: "//site/"` **already means** "the whole site at any depth", and the Asset Picker sends exactly that. Redefining it would silently change what the picker asks for. The general rule both our features are leaning on is the same one, that an existing payload has to keep meaning what it means today, and it constrains where new meanings can be expressed.

Worth flagging so nobody invokes consistency in either direction: `filters` is not currently the home for filters. `contentTypes`, `baseTypes`, `language`, `workflow`, `status` and `userSearchable` all sit at the top level and all of them genuinely *are* chips in the filter bar. Moving those six in is only part of a real cleanup though, so I would not treat it as the benchmark. The contract also lets a caller say one thing twice, `archived: true` and `status: ["ARCHIVED"]`, which `ContentDriveHelper` then has to reconcile, and it carries `live` as a boolean when it actually selects which *version* you get rather than filtering anything. That is a contract PR of its own, worth doing while this endpoint still has exactly two callers, and not something either of our feature PRs should take on.

*Comment written by Claude on @zJaaal's behalf.*

ihoffmann-dot (changes requested):
**Mid: possible conflict with ADR-0018 (Database-First Search for Content Drive)**

ADR-0018's routing table states explicitly:

> Title → **DB ∪ Index** (`contentlet.title` column ∪ index records)

This union exists specifically to preserve read-your-writes: a just-saved/renamed item that the index hasn't caught up to yet must still be findable via the DB `title` column.

FR-008/FR-010 define **Title** scope purely as a narrower Elasticsearch query (drop `catchall`, drop the leading wildcard, gate on `title`). Nothing in the FR/SC list references consulting the `contentlet.title` DB column, and no acceptance scenario or success criterion covers "a document created/renamed just before search, not yet indexed, searched by title."

The ADR does note the DB ∪ Index union is gated on a separate title-persistence effort and may not be fully implemented in the current codebase either, so this may already be a pre-existing gap rather than something this feature introduces. It's worth an explicit call before `/speckit-plan`

No other findings. No other ADRs in `platform-adrs` were found that this spec contradicts.

specs/37479-content-drive-search-scope/spec.md
KevinDavilaDotCMS: Agreed, and renamed. The wide option is now **All Fields**, values `TITLE` / `ALL_FIELDS`.

Your framing is what settled it: this scope does not widen *which content* is searched, it widens *which fields of each document* the term is read against. "All Content" made a claim about breadth of content — which is exactly what your **All** browse scope genuinely means — so the two would have put different meanings of the same word on the same request object. The issue's original wording stays recorded verbatim in **Input**, with a note pointing at Review Decision 7 for why it was rejected.

The tooltip is in as **FR-022**: the control must make available a short explanation distinguishing the item's name from anything written anywhere in the item, and the two labels explicitly may not be relied on to carry that distinction alone. Left as a requirement rather than a design — tooltip, helper text or per-option description is an implementation call.

specs/37479-content-drive-search-scope/spec.md
KevinDavilaDotCMS: Both taken, thank you — this was the most useful note on the PR.

**Asset Picker named explicitly.** FR-017 no longer hides it behind "other callers". It now carries the call site (`with-asset-browse.feature.ts` → `DotContentDriveService.search()`, the same `POST /drive/search`) and states the consequence you were after: any future change to the default has to confront the Asset Picker by name, not merely re-run Content Drive's tests. Verified both files exist on `main`.

**Blast radius pinned.** New **FR-024** confines the change to the text-search branch of the shared listing and names the three other doors I confirmed reach it — `WebAssetHelper` (assets REST API), `BrowserAjax` (legacy admin browser), `DotCMSMacroWebAPI` (Velocity viewtool). Stated as you put it: staying in the search-text branch keeps it at two callers, widening into shared query building takes it to six. **SC-008** measures it, and Legacy Considerations records the two-plus-three split.

Also added your contract-debt paragraph to Legacy Considerations as debt **deliberately not taken on** — `archived` vs `status: ["ARCHIVED"]`, `live` selecting a version rather than filtering, the six top-level fields that genuinely are chips. Worth doing while the endpoint has two callers; not for either of our feature PRs to carry.

KevinDavilaDotCMS:
@zJaaal — naming and placement both taken, in full.

**Naming.** "Search scope" everywhere in prose, never bare "scope", and `searchScope` on the wire. Recorded as Review Decision 5 with your reasoning, and Key Entities now points at the browse scope the way yours points here, so the pair reads unambiguously from either side: `searchScope` says how a search reads what is there, `browseScope` says where "there" is.

**Placement — `filters.searchScope`, agreed.** I checked `AbstractQueryFilters` and it is exactly as you describe: `{ text, filterFolders }`, with `filterFolders`' own Javadoc reading "When text is provided". Both members exist to qualify the text search, and the search scope is a third of the same kind — it says which fields `text` reads and means nothing without `text`. Beside `text` they travel together. I took your "visibly nonsense rather than a validation rule someone has to remember" one step further and wrote it down as **FR-025**: a request carrying a search scope with no text is a contract error, not a silently-ignored field.

Your asymmetry argument is what convinced me the two placements are not inconsistent: the search scope *should* be cleared by "Clear all" — that is FR-020 — whereas a browse scope living in `filters` would let clearing your filters navigate you out of System Host. Same rule, opposite outcome, because one narrows what you see where you are and the other moves you. Review Decision 6 records both halves so nobody later reads this as "scopes go in filters".

Noted on `assetPath: "//site/"` already meaning "the whole site at any depth" and the picker sending exactly that. The general rule is the one this spec already leans on throughout — an existing payload keeps meaning what it means today — and it is why the search scope is additive and optional at every layer rather than a new reading of anything that already exists.

---

@ihoffmann-dot — ADR-0018 gets its own section and an explicit call rather than silence. Short version: **deferred, with reasons**, and I think the evidence is a little stronger than "possibly a pre-existing gap".

1. **The ADR defers its own union.** It states `contentlet.title` "is **not reliably populated**", that the display title is derived and `Contentlet.title()` is `@Nullable`, and that "populating `contentlet.title` reliably is a known gap to be addressed in a separate issue" — the ADR fixes the routing contract that will *consume* the column, not a claim that it is usable today.

2. **No text search consults it today, in either scope.** This is the part I verified rather than assumed, and it became **Premise Correction 5**. The browsing service has two text paths: Elasticsearch, or a SQL predicate running `contentlet_as_json::text ILIKE '%token%'` when `useElasticsearchFiltering` is false (`BrowserAPIImpl.java:2053-2060`, `appendFilterQuery` at `:2222`). The builder default is `false`, so the SQL path is the general fallback — but Content Drive never takes it: `ContentDriveHelper.java:180-184` sets `useElasticsearchFiltering(true)` unconditionally whenever the request carries text. So contentlet text matching is index-only in both scopes, and the exposure to index lag is identical in All Fields and would be identical in Title. Narrowing the clause does not deepen it — an item missing from the index is missing from both scopes equally.

3. **The union arrives for free.** When the title-persistence work lands and the text path becomes `DB-title ∪ index`, Title scope is the scope that benefits most directly, because the column it unions in *is* the title. **FR-026** makes that a requirement: Title scope must not foreclose the union, and adding the DB column later must be additive rather than a rewrite.

There is also a new edge case covering the scenario you said nothing covered — an item saved or renamed moments before the search, not yet indexed — stating plainly that it is missing in **both** scopes, as it is today.

Premise Correction 5 was worth the trip on its own: it rules out a second, SQL-shaped text match that a Title scope could have silently failed to narrow, which would have been a real hole in FR-010.

---

Pushed as `77abd40b77`. PR title updated to match the rename. Both of you please re-review.

zJaaal (dismissed):

KevinDavilaDotCMS:
## Scope change: #37532 folded in — please re-review

Pushed `1ce8cdd1bf`. The spec now resolves **two** issues. @zJaaal your approval predates this, so it needs another look.

[#37532](https://github.com/dotCMS/core/issues/37532) (High, customer ticket 39185) reports that field filters return "no results found" for values containing Lucene query-syntax characters — `ABC Bank (XETRA: DBKGn.DB / NYSE: DB)` and similar. It was raised against the **Content Search portlet**, so at first glance it looked unrelated to this spec. It is not.

**The same defect is in Content Drive's search box**, in the exact clause FR-010 rewrites:

| Defect | Content Drive | Evidence |
|---|---|---|
| Gate built from unescaped input | **yes** | `GlobalSearchAttributeStrategy:37-38` uses the raw `value`; only `:46-48` escapes, and only its own clause |
| `/` missing from the escape set | **yes** | private `SPECIAL_CHARS_TO_ESCAPE` at `:20` — this is #37532's fifth AC verbatim |
| Consecutive separators emit empty clauses | **yes** | `:40-45` splits with no empty-token filter → a term-less `title:^5` |
| Parse failure swallowed → silent zero results | **yes** | `BrowserAPIImpl:893-895`, structurally identical to the `ContentletAjax:1058` the issue names |

Nothing sanitizes upstream — `ContentDriveHelper:183` passes `filters().text()` straight through. So implementing #37479 alone would have meant rewriting the buggy clause and leaving the bug in it.

The issue also asks for this directly: *"Content Drive is intended to replace the Content Search portlet, so the improved field-level search behaviour should land there."*

### What changed in the spec

- **Premise Correction 6** — the evidence above, with the point that the spec's existing edge case ("a term is never allowed to alter the structure of the query") was aspirational, not a property the search box has.
- **Why #37532 lands here** — a per-AC table of what this spec delivers and what it does not.
- **User Story 4** (P1, listed fourth) — the customer's headline, in both scopes.
- **FR-027 to FR-031, SC-009 to SC-012** — literal matching across the full reserved set via the shared `LuceneQueryUtils.escape` rather than a strategy-private set, no empty clauses, error state instead of a silent empty list, and field-filter parity confirmed by test.

### Two things I want you to push back on if you disagree

**1. FR-009 and SC-002 had to yield.** Escaping the gate *changes* All Fields results for affected terms. That contradicts the no-regression promise you approved, so rather than smuggle it in I wrote the carve-out down and bounded it: reserved characters and consecutive separators only, every other term byte-identical, and the tests that change enumerated in the PR. If you'd rather the escaping ride in its own PR to keep FR-009 absolute, say so — it's a defensible call, it just costs a second trip through this code.

**2. The Content Search portlet is NOT fixed.** `ContentletAjax.searchContentletsByUser()` keeps its current behaviour. The issue frames the legacy construction as something to replace rather than patch, which is the direction I followed — but the customer on ticket 39185 is using the portlet **today**, not Content Drive. Closing #37532 on this work is only honest if the team accepts that. I flagged it in Legacy Considerations and in Assumptions rather than assume it; it's the issue owner's call, not the spec's.

Also worth noting: the fix lands in `GlobalSearchAttributeStrategy` itself, which the Search portlet and the Relationships dialog also use, so they stop mis-parsing reserved characters too. That's deliberate (**FR-031**) and it's the one place this feature intentionally changes behaviour beyond Content Drive — unlike the Title scope, which stays a sibling path outside the shared strategy. Calling it out so it doesn't read as scope creep.

One thing that was already fine: Content Drive's **field** filters escape correctly today via `TextFieldStrategy:45-46,53-54`, so that AC is verification rather than implementation (FR-030 / SC-012).

zJaaal (dismissed):

ihoffmann-dot (dismissed):

zJaaal (approved):

PR #37519, opened 2026-09-11
spec(uve): Edit VTL menu opens empty after the selection is re-anchored

### LINKED ISSUE #37499: UVE: "Edit VTL" menu opens empty after the selected contentlet is re-anchored
## Problem Statement

In the Universal Visual Editor, the **Edit VTL** (`</>`) button on the contentlet hover toolbar intermittently opens an **empty menu** — no VTL files are listed, and nothing happens. A customer reported it as *"Sometimes the edit VTL on the UVE is not responding. Happens from time to time."* They were using a `VtlInclude` content type.

This is a **regression**. It affects any user who edits VTL from UVE on a page where the iframe layout changes after selecting a contentlet — which is most pages.

**Root cause — the button and the menu read two different contentlets:**

| Concern | Source | Location |
|---|---|---|
| Is the `</>` button shown? | **Hovered** contentlet | `dot-uve-contentlet-tools.component.ts:213` — `hasVtlFiles()` reads `contentContext()` |
| What files does the menu list? | **Selected** contentlet, when one exists | `dot-uve-contentlet-tools.component.ts:331-338` — `vtlMenuItems()` does `this.selected() ? this.selectedContentContext() : this.contentContext()` |

The selected contentlet's payload then loses `vtlFiles` on every re-anchor:

- `applyBoundsForSelection()` rebuilds `editorSelected.payload` from the SDK bounds payload via `getPageSavePayload()` — `withSelectionAnchor.ts:130-137`, `withEditor.ts:437`
- That bounds payload structurally has no `vtlFiles` — `getDotCMSContentletsBound()` emits only `{container, contentlet}` — `libs/sdk/uve/src/lib/dom/dom.utils.ts:84-110`
- `vtlMenuItems()` has no `?? []`, so it returns `undefined` and PrimeNG renders an empty popup rather than failing loudly

**Why it's intermittent:** `SET_BOUNDS` arrives from the SDK `AUTO_BOUNDS` channel, which observes *all* layout change inside the iframe — scroll, resize, image/font load shifts, container mount/unmount — on a debounce trailing edge (`libs/sdk/uve/src/script/utils.ts:137-145`). A fresh click with no layout change works. Any scroll or resize after selecting breaks it.

**Why the customer's workaround works:** Quick Edit calls `promoteHoverToSelected()`, which writes `payload: this.contentContext()` — the hover context, which still carries `vtlFiles` — into `editorSelected`.

**Note:** the `</>` button exists *only* in the hover toolbar (`component.html:66-79`). `selectedHasVtlFiles` (`:233`) is dead code, never referenced in the template.

**Regression origin:** `5431c635f8` — *Refactor(UVE): Real-Time Canvas* (#34173), 2026-04-14 — changed `vtlMenuItems` from hover-only to selected-preferred while `hasVtlFiles` stayed hover-only. Later touched by `dfdeb9a8db` (#35539).

Backend is behaving correctly — `DotParse.java:59` emits the `vtl-file` marker as expected. No backend or SDK change is required.

## Steps to Reproduce

**Setup** — a widget that renders a VTL file via `#dotParse` (the `VtlInclude` content type in the demo starter does this out of the box). To make the timing deterministic, the included `.vtl` self-resizes on a timer so `SET_BOUNDS` fires without user input:

```html
<div id="uve-repro-grow" style="height:40px;background:#fce8e6;"></div>
<script>
(function () {
    if (window.__uveVtlReproRunning) { return; }
    window.__uveVtlReproRunning = true;
    var box = document.getElementById('uve-repro-grow');
    var big = false;
    setInterval(function () {
        big = !big;
        box.style.height = big ? '180px' : '40px';
    }, 2000);
})();
</script>
```

**Reproduce:**

1. Add the widget to a container on a page and open that page in UVE, Edit mode.
2. **Click** the widget once so it becomes the selected contentlet.
3. Wait ~4 seconds — the block resizes, `SET_BOUNDS` fires, the selection is re-anchored.
4. **Hover** the widget. The `</>` Edit VTL button appears.
5. **Click** the `</>` button.

**Expected:** the menu lists the widget's VTL file.
**Actual:** the menu opens empty.

**Customer workaround (confirms the diagnosis):** click Quick Edit (bolt), then click `</>` again — the file now appears.

**Negative control (confirms the re-anchor is the cause):** reload, hover the widget **without ever clicking it**, click `</>` — the menu populates correctly, because `selected()` is `null` and `vtlMenuItems()` falls back to the hover context.

**Without the self-resizing widget** — the customer's original path: put the widget on a page tall enough to scroll, click it, scroll down and back up, hover, click `</>`.

## Acceptance Criteria

**Happy path**

- [ ] With a contentlet selected and the selection re-anchored by a `SET_BOUNDS` event, clicking `</>` on that contentlet's hover toolbar lists its VTL files.
- [ ] Clicking a file in the menu still opens the VTL in the editor dialog, unchanged.
- [ ] Hovering a contentlet with no prior selection lists its VTL files (existing behavior does not regress).

**Sad path / edge cases**

- [ ] `vtlMenuItems()` returns `[]` rather than `undefined` when the contentlet has no VTL files, so PrimeNG is never handed an undefined model.
- [ ] Whatever gates the `</>` button reads the same contentlet the menu is populated from — hovering contentlet A while contentlet B is selected must never show A's button with B's files.
- [ ] Scrolling, resizing the canvas, or switching device preview after selecting does not change which files the menu lists.
- [ ] A contentlet with multiple `#dotParse` includes lists all of them.

**Store hardening**

- [ ] `applyBoundsForSelection()` carries `vtlFiles` forward from the previous selection when re-anchoring, so `editorSelected.payload` no longer silently loses data the bounds payload cannot supply.

**Cleanup**

- [ ] Dead `selectedHasVtlFiles` computed (`:233`) is removed, or wired up if a selected-toolbar VTL button is intended.

**Tests**

- [ ] `dot-uve-contentlet-tools.component.spec.ts` covers: selection re-anchored → hover → menu still lists the files.
- [ ] `withSelectionAnchor.spec.ts` covers: re-anchor preserves `vtlFiles` on the selected payload.

## dotCMS Version

Latest from `main`. Regression introduced 2026-04-14 by #34173 (`5431c635f8`); first shipped in release `26.04.20-01` and present in every release since. Reproduced locally against `main` on the demo starter site.

## Severity

## Links

NA

rjvelazco:
### Proposed Changes

**PR 1 of 2 — specification only. No implementation code.** Approve the spec and I'll open PR 2 with the fix and tests.

Fixes #37499 (spec; the implementation PR will close it).

* Adds `specs/37499-uve-edit-vtl-empty-menu/spec.md` — one file, nothing else.

---

### The defect

In UVE, the **Edit VTL** (`</>`) button on the contentlet hover toolbar opens an **empty menu** once the contentlet has been selected and a layout shift re-anchors the selection. To the user, the button simply doesn't respond. A customer reported it as *"Sometimes the edit VTL on the UVE is not responding. Happens from time to time."*

It's a regression from #34173 (`5431c635f8`, *Refactor(UVE): Real-Time Canvas*, 2026-04-14), first shipped in `26.04.20-01`.

**Two defects combine** — both verified against current `main`:

| # | Defect | Where |
|---|---|---|
| 1 | The button and the menu read **different contentlets** — `hasVtlFiles()` gates on the *hovered* one, `vtlMenuItems()` prefers the *selected* one | `dot-uve-contentlet-tools.component.ts:213` and `:331-338` |
| 2 | The selected payload **loses `vtlFiles` on every re-anchor** — `applyBoundsForSelection()` rebuilds it from the SDK bounds payload, which structurally cannot carry `vtlFiles` | `withSelectionAnchor.ts:130-137`, `dom.utils.ts:85-110` |

The failure is silent: `vtlMenuItems()` returns `undefined` (allowed because `tsconfig.base.json` sets `"strict": false`) and PrimeNG renders an empty popup rather than failing.

**Why it looks intermittent:** `SET_BOUNDS` fires from the SDK's `AUTO_BOUNDS` channel on *any* layout change inside the iframe — scroll, resize, late image load — on a debounce trailing edge.

Frontend only. Backend is correct; `DotParse.java` emits the `vtl-file` marker as expected. **No backend and no SDK change.**

---

### What a reviewer should focus on

Three decisions came out of `/speckit-clarify`. **Two of them deliberately narrow the work** — please confirm you agree rather than read them as oversights:

1. **The menu follows the hovered contentlet.** `vtlMenuItems()` goes back to hover-only, matching `hasVtlFiles()`. The `</>` button's visibility rule doesn't change. If you think the contract should be the *selected* contentlet instead, the fix direction inverts — say so now.
2. **Carry-forward is `vtlFiles` only, and no follow-up issue is filed.** `getDotCMSContentletsBound()` also drops `baseType`, `onNumberOfPages` and `dotStyleProperties`, so the same class of silent loss survives this fix, untracked. Recorded as an explicit non-goal. The style editor already carries its own rollback-based workaround for the `dotStyleProperties` loss.
3. **No E2E.** AC-002, AC-006 and AC-007 stay manual. The accepted consequence is written into the spec: the SDK→store seam is exercised only by simulation, so a regression in the bounds payload shape wouldn't be caught.

---

### Checklist

- [x] Tests — 10 acceptance criteria mapped to two spec files (`dot-uve-contentlet-tools.component.spec.ts`, `withSelectionAnchor.spec.ts`), with the Principle V justification for the layers not covered. No test code in this PR; TDD gates run in PR 2.
- [ ] Translations — n/a, no new or changed UI strings.
- [x] Security Implications Contemplated — none. No REST contract, DB schema, ES mapping, serialized state or public SDK surface. Not rollback-unsafe. The defect is transient client-side state, so there is no bad persisted data to repair.

### Additional Info

Spec-Kit flow, per [CLAUDE.md](https://github.com/dotCMS/core/blob/main/CLAUDE.md) and [Spec-Kit Quick Start §3](https://github.com/dotCMS/core/blob/main/docs/core/SPEC_KIT_QUICK_START.md). Sized **Tier 1 (Standard)**: changes observable behavior, no new or changed interface, not cross-team or security-sensitive.

The gate is **approval, not merge** — once this is approved I'll run `/speckit-plan` and branch PR 2 off this branch.

### Screenshots

n/a — specification only.

This PR fixes: #37499

### DIFF
--- a/specs/37499-uve-edit-vtl-empty-menu/spec.md
+++ b/specs/37499-uve-edit-vtl-empty-menu/spec.md
@@ -1,0 +1,287 @@
+# Issue Resolution Specification: UVE "Edit VTL" menu opens empty after the selected contentlet is re-anchored
+
+**Feature Branch**: `issue-37499-uve-edit-vtl-empty-menu`
+
+**Created**: 2026-09-11
+
+**Status**: Draft
+
+**Type**: Issue / Bug Resolution
+
+**Related GitHub Issue**: [#37499](https://github.com/dotCMS/core/issues/37499)
+
+**Input**: User description: "In the Universal Visual Editor, the Edit VTL (`</>`) button on the contentlet hover toolbar intermittently opens an empty menu; no VTL files are listed, and nothing happens. A customer reported it as *\"Sometimes the edit VTL on the UVE is not responding. Happens from time to time.\"* They were using a `VtlInclude` content type."
+
+## Problem Statement *(mandatory)*
+
+In the Universal Visual Editor (UVE), the **Edit VTL** (`</>`) button on a contentlet's hover
+toolbar opens a menu that is supposed to list the VTL files that contentlet renders. After the
+user has clicked a contentlet at least once, that menu intermittently opens **empty** — no file
+entries, no error, nothing to click. To the user the button simply does not respond.
+
+The button itself is still shown, because the decision to show it and the decision about what
+to put in the menu are made from two different contentlets. The user is offered an action that
+has already lost the data it needs.
+
+**Severity / Impact**:
+
+- **Severity**: Medium — editing VTL from UVE is blocked; no data is lost or corrupted.
+- **Who**: Any user editing VTL from UVE on a page whose iframe layout shifts after a
+  contentlet is selected — scroll, resize, device-preview switch, or late image/font load. In
+  practice, most pages.
+- **How often**: Intermittent by user perception, but deterministic given the trigger: it
+  happens on every hover that follows a layout shift after a selection.
+- **Workaround exists**: Clicking Quick Edit (bolt) before `</>` repopulates the menu. The
+  customer found this on their own; it is not discoverable.
+- **Regression**: Introduced 2026-04-14 by #34173 (`5431c635f8`, *Refactor(UVE): Real-Time
+  Canvas*), first shipped in `26.04.20-01`, present in every release since. Later touched by
+  `dfdeb9a8db` (#35539).
+
+## Clarifications
+
+### Session 2026-09-11
+
+- Q: The `</>` button is hover-only, but `vtlMenuItems()` prefers the selected contentlet. Which contentlet should the menu list files from? → A: The **hovered** contentlet — `vtlMenuItems()` becomes hover-only, matching `hasVtlFiles()`, restoring the pre-#34173 contract.
+- Q: The bounds payload also drops `baseType`, `onNumberOfPages` and `dotStyleProperties`. How wide should AC-009's carry-forward be? → A: **`vtlFiles` only.** The other dropped fields are out of scope for this fix and no follow-up is filed.
+- Q: Which test layers does this fix get, given Principle V? → A: **Component/store specs only.** No E2E; AC-002, AC-006 and AC-007 stay manual, with the Principle V justification recorded below.
+
+## Reproduction *(mandatory)*
+
+**Environment**: dotCMS `main` (reproduced locally); demo starter site; UVE in Edit mode;
+any browser. Requires a widget that renders a VTL file via `#dotParse` — the `VtlInclude`
+content type in the demo starter does this out of the box.
+
+**Steps to Reproduce**:
+
+To make the timing deterministic, give the included `.vtl` a block that self-resizes on a
+timer, so a layout change fires without user input:
+
+```html
+<div id="uve-repro-grow" style="height:40px;background:#fce8e6;"></div>
+<script>
+(function () {
+    if (window.__uveVtlReproRunning) { return; }
+    window.__uveVtlReproRunning = true;
+    var box = document.getElementById('uve-repro-grow');
+    var big = false;
+    setInterval(function () {
+        big = !big;
+        box.style.height = big ? '180px' : '40px';
+    }, 2000);
+})();
+</script>
+```
+
+1. Add the widget to a container on a page and open that page in UVE, Edit mode.
+2. **Click** the widget once so it becomes the selected contentlet.
+3. Wait ~4 seconds — the block resizes and the selection is re-anchored to the new bounds.
+4. **Hover** the widget. The `</>` Edit VTL button appears.
+5. **Click** the `</>` button.
+
+**Without the self-resizing widget** (the customer's original path): put the widget on a page
+tall enough to scroll, click it, scroll down and back up, hover, click `</>`.
+
+**Expected Behavior**: The menu lists the widget's VTL file(s); clicking one opens it in the
+VTL editor dialog.
+
+**Actual Behavior**: The menu opens empty — an empty popup panel with no items. No console
+error, no toast, no network call.
+
+**Reproducibility**:
+
+- **Always**, once the state is reached: a contentlet must have been selected, and a layout
+  change must have re-anchored that selection afterwards.
+- **Never** before the first click: with nothing selected, the menu populates correctly. This
+  is the negative control — reload, hover the widget without ever clicking it, click `</>`,
+  and the file is listed.
+- Perceived as intermittent only because the triggering layout change is usually incidental
+  (a scroll, a late-loading image) rather than deliberate.
+
+## Scope of Investigation *(mandatory)*
+
+- **Affected area**: Universal Visual Editor — contentlet hover toolbar and the editor store's
+  selection anchoring. Specifically the `edit-ema` portlet in the `core-web` Nx workspace.
+- **Suspected surface**: **Frontend only** (Angular/TypeScript under
+  `core-web/libs/portlets/edit-ema/portlet/`). The modern-vs-legacy Java distinction does not
+  apply here — no Java is involved. Backend is behaving correctly: `DotParse.java` emits the
+  `vtl-file` marker as expected, and the page API returns the VTL file list. **No backend and
+  no SDK change is required.**
+  - Toolbar component:
+    `core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-contentlet-tools/`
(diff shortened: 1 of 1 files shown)


rjvelazco:
> 🤖 Reply written by **Claude Code** on @rjvelazco's behalf, not by Rafael directly.

Good catch on AC-009 — it holds up. I re-verified it independently rather than taking it on trust, and the finding is correct.

**What I checked:**

* `grep -rn "vtlFiles"` across all of `core-web`, excluding specs and coverage, returns exactly four non-spec hits: `hasVtlFiles` (`:213`), `selectedHasVtlFiles` (`:234`), `vtlMenuItems` (`:333-334`), and the `ClientData` type declaration (`models.ts:32`). There is no fifth reader anywhere.
* AC-010 deletes `selectedHasVtlFiles`; fix (1) repoints `vtlMenuItems` at the hover context. Nothing then reads `vtlFiles` off `editorSelected.payload` — confirmed, not just plausible.
* The load-bearing premise also checks out: `events.ts:326-335` (hover) and `:418-430` (click) both compute `vtlFiles` via `findDotCMSVTLData(foundElement)` unconditionally. Only `getDotCMSContentletsBound()` omits it. So the component fix is sufficient on its own.

**Decision: AC-009 is withdrawn** (pushed in `b80c253cd0`). Not relabelled as hardening — dropped. Carrying a field forward that nothing reads, backed by a store spec that cannot fail for any user-visible reason, isn't worth the code. It also sat awkwardly next to the spec's own "the narrowness is deliberate" line, which you were right to point at.

What changed in the spec:

* **AC-009 marked WITHDRAWN, and deliberately not renumbered** — AC-010 stays AC-010 so this review thread keeps resolving to the right items.
* **`withSelectionAnchor.spec.ts` dropped** from the verification plan. One spec file now, not two.
* **New non-goal: no store change at all.** With it, the constraint is written down explicitly: the `SET_BOUNDS` re-anchor drops `vtlFiles`, `baseType`, `onNumberOfPages` and `dotStyleProperties` uniformly, and *any future consumer reading one of those off `editorSelected.payload` — a selected-toolbar VTL button, say — must repopulate it or hit #37499 again*. That's the knowledge your finding surfaced, kept without shipping the code.
* **Regression risk section rewritten.** The store is now entirely outside the diff, which removes the fix's only shared-state risk. What remains is confined to one component.
* **Assumptions now record the hover-event premise** — if `findDotCMSVTLData` ever becomes conditional in the hover path, this fix regresses and the carry-forward comes back on the table.

**On `menuItems()` (`:308-310`)** — confirmed intentionally untouched, and I've added that as an explicit non-goal so PR 2's reviewer doesn't have to re-derive it. It shares the selected-preferred dual-context pattern but never reads `vtlFiles`; it passes the context straight through as the `addContent` payload, where preferring the selected contentlet is the intended behavior, not a bug.

Spec is 10/10 on its quality checklist after the revision. Ready for another look.

dario-daza:
I agree with the approach you're presenting for the menu following the hovered
contentlet, it's more natural for the user. But there's one thing I hit while
testing by hand that I don't think the spec covers.

**Clicking `</>` doesn't change the selection — and after this fix, that becomes visible**

With contentlet A selected, hover contentlet B and click the bolt or the palette
button: B becomes the selected one, the border moves, the side panel follows. Click
`</>` on B instead and nothing moves; A stays selected.

Right now you don't really notice, because the menu is listing A's files anyway.
Once the menu follows the hover, you'd be editing B's VTL while the editor still
shows A as the selected one.

Do you reckon we should leave the selection where it is, or select the hovered
contentlet after clicking its `</>` button? Either way, I think it's worth a line in
the spec so it's a decision rather than just how it happens to behave.

https://github.com/user-attachments/assets/c3a6caac-bec3-4128-813a-c9e801c1a69b

dario-daza (approved):

PR #37524, opened 2026-09-11
fix(content-drive): close the timing gap that lets a fast double click fire two page requests (#37212)

### LINKED ISSUE #37212: Content Drive: pagination buttons stay enabled while a page is loading, allowing duplicate requests
## Problem Statement

The Content Drive pagination controls (next/previous page) remain clickable while a page request is already in flight. Clicking them again before the current request resolves fires additional, redundant requests for the same navigation action.

## Steps to Reproduce
Video: [content drive pagination buttons.mov](https://drive.google.com/file/d/1RTR3O6l-eizLRpik-DhoIjxDxedk3DrY/view?usp=drive_link)

1. In Content Drive, navigate to a folder/search result with multiple pages.
2. Click the "next page" (or "previous page") button while the current page is still loading.
3. Click it again (or multiple times) before the request completes.
4. Observe multiple network requests fired, one per click, instead of the click being ignored/debounced while a request is in flight.

## Expected vs Actual

Expected: pagination controls should be disabled (or clicks ignored) while a page request is in progress, so only one request is made per navigation action.

Actual: pagination buttons remain enabled during loading; repeated clicks fire additional requests.

## Acceptance Criteria

- [ ] Pagination controls are disabled (or non-interactive) while a page request is in flight
- [ ] Rapidly clicking a pagination control while loading results in exactly one request, not one per click
- [ ] No regression to normal pagination behavior once a request completes

## Parent Epic

#36814

ihoffmann-dot:
## Summary
- Content Drive's pagination guard (`if (this.$loading()) return;` in `DotFolderListViewComponent#onPage`) already blocks a second click once the caller's `loading` input reflects the in-flight request — but that input takes at least one microtask to catch up (it round-trips through the caller's store via `paginate` → `setPagination` → a reactive effect → `loadItems()`). A fast double click, or two rapid keyboard activations, can both land inside that gap and slip past the guard, firing two requests for the same navigation action.
- Adds a synchronous local flag (`#pageChangeAccepted`) that closes exactly this gap: set the instant a page change is accepted, cleared by an effect the moment `$loading()` first reflects it. The existing guard is untouched and remains the sole gate for the rest of a request's lifetime.

Fixes #37212.

## Test plan
- [x] `pnpm nx test ui -- dot-folder-list-view` — 241/241 passing, including 2 new tests:
  - a rapid second `onPage()` call before `loading` catches up fires exactly one `paginate` emit
  - a page change is accepted again once `loading` reflects the first one (guard correctly releases)

This PR fixes: #37212

### DIFF
--- a/core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
+++ b/core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
@@ -879,6 +879,14 @@ export class DotFolderListViewComponent implements OnInit, AfterViewInit, OnDest
         this.#seedRovingTabStop();
         this.#applyRovingTabStop();
         this.#trackFocusIntent();
+
+        // Hands the local page-change guard back to `$loading()` the moment the store's own status
+        // catches up with the click that set it (found in review, issue #37212).
+        effect(() => {
+            if (this.$loading()) {
+                this.#pageChangeAccepted = false;
+            }
+        });
     }
 
     readonly #onKeydownCapture = (event: KeyboardEvent): void => {
@@ -1178,6 +1186,20 @@ export class DotFolderListViewComponent implements OnInit, AfterViewInit, OnDest
     }
 
     /**
+     * True from the moment a page-change click is accepted until `$loading()` first reflects it.
+     *
+     * `$loading()` mirrors the caller's request lifecycle one round trip away: this component emits
+     * `paginate`, the caller patches its store, and only once that store's own reactive pipeline
+     * reacts does `loading` flip to `true` and propagate back down here as an input. That round
+     * trip is at least one microtask, and during it `$loading()` still reads `false` — so a second
+     * click in that gap (a fast double click, or two rapid keyboard activations) would sail past the
+     * `$loading()` guard below and queue a second request the first click already started (found in
+     * review, issue #37212). This flag closes exactly that gap; once `$loading()` catches up (see the
+     * constructor's effect), that guard alone is sufficient for the rest of the request's lifetime.
+     */
+    #pageChangeAccepted = false;
+
+    /**
      * Handles pagination events from the PrimeNG Table
      * @param event The lazy load event containing pagination info
      */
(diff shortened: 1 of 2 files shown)


dario-daza (approved):

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
```diff
@@ -879,6 +879,14 @@ export class DotFolderListViewComponent implements OnInit, AfterViewInit, OnDest
         this.#seedRovingTabStop();
         this.#applyRovingTabStop();
         this.#trackFocusIntent();
+
+        // Hands the local page-change guard back to `$loading()` the moment the store's own status
+        // catches up with the click that set it (found in review, issue #37212).
+        effect(() => {
+            if (this.$loading()) {
+                this.#pageChangeAccepted = false;
```
zJaaal, line 887: Should this be a signal?

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
```diff
@@ -1188,15 +1210,18 @@ export class DotFolderListViewComponent implements OnInit, AfterViewInit, OnDest
...
-        // for last. The controls are also visually disabled via `$ptConfig`, but this is the part
-        // that has to hold: a keyboard activation or a fast double click does not wait for CSS.
-        if (this.$loading()) {
+        // A search is already in flight, or one was just accepted and the caller's `loading` has not
+        // caught up yet (see `#pageChangeAccepted`). The paginator stays mounted while the rows are
+        // replaced by skeletons, so without this every further click fires another search for a page
+        // the user cannot see yet, and the last response to arrive wins regardless of which page was
+        // asked for last. The controls are also visually disabled via `$ptConfig`, but this is the
+        // part that has to hold: a keyboard activation or a fast double click does not wait for CSS.
+        if (this.$loading() || this.#pageChangeAccepted) {
```
zJaaal, line 1219: Are we sure we need a double gate? 

Seems like redundancy, since looks like we set it in the same cycle

zJaaal (changes requested):

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
zJaaal: Maybe we need to check the pointers, to prevent forward navigation

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
ihoffmann-dot: `#pageChangeAccepted` is only read in imperative code paths (`onPage`, the constructor's `effect`). It never drives the template, so I think a signal here wouldn't change the behavior. What do you think?

core-web/libs/ui/src/lib/components/dot-folder-list-view/dot-folder-list-view.component.ts
ihoffmann-dot: I don't think it's redundant, they cover different windows. `$loading()` only flips true after the `paginate` emit round trips through the caller's store, at least one microtask away. `#pageChangeAccepted` closes that gap synchronously. Once `$loading()` catches up, the effect hands the gate back to it.

zJaaal (approved):

PR #37525, opened 2026-09-11
spec: Content Drive global search hides matches without content-type filter (#37211)

### LINKED ISSUE #37211: Content Drive: adding a content-type filter to the global search hides matches the search alone found
## Problem Statement

In Content Drive, the global search filter and a content-type filter don't compose correctly: typing a term in the global search should return all matching content, and narrowing further by content type should return a subset of those same matches. Instead, adding a content-type filter surfaces content that the plain global search did not return, and removing the content-type filter again makes that content disappear from the results — even though the same search term is used throughout. The Search All portlet composes global search + content-type correctly for the same term.

## Steps to Reproduce

Video: [content drive general filter difference.mov](https://drive.google.com/file/d/1IFamcKQJKEj_yvZc67CFhYOtF7S8nX7Y/view?usp=drive_link)

1. In Content Drive, use the global search filter only and type "capy".
2. Expected capybara image assets (uploaded and matching by filename) are missing from the results.
3. Without clearing the search term, add a content-type filter for File Asset.
4. The same capybara images now appear.
5. Remove the content-type filter again (keeping the search term) — the capybara images disappear from the results again.
6. In the Search All portlet, typing "capy" in the global search (no content-type filter) returns the capybara images correctly, with no need for an additional filter.

## Expected vs Actual

Expected: the global search filter alone returns every content item matching the term, matching what Search All returns; adding a content-type filter should only narrow that same set, never introduce results that weren't already found by the term alone.

Actual: Content Drive's global search filter alone misses matches; adding a content-type filter surfaces items the unfiltered search should have already found; removing the content-type filter makes them disappear again. The Search All portlet does not exhibit this inconsistency for the same term.

## Acceptance Criteria

- [ ] The global search filter alone returns every content item that would also be returned when a content-type filter is added for the same term
- [ ] Adding a content-type filter to the global search only narrows the existing match set; it does not introduce items absent from the unfiltered search
- [ ] Behavior is consistent with the Search All portlet for the same search term
- [ ] Verified against a large dataset (~718,174 contentlets)

## Parent Epic

#36814

ihoffmann-dot:
## Summary

- PR 1 of 2 (Spec-Kit): carries only `spec.md`, per the dotCMS spec-driven workflow. Needs approval before `/speckit-plan` runs.
- Issue: dotCMS/core#37211 — Content Drive's global (free-text) search filter silently omits matches that a content-type filter (for the same term) does find.

## Root cause (confirmed via code-level investigation, not yet live-reproduced)

Content Drive's unfiltered global search goes through `BrowserAPIImpl`'s default hybrid DB+ES chunked-scan heuristic (`doHybridSingleChunkedQueryES`), which scans the SQL candidate set in fixed-size DB-order pages (`BROWSER_CONTENT_CHUNK_SIZE`, default 900) and stops once `maxResults` is met or `BROWSER_DB_MAX_SCAN_ROWS` (default 50,000) is hit. Without a content-type filter, matching rows can sit far enough down DB order that they're never loaded and sent to ES at all — not an ES matching failure, a candidate-selection one. Adding a content-type filter narrows the SQL candidate set (`struc.inode in (...)`), pulling matches earlier into the scan.

Two alternative theories were checked and refuted with code evidence (see spec): that the ES query itself changes shape based on content-type filter presence, and that `catchall` doesn't index filenames.

## What's NOT in this PR

- No code changes — implementation happens after `/speckit-plan` → `/speckit-tasks` → `/speckit-implement` on a follow-up branch/PR.
- One open `[NEEDS CLARIFICATION]`: pinning the exact class/line backing the legacy Search All portlet's ES-direct path (architecturally confirmed, not yet cited to file:line) — deferred to `/speckit-plan`.

## Test plan

- N/A for this PR (spec only). Verification approach for the eventual fix is documented in the spec's Acceptance & Verification section.

## Issue

Fixes: [#37211](https://github.com/dotCMS/core/issues/37211)

This PR fixes: #37418

### DIFF
--- a/specs/37418-content-drive-global/spec.md
+++ b/specs/37418-content-drive-global/spec.md
@@ -1,0 +1,204 @@
+# Issue Resolution Specification: Content Drive: adding a content-type filter to the global search hides matches the search alone found
+
+**Feature Branch**: `37418-content-drive-global`
+
+**Created**: 2026-09-11
+
+**Status**: Draft
+
+**Type**: Issue / Bug Resolution
+
+**Related GitHub Issue**: [#37211](https://github.com/dotCMS/core/issues/37211)
+
+**Input**: User description: "Content Drive: adding a content-type filter to the global search hides matches the search alone found"
+
+<!--
+  This is the dotCMS ISSUE-RESOLUTION spec (used by /speckit-specify-fix). Unlike the
+  feature spec, it is framed around a defect: what is wrong, how to reproduce it, and how
+  we will know it is fixed. It still flows into /speckit-plan, where the Legacy Impact and
+  ADR Alignment gates apply. Keep this technology-light — root-cause and fix details are
+  refined in the plan.
+-->
+
+## Problem Statement *(mandatory)*
+
+In Content Drive, the global (free-text) search filter and the content-type filter don't
+compose correctly. Typing a search term with no content-type filter should return every
+content item matching that term; adding a content-type filter should only narrow that same
+set. Instead, Content Drive's unfiltered global search silently omits matching content that
+the *same* search term finds as soon as a content-type filter is added — and removing the
+content-type filter again makes that content disappear. The legacy Search All portlet does
+not have this problem: it returns the correct matches for the same term with no content-type
+filter needed.
+
+This is not a "search can't find this term" bug — the term does match the content. It's a
+"content that matches never gets asked" bug: on a large site, the affected content can be
+silently excluded from every unfiltered global search result set, with no error, warning, or
+indication to the user that results are incomplete.
+
+**Severity / Impact**: Any Content Drive user searching without a content-type filter, on a
+site with enough content that the affected item isn't scanned early. Confirmed on File Asset
+content reached by filename; the underlying mechanism is not File-Asset-specific and can
+plausibly bury matches of any content type. Silent data-completeness bug — no error is
+raised, so users have no way to know their global search results are incomplete unless they
+happen to retry with a content-type filter.
+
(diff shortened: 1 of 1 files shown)


zJaaal (dismissed):

dario-daza:
Three things I'd like to have a check from you before this is approved.

**1. The fix has no direction, and two of the three named options conflict with the stated constraint.**

"Exhausting the full scan before giving up" and "preserve the hybrid heuristic's purpose (bounding scan cost on very large sites)" cannot both hold at 718,174 rows — exhausting the scan means loading and ES-filtering up to 718,174 rows per search, which is exactly what `BROWSER_DB_MAX_SCAN_ROWS` exists to prevent. Deferring the decision to `/speckit-plan` is fine, but Fix Scope should at least enumerate the real alternatives (stay DB-first vs. invert to ES-first the way Search All does) and the criterion for choosing, because risk, tests and latency all follow from that choice.

**2. AC-004 isn't verifiable as written.**

The AC requires validation at ~718,174 contentlets, but the verification method proposes seeding "enough content to exceed `BROWSER_CONTENT_CHUNK_SIZE`" — that's **900** rows, not the **50,000** of `BROWSER_DB_MAX_SCAN_ROWS`. Exceeding 900 exercises the chunk loop, not the cutoff that causes the defect. The integration test needs to lower `BROWSER_DB_MAX_SCAN_ROWS` via `Config` so the cutoff path is reachable at fixture scale, and the spec should say so. As written, AC-004 is manual-only.

dario-daza (approved):

PR #37531, opened 2026-09-14
Update starter.deploy.version to [empty_20260914]

### LINKED ISSUE #37507: Publish Starter PR creation fails — duplicate Authorization header between actions/checkout and create-pull-request
## Problem Statement

The **Publish Starter** GitHub Actions workflow fails at the **Create Pull Request** step of the `deploy-artifacts` job, after the artifact deployment and the `pom.xml` update have already completed successfully.

The step fails because `actions/checkout` and `peter-evans/create-pull-request` configure git credentials by different, mutually incompatible mechanisms. Both end up active at the same time, git transmits **two** `Authorization` headers on the same request, and GitHub rejects it with HTTP 400.

No pull request is created, so the `starter.deploy.version` bump produced by the run is discarded and the Empty Starter publication is never recorded in the codebase.

### Root cause (verified)

Workflow run:

- Workflow: `Publish Starter`
- Run: `#77`
- Commit: `44a4026`
- Failed job: `deploy-artifacts`
- Failed step: `Create Pull Request`

Verbatim from the job log:

```text
remote: Duplicate header: "Authorization"
fatal: unable to access 'https://github.com/dotCMS/core/': The requested URL returned error: 400
##[error]The process '/usr/bin/git' failed with exit code 128
```

The failing git invocation is `git remote prune origin` — the action's first network call, during "Checking the base repository state". It fails before any branch is created or pushed.

The mechanism:

1. `actions/checkout@v7.0.1` checks out the repo and, with its default `persist-credentials: true`, leaves credentials behind. Newer checkout versions no longer do this solely via `http.<url>.extraheader`; they write a separate git credentials config file wired in through `includeIf.gitdir` entries. The run's own post-job cleanup confirms this shape — it removes `includeif.gitdir:` entries pointing at `/home/runner/work/_temp/git-credentials-<uuid>.config`.
2. `create-pull-request@v6` probes for persisted credentials by looking **only** for a local `http.https://github.com/.extraheader` matching `^AUTHORIZATION:`. Since checkout v7 supplies them through the `includeIf` file instead, the probe finds nothing and the action injects its own `extraheader` using `CI_MACHINE_TOKEN`.
3. Both credential sources are now live. Git sends two `Authorization` headers, GitHub answers `400 Duplicate header: "Authorization"`, git exits 128.

This is a known upstream incompatibility, not a dotCMS misconfiguration:

- peter-evans/create-pull-request#4228 — "Incompatible with actions/checkout@v6"
- peter-evans/create-pull-request#4272 — the duplicate-header error verbatim

The action's maintainer states the minimum compatible release is **v7.0.9**; a reporter independently confirms the **v8** line works.

**Trigger date**: `actions/checkout` in this workflow was bumped to **v7.0.1** on **2026-08-28** (commit `31102249cc`, PR #36850, a repo-wide action-bump effort). That bump moved the workflow onto the incompatible side of the boundary.

### Why the Node.js 20 deprecation is not the cause

The same job emits a Node.js 20 deprecation notice naming this action. It is a concurrent symptom rather than the cause:

- The notice is a non-fatal **warning**, and it names **two** actions, not one:

  ```text
  ##[warning]Node.js 20 is deprecated. The following actions target Node.js 20 but are being
  forced to run on Node.js 24: jfrog/setup-jfrog-cli@v4, peter-evans/create-pull-request@v6
  ```

- `jfrog/setup-jfrog-cli@v4` carries the identical deprecation **in the same job** and completed with conclusion `success`. Being forced onto Node 24 does not, by itself, fail a step.
- `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true` would **not** have fixed this. Forcing the action back onto Node 20 leaves the credential conflict untouched, so ruling that escape hatch out holds on these grounds too.

Upgrading the action clears both, but it unblocks the workflow by resolving the credential conflict, not by changing the Node.js runtime. **This distinction drives validation**: success is defined by a pull request being created, not by a warning disappearing.

### Not a transient failure — explicitly ruled out

- Run **#76** succeeded five minutes earlier on the **identical commit**. It is not counter-evidence: its `Checkout repository`, `Update pom.xml`, and `Create Pull Request` steps all show conclusion **`skipped`**. It was not an `empty` + non-`dry-run` publication, so it never exercised the failing path.
- Runs **#72–#75** (2026-07-30 → 2026-07-31) likewise skipped both `Update pom.xml` and `Create Pull Request`.
- **There is no run in recent history that successfully executed the pull-request step.** This path has no regression coverage — it runs only on real Empty Starter publications, which are infrequent — so the defect stayed latent from the moment it was introduced until the first run that reached the step.

## Steps to Reproduce

Requires `type: empty` **and** `dry-run: false`; the failing step is gated on both and is skipped otherwise.

1. Trigger the `Publish Starter` workflow manually with `type: empty`, `dry-run: false`, and a valid `issue-number`.
2. Allow `get-starter`, artifact download, deployment, and the `pom.xml` update to complete — all succeed.
3. Reach the `Create Pull Request` step in the `deploy-artifacts` job.
4. **Actual:** the step fails ~1 second in with `Duplicate header: "Authorization"` / HTTP 400 / git exit 128. No branch, no pull request.
5. **Expected:** the workflow pushes the auxiliary branch, opens or updates the version pull request, and completes successfully.

**Reproducibility**: deterministic — a function of the action versions present, not of timing, network, or credential validity.

## Proposed Fix

Replace the incompatible action version:

```yaml
uses: peter-evans/create-pull-request@v6
```

with a release compatible with `actions/checkout` v6+ — **minimum v7.0.9**, with the current **v8** line preferred. Choosing v8 additionally clears the Node.js 20 deprecation notice for this action, resolving both the failure and the deprecation symptom in one move.

Pin the replacement to its full commit SHA with a trailing version comment, per the repository's prevailing convention (116 SHA-pinned vs. 49 tag-pinned usages across `.github/workflows/`; see commit `3881aba812`, "pin every bumped action to a commit SHA"). The current `@v6` is one of the tag-pinned stragglers.

### Migration risk

Assessed against upstream release notes:

- **v7.0.0** behaviour changes: `git-token` renamed to `branch-token`; deprecated `PULL_REQUEST_NUMBER` output env var removed; rate-limited requests now retried. **None apply** — this workflow uses neither `git-token` nor `PULL_REQUEST_NUMBER`.
- **v8.0.0** is a runtime bump only (Node 24). Its sole compatibility requirement — Actions Runner v2.327.1+ — applies to **self-hosted** runners; this job runs on a GitHub-hosted `ubuntu-24.04` runner.
- The six inputs actually in use (`token`, `branch`, `commit-message`, `title`, `body`, `labels`) are unchanged across v6 → v8.

Residual risk is low, but must still be confirmed by an actual run, since no recent run has exercised this step.

## Secondary defect found during investigation (in scope)

The `send-notification` job composes its Slack message from `needs.update-pom.outputs.pull-request-url`. **There is no job named `update-pom`** — `update-pom` is a *step id* inside `deploy-artifacts`, and `send-notification` declares `needs: [deploy-artifacts]` only. That expression resolves to an empty string on every run, and `deploy-artifacts` does not expose the pull request URL as a job output either.

Consequence: even once the pull request is created successfully, the Slack alert meant to prompt operators for approval shows an empty link.

Included in this fix rather than split out: it is the only channel through which an operator learns the pull request exists, so without it the primary fix cannot be observed from where operators actually watch this workflow.

## Acceptance Criteria

**Primary — the failure is resolved**

- [ ] A manual `Publish Starter` run with `type: empty`, `dry-run: false` completes the `Create Pull Request` step successfully
- [ ] The reproduction no longer yields `Duplicate header: "Authorization"` / git exit code 128
- [ ] The run opens a pull request against the intended base branch carrying the `starter.deploy.version` bump in `parent/pom.xml`
- [ ] The pull request uses the expected branch name, title, body, and both `empty-starter` and `automated pr` labels
- [ ] The `deploy-artifacts` job completes successfully end to end

**No regressions**

- [ ] JFrog CLI setup, artifact download, and artifact deployment work unchanged
- [ ] `pom.xml` is updated with the expected version before the PR creation step
- [ ] A gated-off run (`type` other than `empty`, or `dry-run: true`) still skips `Update pom.xml` and `Create Pull Request` and completes successfully, creating no branch, commit, or pull request
- [ ] Re-running when the auxiliary branch already exists updates the existing pull request rather than failing

**Slack notification**

- [ ] The Slack notification contains a working link to the created pull request, sourced from a `deploy-artifacts` job output rather than a reference to a non-existent job
- [ ] A gated-off run's notification degrades cleanly — no broken or placeholder pull request link

**Supply chain**

- [ ] The action is pinned to a full commit SHA with a trailing version comment
- [ ] The fix does not require `ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true`

**Corroborating (not the definition of success)**

- [ ] The Node.js 20 deprecation warning no longer names `peter-evans/create-pull-request`. It may still name `jfrog/setup-jfrog-cli@v4` — that is expected and out of scope for this issue.

**Validation**

- [ ] The workflow is tested from the fix branch before merging
- [ ] The successful run URL is attached to this issue
- [ ] The PR description states that the root cause was the duplicate `Authorization` header rather than the Node.js runtime, so future readers do not re-derive it from the warning alone

## Out of scope

- The Node.js 20 deprecation carried by `jfrog/setup-jfrog-cli@v4` — non-fatal, that step succeeds today, unrelated action with its own upgrade considerations. Tracked separately.
- A repository-wide audit of the remaining ~49 tag-pinned actions.
- Introducing Dependabot or any automated action-update mechanism.
- Restructuring the workflow (job layout, gating, dry-run semantics, Artifactory deployment, Slack message format).
- Adding automated regression coverage for the publication path — a real gap, documented above, but materially larger than this fix.

## dotCMS Version

Latest from `main` at commit `44a4026` when the failure was observed. No dotCMS application code is involved; the defect is entirely in repository CI configuration.

## Severity

Medium–High for release operations, zero for product runtime. Only the Empty Starter publication path is blocked; Artifactory deployment still succeeds and no customer-facing functionality is affected. Every Empty Starter publication currently requires a manual pull request to carry the `pom.xml` change.

## Links

- Failed run: https://github.com/dotCMS/core/actions/runs/34533343387/job/103059171093
- Upstream incompatibility: https://github.com/peter-evans/create-pull-request/issues/4228
- Upstream duplicate-header report: https://github.com/peter-evans/create-pull-request/issues/4272
- Action documentation: https://github.com/peter-evans/create-pull-request
- GitHub Node 20 deprecation: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
- Spec (Spec-Kit): `specs/37507-publish-starter-pr-auth/spec.md` on branch `37507-publish-starter-pr-auth`

dotCMS-Machine-User:
This PR was created automatically to update the **starter.deploy.version** in pom.xml to [**empty_20260914**].

This PR fixes: #37507


dcolina:
Closing unmerged — this PR was produced by a post-merge validation run for #37507 (run #80), not by an intentional Empty Starter publication. The `starter.deploy.version` bump is not meant for `main`.

Note for whoever runs the next real publication: this PR also contains `starter/empty_20260914.zip` (50 KB), committed because the workflow downloads the artifact into the workspace and `create-pull-request` commits any untracked changes. That is a pre-existing defect, newly visible now that the PR step works again. Tracking separately.

PR #37535, opened 2026-09-14
fix(content-import): validate Content Type before streaming sample CSV

### LINKED ISSUE #37461: Import Content: "Click here to download a CSV sample file" link is broken - no sample CSV can be downloaded
## Problem

In the **Content → Import Content** portlet, the link **"Click here to download a CSV sample file"** is broken. Clicking it does not download any file — the browser navigates to a Liferay/Struts URL that fails to resolve and renders Chrome's "webpage might be temporarily down or it may have moved permanently to a new web address" error page.

This blocks the documented starting point for a CSV content import: users have no reference file showing the expected column headers for the selected Content Type.

## Steps to Reproduce

1. Log in to dotAdmin (reproduced on `demo.dotcms.com`).
2. Go to **Content → Search** and open the **Import Content** screen (`/dotAdmin/#/c/content?mId=71b8...`).
3. Click **"Click here to download a CSV sample file"**.

## Expected Behavior

A sample CSV file is downloaded, containing the header row for the selected Content Type / language.

## Actual Behavior

No file is downloaded. The browser leaves dotAdmin and lands on an error page for a URL of the form:

```
/c/portal/layout?p_l_id=71b8a1ca-37b6-4b6e-a43b-...&p_p_id=content&p_p_action=1&p_p_state=maximized
  &angularCurrentPortlet=content&p_p_mode=view
  &_content_struts_action=%2Fext%2Fcontentlet%2Fimport_contentlets&_content_cmd=...
```

Note the link also assumes the legacy portal/Struts navigation, so it breaks out of the Angular admin shell instead of triggering a download in place.

## Notes / Possible Cause

The sample-CSV action still points at the legacy `/c/portal/layout` + `_content_struts_action=/ext/contentlet/import_contentlets` endpoint. It looks like either the endpoint no longer resolves or the required parameters are no longer being built correctly from the Angular portlet, so the request 404s instead of returning the CSV.

Worth confirming whether the sample download should be reimplemented as a direct REST/servlet download (e.g. returning `text/csv` with `Content-Disposition: attachment`) so it works independently of the portal layout URL.

## Environment

- Instance: `demo.dotcms.com`
- Browser: Chrome (macOS)
- Portlet: Import Content (Content Search)

## Screenshots

1. Import Content screen with the "Click here to download a CSV sample file" link.
2. Error page shown after clicking the link.

Neehakethi:
Fixes #37461

### Proposed Changes
* Guard `ImportContentletsAction._downloadCSVTemplate` so a missing Content Type re-renders the Import Content screen with the existing `structure-type-is-required` message instead of throwing after the download headers are committed. The field lookup now happens before any header is set, so a failed lookup can no longer leave the response flagged as an attachment with an empty error body.
* Validate the Content Type client-side in `import_contentlets.jsp#downloadCSVExample()` and show "A Content Type is required" via `showDotCMSSystemMessage`, so the form never submits an empty `structure`.

### Root cause
The link is not dead — it breaks only when no Content Type is selected, which is the screen's default state. In that state the form POSTs `structure=""`; the action set `Content-Type: application/octet-stream` + `Content-Disposition: attachment` and opened the output stream first, then `FieldsCache.getFieldsByStructureInode("")` → `ContentTypeAPI.find("")` threw. The result was an HTTP 500 with attachment headers and an empty body, which Chrome renders inside the portlet iframe as "webpage might be temporarily down or it may have moved permanently" — the error page in the issue screenshots. With a Content Type selected the download worked all along (verified on `demo.dotcms.com` 26.09.03-01: valid `structure` → 200 + correct CSV; empty `structure` → 500 + attachment headers + empty body).

The fix mirrors the validation the `preview` branch of the same `processAction` already applies, reusing its existing i18n key (translations already present).

### Checklist
- [ ] Tests — no automated test added: `ImportContentletsAction` has no existing test coverage and no portlet-action test harness; flagged as a known gap. Verified by reproduction (below).
- [x] Translations — reuses the existing `structure-type-is-required` key (`Language.properties`), no new keys.
- [x] Security Implications Contemplated — no new surface; the change adds input validation to an existing authenticated portlet action.

### Additional Info
Verified on a local build from this branch (empty starter):
* The rendered Import Content screen contains the new client-side guard.
* The exact request that previously produced the broken response (`cmd=downloadCSVTemplate` with empty `structure`) now returns HTTP 200 re-rendering the portlet, with no `Content-Disposition` header.
* With a valid `structure`, `CSV_Template.csv` still downloads with the correct header row for the selected type.
* Regression: the `preview` validation path (whose pattern this reuses) behaves as before.

The issue floated reimplementing the download as a REST endpoint (e.g. `GET /v1/content/_import/_template`). Deliberately out of scope here — the legacy endpoint works once validated; a REST template endpoint can be a follow-up if wanted.

### Screenshots
Chrome error page inside the portlet iframe (see issue #37461 screenshots) | "A Content Type is required" system message; screen stays put. With a type selected, `CSV_Template.csv` downloads.

Before : image

After : 
image

This PR fixes: #37461


danielsolis-dotcms (dismissed):

dotCMS/src/main/java/com/dotmarketing/portlets/contentlet/action/ImportContentletsAction.java
```diff
@@ -376,13 +376,21 @@ private void _downloadCSVTemplate(ActionRequest req, ActionResponse res, Portlet
...
+		ImportContentletsForm importForm = (ImportContentletsForm) form;
+		if (!UtilMethods.isSet(importForm.getStructure())) {
+			SessionMessages.add(req, ERROR, "structure-type-is-required");
+			setForward(req, PORTLET_EXT_CONTENTLET_IMPORT_CONTENTLETS);
+			return;
+		}
+
+		// The fields must be resolved before any download header is committed, otherwise a
+		// failed lookup leaves the response flagged as an attachment with an empty error body
+		List<Field> fields = FieldsCache.getFieldsByStructureInode(importForm.getStructure());
```
danielsolis-dotcms: Maybe we should validate here that importForm.getStructure() contains an existing content type too? Not only that it has a value.

Neehakethi: Good call — addressed in e1533180ef. The lookup is now wrapped in a try/catch: a set-but-nonexistent inode (e.g. type deleted after the screen loaded) logs a warning and re-renders the import screen with the existing `Workflow-does-not-exists-content-type` message ("Content Type does not exist"), through the same SessionMessages + setForward path as the empty-selection guard. Went with catch-on-lookup rather than a separate existence check since `FieldsCache.getFieldsByStructureInode` already performs the `ContentTypeAPI.find`, so a pre-check would run the same query twice.

danielsolis-dotcms (approved):

PR #37537, opened 2026-09-14
docs(spec): GraphQL access to inherited-subtype properties on Image and File fields (#34540)

### LINKED ISSUE #34540: GraphQL Support for dotAssets and FileAssets in Image/File Fields

fabrizzio-dotCMS:
**PR 1 of 2 — spec only.** Carries `spec.md` alone. Approval here gates `/speckit-plan`; no implementation until then.

Refs #34540.

## The problem

Querying an Image or File field over GraphQL returns `DotFileasset` — a flat, six-property object built once in a static block — instead of the real type of the asset the field points at.

```
image { tags }       → Field 'tags' in type 'DotFileasset' is undefined
image { identifier } → Field 'identifier' in type 'DotFileasset' is undefined
image { ... on Images { tags } }
                     → objects of type DotFileasset can never be of type Images
```

So a customer who extends DOTASSET or FILEASSET to model their own assets can author those fields but cannot read them back. The AI tagging workflow is blocked the same way. There is no workaround.

Root cause is `ContentAPIGraphQLTypesProvider` lines 97-98, mapping both `ImageField.class` and `FileField.class` to the single `CustomFieldType.FILEASSET`. The per-content-type generation and the `DotAssetBaseType` / `FileBaseType` interfaces already exist and are already dynamic — verified by introspection against a running instance — so the gap is only in how these two field classes are typed.

## Two decisions the reviewer is being asked to confirm

> **Update — the first decision was reversed, and nothing breaks any more.** This section
> originally asked you to accept breaking existing queries. It no longer does. The corrected
> reasoning is below; the superseded argument is kept in `spec.md`'s decision section, because the
> path the reasoning took is the useful part of the record. The current `spec.md` lives in #37594,
> which branches off this one.

**1. Existing queries keep working (FR-012). No exception to ADR-0022 is needed.**

All six properties the flat view exposed — `fileName`, `description`, `fileAsset`, `metaData`,
`showOnMenu`, `sortOrder` — remain selectable and return the same values. Nothing is removed,
nothing is renamed, no query has to be rewritten.

The one that looked impossible was `description`, and it is worth understanding why it was not.
Two meanings have always shared that name, in production, today:

```
image { description }             → the contentlet TITLE (for an asset, the file name)
ImagesCollection { description }  → the stored description, usually empty
```

Measured on a running instance: the first returns a value for 57 of 57 images, while only 2 of
those 57 have a stored description. The flat view derived it from the title; the content type's
own field holds what an editor typed. Once the asset field is typed by an interface, the concrete
type's definition would normally take over — which is exactly the silent change FR-009a forbids.
So the value is selected by how the asset was reached, and both contracts hold unchanged.

That costs one data fetcher. The earlier spec concluded removal was the only way to honour
FR-009a; it was not.

**What does change**, neither of it a query that stops validating:

| | Before | After |
|---|---|---|
| `__typename` on an asset field | the constant `"DotFileasset"` | the concrete type (`"Images"`, `"BannerImages"`) |
| kind of `DotFileasset` | object | interface |
| asset field pointing at **non-asset** content | resolved the flat view | `null` |

`__typename` is inseparable from the capability: the same resolution that makes
`... on BannerImages` work is what it reports, so pinning it would mean naming a type the value
does not have. The kind change affects clients that generate types from the schema (Apollo
codegen, Relay, graphql-codegen) — **not** the `@dotcms/*` SDK, which sends the fragment text the
customer writes and does not generate types from the schema. The `null` is forced: handing a
non-asset contentlet to the interface raises `UnresolvedTypeException`, which fails the entire
request rather than that one field.

**2. PR #35363 is superseded, not merged (FR-013).**

That PR adds a `DotBinaryLike` interface flattening 12 binary properties onto `DotFileasset`, and states it closes #34540. It delivers only the second limb of the issue; customer-defined fields stay unreachable after it.

Its benefit does not survive this feature. It exists to spare one level of nesting — `image { size }` over `image { fileAsset { size } }` — but once the field returns the asset's real type, that type's `asset` property is already a `DotBinary` carrying all 12 plus `focalPoint`. The client still descends exactly one level, just to the correctly named property. Meanwhile its deliverable is attached to the view this feature removes, its checks have been failing since it was last updated on a check in its own area, and it carries no review.

**The remaining scope is deferred, not dropped**: #34540 stays open after this ships, carrying the second stage, whose design depends on the shape this one establishes.

## What this means for a client's query

Example

### Your query, unchanged

```graphql
query Banners {
  BannerCollection(limit: 10) {
    title
    image {
      fileName
      description
      fileAsset { versionPath size mime width height }
      metaData { key value }
    }
  }
}
```

This validates and returns the same values it did before, `description` included. **That is the
entire migration for a typical client: nothing.**

### After — what the capability buys

```graphql
image {
  fileName  versionPath  size  mime  width  height   # the binary, no extra level

  identifier  title  live  urlMap                     # the asset's own identity
  __typename                                          # which content type it is

  ... on Images       { tags }                        # customer-defined properties
  ... on BannerImages { campaignName adSize }
}
```

The middle two groups were not selectable before at any depth. The last group is the limb of #34540 with no workaround at all — it is what blocks the AI tagging workflow.

### A mixed collection

```graphql
image {
  baseType                                        # DOTASSET or FILEASSET, no clause needed
  __typename
  ... on DotAssetBaseType { asset     { size mime } }
  ... on FileBaseType     { fileAsset { size mime } }
}
```

Each row answers with only the clause that applies; a clause that does not match contributes nothing and does not fail.

Migration guide for clients: `docs/backend/GRAPHQL_ASSET_FIELDS_MIGRATION.md` in #37594.

## Worth flagging separately

`GraphqlAPITest.areFileassetFieldsPresent` uses `allMatch`, so it asserts that the asset view exposes *no property other than* the six it has today — the opposite of what its name says. It is what #35363's CI trips on, and it will block this work too. **FR-014** requires correcting it.

## Not in this PR

No implementation. Planning starts once this is approved.


specs/34540-graphql-asset-subtype-fields/spec.md
```diff
@@ -0,0 +1,324 @@
...
+
+**Type**: New Feature
+
+**Issue**: dotCMS/core#34540 — "GraphQL Support for dotAssets and FileAssets in Image/File Fields"
+
+**Input**: User description: "A client querying an Image or File field over GraphQL must be able to select the properties of the content type the field actually points at — including fields the customer defined on their own content types that extend the DOTASSET or FILEASSET base types — and must be able to tell which type it received."
+
+## Overview
+
+dotCMS lets a customer model their own kinds of assets by extending the two asset base types: a
```
nollymar: Beware that these are just examples. This should be supported in any content type extending from File Asset or DotAsset

nollymar (changes requested):
It seems like the spec is written to try to solve the issue in the provided example, not the issue in general. 

The GraphQL API should expose any dotAsset/fileAsset content type so that I can be able to "cast" any binary to that type. And in case it is not possible, I should return the results that could be casted and the warnings for those that couldn't

fabrizzio-dotCMS:
Pushed a second commit pinning down how narrowing clauses behave on an asset field — four requirements the first draft left implicit, each checked by introspection against a running instance.

**Shared properties are available at both levels (FR-015).** A client can select them directly on the field and again inside a clause that narrows to a specific type:

```graphql
image {
  identifier                       # shared, at the top
  ... on Images { tags }           # plus what's specific to this type
}
```

**Narrowing to a type the asset is not must not fail the request (FR-016).** Worth being precise: there is no cast here to fail. `... on Images` is a *condition*, not a coercion — if the asset is not an `Images`, the clause contributes nothing and the rest of the response is delivered normally. That is already how it behaves. What is being added is a **non-fatal warning** naming the clause that matched nothing, so a client can distinguish "this asset wasn't that type" from "I typed the wrong type name".

**Naming a type that does not exist must still fail (FR-017).** No valid reading, so failing loudly is correct.

**Clauses merge (FR-018).** Asking for the base kind and the specific type in one request yields one object, not two partial ones:

```graphql
... on DotAssetBaseType { asset { size } }
... on Images           { tags }
→ {"asset": {"size": 43692}, "tags": []}
```

**On hierarchy depth** — this came up as an open question and the answer is that it does not arise. Introspection shows the base-kind description sits under nothing, and a specific type names its shared descriptions side by side rather than nested:

```
DotAssetBaseType   kind=INTERFACE   interfaces=NONE
BannerImages       implements=['DotAssetBaseType', 'DotContentlet']
```

So the hierarchy is exactly two levels and cannot grow, because a content type may only extend one of the fixed base types and never another content type. A client never faces a chain of narrowing clauses — at most one for the base kind and one for the specific type, and per FR-018 those merge.

fabrizzio-dotCMS:
@nollymar thanks — you were right on both counts, and chasing the second one turned up a real gap in the proposed shape, not just the framing.

**On the framing.** Fair. The normative requirements were already general (FR-001 says "any property the customer defined on the content type of the asset that field points at"), but the Overview and every acceptance scenario were written around the banner/PDF cases, which is what makes it read as example-driven. Fixed: the Overview now leads with "any content type extending either asset base type" and states outright that those two are illustrations. User Story 1 gains two scenarios stating the general case, one of which says explicitly that passing the two example scenarios does **not** satisfy it.

**On "cast any binary to that type" — this found a defect.** I tested whether an Image field is actually constrained to image-style content. It is not:

```json
"image": {
  "fileName": "storeProductList.vtl",
  "fileAsset": { "mime": "text/plain; charset=ISO-8859-1", "size": 3128 }
}
```

That is an Image field on a real content type, pointing at a FileAsset — a `.vtl` text file — resolving fine today. The field's own kind does not constrain the base type of what it references, and the current `DotFileasset` already quietly handles both (that is what its two base-type ternaries are for).

My "shape of the solution" section typed an Image field as the image-style interface and a File field as the file-style one. That would have **silently dropped content those fields already hold**. Corrected, and pinned as requirements:

- **FR-001a** — the reachable set is every content type extending either base type, including ones a customer creates later; it may not be satisfied by enumerating known types.
- **FR-001b** — an asset-pointing field must span both base types regardless of the field's own kind, with the evidence above recorded.

**On "return the results that could be casted and the warnings for those that couldn't."** That is a collection-level statement and FR-016 was written per-asset, so I added **FR-016a**: across one result set, assets matching a clause come back populated, assets that do not still come back, the request as a whole succeeds, and a single non-matching asset must not suppress the matching ones. Warnings identify *which* clause matched nothing rather than being one opaque flag.

One related thing already in the spec, since it is the same conversation: narrowing to a type that **does not exist** still fails the request (FR-017). The "must not fail" rule is for a type that exists but is not what this particular asset turned out to be. Say the word if you would rather both be warnings.

Pushed as `92f5317c87`.

fabrizzio-dotCMS:
Correction to my last comment: the final paragraph offered to reconsider whether a non-existent type should warn instead of fail. That is not open — it is already decided. **A clause naming a type that does not exist fails the request (FR-017).** It is a client mistake with no valid reading, and a silent pass would hand back missing data with no signal.

The lenient rule applies only to a type that exists but is not what this particular asset turned out to be (FR-016 / FR-016a). No spec change; disregard that paragraph.

nollymar:
@fabrizzio-dotCMS just to keep in mind:  FR-016 and FR-016a require a non-fatal warning per clause that matched nothing, identified by clause. Standard GraphQL has no warnings slot; the realistic homes are the errors array with a path, or extensions

nollymar (dismissed):

fabrizzio-dotCMS:
@nollymar **re-review needed — an approved decision in this spec has been reversed.**

Running the ADR consultation that gates `/speckit-plan` turned up a conflict with the backward-compatibility decision you approved, so I stopped before writing the plan and changed the spec instead.

**What the ADRs say**

- **ADR-0020** (accepted) deprecated a core REST endpoint and kept it **functional**, annotated for removal, with existing integrations explicitly unaffected.
- **ADR-0022** (accepted) states the principle directly — *"a URL, parameter, or response field is a promise to callers in exactly the same way a database column is a promise to queries"* — and names why it bites here: dotCMS is self-hosted and upgraded by customers on their own schedule, so *"'nobody is calling the old endpoint right now' is not the same question as 'it's safe to remove.'"* It prescribes **expand → adopt → bake → retire**, gates removal on a supported-version floor plus a confirmed zero-use window, and requires the deprecation to be marked **in the schema itself**, not only in code.

**What changed in the spec**

| before | after |
|---|---|
| FR-012 — existing queries may break, with announcement and migration guidance | **FR-012** — existing queries keep working unchanged; the new capability ships alongside |
| — | **FR-012a** — the current view is marked superseded in the published API description, naming each property's replacement |
| — | **FR-012b** — removal is out of scope here, gated on the supported-version floor plus a zero-use window, with the retirement ticket opened up front |
| FR-009a — rename the property that would change meaning, so the break is visible | **FR-009a** — do not touch it; reach the corrected value through the new surface instead |
| — | **SC-008 / SC-009** — make the guarantee measurable |

Plus an **ADR Alignment** section. No exception to either ADR is requested.

**Why this is the right way round.** The decisive case is the same one and it now argues the other way. The asset description would have kept returning a string while changing *which* string — populated for 57 of 57 images today, 2 of 57 afterwards. A self-hosted customer who had not upgraded would have seen no failure, just different data. That is precisely the outcome ADR-0022's reasoning exists to prevent.

Nothing else changed — the generalization work from your earlier review (FR-001a, FR-001b, FR-016a) is untouched.

Pushed as `3f3465fb4c`. Planning is on hold until this is re-approved.

nollymar (dismissed):

fabrizzio-dotCMS:
@nollymar @fmontes — **re-review needed. The central decision in this spec has been reversed, and it now requests an exception to an accepted ADR.**

The version approved here (2026-09-16) described a **non-breaking** design. The implementation in #37594 took the **breaking** one, so this spec has been updated to match: an approved spec that does not describe what shipped is worse than no spec.

### What changed

| | before | now |
|---|---|---|
| FR-012 | every existing query keeps working | five of six flat properties survive unchanged; `description` and the non-asset-content case do not |
| FR-012a / FR-012b | mark as superseded, retire later behind a gate | no longer apply — nothing survives to mark, and retirement happens now |
| FR-012c | — | the break ships with announcement and migration guidance |
| ADR Alignment | "no exception to either ADR is requested" | **an exception to ADR-0022 is requested, knowingly** |

`FR-009a` survives intact, and is why the breaks take the shape they do: nothing may change value silently. A removal a client can see is acceptable; a name that keeps working and returns something else is not.

### Why, honestly

The ADR-compliant design was not skipped — it was **built and verified green**: a companion field beside each asset field (`image` gaining `imageContent`), breaking nothing, needing no spec change and no exception. It was set aside because it could not put the narrowing clauses in the same block as the flat properties:

```graphql
imageContent { ... on Images { tags } }              # the compliant option
image        { fileName  ... on Images { tags } }    # what was asked for
```

That ergonomic difference — one block instead of two — was judged by product to be worth the break. **This is a product decision, not a technical one**, and it is recorded as such in the spec.

The constraint that forces the choice is not negotiable and is worth stating plainly: a GraphQL field has exactly one type, and a resolved value has exactly one runtime type. The flat view and the asset itself are two descriptions of the same content, so they cannot occupy the same position. Keeping both means keeping them at different positions — which is exactly what the compliant design did.

### What the exception preserves of ADR-0022's intent

- **Every break is visible.** The ADR's concern is a self-hosted customer who upgrades on their own schedule and cannot tell that data changed. Both removals fail loudly.
- **The blast radius is one property plus one edge case**, not the whole surface. Five of six carry over returning identical values.
- **FR-012c** requires the announcement and migration guidance the bake window would otherwise have provided.

The decisive measurement is unchanged and now argues the other way: `description` would have kept returning a string while changing *which* string — populated for 57 of 57 images before, 2 of 57 after. Removing the name makes that visible rather than silent.

@fmontes you are asked specifically as an author of ADR-0020, which set the deprecate-and-keep-functional precedent this departs from.

Implementation, 76 tests and the full reasoning: #37594. PR #35363 is closed as superseded, its work absorbed there.

fabrizzio-dotCMS:
here an example 
Screenshot 2026-09-21 at 8 51 27 AM

fabrizzio-dotCMS:
## What the spec means for a client's query, in text

Same thing as the screenshot above, written out so it is searchable and copy-pasteable.

### Before

```graphql
query Banners {
  BannerCollection(limit: 10) {
    title
    image {
      fileName
      description
      fileAsset { versionPath size mime width height }
      metaData { key value }
    }
  }
}
```

### After — minimally corrected. One line.

```graphql
query Banners {
  BannerCollection(limit: 10) {
    title
    image {
      fileName
      title                      # was `description`: same value, honest name
      fileAsset { versionPath size mime width height }
      metaData { key value }
    }
  }
}
```

**That is the entire migration for a typical client**, and the reason is the measurement in the spec: `description` on image-style content never returned a stored description — it returned the contentlet *title*, which for an asset is the file name. Populated for 57 of 57 images, against 2 of 57 for the stored value. So `title` renders exactly what the page was already rendering.

The trap worth naming: `... on Images { description }` *looks* like the faithful translation and is not. It returns the stored value, usually empty. A front-end using the old field as alt text would go blank.

### After — what the capability actually buys

```graphql
image {
  fileName  versionPath  size  mime  width  height   # the binary, no extra level

  identifier  title  live  urlMap                     # the asset's own identity
  __typename                                          # which content type it is

  ... on Images       { tags description }            # customer-defined properties
  ... on BannerImages { campaignName adSize }
}
```

The middle two groups were not selectable before at any depth, and the last group is the limb of #34540 that had no workaround at all — it is what blocks the AI tagging workflow.

### A mixed collection

```graphql
image {
  baseType                                        # DOTASSET or FILEASSET, no clause needed
  __typename
  ... on DotAssetBaseType { asset     { size mime } }
  ... on FileBaseType     { fileAsset { size mime } }
}
```

Each row answers with only the clause that applies; a clause that does not match contributes nothing and does not fail.

---

**To restate what is being asked here, since this PR is the spec and not the code:** the break is decided — a product decision, recorded as such. What this PR needs is @fmontes and @nollymar signing off on the **exception to ADR-0022**, which prescribes deprecate-then-retire and which this departs from knowingly. The queries above exist to make the cost of that concrete rather than abstract.

Migration guide for clients: `docs/backend/GRAPHQL_ASSET_FIELDS_MIGRATION.md` in #37594.

nollymar:
@fabrizzio-dotCMS , can you please verify if these comments are still valid? 

- The null case vs. FR-009a/SC-009. A field pointing at non-asset content used to return the flat view and will now return null. A query that validates and runs but returns different data is exactly what SC-009 says never happens (“zero selections keep working while returning different data”). Either count it as an exception in SC-009 or require a warning, as FR-016 does.
- The PR description is stale. It still says “#34540 stays open … carrying the second stage” and “Its benefit does not survive this feature”. The spec now delivers the binary convenience here and closes the issue.
- Leftover inconsistency: Assumptions line 482 still says “The five surviving flat properties”. It should say six.
- Edge cases with no FR: archived/deleted targets, language fallback, collisions between customer property names and general asset properties, deleted properties, and not leaking permission-restricted type names. They're listed but have no requirement or success criterion. The collision case in particular needs an actual rule.
- FR-016 warnings are non-standard GraphQL. Spec-compliant servers silently skip fragments that don't match. In a mixed collection every row would warn for every clause that doesn't match. Define the payload shape (in extensions?), whether it's deduplicated, and whether it can be turned off.

fabrizzio-dotCMS:
@nollymar all five were still valid — thanks. Addressed in the latest spec push:

1. **Null case vs. SC-009** — counted as the explicit second exception in SC-009, and added to the FR-012c announcement. No warning: the misconfiguration is in stored data, which the client developer can't fix.
2. **Stale PR description** — rewritten: the binary convenience ships here (10 of 12 properties; `title`/`modDate` stay under the binary), and #37594 fixes #34540.
3. **"five surviving"** → six.
4. **Edge cases** — now FR-019 (archived/deleted/unreadable/language: resolution unchanged, `null` for that field), FR-020 (deleted property: standard validation error naming it), FR-021 (collisions), FR-022 (permissions), measured by SC-010.
   The collision one was worse than a missing rule: a customer asset type with e.g. a text `size` makes the asset interface invalid and graphql-java rejects the **whole schema**. Rule in FR-021: customer property wins on its name, general value stays under `asset {}`/`fileAsset {}`, incompatible names are dropped from the interface at build time, and new colliding fields are refused on save. Fix goes into #37594.
5. **FR-016 warnings** — defined: `extensions.warnings`, `{path, typeCondition, message}`, deduplicated per request (a clause is reported only if nothing in the whole response matched it, so a mixed collection doesn't warn per row), inline fragments only, not switchable since it's inert for clients that ignore `extensions`.

fabrizzio-dotCMS:
@nollymar follow-up on the name-collision rule (FR-021), now implemented in #37594:

When a customer asset type has a field named like one of the flattened asset properties (`name`, `size`, `width`, …):

- **The customer's field wins its name** on its own type. The binary's value stays reachable through `asset { }` / `fileAsset { }`.
- **Same GraphQL type** (e.g. a text `sha256`): `image { sha256 }` returns the customer's value for that type.
- **Different GraphQL type** (e.g. a text `width` vs. the binary's `Long`): `width` is left off the asset interfaces for the instance, so the schema stays valid. `image { width }` fails validation naming `width`; `... on BannerImages { width }` and `asset { width }` keep working. A warning is logged at schema build naming the type and field.
- **On save**, a generated field variable is steered away from an incompatible name ("Width" → `width1`), same as dotCMS already does for the general content properties. An explicitly chosen variable is not rejected — dotCMS doesn't reject explicit variables anywhere today — so the build-time rule covers it.

Correction to my previous comment: I said new colliding fields are "refused on save"; they are steered, not refused.

nollymar (dismissed):

nollymar (approved):

PR #37423, opened 2026-09-05
Dojo to Angular: dotAI Portlet (#37417)

fmontes:
https://github.com/user-attachments/assets/22dc8b74-25ee-42a0-b0b8-96cec27b8b89

**Spec-Kit PR 2 of 2 — the implementation.** Spec approved in #37418; closes #37417.

Until #37418 merges this diff also carries the spec commit — that is the shared ancestor, not a duplicate.

## What this is

The dotAI portlet, rebuilt in Angular. Five tabs — Search, Chat, Image, Embeddings, Config Values — rendering natively in the admin shell instead of the legacy iframe. `dotai` takes over its existing menu entry so upgrades need no manual step; an unlisted `dotai-legacy` twin keeps the old JSP reachable for rollback.

**No Java.** The only server-side edits are `portlet.xml` and `Language.properties`.

## How to review

The commits are ordered to be read in sequence, and each is self-contained:

| Commit | What |
|---|---|
| `cb84f85`, `1f2fcd3` | Registration + blank routed shell, validated in a browser before any logic |
| `75855a1`, `3b33dd7` | `DotAiService` split + the search / embeddings / stream clients |
| `bbdea36`, `1e51fae` | US1 Search + the shared retrieval-settings panel |
| `ae175de` | US2 Chat (streaming, Stop) |
| `9387dfe` | US3 Embeddings |
| `795ab79` | US4 Image, US5 Config Values |
| `aa587f2`, `2ef2d65` | Convergence findings closed |
| `78254e3` | Four stale instructions corrected in the portlet docs |

**Start with `75855a1`** — the service split is a pure refactor and the foundation everything else sits on. Its carried-over specs passing unchanged is what makes it provably behavior-preserving.

## Six defects browser e2e caught that no unit test could

Validated against a live instance with a real provider throughout, and it earned its keep:

1. **`Accept: text/event-stream` → HTTP 406.** The completions endpoint is a JAX-RS `StreamingOutput`, not SSE. A stubbed `fetch` never negotiates content, so chat would have shipped broken with a green suite.
2. **A bare JSON error with no SSE framing.** When retrieval matches nothing the endpoint answers `{"error":"no matching content found..."}` unframed. The parser dropped it and the user got an empty answer with no explanation.
3. **`[attr.aria-label]` on `<p-button>` lands on the host**, leaving the real `<button>` unnamed — the icon-only Send button had no accessible name.
[...]

PR #37541, opened 2026-09-14
fix(dotai): design and QA follow-ups from the Angular migration (#37538)

### LINKED ISSUE #37538: dotAI portlet: design + QA follow-ups from the Dojo to Angular migration
## Description

Design and QA review of the new Angular dotAI portlet (#37417) turned up seven follow-ups — five UI/copy corrections and two bugs in the Embeddings tab. Grouping them into one issue since they all land in the same portlet and are best fixed and QA'd together.

1. **Empty states don't match the platform.** The portlet rolled its own empty state instead of the `dot-empty-container` every other portlet uses. Replace it everywhere.
2. **`Fields` and `Velocity template` need real descriptions.** The Velocity template has no hint and no placeholder at all, and the Fields hint is wrong — it says "leave empty to embed the whole contentlet" when dotCMS actually picks a subset of fields. The two are also alternatives, not complements: a Velocity template overrides Fields entirely. None of that is visible in the dialog. Needs a hint on both plus an example placeholder on the template.
3. **Remove "View provider config".** The button and its JSON dialog in Config Values aren't needed.
4. **"Rebuild DB" is red.** Should be a regular outlined button — the destructive step is already the confirm dialog.
5. **The Config Values host line is unclear.** It reads `demo.dotcms.com (falls back to system host)`, which doesn't say what it means and isn't translatable.
6. **Build warnings appear behind the modal.** Submitting the Build an index dialog closes it, then shows the outcome — including Lucene query errors — in the tab behind it. The user loses what they typed and has to start over. Keep the modal open and show it there.
7. **A new index doesn't appear in the table.** After building, the row only shows up after a manual page reload.

## Acceptance Criteria

- [ ] Every dotAI empty state uses `dot-empty-container` from `@dotcms/ui` and matches the empty states in other portlets
- [ ] `Velocity template` has a hint and an example placeholder; the `Fields` hint is corrected, and it's clear a template overrides Fields
- [ ] The `View provider config` button and dialog are gone
- [ ] `Rebuild DB` renders as a regular outlined button
- [ ] The Config Values host line reads clearly and is translatable
- [ ] Building an index keeps the modal open and shows success, "nothing matched", and query errors inline, with the typed values preserved; it closes only on success
- [ ] A new index appears in the Embeddings table as `Building` right after submit and flips to `Ready` on its own, with no page reload

## Additional Context

Follow-up to #37417 (Dojo to Angular: dotAI Portlet), implemented in #37423. Feedback from Jorge Conde (9/11) and melissa.rojas (9/10).

Items 6 and 7 are defects; 1–5 are UI/copy. Items 2 and 5 touch the backend; the rest are frontend-only.

### LINKED ISSUE #37543: dotAI: split the build/delete dialog and fix silent index actions
## Description

The Embeddings tab packs two opposite actions into one "Build an index" dialog, and none of the destructive ones tell you what they did. Splitting them and adopting the admin's normal feedback pattern fixes both.

1. **Split the dialog.** The Mode toggle makes the form reshape under the user — Fields and Velocity template disappear when you switch to Delete — and puts a destructive action one click from a create action, sharing the same submit button. "Build an index" should only build.
2. **Move "delete from index" onto the index row.** It always targets exactly one index, yet today you type the index name into a free-text box: a typo silently deletes nothing, a near-miss hits the wrong index. Launched from the row, the name comes from the row and that whole problem goes away. The row's trash icon becomes a small menu — *Remove content…* and *Delete index*.
3. **Say what it actually does.** The dialog states none of this: it removes embeddings, not content; the query matches content *as it is now*, so content that changed and no longer matches is left orphaned and archived or deleted content can't be reached by it at all.
4. **Actions are silent.** Rebuild DB, Delete index and Delete from index all give no feedback whatsoever — you confirm something destructive and get nothing back. Delete from index already returns a count from the server; it is thrown away.
5. **Use the admin's feedback pattern.** The house rule elsewhere is a toast for transient success and inline messages for failures that need a field corrected. dotAI instead leaves a dismissible banner sitting on the Embeddings tab for a successful build, which no other portlet does.

## Open question for the backend

Add and delete do not use the same query semantics: building appends `+live:true` to your query, deleting appends nothing. So the same string typed into the two does not select the same content. Either the two should match, or the difference is deliberate — letting you clean up embeddings for content that is no longer live — and the dialog should say so. Needs a decision before the copy in item 3 can be written.

## Acceptance Criteria

- [ ] "Build an index" has no Mode toggle and only builds
- [ ] Removing content from an index is launched from the index row, with the index as context rather than a typed field
- [ ] That dialog explains that it removes embeddings rather than content, and what its query actually matches
- [ ] Rebuild DB, Delete index and Delete from index all report their outcome, including when nothing matched
- [ ] Success uses the shared toast and failures that need a field corrected stay inline in their dialog, matching the rest of the admin
- [ ] The Embeddings tab no longer carries a persistent banner for a successful build

## Additional Context

Follow-up to #37538 / #37541, which fixed the design and QA round on the Angular dotAI portlet. These came out of reviewing how index editing works: there is no way to edit an index, because nothing persists what built it, so add and delete are the only tools and they need to be clear.

Related and worth its own issue: index builds store none of their inputs — not the Lucene query, the fields, the Velocity template or the model — so an index cannot be inspected, edited or re-run, and builds are additive, meaning re-running with a narrower query silently leaves the old content behind.

fmontes:
Design and QA review of the new Angular dotAI portlet (#37417, implemented in #37423) raised seven items. Researching one of them — how you edit what is in an index — turned up that you cannot, because nothing persists the query that built it, which made the build/delete dialog worth fixing here rather than later (#37543).

Each section below has a slot for its screen recording. Every clip is against a locally-built container on the demo starter.

---

## 1. Empty states use the shared component

The portlet rolled its own `dot-ai-empty-state` — a bare title, no icon — so its tabs looked nothing like the rest of the admin. Deleted, and all eight call sites (Search ×3, Chat, Image, Embeddings ×2, Config Values ×2) now render through `dot-empty-container` from `@dotcms/ui`.

They are centred now too, which the first pass got wrong: `dot-empty-container` centres itself with `h-full`, and a percentage height needs a parent that has one. Every call site nested it under an auto-height wrapper, so each state measured only as tall as its own text and sat at the top of a blank pane. Search and Chat move theirs out of the `container mx-auto` column into the scroll box; Embeddings and Config Values replace the table entirely rather than sitting in a `<td>`, which is sized by its own content.

[01-empty-states-use-the-shared-component.webm](https://github.com/user-attachments/assets/c7889002-a2c8-440e-9b8d-69541493167f)

## 2. `Fields` and `Velocity template` explain themselves

`Velocity template` had **no hint and no placeholder at all**, and the `Fields` hint was wrong on both counts.

The two are not independent, which is the part nobody could have guessed: `BulkEmbeddingsRunner.run()` tries the template **first** and only falls back to the field path when it renders nothing. So a template **silently overrides `Fields`**, and leaving both empty makes dotCMS guess (WYSIWYG, Story Block, Textarea, file, binary) rather than "embed the whole contentlet". Both hints say so now, and the template carries a worked example.

[02-fields-and-velocity-template-copy.webm](https://github.com/user-attachments/assets/c789d88b-dacc-414d-984b-e671309bf147)

## 3. `View provider config` removed

The button, its JSON dialog, the two message keys, and the `maskCredentials` helper that had no other caller.

[03-view-provider-config-removed.webm](https://github.com/user-attachments/assets/a5dbab76-23c9-41c7-aea9-5fac278fe5d2)

## 4. `Rebuild DB` is a plain outlined button

A permanently-red control in the toolbar read as a warning about the screen. The destructive step is the confirm dialog it opens.

[04-rebuild-db-is-a-plain-outlined-button.webm](https://github.com/user-attachments/assets/17c7c39e-9e75-4fbf-abbb-f486dbf148a9)

## 5. The Config Values host line says what it means

It read `demo.dotcms.com (falls back to system host)`, assembled server-side by concatenating the hostname with an English suffix — untranslatable, and it claimed the fallback **whether or not it had happened**.

`GET /api/v1/ai/completions/config` now returns `configHost` and `configHostInherited` as separate fields, derived from whether `ConfigService` actually fell back to the System Host's secrets, and the client renders the label from a message key.

[05-config-host-message-is-clear.webm](https://github.com/user-attachments/assets/edcd45e0-8114-45ef-b72d-2eaaf084be16)

---

## 6. Build errors stay in the modal

Submitting Build an index closed the dialog and reported the outcome on the tab underneath — so a rejected Lucene query arrived **after** the modal had taken the query away with it, and you had to reopen and retype.

The dialog now owns its submit, renders the failure and the "nothing matched" warning inline with the query still in the field, and closes only on success. It also cannot be dismissed at all while its request is outstanding — the answer is coming back *into this form*, so it has to still be there to receive it.

[06-build-error-stays-in-the-modal.webm](https://github.com/user-attachments/assets/2ade6828-9c99-47d4-a323-a1fbb19bd954)

## 7. A new index appears without a page reload

`buildIndex` called `markIndexBuilding` and then `loadIndexes`, and that refresh **erased the flag it was meant to act on**: `applyIndexes` rebuilt the status map from the server's list alone, and embedding is asynchronous so the new index is not in `indexCount` yet. With no `BUILDING` status left, the poll never started, so nothing re-fetched and the row only showed up on a manual reload.

Seeded builds now hold a placeholder row until the server catches up, the fragment snapshot is taken from the server's response only, and a grace period stops the poll if a build never materialises.

[07-new-index-appears-without-a-reload.webm](https://github.com/user-attachments/assets/ca696478-a631-4e97-8e2e-93c371f5acae)

---

## 8. Build and remove are separate dialogs

The Mode toggle made the form reshape under the user (Fields and Velocity template vanished on switching) and put a destructive action one click from a create action, sharing its submit button. *Build an index* only builds.

Removing content is its own toolbar action beside New Index — the two are the same shape of operation, a Lucene query plus an index to scope it to. It is deliberately **not** a row action: nothing records the query that built an index, so the query is written blind either way, and hanging it off a row would promise context the backend cannot give. The index is a picker over the real list rather than a free-text field, and nothing is preselected, because this removes embeddings.

The dialog also states three things the old delete mode said nowhere: it removes embeddings and not content; the query matches content *as it is now*, so content that changed is left behind and archived content cannot be reached; and unlike building it is not limited to live content.

[08-build-and-remove-are-separate-dialogs.webm](https://github.com/user-attachments/assets/b098ecaf-fc0a-4169-866a-d67372c1c7e2)

## 9. Every action reports what it did

Rebuild DB, Delete index and Remove content were **completely silent** — three of the four returned a count the store dropped on the floor. One `indexNotice` channel now carries the operation and its outcome, because where an outcome belongs is decided by the outcome rather than the action: anything the user must act on stays in the dialog that submitted it, everything else is a toast. A removal that matched nothing is `empty`, not success — the server answers 200 with `deleted: 0`.

Toasts, as `dot-locales` and `dot-experiments` do it. The tab no longer carries a standing banner for a successful build.

[09-every-action-reports-what-it-did.webm](https://github.com/user-attachments/assets/3ab3ddf4-9a2e-4a96-acbd-4185e4757a07)

---

## Testing

- `nx test portlets-dot-ai-portlet` — 283 passing
- `nx affected -t test` — 2 267 passing across every affected project
- `CompletionsResourceTest` — 4 new unit tests on the backend rule, including the "nothing configured anywhere" branch it exists for
- `nx affected -t lint` clean across 36 projects; `nx format:check` clean
- Verified in a browser against a locally-built container throughout

`block-editor` has 35 test failures that also fail on `main`; confirmed unrelated by re-running it with this branch's shared-lib changes reverted.

## Notes for the reviewer

**Two review rounds ran over this branch and both found real defects**, which are fixed and pinned by tests that fail without them. Worth knowing where they were, since they all lived in the plumbing rather than the user-facing fixes:

- A deleted index came back as a ghost `Building` row for two minutes, offered in the retrieval picker and eligible to become the selected index — so Search and Chat could point at an index that no longer existed.
- A dialog reopened over an in-flight request closed on the *previous* request's outcome while `exhaustMap` silently dropped the new one.
- A 403 left a modal nobody could close but a page reload, because the dialog settles only on a notice and that path emitted none.
- Outcomes no open dialog would render were suppressed anyway, so a delete that failed while another dialog was up was never reported.

**Open question for whoever owns `EmbeddingsResource`:** building appends `+live:true` to your query and deleting appends nothing, so the same string does not select the same content. The remove dialog states the asymmetry rather than deciding it.

**Worth its own issue:** index builds persist none of their inputs — not the query, the fields, the template or the model — so an index cannot be inspected, edited or re-run, and builds are additive, meaning re-running with a narrower query silently leaves the old content behind.

**Known and deliberate:** deleting an index while its build is still in flight lets the build's response re-seed it; reaching it needs a portlet-tab switch, since the dialog is modal-locked during flight. And the dialog's header X returns on the user's next interaction rather than instantly after a failure — `closable` is a plain property the host reads in a template binding, and the unlock runs inside an effect already part of the pass in progress. Cancel and Escape are live throughout.

Fixes #37538. Closes #37543.

This PR fixes: #37538


fmontes:
Follow-up filed as #37543 — splitting the build/delete dialog, moving "delete from index" onto the index row, and fixing the silent destructive actions (Rebuild DB, Delete index and Delete from index all currently report nothing).

Scope of this PR is unchanged; #37543 came out of researching how index editing works and is better reviewed on its own.

fmontes:
Both Medium findings fixed in `51869fe`.

**`index.name in seeds` walking the prototype chain** — confirmed reachable before fixing, not just plausible:

```
'constructor' in seeds  => true      ← reads as seeded when it is not
hasOwnProperty.call     => false
```

The seed map comes from `Object.fromEntries`, so it carries `Object.prototype`, and `INDEX_NAME_PATTERN = /^[a-zA-Z0-9_-]+$/` accepts all four colliding names. Now `Object.prototype.hasOwnProperty.call(seeds, index.name)`.

You were right that it is a regression rather than a latent quirk: the code it replaced used `Set.has`, and the exposure arrived with the commit that made `indexStatuses` a computed. Pinned with a test over `constructor` / `toString` / `valueOf` / `hasOwnProperty`, which I verified fails against the old `in` check before keeping it.

Worth noting for anyone reading later: the blast radius stops at the badge. `liveSeeds` and `stillBuilding` iterate with `Object.entries`, `withPendingIndexes` uses a real `Set`, and the poll guard is `Object.keys(...).length > 0` — so a colliding name rendered a permanently wrong status but never started a runaway poll or produced a phantom row.

**Stale `BUILD_SEED_TTL_MS` reference** — corrected to `PENDING_SEED_TTL_MS`. I renamed the constant when its meaning narrowed to the window before the server lists an index, and missed the reference in the state model's doc block.

**The M-3 rollback flag** is being taken as a documented caveat rather than actioned, which matches your own read of it: the field is additive, the client defaults it with `raw?.configHostInherited ?? false`, and the worst case after a rollback is a redundant label until the next page load. Flagging it in the release notes instead of adding a compatibility branch for one release.

284 tests passing, lint and format clean.

rjvelazco (approved):

KevinDavilaDotCMS (commented):
Non-blocking review — nothing here should hold the PR up. Left a couple of edge cases and some nits inline; the rest is praise, because there is plenty of it.

**What I really liked**

- `watchIndexOperation` is the best decision in the PR: matching on operation *and* `indexName` kills the whole class of "dialog settles on someone else's outcome" bugs, and documenting the exception pattern in `libs/portlets/CLAUDE.md` is exactly where it belongs.
- I traced the dialog DI chain (PrimeNG 21 `DynamicDialogInjector` -> the tab's node injector -> the shell's `DotAiStore`): the claim/notice machinery resolves the **same store instance** in production, not just under test. The ownership design actually holds at runtime.
- `isInheritedConfig` semantics check out against `ConfigService.config(host)` — including the blank-`providerConfig` gate for the "nothing configured anywhere" case, which `CompletionsResourceTest` covers explicitly. Nice to see the fallback path of `ConfigService` reasoned about rather than assumed.
- `hasOwnProperty` vs `in` in `indexStatuses` is a genuine catch — the name pattern accepts `constructor`/`toString`/`valueOf`/`hasOwnProperty`, and any of those would have rendered `Building` forever.
- Scoping `PENDING_SEED_TTL_MS` to the window *before* the index is listed is the subtle right call: a start-of-build timer would have flipped large builds to Ready mid-flight, which is the exact failure this feature exists to prevent.
- The empty-state rework is thorough: all eight call sites through `dot-empty-container`, with the `h-full`-needs-a-parent-with-height reasoning captured at each restructured call site.
- Backward compatibility handled on both ends (`configHostInherited ?? false` client-side) and the `{0}/{1}` arg swap in `build.ok` applied consistently across every call site, including the two dialog templates.

**One thing to confirm**

- Ticket #37543 AC #2 asked for "remove from the index **row**"; this PR deliberately puts it in the toolbar, with a reasoned argument (nothing records the query that built an index, so row context would promise something the backend cannot give). The reasoning is sound — just worth an explicit sign-off from QA/product on the deviation, since it is the only AC not met literally.

The two medium-ish items inline (stale notice re-toast on tab re-entry, and the dialog that cannot be dismissed if the store dies mid-flight) are both edge cases — happy to file follow-up issues for either.

core-web/libs/portlets/dot-ai/src/lib/tabs/dot-ai-embeddings/dot-ai-embeddings.component.ts
```diff
@@ -52,79 +98,148 @@ const CONFIRM_BUTTONS = {
...
 
     readonly #confirmationService = inject(ConfirmationService);
+    readonly #toast = inject(MessageService);
     readonly #dialogService = inject(DialogService);
     readonly #messageService = inject(DotMessageService);
 
     protected readonly statuses = DOT_AI_INDEX_STATUS;
 
+    /** Identity guard: each operation makes a fresh notice, so this toasts each one once. */
+    #toasted: DotAiIndexNotice | null = null;
```
KevinDavilaDotCMS, line 126: Non-blocking (edge): `#toasted` is a plain field on this component, but `indexNotice` lives on the shell-scoped store and nothing clears it. Leaving the tab and coming back re-runs this effect with the previous notice still in the store, so the last outcome — a success or a failure the user has already seen — toasts a second time.

Two cheap options: clear `indexNotice` / `indexNoticeOwner` in a `DestroyRef.onDestroy` here, or move the "already reported" guard into the store so it survives tab switches. Fine as a follow-up; it is a UX wart, not a correctness bug.

core-web/libs/portlets/dot-ai/src/lib/utils/dot-ai-index-dialog.utils.ts
```diff
@@ -0,0 +1,143 @@
...
+        }
+
+        // Unbound here rather than at teardown: DynamicDialog is destroyed only after its
+        // leave animation, and a live handler on a dialog already closing would close it twice.
+        unbindEscape();
+        deps.close();
+    }
+
+    documentRef.addEventListener('keydown', onEscape);
+    inject(DestroyRef).onDestroy(() => {
```
KevinDavilaDotCMS, line 98: Non-blocking (edge), and the mirror image of the "cannot dismiss while submitting" design: if the user navigates out of the portlet while a request is in flight, the shell — and with it the store — is destroyed, the `rxMethod` subscription dies, and the notice never arrives. The dialog is then stuck `$submitting` forever: no header X, no Escape, Cancel disabled, floating over whatever portlet came next.

Only reachable by clicking the sidebar mid-flight on a slow request, so not a merge blocker. A router-event listener that releases the dialog when the route that opened it is gone would close the hole — happy to file the issue.

core-web/libs/portlets/dot-ai/src/lib/tabs/dot-ai-embeddings/dot-ai-embeddings.component.ts
```diff
@@ -35,6 +40,47 @@ import { DotAiStore } from '../../store/dot-ai.store';
  * itself and the theme defines no `p-button-primary` to ask for — `p-button-secondary` exists
  * there, `p-button-primary` does not. Setting one would resolve to nothing.
  */
+/**
```
KevinDavilaDotCMS, line 43: Nit: the Javadoc block *above* this one (the "Button treatment shared by both confirmations…" one) belongs to `CONFIRM_BUTTONS` further down, but as laid out it reads as the doc for `NOTICE_MESSAGE_KEYS`. Re-attaching it to the const it describes would keep the pairing unambiguous. (Same pattern in `with-ai-indexes.feature.ts`, where the `markIndexBuilding` doc now sits above `forgetIndexBuildSeeds`.)

core-web/libs/portlets/dot-ai/src/lib/store/features/with-ai-indexes.feature.ts
```diff
@@ -112,13 +197,46 @@ export function withAiIndexes() {
                  * "a build just started here", so the first poll does not have to infer it
                  * from a delta that has not appeared yet.
                  */
+                /**
```
KevinDavilaDotCMS, line 200: Nit: the Javadoc block above this one ("Marks an index as building straight away…") documents `markIndexBuilding`, but after the insertion it sits directly on top of `forgetIndexBuildSeeds`'s own doc. Moving it down to its method keeps the two from reading as one double-stacked comment.

core-web/libs/portlets/dot-ai/src/lib/tabs/dot-ai-embeddings/dot-ai-embeddings.component.ts
```diff
@@ -52,79 +98,148 @@ const CONFIRM_BUTTONS = {
...
+        severity: string;
+        summary: string;
+        detail: string;
+        life: number;
+    } {
+        const severity = NOTICE_SEVERITY[notice.outcome];
+
+        return {
+            severity,
+            summary: this.#messageService.get(`dotai.embeddings.toast.${severity}`),
```
KevinDavilaDotCMS, line 233: Nit: an interpolated key here — which the comment on `NOTICE_MESSAGE_KEYS` a few lines up names as the exact thing to avoid (invisible to grep, reported as orphaned by a copy audit). Consistent alternative:

```ts
/** Toast summary key per outcome — spelled out for the same reason NOTICE_MESSAGE_KEYS is. */
const TOAST_SUMMARY_KEYS: Record<DotAiIndexNotice['outcome'], string> = {
    ok: 'dotai.embeddings.toast.success',
    empty: 'dotai.embeddings.toast.warn',
    failed: 'dotai.embeddings.toast.error'
} as const;
```

then `summary: this.#messageService.get(TOAST_SUMMARY_KEYS[notice.outcome])`.

dotCMS/src/main/java/com/dotcms/ai/rest/CompletionsResource.java
```diff
@@ -193,9 +193,12 @@ public final Response getConfig(@Context final HttpServletRequest request,
         final AppConfig appConfig = ConfigService.INSTANCE.config(host);
 
         final Map<String, Object> map = new HashMap<>();
-        map.put(AiKeys.CONFIG_HOST, host.getHostname() + " (falls back to system host)");
-
         final String providerConfig = appConfig.getProviderConfig();
+        final String requestedHost = host.getHostname();
+
+        map.put(AiKeys.CONFIG_HOST, requestedHost);
+        map.put(AiKeys.CONFIG_HOST_INHERITED, isInheritedConfig(requestedHost, appConfig));
```
KevinDavilaDotCMS, line 200: Minor: the `@Operation` description on `GET /config` predates `configHostInherited`. The response schema is `Map.class`, so nothing breaks, but per our REST conventions the annotation is the contract — worth a sentence noting the response now carries `configHost`, `configHostInherited`, `settings` and `providerConfig`. (Will require regenerating `openapi.yaml` alongside, per the usual flow.)

dotCMS/src/main/webapp/WEB-INF/messages/Language.properties
```diff
@@ -8708,9 +8708,27 @@ dotai.chat.stopped=Answer stopped.
 dotai.chat.thinking=Reading your content
 dotai.chat.error.empty=No answer came back. Try asking again, or loosen the closeness threshold so the index returns something to answer from.
 dotai.embeddings.filter.placeholder=Filter indexes
-dotai.embeddings.build.ok=Indexed {0} items into "{1}".
+dotai.embeddings.build.ok=Indexed {1} items into "{0}".
 dotai.embeddings.build.empty=Nothing matched that query, so "{0}" was created with no content. Check the query and build again.
 dotai.embeddings.build.failed=Could not build "{0}". {1} Check the query syntax and try again.
+dotai.embeddings.delete.ok=Deleted "{0}" and the {1} embeddings in it.
```
KevinDavilaDotCMS, line 8714: Tiny copy edge: deleting an empty index renders `Deleted "x" and the 0 embeddings in it.` Not wrong, just slightly odd — something count-agnostic like `Deleted "{0}" and its embeddings ({1}).` would read better in that corner case.

core-web/libs/portlets/dot-ai/src/lib/store/features/with-ai-indexes.feature.ts
```diff
@@ -39,27 +58,86 @@ export function withAiIndexes() {
...
+                return Object.fromEntries(
+                    store.indexes().map((index) => [
+                        index.name,
+                        // `hasOwnProperty`, not `in`: the seed map comes from
+                        // `Object.fromEntries`, so `in` walks its prototype and an index
+                        // named `constructor`, `toString`, `valueOf` or `hasOwnProperty`
+                        // reads as seeded when it is not — then renders Building forever,
+                        // since nothing clears a seed that does not exist. The create
+                        // form's name pattern accepts all four.
+                        Object.prototype.hasOwnProperty.call(seeds, index.name)
```
KevinDavilaDotCMS, line 82: Love this one. An index literally named `constructor` would have sat at `Building` forever, and this is the kind of bug only an adversarial test catches — the spec that pins it reads like documentation.

core-web/libs/portlets/dot-ai/src/lib/store/features/with-ai-indexes.feature.ts
```diff
@@ -15,14 +15,33 @@ import { computed, inject } from '@angular/core';
...
+ *
+ * Deliberately scoped to the window *before* the index is listed at all. Once it appears, the
+ * fragment-count delta owns its lifecycle and no clock is involved — a timer measured from the
+ * start of the build would cut a large one off mid-flight, flipping a still-growing index to
+ * Ready and stopping the poll, which is the exact failure this feature exists to prevent.
+ *
+ * Embedding normally writes its first rows within a second or two, so a build still unlisted
+ * after two minutes failed somewhere the client cannot see.
+ */
+const PENDING_SEED_TTL_MS = 2 * 60 * 1000;
```
KevinDavilaDotCMS, line 43: Scoping this TTL to the window *before* the index is listed is the subtle right call, and the comment explaining why a start-of-build timer would be wrong is exactly the context a future maintainer needs. Well done.

KevinDavilaDotCMS (commented):
Inline comments for the summary review above — all non-blocking.

KevinDavilaDotCMS (approved):

Issue #37464, opened 2026-09-08
Standardize field hint and required-error presentation in the new Edit Contentlet


erickgonzalez:
## Description

In the new Edit Contentlet, the **hint** and the **required** indicators need to follow a single, consistent design across every field type. Today each field component renders its own label/hint/error markup, in three different variants, with behavior that diverges from the target design.

### Target design

<img width="1230" height="824" alt="Image" src="https://github.com/user-attachments/assets/796b9177-733a-4506-a0c2-b6afbbebe63f" />

Four states, using a Date and time field as the example:

| # | State | Presentation |
|---|---|---|
| 1 | No hint, no error | Bold label + **red asterisk** (required), input |
| 2 | Hint present, no error | Same, plus hint text **below the input** |
| 3 | No hint, error on save | **Red border** on the input, `This field is required` in red below it — **no icon** |
| 4 | Hint present, error on save | Red border, red required message below the input, **then** the hint below that |

State 4 is the key behavioral change: the error message and the hint are shown **together**, error first, hint second. Today the hint is suppressed whenever an error is present.

### Current gaps

1. **Hint is suppressed on error.** Every field template guards the hint with `!fieldHasError`, so state 4 is unreachable.
2. **Hint sometimes renders as a tooltip icon, not text.** `dot-card-field-label` accepts a `hint` input and renders a `pi pi-info-circle` tooltip next to the label. The Date/Time field routes the hint into that tooltip whenever a timezone is shown, and the Block Editor uses it unconditionally.
3. **Three duplicated markup variants**, none shared:
   - `<small class="p-field-error">` / `<small class="p-field-hint">` — Text, Textarea, Select, Radio, Checkbox, Multi-select, Date/Time, Custom Field
   - `<div class="error-message">` / `<div class="hint-message">` — Tag, Host/Folder, JSON, Key-Value, WYSIWYG
   - Field-specific markup — Relationship (`.error-message` + own testid), File/Image/Binary, Category (no hint/error slot at all)
4. **`.error-message` / `.hint-message` have no styles** anywhere in `core-web`. Those fields render their required message in default body color, not red. Only `p-field-error` / `p-field-hint` are styled (`apps/dotcms-ui/src/style.css`, scoped under `.form`).
5. **Error appears too early.** `BaseWrapperField` sets `$hasError` from `control.invalid && control.touched`, so a required field turns red as soon as the user tabs out of it — before any save or publish attempt.

### Relevant code

| Concern | Location |
|---|---|
[...]

PR #37542, opened 2026-09-14
docs(edit-content): spec for #37465 Date/Time field width, clearing and picker footer

### LINKED ISSUE #37465: Improve Date, Time, and Date-and-Time field width, clearing, and picker footer in the new Edit Contentlet
## Description

The **Date**, **Time**, and **Date and time** fields in the new Edit Contentlet need three presentation improvements: they should fill the width of their column, they should be clearable directly from the field, and the picker overlay's footer should show the timezone on the left with a `Today` button on the right.

### 1. Field does not fill its column

The date/time control is narrower than its column while every other field type stretches to the full column width.

**Cause:** `calendar-field.component.scss` sets `display: inline-flex` on the `p-datepicker` wrapper, so the control sizes to its content instead of its container. Sibling fields do stretch — e.g. the Text field puts `class="w-full"` on its `<input>`.

```scss
// core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.scss
:host ::ng-deep p-datepicker {
    // Use PrimeNG's default datepicker sizing
    display: inline-flex;   // <-- keeps the field from filling the column
    ...
}
```

### 2. No clear control on the field

`showClear` is hardcoded to the content type's expire-date field only, so every other date/time field has no way to empty a value once set:

```html
<!-- calendar-field.component.html -->
[showClear]="$isExpireDateField()"
```

The mock shows a clear (X) icon between the value and the calendar button on an ordinary `Posting Date` field.

### 3. Picker footer

`[showButtonBar]="true"` renders PrimeNG's default button bar — **`Today` on the left, `Clear` on the right**. The target footer is:

| Slot | Content | Style |
|---|---|---|
| Left | Timezone | Gray, muted text — not a button |
| Right | `Today` (`Now` for Time-only) | Secondary button, **outlined** |

The `Clear` button leaves the footer entirely — clearing is now the on-field X from #2.

### Interaction with #37464

The timezone currently renders **under the input** as `<small class="p-field-hint">{{ systemTimezone.label }}</small>` in `dot-edit-content-calendar-field.component.html`. This issue **moves it into the picker footer** and removes the under-input line.

That supersedes part of [#37464](https://github.com/dotCMS/core/issues/37464), which left timezone placement out of scope and specified *"when a Date/Time field has both a timezone and a hint, show the hint only"*. Once the timezone lives in the picker, the timezone-vs-hint collision disappears and the field footer carries hint and required-error text only. **Whichever issue lands second should drop the now-moot criterion from the other.**

### Relevant code

| Concern | Location |
|---|---|
| Picker markup, `showClear`, `showButtonBar` | `core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.html` |
| `inline-flex` width, focus-ring styling | `.../components/calendar-field/calendar-field.component.scss` |
| `$isExpireDateField`, `onCalendarChange`, value handling | `.../components/calendar-field/calendar-field.component.ts` |
| Per-type config (`showTime` / `timeOnly` / `icon`) | `.../components/calendar-field/calendar-field.util.ts` — `CALENDAR_OPTIONS_PER_TYPE` |
| Under-input timezone line (to be removed) | `.../dot-edit-content-calendar-field.component.html` |
| Server time helper | `getCurrentServerTime(...)` in `calendar-field.util.ts` |
| Message bundle | `dotCMS/src/main/webapp/WEB-INF/messages/Language.properties` |

All three field types route through the same component — `DATE`, `TIME`, and `DATE_AND_TIME` all map to `dot-edit-content-calendar-field` in `dot-edit-content-field.component.html` — so one fix covers all three.

### Width

- [ ] Date, Time, and Date and time fields fill 100% of the width of the column they are rendered in, matching sibling field types (Text, Select, etc.).
- [ ] The input and the calendar/clock trigger button remain visually joined as one control at full width — the trigger stays flush to the right edge.
- [ ] The focus ring still encloses input **and** trigger button as a single unit (the behavior `calendar-field.component.scss` deliberately implements today).
- [ ] Width is correct in single-column and multi-column form layouts, and at narrow viewport widths.

### Clear on the field

- [ ] A clear (X) control renders inside the field, between the value and the calendar/clock trigger, for **all three** field types.
- [ ] The X is shown **only when the field holds a value**, and is not rendered when the field is empty.
- [ ] Clicking the X empties the field: the display value clears, the form control is set to `null`, and the control is marked touched/dirty so validation reacts.
- [ ] Clearing a **required** field leaves it empty and invalid — the X is not suppressed on required fields, and the required error appears per the rules in #37464.
- [ ] The X is not rendered when the field is disabled or read-only.
- [ ] The expire-date field keeps working: its `Never Expires` placeholder still shows when cleared, and it does not end up with two clear controls.
- [ ] The X is keyboard reachable and has an accessible label (e.g. `Clear`), not an icon-only unlabelled button.

### Picker footer — timezone

- [ ] Opening the picker on a **Date and time** or **Time** field shows the timezone as gray, muted, non-interactive text in the **left** of the footer.
- [ ] Opening the picker on a **Date-only** field shows **no** timezone text — the footer holds the button only.
- [ ] The timezone text is the same value shown today (`systemTimezone.label`), and renders nothing (no empty footer slot, no layout shift) when the timezone is unavailable.
- [ ] The under-input `<small class="p-field-hint">` timezone line is **removed** from `dot-edit-content-calendar-field.component.html`.

### Picker footer — Today / Now button

- [ ] The footer's **right** slot holds a single button, styled as **secondary with an outline**, using the shared PrimeNG button styling (not bespoke CSS).
- [ ] The button reads `Today` on **Date** and **Date and time** fields, and `Now` on **Time-only** fields.
- [ ] On a Date and time field, the button sets **today's date plus the current server time**.
- [ ] On a Date-only field, the button sets today's date.
- [ ] On a Time-only field, `Now` sets the current server time.
- [ ] The value the button sets respects the server timezone — it goes through the existing `getCurrentServerTime` / `convertServerTimeToUtc` path, not the browser's local clock.
- [ ] After clicking the button the field shows the new value, the form control updates, and the control is marked touched/dirty.
- [ ] PrimeNG's default `Clear` button no longer appears in the footer for any of the three field types.
- [ ] Button and timezone labels come from `Language.properties` keys (new keys needed — no `today` / `now` key exists under the `edit.content.form.field.calendar.*` namespace today) and are not hardcoded strings.

### Regression

- [ ] Existing timezone conversion behavior is unchanged: values still store as UTC timestamps, Date-only as UTC midnight, Time-only against a consistent date base.
- [ ] `field.defaultValue` handling (`now` and fixed dates) still works.
- [ ] The picker still does not close on time selection (`hideOnDateTimeSelect` is `false`) and still appends to `body`.
- [ ] The error border on the control in the invalid state still renders (`:host.ng-invalid.ng-touched`).

### Tests

- [ ] Unit tests cover: X hidden when empty, X shown when a value is set, X clears the value and marks the control touched.
- [ ] Unit tests cover the footer per field type — timezone present for Date and time and Time-only, absent for Date-only.
- [ ] Unit tests cover the button label (`Today` vs `Now`) and the value it sets for each of the three field types.
- [ ] A test asserts PrimeNG's `Clear` button is absent from the picker footer.
- [ ] Existing `dot-edit-content-calendar-field.component.spec.ts` assertions on the `calendar-field-timezone` testid are updated for the new placement.

## Priority

Medium

## Additional Context

**Mock 1 — the field.** A `Posting Date` field showing `09/04/2026 09:33`, with a clear X and the calendar trigger to its right. A red line extends to the right of the field toward the column edge, marking the width the field should occupy.

**Mock 2 — the open picker.** Month/year header with prev/next arrows, day grid (the 4th selected), divider, hour/minute spinner (`09 : 33`), then the footer. Red annotations: *"TimeZone: text gray"* on the left of the footer and *"Today button sec with outline"* on the right. The current build's footer — `Today` left, `Clear` right — is what the annotations replace.

**Resolved during refinement:**

| Question | Decision |
|---|---|
| Does the picker footer keep a Clear button? | **No** — removed; clearing is the on-field X |
| When is the on-field X visible? | **Only when the field holds a value** |
| Does the under-input timezone line stay? | **No** — moved into the picker footer |
| Which field types show the timezone in the footer? | **Date and time + Time-only**; not Date-only |
| What does Today set on a Date and time field? | **Today's date + the current time** |
| What does the Time-only footer show? | A **`Now`** button, same secondary-outline styling and position |

**Out of scope:** hint and required-error presentation (covered by #37464); moving the timezone to a secondary line under the *label*; changing the stored value format or the timezone conversion logic; the legacy (non-new) Edit Contentlet screen.

adrianjm-dotCMS:
## PR 1 of 2 — spec only

This PR carries `spec.md` **alone**. No implementation, no tests, no product code.
Per the [Spec-Kit flow](https://github.com/dotCMS/core/blob/main/docs/core/SPEC_KIT_QUICK_START.md),
implementation starts only once another dev **approves** this spec — approval, not merge:
PR 2 branches off this branch and is opened without waiting for this one to land.

Related to #37465

## What the feature is

The three temporal field types in the new Edit Contentlet — **Date**, **Time** and
**Date and time** — are the only fields that look and behave differently from their neighbours.
They sit narrower than their column, they cannot be emptied once a value is set, and the picker
overlay's footer offers a *Clear* action the author does not need there while hiding the one thing
they do need: which timezone the value is interpreted in.

Four independently shippable user stories: **P1** clearing on the field → **P2** timezone in the
picker footer → **P2** a server-accurate *Today* / *Now* → **P3** full-column width.
25 functional requirements, 11 success criteria.

## Two things the code said that the issue did not

Both are worth a reviewer's attention, because the issue describes neither:

1. **The timezone renders today only when the field *also* carries a hint.** The field's footer row
   is drawn only when there is a hint or a required error, so on a Date-and-time field with a
   timezone and no hint the timezone is never shown at all — and where it does render, it displaces
   the hint into the label tooltip. So User Story 2 is not only a relocation: it also closes a
   coverage gap, which is what SC-003 measures ("up from only those fields that happen to carry a
   hint today").
2. **The current *Today* action reads the browser's clock**, not the server's, so it can set the
   wrong calendar day when the server is in another timezone. FR-013 and SC-005 make correcting
   that explicit instead of leaving it as a side effect of rebuilding the footer.

## Two decisions the issue left open

Both were resolved before writing rather than carried as `[NEEDS CLARIFICATION]`, and each is
stated directly by the requirement that owns it:

| Question | Decision | Requirement |
|---|---|---|
| Where does the hint live once the timezone leaves the field footer? | **Back in the field footer**, dropping the current label-tooltip special case | FR-009 |
| Does *Today* / *Now* close the picker? | **No** — it stays open on all three types, the same as selecting a day or a time already does. The picker closes only on a click outside. | FR-014a |

## Overlap with #37464

#37464 (*Standardize field hint and required-error presentation*) is open with no spec and no
branch. This spec leaves hint and required-error presentation to it (see Out of scope), but FR-009
removes the under-input timezone line, which four passages in **#37464's description** are written
around — two acceptance criteria, one test criterion, and a row in its refinement table.

Whichever issue lands second should withdraw those from the other. That is cross-issue bookkeeping,
not a product requirement, so it is tracked here rather than as an FR in the spec — and editing
another issue's description needs the developer's approval before anyone acts on it.

## Explicitly out of scope

Standardizing hint and required-error presentation across all field types (owned by #37464); moving
the timezone under the *label*; any change to how values are stored or converted between the server
timezone and UTC; the legacy Edit Contentlet screen; the picker's header, day grid and time spinner;
date/time fields rendered anywhere outside the new Edit Contentlet form.

## Reviewing this

The spec is the contract PR 2 will be judged against, so the useful questions are about the
**requirements**, not the code: are FR-001…FR-019 testable as written, is the out-of-scope boundary
where you would draw it, and are the two decisions above the calls you would have made.

`checklists/requirements.md` is gitignored per the repo's spec-folder commit policy, so it is not in
this diff.

This PR fixes: #37465


specs/37465-calendar-field-width-clear-picker-footer/spec.md
```diff
@@ -0,0 +1,241 @@
...
+Conversion between the server timezone and UTC already works and is not changed by this feature.
+The timezone change is a relocation of an existing label: it moves from under the input into the
+picker footer. The *Today* shortcut is the one timezone-related defect — it reads the browser's
+clock instead of the server's.
+
+All three field types are rendered by the **same** field component, so one change covers all three.
+
+### Out of scope
+
+- Standardizing hint and required-error presentation across all field types — owned by #37464.
```
nicobytes: Out of scope says hint and required-error presentation is owned by #37464, but FR-009 changes hint placement for these three field types. Worth naming here which issue owns the regression test for that placement, so it doesn't end up covered twice or not at all.

specs/37465-calendar-field-width-clear-picker-footer/spec.md
```diff
@@ -0,0 +1,241 @@
...
+- **FR-006**: The clear control MUST be rendered **only** when the field holds a value, and MUST NOT be rendered when the field is empty or disabled. *(The issue's acceptance criterion also names a read-only state. These fields have no read-only state, so that part of the criterion is not valid.)*
+- **FR-007**: Activating the clear control MUST empty the displayed value, set the underlying value to none, and mark the field as touched and dirty so validation and unsaved-change tracking react. Clearing MUST NOT be suppressed on required fields — a cleared required field is empty and invalid.
+- **FR-007a**: The clear control MUST be reachable by keyboard and MUST carry an accessible name.
+- **FR-007b**: The expire-date field MUST continue to show its *Never Expires* placeholder when cleared, and MUST offer exactly one clear control.
+
+#### Picker footer — timezone
+
+- **FR-008**: Opening the picker on a **Date and time** or **Time** field MUST show the system timezone as muted, non-interactive text at the left of the picker footer. It MUST be the same timezone value shown under the input today.
```
nicobytes, line 187: The timezone line renders today with `role="status"` / `aria-live="polite"`. FR-008 describes it only as muted non-interactive text, and no SC covers screen-reader exposure. Once it lives inside an overlay, should it be announced when the picker opens?

specs/37465-calendar-field-width-clear-picker-footer/spec.md
```diff
@@ -0,0 +1,241 @@
...
+- **FR-008a**: Opening the picker on a **Date** field MUST NOT show timezone text.
+- **FR-009**: The timezone line under the input MUST be removed from the field types that render one today — Date-and-time and Time-only; a Date-only field has none to remove — and the field's hint MUST render in the field footer under the input — the same placement the other field types use — rather than being displaced into the label.
```
nicobytes, line 190: The field footer today renders either the required error or the hint, never both. FR-009 doesn't say which wins once the hint comes back — a required field that is invalid *and* carries a hint has two valid readings, which makes the acceptance test ambiguous.

specs/37465-calendar-field-width-clear-picker-footer/spec.md
```diff
@@ -0,0 +1,241 @@
...
+
+#### Picker footer — Today / Now
+
+- **FR-010**: The picker footer MUST hold exactly one button, at its right, styled as a secondary **outlined** button using the shared button styling rather than bespoke CSS.
+- **FR-011**: That button MUST read **Today** on Date and Date-and-time fields, and **Now** on Time-only fields.
+- **FR-012**: Activating it MUST set: today's date and the current time on a Date-and-time field; today's date on a Date-only field; the current time on a Time-only field. "The current time" is the server's time at the moment of activation, resolved to **second** precision — the same precision the field already applies to a `now` default.
+- **FR-013**: The value it sets MUST be derived from the **server** timezone through the field's existing `getCurrentServerTime` / `convertServerTimeToUtc` path, never from the browser's clock.
```
nicobytes: `getCurrentServerTime` still starts from `new Date()` and only reinterprets that instant in the server's timezone, so "never from the browser's clock" holds for the timezone but not for the instant. If the intent is only to fix the timezone, the wording overstates it; if it also means clock skew, the existing path can't satisfy FR-013 and the spec needs a real server-time source.

specs/37465-calendar-field-width-clear-picker-footer/spec.md
```diff
@@ -0,0 +1,241 @@
...
+- **FR-014**: After activation the field MUST show the new value, the underlying value MUST update, and the field MUST be marked touched and dirty.
+- **FR-014a**: Activating it MUST leave the picker open on all three field types, the same as selecting a day or a time does. The picker closes only when the author clicks outside it.
+- **FR-015**: The stock **Clear** button MUST NOT appear in the picker footer for any of the three field types.
+- **FR-015a**: The button label and any accessible name introduced by this feature MUST come from the localized message bundle under the existing calendar-field namespace, not from hardcoded strings.
+- **FR-015b**: The footer action MUST be reachable by keyboard and MUST carry an accessible name, matching the reachability FR-007a requires of the clear control.
+
+#### Preserved behaviour
+
+- **FR-017**: Stored values MUST be unchanged by this feature: Date-and-time values continue to store as UTC timestamps, Date-only as UTC midnight, and Time-only against a consistent date base.
```
nicobytes: The numbering goes FR-015b → FR-017. Was FR-016 dropped on purpose, or is a requirement missing?

specs/37465-calendar-field-width-clear-picker-footer/spec.md
```diff
@@ -0,0 +1,241 @@
...
+- **FR-018**: Default-value handling (`now` and fixed dates) MUST continue to work, and MUST NOT be re-applied to a field the author has explicitly cleared.
```
nicobytes: FR-018 covers the in-session case, but not what happens after clearing, saving empty, and reopening. The default is applied on open when the value is empty, so as written a `now`-default field would repopulate on reload and the clear wouldn't survive it — which would undercut SC-001's independent test.

specs/37465-calendar-field-width-clear-picker-footer/spec.md
```diff
@@ -0,0 +1,241 @@
...
+
+### Measurable Outcomes
+
+- **SC-001**: A content author can empty a Date, Time or Date-and-time field in a single action, on **100%** of such fields — up from only the expire-date field today.
+- **SC-002**: In a form containing a Date, Time or Date-and-time field beside any other field type in the same column, **0** fields stop short of the column edge, in single-column and multi-column layouts.
+- **SC-003**: The timezone governing a Date-and-time or Time value is visible to the author whenever they are choosing that value, on **100%** of such fields — up from only those fields that happen to carry a hint today.
+- **SC-004**: A field's hint is shown to the author on **100%** of Date, Time and Date-and-time fields that define one, in the same position as every other field type.
+- **SC-005**: With the server and the browser on calendar days that differ, the *Today* / *Now* shortcut sets the **server's** day and time in **100%** of attempts across all three field types.
+- **SC-005a**: The value the shortcut sets is resolved to the **second** in **100%** of attempts — not rounded to the minute, and not drifting from the precision a `now` default already produces.
```
nicobytes: SC-005a reads as applying to all three field types, but FR-017 keeps Date-only stored as UTC midnight, so on a Date-only field there's no second to resolve. Scoping this criterion to Date-and-time and Time-only would make it measurable.

nicobytes (commented):
Reviewed the spec against the current calendar-field code on `main`. The requirements read well and are mostly testable; leaving a few questions where I think the wording would let PR 2 be built two different ways.

rjvelazco (dismissed):

specs/37465-calendar-field-width-clear-picker-footer/spec.md
adrianjm-dotCMS: Good catch — this one was a real defect, not just wording.

`handleChangeValue` receives `null` both when a field never held a value and when the author cleared it and saved, and re-applies `processFieldDefaultValue` in both. So on a field with a `now` or fixed default, clearing does not survive a reload — which contradicts US1's own independent test.

FR-017 (renumbered from FR-018) now requires the two states to be distinguished, US1 gains scenario 7, and SC-001a measures it.

Worth flagging for the implementation: the defect predates this feature. Only the expire-date field could be cleared before, so the path was unreachable — FR-005 made the clear control universal and with it this code path. The parent has `$contentlet` but does not pass it down today, so that is where the fix will land.

specs/37465-calendar-field-width-clear-picker-footer/spec.md
adrianjm-dotCMS: Confirmed — `main` has `role="status" aria-live="polite"` on that label and the rebuilt footer renders a plain `<span>` with neither. It was lost in the rewrite, not decided.

Added FR-008b, stated by outcome rather than mechanism: a screen-reader user must learn the timezone while choosing a value, without having to navigate to the picker footer. I deliberately did not mandate restoring `aria-live` — the label was async-arriving content under the input, and in an overlay the right pattern may be `aria-describedby` on the input instead. SC-003a measures the outcome.

specs/37465-calendar-field-width-clear-picker-footer/spec.md
adrianjm-dotCMS: Right, that was ambiguous. Added FR-009a: when a required field is invalid **and** carries a hint, the footer shows **both** — error first, hint below. The hint says what to enter, which is exactly what the author needs while in error, so the error must not evict it.

That is what the implementation already does; the spec just never said so.

specs/37465-calendar-field-width-clear-picker-footer/spec.md
adrianjm-dotCMS: Agreed, the wording overstated it. FR-013 now separates the two:

- the **zone** the instant is read in — corrected, and what makes `Today` land on the server's calendar day;
- the **instant** itself — still `new Date()`, unchanged by this feature.

Correcting clock skew would need a real server time source, which does not exist today, so that is named as out of scope rather than implied by the wording.

specs/37465-calendar-field-width-clear-picker-footer/spec.md
adrianjm-dotCMS: Correct — scoped SC-005a to Date-and-time and Time, since FR-016 keeps Date-only at UTC midnight and there is no second to resolve there.

SC-005 one line above had the same flaw ("day and time across all three field types"), so I scoped it the same way: the day on Date-only, the time on Time-only, both on Date-and-time.

specs/37465-calendar-field-width-clear-picker-footer/spec.md
adrianjm-dotCMS: It was dropped on purpose — FR-016 tracked a cross-issue bookkeeping obligation that was later moved out of the requirements, since it is not a product requirement.

The gap was just untidy, so FR-017..020 are renumbered to FR-016..019 and the one cross-reference updated. Numbering is now continuous FR-001..FR-019.

specs/37465-calendar-field-width-clear-picker-footer/spec.md
adrianjm-dotCMS: Fair point — the boundary was fuzzy. Out of scope now says it explicitly: **this** issue owns the hint's placement on the three calendar field types, because FR-009 moves it, so the regression test for that placement belongs here. #37464 keeps the cross-field-type consolidation and the styling.

nicobytes (approved):

rjvelazco (approved):

Issue #37522, opened 2026-09-11
Search queries are lowercased end to end, so case-sensitive matching is impossible and quoted phrases are split


fabrizzio-dotCMS:
## Description

Every query that goes through the search view tools and the search API is folded to lower case
before it runs, **values included**, so a case-sensitive match is impossible through them. Quoted
phrases are not protected either.

The folding is done by `StringUtils.lowercaseStringExceptMatchingTokens(query, LUCENE_RESERVED_KEYWORDS_REGEX)`:

- `StringUtils.java:335` — the helper splits the query on **spaces** with a `StringTokenizer` and
  lower-cases every token that does not match the regex.
- `ESContentFactoryImpl.java:252` — the regex is `"OR|AND|NOT|TO"`, so only those four tokens survive.

Call sites (both engines, so the behaviour is the same in every migration phase):

| Path | Where |
|---|---|
| Legacy ES search | `enterprise/.../priv/ESSearchAPIImpl.java:64` (`esSearch`) and `:116` (`esSearchRaw`) |
| OpenSearch | `OSSearchAPIImpl.java:92` (`search`) and `:154` (`searchRaw`) |

### What it does to a query

| Written | Executed |
|---|---|
| `+title:Hello` | `+title:hello` |
| `+title:"Hello World"` | split on the space, both halves folded — the quotes protect nothing |
| `+contentType:News AND +live:true` | `+contenttype:news AND +live:true` (`AND` survives) |

Two separate consequences:

1. **No case-sensitive matching.** Against a `keyword`-mapped field holding `News`, a query for
   `News` is executed as `news` and does not match. This hits identifiers, product codes, tags with
   their own capitalisation, and any `_dotraw` field an exact match was expected on.
2. **Quoted phrases are tokenised by the helper.** Splitting on spaces means a phrase query is
   processed as separate tokens before it ever reaches the engine, which is not what the author
   wrote.

The field-name half of this is deliberate and useful — it is what lets `contentType` resolve to the
physical index field `contenttype`. The value half looks like collateral damage rather than intent.

### Not a migration regression

This dates to `c4e2d072cc` (2023-04-19) on the Elasticsearch side. The OpenSearch implementation
added the same call deliberately for parity (`80b6a1bcc6`, #35609) — its own comments say
*"Reuses the existing lowercasing helper for parity"* and *"Symmetric with the ES raw path"*. So
nothing about it changes when an installation migrates; it is raised here because the ES→OS review
[...]

PR #37544, opened 2026-09-14
docs(search-api): fix the VTL examples and document JSON output

### LINKED ISSUE #37520: ES→OS: endpoint to audit and rewrite VTL that calls the deprecated $estool.esSearch / $estool.esRaw
## Description

Customers migrating from Elasticsearch to OpenSearch 3 have to find and rewrite every Velocity
template that calls the two deprecated search methods, `$estool.esSearch(...)` and
`$estool.esRaw(...)`. Today this is entirely manual, and it is the single hardest part of the
migration for a customer to do alone:

- The code is spread across **four** places — templates and containers stored in the database, file
  templates/containers in the assets tree, widget code and custom-field Velocity inside content type
  definitions, and Velocity script steps inside workflow actions. Nothing but the files can be
  grepped.
- The failure is silent. These templates keep working through Phase 1 (reads still come from
  Elasticsearch) and break in Phase 2, and they break by rendering *less*, not by erroring.

This task adds one role-gated endpoint that both **reports** what needs changing and, when explicitly
instructed, **applies** the mechanical part of the change with a manifest that can be reverted.

### Endpoint modes

| Mode | Behaviour |
|---|---|
| `GET /api/v1/index/migration/vtl-audit` | Read-only report: object type, id, name, line, rule, confidence. Touches nothing. |
| `POST /api/v1/index/migration/vtl-audit/_apply` with `dryRun: true` | Returns the exact diff it would write. Writes nothing. |
| `POST /api/v1/index/migration/vtl-audit/_apply` | Writes the **working version only** of each object — never publishes — and returns a `runId` plus the manifest (object, rule applied, previous version). |
| `POST /api/v1/index/migration/vtl-audit/_revert` with a `runId` | Restores the previous version of every object in that manifest. |

### Rules

Auto-rewritable (high confidence):

| From | To |
|---|---|
| `$estool.esRaw(` | `$estool.raw(` |
| `$r.took` (on a variable bound to a raw response) | `$r.tookInMillis` |
| `$r.hits.getAt(n)` | `$r.hits.hits.get(n)` |
| `$estool.esSearch(` **and** the `$item.map.field` uses of its loop variable | `$estool.search(` and `$item.field` |

Reported, never rewritten:

| Found | Why it needs a human |
|---|---|
| `$r.toString()` used as JSON | The neutral response does not serialize to engine JSON; the JSON has to be rebuilt field by field. Highest-risk finding: it fails silently. |
| `$results.hits.maxScore` | No equivalent on the neutral response; the score has to be read per hit. |
| `esSearch` whose result variable escapes the file (`#parse`, `#dotParse`, a macro) | The rewrite of `.map.` uses cannot be attributed safely. |
| Velocity steps in workflow actions | Reported with the action id; rewriting is allowed only with an explicit target list. |

### Design constraints

1. **Only rewrite what is provably local.** If the result variable of an `esSearch` call crosses a
   `#parse` / `#dotParse` / macro boundary, the finding is reported, not rewritten. A wrong rewrite
   here fails the same silent way the migration itself does.
2. **Never publish.** Writing the working version keeps the live site untouched and uses the
   versioning that templates, containers, content types and file assets already have — restoring the
   previous version *is* the rollback. Publishing stays a human decision after reading the diff.
3. **The manifest is the only backup we own.** Workflow actionlets do not version, so for those the
   previous code is stored whole in the manifest.
4. **Idempotent.** A second run finds nothing to do; `_revert` works as long as the `runId` manifest
   exists.
5. **Gated and off by default.** Same gate as the readiness endpoint — CMS Administrator *plus* the
   migration role — since this reads and rewrites customer code.

## Acceptance Criteria

- [ ] `GET .../vtl-audit` returns findings from all four sources: database templates and containers,
      file templates/containers, content type field values (widget code and custom fields), and
      Velocity workflow actionlets.
- [ ] Every finding carries object type, id, human-readable name, line, rule, and whether it is
      auto-rewritable or manual.
- [ ] `_apply` with `dryRun: true` returns a diff and provably writes nothing.
- [ ] `_apply` writes only working versions; nothing is published, and the live site renders
      identically immediately after a run.
- [ ] `_apply` returns a `runId` whose manifest names every object touched and its previous version.
- [ ] `_revert` with that `runId` restores every object to its previous state, including workflow
      actionlets, which have no versions of their own.
- [ ] Re-running `_apply` on an already-converted installation reports and changes nothing.
- [ ] A template whose `esSearch` result escapes through `#parse` or a macro is reported and left
      untouched.
- [ ] `toString()`-as-JSON and `maxScore` findings are never rewritten.
- [ ] Both endpoints answer 403 without CMS Administrator plus the migration role.
- [ ] Integration tests cover a fixture with one case per rule, plus an already-migrated fixture that
      must produce zero findings.

## Additional Context

The rules come from the customer-facing migration guide's template section, which was verified
against the code:

- `ESContentTool.esSearch` returns `ESSearchResults` holding raw `Contentlet`s; `ESContentTool.search`
  returns `ContentSearchResults<ContentMap>`. `ContentMap` has no `getMap()`, which is why
  `$item.map.field` stops resolving after the rename — silently.
- `ContentSearchResponse` exposes `getTookInMillis()`, and `SearchHits` exposes `getHits()` /
  `getTotalHits()`; the old `getAt(n)` and `getMaxScore()` have no counterpart.
- On Elasticsearch the neutral `search()` delegates to the same `esSearch()` underneath
  (`ESSearchAPIImpl.search`), so query handling is unchanged by the rewrite — only the result type is.

Both deprecated methods are marked `@Deprecated(forRemoval = true)` for removal in v26.08.04.

Related: the deprecation inventory in `docs/backend/ES_OS_DEPRECATIONS.md` and the plugin audit skill
requested in #37466 — this issue is the Velocity half, and the two should share one rule source
rather than each keeping its own list.

fabrizzio-dotCMS:
## What

Fixes the Velocity section of `docs/backend/SEARCH_API_MIGRATION.md`, whose examples could
not be followed as written, and adds a section answering the question the guide raises but
never resolves: if `.toString()` no longer returns JSON, what does a template use instead?

## Why

Every snippet in section 3 fails:

| In the doc | Problem | Correct |
|---|---|---|
| `$ESContent.raw($query)` | The viewtool is registered under the key `estool` (`toolbox.xml`). `$ESContent` resolves to nothing | `$estool.raw($query)` |
| `$raw.hits().hits()` | Java call syntax. Velocity resolves properties through bean getters | `$raw.hits.hits` |
| `$hit.id()` | Same | `$hit.id` |
| `hits().totalHits().value()` | Listed under a Velocity heading, but it is Java — and wrong Java | `$raw.hits.totalHits.value` |

The Java example in section 1 was also wrong: `SearchHits` declares its record components
as `getHits` / `getTotalHits` — deliberately, so Velocity can resolve them as properties —
which means there is no `hits()` or `totalHits()` method on it. `raw.hits().hits()` does not
compile; it is `raw.hits().getHits()`.

And `aggregations()` was recommended with no caveat. Its own javadoc says it returns
"first-level terms aggregations… only aggregations that have buckets are included" — it
silently drops nested aggregations and `top_hits`. The full tree is `getAggregations()`,
reached from VTL as `$raw.aggregations`.

## Changes

- Section 3 renamed to `$estool`, with every snippet rewritten in real Velocity.
- The accessor table now has two columns, VTL and Java, instead of mixing the dialects under
  a Velocity heading. Both flavours of `aggregations` are listed, with the flattened one
  marked lossy.
- A note on why `SearchHits` looks the way it does, so the naming stops surprising people
  in both languages.
- New **JSON output** section: `$json.generate(...)` still produces JSON in the neutral
  shape — a path kept working on purpose (`ContentSearchResponse.getAggregations()` is
  deliberately not `@JsonIgnore` so the reflective JSON builder still sees it, issue #36435)
  — while Elasticsearch's wire format is gone by design. Only consumers that parse that
  specific envelope need restructuring. Includes a reduced-payload example and a
  three-case summary table.
- Documents that the query is lowercased before execution
  (`StringUtils.lowercaseStringExceptMatchingTokens`), so an aggregation declared `"tagAgg"`
  comes back keyed `tagagg` — a `#foreach` over the wrong key renders nothing rather than
  failing.

## Verification

Every accessor in the changed text was checked against the source: `ESContentTool`,
`ContentSearchResponse`, `SearchHits`, `TotalHits`, `SearchHit`, `AggregationBucket`,
`JSONTool`, and the viewtool keys in `toolbox.xml`.

Docs only — no code changes.

Related: #37520 (the audit/rewrite tool, which uses these rules as its spec).

This PR fixes: #37520

This PR fixes: #37520


fabrizzio-dotCMS:
Pushed a second commit (`3931917c52`) extending this to `OPENSEARCH_MIGRATION_RUNBOOK.md`, which carried the same class of errors plus two of its own:

- **Neutral response accessors** — same fix as the migration guide: `hits().getHits()` / `hits().getTotalHits().value()`, and `getAggregations()` instead of the flattened `aggregations()`.
- **When `esSearch` / `esRaw` break** — the runbook said Phase 2, in four places. Both delegate to `APILocator.getEsSearchAPI()` in every phase and never reach the phase router, so in Phase 2 they keep answering from Elasticsearch while the rest of the site reads OpenSearch: a silent divergence, not a crash. They fail at Phase 3. The Timing section now says that neither a clean Phase 1 nor a clean Phase 2 tells you anything about plugin readiness.
- **`OS_INDEX_REPLICAS`** — it is declared with a fallback to `ES_INDEX_REPLICAS` (`OSIndexProperty`), so "OpenSearch does not inherit dotCMS's implicit default" was wrong. The advice to set it stands, with the real reason.
- **A contradiction on Site Search** — Stage 3.9 warned about incremental crawls leaving a dynamically-mapped copy while R12 documented the gate that prevents it. Reconciled.

One known issue deliberately left alone: the runbook describes the Phase 2 read fallback as covering every read, which is not true in `main` today — `ESContentFactoryImpl` picks its provider directly and never goes through `PhaseRouter`. That is what #37413 / #37500 fixes, so the wording is correct once that merges and I would rather not churn it twice.

fabrizzio-dotCMS:
Both review notes applied in `553ae3c7a6`:

- **Header vs. audience** — it now names both audiences and says which sections belong to each: Java callers get 1, 2, 4 and 5; template authors get section 3, which is self-contained and written entirely in Velocity. The closing line also distinguishes the two failure modes, since they are not alike: Java fails to compile, templates fail silently.
- **Redundant cast** — dropped; `esSearch(...)` already returns `ESSearchResults`.

On the lowercasing note: the fact is right and worth recording. `esSearch` has normalised the query since `c4e2d072cc` (2023-04-19), but `esSearchRaw` only gained it in `7f8eee8586` (2026-06-16, #36170 — "normalize query in raw() search path"). So on a build older than June the raw path did not lowercase. Agreed the present-tense wording is accurate for anything current, so leaving the text as is.

docs/backend/OPENSEARCH_MIGRATION_RUNBOOK.md
jdcmsd: Step 0.2 says templates "break in Phase 2." Claude is suggesting this may have been missed when R7's timing section was rewritten:

> Since `$estool.esSearch` goes through the same getEsSearchAPI() path, shouldn't it be diverge at Phase 2, fail at Phase 3 like the plugin table now says?

docs/backend/SEARCH_API_MIGRATION.md
```diff
@@ -122,46 +139,114 @@ the elements are still `ContentMap` objects with the same properties.
...
+
+> **Use `$raw.aggregations`, not the flattened `aggregations()` view.** The flattened map
+> keeps only first-level terms aggregations and silently discards nested ones and
+> `top_hits`. From VTL, `$raw.aggregations` resolves to `getAggregations()` and returns the
+> whole tree; each value exposes `.buckets`, and each bucket exposes `.key`,
+> `.keyAsString`, `.keyAsNumber`, `.docCount` and its own nested `.aggregations`.
+
+> **Aggregation names come back lowercased.** Both `search` and `raw` fold the whole query
+> to lower case before running it (`StringUtils.lowercaseStringExceptMatchingTokens`, the
+> same call the deprecated methods made), so an aggregation declared as `"tagAgg"` is keyed
```
jdcmsd, line 196: I had mentioned this flag earlier; what do you think? Do we need any sort of mention of the periodization re: the raw-method update in 26.06.21-01 that brought the lowercasing behavior into parity? or is it not worth bringing up?

jdcmsd (changes requested):
Two comments attached.

docs/backend/SEARCH_API_MIGRATION.md
fabrizzio-dotCMS: Good catch on the version, but I'd leave it out. The June change brought esSearchRaw in line with esSearch — and since the neutral raw() delegates to esSearchRaw on the ES path, old and new behave identically to each other at any given build: before it, neither lowercased; after it, both do. So for someone migrating from esRaw to raw there's no behaviour change in any version, which is exactly what that note is there to say. Dating it would imply the migration is implicated when it isn't. The real audience for the version is anyone upgrading across 26.06.21-01 — their raw queries start lowercasing — but that's a release-note matter rather than a migration-guide one.
https://github.com/dotCMS/core/issues/37522

docs/backend/OPENSEARCH_MIGRATION_RUNBOOK.md
fabrizzio-dotCMS: Good catch — and it was broader than Step 0.2. Fixed in 6a23ae1372.

Four places still said customisations break at Phase 2, so the runbook contradicted itself after the last commit: Step 0.2 (templates), Step 0.3 (plugins), the Phase 2 description in Start here, and the Stage 4 risk heading.

Following the reasoning through changed what Stage 4 is about. A legacy template does not merely diverge at Phase 2 — while the two indices agree it renders exactly as before, so walking the site in Stage 4 proves nothing about it. It fails in Stage 5.

So Stage 4's real exposure is the opposite set: the supported path — content pulls, the admin search, URL maps, REST and GraphQL — which does switch engine there. Any difference between the indices surfaces at that moment: results in a different order where the query gave no explicit sort, or short results where the copy is incomplete. Both headings now say that, with the legacy methods called out separately as the deferred, symptomless risk they are.

Also moved the template/plugin deadline from "before Stage 4" to "before Stage 5", with the reason attached: fix them while Elasticsearch is still live and old and new output can be compared side by side.

jdcmsd (approved):
all comments addressed

PR #37546, opened 2026-09-14
fix(uve): stop the bounds re-anchor degrading the selected payload

### LINKED ISSUE #37499: UVE: "Edit VTL" menu opens empty after the selected contentlet is re-anchored

rjvelazco:
Fixes #37499. Spec approved and merged in #37519, then **amended on this branch** — see *Needs re-approval* below.

## The defect

The `</>` Edit VTL button is gated on the **hovered** contentlet, but its menu was filled from the **selected** one. The selected contentlet's payload loses `vtlFiles` on every `SET_BOUNDS` re-anchor, so after any layout shift the button appeared and the menu opened empty. Customers read it as *"Edit VTL is not responding"*; clicking Quick Edit first happened to repopulate it.

## The cause, and why this PR is bigger than the symptom

`SET_BOUNDS` is a **positional** event, but `applyBoundsForSelection()` rebuilt `editorSelected.payload` wholesale from the bounds snapshot. That snapshot carries only `{identifier, title, inode, contentType, canEdit}`, so every re-anchor silently dropped `vtlFiles`, `baseType`, `onNumberOfPages` and `dotStyleProperties`.

`core-web` already carried two independent workarounds for treating that payload as untrustworthy — the style editor's `extractFromRollback` and `handleOpenFullEditor`'s `$contentletEditData`. The VTL menu was simply the one consumer with no workaround. Fixing only the menu would have left the landmine.

**The payload merges now**, it does not freeze. Freezing was evaluated and rejected: `insertContentletInContainer({...payload})` and `deleteContentletFromContainer(payload)` read `payload.pageContainers` directly, so a frozen copy would make add and delete write a stale container tree.

| Data | Source | On re-anchor |
|---|---|---|
| `pageContainers`, `contentletsId`, `language_id`, `pageId`, `personaTag` | `$pageData()` | **refreshed**, as before |
| `vtlFiles`, `contentlet.baseType`, `onNumberOfPages`, `dotStyleProperties` | DOM at selection time | **preserved** |

## Also in this PR

- `vtlMenuItems()` reads the hovered contentlet, matching the `hasVtlFiles()` gate.
- It returns `[]` rather than `undefined`, honouring its declared `MenuItem[]` type.
- Choosing a file promotes the hovered contentlet to selected — on the **file choice**, not the button click, so a menu opened and dismissed leaves the selection alone. This reverses an earlier decision; @dario-daza raised it on #37519 and his instinct was right.
- Dead `selectedHasVtlFiles` removed.

**Deliberately unchanged**: the SDK bounds payload (ADR-0019 treats `postMessage` as a versioned surface), the pencil and `×` buttons, and the two existing payload workarounds — neither is caused by the re-anchor.

## Verification

- Full project suite: **1506 passed, 5 skipped, 66 files**
- Lint clean; typecheck 0 errors (`--skip-nx-cache`)
- 10 new cases, both Red gates recorded before implementation per Constitution Principle V
- One pre-existing case updated with the contract: `should return undefined when no vtl files` → expects `[]`

## Needs re-approval

The spec was amended after sign-off (AC-011 supersedes the withdrawn AC-009; AC-012 reverses the no-promote decision). Per the two-PR flow this needs a fresh look — @dario-daza, the third commit is the spec diff.

## Open before merge

- Manual verification of AC-002, AC-006, AC-007 and the AC-011 add/delete regression (deliberately not automated — the debounced `AUTO_BOUNDS` trailing edge inside the iframe makes an E2E flaky)
- Two test-strength items from `/speckit-converge` (T049, T050) — no behavior gaps
- `/speckit-docs-converge` is blocked on the task list and has not assessed two known doc drifts: `core-web/CLAUDE.md` still documents the Jest-era `--testPathPatterns=` flag, which does not work under this project's Vitest runner; and the UVEStore `CLAUDE.md` does not record the two-payload-kinds rule this PR establishes

## Video

https://github.com/user-attachments/assets/57d4e72c-7afa-4035-a70e-036c275f313e

This PR fixes: #37499


rjvelazco:
Added the store-level guard you suggested — `e89f55bb01`.

You were right that the existing case asserted the payload *field* and never ran the real consumer over it. The new test feeds the merged payload to `deleteContentletFromContainer` and asserts it operates on the refreshed container tree.

Worth noting what made it more than a shape check: both save-path utilities **mutate** `action.pageContainers` via `.push()` when the container isn't found, so a merge that preserved the click-time tree wouldn't just read stale data — it would write it back.

I mutation-tested the guard rather than trusting a green first run. Reversing the spread order so the stale payload wins:

```
expected [ 'gone-since-selection' ] to deeply equal [ 'kept-a', 'kept-b' ]
```

So it fails for the right reason, and the failure message names the actual consequence.

This converts what was manual-only verification into an automated one. The remaining manual items (AC-002 / AC-006 / AC-007) genuinely need the iframe and stay manual.

The two doc drifts are still open — tracked, not forgotten.

dario-daza (dismissed):

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-contentlet-tools/dot-uve-contentlet-tools.component.ts
```diff
@@ -325,16 +318,34 @@ export class DotUveContentletToolsComponent {
...
-        return vtlFiles?.map((file) => ({
-            label: file?.name,
-            command: () => this.editVTL.emit(file)
-        }));
+        const { vtlFiles } = this.contentContext() ?? {};
+        return (
+            vtlFiles?.map((file) => ({
+                label: file?.name,
+                command: () => {
+                    this.promoteHoverToSelected();
```
zJaaal, line 303: **Medium.** A behavior decision rather than a defect: should choosing a VTL file retarget an *open* side panel?

`promoteHoverToSelected()` here does more than move the selection border, because `editorEditPanelOpen` is independent state that survives a selection change (`edit-ema-editor.component.ts:397`). Both panel tabs bind to `editorSelected`:

* `$contentletEditData()` (`edit-ema-editor.component.ts:229`) reads `this.uveStore.editorSelected()?.payload` and feeds `<dot-uve-contentlet-quick-edit [data]="$contentletEditData()">` (template line 251).
* The style form's identifier computed (`dot-uve-style-editor-form.component.ts:137`) reads `editorSelected()?.payload?.contentlet?.identifier`, and the effect it drives calls `#buildForm`, which does `this.#form.set(null)` at line 203 before rebuilding.

So: a user opens quick edit on contentlet B, the panel stays open, then hovers contentlet A and picks a file from A's `</>` menu. The panel silently swaps to A and the form is torn down and rebuilt, while the VTL editor opens on top of it.

Worth noting the pencil already made the opposite call for the same reason. `handleEditWithCopyDecision` (`edit-ema-editor.component.ts:1541`) states that it reads the contentlet from the event payload, not from `editorSelected`, because the pencil is intentionally stateless with respect to editor selection. Opening a VTL file is arguably the same kind of dialog action, and `handleEditVTL` (line 1608) never reads `editorSelected` either.

This may well be the intended trade, since the PR body argues for it explicitly and @dario-daza's instinct on #37519 pointed this way, so I am raising it as a decision, not a fix request. Promoting the *border* is clearly desirable; promoting the *open panel* is the part that looks like a side effect rather than a choice. If you keep it as is, a case covering "panel open on B, VTL chosen on A" would pin the intent: none of the ten new cases exercise the open-panel state.

_Posted by @zJaaal, written by Claude Code._

core-web/libs/portlets/edit-ema/portlet/src/lib/store/features/editor/withSelectionAnchor.ts
```diff
@@ -128,14 +128,48 @@ export function withSelectionAnchor() {
...
+                                //     (getDotCMSContentletsBound), so rebuilding from it
+                                //     DROPS the rest. Preserving them is what keeps the
+                                //     selection usable across a scroll or resize.
+                                //
+                                // Spread order: previous first so nothing is lost, the
+                                // fresh payload second so live values win. contentlet is
+                                // nest-merged for the same reason — a wholesale replace
+                                // is precisely the bug.
+                                payload: {
+                                    ...previousPayload,
```
zJaaal: **Low.** The blanket spread preserves more than the comment above describes.

The comment enumerates the DOM-sourced data worth keeping (`vtlFiles`, `contentlet.baseType`, `onNumberOfPages`, `dotStyleProperties`), but `...previousPayload` also carries forward the transient `PositionPayload` / `ActionPayload` fields that every prior re-anchor cleared: `position`, `newContentlet`, `newContentletId` (`shared/models.ts:35-49`).

`position` is the one that actually accumulates. `promoteHoverToSelected()` writes `payload: this.contentContext()` (`dot-uve-contentlet-tools.component.ts:592`), and `contentContext()` always stamps `position: this.buttonPosition()` (line 195), which defaults to `'after'` (line 133). `getPageSavePayload()` never sets `position`: it spreads the bounds `positionPayload`, which carries only `{container, contentlet}`. So once set, `position` now survives every re-anchor for the life of the selection.

That matters because `insertContentletInContainer` branches on exactly that field:

```ts
if (action.position) {
    return insertPositionedContentletInContainer(action);
}
```
(`utils/index.ts:216`)

I traced every insert call site (`edit-ema-editor.component.ts:998, 1042, 1106, 1148, 1473`) and each one gets its payload from `menuItems()` / `addContent`, which re-stamps `position` from `buttonPosition()`, so this is latent today rather than a live bug. But an explicit allowlist of the keys the comment names would preserve exactly what is intended, instead of everything that happens to be on the object.

_Posted by @zJaaal, written by Claude Code._

core-web/libs/portlets/edit-ema/portlet/src/lib/store/features/editor/withSelectionAnchor.ts
```diff
@@ -128,14 +128,48 @@ export function withSelectionAnchor() {
...
+                                    ...actionPayload,
+                                    contentlet: {
+                                        ...previousPayload?.contentlet,
+                                        ...actionPayload.contentlet
+                                    }
+                                }
```
zJaaal, line 180: **Low.** One instance of the same landmine is still out there, on the fork path.

The PR's framing is right ("fixing only the menu would have left the landmine"), and this merge fixes the `SET_BOUNDS` instance. But `applyBoundsForSelection` is not the only place that rebuilds `editorSelected.payload` from a partial source. `dot-uve-copy-decision.component.ts:116` does:

```ts
const activeContentlet = this.#uveStore.getPageSavePayload({
    container,
    contentlet: newContentletPayload
});
// ...
this.#uveStore.setSelectedPayload(activeContentlet);   // line 125
```

That is a wholesale replace built from `{container, contentlet}` only, so the forked selection loses `vtlFiles`, `dotStyleProperties` and `baseType`. Its own comment at lines 121-124 then leans on this very code path to recover, saying `pageReload` will re-emit `SET_BOUNDS` via auto-bounds and re-anchor against the new inode. A merge can only preserve what is already there, so after a fork the data is gone until the user re-clicks the contentlet.

No user-visible symptom today, which is why I am filing this Low rather than asking for it in this PR: the VTL menu now reads hover, and `#buildForm` reads `dotStyleProperties` from the page asset via `extractFromRollback` (`dot-uve-style-editor-form.component.ts:197`) rather than from the payload. It is simply the remaining occurrence of the pattern this PR set out to remove, and worth a follow-up issue so the next consumer without a workaround does not rediscover it.

_Posted by @zJaaal, written by Claude Code._

core-web/libs/portlets/edit-ema/portlet/src/lib/edit-ema-editor/components/dot-uve-contentlet-tools/dot-uve-contentlet-tools.component.ts
```diff
@@ -227,13 +227,6 @@ export class DotUveContentletToolsComponent {
         return this.selectedContentContext()?.contentlet?.identifier === 'TEMP_EMPTY_CONTENTLET';
```
zJaaal: **Low.** Three sibling dead members survived the sweep.

Deleting `selectedHasVtlFiles` was correct: I grepped `libs/` and it had no reference outside its own declaration. The same is true of three neighbours in this class, none of which is referenced by the template, by another class, or by a spec:

* `isSelectedContainerEmpty` (this computed, line 226). The template uses `isContainerEmpty()` through the `hoverIsEmpty` `@let` at line 9.
* `isHoveredDifferentFromSelected` (line 160), and with it `isSameContentlet` (line 141), whose only production caller is that computed. Its remaining references are the five cases in `describe('isSameContentlet')` in the spec.
* `dragPayload` (line 501). The template binds `hoverDragPayload()` (line 527) at template line 48, via `[attr.data-item]="hoverDragItem | json"`.

Cheap to fold into the same commit while the file is already open, and it removes the last consumers of `selectedContentContext()` outside `menuItems()`.

_Posted by @zJaaal, written by Claude Code._

rjvelazco:
All four addressed — `bc191ebdc9` and `504be1bd8d`. I verified each against the code rather than taking them as read; here's where each landed.

**C1 (panel retargeting) — kept, and now pinned by a test.**

You were right that I hadn't considered the open-panel case; AC-012 only said "no side panel *opens*". But the behaviour isn't new. The `SET_SELECTED_CONTENTLET` handler already documents that one write *"drives both the floating overlay and the side panel's data binding"*, and nothing closes the panel on a selection change — so plain-clicking a contentlet already retargets an open panel and rebuilds the form. Promotion makes `</>` consistent with a click rather than adding a hazard.

The end state is also the coherent one. Without promotion you get the VTL dialog on A while the border and panel stay on B.

Added the case you asked for — "panel open on B, VTL chosen on A" — so the intent is recorded rather than incidental.

**C2 (blanket spread) — fixed.** Confirmed the mechanism: `contentContext()` always stamps `position`, `getPageSavePayload()` never sets it, so it was being pinned to the selection for its lifetime. Also confirmed latent, not live — I checked all six `insertContentletInContainer` call sites and each re-stamps via `menuItems()`/`selectedContentContext()`. Now an explicit allowlist, so the code matches its comment. Test-first: it failed with `expected 'after' to be undefined` before the change.

**C3 (fork path) — filed as #37551.** Agreed this is a follow-up, not this PR.

**C4 (dead members) — all four removed**, along with their specs. One correction: the comment says none is referenced by a spec, but `dragPayload` had a `describe` block with two cases. Removed those too — a suite over dead code still tests dead code. `menuItems()` is now the only remaining reader of `selectedContentContext()`.

Suite: 1501 passed / 5 skipped, 66 files. Lint and typecheck clean.

zJaaal (dismissed):

adrianjm-dotCMS (approved):

PR #37549, opened 2026-09-14
fix(velocity): harden SecureIntrospector restricted-class coverage

nollymar:
Fixes dotCMS/private-issues#704

## What

Extends the Velocity template sandbox's `SecureIntrospector` so method execution is denied on the file, IO, and network-resource type families. The introspector previously matched restricted classes and packages by exact string, so types outside the enumerated `java.lang.*` / reflection set — including `java.io.File`, `java.nio.file.Path`, IO streams, and `java.net.URL`/`URI` — were not covered.

## Changes

- **`SecureIntrospectorImpl`**: adds a hierarchy-aware restricted-type check alongside the existing `ClassLoader`/`Thread` block. Matching by type hierarchy means concrete platform implementations (for example the JDK's internal `Path` type) are covered as well, which an exact-string list cannot reach. Covered families: `File`, `Path`, `RandomAccessFile`, `InputStream`, `OutputStream`, `Reader`, `Writer`, `Channel`, `FileSystem`, `FileSystemProvider`, `URL`, `URI`, plus the `Files` and `Paths` utility holders by identity.
- **`velocity.properties`**: documents the same network-resource and file-system utility types in the restricted-class list, alongside the existing entries. The code-level check is the enforcement floor.
- The existing `Number`/`Boolean`/`String` and `Class.getName()` fast-paths and all previously restricted classes are preserved unchanged.
- Progressive enhancement: `@Override` added to `getMethod` and `checkObjectExecutePermission`.

## Testing

- **Unit** (`SecureIntrospectorImplTest`): asserts the file/IO/network-resource families (including a concrete platform `Path` implementation) are denied, the previously restricted reflection/system families stay denied, and the common value/collection types stay allowed. Also verifies the config list enumerates the exact-matchable resource types.
- **Integration** (`SecureIntrospectorRenderTest`, registered in `MainSuite1b`): renders templates through the real Velocity engine and confirms the guard is applied at render time, while ordinary value methods still render.

Both suites pass locally (`Tests run: 5` unit, `Tests run: 3` integration).

This PR fixes: #37508


wezell:
Are there any performance implications of this change?  Reflection is already Velocity's weakpoint.  It would be interesting to do a before and after perf test.

dotCMS-Machine-User (dismissed):

nollymar:
Good question @wezell  — net impact is negligible, and I measured it.

**Where the check runs.** Method resolution goes through Velocity's per-context introspection cache (`ClassUtils.getMethod` → `context.icacheGet` / `icachePut`). `checkObjectExecutePermission` only runs on a **cache miss** — roughly once per distinct `(class, method)` per render context, not per invocation. A loop calling `$item.getX()` over N rows resolves once and is served from cache after that, so this isn't on the hot reflection path.

**Cost of the check itself.** The added coverage is a bounded set of `isAssignableFrom` calls plus two identity checks. The common value types (`Number`/`Boolean`/`String`, and `Class.getName()`) short-circuit on the existing fast-paths *before* reaching the new block, so they pay nothing extra.

**Before/after microbenchmark** of `checkObjectExecutePermission` (JDK 25, warmed up, best-of-N), measuring per-resolution cost:

- Mixed class workload (value types, ordinary collections, denied file/IO types): **21.2 ns → 17.7 ns** — slightly *faster*, because denied file/IO types now return early instead of scanning the full string deny-list.
- Worst case, an ordinary *allowed* type (e.g. `ArrayList`) that runs the whole hierarchy scan and then the existing string loops: **27.6 ns → 34.2 ns**, i.e. **~+6.6 ns per resolution**.

So the worst case adds single-digit nanoseconds per *distinct* method resolution, paid once per render and cached after. A render resolving ~100 distinct signatures pays well under 1 µs total, one time — below the noise of parsing / I/O / DB.

Happy to run a full render-level benchmark end-to-end if you'd like to see it, but at the introspection level the delta is in the nanoseconds and partly negative.

dotCMS/src/main/resources/org/apache/velocity/runtime/defaults/velocity.properties
```diff
@@ -278,3 +278,12 @@ introspector.restrict.classes = javax.management.MBeanServer
...
 introspector.restrict.classes = javax.script.ScriptEngineManager
+
+# File / network-resource types. These document the restricted-class intent alongside the
+# entries above; note that this exact-string list cannot match concrete platform
+# implementations (e.g. the JDK's internal Path type), so the hierarchy-aware check in
+# SecureIntrospectorImpl is the actual enforcement floor for these families.
+introspector.restrict.classes = java.net.URL
+introspector.restrict.classes = java.net.URI
+introspector.restrict.classes = java.nio.file.Files
+introspector.restrict.classes = java.nio.file.Paths
```
wezell, line 289: why not go deeper here and include other packages?  

`java.nio`
`java.net`

Also, can we configure (env var) packages to include or exclude?  that would be handy and would allow us to push the fix deeper to a more white listed stance.  We update these deny lists all the time as new holes are discovered.

wezell (changes requested):

dotCMS/src/main/resources/org/apache/velocity/runtime/defaults/velocity.properties
sfreudenthaler: My opinion.  Switch it to an allow list and add the default deny on the fall through.

wezell: I think that could break some implementations but that is where I was suggesting to move, as long as it is configurable and we log the deny message

nollymar: Good call — made it configurable. Pushed a follow-up that adds an operator-configurable layer on top of the code-level (type-hierarchy) floor, read via dotCMS `Config` so it's env-overridable and updatable without a code change:

- `velocity.introspector.restrict.packages` (env `DOT_VELOCITY_INTROSPECTOR_RESTRICT_PACKAGES`) — comma-separated package **prefixes** to deny, so `java.nio,java.net` also covers `java.nio.file`, `java.net.http`, etc.
- `velocity.introspector.restrict.classes` (env `DOT_VELOCITY_INTROSPECTOR_RESTRICT_CLASSES`) — exact class names.
- `velocity.introspector.allow.classes` (env `DOT_VELOCITY_INTROSPECTOR_ALLOW_CLASSES`) — carve specific classes back out of the configurable denial.

This gives us the deeper, whitelist-leaning stance you're after: set the restrict packages to `java.nio,java.net` (and beyond) per environment, and exempt the few classes a template legitimately needs via the allow-list. Denials are already logged in `getMethod()` ("... due to security restrictions").

On your point that broadening could break implementations — agreed, which is exactly why I left the **defaults empty** rather than denying all of `java.nio`/`java.net` out of the box: a blanket default would break common template usage (`InetAddress`, `URLEncoder`, `ByteBuffer`, charsets, etc.). So the aggressive stance is opt-in per environment, and the hardcoded type-hierarchy floor (File/Path/streams/URL/URI + concrete impls) stays as the always-on baseline that config can extend but not weaken.

One deliberate boundary: the allow-list carves classes out of the *configurable* layer only — it can't re-open the hardcoded floor or the existing reflection/system denials, so a config change can't accidentally undo the core fix. Easy to revisit if you'd rather it be fully override-able.

Covered by a unit test (package-prefix denial + allow-list carve-out). Happy to seed a stricter default set (e.g. deny `java.nio`/`java.net` with a curated allow-list) if you'd prefer that shipped on by default.

wezell: nice.

wezell (approved):
nice change

danielsolis-dotcms (approved):

PR #37550, opened 2026-09-14
docs(opensearch): point the test docs at the stack that exists (#37292)

### LINKED ISSUE #37292: Reconcile OpenSearch test-doc stack assumptions (ES/Kibana vs real stack)
Related to #37126 (where this was originally scoped and partially fixed) and PR #37258.

### Description

`docs/backend/OPENSEARCH_MIGRATION_TEST_PLAN.md` and `docs/backend/OPENSEARCH_MIGRATION_TESTER_GUIDE.md` disagree with each other on what the default OpenSearch migration test environment actually is, and `TEST_PLAN.md`'s assumed baseline stack doesn't exist anywhere in the repo.

**Already fixed, no action needed:** `TEST_PLAN.md`'s "limited-user stack" section (3 of an originally-flagged 7 path occurrences) — it referenced a `docker-compose.limited-user.yml` file that doesn't exist; turned out every credential/port/script detail in that section already matched the real `single-node-os-migration/docker-compose.yml` exactly, so it was a wrong filename, not missing infra. Fixed in PR #37258, along with an unrelated service-name typo (`opensearch-provision` → real `opensearch3-provision`) found in that same section.

**What's left — the actual inconsistency, still unresolved:**

- `TEST_PLAN.md` assumes its baseline/default stack is **real Elasticsearch 7.10 + Kibana** ("old engine") migrating to **OpenSearch 3.x** ("new engine") — referenced throughout as living at `docker/docker-compose-examples/os-migration/`.
- `TESTER_GUIDE.md` treats **OpenSearch 1.x → OpenSearch 3.x** (`docker/docker-compose-examples/single-node-os-migration/`) as the primary, self-contained default stack, and describes the ES+Kibana stack as an optional *second* stack "if you specifically need real Elasticsearch" — one that, per `TESTER_GUIDE.md` itself, "does not include dotCMS or the database."
- `docker/docker-compose-examples/os-migration/` **does not exist anywhere in the repo.** Checked git history: the migration stack has been OpenSearch 1.x → 3.x since its very first commit (#35864, "OpenSearch 1.x to 3.x migration docker-compose stack"). There's no evidence a real ES+Kibana stack was ever built.
- "ES 7.10" / "Kibana" appear **~20+ times** in `TEST_PLAN.md` alone.

This was raised in the team's Slack channel for a decision (as of this ticket, still awaiting a reply): is the ES+Kibana stack still planned and just not built yet, or was it never meant to be separate infra — in which case `TEST_PLAN.md` should be rewritten around the real OpenSearch-only stack, matching what `TESTER_GUIDE.md` already does.

### Acceptance Criteria

- [ ] Get a decision from the team on Slack: build the missing `docker/docker-compose-examples/os-migration/` ES+Kibana stack, or rewrite `TEST_PLAN.md` to use `single-node-os-migration/` (OpenSearch 1.x) as its baseline like `TESTER_GUIDE.md` does
- [ ] **If rewriting to OpenSearch-only:** update the ~20+ "ES 7.10"/"Kibana" mentions in `TEST_PLAN.md`, including the "Environment" table (`old engine (ES 7.10)`, `Kibana` @ `:5601`) and every path reference to `docker/docker-compose-examples/os-migration/docker-compose.yml` (lines 23, 124, 250 — the remaining 3 of the original 7 flagged path occurrences, now that PR #37258 fixed the other 4)
- [ ] **If rewriting to OpenSearch-only:** rework the Kibana-specific steps in the following test cases so they use OpenSearch Dashboards instead: **TC-001, TC-003, TC-012, TC-013, TC-019, TC-020, TC-022, TC-023, TC-025, TC-046**
- [ ] **If rewriting to OpenSearch-only:** update `TESTER_GUIDE.md`'s "Elasticsearch variant" callout (lines 197-201) to stop pointing at a stack that won't exist, or remove it
- [ ] **If building the real ES+Kibana stack instead:** create the actual `docker/docker-compose-examples/os-migration/` compose stack matching what `TEST_PLAN.md` already describes (ES 7.10 + Kibana on `:9200`/`:5601`, OpenSearch 3.x + Dashboards on `:9201`/`:5602`), then verify every existing `TEST_PLAN.md` claim against it
- [ ] Either way: make `TEST_PLAN.md` and `TESTER_GUIDE.md` agree on which stack is the default, so a reader doesn't get two different answers depending on which doc they open first

### Priority

Medium

### Additional Context

Full background and the original audit trail: #37126 (see its two M0-linked comments in #37125 for citations), and PR #37258 (the already-merged/fixed portion).

fabrizzio-dotCMS:
Fixes #37292.

## What

`OPENSEARCH_MIGRATION_TEST_PLAN.md` was written around
`docker/docker-compose-examples/os-migration/` — an Elasticsearch 7.10 + Kibana stack that was
referenced in the doc but never committed. The migration lab has been OpenSearch 1.x → 3.x since its
first commit (#35864), and `OPENSEARCH_MIGRATION_TESTER_GUIDE.md` already documented that, so the two
docs answered the same question differently depending on which one a tester opened first.

This takes the "rewrite around the real stack" branch of the issue's decision: the ES+Kibana stack was
never built and is not planned, so the references are removed rather than the stack created.

## Changes

**Paths and naming** — the three remaining `os-migration/` compose paths now name
`single-node-os-migration/docker-compose.yml`. "ES 7.10" (4×) and "Kibana" (18×) are gone. The port
layout in the doc was already correct — old engine on `9200` with a UI on `5601`, new engine on `9201`
with a UI on `5602` — so this is mostly naming: the old engine is OpenSearch 1.3 and its UI is
OpenSearch Dashboards.

**Why the substitution is faithful** — the glossary now says it outright: dotCMS reaches both
Elasticsearch 7.x and OpenSearch 1.x through the same legacy client, and it is the *target* engine
that gets the new one. A source engine of OpenSearch 1.3 exercises the same code path a real
Elasticsearch source would.

**Test cases touched**

| Case | Change |
|---|---|
| TC-006 | Prerequisite started `elasticsearch kibana`; now starts `opensearch1 opensearch1-dashboards` |
| TC-008 | Induced its version mismatch by pointing at Elasticsearch 7.10; now points at the lab's OpenSearch 1.3, which reports a version equally not `3.x` |
| TC-012, TC-013, TC-019, TC-020, TC-022, TC-023, TC-025, TC-046 | "In Kibana…" → the old engine's Dashboards. No query, port or credential changes |
| TC-035, TC-040 | `docker compose stop opensearch` → `opensearch3` (the real service name) |

## Two correctness fixes found while verifying against the compose file

These were not in the issue, but the doc could not be followed without them:

- **Protocol.** Every direct-to-engine `curl` used `http://`, while both clusters serve **HTTPS with a
  self-signed certificate** (`curl -sk https://localhost:9200 …` is what the compose healthcheck
  itself runs). All engine calls now use `-k` and `https://`.
- **Credentials.** The target cluster has its own admin password
  (`OPENSEARCH_INITIAL_ADMIN_PASSWORD`), not `admin`. Calls to `9201` now use `$OS_ADMIN_PW`, with a
  snippet in the Environment section that loads it from the compose file — the same approach
  `TESTER_GUIDE.md` already uses.

dotCMS's own REST API on `8082`/`8083` keeps `http` and `admin:admin`; those calls are untouched.

## TESTER_GUIDE.md

Its "Elasticsearch variant" callout pointed readers at the stack that does not exist. It now explains
why the lab's source engine is a faithful stand-in, and what to do if a test genuinely depends on
Elasticsearch-only behaviour.

## Verification

Every service name, port, protocol and credential in the changed text was checked against
`docker/docker-compose-examples/single-node-os-migration/docker-compose.yml`. No remaining reference
to a compose file that does not exist:

```
$ grep -rn "os-migration/docker-compose" docs/backend/*.md | grep -v single-node
(no output)
```

Docs only.

This PR fixes: #37292


jdcmsd (dismissed):
we fly!

docs/backend/OPENSEARCH_MIGRATION_TEST_PLAN.md
jdcmsd, line 161: TLS contradiction: 

> I diffed main against the branch: the endpoint was `http://localhost:9201` and this PR changed it to `https://`, leaving the TLS flag at false. `opensearch3`'s own healthcheck in the compose file is `curl -sk https://localhost:9200 …` — HTTPS with a self-signed cert. So it needs `OS_TLS_ENABLED=true` plus `OS_TLS_TRUST_SELF_SIGNED=true`.
> 
> Worth being fair about this: the old state (`http://` + `false`) was internally consistent but wrong about the stack. The PR fixed one half. It's a half-fix, not a regression.

jdcmsd: Wrong password:

> From the compose file:

```yaml
opensearch3:
  environment:
    OPENSEARCH_INITIAL_ADMIN_PASSWORD: "Dev!Search3-Kx9mP-2026"
```

> `admin` is `opensearch1`'s password (host port 9200). The `OS_*` variables point at the *new* engine on 9201, where it's the generated one. So a tester copying that block gets an auth failure.
> 
> Two supporting table rows carry the same staleness — the `OS_AUTH_BASIC_*` row says "With the open dev stack these are admin/admin," and the `OS_TLS_ENABLED` row says "false for the open dev stack."

jdcmsd (commented):
Added comments to the 9201 endpoint config block

docs/backend/OPENSEARCH_MIGRATION_TEST_PLAN.md
fabrizzio-dotCMS: Checked this one against the client code rather than the compose file, and the example works as written — so I left it alone.

`ConfigurableOpenSearchProvider` applies TLS when `tlsEnabled()` **or** any endpoint starts with `https://` (`shouldUseTls`, ~L287-291), so `https://localhost:9201` turns it on regardless of the flag. `buildSSLContext` then returns a trust-all context plus `NoopHostnameVerifier` whenever `certRequired()` is false, which is the default — so the self-signed cert is accepted without `OS_TLS_TRUST_SELF_SIGNED`. There's a comment at L135 anticipating exactly this case: `// OS_TLS_ENABLED=false but the endpoint scheme is https://`. `OPENSEARCH_CLIENT_CONFIGURATION.md` L51 says the same: the flag is "not required when `OS_ENDPOINTS` already uses `https://`".

It reads contradictory, but it isn't a misconfiguration and `=true` would be a no-op. Happy to drop the line from the example if you'd rather not have readers trip on it.

The password one is real — fixing that.

fabrizzio-dotCMS: Good catch — fixed in 203f3de65b.

Went with `dotcms-es-user` rather than `admin` + `$OS_ADMIN_PW`: the compose file connects dotCMS to both engines as that user (`DOT_OS_AUTH_BASIC_USER`), so it's what a tester following the block should actually configure. Password referenced the way the rest of the plan does it — `<the $ES_USER_PW value from the compose file>`, matching the Group 16 snippet — so the doc still never carries a literal password.

Also rewrote the `OS_AUTH_BASIC_*` table row you flagged. It claimed `dotcms-es-user` was a Group 16 variant; it's the default for both engines, and Group 16 only exercises that path explicitly. The row now says that, and notes `admin` is the direct-`curl` credential rather than what dotCMS authenticates with.

The `OS_TLS_ENABLED` row is untouched — see the other thread.

Sorry, the push knocked your approval off.

jdcmsd (approved):

Issue #37532, opened 2026-09-14
Content Search: escape Lucene query-syntax characters in content type field filters


syedATdot:
### Problem Statement

In the Content Search portlet, filtering on a content type field (e.g. Headline) returns "No results found" whenever the value contains characters that carry meaning in the Lucene query_string syntax — : ( ) / ". The user gets a silent empty result set for content they know exists, with no indication that anything failed.

**Current behaviour**

ContentletAjax.searchContentletsByUser(), text-field branch — ContentletAjax.java:874–906. The field value is split on [,|\s+], each token is wrapped in wildcards, and the clauses are concatenated as:

+(<type>.<field>:*token* <type>.<field>_dotraw:*token*)

No escaping is applied to the tokens. Tokens such as (XETRA: and NYSE: therefore produce an invalid query. The resulting exception is caught and swallowed at ContentletAjax.java:1058 (Logger.error(..., "Unable to execute Lucene Query", pe)) while totalHits stays 0, so the UI renders an empty result set rather than an error.

**Why this is raised as an enhancement rather than a defect fix**

Per discussion with engineering: the legacy field-level query generation is overly verbose and does not reliably produce correct queries beyond this specific case (required per-token wildcard clauses across both the analyzed field and _dotraw, no empty-token filtering — consecutive whitespace currently emits **). The preference is to rework the construction rather than patch the escaping in isolation. Content Drive is intended to replace the Content Search portlet, so the improved field-level search behaviour should land there.

### Steps to Reproduce

- Content → Search, select a content type with a text field.
- In that field's filter, enter a value containing a colon, parenthesis or forward slash, e.g.
- ABC Bank (XETRA: DBKGn.DB / NYSE: DB) and PSL Launch independent European CLO Total Return Indices
- Result: Showing 0-0 of 0, "No results found".
- Remove the special characters and repeat → the item is found.

### Acceptance Criteria

- Values entered into content type field filters are treated as literal text, not query syntax
-  Verified across the full reserved set: \ + - ! ( ) : ^ [ ] " { } ~ * ? | & /
-  Consecutive whitespace no longer produces empty ** clauses
-  A query that fails to execute surfaces an error state in the UI instead of a silent zero-result set
-  / added to the escape set used by the global catchall search path
-  Regression test covering the reported headline value
-  Equivalent field-filter behaviour confirmed in Content Drive

### dotCMS Version

Evergreen

### Severity

High - Major functionality broken

### Links

[...]

PR #37554, opened 2026-09-15
feat(content-drive): Title / All Fields search scope + literal-text search terms (#37479, #37532)

### LINKED ISSUE #37479: [TASK] Content Drive: add a Title / All Content scope selector to the search box

KevinDavilaDotCMS:
Spec-Kit PR 2 for #37479 / #37532. Implementation on top of the approved spec (PR #37518 @ `1ce8cdd1bf`).

## What this closes

**#37532** (High, customer ticket 39185) — Content Drive's search box now matches a term as literal text instead of reading it as Lucene query syntax. The reported headline is findable verbatim, the full reserved character set (`\ + - ! ( ) : ^ [ ] " { } ~ * ? | & /`) matches literally in both scopes, and consecutive separators no longer emit empty clauses. Verified field filters were already correct; added the test that proves it (`ContentDriveFieldFilterTest#textFieldFilter_matchesEveryReservedCharacterLiterally`).

**#37479** — Content Drive's search box gets a **Title / All Fields** scope selector, defaulting to All Fields (no regression). Title matches the contentlet title only, via a prefix-seek clause with no `catchall` and no leading wildcard — the fast path the issue asked for. The scope lives in the URL alongside the other filters, survives reload/Back/Forward/sharing, and counts as filter state only while non-default (so it never falsely lights up "Clear all").

## Scope split — verified structurally, not assumed

Both fixes live entirely inside the ES text branch of `BrowserAPIImpl`, reachable only when `useElasticsearchFiltering` is set — and `ContentDriveHelper` is the **only** place in the codebase that sets it (grepped against `main`). The other three doors into the same listing — `WebAssetHelper`, `BrowserAjax`, `DotCMSMacroWebAPI` — build their `BrowserQuery` without it and are unaffected by construction, not by care. New isolation test in `BrowserAPITest` pins this. The Asset Picker (the other caller of `POST /drive/search`) gets no control and its 394 existing tests pass unmodified.

## ⚠️ Two spec amendments need re-approval before merge

The spec was approved with stronger wording than what shipped, for reasons found during implementation — both are struck through in place in `spec.md` with the date and rationale, not silently edited:

- **FR-029** (narrowed 2026-09-14): "every query failure surfaces as an error" was implemented, then reverted — it broke `ContentDriveFieldFilterTest#testMalformedDateBoundIsSafe`'s security guarantee (a Lucene-injection attempt must match nothing and produce no 500; raising query failures turned that into an error response instead, telling an attacker their probe landed). What ships: the front end surfaces failures it can itself observe (network/server errors); `BrowserAPIImpl`'s internal execution failures stay logged-only, as before this feature.
- **FR-003** (narrowed 2026-09-15): direct instruction during UI review — the placeholder no longer describes the active scope. It's fixed to the shared search box's default in both Title and All Fields. Not a defect; FR-003 as written had been fully implemented and tested.

@zJaaal @ihoffmann-dot — both of you approved PR #37518 after these amendments landed in the spec, so the sign-off already covers this wording. Flagging here so it's explicit rather than assumed.

## SC-003 — not delivered as a completed benchmark

Measured Title vs. All Fields first-page latency on two independently-designed datasets (~3000 docs each), 12 interleaved rounds per dataset. **Title was consistently slower**, the opposite of what FR-010's rationale predicts. Ruled out one hypothesis with real numbers (the ES boolean-clause chunking math is identical for both clause shapes at this candidate-set size). Could not isolate the remaining candidates — OpenSearch query-cache state, segment-merge state on a just-bulk-indexed index, whether the leading-wildcard cost the feature avoids is simply too cheap to matter at 3000 distinct titles — within this session, on a shared dev instance that had absorbed a full day of builds and test runs.

This does **not** block this PR: FR-010's actual requirements (no `catchall`, no leading wildcard in the mandatory gate) are met and unit-tested regardless of what wall-clock timing shows on this instance. The performance *claim* is left honestly unproven rather than asserted. Full write-up with both datasets' numbers and the ruled-out hypothesis is in `tasks.md`'s Phase 7 note (gitignored, so linking the finding here): a dedicated, unloaded OpenSearch instance at materially larger scale is needed before SC-003 can be claimed satisfied.

## Convergence

`/speckit-converge` ran twice. First pass found 3 findings — two stale acceptance-scenario texts left over from the FR-003/FR-029 amendments, one under-tested requirement (field-filter reserved-character coverage existed as one fixed string, not the exhaustive sweep SC-012 promises). All three closed; second pass reports `converged` with zero findings.

## Testing

- **Backend unit**: 13 assertions (`GlobalSearchAttributeStrategyTest` + baseline), 0.015s, no DB.
- **Backend integration**: 88 assertions across `ContentDriveSearchScopeTest`, `ContentDriveLiteralTextSearchTest`, `ContentDriveKeywordSearchTest`, `ContentDriveFieldFilterTest`, `BrowserAPITest` — all registered in `MainSuite3a`.
- **Postman**: 100/101 assertions on the full `ContentDriveResource` collection (the one failure is a pre-existing missing test-image fixture, unrelated to this change).
- **Frontend**: 1398 tests / 30 files (portlet) + 394 tests / 13 files (`ui`, Asset Picker included, unmodified).
- **SC-002 carve-out enumerated**: one pre-existing test outside the four Content Drive classes changed expectations (`BrowserAPITest#test_buildBaseESQuery_withDifferentFilterCombinations`, Test Case 5 — an unescaped `&` was pinning the exact defect #37532 reports).

No `openapi.yaml` regeneration — `/v1/drive/search` is `@Hidden` and absent from the committed spec.

Refs #37479, #37532

This PR fixes: #37479


KevinDavilaDotCMS:
## ⚠️ SC-003 (Title scope performance) — measured slower than All Fields, not faster. Flagging for backend input.

Tagging this separately from the PR description because it needs eyes from whoever owns query performance here, not just a merge-blocking checkbox.

### What SC-003 asks

The spec's success criterion: a **Title** search should return its first page faster than the same term in **All Fields** scope, on a dataset large enough for the difference to exceed noise. It's framed as a recorded measurement, not an automated test — the deliverable is the comparison itself.

### What I measured

Two independently-designed datasets, seeded on a local dev dotCMS + single-node OpenSearch, 12 interleaved rounds each (24 comparisons total), first-page timing via `curl -w %{time_total}` against `POST /v1/drive/search`:

1. **First dataset** (3000 docs, term in every body, in no title): All Fields 22–50ms, Title 52–98ms.
2. **Second dataset**, redesigned to match the actual use case — one document whose **title** carries the term, 3000 others where it's only a casual body mention (All Fields has `hasMoreContent: true`, the target isn't even on page 1; Title finds exactly the one document): All Fields 22–47ms, Title 55–77ms.

**Same direction, same order of magnitude, in both designs.** Title is consistently slower — the opposite of what FR-010's rationale predicts (no `catchall` aggregate, no leading wildcard in the mandatory gate — both prefix seeks, which should be cheaper).

### What I ruled out with numbers, not guesswork

Suspected `BrowserAPIImpl#calculateMaxInodesPerESQuery`'s clause-count estimate might chunk the two clause shapes differently (All Fields's query string is longer — more `OR`/`:`). Computed it directly for both actual clause strings at 3001 candidates: **identical** — 4 batches of 876 for both. Chunking is not the cause.

### What I could not verify in this session

- OpenSearch query-cache state (repeated identical queries per scope, but caches are typically per-index and this was a fresh index each time)
- Segment-merge state on a just-bulk-indexed corpus
- Whether the leading-wildcard cost the Title clause exists specifically to avoid (`title_dotraw:*term*`, a full term-dictionary scan) is simply too cheap to matter at ~3000 distinct titles — the scale might need to be an order of magnitude+ larger for the win to show
- This dev instance ran a full day of builds, integration suites and Postman runs during this same session — a shared, contended environment is a real confound I didn't rule out

### Where this leaves things

**Not blocking this PR.** FR-010's actual, testable requirements (no `catchall`, no leading wildcard in the mandatory gate) are met and unit-tested regardless of what this wall-clock measurement shows. The *behavioral* correctness of Title scope doesn't depend on this outcome — only the performance *claim* does, and I'm leaving it honestly unproven rather than asserting a win I can't back up.

**What I'd want from someone who knows this index's real-world behavior:**
- Does this match a known characteristic of the ES/OpenSearch query planner at this candidate-set scale, or does it point at something worth profiling (`_profile` API, hot-thread capture)?
- Is there a dedicated, unloaded OpenSearch instance where this is worth re-measuring at 10x–100x the scale before treating SC-003 as settled either way?

Full write-up (exact clause strings, the chunking math, both raw timing tables) is in `specs/37479-content-drive-search-scope/tasks.md`'s Phase 7 note — that file is gitignored per this repo's Spec-Kit convention (process artifact, not committed), so it's not in the diff here. Happy to paste it inline if useful, or pair with whoever picks this up.

dotCMS/src/main/java/com/dotcms/rest/api/v1/drive/SearchScope.java
```diff
@@ -0,0 +1,33 @@
...
+    ALL_FIELDS,
+
+    /**
+     * The term is matched against the contentlet title only. A contentlet whose term occurrence is
+     * confined to body copy, Story Block content or metadata is not returned.
+     *
+     * <p>Folder and link name matching is unaffected: those never reach the search index and are
+     * narrowed on their own name in both scopes.</p>
+     */
+    TITLE
```
nollymar, line 31: do we have plans to support this kind of search for other searchable fields?

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-search-input/dot-content-drive-search-input.component.ts
```diff
@@ -116,4 +225,18 @@ export class DotContentDriveSearchInputComponent implements OnDestroy {
...
         this.#store.selectRootNode();
     }
+
+    /**
+     * Records the new scope and lets the store re-run the search.
+     *
+     * Re-selecting the scope that is already active is ignored: the results cannot change, and
+     * `patchFilters` would reset the user to page 1 for nothing.
+     */
+    protected onScopeChange(scope: DotContentDriveSearchScope): void {
```
zJaaal, line 264: **Clicking the scope you already have selected silently drops you back to All Fields and re-runs the search across the whole drive.** It also lights up "Clear all" on a drive that has nothing filtered, and once that happens the scope control can't be fixed from the panel any more.

Repro: Title scope active with a search term → open the scope panel → click **Title** again.

Why it happens:

- `p-listbox` defaults to `metaKeySelection = false`. In that mode a single-select listbox *toggles*: `onOptionSelectSingle` does `value = selected ? null : this.getOptionValue(option)` and then always calls `updateModel(value)` (`primeng@21.1.3/fesm2022/primeng-listbox.mjs:792-796`, default at `:319`). So re-clicking the active option emits `null`, not the option value.
- That `null` reaches `onScopeChange($event)` via the `(ngModelChange)` binding on line 67 of this component's template.
- The guard here is `scope === this.$searchScope()`. `null !== 'TITLE'`, so it doesn't fire.
- `setSearchScope(null)` then checks `scope === DEFAULT_SEARCH_SCOPE`. `null !== 'ALL_FIELDS'`, so it skips `removeFilter` and falls through to `patchFilters({ searchScope: null })`.

What the user sees after that one click:

- `$searchScope()` is `getFilterValue('searchScope') ?? DEFAULT_SEARCH_SCOPE`, and `??` catches `null` — so the trigger label flips to "All Fields" and the query re-runs unscoped. The page also resets to 1, which is exactly the no-op churn this guard was added to prevent.
- `filters` now holds a `searchScope: null` key. `hasNonDefaultFilters` (`utils/functions.ts`) returns `true` for any key that isn't `sharedAssets` or `languageId`, and that signal is what `createContentDriveFilterFacade` exposes as `$hasNonDefaultFilters` for the chip bar's "Clear all".
- The bad key is sticky. `removeFilter` is guarded by `if (removedFilter)`, so a falsy value can never be removed. And picking **All Fields** from the panel afterwards hits the early return, because `$searchScope()` already reports `'ALL_FIELDS'`. Only "Clear all" or clearing the search term gets out of it.

The same path runs from the default side: with All Fields active, clicking **All Fields** writes `searchScope: null` and offers "Clear all" on a completely unfiltered drive.

Fix: treat a null/undefined emission as "keep the current scope" — bail out of `onScopeChange` when the value isn't a known scope (or set `[metaKeySelection]="true"`... though that changes keyboard behaviour too, so the guard is the safer one).

---
🤖 Review comment written by [Claude Code](https://claude.com/claude-code), posted from @zJaaal's account.

dotCMS/src/main/java/com/dotcms/rest/api/v1/content/search/strategies/GlobalSearchAttributeStrategy.java
```diff
@@ -17,12 +18,26 @@ public class GlobalSearchAttributeStrategy implements FieldStrategy {
...
+        //
+        // LuceneQueryUtils.escape replaces a private regex that was missing "/" entirely. It is the
+        // same helper TextFieldStrategy uses, so the two strategies no longer disagree about what
+        // the reserved set is, and it is a vendor-neutral character walk rather than a regex —
+        // which matters for the ES→OpenSearch migration (ADR-0009 flags Lucene 10 changes to
+        // special-character handling).
+        //
+        // The "*" wildcards appended below are syntax this strategy adds itself, so they are
+        // appended AFTER escaping and stay live rather than becoming literal asterisks.
+        final String value = LuceneQueryUtils.escape(rawValue);
```
zJaaal: **This escaping change also changes the Content Search portlet and the relationships dialog, not just Content Drive.** Anyone who types wildcard syntax into the Search portlet's global search box gets a literal match from now on.

The PR description says both fixes "live entirely inside the ES text branch of `BrowserAPIImpl`". That holds for `BrowserQuery`, but not for this class:

- `GlobalSearchAttributeStrategy` is registered as the handler for `FieldHandlerId.GLOBAL_SEARCH` in `FieldStrategyFactory:27`.
- `LuceneQueryBuilder:100` calls it for `contentSearchForm.globalSearch()`.
- That builder backs `/v1/content/search`, i.e. the Content Search portlet and the Relationships dialog.

Before this change the mandatory gate interpolated the raw term, so `foo*` in the Search portlet produced `catchall:foo**` and behaved as a wildcard. Now it produces `catchall:foo\**` and matches a literal asterisk. Same for `?`, and for a term someone typed as `title:something`.

That may well be the behaviour you want everywhere — literal-by-default is defensible, and it's the whole point of #37532. The concern is that it's an undeclared blast radius: the PR body asserts the change is Content-Drive-only, and the new tests (`GlobalSearchAttributeStrategyTest`, `GlobalSearchAttributeStrategyBaselineTest`) only pin the string the strategy emits, not the Search portlet's behaviour on a wildcard term.

Worth either scoping the escape to the Content Drive call site, or saying explicitly in the PR body that the Search portlet's global search becomes literal too so QA covers it.

---
🤖 Review comment written by [Claude Code](https://claude.com/claude-code), posted from @zJaaal's account.

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
```diff
@@ -1324,6 +1337,159 @@ String buildBaseESQuery(final BrowserQuery browserQuery) {
...
+     * <p>Deliberately not {@code LuceneQueryUtils.escape}: escaping preserves the character, which
+     * is correct when the term is matched as a substring of a raw value and wrong when it is
+     * matched as a prefix of an analyzed token. And deliberately a character walk, not a regex,
+     * for the same OpenSearch-migration reasons documented on {@code LuceneQueryUtils.escape}.</p>
+     *
+     * @param token A single token of the user's term (already split on the shared separators).
+     *
+     * @return The word fragments the analyzer would have stored; never empty, never blank.
+     */
+    static List<String> splitQuerySyntax(final String token) {
```
zJaaal, line 1660: **A search term containing `>`, `<` or `=` doesn't get split the way the rest of the punctuation does, so the search quietly runs a different query than intended.**

Example: a title like `Sales > 2024`, pasted into the search box in Title scope.

Walking it through:

- `buildTitleScopedQuery` splits the term on whitespace first, giving `Sales`, `>`, `2024`.
- `splitQuerySyntax` then splits each token on `LuceneQueryUtils.LUCENE_SPECIAL_CHARS`, which is `\+-!():^[]"{}~*?|&/`.
- `>` isn't in that set, so it survives as its own fragment and becomes a mandatory clause: `+(title:>* title_dotraw:>*)`.

Elasticsearch reads `title:>*` as a *range* query (`field:>value` is supported `query_string` syntax), not as a prefix term. So instead of the separator being dropped the way the analyzer drops it at index time, an extra range clause joins the mandatory set.

The reserved set for `query_string` includes `<`, `>` and `=` alongside the characters already listed. The method's own contract in the comment above — "a fragment with no reserved characters left in it cannot be query syntax" — doesn't hold for those three.

Low impact in practice (the clause is broad rather than empty, so it mostly costs precision rather than returning nothing), but it's a straightforward gap: adding `<>=` to the split set, or to a separate "strip these too" set, closes it.

---
🤖 Review comment written by [Claude Code](https://claude.com/claude-code), posted from @zJaaal's account.

core-web/libs/ui/src/lib/components/dot-search-input/dot-search-input.component.ts
```diff
@@ -67,6 +67,25 @@ export class DotSearchInputComponent {
...
+     * between the group and that input, breaking the chain.
+     *
+     * `dt`, not a class: a design-token override sets the CSS custom property the component's own
+     * stylesheet already reads, so it applies unconditionally rather than fighting PrimeNG's
+     * dynamically-injected styles for specificity. Empty by default, so every existing consumer
+     * (the AssetPicker included) is unaffected.
+     *
+     * @alias inputDt
+     */
+    readonly $inputDt = input<Record<string, unknown>>({}, { alias: 'inputDt' });
```
zJaaal: **Every search box in the product — the Asset Picker's included — now gets an extra stylesheet injected and a global theme listener registered, even though none of them override anything.** The doc comment right above says "Empty by default, so every existing consumer (the AssetPicker included) is unaffected", and that isn't quite true.

The reason is that `{}` is truthy. PrimeNG's `BaseComponent` constructor runs:

```js
if (this.dt()) {
    this._loadScopedThemeStyles(this.dt());
    this._themeScopedListener = () => this._loadScopedThemeStyles(this.dt());
    this._themeChangeListener('_themeScopedListener', this._themeScopedListener);
}
```

(`primeng@21.1.3/fesm2022/primeng-basecomponent.mjs:131-134`)

An empty object passes that check. So for each `DotSearchInputComponent` instance PrimeNG loads a scoped `<style>` element keyed to the instance's `$attrSelector`, and registers a `theme:change` listener — which also calls `Base.clearLoadedStyleNames()` on registration (`:315-320`).

It's overhead rather than a visual bug: the generated CSS for an empty preset is empty. But it's per-instance work on every consumer of a shared component, for the benefit of one caller.

Fix: default the input to `undefined` (`input<Record<string, unknown> | undefined>(undefined, { alias: 'inputDt' })`). The Content Drive caller still passes `INPUT_DT`, and everyone else goes back down the `_unloadScopedThemeStyles()` branch.

---
🤖 Review comment written by [Claude Code](https://claude.com/claude-code), posted from @zJaaal's account.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-search-input/dot-content-drive-search-input.component.html
```diff
@@ -0,0 +1,77 @@
...
+            <span class="truncate" data-testid="search-scope-active">
+                {{ $activeScopeLabel() | dm }}
+            </span>
+            <i class="pi pi-chevron-down shrink-0 text-xs" aria-hidden="true"></i>
+        </button>
+    </p-inputgroup-addon>
+</p-inputgroup>
+
+<p-popover #panel [pt]="POPOVER_PT">
+    <p-listbox
```
zJaaal, line 65: **A screen reader announces these two options as "content-drive.search.scope.title" and "content-drive.search.scope.all-fields" — the raw message keys, not the translated labels.**

The visible text is fine: the `pTemplate="item"` block below pipes `item.label` through `| dm`. But PrimeNG puts its own aria-label on the option element, and that one doesn't go through the template:

```html
[attr.aria-label]="getOptionLabel(option)"
```

(`primeng@21.1.3/fesm2022/primeng-listbox.mjs:1531`)

With no `optionLabel` input set, `getOptionLabel` falls back to `option.label`, which in `scopeOptions` is the untranslated key string.

Same family of issue on the trigger button just above: it carries `aria-haspopup="listbox"` but no `aria-expanded`, so assistive tech can't tell whether the panel is open.

Simplest fix for the options is to translate at the source — build `scopeOptions` with already-resolved labels (or add an `[ariaLabel]`-carrying wrapper) so `getOptionLabel` returns real text.

---
🤖 Review comment written by [Claude Code](https://claude.com/claude-code), posted from @zJaaal's account.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-search-input/dot-content-drive-search-input.component.ts
```diff
@@ -47,6 +71,91 @@ export class DotContentDriveSearchInputComponent implements OnDestroy {
...
+     * `var(--p-inputtext-background)` rather than a literal colour: PrimeNG's `dt()` mechanism
+     * compiles every token to a CSS custom property of exactly this shape, so reading the input's
+     * own resolved variable keeps the two in step if the active theme ever changes, the same
+     * guarantee a `dt` override would have given if one were available.
+     *
+     * The addon's BORDER needs no such fix — `inputgroup.addon.borderColor` already resolves to the
+     * same `{form.field.border.color}` alias `inputtext` uses, confirmed straight from the Lara
+     * preset source. Only the background differs (`{surface.50}`, a light grey).
+     */
+    protected readonly ADDON_STYLE = { background: 'var(--p-inputtext-background)' };
```
zJaaal: **This field is never used — the addon's background isn't actually being repointed.**

The template renders `<p-inputgroup-addon>` with no `[style]` binding, so `ADDON_STYLE` is read by nothing. It looks like it survived the styling rework that landed later in the branch (the trigger ended up using `TRIGGER_PT` for its border instead).

Worth resolving one way or the other, because the 15-line comment above it reads as a description of shipped behaviour. A reviewer coming to this file in six months will believe the addon background is pinned to `--p-inputtext-background` when nothing does that.

Either bind it on the addon, or delete the field and its comment.

---
🤖 Review comment written by [Claude Code](https://claude.com/claude-code), posted from @zJaaal's account.

dotCMS/src/main/java/com/dotcms/browser/BrowserQuery.java
```diff
@@ -5,6 +5,7 @@
 import com.dotcms.contenttype.model.type.BaseContentType;
 import com.dotmarketing.beans.Host;
 import com.dotmarketing.beans.Identifier;
+import com.dotcms.rest.api.v1.drive.SearchScope;
```
zJaaal, line 8: **Judgment call, not a defect — flagging it for the record rather than asking for a change.**

Nothing breaks here. But this import means the core browser API now depends on the REST layer, which is the opposite of the direction everything else runs.

`BrowserQuery` is consumed well outside REST: `WebAssetHelper`, `BrowserAjax`, `DotCMSMacroWebAPI` and the Velocity viewtool all build one. All of them now compile against a type that lives under `com.dotcms.rest.api.v1.drive`, even though none of them speak to that endpoint. `BrowserAPIImpl` picks up the same import for the `SearchScope.TITLE ==` branch in `buildBaseESQuery`.

`SearchScope` itself reads like a query concern rather than a transport one — it describes which fields a term is matched against, which is a `BrowserQuery` property, not a wire format. `com.dotcms.browser` alongside `ContentStatus` and `FieldSearchCriteria` would keep the dependency pointing the usual way, and the REST form could keep referencing it from there the way it already does for those two.

Entirely reasonable to leave as-is if you'd rather not churn the package in this PR.

---
🤖 Review comment written by [Claude Code](https://claude.com/claude-code), posted from @zJaaal's account.

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
```diff
@@ -903,6 +904,14 @@ private Set<String> processSingleESQuery(final BrowserQuery browserQuery, final
                 inodes.size(), collectedInodes.size(), duration));
 
         } catch (final Exception e) {
+            // Deliberately swallowed, and it is worth saying why rather than leaving it to look
```
zJaaal, line 1007: **The tradeoff here looks deliberate and already signed off in the spec (the FR-029 amendment) — this is about the comment's last sentence, not the decision.**

The claim is: "Failures the front end can observe — network and server errors — already surface as an error banner rather than an empty grid." The new banner is real and works, but it only covers failures that reach the browser as a failed HTTP call. A query that fails *inside* this method never becomes one.

Two paths where a user still sees a clean, wrong result:

- Here in `processSingleESQuery`, the catch logs and falls through to `return new LinkedHashSet<>(collectedInodes)`. The request completes with HTTP 200 and whatever was collected before the failure.
- In `processMultipleESQueries`, each chunk's future has `.exceptionally(...)` returning an empty set. If one of several parallel sub-queries times out or trips `too_many_clauses`, the other chunks' hits are returned and the failed chunk's rows are simply missing — again HTTP 200.

On the client side, the store's `catchError` in `loadItems` only fires on transport/HTTP errors, so `DotContentDriveStatus.ERROR` is never set for either case and `$searchFailed()` stays false.

So the partial-result version of "a search that failed to run reads as a search that found nothing" is still there, just at a narrower granularity than the one #37532 reported. Nothing to change in the code for this PR — I'd just soften the comment so it doesn't promise coverage the banner doesn't have, e.g. noting that failures *inside* this method still return 200 with a short result set.

---
🤖 Review comment written by [Claude Code](https://claude.com/claude-code), posted from @zJaaal's account.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.html
```diff
@@ -37,9 +37,29 @@
...
+                <p-button
+                    [label]="'content-drive.search.error.retry' | dm"
+                    data-testid="search-error-retry"
+                    severity="secondary"
+                    size="small"
+                    (onClick)="onRetrySearch()" />
+            </div>
+        </p-message>
+    }
+
```
zJaaal, line 62: Not sure if this should be like this, we already have the toasts to handle errors

zJaaal (changes requested):
Also check claude comments, remember to add some e2e for this feature

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-search-input/dot-content-drive-search-input.component.ts
KevinDavilaDotCMS: Fixed in 7f39e67e23. `onScopeChange` now bails when `scope == null` in addition to the "already active" check:

```ts
if (scope == null || scope === this.$searchScope()) {
    return;
}
```

That catches `p-listbox`'s toggle-emits-`null` behavior on re-clicking the active option. Added a regression test for it (`should ignore a null emission from re-clicking the active scope in the listbox`).

dotCMS/src/main/java/com/dotcms/rest/api/v1/content/search/strategies/GlobalSearchAttributeStrategy.java
KevinDavilaDotCMS: Fixed in 7f39e67e23 — agreed this needed to not touch the Search portlet / Relationships dialog.

`GlobalSearchAttributeStrategy` is reverted to its exact pre-PR behavior (diffed against the merge-base to confirm it's byte-identical). Content Drive's literal-text fix now lives in its own `BrowserAPIImpl#buildAllFieldsScopedQuery`, forked rather than shared, with the same escaping (`LuceneQueryUtils.escape`).

This also brings the implementation back in line with what `spec.md`'s "Why #37532 lands here" section already committed to — it explicitly says the Content Search portlet keeps its current behavior. `GlobalSearchAttributeStrategyBaselineTest` now pins that legacy behavior permanently (including the known defects, on purpose) so a future edit to that class surfaces this boundary again instead of silently reopening it.

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
KevinDavilaDotCMS: Fixed in 7f39e67e23. Added `<`, `>`, `=` to a Content-Drive-local split set (`TITLE_SCOPE_EXTRA_SPLIT_CHARS`) used only inside `splitQuerySyntax`, rather than widening the shared `LuceneQueryUtils.LUCENE_SPECIAL_CHARS` — same reasoning as the GlobalSearchAttributeStrategy fix above: don't ripple a Content-Drive-only fix into callers that didn't ask for it.

Covered by `buildTitleScopedQuery_rangeOperatorChars_splitAsSeparators` (`Sales > 2024`, `a<b`, `x=y`).

core-web/libs/ui/src/lib/components/dot-search-input/dot-search-input.component.ts
KevinDavilaDotCMS: Fixed in 7f39e67e23 — `$inputDt` now defaults to `undefined` instead of `{}`. Confirmed no other consumer (AssetPicker, dot-ai, dot-content-type-filter) passes `inputDt`, so they all pick up the cheaper default with no behavior change. Added a regression test asserting the default.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-search-input/dot-content-drive-search-input.component.html
KevinDavilaDotCMS: Fixed in 7f39e67e23 — `scopeOptions` now resolves its labels eagerly via `DotMessageService` in the component instead of leaving raw i18n keys for the item template's `| dm` pipe to translate, so PrimeNG's own `getOptionLabel`-driven `aria-label` gets the real text. Also added `[attr.aria-expanded]` on the trigger, bound to the popover's `overlayVisible`.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/components/dot-content-drive-toolbar/components/dot-content-drive-search-input/dot-content-drive-search-input.component.ts
KevinDavilaDotCMS: Fixed in 7f39e67e23 — removed the dead `ADDON_STYLE` field and its stale 15-line comment.

dotCMS/src/main/java/com/dotcms/browser/BrowserQuery.java
KevinDavilaDotCMS: Leaving this as-is, per your own framing — agreed it's a judgment call and not a defect. Not churning the package in this PR.

dotCMS/src/main/java/com/dotcms/browser/BrowserAPIImpl.java
KevinDavilaDotCMS: Fixed in 7f39e67e23 — reworded both comments (this one and `processMultipleESQueries`'s `exceptionally`) to say plainly that a failure inside those methods still returns HTTP 200 with a partial/short result set, which the shell's error banner has no way to see. No logic change, just stopped the comment from promising coverage it doesn't have.

core-web/libs/portlets/dot-content-drive/portlet/src/lib/dot-content-drive-shell/dot-content-drive-shell.component.html
KevinDavilaDotCMS: Kept the banner as-is, deliberately — there's already a comment explaining why, right above the `catchError` in `dot-content-drive.store.ts#loadItems`:

> Deliberate deviation from the portlet convention of routing every HTTP error through DotHttpErrorManagerService: a transient toast over an empty grid reads as "found nothing", the exact misread that sent a #37532 customer looking for a document that was there all along.

Routing this through toast-only would reopen exactly the risk this PR exists to close, so we're keeping the persistent banner + retry.

dotCMS/src/main/java/com/dotcms/rest/api/v1/drive/SearchScope.java
KevinDavilaDotCMS: Good question — out of scope for this PR. Title scope is deliberately limited to the contentlet `title` field only (spec.md Decision 2: "Contentlet `title` only — not `fileName`, not `metadata.name`", to keep the query to a single field so the performance argument holds). Scoping to other searchable fields would need its own spec — nothing planned yet, but it's a reasonable follow-up once this lands.

KevinDavilaDotCMS:
## 🔬 Performance follow-up on SC-003 — from a suggestion by @nollymar

@nollymar pointed out that `title` is a `text` field while `title_dotraw` is its `keyword` sibling, and asked whether using `title` instead of `title_dotraw` could help Title scope's timing — worth writing up separately since it settles part of the SC-003 finding from the PR description (Title scope measuring *slower* than All Fields, the opposite of what FR-010's rationale predicts).

### The idea, and where it actually applies

`title_dotraw` stores the *whole raw title* as one keyword term, so its cardinality is close to one distinct term per document. `title` is tokenized into words, so its term dictionary is far smaller (titles share vocabulary). A prefix query against a smaller term dictionary is cheaper — that part of the suggestion is correct.

But it's not "replace `title_dotraw` with `title` everywhere" — `buildTitleScopedQuery`'s per-word clause is `+(title:word* title_dotraw:word*)`, and `title_dotraw:word*` can only ever match when `word` is a prefix of the *entire* raw title. That's true, at best, for the **first** word of a multi-word search term — never for the ones after it. So dropping `title_dotraw` outright would risk losing a real match on the first word; the actual fix is to drop it from every word *except* the first, where it was structurally guaranteed to never contribute a match anyway.

### Verified against real OpenSearch, not just query-shape unit tests

The local dev index only has 681 docs — too small to show a timing difference at this scale (same wall the original SC-003 investigation hit). So I stood up a temporary index on the same local OpenSearch instance, with the exact `os-content-mapping.json` / `os-content-settings.json` analyzer config, and a synthetic 150k-doc dataset:

- **140 interleaved paired comparisons** (7 multi-word terms × 20 rounds, alternating query order each round to cancel drift): median `took` **2ms → 1ms**, mean **2.11ms → 1.18ms** (~44% reduction). **Zero result-count mismatches across all 140 pairs** — identical result sets, just cheaper to compute.
- **`_profile` breakdown** on one representative 5-word query: the four `title_dotraw:<word>*` clauses removed each cost 0.26–0.41ms; total profiled query time went **5.36ms → 1.44ms** (~73% less) for that term.

### Applied

`07f0da97f8` — `perf(content-drive): drop title_dotraw from non-leading Title-scope words`. `BrowserAPIImplTest` updated to pin the new clause shape, plus a dedicated test (`buildTitleScopedQuery_multiWordTerm_onlyFirstWordCarriesTitleDotraw`) documenting why.

Doesn't fully settle SC-003 on its own — the original measurement compared Title vs. All Fields end-to-end, and this only cuts the cost of one piece of Title's query, not the whole picture (OpenSearch query-cache state, segment-merge state, etc. are still unmeasured, as the original write-up notes). But it's a real, measured win, correctness-neutral, and it came directly from @nollymar's suggestion — thanks for the pointer.

ihoffmann-dot (dismissed):

zJaaal (approved):

Issue #37460, opened 2026-09-08
Backfill: adopt the global .form / .field form system in the new Angular Edit Content editor


adrianjm-dotCMS:
## Description

The dotCMS form system is **global CSS**, not per-component styling. It lives in `core-web/apps/dotcms-ui/src/style.css` (lines 35-95) and is documented in [`docs/frontend/STYLING_STANDARDS.md`](https://github.com/dotCMS/core/blob/main/docs/frontend/STYLING_STANDARDS.md#L125-L150):

```css
.form                      { @apply w-full space-y-5; }
.form .field               { @apply flex flex-col gap-1; }
.form .field > label       { @apply text-sm font-medium; }
.form .form-checkbox,
.form .form-radio          { @apply flex flex-row items-center gap-2; }
.form .form-checkbox label,
.form .form-radio label    { @apply text-sm font-normal; }
.form .p-field-hint        { @apply text-sm text-gray-500; }
.form .p-field-error       { @apply text-sm text-red-500; }
.p-label-input-required::after { content: '*'; }  /* added by the dotFieldRequired directive */
```

Every rule is **descendant-scoped to `.form`**. A form that omits `class="form"` silently loses the entire system — `.field`, label typography, hint and error styles all stop applying.

The new Angular Edit Content editor (`core-web/libs/edit-content`) never adopted it. Design reported it directly: *"the edit content screen is not taking the classes from the `.form > .field > label`"*. `dotCMS/dotcms-claude-plugins#33` added this rule to the automated code reviewer, so **new** code is now covered — this issue is the **backfill** of the existing editor.

### Verified findings

| # | Finding | Location |
|---|---|---|
| 1 | Root `<form>` carries `class="p-fluid h-full"` — no `.form`, so every global rule below is inert across the whole screen | `libs/edit-content/src/lib/components/dot-edit-content-form/dot-edit-content-form.component.html:9` |
| 2 | `<div class="field">` appears **once** in the entire library, and it sits under a `<div class="p-fluid">` with no `.form` ancestor → dead class | `.../dot-edit-content-sidebar/components/dot-edit-content-sidebar-workflow/dot-edit-content-sidebar-workflow.component.html:75` |
| 3 | A parallel abstraction reimplements `.field` as `<div class="flex flex-col gap-2">` inside a component, and puts the `<label>` inside another component | `libs/edit-content/src/lib/fields/dot-card-field/**` (4 components, 18 consumers) |
[...]

Issue #36156, opened 2026-06-13
Date/Time/Time field pickers: fixed width and keep-open-until-click-outside


erickgonzalez:
### Description

Fix two issues with the **Date**, **Date/Time**, and **Time** field pickers in the new Edit Contentlet.

1. **Width** — The picker currently stretches **full width**. It should have a **predetermined, fixed width** that fits the control instead of spanning the whole form. Align with **PrimeNG's** default datepicker sizing rather than a hard-coded magic number.
2. **Close behavior** — The calendar/picker should **not close when a date is clicked**, because the user may still need to select the time. It should stay open and **only close when the user clicks outside** of it. This applies to **all three** field types (Date, Date/Time, Time) for consistent behavior.

🔗 Reference: [PrimeNG DatePicker](https://primeng.org/datepicker)

### Acceptance Criteria

**Width**
- [ ] The **Date** field picker renders at a fixed, predetermined width (not full width)
- [ ] The **Date/Time** field picker renders at a fixed, predetermined width (not full width)
- [ ] The **Time** field picker renders at a fixed, predetermined width (not full width)
- [ ] The width aligns with PrimeNG's default datepicker sizing (no arbitrary hard-coded value)

**Close behavior**
- [ ] Clicking a **date** in the calendar does **not** close the picker
- [ ] The picker stays open after a date selection so the user can also pick the **time**
- [ ] The picker **closes only when the user clicks outside** of it
- [ ] This behavior is consistent across all three field types: **Date**, **Date/Time**, and **Time**
- [ ] The selected date/time value is correctly applied to the field when the picker is closed

### Priority

Medium

### Additional Context

Part of the new Edit Contentlet work (`dotCMS: New Edit Contentlet`). The fields are built on PrimeNG — use its DatePicker defaults/configuration for both the width and the "stay open until click-outside" behavior. Keeping the panel open on date selection is required so users editing Date/Time and Time fields can complete the time portion before dismissing.

freddyDOTCMS:
**### QA Failled*  

QA Test Results — #36156
  
  Test plan used: [36156-36365-test-plan.md](https://github.com/user-attachments/files/29811111/36156-36365-test-plan.md)

 -  ❌ TC-008 — Fixed-width picker is fully visible inside the dialog container — FAILED
  
  The date picker panel is partially obscured when opened inside the edit dialog or in full screen. It appears to be clipped by the dialog's overflow boundary, making it difficult to interact with.

[...]

PR #37555, opened 2026-09-15
feat(edit-content): fill the column, clear on the field, and rebuild the picker footer (#37465)

### LINKED ISSUE #37465: Improve Date, Time, and Date-and-Time field width, clearing, and picker footer in the new Edit Contentlet

adrianjm-dotCMS:
## PR 2 of 2 — implementation

Fixes #37465

**Base is the spec branch, not `main`.** PR 1 (#37542) carries `spec.md` and is still awaiting
approval, so this branches off it rather than waiting for it to merge — the
[Spec-Kit flow](https://github.com/dotCMS/core/blob/main/docs/core/SPEC_KIT_QUICK_START.md) gates on
approval, not on merge. Rebase onto `main` once PR 1 lands.

Screenshot 2026-09-15 at 6 28 38 PM
Screenshot 2026-09-15 at 6 28 14 PM
Screenshot 2026-09-15 at 6 28 04 PM
Screenshot 2026-09-15 at 6 28 00 PM
Screenshot 2026-09-15 at 6 27 52 PM

## What changed

Date, Time and Date-and-time were the only fields in the new Edit Contentlet that behaved unlike
their neighbours — narrower than their column, impossible to empty once set, and carrying PrimeNG's
stock picker footer with a redundant *Clear* and a *Today* that read the browser's clock.

| | |
|---|---|
| **Width** | Block-level flex at full width; the input grows, the trigger stays flush right, and the focus ring and invalid border still enclose both as one unit |
| **Clearing** | `showClear` is unconditional, so every type can be emptied — not just the expire-date field |
| **Picker footer** | Rebuilt through `#buttonbar`: timezone left (on the two types carrying a time), a single secondary-outlined `Today` / `Now` right, PrimeNG's `Clear` gone |
| **Today / Now** | Resolves from the **server** clock, and leaves the picker open on all three types — the same as selecting a day or time |
| **Hint** | The timezone line under the input is gone; the hint returns to the field footer, where every other field type puts it |

8 files: 6 modified, 1 new spec, plus three keys in `Language.properties`.

## Two things worth a reviewer's attention

**1. PrimeNG's supplied `todayCallback` is deliberately unused.** It opens with `const date = new Date()`
— the *browser's* clock, which is the defect FR-013 exists to correct. Wiring the footer button to it
would look right, pass a naive test, and reintroduce the bug. The button goes through
`getCurrentServerTime` and the existing `onCalendarChange` conversion instead, so there is still only
one conversion path.

Verified in the browser against a running instance: server on UTC, browser on UTC-4 — `Today` set
**15:57**, not 11:57.

**2. PrimeNG's default clear icon is a bare `<svg>` with a click handler** — unfocusable, no
accessible name, unreachable by keyboard. A real `<button>` is projected through `#clearicon`. The
accessibility tree now announces all three as `button "Clear"`.

## A defect the test suite could not have caught

On reopening saved content, **no field showed a clear control** until the author happened to focus it
— a value with no way to clear it. PrimeNG gates that control on a DOM read of
`inputfieldViewChild.nativeElement.value`, which `updateInputfield()` writes without `markForCheck()`;
the DatePicker is OnPush, so on the load path the condition is evaluated before the value lands and
nothing re-evaluates it. An effect now schedules the pass in a microtask.

It is worth being explicit about why the suite was green over it: the `settle()` test helper had been
calling the DatePicker's own change detector — doing in the test what the component failed to do in
production. It now runs only the host's detection plus a microtask turn, the same turn the browser
gives it, so these tests can fail for this cause again.

Found by manual verification in Chrome against a running dotCMS.

## Testing

`pnpm nx test edit-content` — **2390 passing**, 117 files. `pnpm nx lint edit-content` — 0 errors
(3 pre-existing warnings in untouched files).

Written test-first throughout, with developer approval and confirmed-Red gates per Constitution
Principle V. Two exceptions were declared and signed off rather than left silent:

- **No automated layout test** — jsdom computes no layout, so a unit test could only assert that a CSS declaration exists, which restates the implementation and would pass even if the control rendered wrongly. Covered by manual verification instead.
- **No end-to-end cross-timezone test** — no e2e coverage exists for date fields, and a faithful test needs a server far enough from the runner to cross midnight. The unit test guards the same logic with a faked clock and an offset timezone.

Manually verified in Chrome against a running instance, on throwaway content types since deleted:
field widths identical to a sibling Text field in both single- and two-column layouts (721px / 543px),
input-to-trigger gap 0px, one focus ring over the whole unit, error border over input and trigger,
all three footers correct, and a full save/reopen round-trip confirming the stored shapes are
unchanged — Date-only at UTC midnight, Time-only against a consistent date base, Date-and-time as a
UTC instant. That manual pass predates the last commit, which keeps the picker open after *Today* on
Date-only fields; that change is covered by unit tests.

This PR fixes: #37465


core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.scss
```diff
@@ -51,3 +58,23 @@
...
+    background: none;
+    color: inherit;
+    cursor: pointer;
+
+    &:focus-visible {
+        outline: var(--p-focus-ring-width) solid var(--p-focus-ring-color);
+        outline-offset: var(--p-focus-ring-offset);
+        border-radius: var(--p-border-radius-sm);
+    }
+}
```
nicobytes: This clear control is our own `<button>` in `#clearicon`, not a PrimeNG internal node. The chrome reset (flex alignment, padding, border, background, cursor) and the `:focus-visible` ring are all things Tailwind already covers, and utilities are global so they still apply after the template is projected.

Please move these classes onto the button in `calendar-field.component.html` and drop this `::ng-deep` BEM block. Something along the lines of:

```html
class="inline-flex items-center justify-center cursor-pointer border-0 bg-transparent p-0 text-inherit focus-visible:rounded-sm focus-visible:outline focus-visible:outline-[length:var(--p-focus-ring-width)] focus-visible:outline-offset-(--p-focus-ring-offset) focus-visible:outline-[var(--p-focus-ring-color)]"
```

Keep the `p-datepicker` rules above — those are a real PrimeNG override and belong in SCSS. Only keep an SCSS leftover here if PrimeNG still wins on specificity after the Tailwind move.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/dot-edit-content-calendar-field.component.ts
```diff
@@ -1,4 +1,4 @@
-import { ChangeDetectionStrategy, Component, computed, inject, input } from '@angular/core';
+import { ChangeDetectionStrategy, Component, inject, input } from '@angular/core';
```
nicobytes: `ChangeDetectionStrategy` is no longer needed here. On Angular 22, `OnPush` is the framework default (`docs/frontend/ANGULAR_STANDARDS.md`), so please drop `changeDetection: ChangeDetectionStrategy.OnPush` from the `@Component` decorator and this import.

Same nit on `components/calendar-field/calendar-field.component.ts` — that decorator still sets `OnPush` explicitly too.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.html
```diff
@@ -20,8 +20,54 @@
...
+                </span>
+            }
+            <p-button
+                class="ml-auto"
+                size="small"
+                severity="secondary"
+                [outlined]="true"
+                data-testid="calendar-field-today-button"
+                [label]="$footerActionLabel() | dm"
+                (onClick)="setCurrentServerDateTime()" />
```
nicobytes, line 56: **Keyboard handling regresses inside the overlay.** Every control PrimeNG renders in the panel is wired with `(keydown)="onContainerButtonKeydown($event)"` — including the two stock footer buttons this template replaces. The panel root itself has no keydown handler, only `(click)="onOverlayClick($event)"`, so that per-control binding is the *only* place overlay keyboard behaviour lives.

`onContainerButtonKeydown` implements two things (primeng 21.1.3, `primeng-datepicker.mjs:1771`):
- **Escape** → `inputfieldViewChild.nativeElement.focus()` + `overlayVisible = false`
- **Tab** → `trapFocus(event)`, and `focusTrap` defaults to `true`

So with focus on the new `Today`/`Now` button: Escape no longer closes the picker or returns focus to the input, and Tab walks out of the `body`-appended overlay into whatever follows it in the document. That is a worse keyboard story than the stock footer this PR replaces, in a PR whose stated goal includes accessibility.

One-line fix — name the picker and forward the event:

```html
<p-datepicker #picker ... >
  ...
  <p-button
      class="ml-auto"
      size="small"
      severity="secondary"
      [outlined]="true"
      data-testid="calendar-field-today-button"
      [label]="$footerActionLabel() | dm"
      (keydown)="picker.onContainerButtonKeydown($event)"
      (onClick)="setCurrentServerDateTime()" />
```

The suite currently asserts the footer action is a focusable `<button>` with a name, which is the easy half. Please add the behavioural one: open the picker, focus the footer action, press Escape, assert `overlayVisible === false` and that focus is back on the input. That is the assertion that would have caught this.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.html
```diff
@@ -20,8 +20,54 @@
     [showIcon]="true"
     [icon]="fieldTypeConfig.icon"
     [showButtonBar]="true"
-    [showClear]="$isExpireDateField()"
+    [showClear]="true"
```
nicobytes, line 23: **Dead binding, three lines up** (line 15, outside the diff so I can't anchor there): `(onClearClick)="onCalendarChange(null)"`.

`onClearClick` is emitted from exactly one place — `onClearButtonClick()` (`primeng-datepicker.mjs:3125`) — which is reachable only from the stock footer Clear button this PR deletes, or from the `clearCallback` handed to the `#buttonbar` context, which you deliberately don't use. The on-field X goes through `clear()`, which emits `onClear` only (`:1740`).

So the binding can no longer fire. Please drop it; `(onClear)` on the next line covers the real path. Leaving it there implies a second clear route exists and sends the next reader looking for it.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.ts
```diff
@@ -119,6 +122,17 @@ export class DotCalendarFieldComponent extends BaseControlValueAccessor<number |
...
+        // PrimeNG decides whether to render the clear control by reading the input element's DOM
+        // value, and `updateInputfield()` writes that value without marking the component dirty.
+        // The DatePicker is OnPush, so on the load path — a value arriving from a saved
+        // contentlet — the condition is evaluated before the value lands and nothing ever
+        // re-evaluates it: the field shows a value the author has no way to clear until they
+        // happen to focus it. The microtask lets PrimeNG's writeValue finish first.
+        effect(() => {
+            this.$value();
+            queueMicrotask(() => this.$picker()?.cd.detectChanges());
+        });
```
nicobytes: **The stated cause isn't what PrimeNG 21.1.3 does, and that changes the fix.**

The comment says `updateInputfield()` "writes that value without marking the component dirty". It does write the DOM value without marking — but its only caller on the load path does mark:

```js
// primeng-datepicker.mjs:3248
writeControlValue(value) {
    this.value = value;
    ...
    this.updateInputfield();
    this.updateUI();
    this.cd.markForCheck();   // <- :3262
}
```

The bug you found is real; the mechanism is a **view-query race on the first render**, not a missing `markForCheck`. The clear-icon condition is `showClear && !$disabled() && inputfieldViewChild?.nativeElement?.value` (`:3322`), and `inputfieldViewChild` is a non-static `ViewChild` — still unresolved during the first update pass. So on that pass the condition is evaluated against `undefined`, and `markForCheck()` on a view that has already been refreshed in this cycle only takes effect on the *next* cycle, which on the load path never comes.

Two consequences worth fixing:

1. **The note misdirects the next maintainer.** On the next PrimeNG bump someone will check "did they add `markForCheck`?", find it already there, and conclude the workaround is obsolete — while the actual condition (`inputfieldViewChild` in the `*ngIf`) is unchanged. Please restate the cause and cite the line, so the upgrade check looks at the right thing.

2. **If it is only the first render, the effect is broader than the problem.** As written this re-runs on *every* `$value()` change — every keystroke schedules a forced `detectChanges()` over the whole picker subtree, including the overlay when open. `afterNextRender` is the hook that matches a first-render race, runs once, and is torn down with the component:

```ts
afterNextRender(() => this.$picker()?.cd.detectChanges());
```

If later writes turn out to need it too, say so in the comment and keep the effect — but then it's a different claim than the one the comment makes now, and it should be the one that's written down.

Separately: `$picker().cd` reaches into `BaseComponent`'s injected `ChangeDetectorRef`. It is a public field, so this compiles, but it isn't part of the documented DatePicker API and nothing will warn you when it moves. Worth a line acknowledging that, since it's the only thing holding the workaround up.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.ts
```diff
@@ -119,6 +122,17 @@ export class DotCalendarFieldComponent extends BaseControlValueAccessor<number |
...
 
+        // PrimeNG decides whether to render the clear control by reading the input element's DOM
+        // value, and `updateInputfield()` writes that value without marking the component dirty.
+        // The DatePicker is OnPush, so on the load path — a value arriving from a saved
+        // contentlet — the condition is evaluated before the value lands and nothing ever
+        // re-evaluates it: the field shows a value the author has no way to clear until they
+        // happen to focus it. The microtask lets PrimeNG's writeValue finish first.
+        effect(() => {
+            this.$value();
+            queueMicrotask(() => this.$picker()?.cd.detectChanges());
```
nicobytes: **No destroy guard on the deferred `detectChanges()`.**

The microtask runs after the current task, so the view can be gone by the time it fires. That isn't hypothetical here — in `dot-edit-content-form.component.html` the field sits under `@if (isPreservedField(...) ? $shouldRenderPreservedFields() : $shouldRenderFields())`, inside `@if (form)`. A single store update that both patches form values and tears the form down (version restore, historical-version preview, language switch, loading a different contentlet) runs the effect and destroys the view in the same task; the microtask then calls `detectChanges()` on a destroyed `ChangeDetectorRef`, which asserts in dev builds.

`afterNextRender` (see the other comment) solves this for free — its callbacks are unregistered on destroy. If you keep the effect for another reason, guard it explicitly:

```ts
private readonly destroyRef = inject(DestroyRef);
private isDestroyed = false;
// in the constructor
this.destroyRef.onDestroy(() => (this.isDestroyed = true));
// in the microtask
if (!this.isDestroyed) this.$picker()?.cd.detectChanges();
```

A test is cheap: render with a value, destroy the fixture in the same turn, flush microtasks, assert nothing throws.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.scss
```diff
@@ -2,14 +2,21 @@
...
+
     transition:
         border-color var(--p-inputtext-transition-duration),
         box-shadow var(--p-inputtext-transition-duration);
 
     .p-datepicker-input {
+        // Grows so the input, not the wrapper, absorbs the column's extra width.
+        flex: 1 1 auto;
+        min-width: 0;
         border-right: 0;
```
nicobytes: **PrimeNG already ships this, and a sibling field in this feature already uses it.**

`display: flex` + `width: 100%` + `flex: 1 1 auto` on `.p-datepicker-input` is a re-implementation of `[fluid]`. From `@primeuix/styles/datepicker`:

```css
.p-datepicker        { display: inline-flex; max-width: 100%; }
.p-datepicker-fluid  { display: flex; }                 /* <- what [fluid] adds */
.p-datepicker-input  { flex: 1 1 auto; width: 1%; }     /* <- already there */
```

`[fluid]="true"` sets `hasFluid`, which puts `p-datepicker-fluid` on the root (`primeng-datepicker.mjs:42`) → `display: flex`, and a block-level flex container fills its column with no explicit `width`. It also forwards `[fluid]` to the inner `pInputText` (`:3317`), so the input stretches without your `flex` rule.

`tag-field.component.html:16`, in this same `fields/` tree, already does exactly this.

So:
```html
<p-datepicker [fluid]="true" ... >
```
and the first three declarations plus the `flex: 1 1 auto` come out. Keep `min-width: 0` if you measured that you need it (PrimeNG's `width: 1%` usually covers the same overflow case) — worth confirming against a long value in a narrow column either way.

The rest of this stylesheet — moving the focus ring to the wrapper, the invalid border over both halves — is doing real work that PrimeNG doesn't offer, and the comments on it are good. This is only about the four lines that duplicate a supported input.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/dot-edit-content-calendar-field.component.spec.ts
```diff
@@ -260,35 +295,42 @@ describe('DotEditContentCalendarFieldComponent', () => {
...
                         contentlet: createFakeContentlet({
                             [DATE_FIELD_MOCK.variable]: null
                         })
                     }
                 }
             );
             spectator.detectChanges();
+            // The DatePicker is OnPush and writes the input's DOM value without marking itself
+            // dirty, so the clear control's condition is only re-evaluated by its own detector.
+            spectator.query(DatePicker).cd.detectChanges();
```
nicobytes: **This is the anti-pattern the new `settle()` was written to remove.**

`calendar-field.component.spec.ts` documents, at length and correctly, why calling the DatePicker's own detector in a test is what kept the load-path bug green: "an earlier version of this helper called `picker.cd.detectChanges()`, which turned these tests green while the real app was broken". This test does exactly that, and the comment above it restates the pre-fix behaviour as if it were still current.

Concretely: delete the `effect` in `calendar-field.component.ts` and this test still passes, because the test supplies the pass the component is supposed to schedule. It cannot fail for the cause it looks like it's guarding.

Use the same microtask shape as the other spec:

```ts
spectator.detectChanges();
await Promise.resolve();
spectator.detectChanges();
```

(`it` becomes `async`.) Then drop the comment — with the production fix in place the DatePicker's condition *is* re-evaluated without the test's help, which is the whole point.

nicobytes (commented):
## Review — five axes (correctness, readability, architecture, security, performance)

The reasoning in the description is unusually good, and two things are genuinely well done: routing `Today`/`Now` through `getCurrentServerTime` instead of PrimeNG's `todayCallback` is the right call and is pinned by a test that fakes the clock **and** sits on an offset zone, so it can fail again for the original cause; and rewriting `settle()` so it no longer runs the DatePicker's own detector is the kind of test fix that costs you a green build today and saves you a production bug later. I checked the disabled case against PrimeNG rather than assuming: `showClear && !$disabled() && inputfieldViewChild?.nativeElement?.value` (`primeng/datepicker` 21.1.3), so making `showClear` unconditional cannot expose a clear control on a read-only field — and there is a test for it. No security or performance concerns in scope; no untrusted input, no new network or data access.

Three things I'd want addressed before merge, all verified against the installed PrimeNG 21.1.3 source rather than inferred:

1. **Keyboard handling is lost in the new picker footer.** The custom `<p-button>` does not wire `onContainerButtonKeydown`, which is what implements Escape and the Tab focus trap inside the overlay. Details inline.
2. **The comment explaining the OnPush workaround states a cause that is not true in 21.1.3.** `writeControlValue()` *does* call `cd.markForCheck()`. The symptom is real; the mechanism is a view-query race on the first render. That distinction changes both the right hook and how narrow the fix can be. Details inline.
3. **The forced `detectChanges()` runs in a microtask with no destroy guard**, and field components live under `@if (...)` in the form template. Details inline.

Plus one architecture cleanup worth taking: PrimeNG already ships the full-width behaviour as `[fluid]`, and the sibling tag field in this same feature already uses it — the `::ng-deep` block re-implements it.

The rest are small: one dead output binding, one test that re-introduces the anti-pattern the new `settle()` was written to remove, and three stale comments.

### Not blocking, noted for the record
- `dot-edit-content-custom-field.component.html:33` still carries `data-testid="calendar-field-hint"`, a copy-paste leftover from this field. Not this PR's job, but the id is free now.
- Dropping the `console.warn` on non-number values is correct per the repo's Critical Rules, and `$value` is typed `number | null`, so nothing is being silently swallowed.
- The two declared test exceptions (no jsdom layout test, no cross-timezone e2e) are the right calls and are argued rather than hidden.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.spec.ts
```diff
@@ -0,0 +1,476 @@
...
+    const ALL_TYPES = [FIELD_TYPES.DATE, FIELD_TYPES.TIME, FIELD_TYPES.DATE_AND_TIME] as const;
+
+    /** A value the picker will render into the input, so the clear control becomes eligible. */
+    const A_VALUE = new Date(2026, 3, 9, 9, 33).getTime();
+
+    /**
+     * The clear control as the user meets it: something focusable and activatable, not the
+     * bare <svg> PrimeNG renders by default. Queried by role rather than by class so the
+     * test fails if the control is present but not a real button.
+     */
```
nicobytes, line 118: Nit — the doc contradicts the code. It says "Queried by role rather than by class so the test fails if the control is present but not a real button", but the query is `[data-testid="calendar-clear-button"]`, which is neither. What actually enforces "real button" is the separate `tagName`/`type`/`aria-label` test below, and `byTestId` is the repo convention — so the query is right and the sentence is wrong.

Suggest: "Queried by test id, the repo convention; that it is a real focusable button is asserted separately (FR-007a)."

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.spec.ts
```diff
@@ -0,0 +1,476 @@
...
+            const control = spectator.hostComponent.formGroup.get(DATE_FIELD_MOCK.variable);
+
+            expect(control?.touched).toBe(true);
+            expect(control?.dirty).toBe(true);
+        });
+
+        // T037 — FR-014a. A date-only pick is complete, so it closes; the two types carrying a
+        // time stay open so the hour can still be adjusted.
+        // The footer action never closes the picker, matching a day click: hideOnDateTimeSelect
+        // is false for all three types, so selecting a value keeps the overlay open until the
```
nicobytes: Stale first sentence, and it says the opposite of the test. "A date-only pick is complete, so it closes; the two types carrying a time stay open" survives from an earlier behaviour; the next sentence, and the test itself (`it.each([...ALL_TYPES])` … `expect(overlayVisible).toBe(true)`), say the picker stays open on all three. Please delete the first sentence — a comment that contradicts its own assertion is worse than none, and this one describes behaviour a reader might go and "restore".

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.spec.ts
```diff
@@ -0,0 +1,476 @@
...
+        });
+
+        /**
+         * The inner <button> of the p-button, not its host element: clicking the host does not
+         * fire `onClick`, it just bubbles to the document and PrimeNG closes the overlay — which
+         * looks like the handler ran and did the wrong thing.
+         */
+        const queryActionButton = (): HTMLElement | null =>
+            queryInOverlay('[data-testid="calendar-field-today-button"] button') ??
+            queryInOverlay('[data-testid="calendar-field-today-button"]');
```
nicobytes: Nit — the `??` fallback softens the assertions it feeds. The docblock above explains precisely why the inner `<button>` is the element that matters (clicking the `p-button` host doesn't fire `onClick`), and then the fallback hands back the host anyway when the inner button isn't found. If `p-button` ever stops rendering a native button, the label test and the "exactly one footer action" test keep passing against the host element, and only the `tagName` test fails — so you'd get one red instead of four and a misleading diagnosis.

Drop the fallback and let it be `null`; the failure then points straight at the cause. If the host-vs-inner-button indirection is the real irritant, `[buttonProps]="{ 'data-testid': 'calendar-field-today-button' }"` puts the id on the rendered button and the query becomes direct.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.ts
```diff
@@ -199,6 +246,18 @@ export class DotCalendarFieldComponent extends BaseControlValueAccessor<number |
...
 
+    /**
+     * Fills the field with the current moment, as the SERVER sees it.
+     *
+     * Deliberately does not use the `todayCallback` PrimeNG hands to the button-bar template:
+     * that callback reads `new Date()`, the browser's clock, which on a server in another
+     * timezone resolves to the wrong calendar day. Routing through `onCalendarChange` also
+     * means the per-type UTC conversion stays in one place rather than being duplicated here.
+     */
+    setCurrentServerDateTime(): void {
```
oidacra, line 349: The footer button is rendered unconditionally for all three types, but `$showFooterTimezone()` deliberately hides the timezone label when the timezone has not resolved. That leaves a window where the button is clickable and the label is absent.

`getCurrentServerTime(null)` takes its `!systemTimezone?.id` branch and returns the browser's UTC components reinterpreted as local. On a non-UTC server that commits the wrong instant, with no error, no log, and no visible timezone to hint anything is off. `$systemTimezone` resolves from `#globalStore.systemConfig()?.systemTimezone ?? null`, so this is reachable both as a load race and permanently if that config is ever missing.

Worth flagging that this utility previously only fed `$defaultDate` for calendar navigation; this is the first call site that persists its result, which turns a cosmetic fallback into a stored-value bug.

Should the button be disabled until `$systemTimezone()` is available, mirroring what `$showFooterTimezone()` already does for the label beside it? Or is the null-timezone case considered unreachable in production?

(Added by this PR. `getCurrentServerTime` itself is pre-existing and unchanged.)

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.scss
```diff
@@ -2,14 +2,21 @@
 // the ring encloses both the input and the calendar icon button as one unit.
 
 :host ::ng-deep p-datepicker {
-    // Use PrimeNG's default datepicker sizing
-    display: inline-flex;
+    // Block-level flex so the control fills the column it sits in, matching every other field
+    // type. The input is the part that grows; the trigger keeps its intrinsic width and stays
+    // flush against the right edge.
+    display: flex;
```
oidacra: This reverses a decision that shipped deliberately, and I think it is worth confirming before merge.

`inline-flex` was not an oversight. Issue #37465 (this one) names it as the cause of the width defect, but issue #36156 (Date/Time/Time field pickers: fixed width and keep-open-until-click-outside, closed) reads: "The picker currently stretches full width. It should have a predetermined, fixed width that fits the control instead of spanning the whole form." It was fixed by PR #36365 (commit `77beaa3cae`), which removed `w-full`, introduced this `inline-flex`, and shipped a test literally named "should render the %s picker at PrimeNG default width, not full width" — the test this PR renames.

So #37465 and #36156 ask for opposite things, and #37465's Cause section does not acknowledge the earlier decision. This PR faithfully implements the issue it was given; the contradiction is upstream of the code.

Was #36156's original complaint re-examined when #37465 was written? Whatever visual problem prompted it (crowded or misaligned pickers in multi-column layouts) would come back now that the control stretches again — and the width ACs here mention multi-column and narrow viewports, which no automated test covers.

(Changed by this PR.)

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/dot-edit-content-calendar-field.component.html
```diff
@@ -20,36 +14,40 @@
...
+                        apps/dotcms-ui/src/style.css already declare, because those rules only
+                        apply under a `.form` ancestor and the new Edit Content editor has none —
+                        so the classes alone render as unstyled text. Adopting `.form` here is
+                        #37460; making these rules stand on their own is an AC of #37464. Until
+                        one of those lands this field states its own colours. Drop the utilities,
+                        keeping the semantic classes, once either does.
+                    -->
                     @if (fieldHasError && isRequired) {
-                        <small class="p-field-error">
+                        <small class="p-field-error text-sm text-red-500">
```
oidacra, line 40: Two things converge on this block.

First, scope. Issue #37465 lists "hint and required-error presentation" as out of scope, owned by #37464. Removing the under-input timezone line only required dropping the `hasTimezone` gate from the hint condition. Stacking the hint alongside the error (it previously suppressed the hint entirely), fixing their order, and adding the colour utilities all go past that. It is covered by a test, so it is not risky — but it is new user-visible behaviour in an area the issue explicitly fenced off.

Second, collision. PR #37597 (branch `37460-37464-edit-content-form-hint-error`, opened the same day, still open) modifies this same file's hint/error markup to implement #37460 and #37464 globally. Landing both produces a conflict on exactly the lines rewritten here, and the "temporary until #37464 lands" utilities this comment describes become dead weight the moment that PR merges.

Has this been coordinated with whoever owns #37597 — and would it be cleaner to let that PR own the hint/error presentation and keep this one to dropping the `hasTimezone` gate?

(Added by this PR.)

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/dot-edit-content-calendar-field.component.spec.ts
oidacra: This calls the DatePicker's own change detector directly — which is the pattern the new spec's `settle()` doc comment condemns as having "turned these tests green while the real app was broken."

The production fix for that defect is the `queueMicrotask(() => this.$picker()?.cd.detectChanges())` effect. By supplying the pass synchronously here, this test bypasses that mechanism: delete the effect and the test still passes. So it proves the clear control can render given some change-detection pass, not that the app actually supplies one — which is the regression FR-007b is meant to guard.

Would awaiting a microtask turn (the `settle()` pattern from the sibling spec) work here instead, so the test exercises the real path?

(Added by this PR — the whole test body was rewritten in this diff.)

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.spec.ts
```diff
@@ -0,0 +1,476 @@
...
+
+    describe('Picker footer — timezone (US2)', () => {
+        // T018 — FR-008
+        it('should show the timezone in the picker footer for a Date-and-Time field', async () => {
+            spectator = buildHost(FIELD_TYPES.DATE_AND_TIME);
+            openPicker();
+
+            const timezone = queryInOverlay('[data-testid="calendar-field-timezone"]');
+
+            expect(timezone).not.toBeNull();
```
oidacra, line 417: Every test in this suite injects `MOCK_TIMEZONE`, so the timezone-unavailable branch is never exercised.

One AC asks that the timezone "renders nothing (no empty footer slot, no layout shift) when the timezone is unavailable." `$showFooterTimezone()` gates on `!!this.$systemTimezone()?.label` and the `ml-auto` on the button keeps the row correct, so the code looks right — but nothing pins it.

A single case with `utcTimezone: null` on a DATE_AND_TIME field would close that AC, and it is also the case that would surface the `Today`/`Now` behaviour I raised on `calendar-field.component.ts:257` — the button stays clickable in exactly this state and commits a UTC-assumed value. Worth adding for both reasons?

(Added by this PR.)

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.html
```diff
@@ -20,8 +20,54 @@
...
+        three. The click bubbles to the wrapping span PrimeNG puts the handler on, so Enter
+        and Space work through the native button behaviour without extra key handling.
+    -->
+    <ng-template #clearicon>
+        <button
+            type="button"
+            class="calendar-field__clear"
+            data-testid="calendar-clear-button"
+            [attr.aria-label]="'edit.content.form.field.calendar.clear' | dm">
+            <i class="pi pi-times" aria-hidden="true"></i>
```
oidacra: ANGULAR_STANDARDS asks for Material Symbols in markup you author, and treats PrimeIcons as legacy-only. It does carve out "icons rendered internally by PrimeNG components" — but that covers glyphs PrimeNG draws in its own template, and here the `<button>` and this `<i>` are written by this component into a projection slot; PrimeNG only decides where the node lands, not what is in it.

`edit-content` already renders this exact action as a Material Symbol elsewhere — `dot-form-file-editor.component.html` and `dot-image-editor-history-panel.component.html` both use `<i class="material-symbols-outlined">close</i>`.

Would `<span class="material-symbols-outlined" aria-hidden="true">close</span>` be the right call here, to match those?

(Added by this PR.)

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.spec.ts
```diff
@@ -0,0 +1,476 @@
...
+        // T035 — FR-012
+        it('should set the server time on a Time-only field', async () => {
+            spectator = buildHost(FIELD_TYPES.TIME);
+            openPicker();
+
+            spectator.click(queryActionButton() as HTMLElement);
+            await settle();
+
+            expect(spectator.component.internalFormControl.value?.getHours()).toBe(SERVER_HOUR);
+            expect(controlValue()).not.toBeNull();
```
oidacra: This is the weakest of the three server-clock assertions. The Date test asserts an exact `Date.UTC(...)`, the Date-and-Time test asserts exact `getDate()`/`getHours()` plus an exact `controlValue()` — but the committed value here is only checked for presence.

The AC is "On a Time-only field, `Now` sets the current server time", and a regression that stored the right hour against the wrong date base would still pass this.

The TIME branch of `onCalendarChange` mixes `new Date()` for the calendar day with the server-derived time, so I can see why an exact literal is awkward here. Is that the reason, or is this just left over from iteration? Even asserting the value against a base derived the same way the production code derives it would be stronger than a null check.

(Added by this PR.)

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.html
```diff
@@ -20,8 +20,54 @@
...
+    <ng-template #buttonbar>
+        <!--
+            Tailwind utilities, not component SCSS: the overlay is appended to body, so a
+            ':host ::ng-deep' rule never matches it — the selector requires the element to be a
+            descendant of the host. Tailwind's classes are global, so they apply wherever
+            PrimeNG moves these nodes.
+        -->
+        <div class="flex w-full items-center justify-between gap-2">
+            @if ($showFooterTimezone()) {
+                <span class="text-sm text-muted-color" data-testid="calendar-field-timezone">
```
oidacra: This uses `text-muted-color`, while the hint in the same feature — `dot-edit-content-calendar-field.component.html:46` — uses `text-gray-500`. Before this PR the timezone was that same hint element, styled `text-gray-500`; it picked up the new scale when it moved into the footer.

STYLING_STANDARDS asks to "use one gray scale [...] `text-gray-*` is the app's majority convention — don't mix the two in one screen." `text-muted-color` is used elsewhere in the app so it is not wrong on its own, but having both scales on secondary text two elements apart in the same field seems unintended.

Was the switch deliberate (e.g. because the footer sits on a PrimeNG overlay surface), or should this stay `text-gray-500`?

(Added by this PR.)

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.html
```diff
@@ -20,8 +20,54 @@
...
+                    {{ $systemTimezone().label }}
```
oidacra: `$systemTimezone` is typed `DotSystemTimezone | null`, and `.label` is read here without `?.`. The enclosing `@if ($showFooterTimezone())` does check `!!this.$systemTimezone()?.label` internally, but that is a separate computed the template type-checker cannot correlate with this expression.

It compiles today because `edit-content`'s tsconfig sets `strict: false` (so `strictNullChecks` is off), even though `strictTemplates` is on. The repo's diff-scoped strict typecheck gate runs with null checks enabled, so this is the kind of line that trips it.

Would `{{ $systemTimezone()?.label }}` be worth it here, rather than relying on a separately-named computed to establish non-null?

(Added by this PR.)

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.spec.ts
```diff
@@ -0,0 +1,476 @@
...
+        ])('should label the footer action from the %s key', (fieldType, expectedKey) => {
+            spectator = buildHost(fieldType as FIELD_TYPES);
+            openPicker();
+
+            expect(queryActionButton()?.textContent).toContain(expectedKey);
+        });
+
+        // T038 — FR-015. Clearing is the on-field X now; a second way to do it in the footer
+        // is the redundancy this feature removes.
+        it.each([...ALL_TYPES])(
```
oidacra, line 495: This test is declared `async` but never awaits anything — `openPicker()` is synchronous and there is no `await settle()`. Same for the label test just above it.

Was an `await settle()` intended here and dropped? If the synchronous double-`detectChanges()` inside `openPicker()` is genuinely sufficient for overlay content, the `async` is vestigial and could go; if it is not, these two tests are relying on timing rather than on the settle discipline the rest of the file uses.

Related: `openPicker()` calls `picker.cd.detectChanges()` directly, which is the pattern `settle()` was rewritten to avoid. I can see the argument that they are different situations — `showOverlay()` is called on the instance rather than through a real click, so nothing marks the OnPush picker dirty — but the file does not say so, and a future reader has no way to tell the two apart. Worth a line in `openPicker()`'s doc comment explaining why it is exempt?

(Added by this PR.)

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.html
adrianjm-dotCMS: Fixed. Confirmed the mechanism in primeng 21.1.3 before touching it: `onContainerButtonKeydown` handles Tab via `trapFocus` (:1776) and Escape via `inputfieldViewChild.focus()` + `overlayVisible = false` (:1793), and the panel root really does bind only `(click)`.

Forwarded through `onFooterButtonKeydown`, using the `viewChild` the component already held rather than a new template ref.

Added the behavioural test you asked for — open, focus the footer action, Escape, assert `overlayVisible === false` and focus back on the input. Verified it earns its keep: removing the binding turns it red.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.ts
adrianjm-dotCMS: You are right, and this was the most useful comment in the review — the note would have sent the next maintainer to the wrong place.

Verified: `writeControlValue` does call `cd.markForCheck()` at :3262, and the view query really is non-static (`viewQueries: [{ propertyName: "inputfieldViewChild", first: true, predicate: ["inputfield"] }]`, no `static: true`). The comment now states the view-query race, cites :3322 and :3262, and says explicitly that a missing `markForCheck` is *not* the bug so the next PrimeNG bump checks the right thing.

Also took your second point: switched to `afterNextRender`. It matches a first-render race, runs once instead of on every `$value()`, and — per your other comment — is torn down with the component. Disabling it reds eight tests, so the guard holds.

Added the caveat about `$picker().cd` reaching into `BaseComponent`'s injected `ChangeDetectorRef`: public, undocumented, nothing warns if it moves.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.ts
adrianjm-dotCMS: Fixed by the `afterNextRender` switch — its callbacks are unregistered on destroy, so there is no deferred call left to fire against a torn-down view. Thanks for tracing the concrete path through `@if (form)`; that is what made it clear the risk was not hypothetical.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.scss
adrianjm-dotCMS: Applied — `[fluid]="true"` and the four declarations are gone (seven lines with the comment).

Verified in the browser rather than trusting the CSS: `p-datepicker-fluid` lands on the root and `p-inputtext-fluid` on the input, and across a two-column layout all four fields (Text, Date, Time, Date-and-time) measure exactly 520px with the input-to-trigger gap at 0. Also dropped `min-width: 0` — PrimeNG's `width: 1%` covers it, confirmed against a full value in a narrow column.

The focus-ring and invalid-border rules stayed; re-checked both in the browser after the removal (`box-shadow` on the wrapper at `:focus-within`, invalid border on input *and* trigger).

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.html
adrianjm-dotCMS: Confirmed and removed. `onClearClick` is emitted only from `onClearButtonClick` (:3129), reachable from the stock Clear button this PR deletes or from the buttonbar `clearCallback` we deliberately do not use; `clear()` emits `onClear` only (:1746).

Its three tests went with it — they drove the event synthetically via `triggerEventHandler`, so they asserted a path production cannot take. Left a comment where they were explaining why, with the line numbers.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.scss
adrianjm-dotCMS: Done — the chrome moved to Tailwind utilities on the button and the `::ng-deep` BEM block is gone. You are right that it was never a PrimeNG internal, so it had no business in that stylesheet. The `p-datepicker` rules above stayed.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/dot-edit-content-calendar-field.component.ts
adrianjm-dotCMS: Removed from both components, along with the import. Thanks for catching the second one.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/dot-edit-content-calendar-field.component.spec.ts
adrianjm-dotCMS: Fixed — it now uses the same `detectChanges()` / `await Promise.resolve()` / `detectChanges()` shape as the other spec, and the comment is gone.

You are right that it could not fail for the cause it looked like it was guarding. That is the same mistake this file's `settle()` docblock was written about, so having it survive two files away was not a great look.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.spec.ts
adrianjm-dotCMS: Reworded to your suggestion — the query is by test id, and that it is a real focusable button is asserted separately.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.spec.ts
adrianjm-dotCMS: Deleted. It described behaviour from an earlier iteration and contradicted the assertion right below it.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.spec.ts
adrianjm-dotCMS: Dropped the fallback. Agreed on the reasoning — one red pointing at the cause beats four reds pointing at symptoms.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.ts
adrianjm-dotCMS: Good catch, and the detail that decided it was yours: this is the first call site that *persists* what `getCurrentServerTime` returns. While it only fed `$defaultDate` the fallback was cosmetic; committing it is not.

The button is now disabled until `$systemTimezone()` resolves, mirroring what `$showFooterTimezone()` does for the label. A server explicitly on UTC stays enabled — there the branch is correct, not a fallback — so the guard is `!!$systemTimezone()?.id` rather than a check on the label.

Covered by three tests (one per field type) plus one asserting UTC stays enabled. Removing the `[disabled]` reds them.

On the obvious follow-up — a disabled button with no explanation: left as is, deliberately. `disabled` suppresses PrimeNG tooltips, so explaining it would need a wrapper and a new message key for a state that is either a few milliseconds at startup or a failed `systemConfig`, in which case the editor has larger problems than this button. Not worth the machinery; noting it here so the omission is a decision rather than an oversight.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.spec.ts
adrianjm-dotCMS: Added — three cases with `utcTimezone: null`: no timezone element in the footer (the AC you named), and the footer action disabled on each of the three field types. Plus one for a server explicitly on UTC staying enabled, so the two states do not get conflated.

You were right that it was also the state that surfaced the `Today`/`Now` problem; both are fixed together.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.spec.ts
adrianjm-dotCMS: Fixed — it now asserts the stored instant exactly, derived the same way the TIME branch derives it (today's date in the runner's zone, server hour/minute/second). Your read was right: a regression storing the right hour against the wrong date base would have passed the null check.

It was iteration leftovers, not a deliberate call.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.html
adrianjm-dotCMS: Changed to `text-gray-500`. Not deliberate — it picked up the new scale when the label moved into the footer. Agreed it should match the hint two elements away.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.html
adrianjm-dotCMS: Changed to `$systemTimezone()?.label`. You are right that the `@if` on a separately-named computed does not narrow anything for the template type-checker, and that the diff-scoped strict gate is exactly where this would surface.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.html
adrianjm-dotCMS: Changed to `<span class="material-symbols-outlined" aria-hidden="true">close</span>`, matching `dot-form-file-editor` and `dot-image-editor-history-panel`.

Agreed on the reading of the carve-out: PrimeNG decides where the projected node lands, but the `<button>` and its icon are ours.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.spec.ts
adrianjm-dotCMS: Both fixed. The `async` was vestigial on five tests, not two — all overlay-content tests that only call `openPicker()`. Removed.

And you are right that `openPicker()` deserved the line: it now says why it is exempt — `showOverlay()` is called on the instance rather than through a real click, so nothing marks the OnPush picker dirty the way a user interaction would. It stands in for the click, not for a pass the component owes. That is the distinction a future reader had no way to make.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/components/calendar-field/calendar-field.component.scss
adrianjm-dotCMS: Checked this with the issue author before touching it, since you are right that the two issues ask for opposite things.

Confirmed: #36156 was about the **field** — the control itself, not the overlay panel — so #37465 does reverse it rather than correct an over-application. It was a deliberate decision, made knowing that.

On the visual risk you raise: verified in the browser across a two-column layout at several widths. All four fields (Text, Date, Time, Date-and-time) line up at 520px per column, the input and trigger stay joined, no overflow. The width ACs are still not covered by an automated test — jsdom computes no layout and there is no visual-regression harness — so that is manual verification, recorded as a declared Principle V exception in the spec rather than left implicit.

core-web/libs/edit-content/src/lib/fields/dot-edit-content-calendar-field/dot-edit-content-calendar-field.component.html
adrianjm-dotCMS: Coordinated — #37597 is mine, and it is stacked on this branch rather than parallel to it (its base is `issue-37465-calendar-field-width-clear-picker-footer-impl`), so there is no merge conflict.

On the substance, you are right that this markup is temporary: #37597 deletes the wrappers, the colour utilities and the comment, keeping only the stacking semantics. We considered pulling it back to just dropping the `hasTimezone` gate and letting #37597 own the rest. Decided against it, deliberately: the merge order is spec → this PR → #37597, so this one lands alone on main for a while, and during that window the utilities are what keep the required error red — the rest of the editor has that problem today, which is what #37460 exists to fix.

So it is scaffolding with an owner and an expiry, one commit away, not something left behind. The test for the stacking stays in this file either way; #37597 adapts it to the new `$hasError` mechanism.

oidacra (approved):

PR #37559, opened 2026-09-15
feat(ai): OpenAI-compatible inference endpoints at /api/inference/v1 (#37431)

### LINKED ISSUE #37431: Add OpenAI-compatible inference endpoints at /api/inference/v1

fmontes:
OpenAI-wire-format endpoints at `/api/inference/v1`, so any standard client or agent framework can point at dotCMS and use the customer's own provider credentials, resolved per site.

**Includes the hardening work from #37561, now merged into this branch.**

## Architecture

New package `com.dotcms.inference.rest`, sitting *beside* `com.dotcms.ai.rest` rather than inside it.

```
inference/rest/      4 resources + 3 name-bound filters (@InferenceEndpoint)
inference/rest/view/ wire records (Jackson-annotated)
inference/model/     internal representation (deliberately no Jackson)
InferenceAIClient    new client, owns wire-format semantics
       ↓ borrows models from
LangChain4jAIClient  existing cache, fallback chain, AIAppListener flush
AiHostResolver       existing site resolution, widened
ConfigService        existing per-site App secrets
```

Nothing new for configuration, credentials, provider selection or model caching — it all comes from the existing dotAI plumbing.

## Where to look

- **`AiHostResolver`** — widened to public, gains `resolve(request, siteOverride, user)` returning `ResolvedAiContext(User, Host, AppConfig)`. Additive: `resolveHost`/`resolveHostStrict` untouched, so shipped endpoints are unaffected.
- **`LangChain4jAIClient`** — `executeWithFallback` generalised to a typed variant the original delegates to; new `withChatModel`/`withStreamingChatModel`/`withEmbeddingModel`/`withImageModel` accessors. Models are still built and cached here, which is why `AIAppListener`'s cache flush keeps working on key rotation.
- **`InferenceAIClient`** — everything wire-format-specific. Tool specs, tool-call identity, stream assembly, usage.
- **`SseSerializer`** — chunk framing and the `[DONE]` rules.

## Decisions

**A separate client, not an extension of `LangChain4jAIClient`.** `AIProxyClient` registers exactly one client with `currentProvider` hardcoded, and the shipped `toSseChunk` emits a shape the legacy endpoint depends on. Changing either would move behaviour under existing callers. The new client borrows *models* but owns *semantics*.

**A separate cache would have been a security bug.** `AIAppListener` flushes `LangChain4jAIClient`'s caches by host on secret change. A second cache would keep serving revoked keys — so the new client deliberately takes models from the existing one.

**Model is required. No default, no alias.** Every comparable API requires it, and it keeps the accepted set exactly equal to what the site configured, with no reserved word a provider's naming could collide with. Consequence, accepted knowingly: a provider swap is *not* invisible — a caller pinning an old vendor model gets a 404 and must re-read `/models`.

**Site resolution stays standard, fallbacks included.** These callers are servers presenting internal DNS, container names, `localhost` — none of which is a site alias. Refusing them would break the primary deployment. The risk (nobody can tell which site paid) is answered by logging the fallback and naming the serving site in `X-dotCMS-Resolved-Site` on every response, refusals included.

**Bearer tokens only.** The credential is long-lived with its owner's full authority. Session and basic are refused. The check is a static method both the filter and the resource call — a filter alone only exists inside the JAX-RS chain, and every integration test here invokes resource methods directly, so a filter-only control passes CI while proving nothing.

**Message content accepts the array form.** A turn's content may be a string or an array of typed parts, and the client chooses: the OpenAI libraries send a string for plain text and switch to the array as soon as a turn carries an image or more than one part. Binding it to a String failed inside Jackson with a message naming a Java type rather than a refusal a caller could act on. Text parts are now flattened — joined with a newline, so two parts cannot fuse into a word that was in neither — and any other part type is refused by name as a 400. Refused rather than dropped: discarding an image part would send the surrounding text alone and return a confident answer about a picture the model never received. Vision is therefore unsupported, and says so.

**The internal model has no Jackson annotations.** Structural, not stylistic: it stops the chat-completions shape leaking into the core when a second wire format arrives.

## Two defects this found in shipped code

`host_id` / `Host` request parameters were honoured on the unmatched-host fallback path — a caller could append `?host_id=<other site>` and spend that site's credentials. Ignored where the server name matched, live exactly where it did damage. Fixed for this family only; the shipped endpoints still have it (see #37491).

Bearer-only had never been implemented — a session cookie authenticated.

## Notes for running it

A new REST package must be registered **twice**: `DotRestApplication`'s scanned packages, and `swagger-maven-plugin`'s `resourcePackages`. Miss the second and the endpoint works while never appearing in `openapi.yaml`, with CI green.

Running several of these test classes together needs `-Dit.test.forkcount=1` — they share the fixed WireMock port and the module defaults to four forks. Without it you get ten bogus bind failures that look like a broad regression.

## Test-first gates did not hold

The authorization and site-resolution work was built before its tests existed, so those tests could not have gated it. The models, embeddings and image endpoints were written test-first and did fail first, but nobody approved the test sets beforehand and the failing run was not observed by a developer. Eleven mutations were applied afterwards — each deleting one behaviour, building, running the test that guards it, reverting — and all eleven were caught. That establishes the tests detect absence; it does not reconstruct the process. **Worth deciding how much extra scrutiny those areas warrant.**


dotcms-integration/src/test/java/com/dotcms/MainSuite2b.java
wezell, line 1: Don't put these tests in MainSuite2b by default.  Put them in the test suite that runs the fastest. Our CICD is only as fast as our slowest test suite

dotCMS/src/main/java/com/dotcms/ai/client/langchain4j/LangChain4jAIClient.java
```diff
@@ -72,6 +75,21 @@ public class LangChain4jAIClient implements AIClient {
...
+     * Response format asked of an image provider by the {@code /api/inference/v1} family, so the
+     * bytes arrive inline and no separately-addressable artifact is minted upstream.
+     */
+
+    /**
+     * Cache-key discriminator for models the inference family borrows. Those differ from the ones
+     * the legacy endpoints get — they are built asking for the inline image form — so they cannot
+     * share a cache entry with them even for the same site, model and size.
+     */
+    private static final String INFERENCE_KEY_SEGMENT = ":inference";
```
wezell, line 106: It is not clear to me if these caches need to be invalidated across a cluster?  How do they load?  Does the user make changes to them in the UI?

dotCMS/src/main/java/com/dotcms/inference/rest/ChatCompletionsResource.java
```diff
@@ -0,0 +1,594 @@
...
+                BearerOnlyAuthFilter.bearerCredentialProblem(
+                        request.getHeader(javax.ws.rs.core.HttpHeaders.AUTHORIZATION));
+        if (credentialProblem.isPresent()) {
+            return errorResponse(credentialProblem.get());
+        }
+
+        // Any authenticated user, backend or frontend; an anonymous caller is rejected here with a
+        // 401 the builder produces itself.
+        final User user = new WebResource.InitBuilder(request, response)
+                .requiredBackendUser(true)
```
wezell, line 245: Is this check an `OR`?  It might be an `AND`,  requiring both a backend and front end role.

wezell (dismissed):
Left some comments

dotCMS/src/main/java/com/dotcms/inference/rest/ChatCompletionsResource.java
fmontes: It's an `OR`. Both builder calls add to the same `requiredRolesSet`, and `WebResource.checkRolePermissions` returns on the first role that matches:

```java
for (String roleKey : builder.requiredRolesSet) {
    if (doesUserHaveRole(user, roleKey)) return;   // any match passes
}
throw new SecurityException("... lacks one of the required role ...");
```

So this is "backend **or** frontend", which is the intent — any authenticated user, anonymous rejected. Fair question though: `requiredBackendUser(true).requiredFrontendUser(true)` reads like it should be an AND, and nothing at the call site tells you otherwise.

dotCMS/src/main/java/com/dotcms/ai/client/langchain4j/LangChain4jAIClient.java
fmontes: Good catch — they do need it, and today they don't get it.

**How they load:** lazily on first use. Caffeine, `maximumSize(128)`, 1-hour TTL, keyed by `host + ":" + providerConfigHash`.

**Does the user change them in the UI:** yes — dotAI App secrets in the Apps portlet. Saving fires `AppSecretSavedEvent`, and `AIAppListener` calls `flushCachesForHost`.

**Across a cluster: no.** `AppsAPIImpl` publishes with `localSystemEventsAPI.asyncNotify(...)`, which is node-local. Rotate a key on node A and node B keeps serving models built with the old credential until the TTL expires — up to an hour. Where the rotation is a response to a leaked key, the revoked key stays in use on those nodes for that window.

Two things worth noting on scope. It's pre-existing rather than something this PR introduces — all four caches are already in `main`, and the same local-only listener pattern is used by `AnalyticsAppListener` and `ContentAnalyticsAppListener`, so it's platform-wide. And this PR doesn't widen it: the `:inference` cache keys it adds are appended after the hostname, so the existing `startsWith(hostname + ":")` flush still evicts them.

Filed as #37623, scoped to all three subscribers rather than just the AI one.

dotcms-integration/src/test/java/com/dotcms/MainSuite2b.java
fmontes: Done in 9d4b62c9eb — moved to `MainSuite2a`.

The premise checks out: `.github/test-matrix.yml` gives each suite its own matrix job, so they run in parallel and the stage costs whatever the slowest one costs. Class counts:

| Suite | Before | After |
|---|---|---|
| MainSuite2b | 314 | 301 |
| MainSuite1b | 87 | 87 |
| MainSuite1a | 77 | 77 |
| MainSuite3a | 66 | 66 |
| MainSuite2a | 53 | 66 |

`InferenceTestsAreRegisteredTest` had to move with them — it names the suite it checks, so left pointing at `MainSuite2b` it would have reported all 13 as unregistered. Verified by removing a class from the suite and confirming it fails with the missing name.

One caveat: class count is a proxy for runtime, not a measurement. This moves 13 classes off the job most likely to be setting the pace and onto the smallest one, but I can't show `MainSuite2a` is the fastest job without CI timings. If you know which suite actually finishes first, happy to move them there instead — it's a one-line change.

wezell: @freddyDOTCMS please point the ai at a few of our PR merge runs to pull real timings so it can optimize here

dotCMS/src/main/java/com/dotcms/ai/client/langchain4j/LangChain4jAIClient.java
wezell: Maybe for this PR set Caffeine to expireAfter write to like 5m?  Or to refresh in the background every 5m?

dotcms-integration/src/test/java/com/dotcms/MainSuite2b.java
fmontes: Pulled real job durations from three successful PR Check runs (`31131753266`, `31130388103`, `31120501443`):

| Suite | Classes | Run 1 | Run 2 | Run 3 | Avg |
|---|---|---|---|---|---|
| MainSuite 2b | 314 | 38m28 | 36m50 | 41m00 | ~38.8m |
| MainSuite 1b | 87 | 41m31 | 37m20 | 36m44 | ~38.5m |
| MainSuite 1a | 77 | 38m12 | 36m57 | 34m31 | ~36.6m |
| MainSuite 3a | 66 | 31m18 | 31m25 | 32m30 | ~31.7m |
| MainSuite 2a | 53 | 33m31 | 21m22 | 28m09 | ~27.7m |
| Junit5 Suite 1 | 15 | 7m22 | 15m37 | 9m02 | ~10.7m |
| OpenSearch Upgrade | — | 10m05 | 11m09 | 10m21 | ~10.5m |

Two things the data says that the class counts did not.

**Class count is a poor proxy for runtime.** MainSuite 1b carries 87 classes and runs as long as 2b's 314. So 2b being the largest was not the reason it was worth moving off — though it is still the slowest, so the move helps.

**MainSuite 2a is the right target on the numbers, not just on the counts.** At ~27.7m it is the fastest MainSuite by ~4m over 3a and ~11m over 2b. The 13 inference classes run ~2.5m locally, which leaves 2a still comfortably off the critical path.

**The real ceiling is elsewhere.** 2b and 1b are effectively tied at ~38.5-38.8m and together set the integration stage's wall time, so shaving either one is what would actually speed up CI — the difference between the slowest and fastest MainSuite is ~11m of idle capacity. Junit5 Suite 1 finishes in ~10m; migrating suites to JUnit 5 looks like the bigger lever, though that is well outside this PR.

Happy to open an issue for rebalancing 2b/1b against these timings if that is useful.

dotCMS/src/main/java/com/dotcms/ai/client/langchain4j/LangChain4jAIClient.java
fmontes: Done — `expireAfterWrite` is now 5m on all four model caches (was 1h).

It is the right lever for this. It cannot make invalidation cluster-wide, but it bounds the part that actually hurts: how long a revoked credential stays usable on the nodes that did not handle the save. An hour was long enough to matter when the rotation is a response to a leak; five minutes is not.

Cost is low. `LangChain4jModelFactory` constructs a provider client rather than calling one, so a rebuild is local object construction — the price of the shorter life is object and connection-pool churn, not latency on the request that triggers the rebuild.

Two things worth flagging:

**This changes shipped behaviour, not just the new family.** `chatModelCache` and friends are shared — the legacy `/api/v1/ai/*` path uses them through `executeWithFallback` at the same time `/api/inference/v1` uses them through the typed accessors. So the older endpoints get the shorter TTL too. That seems right to me (the stale-key exposure was never specific to the new family), but it is a behaviour change outside this PR's nominal scope, so worth saying out loud. I am running the `com.dotcms.ai.**` integration tests alongside the inference ones for that reason.

**I went with `expireAfterWrite` rather than `refreshAfterWrite`.** Background refresh keeps the old model serving until the replacement is ready, which means a revoked key can still answer requests slightly past the window — the opposite of what we want from this particular knob. Given rebuilds are cheap and never block on the network, plain expiry is both simpler and stricter.

The reason for the number is recorded on the constant, so the next person to look at a 5-minute TTL on a cache of rarely-changing objects sees why it is small before shortening the rebuild churn by lengthening it again.

Cluster-wide invalidation is filed as #37623, scoped to all three `AppSecretSavedEvent` subscribers — `AIAppListener`, `AnalyticsAppListener` and `ContentAnalyticsAppListener` all have the same gap.

wezell: nice, thanks!

wezell (dismissed):
Let's go

wezell (approved):
let's go! (again)

